PR #5, opened 2022-08-05
Inject BASE_URL for SafeConfigService

### LINKED ISSUE #2: Add BaseUrl to SafeConfigService
The `BaseURL` is hardcoded for the `SafeConfigService`.

This value should be injected instead of being hardcoded. On an initial phase it doesn't need to be read from a configuration/environment variable

fmrsabino:
Closes #2 

- Add `BASE_URL_PROVIDER`, a provider that can be injected with `@Inject('SAFE_CONFIG_BASE_URL')` and has the value of the Safe Config service url

### DIFF
--- a/src/services/safe-config/safe-config.module.ts
+++ b/src/services/safe-config/safe-config.module.ts
@@ -3,7 +3,7 @@ import { Module } from '@nestjs/common';
 import { SafeConfigService } from './safe-config.service';
 
 const BASE_URL_PROVIDER = {
-  provide: 'BASE_URL',
+  provide: 'SAFE_CONFIG_BASE_URL',
   useValue: 'https://safe-config.gnosis.io', // TODO extract to a config file
 };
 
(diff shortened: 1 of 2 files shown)


src/services/safe-config/safe-config.module.ts
```diff
@@ -2,9 +2,14 @@ import { HttpModule } from '@nestjs/axios';
 import { Module } from '@nestjs/common';
 import { SafeConfigService } from './safe-config.service';
 
+const BASE_URL_PROVIDER = {
+  provide: 'BASE_URL',
```
hectorgomezv, line 6: This is module-scoped, right? Otherwise we could use a prefix, like CONFIG_SERVICE_BASE_URL :)

fmrsabino: Yeah good point. I actually would prefer to have it as just `BASE_URL_PROVIDER` but scoped only to this module. It should not be possible for this to be injected elsewhere directly.

fmrsabino: Done in [2364b76](https://github.com/5afe/safe-client-gateway-nest/pull/5/commits/2364b76f6a01f8ed1f74f44768a677a166dcfe0d)

hectorgomezv (approved):

fmrsabino:
@hectorgomezv had to rebase because of the new Github actions on `main`

PR #9, opened 2022-08-05
Add husky

fmrsabino:
- Adds Husky – https://typicode.github.io/husky/#/
- Upon installation these adds the following git hooks: 
  * `yarn run lint`
  * `yarn run format`

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
-    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
   "scripts": {
     "prebuild": "rimraf dist",
     "build": "nest build",
-    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
+    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\" --no-error-on-unmatched-pattern",
```
fmrsabino, line 11: I've added `--no-error-on-unmatched-pattern` here so that the commit would not throw an error if it fails matching some files

hectorgomezv (approved):

PR #11, opened 2022-08-08
Extract logic for http error mapping

### LINKED ISSUE #3: Extract logic for http error mapping
Currently we don't have a centralised place to handle errors coming from the Config/Transaction service. A common handler can be implemented to manage these errors and provide a common/normalised behaviour when they happen, so we can also provide a standard response to the clients. 

**Acceptance criteria**
- Logic for handling Config/Transaction service errors will be extracted to a common handler.
- The handler should provide a standard payload format to inform errors to the clients.
- Error codes and messages coming from Config/Transaction services should be provided to the clients.
- On unexpected/unknown error the service should return http 503 to the clients.

hectorgomezv:
Closes #3 

This PR implements a generic error handler which maps errors coming from http services (config service, transaction service) to a common `BaseError` interface.

The handler parses the incoming error and sets the appropriate status code and payload. If the incoming error has a status code, it's kept, otherwise a `502` status code is returned to the client.


src/services/errors/interfaces/base-error.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface BaseError {
+  message: string;
+  statusCode: number;
+  arguments: string[];
+}
```
fmrsabino, line 5: Is this based on `ErrorDetails` of the Safe Client Gateway? https://github.com/safe-global/safe-client-gateway/blob/d28bf1fb9feb069217ef52db1fde13bfb2b2aa58/src/utils/errors.rs#L30-L38

src/services/errors/interfaces/base-error.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface BaseError {
```
fmrsabino, line 1: nitpick: I'd probably pick a more concrete name for this one as `Base` seems to imply that all the other errors that we have derive from this one.

Since this seems to come from the Http side and it's only used as the payload for `HttpException` I'd either rename it to something like `HttpExceptionPayload` or maybe just a simple `Record` (and we remove this class).

What do you think?

fmrsabino (commented):
Nice 👏 

There are some lint warnings that need to be addressed.

src/services/errors/interfaces/http-service-error-response.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface HttpServiceErrorResponseData {
+  code: number;
+  message: string;
+  arguments: string[];
+}
```
fmrsabino, line 5: Similar to what I commented here https://github.com/5afe/safe-client-gateway-nest/pull/11/files#r940172896

Maybe we can remove the `export` on this one since this mapping should only be known to the mapper itself what do you think?

src/services/errors/interfaces/base-error.interface.ts
hectorgomezv: Not explicitly, I was reviewing some errors and did the mapping. Do you know what is this `debug: Option<Value>` property? Should we map it as well?

src/services/errors/interfaces/base-error.interface.ts
hectorgomezv: I was thinking in having a common error class for all errors in the CGW, and extend it for specific needs. But you're right, maybe we can start by having this payload just for HttpException and see if we can find an standard for all the error in the future (and inherit from it).

hectorgomezv: `HttpExceptionPayload` seems a good name to me 👍🏻

src/services/errors/interfaces/base-error.interface.ts
fmrsabino: I think we can start with this payload but instead of `statusCode` let's just have `code` to match what is serialised to the clients. We can get back to this if we need to.

I would not add the `debug` for now 👍

src/services/errors/interfaces/http-service-error-response.interface.ts
hectorgomezv: You mean having this interface into the `http-error-handler.ts` file? It's used on the unit tests as well :)

fmrsabino: Ah ok! I need to review how Typescript handles modules in general. For me this makes sense to be included here and in the tests but not in other modules.

hectorgomezv: I agree, I think TS/JS is more permissive than other languages (like Java) with importing/exporting things from different places in the code. A lot of things fall in developer's responsibility and good practices, which sometimes is not so good 😄

hectorgomezv: @fmrsabino I removed this class because since the last refactor it's redundant with `HttpExceptionPayload` 🙂

fmrsabino (approved):
🚀

PR #13, opened 2022-08-08
Add some VSCode debug configurations

hectorgomezv:
This add a first version of some custom Visual Studio Code debug configurations.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,54 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "node",
+      "request": "launch",
+      "name": "Launch Program",
+      "skipFiles": [
+          "<node_internals>/**"
+      ],
+      "program": "${workspaceFolder}/src/main.ts",
+      "preLaunchTask": "tsc: build - tsconfig.json",
+      "outFiles": [
+          "${workspaceFolder}/dist/**/*.js"
+      ]
+  },
+  {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug tests",
+      "skipFiles": [
+          "<node_internals>/**"
+      ],
+      "runtimeArgs": [
+        "--inspect-brk",
+        "${workspaceRoot}/node_modules/.bin/jest",
+        "--runInBand",
+        "--coverage",
+        "false"
+      ],
+      "console": "integratedTerminal"
+  },
+  {
+      "type": "node",
+      "request": "launch",
+      "name": "Debug current test file",
+      "skipFiles": [
+          "<node_internals>/**"
+      ],
+      "runtimeArgs": [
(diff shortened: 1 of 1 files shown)


fmrsabino (commented):
We have an `e2e` test in the `main` branch that is not compiling right now (because we were fixing this on a separate task).

But it seems that the `Launch Program` task is not the same as `yarn run start:dev` which currently executes successfully 🤔 Why is that?

hectorgomezv:
> We have an `e2e` test in the `main` branch that is not compiling right now (because we were fixing this on a separate task).
> 
> But it seems that the `Launch Program` task is not the same as `yarn run start:dev` which currently executes successfully 🤔 Why is that?

These configuration are not using any package manager or nest binary to start, they're meant to be used for debugging, so I think we can just transpile the TS code and launch using node. The `yarn run dev:start` script watches for changes in the code, I'm not sure this is compatible with the VSCode debugger... (and the debugger of course can also restart after any change in the code). Maybe I'm not getting the question... :)

fmrsabino:
> > We have an `e2e` test in the `main` branch that is not compiling right now (because we were fixing this on a separate task).
> > But it seems that the `Launch Program` task is not the same as `yarn run start:dev` which currently executes successfully 🤔 Why is that?
> 
> These configurations are not using any package manager or nest binary to start, they're meant to be used for debugging, so I think we can just transpile the TS code and launch using node. The `yarn run dev:start` script watches for changes in the code, I'm not sure this is compatible with the VSCode debugger... (and the debugger of course can also restart after any change in the code). Maybe I'm not getting the question... :)

If I use `Launch Program` what does it actually do?

With the package manager we have `yarn run start:dev` but if we ignore the package manager our application is launched with `nest start --watch`. But I don't see this in the PR 🤔 

It seems to invoke `main.ts`. Is this the same as launching it with `nest`?

hectorgomezv:
> > > We have an `e2e` test in the `main` branch that is not compiling right now (because we were fixing this on a separate task).
> > But it seems that the `Launch Program` task is not the same as `yarn run start:dev` which currently executes successfully 🤔 Why is that?
> 
> These configurations are not using any package manager or nest binary to start, they're meant to be used for debugging, so I think we can just transpile the TS code and launch using node. The `yarn run dev:start` script watches for changes in the code, I'm not sure this is compatible with the VSCode debugger... (and the debugger of course can also restart after any change in the code). Maybe I'm not getting the question... :)

If I use `Launch Program` what does it actually do?

With the package manager we have `yarn run start:dev` but if we ignore the package manager our application is launched with `nest start --watch`. But I don't see this in the PR 🤔 

It seems to invoke `main.ts`. Is this the same as launching it with `nest`? 

Yes, it does the same thing under the hood. 
- Transpiles TS code to JS code: creates both *.js files and *.js.map files (for debugging)
- Executes the code in the NodeJS V8 runtime as it would be executed in a production server (JS code).

It also attaches a debugger to the Node.js runtime, so you can do all the things you expect from a debugger: create breakpoints, examine the memory contents, execute random code in a debug console, etc.

This configuration is intended for debugging in the developers machine, you always can execute the Nest server (with hot reload) from the terminal.

fmrsabino (approved):

PR #14, opened 2022-08-09
Add managed API versioning

### LINKED ISSUE #4: Use automatic API versioning
**Summary**
We could take advantage of Nest automatic API versioning feature. By using this, route prefixing (`v1`, `v2`, etc,) is handled automatically by the framework, by prepending the adequate `v` to the URLs [this way](https://docs.nestjs.com/techniques/versioning#uri-versioning-type).

After configuring this, we could simply pass a `{ version: 'X' }` configuration object to [each controller](https://docs.nestjs.com/techniques/versioning#controller-versions), or even to each [particular route](https://docs.nestjs.com/techniques/versioning#route-versions). 

This feature is pretty flexible, it can apply multiple versions to each controller/route, and it also allows "version-neutral" controller/routes.

hectorgomezv:
Closes #4 

This PR implements managed API versioning feature from Nest. Using this feature we could have an explicit and clear route mapping. I used `@Controller` versioning only, but route-level versioning is available as well.

### DIFF
--- a/src/chains/chains.controller.ts
+++ b/src/chains/chains.controller.ts
@@ -1,9 +1,12 @@
-import { Controller, Get } from '@nestjs/common';
+import { Controller, Get, VERSION_NEUTRAL } from '@nestjs/common';
 import { ChainsService } from './chains.service';
 import { Chain } from './entities/chain.entity';
 import { Page } from './entities/page.entity';
 
-@Controller('chains')
+@Controller({
+  path: 'chains',
+  version: VERSION_NEUTRAL,
+})
 export class ChainsController {
   constructor(private readonly chainsService: ChainsService) {}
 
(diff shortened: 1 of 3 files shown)


src/chains/chains.controller.ts
```diff
@@ -1,9 +1,12 @@
-import { Controller, Get } from '@nestjs/common';
+import { Controller, Get, VERSION_NEUTRAL } from '@nestjs/common';
 import { ChainsService } from './chains.service';
 import { Chain } from './entities/chain.entity';
 import { Page } from './entities/page.entity';
 
-@Controller('chains')
+@Controller({
+  path: 'chains',
+  version: VERSION_NEUTRAL,
```
fmrsabino, line 8: This will change because in production it's under `v1` too. You can either change it here or leave this one as is and we tackle in a different one.

fmrsabino (approved):

src/chains/chains.controller.ts
hectorgomezv: Oh, right, you used the root just for testing I guess, I'll move it to the `/v1` path in this PR. At least we know this is working fine with no version-prefixed endpoints 👍

fmrsabino: Yeah that's right! Thanks Hector!

fmrsabino (approved):

PR #15, opened 2022-08-09
Add unit test for Balances Controller

fmrsabino:
- Adds a unit test to the Balances Controller
- This unit test considers a unit to be the `BalancesModule` – ie.: it will use the real component implementation except for the provided HttpClient (which is mocked)


src/balances/balances.controller.spec.ts
```diff
@@ -1,22 +1,68 @@
...
+      data: [{ tokenAddress: '0x000000', balance: '1' }],
+    };
+    const safeConfigServiceResponse = {
+      data: {
+        chainId: '1',
+        chainName: 'mainnet',
+        transactionService: 'https://test.safe.transaction.service',
+        vpcTransactionService: 'https://test.safe.transaction.service',
+      },
+    };
```
fmrsabino, line 35: We can probably extract this into a separate folder that contains mocked responses/json.

We can do this as soon as we start testing multiple scenarios 👍

hectorgomezv: Yes, cool. I have mixed opinions about this to be honest, could be good to have a folder for fixtures but I think we need to be cautious with that as it could increase coupling between test suites (from [my fav testing best practices repository :)](https://github.com/goldbergyoni/javascript-testing-best-practices#-%EF%B8%8F16-dont-foo-use-realistic-input-data))

hectorgomezv: I also think it could be useful to generate test data using a tool which generates different inputs in each test execution, like [Faker](https://www.npmjs.com/package/@faker-js/faker) (it even has an [Eth address generator](https://fakerjs.dev/api/finance.html#ethereumAddress) 😄)

src/balances/balances.controller.spec.ts
```diff
@@ -1,22 +1,68 @@
...
+import * as request from 'supertest';
+import axios from 'axios';
+import { BalancesModule } from './balances.module';
 
-describe('BalancesController', () => {
-  let controller: BalancesController;
+jest.mock('axios');
+const axiosMock = axios as jest.Mocked<typeof axios>;
+
+describe('AppController (e2e)', () => {
```
hectorgomezv, line 10: Nitpick: maybe a more descriptive name could be better? Like `BalancesController` or something similar you like.

hectorgomezv (approved):

src/balances/balances.controller.spec.ts
fmrsabino: Regarding Faker we do use it on our Django services and I was thinking about introducing it but in another PR!

src/balances/balances.controller.spec.ts
fmrsabino: Changed it here [17d55ed](https://github.com/5afe/safe-client-gateway-nest/pull/15/commits/17d55ede8202804e95b3c194ea9f496d29220a8d)

hectorgomezv (approved):
Great work 🚀

PR #22, opened 2022-08-10
Add logger

hectorgomezv:
This PR changes the usage on `console.log` to the Nest.js default logging feature.

I did a brief research of the options we have, and both [Winston](https://github.com/gremo/nest-winston) and [Pino](https://www.npmjs.com/package/nestjs-pino) Nest integrations seems great, but not sure the features they offer are valuable for us right now, maybe Pino could be used in the future since it's very performant, but in my opinion it would be a premature optimisation at this point.

@fmrsabino What do you think about logging incoming http requests? I think it's pretty useful sometimes and, if I'm not wrong, old client gateway implementation uses that. We could achieve this easily by applying a [middleware](https://docs.nestjs.com/middleware) to the modules we want. If you find it useful I could include it in this PR (or create a new issue).

### DIFF
--- a/src/services/errors/http-error-handler.ts
+++ b/src/services/errors/http-error-handler.ts
@@ -1,4 +1,4 @@
-import { HttpException, HttpStatus, Injectable } from '@nestjs/common';
+import { HttpException, HttpStatus, Injectable, Logger } from '@nestjs/common';
 import axios, { AxiosError } from 'axios';
 import { HttpExceptionPayload } from './interfaces/http-exception-payload.interface';
 
@@ -9,6 +9,8 @@ import { HttpExceptionPayload } from './interfaces/http-exception-payload.interf
  */
 @Injectable()
 export class HttpErrorHandler {
+  private readonly logger = new Logger(HttpErrorHandler.name);
+
   private mapError(err: AxiosError | Error): HttpExceptionPayload {
     if (axios.isAxiosError(err) && err.response) {
       const axiosError = err as AxiosError;
@@ -27,7 +29,8 @@ export class HttpErrorHandler {
   }
 
   handle(err: AxiosError | Error) {
-    const error = this.mapError(err);
-    throw new HttpException(error, error.code);
+    const errPayload = this.mapError(err);
+    this.logger.error(errPayload);
+    throw new HttpException(errPayload, errPayload.code);
   }
 }
--- a/src/services/safe-config/safe-config.service.ts
+++ b/src/services/safe-config/safe-config.service.ts
@@ -2,12 +2,14 @@ import { HttpService } from '@nestjs/axios';
 import { Inject, Injectable } from '@nestjs/common';
 import { SafeConfigPage } from './entities/page.entity';
 import { SafeConfigChain } from './entities/chain.entity';
+import { HttpErrorHandler } from '../errors/http-error-handler';
 
 @Injectable()
 export class SafeConfigService {
   constructor(
     @Inject('SAFE_CONFIG_BASE_URL') private readonly baseUrl,
     private readonly httpService: HttpService,
+    private readonly httpErrorHandler: HttpErrorHandler,
   ) {}
 
   async getChains(): Promise<SafeConfigPage<SafeConfigChain>> {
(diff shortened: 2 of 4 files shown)


src/services/safe-config/safe-config.service.ts
```diff
@@ -2,22 +2,23 @@ import { HttpService } from '@nestjs/axios';
...
   async getChains(): Promise<SafeConfigPage<SafeConfigChain>> {
     try {
       const url = this.baseUrl + '/api/v1/chains';
       const response = await this.httpService.axiosRef.get(url);
       return response.data;
-    } catch (error) {
-      console.log(error);
-      throw error;
+    } catch (err) {
+      this.httpErrorHandler.handle(err);
```
katspaugh, line 21: This pattern will probably repeat in all GET endpoints, maybe it's worth making a helper function like
```js
getData<T>(path: string): Promise<T> {
  try {
     const url = this.baseUrl + path
     const response = await this.httpService.axiosRef.get(url);
     return response.data;
   } catch (err) {
     this.httpErrorHandler.handle(err);
   }
}
```

src/services/errors/http-error-handler.ts
```diff
@@ -27,7 +29,8 @@ export class HttpErrorHandler {
   }
 
   handle(err: AxiosError | Error) {
-    const error = this.mapError(err);
-    throw new HttpException(error, error.code);
+    const errPayload = this.mapError(err);
+    this.logger.error(errPayload);
+    throw new HttpException(errPayload, errPayload.code);
```
katspaugh, line 34: Question: where does this `errPayload.code` come from? In case of the chains endpoint, from the config-service?

Maybe it would make sense to have a centralized error registry where you would create your own errors with distinct CGW-only error codes, and send that to the client.

src/services/safe-config/safe-config.service.ts
hectorgomezv: I agree, I think it could be useful to have a separate layer in between `Service` and `HttpService` (something like `HttpClient`) which encapsulates this kind of functionality. So we can just call the `HttpClient` from any `Service` and reuse it. Maybe it's too much for this PR as it's just for logging, but I'd create an issue if @fmrsabino agrees 🙂

src/services/errors/http-error-handler.ts
hectorgomezv: At this point the error could come from any http service we're calling (both Config and Transaction services). I think having a centralised registry it's a good idea, probably this will be done in a further development step, we're keeping it simple for now but we'll do it. In fact this explicit CGW-only error codes is something you already have in place in the old CGW implementation, right?

src/services/safe-config/safe-config.service.ts
katspaugh: Sure, it wasn't meant for this PR. Also you'll probably see larger patterns emerging eventually. I'll resolve this.

src/services/errors/http-error-handler.ts
katspaugh: AFAIK, we only received 42 and 1337 error codes in the old CGW.

src/services/safe-config/safe-config.service.ts
fmrsabino: I would not try to extract helpers at this stage. Let's see how things evolve first. It's still in the early stages.

@katspaugh the helper function that you proposed doesn't have any query parameters in place which we already use. Even thought it is an easy change that's one of the reasons why I prefer to not have helper functions right now.

src/services/errors/http-error-handler.ts
katspaugh: The only use case I know so far is when we need to check if a safe exists or just failed to load.

But obviously, if we had a registry of backend error codes with explanations, we could potentially handle errors better on the frontend end. Or even display them to the user directly.

fmrsabino: This is something that we can (and should) review alongside the clients' teams. 

If the clients' use the current error codes then we will match them on an initial phase so that there are no surprises on your end. Either way as you mentioned, we should align this either for the initial release or one that comes after that 👍

fmrsabino (approved):
Nice 👏 

As you suggested we can also have a look into https://github.com/winstonjs/winston for our production needs. Seems to also be a [recommendation from NestJS](https://docs.nestjs.com/techniques/logger#use-external-logger).

PR #25, opened 2022-08-12
Fix invalid imports

hectorgomezv:
This PR fixes a couple of invalid import references which probably were introduced by an IDE automatic import.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,3 +1,4 @@
 {
-  "editor.defaultFormatter": "esbenp.prettier-vscode"
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "typescript.preferences.importModuleSpecifier": "relative"
 }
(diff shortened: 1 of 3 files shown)


hectorgomezv:
> 👏 I might have fixed some of those in my current branch for the Balances but I also prefer to have a smaller PR fixing it!
> 
> Do you know if there's any way to force relative imports? (with the IDE)

I found a way to do it, I included it in `.vscode/settings`, I think this could be uploaded to the repo, wdyt? 🙂

fmrsabino:
> > 👏 I might have fixed some of those in my current branch for the Balances but I also prefer to have a smaller PR fixing it!
> > Do you know if there's any way to force relative imports? (with the IDE)
> 
> I found a way to do it, I included it in `.vscode/settings`, I think this could be uploaded to the repo, wdyt? 🙂

Awesome! Yeah let's do it 👍

fmrsabino (approved):

PR #28, opened 2022-08-12
Rename SafeConfig* to Config*

fmrsabino:
- Renames `safe-config` folder to `config-service`
- Renames `SafeConfigModule` to `ConfigServiceModule`
- Renames `SafeConfigService` to `ConfigService`
- Renames `SafeConfigChain` to `Chain` (if the name conflicts in a feature, import aliases should be used)
- Renames `SafeConfigPage` to `Page` (if the name conflicts in a feature, import aliases should be used)

### DIFF
--- a/src/services/config-service/entities/chain.entity.ts
+++ b/src/services/config-service/entities/chain.entity.ts
@@ -1,4 +1,4 @@
-export interface SafeConfigChain {
+export interface Chain {
   chainId: string;
   chainName: string;
   transactionService: string;
--- a/src/chains/chains.controller.spec.ts
+++ b/src/chains/chains.controller.spec.ts
@@ -1,5 +1,5 @@
 import { Test, TestingModule } from '@nestjs/testing';
-import { SafeConfigModule } from '../services/safe-config/safe-config.module';
+import { ConfigServiceModule } from '../services/config-service/config-service.module';
 import { ChainsController } from './chains.controller';
 import { ChainsService } from './chains.service';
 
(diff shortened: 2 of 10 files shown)


src/services/config-service/entities/chain.entity.ts
```diff
@@ -1,4 +1,4 @@
-export interface SafeConfigChain {
```
hectorgomezv, line 1: Both `Page` and `Chain` have exact duplicates under `src/chains/entities`. I was thinking about we could use those interfaces directly... from a DDD/SOLID perspective I think it's not incorrect to make use of _domain_ classes into this kind of _adapters_ layer. And if we found any use case which forces us to have differences between them, we could create a DTO in the _adapters_ layer.

hectorgomezv: For `Page`, if it's a common interface, maybe we could extract it to a `src/shared` folder (and remove it from `src/chains`), what do you think?

fmrsabino: I agree. In this case however `src/chains` was created initially as a demo. Its `Page` should actually be more like the following: https://github.com/safe-global/safe-client-gateway/blob/11ce5e74a8bd670e91fe547e37eb71afe5b835b6/src/common/models/page.rs#L5-L12

The concept of `Page` would actually need to be used across every feature. For this I'd create an entity under a new folder (like `common`).

If `Chain` ends up being the same object then we can move it to this `/common` folder that it can be used by both the `ConfigService` and the features.

hectorgomezv (approved):

Issue #19, opened 2022-08-10
Implement /safes/<safe_address>/balances/<fiat> route


fmrsabino:
- Implement `/v1/chains/<chain_id>/safes/<safe_address>/balances/<fiat>?<trusted>&<exclude_spam>` with the same schema as the current Safe Client Gateway.
- Caching is not required for this ticket
- Unit tests should be added for this route

fmrsabino:
Closed with #29 and #31

PR #29, opened 2022-08-15
Add /balances/:fiatCode endpoint

fmrsabino:
Related with https://github.com/5afe/safe-client-gateway-nest/issues/19 – I'd close it once this PR https://github.com/5afe/safe-client-gateway-nest/pull/31 is merged (it has more tests and includes Faker)

- Adds `@nestjs/config` – this dependency allows us to introduce an application level configuration which reads from the local environment
- Adds an `ExchangeService` which depends on an external API to retrieve the fiat conversion prices (which is used by the `/balances` endpoint)
- Introduce `.env.sample` – this file should be copied to `.env` if a local configuration should be changed. `.env` will be automatically read by `configuration.ts`

@hectorgomezv here I just adjusted the test that we had already in place. Since the payload is already increasing in size I'll have a look into `Faker` and ways to better associate the request with the response in the tests (possibly in a separate PR)


fmrsabino:
@hectorgomezv the idea of providing Fake data would be similar to this: https://github.com/5afe/safe-client-gateway-nest/blob/add-fake-data/src/services/transaction-service/entities/__tests__/balance.factory.ts

src/balances/balances.service.ts
```diff
@@ -1,16 +1,89 @@
...
+    fiatCode: string,
+  ): Promise<Balances> {
+    const safeTransactionService =
+      await this.safeTransactionManager.getTransactionService(chainId);
+    const txServiceBalances: TransactionServiceBalance[] =
+      await safeTransactionService.getBalances(safeAddress);
+
+    const usdToFiatRate: number = await this.exchangeService.convertRates(
+      fiatCode,
+      'USD',
```
katspaugh, line 32: I would extract this string into a constant somewhere on top of this file.

src/services/exchange/exchange.service.ts
```diff
@@ -0,0 +1,49 @@
...
+    const fromExchangeRate = exchangeResult.rates[from];
+    if (fromExchangeRate === undefined)
+      throw new InternalServerErrorException(
+        `Exchange rate for ${from} is not available`,
+      );
+    const toExchangeRate = exchangeResult.rates[to];
+    if (toExchangeRate === undefined)
+      throw new InternalServerErrorException(
+        `Exchange rate for ${to} is not available`,
+      );
```
fmrsabino, line 31: This can probably be a 4XX – a `404` might be adequate since the fiat code is in the path. We can optionally return the rates available for conversion to the clients.

However this assumes that we do not default to `0` in case we have an error (separate ongoing discussion).

src/balances/balances.service.ts
fmrsabino: Extracted here: [e3deb67](https://github.com/5afe/safe-client-gateway-nest/pull/29/commits/e3deb6736c1038922b1ba19d8aa892c305eb6bcf)

hectorgomezv:
> Related with #19 – I'd close it once this PR #31 is merged (it has more tests and includes Faker)
> 
> * Adds `@nestjs/config` – this dependency allows us to introduce an application level configuration which reads from the local environment
> * Adds an `ExchangeService` which depends on an external API to retrieve the fiat conversion prices (which is used by the `/balances` endpoint)
> * Introduce `.env.sample` – this file should be copied to `.env` if a local configuration should be changed. `.env` will be automatically read by `configuration.ts`
> 
> @hectorgomezv here I just adjusted the test that we had already in place. Since the payload is already increasing in size I'll have a look into `Faker` and ways to better associate the request with the response in the tests (possibly in a separate PR)

Great, thanks! Yes, for `faker` we have [this issue](https://github.com/5afe/safe-client-gateway-nest/issues/24), I'll try to take it soon 🙂

src/config/configuration.ts
```diff
@@ -0,0 +1,8 @@
+export default () => ({
+  exchange: {
+    baseUri:
+      process.env.EXCHANGE_API_BASE_URI ||
```
hectorgomezv, line 4: This is unrelated to this PR, but as a personal preference I'd say 80 chars per line is too short for JS, I'd increase it to 100, 110 or even 120.

fmrsabino: Yes I agree. I haven't change the defaults so far but we can review the number of columns 👍

fmrsabino:
> > Related with #19 – I'd close it once this PR #31 is merged (it has more tests and includes Faker)
> > 
> > * Adds `@nestjs/config` – this dependency allows us to introduce an application level configuration which reads from the local environment
> > * Adds an `ExchangeService` which depends on an external API to retrieve the fiat conversion prices (which is used by the `/balances` endpoint)
> > * Introduce `.env.sample` – this file should be copied to `.env` if a local configuration should be changed. `.env` will be automatically read by `configuration.ts`
> > 
> > @hectorgomezv here I just adjusted the test that we had already in place. Since the payload is already increasing in size I'll have a look into `Faker` and ways to better associate the request with the response in the tests (possibly in a separate PR)
> 
> Great, thanks! Yes, for `faker` we have [this issue](https://github.com/5afe/safe-client-gateway-nest/issues/24), I'll try to take it soon 🙂

Maybe I can take that one since I introduced it in #31

src/services/exchange/exchange.service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+    return toExchangeRate / fromExchangeRate;
```
hectorgomezv, line 33: This could lead to divisions by zero? I know having a 0 rate value shouldn't happen, do you think is too much defensive programming to consider that case. 

Apart from that, maybe an unit this for this function could help visualize the cases (this is checking for `undefined`, could the `from/to` value be null?) and refactor. (I can do it in a separate PR, no problem 😄)

fmrsabino: For the test cases I can consider it for https://github.com/5afe/safe-client-gateway-nest/pull/31 (this PR was already big).

For the implementation part I'm going to double check the current implementation regarding this! 👍 Good call. For the implementation we can tackle in this PR.

Ideally I'd have feature+tests here but I was playing a little bit with Faker due to the amount of instances that we were creating for the tests so I decided to do it separately.

hectorgomezv (approved):
Great job! 😀

src/balances/entities/balances.entity.ts
```diff
@@ -0,0 +1,6 @@
+import { Balance } from './balance.entity';
+
+export interface Balances {
+  fiatTotal: number;
+  items: [Balance];
```
usame-algan, line 5: This is an interesting syntax. Is it the same as `Balance[]` or `Array<Balance>`?

fmrsabino: > This is an interesting syntax. Is it the same as `Balance[]` or `Array<Balance>`?

Yeah good point! I've been writing with `T[]` but I might have missed this one. This seems to be a `tuple` type. 

However with tuple types the size should be known beforehand so this would mean a tuple of size 1 which corresponds to a `Balance` (but it actually outputs a bigger collection 🤔 ). Do you know why this might be the case?

I will change it.

fmrsabino: Changed it here [183e0a6](https://github.com/5afe/safe-client-gateway-nest/pull/29/commits/183e0a6448be4b0967d87b95c01e9dde0445f6ee)

katspaugh: > Do you know why this might be the case?

You mean why it's not catching that the actual array is bigger than 1? Because the types are static and it cannot see what's actually inserted into this array in runtime unless you're literally pushing things one by one.

fmrsabino: > > Do you know why this might be the case?
> 
> You mean why it's not catching that the actual array is bigger than 1? Because the types are static and it cannot see what's actually inserted into this array in runtime unless you're literally pushing things one by one.

I took another look and it makes more sense now since we are setting the tuple to a `Balance[]` which is compatible. If later we try to access the second index (`1`) then we get this warning `TS2493: Tuple type '[Balance]' of length '1' has no element at index '1'.`

src/services/exchange/exchange.service.ts
fmrsabino: > For the implementation part I'm going to double check the current implementation regarding this! 👍 Good call. For the implementation we can tackle in this PR.

@hectorgomezv For the current implementation it seems that code might `panic` and return a `500` https://github.com/safe-global/safe-client-gateway/blob/dd162f73150e8de4ea6eda481da1ee802692bdbe/src/providers/fiat.rs#L47

I will add a condition to this to have a better error message from the service.

PR #30, opened 2022-08-16
Implement /chains/:chain_id/about/backbone

hectorgomezv:
This PR implements the retrieval of the _backbone_ metadata from the transaction service.


src/services/transaction-service/transaction-service.service.ts
```diff
@@ -14,12 +15,22 @@ export class TransactionService {
...
   ): Promise<Balance[]> {
-    const url = this.baseUrl + `/api/v1/safes/${safeAddress}/balances/usd/`;
+    const url = `${this.baseUrl}/api/v1/safes/${safeAddress}/balances/usd/`;
     try {
-      const { data } = await this.httpService.axiosRef.get(url, {
+      const res = await this.httpService.axiosRef.get(url, {
         params: { trusted: trusted, excludeSpam: excludeSpam },
       });
-      return data;
+      return res?.data;
```
usame-algan, line 23: Is there a way to deal with `res` having `any` type? It looks like this function could return `undefined` but it is not reflected in the signature.

hectorgomezv: The function shouldn't return `undefined`. In fact if `res.data` is not present it should throw an error, I'll back to the old implementation, thanks!

src/chains/entities/index.ts
```diff
@@ -0,0 +1,3 @@
+export * from './backbone.entity';
+export * from './chain.entity';
+export * from './page.entity';
```
fmrsabino, line 3: What is the benefit of doing it like this?

src/chains/chains.controller.spec.ts
```diff
@@ -1,22 +1,60 @@
...
-      imports: [ConfigServiceModule],
       controllers: [ChainsController],
-      providers: [ChainsService],
+      providers: [
+        {
+          provide: ChainsService,
+          useValue: {
+            getChains: jest.fn().mockResolvedValue(chainsResponse),
+            getBackbone: jest.fn().mockResolvedValue(backboneResponse),
+          },
```
fmrsabino, line 42: Here we can use the real `ChainsService`

You'd still need to mock `axios` unless you want to isolate the `ChainsController`.

src/chains/chains.service.spec.ts
```diff
@@ -1,20 +1,43 @@
...
+
+const BACKBONE: Backbone = {
+  name: 'testName',
+  version: '',
+  api_version: '',
+  secure: false,
+  host: '',
+  headers: [],
+  settings: undefined,
+};
```
fmrsabino, line 15: nitpick: we can consider adding a factory function or similar for this one 😄

fmrsabino (commented):
Nice 👏 🚀

src/chains/entities/index.ts
hectorgomezv: I'd say it's more concise to import the entities in multiple places this way:
```javascript
import { Backbone, Chain, Page } from './entities';
```
Another way to do it would be managing the exported interfaces:
```javascript
export { Backbone } from './backbone.entity';
export { Chain } from './chain.entity';
export { Page } from './page.entity';
```
I don't have a strong opinion on wether using `*` or not, it depends of the level of control you want to have over the exports. But doing it the manual way implies you'll need to edit the index if you plan to export several interfaces from the same file.

fmrsabino: Alright 👍 I'm going to comment on where we can use this in this PR (I see that some imports can then be shortened).

src/chains/chains.service.spec.ts
```diff
@@ -1,20 +1,43 @@
-import { Test, TestingModule } from '@nestjs/testing';
-import { ConfigServiceModule } from '../services/config-service/config-service.module';
+import { ConfigService } from '../services/config-service/config-service.service';
+import { TransactionServiceManager } from '../services/transaction-service/transaction-service.manager';
+import { TransactionService } from '../services/transaction-service/transaction-service.service';
 import { ChainsService } from './chains.service';
+import { Backbone } from './entities/backbone.entity';
```
fmrsabino, line 5: nitpick: This import can be shortened.

src/services/transaction-service/transaction-service.service.spec.ts
```diff
@@ -2,14 +2,25 @@ import { HttpService } from '@nestjs/axios';
 import { Balance } from './entities/balance.entity';
 import { HttpErrorHandler } from '../errors/http-error-handler';
 import { TransactionService } from './transaction-service.service';
+import { Backbone } from '../../chains/entities/backbone.entity';
```
fmrsabino, line 5: nitpick: This import can be shortened.

src/services/transaction-service/transaction-service.service.ts
```diff
@@ -1,6 +1,7 @@
 import { HttpService } from '@nestjs/axios';
 import { Balance } from './entities/balance.entity';
 import { HttpErrorHandler } from '../errors/http-error-handler';
+import { Backbone } from '../../chains/entities/backbone.entity';
```
fmrsabino, line 4: nitpick: This import can be shortened.

src/chains/chains.controller.ts
```diff
@@ -1,5 +1,6 @@
-import { Controller, Get } from '@nestjs/common';
+import { Controller, Get, Param } from '@nestjs/common';
 import { ChainsService } from './chains.service';
+import { Backbone } from './entities/backbone.entity';
 import { Chain } from './entities/chain.entity';
 import { Page } from './entities/page.entity';
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [b85875d](b85875df2ce3a7b0bcad7eed23e1f3e7a0d83f0a)

src/chains/chains.service.spec.ts
```diff
@@ -1,20 +1,43 @@
...
+  } as unknown as TransactionServiceManager;
+
+  const service: ChainsService = new ChainsService(
+    configService,
+    transactionManager,
+  );
+
+  it('should retrieve the backbone metadata from the proper TransactionService', async () => {
+    const chainId = '1';
+    const backbone = await service.getBackbone(chainId);
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [ecfab32e](https://github.com/5afe/safe-client-gateway-nest/commit/ecfab32e5a912764b5e65f73223185affd32e04c)

src/chains/chains.service.spec.ts
hectorgomezv: Fixed in [857b5ab](https://github.com/5afe/safe-client-gateway-nest/pull/30/commits/857b5ab863510c87140f573ba4405527c1914975)

src/services/transaction-service/transaction-service.service.spec.ts
hectorgomezv: Fixed in [857b5ab](https://github.com/5afe/safe-client-gateway-nest/pull/30/commits/857b5ab863510c87140f573ba4405527c1914975)

src/services/transaction-service/transaction-service.service.ts
hectorgomezv: Fixed in [857b5ab](https://github.com/5afe/safe-client-gateway-nest/pull/30/commits/857b5ab863510c87140f573ba4405527c1914975)

src/chains/chains.controller.spec.ts
hectorgomezv: @fmrsabino 
I was thinking in these tests as _controller tests_ in isolation, but later I reviewed what you did for `balances.controller.spec.ts` and maybe makes sense to have this `*.controller.spec.ts` as some kind of _almost-e2e-tests_ where nothing is mocked except the external service's data provider in the most outer layer.

I think it's ok by now, but I have to admit that I have some concerns about how this will evolve in time. I mean, we're coding something like the external service will be called one time with this arguments, later it will be called with these other arguments, later another call which would return this thing, etc. etc. Maybe it's too tied to the concrete implementation we have now, and if for any reason the service logic changes these tests would be hard to maintain. 

Plus, if a payload coming from an external service changes, our tests will be passing anyway.

I'm wondering if having isolated unit tests _per layer_, and real _e2e tests_ (where nothing is mocked, even the external services) would be more maintainable. With the real e2e tests you only have to code the expected result the client would receive, and coupling between tests and implementation is near 0.

(Sorry for the long comment) 😅

fmrsabino: > @fmrsabino I was thinking in these tests as _controller tests_ in isolation, but later I reviewed what you did for `balances.controller.spec.ts` and maybe makes sense to have this `*.controller.spec.ts` as some kind of _almost-e2e-tests_ where nothing is mocked except the external service's data provider in the most outer layer.
> 
> I think it's ok by now, but I have to admit that I have some concerns about how this will evolve in time. I mean, we're coding something like the external service will be called one time with this arguments, later it will be called with these other arguments, later another call which would return this thing, etc. etc. Maybe it's too tied to the concrete implementation we have now, and if for any reason the service logic changes these tests would be hard to maintain.
> 
> Plus, if a payload coming from an external service changes, our tests will be passing anyway.

The idea was to Unit test the the endpoint without having any external dependencies. The mocks on the HTTP level are required just because we need some "fake" http service where we say that if X is requested return Y. Besides that there's nothing that ties to the implementation.

If we isolate the test (each Unit test) then it is more tied to the implementation because we need to mock each dependency and if the implementation changes we also need to change the test.

I'm aware that mocking the Axios client can be verbose and we can think of ways to improve it but I think these tests are valuable in a way that they treat the actual implementation as a black box where we just need to write the endpoint that we are hitting, the responses of the external services would be and the expected response. If change the implementation the test should still pass (unless we change how the external services usage).

> I'm wondering if having isolated unit tests _per layer_, and real _e2e tests_ (where nothing is mocked, even the external services) would be more maintainable. With the real e2e tests you only have to code the expected result the client would receive, and coupling between tests and implementation is near 0.

My approach with the tests in `*.controller.spec.ts` does not mean that we should not have tests that test components in isolation. I find those tests still really valuable 👍 

For real E2E tests it can be more complex... specially when creating transactions and submitting transactions. We need to have some kind of test database or similar infrastructure to run this. Even though they provide a lot of value, some flows can be tricky to test.

hectorgomezv: > If we isolate the test (each Unit test) then it is more tied to the implementation because we need to mock each dependency and if the implementation changes we also need to change the test.

I partially agree, I think too much spying would led to coupling between tests and implementation, but by just mocking the dependency you're just tied to its interface, not its internal implementation. By mocking axios, which is an internal detail of the dependency (in the future the dependency could even use a different http client) you're tied to the actual implementation of the dependency. And the test have to know how an internal detail of the dependency needs to operate (which calls needs to make, to where, with which arguments, which should expect in the response, etc.)

Just to clear a bit my view 🙂: for me the http services we're calling (config service and transaction service) are just data providers, they just provide data, the protocol (http) it's just a detail, and could be treated like a database, a queue, etc. In my mind mocking `axios` here is the same that mocking `mongodb` driver for example.

But I get your point, I see these tests valuable as well, and maybe when e2e test are in place we'll see we're doing something very close to this, and we can refactor. I will implement the test as you did for `balances.controller.spec.ts` 👍 

> For real E2E tests it can be more complex... specially when creating transactions and submitting transactions. We need to have some kind of test database or similar infrastructure to run this. Even though they provide a lot of value, some flows can be tricky to test.

Got it, OK, you have all the experience on those scenarios. As I said my main concern was we were doing an hybrid approach between unit and e2e tests, and when we have real e2e tests all this code would be duplicated. But we'll tackle that in the future. 

Thanks for discussing this, glad to share opinions 😄

fmrsabino: > I partially agree, I think too much spying would led to coupling between tests and implementation, but by just mocking the dependency you're just tied to its interface, not its internal implementation. By mocking axios, which is an internal detail of the dependency (in the future the dependency could even use a different http client) you're tied to the actual implementation of the dependency. And the test have to know how an internal detail of the dependency needs to operate (which calls needs to make, to where, with which arguments, which should expect in the response, etc.)
> Just to clear a bit my view 🙂: for me the http services we're calling (config service and transaction service) are just data providers, they just provide data, the protocol (http) it's just a detail, and could be treated like a database, a queue, etc. In my mind mocking `axios` here is the same that mocking `mongodb` driver for example.
> 
> But I get your point, I see these tests valuable as well, and maybe when e2e test are in place we'll see we're doing something very close to this, and we can refactor. I will implement the test as you did for `balances.controller.spec.ts` 👍

Ah completely agree! Mocking `axios` does seem weird. My original idea was to have a `TestHttpModule` which is used instead of the actual one. Or create an interface of our own. This is something that we should sync on.

The same would be the same for external "services" like caching solutions or even the environment that we use (`.env`).

For testing it should be easy enough to replace these external components with a test one but one that is not tied directly to the "actual" implementation (in this case mocking `axios`).

> > For real E2E tests it can be more complex... specially when creating transactions and submitting transactions. We need to have some kind of test database or similar infrastructure to run this. Even though they provide a lot of value, some flows can be tricky to test.
> 
> Got it, OK, you have all the experience on those scenarios. As I said my main concern was we were doing an hybrid approach between unit and e2e tests, and when we have real e2e tests all this code would be duplicated. But we'll tackle that in the future.
> 
> Thanks for discussing this, glad to share opinions 😄

I'd still consider the tests that I mentioned above as "unit" tests (the unit is just bigger instead of being a single class) but there's no golden rule for this. We should get to a solution that it's easy to read and maintain 🙌

package.json
```diff
@@ -74,6 +74,7 @@
       "**/*.(t|j)s"
     ],
     "coverageDirectory": "../coverage",
+    "coveragePathIgnorePatterns": ["index.ts"],
```
hectorgomezv, line 77: No related to this PR, but it was easy and quick. We also have other file patterns which could be ignored by the coverage report (`main.ts`, `app.module.ts`), what do you think on flagging those?

fmrsabino (approved):
🚀 🚀

PR #31, opened 2022-08-16
Add tests to Balances Controller. Add factory functions.

### LINKED ISSUE #24: Use an automatic fake data generator on tests
**Summary**
Currently we are generating our test data manually per test. It would be nice to integrate a tool to generate this fake data to favor tests robustness. [Faker](https://www.npmjs.com/package/@faker-js/faker) seems a good candidate for this.

**Acceptance criteria**
- An automatic tool should be added as development-time dependency only.
- Current and further test classes should use this tool instead of manually generate test/fake data.

fmrsabino:
Closes #24, #19

- Add `@faker-js/faker` as a dev dependency – it allows to generate random data that can be used for testing
- Add factories for `Balance`, `BalanceToken` (from the Transaction Service)
- Add factories for `Chain` (from the Safe Config Service)
- Add factories for the Exchange Service response
- Add more unit tests to Balances Controller. Cases:
  * Success case
  * Exchange API Error – 500 error response
  * Exchange API Error – No rates returned
  * Exchange API Error – from-rate missing
  * Exchange API Error – from-rate is 0
  * Exchange API Error – to-rate missing
  * Transaction Service API Error – 500 error response
- `Balance` from Transaction Service is now a `bigint`

### DIFF
--- a/src/balances/balances.controller.spec.ts
+++ b/src/balances/balances.controller.spec.ts
@@ -31,7 +31,7 @@ describe('Balances Controller (Unit)', () => {
     it(`Success`, async () => {
       const chainId = '1';
       const safeAddress = '0x0000000000000000000000000000000000000001';
-      const safeTransactionServiceBalancesResponse = safeBalanceFactory(1);
+      const transactionServiceBalancesResponse = safeBalanceFactory(1);
       const exchangeResponse = exchangeResultFactory({ USD: 2.0 });
       const chainResponse = chainFactory(chainId);
       axiosMock.get.mockImplementation((url) => {
@@ -42,7 +42,7 @@ describe('Balances Controller (Unit)', () => {
           `${chainResponse.transactionService}/api/v1/safes/${safeAddress}/balances/usd/`
         ) {
           return Promise.resolve({
-            data: safeTransactionServiceBalancesResponse,
+            data: transactionServiceBalancesResponse,
           });
         } else if (url == configuration().exchange.baseUri) {
           return Promise.resolve({ data: exchangeResponse });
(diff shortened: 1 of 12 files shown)


src/balances/balances.service.ts
```diff
@@ -83,7 +83,7 @@ export class BalancesService {
         name: txBalance.token?.name,
         logoUri: logoUri,
       },
-      balance: txBalance.balance,
+      balance: txBalance.balance.toString(),
```
fmrsabino, line 86: This is because of the type change. `balance` is now a `bigint`

hectorgomezv: I was wondering if `txBalance.balance` could be undefined, in which case this would produce a runtime error. I assume it's ok to fail fast and loud in that scenario, but this leads me to another question, should we validate payloads coming from Config/Transaction services? That could be done in a separate PR, of course, but could provide a more controlled behaviour when an unexpected payload shape comes into the service. I've happily used Joi (https://joi.dev/) in the past for that, but Nest recommends this: https://github.com/typestack/class-validator

What do you think? 🙂

src/balances/balances.controller.spec.ts
```diff
@@ -19,95 +23,282 @@ describe('Balances Controller (Unit)', () => {
...
-              address: '0x6810e776880C02933D47DB1b9fc05908e5386b96',
-              decimals: 16,
-              symbol: 'GNO',
-              name: 'Gnosis',
-              logoUri: null,
+  describe('GET /balances', () => {
+    it(`Success`, async () => {
+      const chainId = '1';
+      const safeAddress = '0x0000000000000000000000000000000000000001';
+      const safeTransactionServiceBalancesResponse = safeBalanceFactory(1);
```
hectorgomezv, line 34: Do you think it's better to keep the `safe` prefix here? (I think we removed it in the service implementation)

src/balances/balances.controller.spec.ts
```diff
@@ -19,95 +23,282 @@ describe('Balances Controller (Unit)', () => {
...
+        } else {
+          return Promise.reject(new Error(`Could not match ${url}`));
+        }
+      });
+
+      await request(app.getHttpServer())
+        .get(`/chains/${chainId}/safes/${safeAddress}/balances/usd`)
+        .expect(200)
+        .expect({
+          fiatTotal: safeTransactionServiceBalancesResponse[0].fiatBalance,
```
hectorgomezv, line 58: NP: maybe if you extract the first element in the array beforehand it could be (in my opinion) a bit more readable? What do you think? 🙂
Screenshot 2022-08-17 at 09 38 58

hectorgomezv (approved):
Great work! I just left minor comments 👍

src/balances/balances.service.ts
fmrsabino: > I was wondering if `txBalance.balance` could be undefined, in which case this would produce a runtime error. I assume it's ok to fail fast and loud in that scenario, but this leads me to another question, should we validate payloads coming from Config/Transaction services? That could be done in a separate PR, of course, but could provide a more controlled behaviour when an unexpected payload shape comes into the service. I've happily used Joi (https://joi.dev/) in the past for that, but Nest recommends this: https://github.com/typestack/class-validator
> 
> What do you think? 🙂

I agree. We should look into these solutions for validating the incoming Json. Throwing another one into the list: https://ajv.js.org/

src/balances/balances.controller.spec.ts
fmrsabino: [ca8b3d8](https://github.com/5afe/safe-client-gateway-nest/pull/31/commits/ca8b3d84b29e134fddb0f2fdb917ce363230b7d1)

src/balances/balances.controller.spec.ts
fmrsabino: [ca8b3d8](https://github.com/5afe/safe-client-gateway-nest/pull/31/commits/ca8b3d84b29e134fddb0f2fdb917ce363230b7d1)

hectorgomezv (approved):

PR #34, opened 2022-08-16
Add logger middleware

hectorgomezv:
This implements logging for every request/response the service handles.
The format for the request and response logger messages is totally open to suggestions 🙂

This is an example of the actual output:
Screenshot 2022-08-16 at 18 35 58

### DIFF
--- a/src/common/middleware/logger.middleware.ts
+++ b/src/common/middleware/logger.middleware.ts
@@ -1,0 +1,34 @@
+import { Injectable, Logger, NestMiddleware } from '@nestjs/common';
+import { NextFunction, Request, Response } from 'express';
+
+@Injectable()
+export class LoggerMiddleware implements NestMiddleware {
+  private readonly logger = new Logger(LoggerMiddleware.name);
+
+  private formatRequestMessage = (req: Request) =>
+    `[==>] ${req.method} ${req.url}`;
+
+  private formatResponseMessage = (res: Response) => {
+    const contentLength = res.get('content-length') || '';
+    const contentType = res.get('content-type') || '';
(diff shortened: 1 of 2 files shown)


src/common/middleware/logger.middleware.ts
```diff
@@ -0,0 +1,34 @@
...
+export class LoggerMiddleware implements NestMiddleware {
+  private readonly logger = new Logger(LoggerMiddleware.name);
+
+  private formatRequestMessage = (req: Request) =>
+    `[==>] ${req.method} ${req.url}`;
+
+  private formatResponseMessage = (res: Response) => {
+    const contentLength = res.get('content-length') || '';
+    const contentType = res.get('content-type') || '';
+    return `[<==] ${res.statusCode} ${contentLength} ${contentType}`;
```
fmrsabino, line 14: We don't need to tackle it in this PR but we need to consider the current format that we use: https://github.com/safe-global/safe-client-gateway/blob/dd162f73150e8de4ea6eda481da1ee802692bdbe/src/monitoring/performance.rs#L22-L49

This can be discussed with the team. If we want to keep the same format or not.
I'd also consider logging a Request ID and making it available throughout the request flow (this allows us to see what a specific request did). Also something for another PR 👍

fmrsabino (approved):
Nice 👏 

Testing this one can be more tricky. Probably by spying on `Logger` and triggering `use()`.

src/common/middleware/logger.middleware.ts
hectorgomezv: Awesome, I created [an issue](https://github.com/5afe/safe-client-gateway-nest/issues/35) for that 🙂

PR #36, opened 2022-08-17
Add SAFE_CONFIG_BASE_URI to configuration.ts

fmrsabino:
- The Safe Config Base URI is now read from the `configuration.ts`
- If `SAFE_CONFIG_BASE_URI` is set, then its value is used. Else defaults to `https://safe-config.gnosis.io`

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -3,3 +3,7 @@
 # The EXCHANGE_API_KEY should always be set
 #EXCHANGE_API_BASE_URI=
 #EXCHANGE_API_KEY=
+
+# The base url for the Safe Config Service
+# Default if none is set: https://safe-config.gnosis.io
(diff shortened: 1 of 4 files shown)


hectorgomezv (approved):
👏🏻

fmrsabino:
I think we should keep `src/chains/chains.controller.spec.ts` and `src/chains/chains.service.spec.ts` on a 2nd thought what do you think @hectorgomezv ?

It basically tests that the module can be built.

hectorgomezv:
> I think we should keep `src/chains/chains.controller.spec.ts` and `src/chains/chains.service.spec.ts` on a 2nd thought what do you think @hectorgomezv ?
> 
> It basically tests that the module can be built.

Yes, I think we can keep them as they are, and maybe do proper testing in a separate issue/PR 🙂

PR #38, opened 2022-08-17
Add injectable INetworkService

### LINKED ISSUE #37: Research how to abstract the HttpClient for testing
The current endpoints are exposed to the `HttpModule` provided by NestJS. Even though this works it makes testing difficult as this is one component that we'd like to mock/provide fake data (remote http calls).

An alternative solution should be considered where we avoid mocking the axios client directly (which ties our tests directly with the http client being used)

fmrsabino:
Closes #37

- Adds `INetworkService` – an interface which should be injected in order to perform network related operations. The default implementation provided is `AxiosNetworkService`
- Adds a `NetworkModule` – a Global module which is provided by the `AppModule`. This means that new features can assume that `INetworkService` will be provided by the main application without depending directly on it
- A TestNetworkModule can now be used in the tests where a mock HTTP client is needed

### DIFF
--- a/src/app.module.spec.ts
+++ b/src/app.module.spec.ts
@@ -1,0 +1,14 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { AppModule } from './app.module';
+
+describe('AppModule', () => {
+  it(`AppModule is successfully created`, async () => {
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    const app = moduleFixture.createNestApplication();
+    await app.init();
+    await app.close();
+  });
+});
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -8,9 +8,10 @@ import {
 import { ChainsModule } from './chains/chains.module';
 import { BalancesModule } from './balances/balances.module';
 import { LoggerMiddleware } from './common/middleware/logger.middleware';
+import { NetworkModule } from './common/network/network.module';
 
 @Module({
-  imports: [ChainsModule, BalancesModule],
+  imports: [BalancesModule, ChainsModule, NetworkModule],
 })
 export class AppModule implements NestModule {
   configure(consumer: MiddlewareConsumer) {
--- a/src/balances/balances.controller.spec.ts
+++ b/src/balances/balances.controller.spec.ts
@@ -1,15 +1,15 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { INestApplication } from '@nestjs/common';
 import * as request from 'supertest';
-import axios from 'axios';
-import { BalancesModule } from './balances.module';
 import safeBalanceFactory from '../services/transaction-service/entities/__tests__/balance.factory';
 import exchangeResultFactory from '../services/exchange/entities/__tests__/exchange.factory';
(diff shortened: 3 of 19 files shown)


fmrsabino:
@hectorgomezv I'm considering the alternative where the features do not depend on "common" modules like the HTTP one directly but would expect one to be provided by the app.

Nest only allows this via Global modules. The advantage of this would be that we could remove most of the Testing modules and we register the production HTTP module in production code and the test one in the tests.

hectorgomezv (approved):
Awesome job! I really like this approach, way better than depending on Axios explicitly! 🚀

> @hectorgomezv I'm considering the alternative where the features do not depend on "common" modules like the HTTP one directly but would expect one to be provided by the app.

> Nest only allows this via Global modules. The advantage of this would be that we could remove most of the Testing modules and we register the production HTTP module in production code and the test one in the tests. 

Sounds good 👍🏻 For this case I think we'll only use the `NetworkModule` for config/transaction services but making it `Global` it's OK. When implementing caching this `Global` feature would be very useful.

BTW I have [this PR](https://github.com/5afe/safe-client-gateway-nest/pull/30) without using this `NetworkModule` obviouly, I think we could merge it anyway to maintain it concise, and I will create an issue to make that feature use `NetworkModule`, what do you think?

fmrsabino:
> Awesome job! I really like this approach, way better than depending on Axios explicitly! 🚀

Yeah I agree 🙌 

> > @hectorgomezv I'm considering the alternative where the features do not depend on "common" modules like the HTTP one directly but would expect one to be provided by the app.
> 
> > Nest only allows this via Global modules. The advantage of this would be that we could remove most of the Testing modules and we register the production HTTP module in production code and the test one in the tests.
> 
> Sounds good 👍🏻 For this case I think we'll only use the `NetworkModule` for config/transaction services but making it `Global` it's OK. When implementing caching this `Global` feature would be very useful.

Yeah Caching would also be something that the `AppModule` would provide. Features can assume it exists. If it doesn't it won't work either way (error message).

Same for the Environment Variables 🙌 

> BTW I have [this PR](https://github.com/5afe/safe-client-gateway-nest/pull/30) without using this `NetworkModule` obviouly, I think we could merge it anyway to maintain it concise, and I will create an issue to make that feature use `NetworkModule`, what do you think?

Yes! It's ok! 👍

fmrsabino:
> > BTW I have [this PR](https://github.com/5afe/safe-client-gateway-nest/pull/30) without using this `NetworkModule` obviouly, I think we could merge it anyway to maintain it concise, and I will create an issue to make that feature use `NetworkModule`, what do you think?
> 
> Yes! It's ok! 👍

I adjusted the tests in this PR. It was simple change @hectorgomezv

src/app.module.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+  it(`AppModule is successfully created`, async () => {
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    const app = moduleFixture.createNestApplication();
+    await app.init();
+    await app.close();
+  });
+});
```
fmrsabino, line 14: A module test can simply import the module and use it in the `App` if the module is missing a dependency it'd fail the test

hectorgomezv (approved):

PR #41, opened 2022-08-19
Add Injectable IConfigurationService

### LINKED ISSUE #40: Create wrapper around ConfigModule
A wrapper should be created around Nest's `ConfigModule`.

The reason is that this module reads from the local environment which is external to the service and in order to provide a better testing setup we should be able to replace the `ConfigModule` with a fake one.

fmrsabino:
Closes #40 

- `IConfigurationService` is an injectable class which provides an interface which can be used to read from the loaded configuration
- The default implementation provided uses `ConfigService` from Nest which reads values from `configuration.ts`
- Adds a `TestConfigurationModule` which includes a `FakeConfigurationService` – this `FakeConfigurationService` can be used to override environment variables in a test setup.


hectorgomezv (approved):
Looks good to me. Great job! 👏🏻

hectorgomezv (approved):

PR #42, opened 2022-08-19
Add Dockerfile

hectorgomezv:
This PR implements a first approach to having a Docker container to run the service. It uses a multi-stage build to avoid populating the final image with unnecessary files, and it also runs as non-root user for improved security.


hectorgomezv:
I'm including @luarx to reviewers. It would be also nice to check further steps (integrating it with CI runner, external configuration, Kubernetes, etc.) with DevOps 🙂

Dockerfile
```diff
@@ -0,0 +1,21 @@
+#
+# BUILD CONTAINER
+#
+FROM node:16 as base
+USER node
```
fmrsabino, line 5: Do we need to set a user?

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ENV NODE_ENV production
+WORKDIR /usr/src/app
+COPY --chown=node:node package.json yarn.lock .yarnrc.yml tsconfig*.json ./
```
fmrsabino, line 8: ooc: does copying `.yarnrc.yml` change the outcome of the image? I was under the impression that `corepack` had to be enabled for this to work and I'm not sure this is the case with this image.

fmrsabino: After inspecting the image it does seem that it doesn't run with the bundled yarn version 🤔 

```
$ yarn --version
1.22.19
```

fmrsabino: Actually with the following two lines I'd expect it to use the bundled one

```
COPY --chown=node:node .yarn/releases ./.yarn/releases
RUN yarn install --immutable 
```

Can you confirm this behaviour?

Dockerfile
```diff
@@ -0,0 +1,21 @@
+#
+# BUILD CONTAINER
+#
+FROM node:16 as base
+USER node
+ENV NODE_ENV production
+WORKDIR /usr/src/app
```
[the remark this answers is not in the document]
hectorgomezv: Not really, it's one of the usual options. I'll change it then to keep the same path that the old image, makes sense 👍

Dockerfile
hectorgomezv: I think it's better to run containers as non-root user (the default): https://docs.bitnami.com/tutorials/why-non-root-containers-are-important-for-security

Dockerfile
hectorgomezv: Good catch, I didn't verify this was using Yarn 3.X, I will research it and update the PR, thanks.

hectorgomezv: It's using Yarn 3.2.2. I added a step to the build container to check it:
```
RUN yarn --version
```
And this is the output.
```
#11 [base 5/8] RUN yarn --version
#11 sha256:6f7a19c95f998d36997d3f1ce06e6242edeb00b0837490b4b8cd06c13d87f41d
#11 0.438 3.2.2
#11 DONE 0.5s
``` 
The Yarn version it's different in the prod container as we suspected (and of course it doesn't matter 🙂).

fmrsabino (approved):
🚀

PR #46, opened 2022-08-22
Fix prettier/eslint integration

hectorgomezv:
I've adjusted it to 120 in `eslint` and 119 in `prettier`. After some tests I concluded that with this configuration we would be able to keep using VSCode keybindings to automatically format the code.

Please feel free of share your opinions about the size, or raise any problem you may encounter formatting the code.


src/app.module.ts
```diff
@@ -23,8 +18,6 @@ import { ConfigurationModule } from './common/config/configuration.module';
 })
 export class AppModule implements NestModule {
   configure(consumer: MiddlewareConsumer) {
-    consumer
-      .apply(LoggerMiddleware)
-      .forRoutes({ path: '*', method: RequestMethod.ALL });
```
fmrsabino, line 28: Was this done automatically? I actually have a slight preference for the left part since it looks like a builder pattern.

.prettierrc
```diff
@@ -1,4 +1,5 @@
 {
   "singleQuote": true,
-  "trailingComma": "all"
-}
\ No newline at end of file
+  "trailingComma": "all",
+  "printWidth": 119
```
fmrsabino, line 4: After reading https://prettier.io/docs/en/options.html#print-width do you know how different it'd be if we'd just have the length set with eslint?

src/app.module.ts
hectorgomezv: Yes, almost all changes were done automatically, I'm ok with both versions, but I can preserve the old one if you prefer 👍

.prettierrc
hectorgomezv: From my experience that would cause the editor to format the code using the default 80 chars per line when using the automatic formatting, it takes `.prettierrc` file as source of truth to do it.

src/app.module.ts
hectorgomezv: I was thinking we could preserve the left part if the line length is not reached but it seems prettier enforces styling even if the line length is not reached. At this point we have automatic formatting when committing, so I would take one of the options: 1. close this PR and stay on the 80 chars limit. 2. assume that prettier will format this.

What do you think @fmrsabino? Do you see other options here? 🙂

hectorgomezv:
@fmrsabino I just removed plugins and integrations between `eslint` and `prettier`. Now I'd expect both tools to do their job separately. As we talked `prettier` will keep formatting the code on a pre-commit hook, but lets keep in mind we could remove this behaviour anytime if we're not confortable with it. 🙂

fmrsabino (approved):
🚀 🚀 🚀

PR #47, opened 2022-08-23
Implement /balances/supported-fiat-codes

### LINKED ISSUE #20: Implement /balances/supported-fiat-codes
- Implement /v1/balances/supported-fiat-codes with the same schema as the current Safe Client Gateway.
- Caching is not required for this ticket
- Unit tests should be added for this route

hectorgomezv:
Closes #20 

@fmrsabino I used the `Supported Symbols Endpoint` ([link](https://exchangeratesapi.io/documentation/) to the docs) for this. I think it's sightly more appropriate to use this specific endpoint to retrieve the currencies. I was reading the old implementation code and it seems it was using the exchange `/latest` endpoint to retrieve the currencies as well. What do you think about this? Thanks!


src/balances/balances.service.ts
```diff
@@ -88,4 +88,13 @@ export class BalancesService {
       fiatConversion: fiatConversion,
     };
   }
+
+  async getSupportedFiatCodes(): Promise<string[]> {
```
hectorgomezv, line 92: I'm not sure of this business logic. For brevity I'm assuming here that both `USD` and `EUR` we'll be always available, instead of checking their presence in the array 🤔

src/services/exchange/exchange.service.ts
```diff
@@ -41,14 +44,29 @@ export class ExchangeService {
...
 
-  private async getExchangeResult(): Promise<ExchangeResult> {
+  async getFiatCodes(): Promise<string[]> {
+    const fiatCodesResult: FiatCodesExchangeResult =
+      (await this.getExchangeResult('symbols')) as FiatCodesExchangeResult;
+
+    return Object.keys(fiatCodesResult.symbols);
+  }
+
+  private async getExchangeResult(
```
hectorgomezv, line 54: I've modified this function to serve both endpoints and avoid repeating ourselves. But in fact I think we could extract this function and create something like a `ExchangeClient` to encapsulate the data retrieval and make `ExchangeService` agnostic. Wdyt?

src/services/exchange/exchange.service.ts
```diff
@@ -41,14 +44,29 @@ export class ExchangeService {
...
         params: { access_key: apiKey },
       });
+
+      if (!data?.success) {
+        throw new InternalServerErrorException('Bad response from Exchange');
+      }
+
       return data;
     } catch (error) {
       this.httpErrorHandler.handle(error);
```
hectorgomezv, line 72: Also, by extracting the function to an `ExchangeClient` we could also create a specific `ErrorHandler` for the Exchange errors, since their payload is sightly different from our Safe services error payloads. 🤔

hectorgomezv: Regarding this, I think we could create a general application-wide error handler which parser all the errors and defines a common payload interface. So every data provider kind should have an error handler which maps specific errors to the general error format.

src/services/exchange/exchange.service.ts
fmrsabino: I'd separate both calls. We can do this in the `ExchangeService` for now without creating another class (at least for now).

But I'd split both calls as we might want to handle error cases differently (and I also think that mapping 1 endpoint to 1 function makes it easier to read overall).

fmrsabino: In order to not read from the Config everytime we call these functions I'd suggest the following:

```typescript
  private readonly baseUrl: string;
  private readonly apiKey: string;

  constructor(
    @Inject(NetworkService) private readonly networkService: INetworkService,
    private readonly httpErrorHandler: HttpErrorHandler,
  ) {
    this.baseUrl =
      this.configurationService.getOrThrow<string>('exchange.baseUri');
    this.apiKey =
      this.configurationService.getOrThrow<string>('exchange.apiKey');
  }
```

src/services/exchange/exchange.service.ts
fmrsabino: The discussion around error codes might also be helpful for this one. Where we have a generic error case but a specific error code that helps us identify the source of the issue.

Also Nest already sets some generic error handler. For this PR I wouldn't say it's expected to "set" the appropriate messages and error codes. I think if you remove our error handler usage here and just throw the error it would be fine at this stage 👍

src/balances/balances.controller.spec.ts
```diff
@@ -308,11 +308,39 @@ describe('Balances Controller (Unit)', () => {
...
+
+  describe('GET /balances/', () => {
+    it('Success', async () => {
+      const fiatCodesResult = fiatCodesResultFactory();
+      mockNetworkService.get.mockResolvedValueOnce({ data: fiatCodesResult });
+
+      await request(app.getHttpServer())
+        .get('/balances/supported-fiat-codes')
+        .expect(HttpStatus.OK)
+        .expect(['USD', 'EUR', 'AED', 'AFN', 'ALL']);
```
fmrsabino, line 330: nitpick: if you change the `fiatCodesResultFactory` this test ends up failing so it seems that you are testing what the factory returns.

I'd use the factory if you don't care about its internal data (just that it has "correct" data). In order to test specific data I'd do that setup in the test itself.

src/balances/entities/__tests__/fiat-codes.factory.ts
```diff
@@ -0,0 +1,3 @@
+export default function (): string[] {
+  return ['AED', 'AFN', 'EUR', 'ALL', 'USD'];
+}
```
fmrsabino, line 3: Since this is only used in one place I'd just use this payload in the test itself. It also makes it easier to read because you can see what the function returns and what is expected in the same function.

These factory functions would probably live closer to the data involved (for this one probably under `src/services/exchange/entities/__tests__/` since then the factory should be useful for other modules to use)

fmrsabino (changes requested):

src/services/errors/http-error-handler.ts
```diff
@@ -22,7 +22,7 @@ export class HttpErrorHandler {
       };
     } else {
       return <HttpExceptionPayload>{
-        message: 'Service unavailable',
+        message: err.message || 'Service unavailable',
```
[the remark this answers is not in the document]
hectorgomezv: Exactly, that's why I was wondering if we should have independent clients and error handlers for each data provider (config/transaction/exchange), and map every specific error to a general error interface. I agree that could be done in a further step by the way 🙂 I'll keep this as it was.

src/balances/balances.controller.spec.ts
```diff
@@ -125,19 +126,19 @@ describe('Balances Controller (Unit)', () => {
...
+            return Promise.reject({ status: HttpStatus.INTERNAL_SERVER_ERROR });
           } else {
             return Promise.reject(new Error(`Could not match ${url}`));
           }
         });
 
         await request(app.getHttpServer())
           .get(`/chains/${chainId}/safes/${safeAddress}/balances/usd`)
-          .expect(503)
+          .expect(HttpStatus.SERVICE_UNAVAILABLE)
```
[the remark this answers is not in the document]
hectorgomezv: I get your point. These codes are HTTP statuses which won't never change, and most likely we won't be using our our codes instead of the protocol ones, but I get your point about being more explicit in the tests and remove dependencies. Keeping the numeric value in the test, and the enum reference in the production code, it's also a safe net for human errors setting the codes, so I will keep numeric codes in the tests 👍

src/balances/balances.controller.spec.ts
```diff
@@ -309,10 +310,48 @@ describe('Balances Controller (Unit)', () => {
...
+            message: 'Service unavailable',
+            code: 503,
+          });
 
         expect(mockNetworkService.get.mock.calls.length).toBe(2);
       });
     });
   });
+
+  describe('GET /balances/', () => {
```
fmrsabino, line 323: nitpick:

```suggestion
  describe('GET /balances/supported-fiat-codes', () => {
```

fmrsabino (commented):
Looks good! Some small comments.

Also we have some lint warnings. I'm considering adding the imports one as an error actually what do you think?

fmrsabino:
> Looks good! Some small comments.
> 
> Also we have some lint warnings. I'm considering adding the imports one as an error actually what do you think?

Added a PR for it here: https://github.com/5afe/safe-client-gateway-nest/pull/50

hectorgomezv:
> Also we have some lint warnings. I'm considering adding the imports one as an error actually what do you think?

Yes, I always have issues with that 😄 definitely would be great to mark them as error 👍🏻

hectorgomezv:
> Looks good! Some small comments.
> 
> Also we have some lint warnings. I'm considering adding the imports one as an error actually what do you think?

Fixed [here](https://github.com/5afe/safe-client-gateway-nest/pull/47/commits/575fa687c7c2224b660ab27eda60f22cbf29ccdb) 😄

.eslintrc.js
```diff
@@ -21,5 +18,6 @@ module.exports = {
     '@typescript-eslint/explicit-function-return-type': 'off',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
     '@typescript-eslint/no-explicit-any': 'off',
+    '@typescript-eslint/no-unused-vars': 'error',
```
fmrsabino, line 21: Can we remove this one in favour of https://github.com/5afe/safe-client-gateway-nest/pull/50?

I think it's more clear if it's a separate PR and since it's already open

hectorgomezv: I merged #50 here to see the effect when setting this config. Now it's #50 merged in main so I merged main here as well. 🙂 I usually don't proceed this way but this was trivial, sorry about the misunderstandings.

fmrsabino (approved):
Great work! 🚀

PR #49, opened 2022-08-25
Refactor external data providers naming

### LINKED ISSUE #48: Refactor external data providers naming
**Summary**
Currently we're using the term `service` to refer to the external data providers (Safe Transaction Service, Safe Config Service). Now we we also have an `Exchange` data provider in place which provides currencies data. 

We adopted Nest (an other frameworks as well) convention to place our internal business logic into a `services` layer. Due to this, we have a name collision between two concepts which should be differentiated: external data providers and internal business logic.

This conflict is even worse thinking about `Transaction`, because at some point we'll have a `Transaction` internal service (to manage transaction endpoints/entities) which directly collides with the external `Transaction` data provider (Safe Transaction Service).

**Proposal**
The idea here is to drop a proposal and comment/iterate over it:
- Rename folder `src/services` to `src/datasources` (other options: `src/external`, `src/providers`).
- Rename folder `config-service` to `config-api` (and it's internal classes accordingly).
- Rename folder `transaction-service` to `transaction-api` (and it's internal classes accordingly).
- Rename folder `exchange` to `exchange-api` (and it's internal classes accordingly).

hectorgomezv:
This PR implements the renaming strategy we discussed in the related issue.

Closes #48


moisses89 (approved):
Looks good for me

fmrsabino (approved):
LGTM! 🚀 

Strange that for some files it was set as added/removed.

hectorgomezv:
> LGTM! 🚀
> 
> Strange that for some files it was set as added/removed.

Yes, I think when path changes and also the file content changes (even sightly) git interprets that you are deleting and adding the file.

fmrsabino:
> > LGTM! 🚀
> > Strange that for some files it was set as added/removed.
> 
> Yes, I think when path changes and also the file content changes (even sightly) git interprets that you are deleting and adding the file.

Maybe import related 👍

PR #51, opened 2022-08-25
Run build step in tests job

fmrsabino:
- The build step should be executed in order to catch files that fail on that step
- The test, linter and formatter steps would most likely catch those, but for the compilation itself we should rely on the build step.


hectorgomezv:
🚀

PR #53, opened 2022-08-25
Use number instead of bigint

fmrsabino:
- Our current target is `es2017`. BigInt is available on `esnext`
- `number` right now provides what we need (given the current stage of the project)


hectorgomezv (approved):
👍🏻

PR #54, opened 2022-08-26
Use flag strictNullChecks on TypeScript compiler

### LINKED ISSUE #52: Use flag strictNullChecks on TS compiler
We need to activate [strict null checks](https://www.typescriptlang.org/tsconfig#strictNullChecks) in the TS compiler configuration as a best practice to avoid null/undefined references at runtime.

hectorgomezv:
Closes #52 

I adapted some parts of the codebase to the usage of the strict null checks while compiling TS into JS. This feature prevents us from introducing undefined or null references in the code which could led to runtime errors. It is also required to use AJV, which we plan to use for the external datasources data validations.

All the changes should be pretty straightforward. Maybe the most relevant is to change the http error handler we had to make a http error factory. This way instead of delegating the error throwing when an error happens, we use the error factory to just create the appropriate error. This avoids TS warnings about functions returning `undefined`, and removes the necessity of handling `undefined` in the caller functions. In my opinion this makes the code cleaner.


src/datasources/errors/http-error-factory.ts
```diff
@@ -3,21 +3,21 @@ import axios, { AxiosError } from 'axios';
 import { HttpExceptionPayload } from './interfaces/http-exception-payload.interface';
 
 /**
```
hectorgomezv, line 5: This class is named `*ErrorFactory` but it could be `*ErrorMapper` as well. I'm open to hear name suggestions @fmrsabino 🙂

src/datasources/exchange-api/exchange-api.service.ts
```diff
@@ -25,6 +27,14 @@ export class ExchangeApi {
   ) {
     this.baseUrl = this.configurationService.get<string>('exchange.baseUri');
     this.apiKey = this.configurationService.get<string>('exchange.apiKey');
+
```
hectorgomezv, line 30: This is not strictly necessary, but TS compiler was complaining about `baseUrl`/`apiKey` could be `undefined`, so I realised it could be nice to warn about this situation at startup time if it happens. I chose to use `logger.warn` but it could be `logger.error` as well, no strong preference here.

fmrsabino: I would suggest the following: setting both as `getOrThrow`. This class should only work if we have at least these set.

If you use that you might be able to remove the `undefined` from the `baseUrl` and `apiKey`

fmrsabino (changes requested):
Some smaller comments. I think `HttpErrorFactory` better describes what it does now! 🙌

src/datasources/errors/http-error-factory.spec.ts
```diff
@@ -0,0 +1,58 @@
+import { AxiosError, AxiosResponse } from 'axios';
+import { HttpErrorFactory } from './http-error-factory';
+import { HttpExceptionPayload } from './interfaces/http-exception-payload.interface';
+
+describe('HttpErrorFactory', () => {
```
fmrsabino, line 5: nitpick: One additional test that can be written is if you get an error response with an `undefined` or `null` status.

In those cases we map it to a `500`

src/datasources/exchange-api/exchange-api.service.ts
hectorgomezv: I agree, much better solution, thanks!

src/balances/balances.controller.spec.ts
```diff
@@ -81,10 +81,10 @@ describe('Balances Controller (Unit)', () => {
...
                 tokenType: 'ERC20',
                 address: expectedBalance.tokenAddress,
-                decimals: expectedBalance.token.decimals,
-                symbol: expectedBalance.token.symbol,
-                name: expectedBalance.token.name,
-                logoUri: expectedBalance.token.logo_uri,
+                decimals: expectedBalance?.token?.decimals,
+                symbol: expectedBalance?.token?.symbol,
+                name: expectedBalance?.token?.name,
+                logoUri: expectedBalance?.token?.logo_uri,
```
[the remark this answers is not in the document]
hectorgomezv: No, my bad, I fix this warning too quickly. `expectedBalance` shouldn't be undefined/null, the only possibility here is to have a undefined reference in the `token` property. I'll change this, thanks!

fmrsabino (approved):
🚀

PR #56, opened 2022-08-26
Add tests to AxiosNetworkService

fmrsabino:
- Adds test cases to AxiosNetworkService

### DIFF
--- a/src/common/network/axios.network.service.spec.ts
+++ b/src/common/network/axios.network.service.spec.ts
@@ -1,0 +1,57 @@
+import { AxiosNetworkService } from './axios.network.service';
+import { AxiosInstance } from 'axios';
+import { HttpService } from '@nestjs/axios';
+import { faker } from '@faker-js/faker';
+import { NetworkRequest } from './entities/network.request.entity';
+import { jest } from '@jest/globals';
+
+const axiosRef = {
+  get: jest.fn(),
+} as unknown as AxiosInstance;
+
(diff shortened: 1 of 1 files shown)


src/common/network/axios.network.service.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+  // TODO Property 'mockResolvedValueOnce' does not exist on type '<T = any, R = AxiosResponse<T, any>, D = any>(url: string, config?: AxiosRequestConfig<D>) => Promise<R>'
+  // it(`get forwards error`, async () => {
+  //   const url = faker.internet.url();
+  //   axiosRefMock.get.mockResolvedValueOnce();
+  //
+  //   await target.get(url);
+  //
+  //   expect(axiosRefMock.get).toBeCalledTimes(1);
+  //   expect(axiosRefMock.get).toBeCalledWith(url);
+  // });
```
fmrsabino, line 56: @hectorgomezv I was having some issues on mocking the actual value returned (I think it's because of the generic).

We can discuss this on Monday.

fmrsabino: Fixed it by casting to `any` [b4d46a6](https://github.com/5afe/safe-client-gateway-nest/pull/56/commits/b4d46a65141c40a1851672adf191deb808ddbccf).

Should be fine as we are only interested in asserting mock interactions.

PR #58, opened 2022-08-29
Add cache layer

### LINKED ISSUE #44: Integrate caching solution
Similar to the the Rust implementation, a caching solution should be introduced. 

- Redis is already deployed in production so we should keep using the same solution.
- The implementation used (for caching) should be easily replaceable by a mock/fake for testing
- Caching should only be done for external requests done by the service

Caching should be done in a more effective manner: see https://github.com/safe-global/safe-client-gateway/issues/849

fmrsabino:
Closes #44

- Adds a caching layer which is exposed under `CacheModule` – this module is a global module which means it is available to every module registered in `AppModule`.
- For testing purposes, `TestCacheModule` should be used for a more refined control over the current state of the cache
- Adds `CacheFirstDataSource` – this `DataSource` is available under `CacheFirstModule`.
  * This `DataSource` should be used if access to caching is required by the features
  * It relies on an HTTP fallback in case the cache entry has either expired or is not available
- Adds `Redis` as the main cache service for the application.
  * It can be deployed with `docker compose`
  * Port is mapped under the environment variable `REDIS_PORT`
  * Redis Insight is also available via the docker image – it can be accessed for development purposes via `localhost:8001`


src/common/cache/__tests__/test.cache.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Global, Module } from '@nestjs/common';
+import { CacheService } from '../cache.service.interface';
+import { FakeCacheService } from './fake.cache.service';
+
+@Global()
+@Module({
+  providers: [{ provide: CacheService, useClass: FakeCacheService }],
```
fmrsabino, line 7: @hectorgomezv similar to the `TestConfigurationModule` I think we should expose a test instance via `useValue`.

I've used `useClass` which works but if we want to manipulate the cache state in a test a reference to it should be provided instead. Wdyt?

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -0,0 +1,64 @@
...
+   * present tries to get the respective JSON payload from {@link url}
+   * Errors are not cached.
+   *
+   * @param key - the key to be used to retrieve from cache
+   * @param url - the HTTP endpoint to retrieve the JSON payload
+   * @param params - the parameters to be used for the HTTP request
+   * @param expireTimeSeconds - the time to live in seconds for the payload
+   * behind {@link key}
+   */
+  async get<T>(
```
hectorgomezv, line 44: I know this is a first approach, but I'm tempted to comment that currently we're not using `params` at all regarding the cache (we're just using them for calling `NetworkService`), but probably we'll need to use those `params` to build the cache key in the future. Does it make sense to move the logic to compose the key to this class, in order to derive it from url and params, instead of making the caller responsible for sending the key? 

I mean, we could remove the `key` param from the signature of this method, and just deterministically derive the key inside this class.

We can think about this later and we can evaluate pros and cons (and feasibility) but it seems a good idea to decouple things and remove the responsibility from the caller of so it could even not know there's a cache in the next layer.

src/common/cache/__tests__/test.cache.module.ts
hectorgomezv: I'm not sure I get the point, why not use `useValue` then as we do with `TestConfigurationModule` and `TestNetworkModule`? Do you see any cons/impediments?

fmrsabino: Yes I was considering using the same setup as `TestConfigurationModule`. With `useClass` I think you cannot get to the instance which is being used in this module from the tests. That makes it more difficult to setup a specific cache state.

The "negative" part of this approach is just that you are exposing a global reference to the testing instance. It shouldn't be an issue with our current setup.

I will adjust it to be in the same format as we have `TestConfigurationModule`.

README.md
```diff
@@ -13,12 +14,20 @@ corepack enable && yarn install
...
+
+2. Start the Safe Client Gateway
+
 ```bash
 # development
 yarn run start
 
 # watch mode
-yarn run start:dev
+yarn run start:watch
```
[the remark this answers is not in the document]
fmrsabino: It is! 🙈 Will fix it!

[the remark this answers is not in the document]
fmrsabino: Fixed here [b5c0180](https://github.com/5afe/safe-client-gateway-nest/pull/58/commits/b5c0180ed5caa87ca2d274e0449ae8bd0901e171)

src/common/cache/__tests__/test.cache.module.ts
fmrsabino: Address this here: [5e993fd](https://github.com/5afe/safe-client-gateway-nest/pull/58/commits/5e993fd94c8658e4c9c8a878ca1ad723ac1c22d4)

Just like we use `jest.clearAllMocks()` to reset any mocked state between the tests we now should use `fakeCacheService.clear()` to "reset" the state of the cache between tests.

src/datasources/cache/cache.first.data.source.ts
fmrsabino: > I know this is a first approach, but I'm tempted to comment that currently we're not using `params` at all regarding the cache (we're just using them for calling `NetworkService`), but probably we'll need to use those `params` to build the cache key in the future. Does it make sense to move the logic to compose the key to this class, in order to derive it from url and params, instead of making the caller responsible for sending the key?
> I mean, we could remove the `key` param from the signature of this method, and just deterministically derive the key inside this class.
> 
> We can think about this later and we can evaluate pros and cons (and feasibility) but it seems a good idea to decouple things and remove the responsibility from the caller of so it could even not know there's a cache in the next layer.

The idea that I had in mind with `CacheFirstDataSource` was that it would be just a layer in front of our cache that fallbacks to networking. I also wrote it in a way to be feature agnostic (from the cache perspective).

Features themselves currently do know about the cache via `CacheFirstDataSource` (ie.: it's not exposed as a "generic" `DataSource` which is something that we can also consider). I think it's an interesting alternative but by removing the cache key from the feature side we are tying the key to the parameters of HTTP.

I was considering that each feature could set a cache key that might independent from any URL parameter. If we remove the key from the feature side and rely on the URL (and additional params) as a key we need to make sure that the same keys are built when we also want to manually invalidate the cache. 🤔 

I agree that we should evaluate the next steps for this as any feature (even the current ones) will be based on this 👍

hectorgomezv: Yes, invalidations were my main concern when suggesting this. Of course you have much more context than me and you know how we'll need to invalidate data. I'd say if invalidations are simple enough (data with a given key needs to be refreshed, so the key gets invalidated and that's it) we could push the key composition to the cache layer, but if invalidations could get complex (invalidate a collection of things, chaining invalidations, etc.) for me it's ok to let the services layer manage the keys and the logic. Let's discuss in the future, I'm happy with the current solution 👍

hectorgomezv (approved):

PR #59, opened 2022-08-29
Remove @nestjs/axios

fmrsabino:
- Removes @nestjs/axios and adds uses Axios directly
- We are not currently using any feature from the HttpModule of NestJS


hectorgomezv (approved):
Great work, this simplifies things! 👌🏻

PR #66, opened 2022-08-31
Validate external service calls payloads

### LINKED ISSUE #45: Validate external service calls payloads
**Summary**
Currently we consume two Safe services: Config Service and Transaction Service. To assure this services are providing the expected payloads, it would be nice to validate those into the Client Gateway.

**Technical details**
We have several options to implement this:
- [Joi](https://joi.dev/api/?v=17.6.0)
- [AJV](https://ajv.js.org/)
- [Class Validator](https://github.com/typestack/class-validator) (Nest.js recommended)

We need to do a brief research of the pros and cons of each alternative.

**Acceptance criteria**
- Define a validation layer which will validate payloads coming from external services against some object interfaces.
- Throw an error when the validation doesn't pass.

hectorgomezv:
Closes #45 

This PR aims to implement data validation when dealing with external data sources.

We evaluated several solutions for doing the validation step, and finally chose [AJV](https://ajv.js.org/), mostly due to its language-agnosticity (it relies into JSONSchema) , performance, and good documentation. It's also the most common utility to perform data validation according to NPM statistics.

The idea is to make use of `JSONSchemaType` class from AJV [to link schemas to Typescript interfaces](https://ajv.js.org/guide/typescript.html), so this way we assure the validation layer is synced to our internal type definitions.

I left some `TODO` comments in the code to address the fact that when validation fails, cache should be cleared to avoid caching malformed data coming from data sources. We probably want to address this in a separated issue because the behaviour in this scenario could be optimised in order to avoid flooding the data source (timeout before invalidation, rate-limiting, etc. could be some options there).


src/datasources/config-api/config-api.service.ts
```diff
@@ -13,20 +16,38 @@ export class ConfigApi {
...
 
   async getChains(): Promise<Page<Chain>> {
     const key = 'chains'; // TODO key is not final
-    const url = this.baseUri + '/api/v1/chains';
-    return await this.dataSource.get(key, url);
+    const url = `${this.baseUri}/api/v1/chains`;
+    const page: Page<Chain> = await this.dataSource.get(key, url);
+    const chains = page.results;
+
+    if (!chains.every((chain) => isValidChain(chain))) {
```
hectorgomezv, line 31: I took this approach where `Page` is not validated explicitly and the page contents are. This is an opinionated thing, other option could be create multiple `PageWhateverSchema` and validate the `Page` as well. I don't have a lot of expertise in JSONSchema definitions, maybe there's a way to compose the validation schema dynamically 🤔

src/balances/entities/schemas/balance.schema.ts
```diff
@@ -0,0 +1,44 @@
...
+  required: [],
+};
+
+ajv.addSchema(tokenInfoSchema, 'tokenInfo');
+
+const balanceSchema: JSONSchemaType<Balance> = {
+  type: 'object',
+  properties: {
+    tokenInfo: { $ref: 'tokenInfo' },
+    balance: { type: 'string' },
```
hectorgomezv, line 35: @fmrsabino I'm not sure about this types. In the current CGW implementation `balance`, `fiatBalance` and `fiatConversion` are all `string`:
```
"balance": "25000000000000000",
"fiatBalance": "39.03450",
"fiatConversion": "1561.38"
```
But in this service we're returning on string and two numbers:
```
"balance": "25000000000000000",
"fiatBalance": 39.0503,
"fiatConversion": 1562.01
```

Do you have more context on this? I just created the JSONSchema taking the definition in `balance.entity.ts` but maybe it's wrong 🤔 I think this could affect clients expectations. For sure this could be addressed further, when we have the common DTOs in place, but just curious about it 🙂

src/datasources/config-api/config-api.service.ts
```diff
@@ -13,20 +16,37 @@ export class ConfigApi {
...
   async getChains(): Promise<Page<Chain>> {
     const key = 'chains'; // TODO key is not final
-    const url = this.baseUri + '/api/v1/chains';
-    return await this.dataSource.get(key, url);
+    const url = `${this.baseUri}/api/v1/chains`;
+    const page: Page<Chain> = await this.dataSource.get(key, url);
+
+    if (!page?.results.every((chain) => isValidChain(chain))) {
+      // TODO: probably we want to invalidate cache at this point
+      const errors = isValidChain.errors as DefinedError[];
```
fmrsabino, line 32: Here `errors` seems to be a property of `isValidChain` (which is global).

Given a scenario where:
1. First call `isValidChain(chain_error)` should return error
2. Second call `isValidChain(chain_success)` should be successful.

Would the second call also return an error? Ie.: how are the `errors` state managed?

src/datasources/config-api/config-api.service.ts
```diff
@@ -13,20 +16,37 @@ export class ConfigApi {
...
   }
 
   async getChains(): Promise<Page<Chain>> {
     const key = 'chains'; // TODO key is not final
-    const url = this.baseUri + '/api/v1/chains';
-    return await this.dataSource.get(key, url);
+    const url = `${this.baseUri}/api/v1/chains`;
+    const page: Page<Chain> = await this.dataSource.get(key, url);
+
+    if (!page?.results.every((chain) => isValidChain(chain))) {
```
fmrsabino, line 30: `every` seems to test every element of the array. Do we want that? As in, is it important for the error collection?

We can also consider `some` if we want to exit "early".

fmrsabino (commented):
Great work 👏 

We are currently using four instances of `Ajv`. We should strive to use only one due to caching of the schemas.

One concern that I have with this is regarding modularisation. I'd say that we can keep the `app` with the Ajv instance while each module contributes with the schemas.

This can be accomplished via Nest modules. A module would only provide the `Ajv` instance which can be reused and the feature modules would compile and contribute to that module.

src/balances/entities/schemas/balance.schema.ts
fmrsabino: I think some of these fields were done due to how rational numbers were handled (internally by the service) and converting/forwarding them as string was what was established as part of the API contract.

In this case I'd say that we should keep them as `string` when serialising for the clients but keep it as a `number` on the DTO side as long as it doesn't loose precision on the value received.

src/datasources/config-api/config-api.service.ts
fmrsabino: I think having a schema for `Page` can be useful as it allow also to validate incoming/outgoing urls for the `next` and `prev`.

We can tackle this on a different PR 👍

hectorgomezv: You mean running explicitly two validations, one against `Page` schema, and another one against the `Item` schema which is wrapped into the `Page`? Or composing schemas in advance? For sure we can tackle this in a new issue/PR, I can create one 👍

src/datasources/config-api/config-api.service.ts
hectorgomezv: I think both `some` and `every` methods return immediately if the condition is met. From MDN docs:
```
The every method executes the provided callbackFn function once for each element present in the array until it finds the one where callbackFn returns a [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) value. If such an element is found, the every method immediately returns false
```

Maybe `some` is more readable in this case? Wdyt?
```
if (page?.results.some((chain) => !isValidChain(chain)))
```

src/datasources/config-api/config-api.service.ts
hectorgomezv: As I commented below, `every` exits immediately when the condition is met, so `errors` should be pointed to the failed item.

src/balances/entities/__tests__/balance.factory.ts
```diff
@@ -0,0 +1,20 @@
...
+  fiatConversion?: number,
+): Balance {
+  return <Balance>{
+    tokenInfo: tokenInfo ?? tokenInfoFactory(),
+    balance: balance ?? faker.datatype.number({ precision: 0.01 }).toString(),
+    token: token ?? faker.random.word(),
+    fiatBalance: fiatBalance ?? faker.datatype.number({ precision: 0.01 }),
+    fiatConversion: fiatConversion ?? faker.datatype.number(),
+  };
+}
```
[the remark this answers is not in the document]
hectorgomezv: It's not used, probably duplicated the balance-factory class doing some intermediate tests but it's not needed anymore, I'll delete it, thanks! 🙂

src/balances/entities/__tests__/token-info.factory.ts
```diff
@@ -0,0 +1,21 @@
...
+): TokenInfo {
+  return <TokenInfo>{
+    tokenType: tokenType ?? TokenType.Erc20,
+    address: address ?? faker.finance.ethereumAddress(),
+    decimals: decimals ?? faker.datatype.number(32),
+    symbol: symbol ?? faker.finance.currencySymbol(),
+    name: name ?? faker.random.word(),
+    logoUri: logoUri ?? faker.internet.url(),
+  };
+}
```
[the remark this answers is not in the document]
hectorgomezv: I'll delete this as well, thanks! 🙂

src/balances/entities/__tests__/balance.factory.ts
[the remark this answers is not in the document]
hectorgomezv: Fixed in [74a5df0](https://github.com/5afe/safe-client-gateway-nest/commit/74a5df00dffa2096675f783e59b96bca90c49b95)

src/balances/entities/__tests__/token-info.factory.ts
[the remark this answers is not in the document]
hectorgomezv: Fixed in [74a5df0](https://github.com/5afe/safe-client-gateway-nest/commit/74a5df00dffa2096675f783e59b96bca90c49b95)

hectorgomezv:
> Great work 👏
> 
> We are currently using four instances of `Ajv`. We should strive to use only one due to caching of the schemas.
> 
> One concern that I have with this is regarding modularisation. I'd say that we can keep the `app` with the Ajv instance while each module contributes with the schemas.
> 
> This can be accomplished via Nest modules. A module would only provide the `Ajv` instance which can be reused and the feature modules would compile and contribute to that module.
> 
> See: https://ajv.js.org/guide/managing-schemas.html#compiling-during-initialization

You're totally right about this, we need to centralise the schemas compilation and use a common instance. This leads me to think about having also `ValidationErrorFactory` as a dependency of this centralised instance instead of the specific service requesting validations.

Also, one step further would be encapsulate all the validation logic into this centralised instance, something like `ValidationService` (?) and let this class do the link between the type of the object to validate and the schema to use. Probably we could take advantage of Typescript usage on this. But maybe I'm going too far, which level of encapsulation do you think is more appropriated here @fmrsabino?

fmrsabino:
> You're totally right about this, we need to centralise the schemas compilation and use a common instance. This leads me to think about having also `ValidationErrorFactory` as a dependency of this centralised instance instead of the specific service requesting validations.
> 
> Also, one step further would be encapsulate all the validation logic into this centralised instance, something like `ValidationService` (?) and let this class do the link between the type of the object to validate and the schema to use. Probably we could take advantage of Typescript usage on this. But maybe I'm going too far, which level of encapsulation do you think is more appropriated here @fmrsabino?
> 
> PS: I'd say to do it here or in a different PR depending on the complexity of the solution we're taking.

I would need to think more about a possible solution but I was thinking along the following lines:

- Having one module called `JsonSchemaModule`
  * this is the module that provides AJV and doesn't have any knowledge about any features
- Each feature would contribute schemas to this instance on each "feature" module.
- Each feature would hold a reference to the validation functions generated

This way you get the benefits of a single instance without coupling all the schemas in this module.

Something similar to what we did with the `CacheModule`, `ConfigModule` and `NetworkModule` but without it being `@Global`. I think that each feature can still specify that they need this `JsonSchemaModule` as long as the module's AJV instance is shared across different features.

Let me know what you think!

src/datasources/config-api/config-api.service.ts
fmrsabino: Ah my bad. The way you wrote it good! No need to change it 👍 

I was just considering if we'd fail early which seems to be the case. Thanks for explaining!

src/datasources/config-api/config-api.service.ts
fmrsabino: With this I actually meant calling `isValidChain` in different contexts. Even if the first "call"/"context" results in an error.

Asking because this seems to be a "lone" `const` global reference and I'm not sure how the `errors` state is cleared.

hectorgomezv: I don't like a lot how AJV manages the errors in this case, but it attaches an `errors` property to the validation function. This `errors` is an array containing the error(s) for the item which didn't pass the validation. Since `every()` is retuning early it is guaranteed that the loop will exit and `errors` property will contain the errors for the failed item. `const errors = ...` will be called just one time, and will be freed after the function throws, I don't see any possible inconsistent state here... 🤔

hectorgomezv:
> > You're totally right about this, we need to centralise the schemas compilation and use a common instance. This leads me to think about having also `ValidationErrorFactory` as a dependency of this centralised instance instead of the specific service requesting validations.
> > Also, one step further would be encapsulate all the validation logic into this centralised instance, something like `ValidationService` (?) and let this class do the link between the type of the object to validate and the schema to use. Probably we could take advantage of Typescript usage on this. But maybe I'm going too far, which level of encapsulation do you think is more appropriated here @fmrsabino?
> > PS: I'd say to do it here or in a different PR depending on the complexity of the solution we're taking.
> 
> I would need to think more about a possible solution but I was thinking along the following lines:
> 
> * Having one module called `JsonSchemaModule`
>   
>   * this is the module that provides AJV and doesn't have any knowledge about any features
> * Each feature would contribute schemas to this instance on each "feature" module.
> * Each feature would hold a reference to the validation functions generated
> 
> This way you get the benefits of a single instance without coupling all the schemas in this module.
> 
> Something similar to what we did with the `CacheModule`, `ConfigModule` and `NetworkModule` but without it being `@Global`. I think that each feature can still specify that they need this `JsonSchemaModule` as long as the module's AJV instance is shared across different features.
> 
> Let me know what you think!

So if I'm getting it correctly, in essence you're thinking in a module which is the only one interacting with the AJV instance, and has a function which takes a JSONSchema as argument and returns a validation function associated to the JSONSchema received. 

I like this idea 🙂. The other _feature_ modules interacting with this one should send their JSONSchemas in its constructor and retrieve and store the validation functions, wdyt?

src/datasources/config-api/config-api.service.ts
hectorgomezv: More here: https://ajv.js.org/guide/getting-started.html

Screenshot 2022-09-02 at 11 45 54

fmrsabino: Awesome! That was it. So it holds the `errors` state until the next `validate` call! That was my question yeah! 👍

hectorgomezv:
@fmrsabino I've written a _façade class_ for AJV as we discussed: `JsonSchemaService`, which exposes `compile` and `addSchema` to the api services. As we commented, this forces the callers to store references to the `ValidateFunctions` returned from `compile` internally as private members, and I'm not totally happy with the result of that, to be honest.

I think the api providers should be clean and pretty easy to read (they perform simple operations) and they're getting a bit polluted with validation logic, schemas management, and validation error-handling. And this pollution will increase in the future as we will be adding more datasource endpoints with more schemas.

As I commented I think we need to hide all the interactions with AJV from callers, I was thinking in a central place, something like `DatasourceValidator`, inside `/datasources` folder which which:

1. Encapsulates all AJV interactions, all its types and interfaces. (no more `import whatever from 'ajv'` outside this class)
2. Compiles all the schemas with their dependencies, and stores a reference to the associated `ValidateFunctions`.
3. Links each `ValidateFunction` to a specific Typescript class.
4. (Thanks to 3.) It could have just one public function: `validate(data, Type): void` where `data` is an arbitrary object and `Type` is a Typescript interface. This function would just return `void` or use `ValidationErrorFactory` to create and throw an exception.
5. By having `ValidationErrorFactory` injected, the caller don't have to catch errors, just fire & forget about validation.

All this will make api services thiner and readable, and will also reduce the amount of dependencies to inject/manage in each of them. Of course this is a proposal for another issue/PR 🙂

src/datasources/errors/validation-error-factory.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  from(errors: DefinedError[]): HttpException {
+    const errPayload: HttpExceptionPayload = {
+      message: 'Validation failed',
+      code: this.VALIDATION_ERROR_CODE,
+      arguments: [],
+    };
+
+    const detail = errors.map((err) => JSON.stringify(err));
+    this.logger.error({ ...errPayload, detail });
```
fmrsabino, line 23: I believe that `JSON.stringify` can be quite an expensive operation if we have a complex/sizeable object. We are using this just to log the error. Do you have a use-case that I might be missing if we don't `stringify` it?

src/datasources/exchange-api/entities/exchange-result.entity.ts
```diff
@@ -1,5 +1,5 @@
 export interface ExchangeResult {
   success: boolean;
-  rates?: Record<string, number>;
+  rates?: object;
```
fmrsabino, line 3: Can we leave this as `Record`? We are loosing the type information which is useful when working with this class.

src/datasources/common/json-schema.service.ts
```diff
@@ -0,0 +1,20 @@
+import { Injectable } from '@nestjs/common';
+import Ajv, { JSONSchemaType, ValidateFunction } from 'ajv';
+
+@Injectable()
+export class JsonSchemaService {
```
fmrsabino, line 5: What do you think about moving this to `src/common/schema` or similar?

I don't see this as a "datasource" for the routes 🤔

src/datasources/config-api/config-api.service.ts
```diff
@@ -1,32 +1,62 @@
...
 import { CacheFirstDataSource } from '../cache/cache.first.data.source';
+import { Page } from '../../common/entities/page.entity';
+import { Chain } from '../../chains/entities';
+import { DefinedError, ValidateFunction } from 'ajv';
+import { ValidationErrorFactory } from '../errors/validation-error-factory';
+import { JsonSchemaService } from '../common/json-schema.service';
+import {
+  chainSchema,
+  nativeCurrencySchema,
+} from '../../chains/entities/schemas/chain.schema';
```
fmrsabino, line 13: This introduces a dependency between the datasource and a route.

We should have a schema for the Safe Config Service and a schema for the Safe Transaction Service. Additionally we would have a schema for the Safe Client Gateway if we want to validate the responses returned by this service.

I'd keep these schemas separate. If they do match (external service + cgw) we can add it to `/common`.

The `routes` should depend on the `datasources`. `Datasources` should depend on internal components (cache, networking). `/common` is still a baseline that can be used by both at this point. Ideally we'd have a common module for the routes only and a common module for the data sources only for better encapsulation.

src/balances/entities/schemas/balance.schema.ts
```diff
@@ -0,0 +1,38 @@
...
+      ],
+    },
+    address: { type: 'string' },
+    decimals: { type: 'number' },
+    symbol: { type: 'string' },
+    name: { type: 'string' },
+    logoUri: { type: 'string', nullable: true },
+  },
+  required: [],
+};
```
fmrsabino, line 25: `TokenInfo` is part of common as it will be shared by different routes. The schema should live also in `/common` because other routes will end up depending on it too.

src/datasources/config-api/config-api.service.ts
```diff
@@ -1,32 +1,62 @@
...
-    const url = this.baseUri + '/api/v1/chains';
-    return await this.dataSource.get(key, url);
+    const url = `${this.baseUri}/api/v1/chains`;
+    const page: Page<Chain> = await this.dataSource.get(key, url);
+
+    if (!page?.results.every((chain) => this.isValidChain(chain))) {
+      // TODO: probably we want to invalidate cache at this point
+      const errors = this.isValidChain.errors as DefinedError[];
+      throw this.validationErrorFactory.from(errors);
+    }
```
fmrsabino, line 44: We should test when a chain is not valid.

src/datasources/config-api/config-api.service.ts
```diff
@@ -1,32 +1,62 @@
...
-    const url = this.baseUri + `/api/v1/chains/${chainId}`;
-    return await this.dataSource.get(key, url);
+    const url = `${this.baseUri}/api/v1/chains/${chainId}`;
+    const data = await this.dataSource.get(key, url);
+
+    if (!this.isValidChain(data)) {
+      // TODO: probably we want to invalidate cache at this point
+      const errors = this.isValidChain.errors as DefinedError[];
+      throw this.validationErrorFactory.from(errors);
+    }
```
fmrsabino, line 58: We should test when a chain is not valid.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -17,18 +38,34 @@ export class TransactionApi {
...
         trusted: trusted,
         excludeSpam: excludeSpam,
       },
     });
+
+    if (!balances.every((balance) => this.isValidBalance(balance))) {
+      // TODO: probably we want to invalidate cache at this point
+      const errors = this.isValidBalance.errors as DefinedError[];
+      throw this.validationErrorFactory.from(errors);
+    }
```
fmrsabino, line 52: We should test when `balance` is not valid.

fmrsabino (changes requested):

src/datasources/exchange-api/entities/exchange-result.entity.ts
hectorgomezv: I tried, but I wasn't able to keep the `Record` type and make the `JSONSchemaType` at `exchange.result.schema` compatible, there isn't a lot of documentation regarding that integration. I didn't want to get this feature blocked due to this, so maybe could we open an issue to tackle this? What do you think?

src/datasources/common/json-schema.service.ts
hectorgomezv: I left this into the `datasources` tree because I thought this validation tools would be used just to validate external services (APIs) data. But I'm open to use this to validate client requests instead of [Nest standard validation pipes](https://docs.nestjs.com/techniques/validation#validation). In that case it makes sense to move this outside `datasources`, of course.

hectorgomezv: Fixed in [dc0377f](https://github.com/5afe/safe-client-gateway-nest/commit/dc0377fa957d443f10ee5d31c65f3da203e40628)

src/datasources/errors/validation-error-factory.ts
hectorgomezv: Fixed in [dc0377f](https://github.com/5afe/safe-client-gateway-nest/commit/dc0377fa957d443f10ee5d31c65f3da203e40628)

src/datasources/config-api/config-api.service.ts
hectorgomezv: Fixed in [84c402f](https://github.com/5afe/safe-client-gateway-nest/commit/84c402fcafe4e82ec8817e0092b9fd35427ec1a3) (I moved the schema to the `datasources` layer as we discussed, maybe we could give this a second thought when we take a final decision on layers architecture as we discussed 🙂)

src/datasources/exchange-api/entities/exchange-result.entity.ts
hectorgomezv: Commented out in [84c402f](https://github.com/5afe/safe-client-gateway-nest/commit/84c402fcafe4e82ec8817e0092b9fd35427ec1a3).

I plan to give this a second chance when tacking [this](https://github.com/5afe/safe-client-gateway-nest/issues/72) 🙂

src/datasources/config-api/config-api.service.ts
hectorgomezv: Sorry I forgot to say this in our prior conversation but this class doesn't have any test yet. I'm adding these tests to the controllers in [1a50781](https://github.com/5afe/safe-client-gateway-nest/pull/66/commits/1a5078151be359884f206f1691f9feef39010174) 🙂

src/datasources/transaction-api/transaction-api.service.ts
hectorgomezv: Added controller tests in [1a50781](https://github.com/5afe/safe-client-gateway-nest/pull/66/commits/1a5078151be359884f206f1691f9feef39010174)

src/datasources/config-api/entities/chain.entity.ts
```diff
@@ -1,9 +0,0 @@
-import { NativeCurrency } from './native.currency.entity';
-
-export interface Chain {
-  chainId: string;
-  chainName: string;
-  transactionService: string;
-  vpcTransactionService: string;
-  nativeCurrency: NativeCurrency;
-}
```
fmrsabino, line 9: Why was this moved out of `datasource`?

fmrsabino (commented):
My impression from the sync yesterday is that we'd keep the routes depending on the datasources but not the other way around.

Some entities were moved out of datasource into the routes. This creates future coupling to routes that depend on those entities.


Eg.: `NativeCurrency` is now on the `chains` route. If there's any other route that depends on `NativeCurrency` it'd need to depend on the `chains` route.

src/datasources/config-api/entities/chain.entity.ts
hectorgomezv: In [c8e5080](https://github.com/5afe/safe-client-gateway-nest/commit/c8e5080c1d7a313fc6a1b7d32441120abd78cc38) I moved all the schemas and also the shared entities to `datasources` 👍

fmrsabino (changes requested):

src/chains/entities/__tests__/backbone.factory.ts
```diff
@@ -1,5 +1,5 @@
 import { faker } from '@faker-js/faker';
-import { Backbone } from '../backbone.entity';
+import { Backbone } from '../../../datasources/transaction-api/entities/backbone.entity';
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [cd975e9](https://github.com/5afe/safe-client-gateway-nest/pull/66/commits/cd975e914d916bec34f6591d5f2c22594b39ce7f)

src/datasources/exchange-api/entities/fiat-codes-result.entity.ts
```diff
@@ -1,4 +1,4 @@
 export interface FiatCodesExchangeResult {
   success: boolean;
-  symbols: Record<string, string>;
+  symbols: object;
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [cd975e9](https://github.com/5afe/safe-client-gateway-nest/pull/66/commits/cd975e914d916bec34f6591d5f2c22594b39ce7f)

src/datasources/transaction-api/entities/schemas/balance.schema.ts
```diff
@@ -0,0 +1,15 @@
+import { JSONSchemaType } from 'ajv';
+import { Balance } from '../../../../balances/entities/balance.entity';
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [cd975e9](https://github.com/5afe/safe-client-gateway-nest/pull/66/commits/cd975e914d916bec34f6591d5f2c22594b39ce7f)

fmrsabino (approved):

fmrsabino (approved):
🚀 🚀 🚀

PR #70, opened 2022-09-02
Add support for cursor pagination

### LINKED ISSUE #62: Add support for cursor query parameter
- Current paginated endpoints should have a `cursor` parameter (as implemented in the current version of the Safe Client Gateway).
- We should add support for it while mapping the contents of the cursor to the correct pagination query parameters of the Safe Transaction Service and the Safe Config Service.

Reference implementation:
- https://github.com/safe-global/safe-client-gateway/blob/main/src/common/models/page.rs
- https://github.com/safe-global/safe-client-gateway/blob/main/src/common/converters/page_metadata.rs

fmrsabino:
Closes #62

- Adds two decorators: `@RouteUrlDecorator` and `@PaginationDataDecorator`
  * `@RouteUrlDecorator` can be used to get the route URL (with query parameters)
  * `@PaginationDataDecorator` can be used to retrieve the pagination data associated with a `cursor` query param
- We have two ways of representing pagination data:
  * With `limit&offset` – this is mostly used by the Safe Config Service and Safe Transaction Service. `PaginationData` can be retrieved from these URLs with `PaginationData.fromLimitAndOffset`
  * With `cursor` – this is mostly used by the Safe Client Gateway. It encodes the `limit&offset` which can be forwarded to the other services. `PaginationData` can be retrieved from these URLs with `PaginationData.fromCursor`

### DIFF
--- a/src/common/decorators/pagination.data.decorator.spec.ts
+++ b/src/common/decorators/pagination.data.decorator.spec.ts
@@ -1,0 +1,55 @@
+import { PaginationDataDecorator } from './pagination.data.decorator';
+import { PaginationData } from '../pagination/pagination.data';
+import { Controller, Get, INestApplication, Module } from '@nestjs/common';
+import { fakeCacheService } from '../cache/__tests__/test.cache.module';
+import { Test, TestingModule } from '@nestjs/testing';
+import * as request from 'supertest';
+
+describe('PaginationDataDecorator', () => {
+  let app: INestApplication;
+  let paginationData;
+
+  @Controller()
+  class TestController {
+    @Get()
+    route(@PaginationDataDecorator() data?: PaginationData) {
+      paginationData = data;
+      return;
+    }
+  }
+
+  @Module({ controllers: [TestController] })
+  class TestModule {}
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+    fakeCacheService.clear();
+
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [TestModule],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+    await app.init();
+  });
+
+  it('no cursor', async () => {
+    await request(app.getHttpServer())
+      .get('')
+      .query('some_param=test')
+      .expect(200);
+
+    expect(paginationData.limit).toBe(undefined);
+    expect(paginationData.offset).toBe(undefined);
+  });
+
+  it('with cursor', async () => {
+    await request(app.getHttpServer())
+      .get('')
+      .query('cursor=limit%3D1%26offset%3D0')
+      .expect(200);
+
+    expect(paginationData.limit).toBe(1);
+    expect(paginationData.offset).toBe(0);
+  });
+});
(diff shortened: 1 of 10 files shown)


src/common/pagination/pagination.data.ts
```diff
@@ -0,0 +1,88 @@
...
+   * not exist
+   *
+   * @param url - url which contains the cursor query param
+   */
+  static fromCursor(url: Readonly<URL>): PaginationData {
+    // If we don't have a cursor, assume no pagination data
+    const cursorValue = url.searchParams.get(QUERY_PARAM_CURSOR) ?? '';
+    const cursorSearchParams = new URLSearchParams(cursorValue);
+    const limit = Number(cursorSearchParams.get(QUERY_PARAM_LIMIT) ?? NaN);
+    const offset = Number(cursorSearchParams.get(QUERY_PARAM_OFFSET) ?? NaN);
```
fmrsabino, line 23: I'm defaulting to `NaN` here as a `Number` of a an empty string is `0`. This results in incorrect `limit` and `offset` values (where `0` should be different than no value)

src/common/decorators/pagination.data.decorator.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+describe('PaginationDataDecorator', () => {
+  let app: INestApplication;
+  let paginationData;
+
+  @Controller()
+  class TestController {
+    @Get()
+    route(@PaginationDataDecorator() data?: PaginationData) {
+      paginationData = data;
```
hectorgomezv, line 16: Nitpick: maybe would be better to return the `PaginationData` instead using a class variable? Not a big deal here because there're only 2 simple tests, but I think it's a general best practice, and that will make the route a pure function with no side effects.

src/common/pagination/pagination.data.ts
hectorgomezv: Perfect! JS is pretty weird sometimes 🙂

src/datasources/config-api/config-api.service.ts
```diff
@@ -18,10 +18,15 @@ export class ConfigApi {
...
   }
 
-  async getChains(): Promise<Page<Chain>> {
-    const key = 'chains'; // TODO key is not final
+  async getChains(limit?: number, offset?: number): Promise<Page<Chain>> {
+    const key = `chains-limit=${limit}-offset=${offset}`; // TODO key is not final
     const url = this.baseUri + '/api/v1/chains';
-    return await this.dataSource.get(key, url);
+    return await this.dataSource.get(key, url, {
+      params: {
```
hectorgomezv, line 25: NP: I think this could be simplyfied (__Shorthand Properties__): 
```
return await this.dataSource.get(key, url, { params: { limit, offset } });
```

hectorgomezv (approved):
Great work! 🚀

src/common/decorators/pagination.data.decorator.spec.ts
fmrsabino: I do agree but that leaves me with another question: we'd end up testing an HTTP response instead of the actual `PaginationData` object. My goal with these tests was to test if the decorator was working as intended and not to test the result of wrapping pagination data in the body of an HTTP response.

If we return from the route we end up with the following assertions (do notice the `body`):

```typescript
expect(actual.body.limit).toBe(undefined);
expect(actual.body.offset).toBe(undefined);
```

What do you think?

hectorgomezv (approved):

PR #74, opened 2022-09-06
Add domain layer

fmrsabino:
- Adds a new `domain` layer
- This layer as no dependencies and it is the central component of the app
- `routes` can and should depend on this layer. This layer has the following repositories which can be used: `IBackboneRepository`, `IBalancesRepository`, `IChainsRepository` and `IExchangeRepository`
- `datasources` feed data into this layer through different interfaces. More specifically: `IConfigApi`, `IExchangeApi`, `ITransactionApiManager` and `ITransactionApi`


hectorgomezv:
Great work!

As we agreed, I think these modifications would be very nice, just putting them here for the record:
- Group features and common folders inside the same directory (`routes`, `features`, whatever you prefer), to make the layers/dependency rules more explicit in the folders tree. (This could be done in another PR to avoid conflicts as you suggested).
- Group domain interfaces in an `interfaces` folder.
- Rename DomainRepository to something else, just `Repository` maybe, or split it in a repository per feature (maybe it's better as you suggested).

Thanks!! 🚀

fmrsabino:
> * Group features and common folders inside the same directory (`routes`, `features`, whatever you prefer), to make the layers/dependency rules more explicit in the folders tree. (This could be done in another PR to avoid conflicts as you suggested).

I will tackle this one in a different Pull Request 👍 

> * Group domain interfaces in an `interfaces` folder.
> * Rename `DomainRepository` to something else, just `Repository` maybe, or split it in a repository per feature (maybe it's better as you suggested).

This was done in [34af455](https://github.com/5afe/safe-client-gateway-nest/pull/74/commits/34af4554bdee0fcd6fea798a952966d430579dbd)

hectorgomezv:
Awesome! I really like having a separate `domain` layer which is not depending on anything (or using dependency injection to depend on abstractions). I'm a bit concerned about if we're introducing too much complexity, but I think at this point our design is pretty flexible and it will adapt well to the future use cases 🙂

hectorgomezv (approved):

fmrsabino:
> Awesome! I really like having a separate `domain` layer which is not depending on anything (or using dependency injection to depend on abstractions). I'm a bit concerned about if we're introducing too much complexity, but I think at this point our design is pretty flexible and it will adapt well to the future use cases 🙂

Yes this more about grouping functionality. We don't have separate entities between external services (datasources) and our domain as I think that would increase the complexity without any benefit.

So right now datasources would contribute to the domain (with the respective entities) and the routes can use those entities in whatever way they like in order to build the response.

PR #78, opened 2022-09-07
Remove TODO from ExchangeApi

fmrsabino:
As agreed on https://github.com/5afe/safe-client-gateway-nest/pull/80, we will continue injecting `IConfigurationService` (ie.: we will not inject the `BASE_URL` in the constructor)

### DIFF
--- a/src/datasources/exchange-api/exchange-api.service.ts
+++ b/src/datasources/exchange-api/exchange-api.service.ts
@@ -19,8 +19,6 @@ import { exchangeResultSchema } from './entities/schemas/exchange-result.schema'
 
 @Injectable()
 export class ExchangeApi implements IExchangeApi {
-  // TODO can we depend on the base url instead?
-
   private readonly baseUrl: string;
   private readonly apiKey: string;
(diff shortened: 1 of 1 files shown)


src/datasources/exchange-api/exchange-api.module.ts
```diff
@@ -4,10 +4,29 @@ import { ExchangeApi } from './exchange-api.service';
 import { IExchangeApi } from '../../domain/interfaces/exchange-api.interface';
 import { ValidationErrorFactory } from '../errors/validation-error-factory';
 import { JsonSchemaService } from '../../common/schemas/json-schema.service';
+import { IConfigurationService } from '../../common/config/configuration.service.interface';
+
+const baseUrlFactory = {
+  provide: 'EXCHANGE_SERVICE_BASE_URL',
```
fmrsabino, line 10: Unfortunately it seems that Nest does not support the `provide` to be a `Symbol` at this level (so I had to fallback to `string` as the identifier)

hectorgomezv (approved):
👍🏻

PR #80, opened 2022-09-07
Inject base url in ConfigApi

fmrsabino:
Adds a provider for `CONFIG_API_BASE_URL` allowing the injection of the Safe Config Service base url directly in ConfigApi

### DIFF
--- a/src/datasources/config-api/config-api.module.ts
+++ b/src/datasources/config-api/config-api.module.ts
@@ -4,11 +4,21 @@ import { CacheFirstDataSourceModule } from '../cache/cache.first.data.source.mod
 import { IConfigApi } from '../../domain/interfaces/config-api.interface';
 import { ValidationErrorFactory } from '../errors/validation-error-factory';
 import { JsonSchemaService } from '../../common/schemas/json-schema.service';
+import { IConfigurationService } from '../../common/config/configuration.service.interface';
+
+const baseUrlFactory = {
+  provide: 'CONFIG_API_BASE_URL',
+  useFactory: (configurationService: IConfigurationService) => {
+    return configurationService.getOrThrow<string>('safeConfig.baseUri');
+  },
+  inject: [IConfigurationService],
+};
 
 @Global()
 @Module({
   imports: [CacheFirstDataSourceModule],
   providers: [
+    baseUrlFactory,
     { provide: IConfigApi, useClass: ConfigApi },
     ValidationErrorFactory,
     JsonSchemaService,
--- a/src/datasources/config-api/config-api.service.ts
+++ b/src/datasources/config-api/config-api.service.ts
@@ -1,7 +1,6 @@
 import { Inject, Injectable } from '@nestjs/common';
 import { Page } from '../../domain/entities/page.entity';
 import { Chain } from '../../domain/chains/entities/chain.entity';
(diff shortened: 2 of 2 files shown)


hectorgomezv:
I'm not against this change but I'd like to know why is preferable to inject just the base URL instead of injecting the configuration service and let the `ConfigApi` pick whatever it needs from it. Using the proposed solution we'll be creating a `whateverPropertyFactory` into the module definition for every configuration param the module need?

fmrsabino:
> I'm not against this change but I'd like to know why is preferable to inject just the base URL instead of injecting the configuration service and let the `ConfigApi` pick whatever it needs from it. Using the proposed solution we'll be creating a `whateverPropertyFactory` into the module definition for every configuration param the module need?

I agree. The other take that I had on this is that the Base URL is such a core component for this service that having it explicitly set in the constructor would make it easier to build one without depending on a local configuration (imagine removing the local configuration dependency in a testing scenario).

I have no strong opinion with this one. If we rely more on the local configuration we can also have both the Base URL and the `ConfigurationService` injected. Or we just leave the `ConfigurationService` and get the base url in the constructor (as we have right now) 👍 

Do you have any take on this @Uxio0 ?

fmrsabino:
> > > I'm not against this change but I'd like to know why is preferable to inject just the base URL instead of injecting the configuration service and let the `ConfigApi` pick whatever it needs from it. Using the proposed solution we'll be creating a `whateverPropertyFactory` into the module definition for every configuration param the module need?
> > 
> > 
> > I agree. The other take that I had on this is that the Base URL is such a core component for this service that having it explicitly set in the constructor would make it easier to build one without depending on a local configuration (imagine removing the local configuration dependency in a testing scenario).
> > I have no strong opinion with this one. If we rely more on the local configuration we can also have both the Base URL and the `ConfigurationService` injected. Or we just leave the `ConfigurationService` and get the base url in the constructor (as we have right now) 👍
> > Do you have any take on this @Uxio0 ?
> 
> Another option, if you don't want to inject the `ConfigurationService`, could be define a `ConfigApiConfiguration` interface and populate it into the factory you defined in the module, instead of just building `baseUri`, build a `ConfigApiConfiguration`:
> 
> ```
> const configurationFactory = {
>   provide: 'CONFIG_API_CONFIGURATION',
>   useFactory: (configurationService: IConfigurationService) => {
>     return <ConfigApiConfiguration> {
>       baseUri: configurationService.getOrThrow<string>('safeConfig.baseUri'),
>       // rest of configuration needed by the component
>     };
>   },
>   inject: [IConfigurationService],
> };
> ```
> 
> But maybe we're introducing too much complexity 🤔

I do prefer either leaving it as it is or injecting the `BASE_URL` 😅 . Both seem good options. The `TransactionApi` relies on the `baseUrl` because that can be provided without the configuration though which can make it "weird" that the other `Api` on the same level rely on the `ConfigurationService` for this.

But I don't think this is critical or that it should change. I changed it here because it was easy enough to do so 😄

Uxio0:
I agree with @hectorgomezv , I think this is introducing a lot of complexity, both for testing (need to mock all the dependencies) and I don't see any advantage, as if we inject the `ConfigService` and there's something not defined the service will fail fast

fmrsabino:
I will close this PR then! Thanks for the feedback 🙌

Issue #72, opened 2022-09-05
Centralise datasource data validations logic


hectorgomezv:
More context [here](https://github.com/5afe/safe-client-gateway-nest/pull/66#pullrequestreview-1093150104).

**Summary**
We want to centralise data validation on the domain layer in a single place per repository. 

Currently we are using a façade class for AJV (our validation utility): `JsonSchemaService`, which exposes `compile` and `addSchema` functions to the repositories. This forces the callers to store references to each `ValidateFunction` returned from `compile` internally as private members. This _pollution_ will increase in the future as we will be adding more repository methods with more schemas.

**Acceptance criteria**
We need to hide all the interactions with AJV from callers, so we need to create a `Validator` **per repository**, which:

- Encapsulates all AJV interactions, all its types and interfaces. (no more `import whatever from 'ajv'` outside these classes)
- Compiles all the schemas with their dependencies regarding the repository, and stores a reference to the associated `ValidateFunctions`.
- Links each `ValidateFunction` to a specific `Typescript` class (if possible).
- (Thanks to the previous point, if feasible) It has just one public function: `validate(data, Type): void` where data is an arbitrary object and `Type` is a Typescript interface. This function would just return void or use `ValidationErrorFactory` to create and throw an exception.
- By having `ValidationErrorFactory` injected, the caller don't have to catch errors, just fire & forget about validation.

Issue #64, opened 2022-08-31
Add unit tests to ExchangeRepository


hectorgomezv:
Add unit testing to ExchangeRepository. Something similar to the tests for other repositories we have should work. These tests should check all the logic contained in the repository, including validation logic.

PR #82, opened 2022-09-08
Move validation logic to domain layer

hectorgomezv:
This PR moves all the validation logic to the `domain` layer, so:
- API components have less responsibilities now, they just pick data from their data sources.
- Repositories have to validate the data, picking this responsibility from API components.
- Tests were moved accordingly.
- I added required properties to schemas, so validation is a bit stronger now.

Also I edited #64 to adapt the goal to this change.

And apart from that, keep in mind this validation will be even more encapsulated and transparent to repositories when we implement #72


src/datasources/config-api/config-api.module.ts
```diff
@@ -2,7 +2,7 @@ import { Global, Module } from '@nestjs/common';
 import { ConfigApi } from './config-api.service';
 import { CacheFirstDataSourceModule } from '../cache/cache.first.data.source.module';
 import { IConfigApi } from '../../domain/interfaces/config-api.interface';
-import { ValidationErrorFactory } from '../errors/validation-error-factory';
+import { ValidationErrorFactory } from '../../domain/errors/validation-error-factory';
```
fmrsabino, line 5: Do we need the `ValidationErrorFactory` here?

src/datasources/exchange-api/exchange-api.module.ts
```diff
@@ -2,7 +2,7 @@ import { Global, Module } from '@nestjs/common';
 import { HttpErrorFactory } from '../errors/http-error-factory';
 import { ExchangeApi } from './exchange-api.service';
 import { IExchangeApi } from '../../domain/interfaces/exchange-api.interface';
-import { ValidationErrorFactory } from '../errors/validation-error-factory';
+import { ValidationErrorFactory } from '../../domain/errors/validation-error-factory';
```
fmrsabino, line 5: Do we need the `ValidationErrorFactory` here?

src/datasources/transaction-api/transaction-api.module.ts
```diff
@@ -2,7 +2,7 @@ import { Global, Module } from '@nestjs/common';
 import { TransactionApiManager } from './transaction-api.manager';
 import { CacheFirstDataSourceModule } from '../cache/cache.first.data.source.module';
 import { ITransactionApiManager } from '../../domain/interfaces/transaction-api.manager.interface';
-import { ValidationErrorFactory } from '../errors/validation-error-factory';
+import { ValidationErrorFactory } from '../../domain/errors/validation-error-factory';
```
fmrsabino, line 5: Do we need the `ValidationErrorFactory` here?

src/balances/balances.controller.spec.ts
```diff
@@ -118,31 +118,6 @@ describe('Balances Controller (Unit)', () => {
...
-          .expect(500)
-          .expect({
-            message: 'Validation failed',
-            code: 42,
-            arguments: [],
-          });
-
-        expect(mockNetworkService.get.mock.calls.length).toBe(1);
-      });
-    });
```
fmrsabino, line 144: I still find this test valuable but given the introduction of the schema validation I would say that this should test instead what happens when the Config Service returns a `500`.

src/domain/backbone/backbone.repository.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+  getBackbone: jest.fn().mockResolvedValue(BACKBONE),
+} as unknown as ITransactionApi;
+
+const transactionApiManager = {
+  getTransactionApi: jest.fn().mockResolvedValue(transactionApi),
+} as unknown as ITransactionApiManager;
+
+const validationErrorFactory = {
+  from: jest.fn().mockReturnValue(new HttpException('testErr', 500)),
+} as unknown as ValidationErrorFactory;
```
fmrsabino, line 23: I know that we have some tests with similar setup but what is the advantage of mocking the return value here?

In my opinion it's usually more readable if the in test you see the setup, action, assertions. In this case part of the setup happens outside of the test.

src/domain/backbone/backbone.repository.ts
```diff
@@ -1,16 +1,36 @@
...
+    ) as ValidateFunction<Backbone>;
+  }
 
   async getBackbone(chainId: string): Promise<Backbone> {
     const api = await this.transactionApiManager.getTransactionApi(chainId);
-    return api.getBackbone();
+    const backbone = await api.getBackbone();
+
+    if (!this.isValidBackbone(backbone)) {
+      // TODO: probably we want to invalidate cache at this point
```
fmrsabino, line 29: This can now be more tricky. Since at this level you wouldn't know that there's a `cache`.

src/datasources/config-api/config-api.module.ts
hectorgomezv: We don't! Thanks for noticing! 🙂

Fixed in [d376ff0](https://github.com/5afe/safe-client-gateway-nest/pull/82/commits/d376ff082174dc6624e1eeb6ec396f4bfbc36045)

src/datasources/exchange-api/exchange-api.module.ts
hectorgomezv: We don't! Thanks for noticing! 🙂

Fixed in [d376ff0](https://github.com/5afe/safe-client-gateway-nest/pull/82/commits/d376ff082174dc6624e1eeb6ec396f4bfbc36045)

src/datasources/transaction-api/transaction-api.module.ts
hectorgomezv: We don't! Thanks for noticing! 🙂

Fixed in [d376ff0](https://github.com/5afe/safe-client-gateway-nest/pull/82/commits/d376ff082174dc6624e1eeb6ec396f4bfbc36045)

src/balances/balances.controller.spec.ts
hectorgomezv: Sure! I added another test in [d376ff0](https://github.com/5afe/safe-client-gateway-nest/pull/82/commits/d376ff082174dc6624e1eeb6ec396f4bfbc36045)

src/domain/backbone/backbone.repository.spec.ts
hectorgomezv: I agree, I usually set here the "default" behaviour to avoid repeating it in each test, but maybe it's better to have it defined per test, specially here where we're just mocking one function. 👍🏻

I changed it for this test class in [d376ff0](https://github.com/5afe/safe-client-gateway-nest/pull/82/commits/d376ff082174dc6624e1eeb6ec396f4bfbc36045), we can start using this convention when modifying test classes in other PRs 🙂

src/domain/backbone/backbone.repository.ts
hectorgomezv: I agree, in fact I removed these comments in my other PR for the same reason. Removing it here as well! 🙂

Fixed in [d376ff0](https://github.com/5afe/safe-client-gateway-nest/pull/82/commits/d376ff082174dc6624e1eeb6ec396f4bfbc36045)

fmrsabino: It's definitely something that we should consider. The invalidation can still happen on the `datasources` side but triggering it from `domain` needs some adjustment.

One thing that we need to implement are the invalidation hooks (route level) – so at some point you'd be able to invalidate the cache from `routes`

src/domain/backbone/backbone.repository.spec.ts
fmrsabino: > I agree, I usually set here the "default" behaviour to avoid repeating it in each test, but maybe it's better to have it defined per test, specially here where we're just mocking one function. 👍🏻

I do think that repetition in tests is ok – so that you can look at a single test and understand the setup and the assertions just from the test block.

If someone wants to write more test cases they can assume a "clean" setup for the test (ie.: no assumptions).

fmrsabino (approved):
Great work! 🚀 🚀

PR #83, opened 2022-09-08
Stop Redis connection from preventing shutting down the service

### LINKED ISSUE #73: Shutdown hook not working while trying to connect to Redis
When Redis is not available the service retries to connect repeatedly. After sending a Control-C (SIGTERM) the process still running and trying to connect to the Redis server.

We are currently using  Nest hooks, we need to investigate why they is not shutting down the process when SIGTERM happens.

```
  // Starts listening for shutdown hooks
  app.enableShutdownHooks();
```

Image

hectorgomezv:
Closes #73 

This PR aims to prevent Redis connection retries from stopping service shutdown.

I think the main problem was the service used `client.quit()` to gracefully disconnect from Redis instance, and this forces the app to wait until the Redis instance acknowledges the connection end. The problem arises when the Redis instances is not reachable for any reason (socket issues, the instance is not running, etc.), there is a [similar issue open in the node-redis repository](https://github.com/redis/node-redis/issues/2205).

So the solution I implemented is to define a timeout which will fire a `client.disconnect()` if the instance doesn't respond in time. I've set a 2 seconds grace period for the server to respond, but we could change this if you think so, or even we could extract this into a configuration param/env variable.


src/datasources/cache/redis.cache.service.ts
```diff
@@ -42,11 +43,26 @@ export class RedisCacheService implements ICacheService, OnModuleDestroy {
...
+    clearTimeout(forceQuitTimeout);
+    this.logger.verbose('Redis connection closed');
+  }
+
+  /**
+   * Forces the closing of the Redis connection associated with this service.
+   */
+  private forceQuit() {
+    this.logger.verbose('Forcing Redis connection close');
+    this.client.disconnect();
```
fmrsabino, line 65: `disconnect` returns a `Promise` shouldn't we make `forceQuit` async? How does that play with the `bind()` in `setTimeout`?

hectorgomezv: Yes, maybe it's more correct to define it as `async` and wait for the Promise to be resolved. Solved [here.](https://github.com/5afe/safe-client-gateway-nest/pull/83/commits/968af6e807c8c9e3da58ec472ae40d3512573e83) Thanks.

`bind()` creates a new function but doing a closure over `this` (i.e. save the context of the caller). Otherwise `setTimeout` callback (`forceQuit` function) won't have access to the context when fired. This is necessary to still having access to `this.client` when invoking `disconnect`.

fmrsabino: Do you know about: https://nodejs.org/api/timers.html#timers-promises-api

I did a quick mockup with the following:

```typescript
import { setTimeout } from 'timers/promises';

...

async onModuleDestroy(): Promise<void> {
  const abortController = new AbortController();
  this.logger.verbose('Closing Redis connection');

  (async () => {
    await setTimeout(this.quitTimeoutInSeconds * 1000, this.forceQuit(), {
      signal: abortController.signal,
    });
  })();

  await this.client.quit();
  abortController.abort();
  this.logger.verbose('Redis connection closed');
}
```

fmrsabino (approved):

PR #84, opened 2022-09-08
Add swagger

### LINKED ISSUE #76: Add support for Swagger
- Swagger support should be added
- The Json Schema for the response types should be correctly generated

fmrsabino:
Closes #76

- Adds `@nestjs/swagger` – a framework which helps in building and providing a Swagger interface for NestJS
- Swagger is accessible in the root path of the server
- Adds `openapi` folders to `routes`
  * Here is where you can find the API representations that will be used in Swagger
  * They implement the respective domain interface and add the provided decorators required for generating Swagger stubs

### DIFF
--- a/src/common/openapi/api-token-info.ts
+++ b/src/common/openapi/api-token-info.ts
@@ -1,0 +1,18 @@
+import { TokenInfo as DomainTokenInfo } from '../entities/token-info.entity';
+import { ApiProperty } from '@nestjs/swagger';
+import { TokenType } from '../entities/token-type.entity';
+
+export class TokenInfo implements DomainTokenInfo {
+  @ApiProperty()
+  address: string;
+  @ApiProperty()
+  decimals: number;
+  @ApiProperty()
+  logoUri: string;
+  @ApiProperty()
+  name: string;
+  @ApiProperty()
+  symbol: string;
+  @ApiProperty({ enum: Object.values(TokenType) })
+  tokenType: TokenType;
+}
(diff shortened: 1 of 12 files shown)


src/common/openapi/api-token-info.ts
```diff
@@ -0,0 +1,18 @@
+import { TokenInfo as DomainTokenInfo } from '../entities/token-info.entity';
+import { ApiProperty } from '@nestjs/swagger';
+import { TokenType } from '../entities/token-type.entity';
+
+export class TokenInfo implements DomainTokenInfo {
```
fmrsabino, line 5: I would have preferred to have `ApiTokenInfo` or similar – that would also remove the import alias `as DomainTokenInfo` (we could just use `TokenInfo`).

However it seems that `@nestjs/swagger` generates the name of the stubs based on the class name (I could not find a decorator on the class level to provide a name for it).

src/routes/chains/openapi/api-chain-page.ts
```diff
@@ -0,0 +1,16 @@
...
+  @ApiProperty()
+  count: number;
+  @ApiProperty()
+  next: string;
+  @ApiProperty()
+  previous: string;
+  @ApiProperty({
+    type: ApiChain,
+  })
+  results: ApiChain[];
```
fmrsabino, line 15: In order to correctly generate the stubs for a `Page` (which is polymorphic) we need to specify the types that can be part of `Page`.

Since we are taking a modular approach I've created a `Page` just for `Chain`. The alternative would be to have a `Page` that would set here all possible types that we expose with `Page`.

This will result in more repeated classes but I see these classes purely for documentation purposes (ie.: they are not used outside of our routes).

hectorgomezv (approved):
Great! 🚀

PR #85, opened 2022-09-09
72 centralise datasource data validations logic

hectorgomezv:
Depends on #82 

This PR aims to centralise the get away validation internals from repositories. To achieve that, I've done several things:
- I created a generic `IValidator<T>` interface to define a common contract to all validators.
- I moved all the validation logic to `Validator` classes, which will be responsible of registering `Schemas` and managing `ValidationFunctions`.
- Those `Validators` will also throw validation errors.
- I adapted testing classes to this new flow.

Note: `ExchangeValidator` is the most interesting because it handles several schemas, so we're hiding from the caller (repository) all the logic regarding linking types and schemas. Thats the reason I refactored a bit the interfaces into the `Exchange` domain and I created the `ExchangeResult` hierarchy and the type guards, to get advantage of Typescript capabilities. I think this design is pretty extensible to multiple schemas and type, so just one `Validator` pero domain would suffice.


src/domain/backbone/backbone.repository.spec.ts
```diff
@@ -18,53 +16,27 @@ const transactionApiManager = {
...
-
-const validationFunction = jest.fn();
-validationFunction.mockImplementation(() => true);
-
-const jsonSchemaService = {
-  addSchema: jest.fn(),
-  compile: jest.fn().mockImplementation(() => validationFunction),
-} as unknown as JsonSchemaService;
+const backboneValidator = {
+  validate: jest.fn().mockResolvedValue(BACKBONE),
```
fmrsabino, line 20: nitpick: from a previous discussion I was under the impression that we'd not mock at this level but inside each test right?

src/domain/interfaces/validator.interface.ts
```diff
@@ -0,0 +1,5 @@
+export interface IValidator<T> {
+  validate(data: unknown): T;
+  // TODO: deprecate validateMany in favor of: validatePage(data: unknown): Page<T>
+  validateMany(data: unknown[]): T[];
```
fmrsabino, line 4: When you mention to replace `validateMany` with `validatePage` does it mean at this level on the interface?

`validateMany` seems to put push functionality implementation to some validators even if it's not required. Eg.: Validating `T[]` would be done not by the Validator of `T` but by the Validator of `[]` (ArrayValidator). This way we can decouple `T` from Collection types when it comes to validation.

src/domain/exchange/entities/exchange-result.entity.ts
```diff
@@ -1,6 +1,3 @@
 export interface ExchangeResult {
   success: boolean;
```
fmrsabino, line 2: I think having an interface with just `success` for two payloads is not required and might make some parts of the code more difficult to read (eg.: validators where we distinguish by type).

My suggestion would be to have this field set in `RatesExchangeResult` and `FiatCodesExchangeResult`.

Then we have a validator for `RatesExchangeResult` and one for `FiatCodesExchangeResult`. Wdyt?

src/domain.module.ts
```diff
@@ -21,6 +25,10 @@ import { JsonSchemaService } from './common/schemas/json-schema.service';
     { provide: IBalancesRepository, useClass: BalancesRepository },
     { provide: IChainsRepository, useClass: ChainsRepository },
     { provide: IExchangeRepository, useClass: ExchangeRepository },
+    BackboneValidator,
+    BalancesValidator,
+    ExchangeValidator,
+    ChainsValidator,
```
fmrsabino, line 31: Maybe we need to consider a module for each "sub-domain" I can expect that the list is only going to grow 😅

hectorgomezv: I totally agree, will address this in a separate PR if you are ok with it 🙂

src/domain/interfaces/validator.interface.ts
hectorgomezv: Agree, I simplified the interface to have just a `validate` function so the caller keeps the responsibility to iterate the validation targets if needed.

src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -1,17 +1,21 @@
 import { ExchangeApi } from './exchange-api.service';
-import exchangeResultFactory from '../../domain/exchange/entities/__tests__/exchange.factory';
-import exchangeFiatCodesFactory from '../../domain/exchange/entities/__tests__/fiat-codes.factory';
+import ratesExchangeResultFactory from '../../domain/exchange/entities/__tests__/rates-exchange-result.factory';
+import fiatCodesExchangeResultFactory from '../../domain/exchange/entities/__tests__/fiat-codes-exchange-result.factory';
```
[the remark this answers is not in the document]
hectorgomezv: From my understanding `Exchange` is a data provider, which provides several data types. Having one of the data types named simply `ExchangeResult` it's not very expressive. I think both response data type are `ExchangeResults`, one returns rates and the other returns fiat codes. That's why I named them using the same logic. Wdyt? I'm open to suggestions :)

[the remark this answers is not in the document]
moisses89: Agree but maybe `ratesExchangeFactory` and `fiatCodesExchangeFactory` ? Result is implicit.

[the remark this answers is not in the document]
hectorgomezv: I renamed several classes/functions using `ExchangeRates` and `ExchangeFiatCodes` as entity names so we can get rid of the `Result` part 👍

src/domain/backbone/backbone.validator.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+} as unknown as ValidationErrorFactory);
+
+const validationFunction = jest.fn();
+const mockJsonSchemaService = jest.mocked({
+  addSchema: jest.fn(),
+  compile: jest.fn().mockImplementation(() => validationFunction),
+} as unknown as JsonSchemaService);
+
+describe('Backbone validator', () => {
+  const validator = new BackboneValidator(
```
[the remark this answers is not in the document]
hectorgomezv: I agree in some cases this could be useful, but I think doing this on a regular basis is not necessary and could hurt performance while running tests (for N test cases, the class under test will be built N times). I assume the classes under test (services, repositories, validators, whatever) should be stateless in most cases. I'd even say having internal states in that classes could be a code smell.

src/domain/balances/balances.repository.spec.ts
```diff
@@ -1,62 +1,46 @@
...
+const mockBalancesValidator = jest.mocked({
+  validate: jest.fn(),
+} as unknown as BalancesValidator);
 
 describe('Balances Repository', () => {
   const repository = new BalancesRepository(
     mockTransactionApiManager,
-    mockValidationErrorFactory,
-    mockJsonSchemaService,
+    mockBalancesValidator,
```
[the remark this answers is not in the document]
hectorgomezv: Please refer to [the previous comment](https://github.com/5afe/safe-client-gateway-nest/pull/85#discussion_r988867637).

src/domain/balances/balances.validator.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+  compile: jest.fn().mockImplementation(() => validationFunction),
+} as unknown as JsonSchemaService);
+
+describe('Balances validator', () => {
+  const validator = new BalancesValidator(
+    mockValidationErrorFactory,
+    mockJsonSchemaService,
+  );
+
+  const balances = [balanceFactory(), balanceFactory()];
```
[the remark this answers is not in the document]
hectorgomezv: Changed! Thanks!

src/domain/chains/chains.repository.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Chain } from './entities/chain.entity';
+import { Page } from '../entities/page.entity';
+import { MasterCopyValidator } from './master-copy.validator';
+import { ITransactionApiManager } from '../interfaces/transaction-api.manager.interface';
+
+const CHAIN = chainFactory();
+const CHAINS: Page<Chain> = {
+  count: faker.datatype.number(),
+  results: [chainFactory(), chainFactory()],
+};
```
moisses89, line 15: Should be in the test

src/domain/chains/chains.repository.spec.ts
```diff
@@ -0,0 +1,63 @@
...
+    mockChainValidator.validate = jest
+      .fn()
+      .mockResolvedValue(CHAINS.results[0]);
+
+    const data = await repository.getChains();
+
+    expect(mockChainValidator.validate).toBeCalledTimes(CHAINS.results.length);
+    expect(data).toBe(CHAINS);
+  });
+});
```
moisses89, line 63: Could you add a test for mastercopies?

src/domain/chains/chains.validator.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+} as unknown as JsonSchemaService);
+
+describe('Chains validator', () => {
+  const validator = new ChainsValidator(
+    mockValidationErrorFactory,
+    mockJsonSchemaService,
+  );
+
+  const chain = chainFactory();
+  const chains = [chainFactory(), chainFactory()];
```
moisses89, line 24: The same here we should move this to the test

moisses89 (commented):
Good work! this is going to simplify the validation implementation :) 
Only a few comments

src/domain/chains/chains.validator.spec.ts
hectorgomezv: Done in [ae890e7b0a9f73e8ea70d1ed2b7bf12f2109481a](https://github.com/5afe/safe-client-gateway-nest/commit/ae890e7b0a9f73e8ea70d1ed2b7bf12f2109481a)

moisses89 (approved):
Looks good for me :rocket:

PR #90, opened 2022-09-15
Add /hooks/events route

### LINKED ISSUE #86: Implement cache invalidation route
- The following route should be implemented: `/v1/chains/<chain_id>/hooks/events`.
- The Safe Transaction Service depends on this route. Any update done regarding payload should be adjusted on the Safe Transaction Service side.
- The following payload should be used: https://github.com/safe-global/safe-client-gateway/blob/da658a17153d88ce4964bf3889d70ae0bf79ddba/src/common/models/backend/hooks.rs#L3-L11
- Authentication is not required for the initial implementation

fmrsabino:
Closes #86

- Adds `POST /chains/:chainId/hooks/events`
  * This route accepts the following events: `NEW_CONFIRMATION`, `EXECUTED_MULTISIG_TRANSACTION`, `PENDING_MULTISIG_TRANSACTION`
  * If the payload is not recognised the route returns `400`
  * Payload validation is done with the `EventValidationPipe` which uses AJV to validate the schema of the request body.
- `ITransactionApi` now exposes `clearLocalBalances(safeAddress)` – this method can be used on the route level whenever the locally stored (cached) balances of a specific `address` should be removed.
- `ICacheService` now exposes `delete(key)` – it can be used by other `datasources` if a key needs to be deleted.
  * The Redis implementation calls `unlink` on the key https://redis.io/commands/unlink/

### DIFF
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -11,8 +11,9 @@ export class FakeCacheService implements ICacheService {
     this.cache = {};
   }
 
-  delete(key: string) {
+  delete(key: string): Promise<number> {
     delete this.cache[key];
+    return Promise.resolve(1);
   }
 
   get(key: string, field: string): Promise<string | undefined> {
--- a/src/routes/cache-hooks/cache-hooks.service.ts
+++ b/src/routes/cache-hooks/cache-hooks.service.ts
@@ -1,0 +1,32 @@
+import { Inject, Injectable } from '@nestjs/common';
+import { ExecutedTransaction } from './entities/executed-transaction.entity';
+import { NewConfirmation } from './entities/new-confirmation.entity';
+import { PendingTransaction } from './entities/pending-transaction.entity';
+import { EventType } from './entities/event-payload.entity';
+import { IBalancesRepository } from '../../domain/balances/balances.repository.interface';
+
+@Injectable()
+export class CacheHooksService {
+  constructor(
+    @Inject(IBalancesRepository)
+    private readonly balancesRepository: IBalancesRepository,
+  ) {}
+
+  async onEvent(
+    chainId: string,
+    event: ExecutedTransaction | NewConfirmation | PendingTransaction,
+  ): Promise<void> {
+    switch (event.type) {
+      case EventType.PENDING_MULTISIG_TRANSACTION:
+        break;
+      case EventType.EXECUTED_MULTISIG_TRANSACTION:
+        await this.balancesRepository.clearLocalBalances(
+          chainId,
+          event.address,
+        );
+        break;
+      case EventType.NEW_CONFIRMATION:
+        break;
+    }
+  }
+}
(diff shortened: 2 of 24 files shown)


src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -11,8 +11,9 @@ export class FakeCacheService implements ICacheService {
     this.cache = {};
   }
 
-  delete(key: string) {
+  delete(key: string): Promise<number> {
     delete this.cache[key];
+    return Promise.resolve(1);
```
fmrsabino, line 16: `number` represents the number of entries deleted by this operation. We assume `1` for delete operations in `FakeCacheService` as a key should be unique.

src/routes/cache-hooks/cache-hooks.service.ts
```diff
@@ -0,0 +1,32 @@
...
+  ): Promise<void> {
+    switch (event.type) {
+      case EventType.PENDING_MULTISIG_TRANSACTION:
+        break;
+      case EventType.EXECUTED_MULTISIG_TRANSACTION:
+        await this.balancesRepository.clearLocalBalances(
+          chainId,
+          event.address,
+        );
+        break;
```
fmrsabino, line 27: For each event we call the respective functions to clear the local cache.

This can probably be accomplished with some shared behaviour with type unions or by introducing the concept of `InvalidationScope`. For an initial implementation I think this is ok and also easy to read. We can see how this evolves as more endpoints are added in the future.

src/routes/cache-hooks/pipes/event-validation.pipe.ts
```diff
@@ -0,0 +1,49 @@
...
+    value: any,
+  ): ExecutedTransaction | NewConfirmation | PendingTransaction {
+    this.isExecutedTransactionEvent(value);
+
+    if (
+      this.isExecutedTransactionEvent(value) ||
+      this.isNewConfirmationEvent(value) ||
+      this.isPendingTransactionEvent(value)
+    ) {
+      return value;
```
fmrsabino, line 45: The type of `value` is `ExecutedTransaction` or `NewConfirmation` or `PendingTransaction` depending on which branch is successful.

src/routes/cache-hooks/cache-hooks.service.ts
hectorgomezv: Great! Related to this, just for the sake of curiosity: I was reviewing the current Rust implementation of this, I assume these events are the only ones which could trigger a cache invalidation, right? So `EXECUTED_MULTISIG_TRANSACTION` would invalidate balances cache, and both `PENDING_MULTISIG_TRANSACTION` and `NEW_CONFIRMATION` would invalidate confirmations (?) cache?

hectorgomezv (approved):
Great! 🚀

src/routes/cache-hooks/cache-hooks.service.ts
fmrsabino: > Great! Related to this, just for the sake of curiosity: I was reviewing the current Rust implementation of this, I assume these events are the only ones which could trigger a cache invalidation, right? So `EXECUTED_MULTISIG_TRANSACTION` would invalidate balances cache, and both `PENDING_MULTISIG_TRANSACTION` and `NEW_CONFIRMATION` would invalidate confirmations (?) cache?

This is something that we will need to iteratively work on. The current CGW performs a scan over a patter with either the Safe Transaction Hash. This is incredibly inefficient and one of the main goals for the new caching solution is to optimize this step.

Therefore we are now grouping related entries in Sets (in Redis) and we invalidate those not by scanning a pattern but by being more explicit – this means that for each event, we need to figure what what sets need to be invalidated.

PR #91, opened 2022-09-19
Add exchange-api test

### LINKED ISSUE #64: Add unit tests to ExchangeRepository

moisses89:
- Add exchange-api test

### DIFF
--- a/src/datasources/exchange-api/exchange-api.service.spec.ts
+++ b/src/datasources/exchange-api/exchange-api.service.spec.ts
@@ -1,6 +1,6 @@
 import { ExchangeApi } from './exchange-api.service';
 import exchangeResultFactory from '../../domain/exchange/entities/__tests__/exchange.factory';
-import { FiatCodesExchangeResult } from '../../domain/exchange/entities/fiat-codes-result.entity';
+import exchangeFiatCodesFactory from '../../domain/exchange/entities/__tests__/fiat-codes.factory';
 import { FakeConfigurationService } from '../../common/config/__tests__/fake.configuration.service';
 import { mockNetworkService } from '../../datasources/network/__tests__/test.network.module';
 
@@ -27,10 +27,7 @@ describe('ExchangeApi', () => {
   });
 
   it('Should return the fiatcodes', async () => {
-    const expectedFiatCodes: FiatCodesExchangeResult = {
-      success: true,
-      symbols: {},
-    };
+    const expectedFiatCodes = exchangeFiatCodesFactory(true, { USD: 'Dollar' });
     mockNetworkService.get.mockResolvedValue({ data: expectedFiatCodes });
 
     const fiatcodes = await service.getFiatCodes();
--- a/src/datasources/exchange-api/exchange-api.service.spec.ts
+++ b/src/datasources/exchange-api/exchange-api.service.spec.ts
@@ -1,28 +1,23 @@
 import { ExchangeApi } from './exchange-api.service';
-import { ExchangeResult } from '../../domain/exchange/entities/exchange-result.entity';
+import exchangeResultFactory from '../../domain/exchange/entities/__tests__/exchange.factory';
 import { FiatCodesExchangeResult } from '../../domain/exchange/entities/fiat-codes-result.entity';
 import { FakeConfigurationService } from '../../common/config/__tests__/fake.configuration.service';
 import { mockNetworkService } from '../../datasources/network/__tests__/test.network.module';
 
-const FIATCODES: FiatCodesExchangeResult = {
-  success: true,
-  symbols: {},
-};
-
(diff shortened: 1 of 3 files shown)


src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+describe('ExchangeApi', () => {
+  const fakeConfigurationService = new FakeConfigurationService();
+  fakeConfigurationService.set('exchange.baseUri', 'http://exchange.test');
+  fakeConfigurationService.set('exchange.apiKey', 'aaaaaaaabbbbbbb');
+  const service: ExchangeApi = new ExchangeApi(
+    fakeConfigurationService,
+    mockNetworkService,
+  );
```
fmrsabino, line 25: Here the instance will be reused across tests in this file. This can cause issues if we have a state stored in `ExchangeApi`.

It is better to create and assign the instance in `beforeEach`.

You should also use `jest.clearAllMocks()` so that any mocked state is cleared.

src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+const EXCHANGERESULT: ExchangeResult = {
+  success: true,
+  rates: {},
+  base: '',
+};
+
+describe('ExchangeApi', () => {
+  const fakeConfigurationService = new FakeConfigurationService();
+  fakeConfigurationService.set('exchange.baseUri', 'http://exchange.test');
+  fakeConfigurationService.set('exchange.apiKey', 'aaaaaaaabbbbbbb');
```
fmrsabino, line 21: Probably setting this in `beforeAll` is better as we know that this configuration will "run before all the tests". It's similar to what is being done here but with a more "semantic" way for tests.

src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+const FIATCODES: FiatCodesExchangeResult = {
+  success: true,
+  symbols: {},
+};
+
+const EXCHANGERESULT: ExchangeResult = {
+  success: true,
+  rates: {},
+  base: '',
+};
```
fmrsabino, line 16: I would move this to inside each test. Tests referencing a more global context are more confusing to read in my opinion as any potential state change might affect the outcome of the tests.

You can use `exchangeResultFactory` in each test to get a `ExchangeResult`

src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -0,0 +1,45 @@
...
+    mockNetworkService.get.mockResolvedValue({ data: FIATCODES });
+    const fiatcodes = await service.getFiatCodes();
+    expect(fiatcodes).toBe(FIATCODES);
+  });
+
+  it('Should return the exchange result', async () => {
+    mockNetworkService.get.mockResolvedValue({ data: EXCHANGERESULT });
+    const exchangeresult = await service.getExchangeResult();
+    expect(exchangeresult).toBe(EXCHANGERESULT);
+  });
```
fmrsabino, line 44: nitpick: setup like:

```
// setup
// newline
// action
// newline
// assertions
```

fmrsabino (changes requested):

src/datasources/exchange-api/exchange-api.service.spec.ts
moisses89: I haven't a strong opinion about this, I agree with you :)

src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -0,0 +1,49 @@
...
+
+  it('should error if configuration is not defined', async () => {
+    const fakeConfigurationService = new FakeConfigurationService();
+    await expect(
+      () => new ExchangeApi(fakeConfigurationService, mockNetworkService),
+    ).toThrow();
+  });
+
+  it('Should return the fiatcodes', async () => {
+    const expectedFiatCodes: FiatCodesExchangeResult = {
```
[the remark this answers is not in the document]
fmrsabino: Yes I agree 👍

fmrsabino (dismissed):
Nice! 🚀 👏

fmrsabino (changes requested):
Looks good!

Just one request: to rename `fiatcodes.factory.ts` to `fiat-codes.factory.ts`. This follows more the naming standard that we have in the project.

fmrsabino (approved):
🚀

PR #93, opened 2022-09-19
Add remaining fields to Chain entity

### LINKED ISSUE #92: Complete Chain model schema
- Currently our domain `Chain` entity only has a subset of what is required by the clients
- We should add the remaining data with the appropriate schemas in place

fmrsabino:
Closes #92 

- Adds remaining fields to Chain entity
- Adds respective JSON schemas to the fields
- Adds respect OpenApi schemas to the new fields


src/common/schemas/json-schema.service.ts
```diff
@@ -8,6 +9,7 @@ export class JsonSchemaService {
   constructor() {
     // coerceTypes param shouldn't be necessary when serialization is implemented.
     this.ajv = new Ajv({ coerceTypes: true });
+    addFormats(this.ajv, { formats: ['uri'] });
```
fmrsabino, line 12: This adds support for additional checks on AJV. In this case it adds support for checking `uri`.

src/common/schemas/json-schema.service.ts
```diff
@@ -1,5 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import Ajv, { JSONSchemaType, Schema, ValidateFunction } from 'ajv';
+import addFormats from 'ajv-formats';
```
fmrsabino, line 3: See https://ajv.js.org/guide/formats.html

src/datasources/config-api/config-api.service.spec.ts
```diff
@@ -1,44 +1,7 @@
...
-        symbol: 'XDAI',
-        decimals: 18,
-        logoUri:
-          'https://safe-transaction-assets.gnosis-safe.io/chains/100/currency_logo.png',
-      },
-    },
-  ],
-};
-
-const CHAIN: Chain = CHAINS.results[0];
```
fmrsabino, line 41: I moved these to the tests because I find that using the factories is more readable (and also we avoid sharing state between the tests)

src/domain/chains/entities/__tests__/block-explorer-uri-template.factory.ts
```diff
@@ -1,3 +1,5 @@
+/* istanbul ignore file */
+
```
fmrsabino, line 2: For now, I added these to the the factories of chains. We can think of a rule that would match any file under `__tests__` and that ends with `.factory.ts`.

I don't think the factories should be included in the coverage as they serve more of a utility for the tests. The branches that were being evaluated in the coverage were specific to having an argument provided or not (nullability check `??`).

fmrsabino: Proposal under https://github.com/5afe/safe-client-gateway-nest/pull/94

src/datasources/config-api/config-api.service.spec.ts
moisses89: Yes I agree, but could be better in a custom PR with this purpose, because if we need to revert this PR shouldn't revert this. It's only an idea.

fmrsabino: I've changed it here because this was not compiling either way due to the new schema.

src/domain/chains/entities/__tests__/block-explorer-uri-template.factory.ts
fmrsabino: Reverted in [d6ed2b7](https://github.com/5afe/safe-client-gateway-nest/pull/93/commits/d6ed2b745d8c6032e314033ff3b5192978e8f60a) because #94 was merged

src/domain/chains/entities/chain.entity.ts
```diff
@@ -1,9 +1,24 @@
...
+  safeAppsRpcUri: RpcUri;
+  publicRpcUri: RpcUri;
+  blockExplorerUriTemplate: BlockExplorerUriTemplate;
+  nativeCurrency: NativeCurrency;
   transactionService: string;
   vpcTransactionService: string;
-  nativeCurrency: NativeCurrency;
+  theme: Theme;
+  gasPrice: Array<GasPriceOracle | GasPriceFixed>;
+  ensRegistryAddress: string;
```
moisses89, line 21: This is an optional value on the current client-gateway.
` pub ens_registry_address: Option<String>`

src/domain/chains/entities/gas-price-oracle.entity.ts
```diff
@@ -0,0 +1,8 @@
+import { GasPrice } from './gas-price.entity';
+
+export interface GasPriceOracle extends GasPrice {
+  type: 'oracle';
+  uri: string;
+  gasParameter: string;
+  gweiFactor: number;
```
moisses89, line 7: Maybe this was an issue in the cgw but `gweiFactor` is a string in the current cgw.` gwei_factor: String`

src/domain/chains/entities/rpc-uri-authentication.entity.ts
```diff
@@ -0,0 +1,4 @@
+export enum RpcUriAuthentication {
+  ApiKeyPath = 'API_KEY_PATH',
+  NoAuthentication = 'NO_AUTHENTICATION',
+}
```
moisses89, line 4: In the current CGW exists a third option `Unknown`

src/domain/chains/entities/schemas/chain.schema.ts
```diff
@@ -1,28 +1,118 @@
...
+    'rpcUri',
+    'safeAppsRpcUri',
+    'publicRpcUri',
+    'blockExplorerUriTemplate',
+    'nativeCurrency',
+    'transactionService',
+    'vpcTransactionService',
+    'theme',
+    'gasPrice',
+    'ensRegistryAddress',
```
moisses89, line 114: Maybe shouldn't be required

moisses89 (changes requested):
Good work! I was checking the models with the current CGW and I noticied some differences.

src/domain/chains/entities/gas-price-oracle.entity.ts
fmrsabino: I will keep it as `string` since we don't perform any numeric manipulation and we need to evaluate the usage of a big decimal numbers in the CGW.

src/domain/chains/entities/schemas/chain.schema.ts
fmrsabino: Done in [34d40f2](https://github.com/5afe/safe-client-gateway-nest/pull/93/commits/34d40f226f67c3b006a3f1eccccfe9dc022ad746)

src/domain/chains/entities/rpc-uri-authentication.entity.ts
fmrsabino: Done in [34d40f2](https://github.com/5afe/safe-client-gateway-nest/pull/93/commits/34d40f226f67c3b006a3f1eccccfe9dc022ad746)

src/domain/chains/entities/gas-price-oracle.entity.ts
fmrsabino: Done in [34d40f2](https://github.com/5afe/safe-client-gateway-nest/pull/93/commits/34d40f226f67c3b006a3f1eccccfe9dc022ad746)

src/domain/chains/entities/chain.entity.ts
fmrsabino: Done in [34d40f2](https://github.com/5afe/safe-client-gateway-nest/pull/93/commits/34d40f226f67c3b006a3f1eccccfe9dc022ad746)

src/datasources/config-api/config-api.service.spec.ts
```diff
@@ -48,34 +11,54 @@ const mockDataSource = jest.mocked(dataSource);
...
 
   const service: ConfigApi = new ConfigApi(
     dataSource,
     fakeConfigurationService,
   );
 
+  beforeEach(async () => {
+    jest.clearAllMocks();
+  });
+
```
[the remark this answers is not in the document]
fmrsabino: Done in [34d40f2](https://github.com/5afe/safe-client-gateway-nest/pull/93/commits/34d40f226f67c3b006a3f1eccccfe9dc022ad746)

src/domain/chains/entities/__tests__/block-explorer-uri-template.factory.ts
```diff
@@ -0,0 +1,14 @@
...
+export default function (
+  address?: string,
+  txHash?: string,
+  api?: string,
+): BlockExplorerUriTemplate {
+  return <BlockExplorerUriTemplate>{
+    address: address ?? faker.internet.url(),
+    txHash: txHash ?? faker.internet.url(),
+    api: api ?? faker.internet.url(),
```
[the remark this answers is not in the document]
fmrsabino: While I agree that this can be used for documentation purposes, there is no logic in the CGW to validate how the url is formed so in a testing scenario, any `url` should be valid.

If we add validations to the format of the URL then this needs to be change to generate URLs in that format too.

moisses89 (approved):
Looks good for me :)

PR #105, opened 2022-09-22
Add cursor parameter to Swagger

fmrsabino:
- Adds cursor parameter to /v1/chains route
- This is for documentation purposes only. Cursor is already read and validated if provided

image


moisses89 (approved):
Looks good for me :)

PR #106, opened 2022-09-22
Add support for Record schema validation

fmrsabino:
- Adds validation support for types which had `Record<string, T>`
- `ExchangeResult`, `Backbone` and `FiatCodesExchangeResult` now use `Record<string, string>`

### DIFF
--- a/src/routes/chains/entities/backbone.entity.ts
+++ b/src/routes/chains/entities/backbone.entity.ts
@@ -13,7 +13,7 @@ export class Backbone implements DomainBackbone {
   @ApiProperty()
   secure: boolean;
   @ApiProperty()
-  settings: object;
+  settings: Record<string, string>;
   @ApiProperty()
   version: string;
 }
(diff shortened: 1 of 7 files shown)


src/routes/chains/entities/backbone.entity.ts
```diff
@@ -13,7 +13,7 @@ export class Backbone implements DomainBackbone {
   @ApiProperty()
   secure: boolean;
   @ApiProperty()
-  settings: object;
+  settings: Record<string, string>;
```
moisses89, line 16: Why are two backbone entities?

fmrsabino: One is what we consider the domain entity that can be used by any route: `src/domain/backbone/entities/backbone.entity.ts`

The other one is what the route returns including the OpenAPI documentation for it: `src/routes/chains/entities/backbone.entity.ts`

moisses89 (approved):

PR #110, opened 2022-09-27
Fix balances/usd/ exclude_spam query parameter

fmrsabino:
- The query parameter in `balances/usd/` endpoint is not `excludeSpam` but `exclude_spam`


moisses89 (approved):
Looks good for me!

PR #111, opened 2022-09-27
Move token entities to balances route

fmrsabino:
- Currently, `token-info` and `token-type` are only being used in the `balances` route – they might end up being used later by other routes but at that point we can move them to a `common` entity layer which can be shared between routes
- Removed `tokenInfoSchema` since we were not using it – currently we only validate the JSON coming from external services


moisses89 (approved):
Agree with the PR,

PR #119, opened 2022-09-30
Remove openapi folder from routes/balances

fmrsabino:
- Entities at the route level should be under `/entities`
- A separate OpenApi entity should not be required – we assume that route level entities have OpenApi related decorators


src/routes/balances/entities/balance.entity.ts
```diff
@@ -1,8 +1,14 @@
+import { Balance as DomainBalance } from '../entities/balance.entity';
+import { ApiProperty } from '@nestjs/swagger';
 import { TokenInfo } from './token-info.entity';
 
-export interface Balance {
-  tokenInfo: TokenInfo;
+export class Balance implements DomainBalance {
```
moisses89, line 5: `DomainBalance` could be confusing because it's not really a domain entity.

src/routes/balances/entities/token-info.entity.ts
```diff
@@ -1,10 +1,18 @@
+import { TokenInfo as DomainTokenInfo } from '../entities/token-info.entity';
+import { ApiProperty } from '@nestjs/swagger';
 import { TokenType } from './token-type.entity';
 
-export interface TokenInfo {
-  tokenType: TokenType;
+export class TokenInfo implements DomainTokenInfo {
```
moisses89, line 5: The same as above

src/routes/balances/entities/token-info.entity.ts
```diff
@@ -1,10 +1,18 @@
...
+  @ApiProperty()
   address: string;
+  @ApiProperty()
   decimals: number;
-  symbol: string;
+  @ApiProperty()
+  logoUri: string;
+  @ApiProperty()
   name: string;
-  logoUri?: string;
```
moisses89, line 9: Are you aware about this, now `logoUri` is mandatory.

src/routes/balances/entities/balance.entity.ts
fmrsabino: What do you suggest?

src/routes/balances/entities/token-info.entity.ts
fmrsabino: Just to be clear, `logoUri` changed recently from an optional field to a non-optional field?

moisses89: Nope, I meant that was optional until this change. I want to be sure that you are aware about this.

src/routes/balances/entities/balance.entity.ts
moisses89: I would like use `domainBalance` for domain entities and `balance` for route entities.

hectorgomezv: I don't get this, it should be importing/implementing the domain entity interface, right? It seems you're importing the same file (`'../entities/balance.entity'`)? 

Having that fixed, I think the naming it's ok. 🙂

The rest of the files have the same mistake I think.

hectorgomezv (changes requested):

src/routes/balances/entities/token-info.entity.ts
hectorgomezv: I think the problem is a bad import, this should be indeed implementing a domain entity interface 🙂

src/routes/balances/entities/balance.entity.ts
fmrsabino: Yes you are right. I've fixed the imports either by removing them or by pointing them to the correct place. [cf549a9](https://github.com/5afe/safe-client-gateway-nest/pull/119/commits/cf549a953d7a617e18c8004937501fa7eac2edb7)

moisses89 (approved):
Looks good for me!

hectorgomezv (approved):
🚀

PR #121, opened 2022-10-05
Add GET /chains/:chainId/contracts/:contractAddress 

### LINKED ISSUE #100: Add contract/<address> route
- Add `/v1/chains/<chain_id>/contracts/<contract_address>` route
- Tests should be added

Reference: https://github.com/safe-global/safe-client-gateway/blob/89bd81400b6356d2c34804fe5f097fc983d7f1d0/src/routes/contracts/routes.rs#L67-L90

hectorgomezv:
This PR implements the GET contract endpoint.

- [X] Add /v1/chains/<chain_id>/contracts/<contract_address> route
- [X] Caching support should be added
- [X] Tests should be added

A validator is used, following the strategy from #85 

Closes #100


src/domain/balances/balances.validator.spec.ts
```diff
@@ -24,6 +22,10 @@ describe('Balances validator', () => {
 
   it('should return the data when validation succeed', () => {
     const balances = [balanceFactory(), balanceFactory()];
+    const expectedErrMessage = 'testErrMessage';
+    mockValidationErrorFactory.from.mockReturnValueOnce(
+      new HttpException(expectedErrMessage, 500),
+    );
```
moisses89, line 28: I din`t get this, the return value should be 500 but validator succeed?

src/domain/contracts/entities/contract.entity.ts
```diff
@@ -0,0 +1,8 @@
+export class Contract {
+  address: string;
+  name: string;
+  displayName: string;
+  logoUri?: string;
+  contractAbi?: object;
+  trustedForDelegateCall: boolean;
+}
```
moisses89, line 8: Following the tx-service model https://github.com/safe-global/safe-transaction-service/blob/master/safe_transaction_service/contracts/models.py#L160 I think logo shouldn't be optional. 
I mean if no logo, the service should return empty string not `null`. 
What do you think?

src/routes/contracts/contracts.module.ts
```diff
@@ -0,0 +1,9 @@
+import { Module } from '@nestjs/common';
+import { ContractsController } from './contracts.controller';
+import { ContractsService } from './contracts.service';
+
+@Module({
+  controllers: [ContractsController],
+  providers: [ContractsService],
+})
+export class ContractsModule {}
```
moisses89, line 9: Why modules here? :thinking:

moisses89 (changes requested):
Looks good :) only some comments that only I would be ensure that you aware.

src/domain/contracts/entities/contract.entity.ts
hectorgomezv: Changed [here.](https://github.com/5afe/safe-client-gateway-nest/commit/7e1b1aa3dca4b4723a3768cb0241b527117a25a3)

Thanks!

moisses89 (approved):
:rocket:

Issue #101, opened 2022-09-20
Add GET Delegates


fmrsabino:
- Add `/v1/chains/<chain_id>/delegates?<safe>&<delegate>&<delegator>&<label>` route
- Tests should be added
- OpenApi specification should be added

PR #123, opened 2022-10-11
Add get delegates repository

moisses89:
Related with #101 
- Add get delegates repository into safe domain.

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.ts
+++ b/src/datasources/transaction-api/transaction-api.service.ts
@@ -113,7 +113,7 @@ export class TransactionApi implements ITransactionApi {
     }
   }
 
- async getContract(contractAddress: string): Promise<Contract> {
+  async getContract(contractAddress: string): Promise<Contract> {
     try {
       const cacheKey = contractCacheKey(this.chainId, contractAddress);
       const url = `${this.baseUrl}/api/v1/contracts/${contractAddress}`;
@@ -133,7 +133,7 @@ export class TransactionApi implements ITransactionApi {
   ): Promise<Page<Delegate>> {
     try {
       const cacheKey = `${this.chainId}_delegates`;
-      const cacheKeyField = `${safeAddress}_${delegate}_${delegator}_${label}`;
+      const cacheKeyField = `${safeAddress}_${delegate}_${delegator}_${label}_${limit}_${offset}`;
       const url = `${this.baseUrl}/api/v1/delegates/`;
       return await this.dataSource.get(cacheKey, cacheKeyField, url, {
         params: {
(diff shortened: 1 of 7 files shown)


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -106,4 +107,31 @@ export class TransactionApi implements ITransactionApi {
...
+    safeAddress?: string,
+    delegate?: string,
+    delegator?: string,
+    label?: string,
+    limit?: number,
+    offset?: number,
+  ): Promise<Page<Delegate>> {
+    try {
+      const cacheKey = `${this.chainId}_delegates`;
+      const cacheKeyField = `${safeAddress}_${delegate}_${delegator}_${label}`;
```
hectorgomezv, line 121: Should we include both limit and offset as part of cache key? I mean, otherwise we would be caching different offset/limit results in the same key 🤔

src/domain/interfaces/transaction-api.interface.ts
```diff
@@ -27,4 +27,13 @@ export interface ITransactionApi {
   getMasterCopies(): Promise<MasterCopy[]>;
 
   getSafe(safeAddress: string): Promise<Safe>;
+
+  getDelegates(
```
hectorgomezv, line 31: Just as a comment, not required for this PR, but we could start thinking in wrapping arguments into DTO types when they're a bunch as in this case.

src/datasources/transaction-api/transaction-api.service.ts
moisses89: yes, updated

hectorgomezv (dismissed):
🚀

src/domain/safe/entities/schemas/delegate.schema.ts
```diff
@@ -0,0 +1,13 @@
+import { JSONSchemaType } from 'ajv';
+import { Delegate } from '../delegate.entity';
+
+export const delegateSchema: JSONSchemaType<Delegate> = {
+  type: 'object',
+  properties: {
+    safe: { type: 'string' },
```
fmrsabino, line 7: `safe` should be an optional field

fmrsabino (changes requested):

src/domain/safe/entities/schemas/delegate.schema.ts
moisses89: Done [here](https://github.com/5afe/safe-client-gateway-nest/pull/123/commits/f796ea02dd39a5701bcd760dda7c8924c0dcdf4c)

fmrsabino (approved):
🚀

PR #132, opened 2022-10-17
Add data decoder route

### LINKED ISSUE #99: Add data-decoder route
- Add `/v1/chains/<chain_id>/data-decoder` route
- Tests should be added

Reference: https://github.com/safe-global/safe-client-gateway/blob/89bd81400b6356d2c34804fe5f097fc983d7f1d0/src/routes/contracts/routes.rs#L10-L65

hectorgomezv:
Closes #99 

This PR aims to create a `POST /v1/chains/<chain_id>/data-decoder` endpoint into the service. 

### Tasks
- [X] Add /v1/chains/<chain_id>/data-decoder route
- [X] Tests should be added

Any feedback regarding the expected validation schema for the payload coming from Transaction Service is welcomed (see comments) 🙂


src/domain.module.ts
```diff
@@ -41,20 +44,22 @@ import { DelegateRepository } from './domain/delegate/delegate.repository';
     { provide: IChainsRepository, useClass: ChainsRepository },
     { provide: ICollectiblesRepository, useClass: CollectiblesRepository },
     { provide: IContractsRepository, useClass: ContractsRepository },
+    { provide: IDataDecodedRepository, useClass: DataDecodedRepository },
```
hectorgomezv, line 47: Just ordering import/exports alphabetically here 🙂

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
```diff
@@ -0,0 +1,28 @@
+import { JSONSchemaType, Schema } from 'ajv';
+import { DataDecodedParameter } from '../data-decoded.entity';
+
+export const dataDecodedParameterSchema: JSONSchemaType<DataDecodedParameter> =
```
hectorgomezv, line 4: Any feedback on this schema would be welcomed. I tried to be flexible when defining this because I think the data decoded payload has a pretty open format. But maybe I lack of the knowledge to define a more fine-grained schema.

hectorgomezv: Of course if you feel we need to change any definition here, I'll include the changes in the entity and OpenAPI definitions as well 🙂

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -56,4 +56,26 @@ export class CacheFirstDataSource {
     await this.cacheService.set(key, field, rawJson, expireTimeSeconds);
     return data;
   }
+
+  /**
+   * Executes a POST request against an {@link url}, by using
+   * the injected NetworkService.
+   * Cache is not used by this function.
```
fmrsabino, line 63: I think this can be quite confusing – from 2 of the methods exposed by the `CacheFirstDataSource` one of the doesn't "cache".

I would say that it is better for consumers of this API to use the network data source directly as it is more explicit.

fmrsabino (changes requested):

src/routes/data-decode/entities/data-decoded.entity.ts
```diff
@@ -0,0 +1,12 @@
...
+import {
+  DataDecoded as DomainDataDecoded,
+  DataDecodedParameter,
+} from '../../../domain/data-decoder/entities/data-decoded.entity';
+
+export class DataDecoded implements DomainDataDecoded {
+  @ApiProperty()
+  method: string;
+  @ApiProperty()
+  parameters?: DataDecodedParameter[] | undefined;
```
fmrsabino, line 11:
```suggestion
  @ApiPropertyOptional()
  parameters?: DataDecodedParameter[];
```

fmrsabino (changes requested):

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: I had the exact same thought when I wrote this part 🙂. I forgot to comment that in the PR but I was doubting we could just use `NetworkService` in the caller as you said, or create an `IDataSource` and make `CacheFirstDataSource` (with cache) and a new `DataSource` (without cache) implement it, or even have `getWithCache/postWithCache` counterparts to `get/post` functions.

Probably just using `NetworkService` is the simpler one, I'll make that change 🙂

src/routes/data-decode/__tests__/create-data-decoded.dto.factory.ts
```diff
@@ -0,0 +1,9 @@
+import { faker } from '@faker-js/faker';
+import { CreateDataDecodedDto } from '../entities/create-data-decoded.dto';
+
+export default function (data?: string, to?: string): CreateDataDecodedDto {
+  return <CreateDataDecodedDto>{
```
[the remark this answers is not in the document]
hectorgomezv: True, I was probably thinking this was an interface. I will change it 👍🏻

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Changed in [1d66427](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/1d6642705e425416d8d1842dc5d02f6792202098)

src/routes/data-decode/entities/data-decoded.entity.ts
hectorgomezv: Changed in [1d66427](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/1d6642705e425416d8d1842dc5d02f6792202098)

src/domain/data-decoder/data-decoded.repository.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface IDataDecodedRepository {
+  /**
+   * Gets the {@link DataDecoded} associated with {@link chainId} the for the {@link data}
+   * and the address pointed by {@link to}.
+   *
+   * @param chainId
+   * @param data
+   * @param to
+   */
+  decode(chainId: string, data: string, to: string): Promise<DataDecoded>;
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [1d66427](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/1d6642705e425416d8d1842dc5d02f6792202098)

src/routes/data-decode/__tests__/create-data-decoded.dto.factory.ts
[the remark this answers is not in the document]
fmrsabino: > Giving this a second thought... to instantiate the class directly we would need a constructor in the class. This is the first DTO in the project and maybe we need an alignment here... I left the DTO into the _routes_ layer because is never passed in to the _domain_ layer (`DataDecodedRepository` in this case), which receives primitives only.

Let's go with the constructor. I have a branch where I also have a constructor for it: https://github.com/5afe/safe-client-gateway-nest/blob/088dfca3d2e5b8ba16fe33a0a37051b13eccc060/src/routes/safes/entities/safe-info.entity.ts

> That's the reason why the class `CreateDataDecodedDto` is annotated with OpenAPI stuff without implementing any _domain_ entity. This pattern is sightly different, so I was thinking in moving DTOs to the _domain_, and I think _services_ could be moved as well (in another PR) to _domain_, what do you think @fmrsabino?

I don't think this is the way to go. We would be moving the OpenApi schema to domain and I think the domain should not be aware of it. I see `CreateDataDecodedDto` as a route-only entity – it only encapsulates the request payload.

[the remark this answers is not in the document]
hectorgomezv: Ok, I'm fine with keeping the DTO into the routes layer, just thinking in having a counterpart in the domain layer as the rest of the entities (to keep OpenAPI definitions away from domain). Based on how the project evolves and where we place the business logic we could think if _services_ are part of the routes or domain layer. I mean, the project is small and maybe is a futile decision, but I would expect to have _services_ available even if we implement another entry layer aside REST (cli, websockets, etc.), in that case I would move them to _domain_.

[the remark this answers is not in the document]
hectorgomezv: I created a constructor as you did, it's simpler and helps to keep in mind this is a route-only entity (at least for now) 👍 

Changed in [5bfc6006c96742d8521aa23768926e18aa9eb55f](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/5bfc6006c96742d8521aa23768926e18aa9eb55f)

src/domain/data-decoder/entities/data-decoded.entity.ts
```diff
@@ -0,0 +1,11 @@
+export interface DataDecodedParameter {
+  name: string;
+  param_type: string;
+  value: string | number;
+  value_decoded?: Record<string, any>;
```
fmrsabino, line 5: Shouldn't this be `valueDecoded`?

hectorgomezv: Hmm, yes I'm seeing `valueDecoded` now as well as part of the TX service responses. But I think I copied this from another response... not sure. Due to the misalignment of names, this wasn't being actually validated. I adapted the schema to accept both an object and an array in the `valueDecoded` field (I'm seeing an array there for a multiSend transaction), but as I commented I lack of the knowledge to define a more fine-grained schema here. 

The change is on [34fb5c7ad694e4f957b0d3b7c748cf27c0cac630](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/34fb5c7ad694e4f957b0d3b7c748cf27c0cac630). I needed to define it as `Schema` instead of `JsonSchemaType`. Maybe another option could be just define the required properties only (`name` and `value`) in `dataDecodedParameterSchema` to have a more flexible schema, what do you think?

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
```diff
@@ -0,0 +1,31 @@
+import { Schema } from 'ajv';
+
+export const dataDecodedParameterSchema: Schema = {
+  type: 'object',
+  properties: {
+    name: { type: 'string' },
+    param_type: { type: 'string' },
```
fmrsabino, line 7: The name of this parameter should be `type`. However the CGW renames it to `paramType` when serialising to the clients.

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
```diff
@@ -0,0 +1,31 @@
...
+        { type: 'object', nullable: true },
+        { type: 'array', nullable: true },
+      ],
+    },
+  },
+  required: ['name', 'value'],
+  additionalProperties: true,
+};
+
+export const dataDecodedSchema: Schema = {
```
fmrsabino, line 20: Why not constrain the schema with `JSONSchemaType`? It would give some better error hints if the schema doesn't match a type `T`.

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
```diff
@@ -0,0 +1,31 @@
...
+  type: 'object',
+  properties: {
+    method: { type: 'string' },
+    parameters: {
+      anyOf: [
+        { type: 'null' },
+        { type: 'array', items: { $ref: 'dataDecodedParameter' } },
+      ],
+    },
```
fmrsabino, line 29: Isn't it possible to use `nullable` here instead of having a one of the possible types as `null`?

fmrsabino (changes requested):

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
hectorgomezv: This is something I think we discussed when introduced AJV validation... for some reason `JSONSchemaType` is not fully compatible with nullable fields (TS compilation fails), or maybe I'm missing something in the documentation. You can try to change the definition to see if you find the way. We could create a new issue to investigate this further if you like, but I was already blocked by this for a while in the past.

Another solution, in the case of the data-decoder endpoint, due to its nature (payloads are pretty free I think), we could just skip validation for some fields in the payload as I suggested, wdyt?

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
hectorgomezv: I'm testing payloads in https://www.jsonschemavalidator.net/ and it seems nullable doesn't work in the same way for nested objects.
Screenshot 2022-10-18 at 12 11 28
Screenshot 2022-10-18 at 12 12 33

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
hectorgomezv: Changed in [50ee6e966f4925eeba8f74f82c55e3ff8a216c16](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/50ee6e966f4925eeba8f74f82c55e3ff8a216c16)

hectorgomezv: Hmm, I'm thinking we should change `paramType` when serialising as well then, right @fmrsabino ? What do you think about opening a new issue to implement the first serializer? It could be decisions to make there as well.

fmrsabino: > Hmm, I'm thinking we should change `paramType` when serialising as well then, right @fmrsabino ? What do you think about opening a new issue to implement the first serializer? It could be decisions to make there as well.

We should keep using `paramType` for the clients. For now I would just map this payload to the OpenApi entity (just simple instantiation at the route level)

src/domain/data-decoder/entities/data-decoded.entity.ts
hectorgomezv: I didn't found a better solution for now, so I kept the `Schema` definition. We may revisit this in the future to try to get all the benefits from the TS/AJS integration 🙂

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
hectorgomezv: I mapped it here [afdc03e954157b284956f02a15792f43bf8a70fd](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/afdc03e954157b284956f02a15792f43bf8a70fd)

I mapped it inline into `src/routes/data-decode/data-decoded.service.ts` for now, a more elegant solution could be developed when we integrate or response types with frontend DTOs I think.

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
hectorgomezv: Please refer to [the previous comment](https://github.com/5afe/safe-client-gateway-nest/pull/132#discussion_r998412182)

hectorgomezv:
Added a simple validation using TS type guards [here](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/206ab186563e3f3945facd881c6da292aca9231e) 🙂

PR #133, opened 2022-10-18
Add delegates route

### LINKED ISSUE #101: Add GET Delegates

moisses89:
Closes #101 
- Add the endpoint GET delegates
- Add some tests

### DIFF
--- a/src/domain/delegate/entities/__tests__/delegate.factory.ts
+++ b/src/domain/delegate/entities/__tests__/delegate.factory.ts
@@ -1,0 +1,15 @@
+import { faker } from '@faker-js/faker';
+import { Delegate } from '../delegate.entity';
+export default function (
+  safe?: string,
+  delegate?: string,
+  delegator?: string,
+  label?: boolean,
+): Delegate {
+  return <Delegate>{
+    safe: safe ?? faker.random.word(),
+    delegate: delegate ?? faker.random.word(),
+    delegator: delegator ?? faker.random.word(),
+    label: label ?? faker.random.word(),
+  };
+}
--- a/src/domain/delegate/entities/__tests__/delegate.factory.ts
+++ b/src/domain/delegate/entities/__tests__/delegate.factory.ts
@@ -7,9 +7,9 @@ export default function (
   label?: boolean,
 ): Delegate {
   return <Delegate>{
-    safe: safe ?? faker.random.word(),
-    delegate: delegate ?? faker.random.word(),
-    delegator: delegator ?? faker.random.word(),
+    safe: safe ?? faker.finance.ethereumAddress(),
+    delegate: delegate ?? faker.finance.ethereumAddress(),
+    delegator: delegator ?? faker.finance.ethereumAddress(),
     label: label ?? faker.random.word(),
   };
 }
--- a/src/routes/delegates/delegates.controller.spec.ts
+++ b/src/routes/delegates/delegates.controller.spec.ts
@@ -100,16 +100,14 @@ describe('Delegates controller', () => {
       const chainId = '1';
       const chainResponse = chainFactory(chainId);
       mockNetworkService.get.mockResolvedValueOnce({ data: chainResponse });
-      mockNetworkService.get.mockRejectedValueOnce({
-        data: {},
-        status: 400,
-      });
 
       await request(app.getHttpServer())
         .get(`/chains/${chainId}/delegates`)
-        .expect(400);
-
-      expect(mockNetworkService.get).toBeCalledTimes(2);
+        .expect(400)
+        .expect({
+          message: 'At least one query param must be provided',
+          statusCode: 400,
+        });
     });
   });
 });
--- a/src/routes/delegates/delegates.controller.spec.ts
+++ b/src/routes/delegates/delegates.controller.spec.ts
@@ -1,0 +1,151 @@
+import { INestApplication } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import * as request from 'supertest';
+import delegateFactory from '../../domain/delegate/entities/__tests__/delegate.factory';
+import { DelegatesModule } from './delegates.module';
+import {
+  mockNetworkService,
+  TestNetworkModule,
+} from '../../datasources/network/__tests__/test.network.module';
+import {
+  fakeConfigurationService,
+  TestConfigurationModule,
+} from '../../config/__tests__/test.configuration.module';
+import {
+  fakeCacheService,
+  TestCacheModule,
+} from '../../datasources/cache/__tests__/test.cache.module';
+import chainFactory from '../../domain/chains/entities/__tests__/chain.factory';
+import { Delegate } from './entities/delegate.entity';
+import { Page } from '../../domain/entities/page.entity';
+import { DomainModule } from '../../domain.module';
+import { DataSourceErrorFilter } from '../common/filters/data-source-error.filter';
+
+describe('Delegates controller', () => {
+  let app: INestApplication;
+
+  beforeAll(async () => {
+    fakeConfigurationService.set(
+      'safeConfig.baseUri',
+      'https://test.safe.config',
+    );
+
+    fakeConfigurationService.set(
+      'exchange.baseUri',
+      'https://test.exchange.service',
+    );
+
+    fakeConfigurationService.set('exchange.apiKey', 'https://test.api.key');
+  });
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+    fakeCacheService.clear();
+
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      imports: [
+        // feature
+        DelegatesModule,
+        // common
+        DomainModule,
+        TestCacheModule,
+        TestConfigurationModule,
+        TestNetworkModule,
+      ],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+    app.useGlobalFilters(new DataSourceErrorFilter());
+
+    await app.init();
+  });
+
+  describe('GET delegates for a Safe', () => {
+    it('Success', async () => {
+      const safe = '0x5afe3855358e112b5647b952709e6165e1c1eeee';
+      const chainId = '1';
+      const chainResponse = chainFactory(chainId);
+      mockNetworkService.get.mockResolvedValueOnce({ data: chainResponse });
+      const pageDelegates = <Page<Delegate>>{
+        count: 2,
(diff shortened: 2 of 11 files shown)


src/routes/delegates/delegates.controller.ts
```diff
@@ -0,0 +1,65 @@
...
+    required: false,
+    type: String,
+  })
+  @ApiImplicitQuery({
+    name: 'cursor',
+    required: false,
+    type: String,
+  })
+  @Get('chains/:chainId/delegates')
+  async getDelegates(
```
moisses89, line 46: We should validate here that at least one query parameter between (safe, delegate, delegator or label) is not `null` because if not, the `tx-service` returns `400 bad request `

src/domain.module.ts
```diff
@@ -66,6 +66,7 @@ import { DelegateRepository } from './domain/delegate/delegate.repository';
     IExchangeRepository,
     ISafeRepository,
     IContractsRepository,
+    IDelegateRepository,
```
fmrsabino, line 69: Please use lexicographical order

src/routes/delegates/entities/delegate.entity.ts
```diff
@@ -0,0 +1,12 @@
+import { ApiProperty } from '@nestjs/swagger';
+
+export class Delegate {
+  @ApiProperty()
```
fmrsabino, line 4: You should ApiPropertyOptional for optional fields

src/routes/delegates/entities/delegate.entity.ts
```diff
@@ -0,0 +1,12 @@
...
+  safe?: string;
+  @ApiProperty()
+  delegate: string;
+  @ApiProperty()
+  delegator: string;
+  @ApiProperty()
+  label: string;
```
fmrsabino, line 11: Shouldn't all of these fields be optional?

src/domain/delegate/entities/__tests__/delegate.factory.ts
```diff
@@ -0,0 +1,15 @@
...
+  delegator?: string,
+  label?: boolean,
+): Delegate {
+  return <Delegate>{
+    safe: safe ?? faker.random.word(),
+    delegate: delegate ?? faker.random.word(),
+    delegator: delegator ?? faker.random.word(),
+    label: label ?? faker.random.word(),
+  };
+}
```
fmrsabino, line 15: You should use faker.finance.ethereumAddress() for these fields

src/routes/delegates/delegates.controller.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    app = moduleFixture.createNestApplication();
+    app.useGlobalFilters(new DataSourceErrorFilter());
+
+    await app.init();
+  });
+
+  describe('GET delegates for a Safe', () => {
+    it('Success', async () => {
+      const safe = '0x5afe3855358e112b5647b952709e6165e1c1eeee';
```
fmrsabino, line 65: nitpick: you can generate a random ethereum address with faker

src/routes/delegates/delegates.controller.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+        params: {
+          safe: safe,
+          delegate: undefined,
+          delegator: undefined,
+          label: undefined,
+          limit: undefined,
+          offset: undefined,
+        },
+      });
+    });
```
fmrsabino, line 97: From the discussion that we had yesterday I'd say that you can skip these checks at this level.

We have two different takes for this one:
- They have some value in validating that we actually call external services with these parameters (testing the boundaries)
- They add complexity to the test which makes it a little bit harder to read.

I have no strong opinion on this one but I think given the feedback that I got I'm leaning towards what the team finds more readable.

src/routes/delegates/delegates.controller.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+
+    it('Should failure with bad request', async () => {
```
fmrsabino, line 134: nitpick: "Should fail"

src/routes/delegates/delegates.service.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+    delegateRepositoryMock,
+  );
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+  });
+
+  it('should retrieve the delegates from a safe', async () => {
+    const chainId = '1';
+    const safe = '0x5afe3855358e112b5647b952709e6165e1c1eeee';
```
fmrsabino, line 25: nitpick: you can generate a random ethereum address with faker

src/routes/delegates/delegates.service.ts
```diff
@@ -0,0 +1,49 @@
...
+      delegates.previous,
+    );
+
+    return <Page<Delegate>>{
+      count: delegates.count,
+      next: nextURL?.toString(),
+      previous: previousURL?.toString(),
+      results: delegates.results,
+    };
+  }
```
fmrsabino, line 48: We should consider returning the OpenApi entity at this level. Currently it works because the interface matches.

For this we would need to add a constructor to the Page entity on the route level (which should be done a in a separate PR).

fmrsabino (changes requested):

src/routes/delegates/delegates.controller.spec.ts
moisses89: Ok, removing the more complex checks.

src/routes/delegates/delegates.controller.spec.ts
moisses89: @fmrsabino should return a bad request, and returns 503 something is wrong

src/routes/delegates/delegates.controller.ts
hectorgomezv: If we agree on using TS guards as validation tool for request, you can find [here](https://github.com/5afe/safe-client-gateway-nest/pull/132/commits/206ab186563e3f3945facd881c6da292aca9231e#diff-12bd9e35019ab1c8ce38b567178f33911bc1253984905fb8b274f6779a804d8a) an example (pending to be reviewed). 

In my opinion would be nice to map the values in a DTO and validate it, but you could also implement the validation at `DelegatesService` level manually, no strong opinion here. And probably we could do it in a separate issue/PR as well if you prefer.

hectorgomezv (dismissed):
LGTM, just left a minor comment on validation. Feel free to implement it here or we can create another issue/PR.

src/routes/delegates/delegates.service.ts
hectorgomezv: My two cents: before returning OpenApi entities from services, I think we should think if _services_ do really belong to the _routes_ layer. I think we can all agree OpenApi definitions belong to _routes_ (they're REST-specific), but do the _services_ belong to _routes_? As I said in a previous PR, if we integrate another presentation technology (websockets, cli, gRPC, whatever) I'd expect to have _services_ available in _domain_ 🤔

fmrsabino: Services do the "business" logic associated with the route ie.: presentation layer. The term itself shouldn't be a big concern – it was taken from NestJS.

If we move the Service to the domain:
1.  What is the role of the service in the domain?
2. How do you encapsulate business logic for the route?

src/routes/delegates/delegates.controller.spec.ts
fmrsabino: > @fmrsabino should return a bad request, and returns 503 something is wrong

Your comment is unrelated with my comment right? Maybe I misunderstood. But are you saying that it should return 400 instead of 503 in this test?

src/routes/delegates/delegates.service.ts
hectorgomezv: After an offline talk, I think both of us agree this possible refactor is not needed right now, and we should focus in having feature parity with the current CGW. I'll clarify my concerns/proposal in the future 🙂 thanks.

src/routes/delegates/entities/delegate.entity.ts
fmrsabino: Sorry yes you are right. I was mixing these parameters with the query ones.

moisses89:
@fmrsabino  @hectorgomezv  Added validation of query parameters https://github.com/5afe/safe-client-gateway-nest/pull/133/commits/ca373efa3eafadba01813394915ab90945b5d80e

hectorgomezv (approved):

src/routes/delegates/entities/delegate-params.entity.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.delegator = delegator;
+    this.label = label;
+  }
+}
+
+export function isDelegateParamsDto(
+  dto: DelegateParamsDto,
+): dto is DelegateParamsDto {
+  return !!(dto.safe || dto.delegate || dto.delegator || dto.label);
+}
```
fmrsabino, line 30: I think this should not be used as a type predicate. If you look into `DelegateParamsDto` almost any TS object can matched with `DelegateParamsDto` because all the fields are nullable. So the question is not if it is a `DelegateParamsDto` but if at least one of the fields is not null.

So the following is fine:

``` typescript
if (
      !(
        delegateParams.safe ||
        delegateParams.delegate ||
        delegateParams.delegator ||
        delegateParams.label
      )
    ) {
      throw new HttpException(
        'At least one query param must be provided',
        HttpStatus.BAD_REQUEST,
      );
    }
```

Also sidenote: you have double negation `!!` in the original condition which is not needed.

fmrsabino (changes requested):

src/routes/delegates/entities/delegate-params.entity.ts
moisses89: Ok, agree maybe the function could be called isValid... instead isType. But is ok. 
Regarding double negation I was aware, because should return `true` when was valid. 
```
> !(("") || ("") || (null) )
true
> !!(("Hello") || ("") || (null) )
true
```

moisses89: Done [bfc29e8c8c0a08242b33faf5fcedf88679d43664](https://github.com/5afe/safe-client-gateway-nest/pull/133/commits/bfc29e8c8c0a08242b33faf5fcedf88679d43664)

fmrsabino (approved):

PR #137, opened 2022-10-19
Use allowUnionTypes AJV configuration option

hectorgomezv:
This PR aims to remove the warning in test runs regarding the usage of union types in JSON Schemas. Example:

```
console.warn
      strict mode: use allowUnionTypes to allow union type keyword at "dataDecodedParameter/properties/valueDecoded" (strictTypes)
```

This is necessary for `DataDecoded` entity, since this is the first time we have an entity with a field which can be both an object or an array.


fmrsabino (approved):
🚀

PR #138, opened 2022-10-19
Set data field as optional in NetworkResponseError

fmrsabino:
- A network responses do not require a body – the same applies to an error response
- `DataSourceError` now sets its error message as "An error occurred" if no message was present in the body (in the format of `{ message: <string> }`
- Improves `NetworkResponseError` type predicate – it now checks for the error range
- Adjusts tests that were relying on error responses – errors are now correctly forwarded to the routes

### DIFF
--- a/src/datasources/errors/http-error-factory.ts
+++ b/src/datasources/errors/http-error-factory.ts
@@ -37,8 +37,9 @@ export class HttpErrorFactory {
   }
 }
 
-function isNetworkResponseError(error: any): error is NetworkResponseError {
-  return (
-    error.status !== undefined && error.status >= 400 && error.status < 600
-  );
+function isNetworkResponseError(
+  error: NetworkError | Error,
+): error is NetworkResponseError {
+  const responseError = error as NetworkResponseError;
+  return responseError.status >= 400 && responseError.status < 600;
 }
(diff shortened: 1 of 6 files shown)


src/datasources/errors/http-error-factory.ts
```diff
@@ -36,9 +37,8 @@ export class HttpErrorFactory {
...
 
-function isNetworkResponseError(
-  error: NetworkError,
-): error is NetworkResponseError {
-  const e = error as NetworkResponseError;
-  return e.data !== undefined && e.status !== undefined;
+function isNetworkResponseError(error: any): error is NetworkResponseError {
+  return (
+    error.status !== undefined && error.status >= 400 && error.status < 600
+  );
```
hectorgomezv, line 43: I think the `!== undefined` check should be covered by the error range check, right? I mean, if `status` is in range it is defined, and if is undefined, it isn't in range. Maybe I'm missing something... 

Apart from that, why is `any` used here? I mean, `unknown` could be used, or even `error: NetworkError | Error)`, and keep the cast.

```javascript
function isNetworkResponseError(error: NetworkError | Error) ...
```

fmrsabino: I agree. Updated!

Issue #134, opened 2022-10-18
Test Redis


Uxio0:
## Background
As Redis is the core data storage for this service, we should test it properly, as getting the service to production just with mocks is not desirable.

## Problems
- Redis keeps state, so it should be empty before/after tests and we should be aware of that.
- Redis should be configured in Github Actions

## Objective
- We at least should have a test involving redis writting/reading to correctly check if commands and serializing/deserializing is working.
- Redis test during E2E tests should be enough for now.

PR #140, opened 2022-10-20
Add E2E tests to CI

### LINKED ISSUE #128: Add end-to-end tests into CI execution
In order to have a greater test coverage and detect defects on development time, it would be very useful to implement an end to end tests suite into the CI pipeline.

### Requirements
- A fresh Redis instance should be created/destroyed during the CI execution. 
- Some environment variables should be set in the CI environment, so Github Secrets could be used.

### Decision points
- This service uses both Safe Transaction Service and Safe Config Service as data providers. Which environment should be used by the end to end tests? Production?
- Related to the previous question: which chain(s) should be used? Goerli? Mainnet?
- Also related: should we test side-effect (post, put, delete, etc.) endpoints just for Goerli and non-side-effect ones (get) for several networks including Mainnet?

Please comment 🙂 @fmrsabino @Uxio0 @moisses89

hectorgomezv:
Closes #128 

This PR aims to create the first end to end test in the project. For that, several things are introduced:
- A Redis instance which will be created/destroyed in the CI execution.
- Request payload and also expected response payload which are stored in the route's `__tests__/resources/` folder.
- Several environment variables are set, and some Github secrets regarding Rates Exchange external service are also in this project settings as well. The environment for the test is pointing **staging** Config Service, and **Görli** testnet (hardcoded per test class, could be parametrised).
- A new step in CI execution is created to execute E2E tests, this prevents E2E tests execution while developing.

**Note:** this is not closing #134 since it's just spinning up a Redis instance, but no Redis writes/reads are implied in the tested endpoint. I'll create a separate PR implementing an E2E test which actually involves Redis.


hectorgomezv:
**Update:** I created an specific task for CI execution: `yarn test:ci`. By using this task both E2E and Unit tests are executed, and their coverage is merged. Both `yarn:test` and `yarn:e2e` still available to use them during development.

src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
```diff
@@ -0,0 +1,56 @@
...
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleRef.createNestApplication();
+    await app.init();
+  });
+
+  it('should init the app', () => {
+    expect(app).toBeDefined();
+  });
```
fmrsabino, line 24: I would extract this to a different test since `AppModule` is the main application module and this test is more specific to the Data Decoder

jest-ci.json
```diff
@@ -0,0 +1,10 @@
+{
+  "moduleFileExtensions": ["js", "json", "ts"],
+  "rootDir": ".",
+  "roots": ["./src"],
+  "testEnvironment": "node",
+  "testRegex": [".*\\.spec\\.ts$", ".*\\e2e-spec\\.ts$"],
+  "transform": {
+    "^.+\\.(t|j)s$": "ts-jest"
+  }
+}
```
fmrsabino, line 10: I think it makes sense to move this file to `src/test`. This is where the E2E configuration is also stored.

src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
hectorgomezv: Makes sense, will do it 👍🏻

jest-ci.json
hectorgomezv: I tried it but I got some problems with the paths and the coverage reports, I will go a bit deeper with it to try to found a solution 👍🏻

src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
fmrsabino: I would include it in `app.module.spec.ts`. Now it might make sense to not skip the existing test in this setup

jest-ci.json
fmrsabino: It should work if you set `"coverageDirectory": "../coverage"` 👍 

```json
{
  "moduleFileExtensions": ["js", "json", "ts"],
  "rootDir": ".",
  "roots": ["../src"],
  "testEnvironment": "node",
  "testRegex": [".*\\.spec\\.ts$", ".*\\e2e-spec\\.ts$"],
  "transform": {
    "^.+\\.(t|j)s$": "ts-jest"
  },
  "coverageDirectory": "../coverage"
}

```

fmrsabino: Scratch the above ☝️ Coverage is generated but it's empty.

If you can't find a solution right now to this I would not spend too much time on it. We can take a look another time into it.

src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
hectorgomezv: Changed in [1d100ea921619171a818544aff2a1defbe196400](https://github.com/5afe/safe-client-gateway-nest/pull/140/commits/1d100ea921619171a818544aff2a1defbe196400)

jest-ci.json
hectorgomezv: Changed in [0eeb45cd623c9ecc22644f8be9c13d3551685aac](https://github.com/5afe/safe-client-gateway-nest/pull/140/commits/0eeb45cd623c9ecc22644f8be9c13d3551685aac) 

Setting `coverageDirectory` did the trick 👍

src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
fmrsabino: > I would include it in `app.module.spec.ts`. Now it might make sense to not skip the existing test in this setup

@hectorgomezv did you consider this one?

hectorgomezv: I merged both test files, thanks!

fmrsabino (approved):
Nice! 🚀

PR #141, opened 2022-10-20
Fix error instantiation in AxiosNetworkService

fmrsabino:
- We have recently changed the body of `NetworkResponseError` to be optional – with this change the order of the parameters also changed
- However this was not reflected in `AxiosNetworkService.handleError` – this is now fixed and status comes as the first argument followed by data


moisses89 (approved):
:rocket:

PR #142, opened 2022-10-20
Add post delegates

### LINKED ISSUE #102: Add POST Delegates
- Add `POST /v1/chains/<chain_id>/delegates` route
- Tests should be added

moisses89:
closes #102 
- Add post delegates
- Test validation 
- Test created successfully

### DIFF
--- a/src/domain/delegate/delegate.repository.ts
+++ b/src/domain/delegate/delegate.repository.ts
@@ -4,7 +4,6 @@ import { IDelegateRepository } from './delegate.repository.interface';
 import { Delegate } from './entities/delegate.entity';
 import { Page } from '../entities/page.entity';
 import { DelegateValidator } from './delegate.validator';
-import { NetworkResponse } from '../../datasources/network/entities/network.response.entity';
 
 @Injectable()
 export class DelegateRepository implements IDelegateRepository {
@@ -45,7 +44,7 @@ export class DelegateRepository implements IDelegateRepository {
     delegator?: string,
     signature?: string,
     label?: string,
-  ): Promise<NetworkResponse<any>> {
+  ): Promise<unknown> {
     const transactionService =
       await this.transactionApiManager.getTransactionApi(chainId);
     const result = await transactionService.postDelegates(
--- a/src/domain/delegate/delegate.repository.ts
+++ b/src/domain/delegate/delegate.repository.ts
@@ -4,6 +4,7 @@ import { IDelegateRepository } from './delegate.repository.interface';
 import { Delegate } from './entities/delegate.entity';
 import { Page } from '../entities/page.entity';
 import { DelegateValidator } from './delegate.validator';
+import { NetworkResponse } from '../../datasources/network/entities/network.response.entity';
 
 @Injectable()
 export class DelegateRepository implements IDelegateRepository {
@@ -35,5 +36,25 @@ export class DelegateRepository implements IDelegateRepository {
 
     page?.results.map((result) => this.delegateValidator.validate(result));
     return page;
+  }
+
+  async postDelegates(
+    chainId: string,
+    safeAddress?: string,
+    delegate?: string,
+    delegator?: string,
+    signature?: string,
+    label?: string,
+  ): Promise<NetworkResponse<any>> {
+    const transactionService =
+      await this.transactionApiManager.getTransactionApi(chainId);
+    const result = await transactionService.postDelegates(
+      safeAddress,
+      delegate,
+      delegator,
+      signature,
+      label,
+    );
+    return result;
   }
 }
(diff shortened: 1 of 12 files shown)


src/domain/delegate/delegate.repository.ts
```diff
@@ -36,4 +37,24 @@ export class DelegateRepository implements IDelegateRepository {
...
   }
+
+  async postDelegates(
+    chainId: string,
+    safeAddress?: string,
+    delegate?: string,
+    delegator?: string,
+    signature?: string,
+    label?: string,
+  ): Promise<NetworkResponse<any>> {
```
fmrsabino, line 48: The domain should not expose data-source related entities.

If we don't care about the object returned we should probably use `unknown` here – `Promise<unknown>`

src/routes/delegates/entities/create-delegate.entity.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export function isCreateDelegateDto(
+  dto: CreateDelegateDto,
+): dto is CreateDelegateDto {
+  if (!dto.safe) {
+    return (
+      isHex(dto.delegate) &&
+      isHex(dto.delegator) &&
+      isHex(dto.signature) &&
+      typeof dto.label === 'string'
```
fmrsabino, line 39: Checking if `label` is a `string` is redundant as this is already provided by the type system (`label: string`)

src/routes/delegates/entities/create-delegate.entity.ts
```diff
@@ -0,0 +1,50 @@
...
+    );
+  } else {
+    return (
+      isHex(dto.delegate) &&
+      isHex(dto.delegator) &&
+      isHex(dto.signature) &&
+      isHex(dto.safe) &&
+      typeof dto.label === 'string'
```
fmrsabino, line 47: Same as above

src/routes/delegates/entities/create-delegate.entity.ts
```diff
@@ -0,0 +1,50 @@
...
+    );
+  }
```
fmrsabino, line 49: This expression can be simplified with the following:

```typescript
  const validSafe = dto.safe ? isHex(dto.safe) : true;
  return (
    isHex(dto.delegate) &&
    isHex(dto.delegator) &&
    isHex(dto.signature) &&
    validSafe
  );
```

src/routes/delegates/delegates.controller.spec.ts
```diff
@@ -110,4 +111,33 @@ describe('Delegates controller', () => {
...
+      const chainResponse = chainFactory(chainId);
+      mockNetworkService.get.mockResolvedValueOnce({ data: chainResponse });
+      mockNetworkService.post.mockResolvedValueOnce({ status: 201 });
+
+      await request(app.getHttpServer())
+        .post(`/chains/${chainId}/delegates/`)
+        .send(body)
+        .expect(201);
+    });
+  });
```
fmrsabino, line 142: Additionally I would add a test where you send a malformed POST body and another one that tests what is returned in case the POST request fails

src/routes/delegates/delegates.controller.ts
```diff
@@ -57,4 +59,13 @@ export class DelegatesController {
       paginationData,
     );
   }
+
+  @ApiCreatedResponse()
+  @Post('chains/:chainId/delegates')
+  async getDataDecoded(
```
fmrsabino, line 65: `postDelegates` or similar

src/routes/delegates/delegates.service.ts
```diff
@@ -58,4 +63,25 @@ export class DelegatesService {
...
+  async postDelegates(
+    chainId: string,
+    createDelegateDto: CreateDelegateDto,
+  ): Promise<NetworkResponse> {
+    if (!isCreateDelegateDto(createDelegateDto)) {
+      throw new HttpException(
+        'Invalid payload',
+        HttpStatus.UNPROCESSABLE_ENTITY,
+      );
+    }
```
fmrsabino, line 76: What about forwarding the error response for the Safe Transaction Service? Doesn't it provide more information on what actually failed (I believe this is the current behaviour of the current CGW).

fmrsabino (changes requested):

src/routes/delegates/delegates.controller.spec.ts
moisses89: Ok agree

src/routes/delegates/delegates.service.ts
moisses89: I will check it to return the same in the message :slightly_smiling_face:

moisses89: Ok, I remove then `isCreateDelegate` function to return the error from the `transaction-api` and added a test that check the error is correctly forwarded.

moisses89: Done here https://github.com/5afe/safe-client-gateway-nest/pull/142/commits/90a353171bf0a777e5a2d6828b35e5b80f324a39

fmrsabino (approved):
🚀

PR #143, opened 2022-10-21
Test Redis integration on E2E CI tests

### LINKED ISSUE #134: Test Redis

hectorgomezv:
Closes #134 

This PR aims to test Redis integration during CI pipeline E2E tests execution.

I worked under the premise that abstracting the cache provider is not a priority here. Of course this is debatable, if you feel more confortable we could develop an specific cache client or an implementation on `ICacheService`, but since this is an E2E/blackbox test, I prefer to stay low-level, and avoid importing code from the service itself (i.e. `RedisCacheService`).

Of course this has the downside that if we want to change Redis by another caching solution in the future, we would need to rewrite some parts of the E2E tests, but this seems unlikely, and we could even encapsulate Redis-specific code in a E2E-related cache client as well if you prefer. 

Any feedback is appreciated of course :)

### DIFF
--- a/src/routes/contracts/__tests__/get-contract.e2e-spec.ts
+++ b/src/routes/contracts/__tests__/get-contract.e2e-spec.ts
@@ -1,0 +1,59 @@
+import * as request from 'supertest';
+import { INestApplication, Logger } from '@nestjs/common';
+import { Test } from '@nestjs/testing';
+import { AppModule } from '../../../app.module';
+import { readFileSync } from 'fs';
+import { Contract } from '../entities/contract.entity';
+import { RedisClientType, createClient } from 'redis';
+
+const logger = new Logger(redisClientFactory.name);
+
+async function redisClientFactory(): Promise<RedisClientType> {
+  const { REDIS_HOST = 'localhost', REDIS_PORT = 6379 } = process.env;
+  const client: RedisClientType = createClient({
+    url: `redis://${REDIS_HOST}:${REDIS_PORT}`,
+  });
+  client.on('error', (err) => logger.error('Redis Client Error', err));
+  client.connect();
+  return client;
+}
+
+describe('Get contract e2e test', () => {
+  let app: INestApplication;
+  let redisClient: RedisClientType;
+  const chainId = '5'; // Görli testnet
+
+  beforeAll(async () => {
+    const moduleRef = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleRef.createNestApplication();
+    redisClient = await redisClientFactory();
+    await app.init();
+  });
+
+  it('GET /contracts/<address>', async () => {
+    const contractAddress = '0x7cbB62EaA69F79e6873cD1ecB2392971036cFAa4';
+    const expectedResponse: Contract = JSON.parse(
+      readFileSync(
+        'src/routes/contracts/__tests__/resources/contract-expected-response.json',
+        {
+          encoding: 'utf-8',
+        },
+      ),
+    );
+
+    await request(app.getHttpServer())
+      .get(`/chains/${chainId}/contracts/${contractAddress}`)
+      .expect(200)
+      .then(({ body }) => {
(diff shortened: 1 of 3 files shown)


src/routes/contracts/__tests__/get-contract.e2e-spec.ts
```diff
@@ -0,0 +1,63 @@
...
+        expect(body).toEqual(expectedResponse);
+      });
+
+    const cacheContent = await redisClient.hGet(contractCacheKey, '');
+    expect(cacheContent).toEqual(JSON.stringify(expectedResponse));
+  });
+
+  afterAll(async () => {
+    await app.close();
+    await redisClient.flushAll();
```
Uxio0, line 51: Nice, flushing before and after

Uxio0 (approved):

src/routes/contracts/__tests__/get-contract.e2e-spec.ts
```diff
@@ -0,0 +1,63 @@
...
+});
+
+async function redisClientFactory(): Promise<RedisClientType> {
+  const { REDIS_HOST = 'localhost', REDIS_PORT = 6379 } = process.env;
+  const client: RedisClientType = createClient({
+    url: `redis://${REDIS_HOST}:${REDIS_PORT}`,
+  });
+  client.connect();
+  return client;
+}
```
fmrsabino, line 63: Why not use the same factory that we have in `cache.module.ts` (`redisClientFactory`)?

src/routes/contracts/__tests__/get-contract.e2e-spec.ts
```diff
@@ -0,0 +1,63 @@
...
+  beforeAll(async () => {
+    const moduleRef = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleRef.createNestApplication();
+    await app.init();
+
+    redisClient = await redisClientFactory();
+    await redisClient.flushAll();
```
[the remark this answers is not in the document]
hectorgomezv: In this case it doesn't matter so much since there's only a test (and it won't be more tests for /contracts I think), but I agree that _in the majority cases_ is a best practice to flush per test. I will send a PR to change it 🙂

src/routes/contracts/__tests__/get-contract.e2e-spec.ts
hectorgomezv: Do you mean copying the exact code, importing the function or injecting the client?

src/routes/contracts/__tests__/get-contract.e2e-spec.ts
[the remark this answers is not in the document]
fmrsabino: In this case that is true I think it's better to avoid a situation where we are debugging the state of one test only to find out that the state is reset only after the suite finishes.

So even though both accomplish the same thing I still think `beforeEach` and `afterEach` are more correct and represent better what needs to happen at the beginning, before each test, after each test and at the end.

src/routes/contracts/__tests__/get-contract.e2e-spec.ts
fmrsabino: Using as in importing the function. But this is also fine.

Importing would require additional dependencies.

src/routes/contracts/__tests__/get-contract.e2e-spec.ts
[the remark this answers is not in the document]
hectorgomezv: Changed [here](https://github.com/5afe/safe-client-gateway-nest/pull/145) 👍

PR #146, opened 2022-10-24
Add delete delegate

### LINKED ISSUE #144: Add DELETE delegates endpoint
- [x] Add DELETE /v1/chains/<chain_id>/delegates route
- [x] Tests should be added
- [x] OpenApi specification should be added

moisses89:
Closes #144 
- Add delete method
- Add tests

### DIFF
--- a/src/datasources/network/axios.network.service.spec.ts
+++ b/src/datasources/network/axios.network.service.spec.ts
@@ -12,6 +12,7 @@ import {
 const axios = {
   get: jest.fn(),
   post: jest.fn(),
+  delete: jest.fn(),
 } as unknown as Axios;
 
 const axiosMock = jest.mocked<Axios>(axios);
@@ -156,6 +157,63 @@ describe('AxiosNetworkService', () => {
 
       expect(axiosMock.post).toBeCalledTimes(1);
       expect(axiosMock.post).toBeCalledWith(url, data, undefined);
+    });
+  });
+
+  describe('DELETE requests', () => {
+    it(`delete calls axios delete`, async () => {
+      const url = faker.internet.url();
+      const data = { some_data: 'some_data' };
+
+      await target.delete(url, data);
+
+      expect(axiosMock.delete).toBeCalledTimes(1);
+      expect(axiosMock.delete).toBeCalledWith(url, { data: data });
+    });
+
+    it(`delete forwards unknown error as NetworkOtherError`, async () => {
+      const url = faker.internet.url();
+      const data = { some_data: 'some_data' };
+      (axiosMock.delete as any).mockRejectedValueOnce(new Error('Axios error'));
+
+      await expect(target.delete(url, data)).rejects.toThrowError(
+        new NetworkOtherError('Axios error'),
+      );
+
+      expect(axiosMock.delete).toBeCalledTimes(1);
+      expect(axiosMock.delete).toBeCalledWith(url, { data: data });
+    });
+
+    it(`delete forwards response error as NetworkResponseError`, async () => {
+      const url = faker.internet.url();
+      const data = { some_data: 'some_data' };
+      const error = {
+        response: { data: 'data', status: 100 },
+      };
+      (axiosMock.delete as any).mockRejectedValueOnce(error);
+
+      await expect(target.delete(url, data)).rejects.toThrowError(
+        new NetworkResponseError(error.response.data, error.response.status),
+      );
+
+      expect(axiosMock.delete).toBeCalledTimes(1);
+      expect(axiosMock.delete).toBeCalledWith(url, { data: data });
+    });
+
+    it(`delete forwards response error as NetworkRequestError`, async () => {
+      const url = faker.internet.url();
+      const data = { some_data: 'some_data' };
+      const error = {
+        request: 'some error',
+      };
+      (axiosMock.delete as any).mockRejectedValueOnce(error);
+
+      await expect(target.delete(url, data)).rejects.toThrowError(
+        new NetworkRequestError(error.request),
+      );
+
+      expect(axiosMock.delete).toBeCalledTimes(1);
+      expect(axiosMock.delete).toBeCalledWith(url, { data: data });
     });
   });
 });
--- a/src/datasources/network/axios.network.service.spec.ts
+++ b/src/datasources/network/axios.network.service.spec.ts
@@ -174,7 +174,7 @@ describe('AxiosNetworkService', () => {
     it(`delete forwards unknown error as NetworkOtherError`, async () => {
       const url = faker.internet.url();
       const data = { some_data: 'some_data' };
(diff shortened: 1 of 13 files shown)


src/datasources/network/axios.network.service.ts
```diff
@@ -39,6 +39,17 @@ export class AxiosNetworkService implements INetworkService {
     }
   }
 
+  async delete<T = any, R = NetworkResponse<T>>(
+    url: string,
+    data?: object,
```
moisses89, line 44: @fmrsabino @hectorgomezv this is working correctly but in the case of the params an entity was create should I create an entity for data?

hectorgomezv: In my opinion having this defined as `object` is ok, we don't need to map it.

src/datasources/network/axios.network.service.spec.ts
```diff
@@ -158,4 +159,61 @@ describe('AxiosNetworkService', () => {
...
+      expect(axiosMock.delete).toBeCalledWith(url, { data: data });
+    });
+
+    it(`delete forwards response error as NetworkResponseError`, async () => {
+      const url = faker.internet.url();
+      const data = { some_data: 'some_data' };
+      const error = {
+        response: { data: 'data', status: 100 },
+      };
+      (axiosMock.delete as any).mockRejectedValueOnce(error);
```
fmrsabino, line 193: The cast shouldn't be required

src/routes/delegates/delegates.controller.ts
```diff
@@ -67,4 +77,19 @@ export class DelegatesController {
   ): Promise<unknown> {
     return this.service.postDelegates(chainId, createDelegateDto);
   }
+
+  @ApiCreatedResponse()
```
fmrsabino, line 81: Why do you need to annotate with this decorator?

fmrsabino (changes requested):

src/datasources/network/axios.network.service.spec.ts
moisses89: Yes! done [here](https://github.com/5afe/safe-client-gateway-nest/pull/146/commits/0b1bb4165478747d8ffa85b5f9c57b07d9c6a657)

src/routes/delegates/delegates.controller.ts
moisses89: Was a copy paste error from post :man_facepalming: fixed [here](https://github.com/5afe/safe-client-gateway-nest/pull/146/commits/0b1bb4165478747d8ffa85b5f9c57b07d9c6a657)

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -168,7 +168,7 @@ export class TransactionApi implements ITransactionApi {
     }
   }
 
-  async postDelegates(
+  async postDelegate(
```
moisses89, line 171: Renamed the names because we can create one and we can delete one delegate.

src/routes/delegates/delegates.service.ts
```diff
@@ -73,4 +74,17 @@ export class DelegatesService {
       createDelegateDto.label,
     );
   }
+
+  async deleteDelegate(
+    chainId: string,
+    delegateAddress: string,
+    deleteDelegateDto: DeleteDelegateDto,
+  ): Promise<unknown> {
+    return await this.repository.deleteDelegate(
```
hectorgomezv, line 83: As we talked offline it seems `delegateAddress` is sent twice by the client (as path param and as part of the body).  As you mentioned @moisses89 it seems the Transaction Service overrides the content of the body with the `delegateAddress` value in the path ([see this](https://github.com/safe-global/safe-transaction-service/blob/0ac2f06b4e2c63e039a60f99fcb4cf2ffd91ac8f/safe_transaction_service/history/views.py#L851)) so I know we have to focus in feature-parity with the current CGW, but is this maybe a bit misleading for clients? Should we check the path and the body values match? cc. @Uxio0 @fmrsabino

src/routes/delegates/delegates.controller.spec.ts
```diff
@@ -180,4 +181,64 @@ describe('Delegates controller', () => {
...
+    });
+
+    it('Should return the tx-service error message', async () => {
+      const delegate = faker.finance.ethereumAddress();
+      const body = {
+        delegate: delegate,
+        delegator: 'delegator',
+        signature: 'signature',
+      };
+      const chainId = '99';
```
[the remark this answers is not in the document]
hectorgomezv: Yes, `faker.random.numeric()` returns a string representation of an integer 👍🏻

src/routes/delegates/delegates.controller.spec.ts
```diff
@@ -180,4 +181,64 @@ describe('Delegates controller', () => {
...
+
+      await request(app.getHttpServer())
+        .delete(`/chains/${chainId}/delegates/${body.delegate}`)
+        .send(body)
+        .expect(204);
+    });
+
+    it('Should return the tx-service error message', async () => {
+      const delegate = faker.finance.ethereumAddress();
+      const body = {
```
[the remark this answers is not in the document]
moisses89: Yeah, really it's not necessary a malformed body, but anyway maybe moving this to an e2e test could be more sense, what do you think?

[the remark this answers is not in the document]
moisses89: I will add e2e tests in a separated PR.

src/routes/delegates/delegates.controller.spec.ts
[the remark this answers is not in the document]
moisses89: Done [here](https://github.com/5afe/safe-client-gateway-nest/pull/146/commits/17f28090f22f967fe29d44ce357588391b029f3e)

src/routes/delegates/delegates.controller.spec.ts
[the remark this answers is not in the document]
moisses89: Removed the body [here](https://github.com/5afe/safe-client-gateway-nest/pull/146/commits/17f28090f22f967fe29d44ce357588391b029f3e)

fmrsabino (approved):

src/routes/delegates/delegates.controller.spec.ts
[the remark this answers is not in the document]
hectorgomezv: Great, thanks! 🙂

hectorgomezv (approved):

src/routes/delegates/delegates.service.ts
Uxio0: I think that's something that I started doing with the `safe address` and I extended that to the delegates. There are a lot of places in the **tx service** where `safe address` can be part of the body but it will be derived from the `url`, and I did the same for the delegate address

Uxio0: https://github.com/safe-global/safe-transaction-service/pull/1135

PR #149, opened 2022-10-26
Fix to import route entities instead of domain entities for ChainsController

hectorgomezv:
This PR fixes some bad imports in the ChainsController class. Domain entities were imported instead of the routes ones.


src/routes/common/entities/page.entity.ts
```diff
@@ -18,8 +18,8 @@ export abstract class Page<T> implements DomainPage<T> {
   @ApiProperty()
   count: number;
   @ApiProperty()
-  next: string;
+  next?: string;
   @ApiProperty()
-  previous: string;
+  previous?: string;
```
moisses89, line 23: @ApiPropertyOptional

hectorgomezv: Changed!

moisses89 (approved):
Looks good to me!

Issue #165, opened 2022-10-27
Add support for /v1/safe-apps/


fmrsabino:
Support should be added to `IConfigApi`

hectorgomezv:
Closed in https://github.com/5afe/safe-client-gateway-nest/pull/169

PR #169, opened 2022-10-27
Add /v1/chains/<chain_id>/safe-apps route

### LINKED ISSUE #147: Add GET SafeApps endpoint
- Add /v1/chains/<chain_id>/safe-apps route
- Tests should be added

hectorgomezv:
Closes #147 and #165 

- [X] Add /v1/chains/<chain_id>/safe-apps route
- [X] Caching support should be added
- [X] Tests should be added


src/domain/safe-apps/entities/safe-app.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { SafeAppAccessControl } from './safe-app-access-control.entity';
+import { SafeAppProvider } from './safe-app-provider.entity';
+
+export interface SafeApp {
+  id: number;
+  url: string;
+  name: string;
+  iconUrl: string;
+  description: string;
+  chainIds: string[];
```
moisses89, line 10: Should be an array of integers. https://github.com/safe-global/safe-config-service/blob/main/src/safe_apps/models.py#L48

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export const safeAppSchema: JSONSchemaType<SafeApp> = {
+  type: 'object',
+  properties: {
+    id: { type: 'number' },
+    url: { type: 'string' },
+    name: { type: 'string' },
+    iconUrl: { type: 'string' },
+    description: { type: 'string' },
+    chainIds: { type: 'array', items: { type: 'string' } },
```
moisses89, line 33: The same as above.

moisses89 (changes requested):
`chainIds` should be an array of integers, but the e2e test pass without error? We should review this.

src/domain/safe-apps/entities/safe-app.entity.ts
hectorgomezv: Correct, this is an array of integers in the data source (Config Service), but the current implementation of the client gateway converts the integers to strings (see [this](https://safe-client.staging.5afe.dev//v1/chains/5/safe-apps)). So by defining this as an array of strings in the AJV schema, in combination with `coerceTypes: true` in `json-schema.service.ts` performs a type conversion when validating. 

I left a comment in `json-schema.service.ts` precisely to document that behaviour, and thought this will be changed when implementing serialisation. This works, and the E2E test is getting an array of strings (because AJV is doing the conversion) so nothing is wrong with E2E in this case.

Since we recently decided to implement ad hoc mappings for serialisation, maybe it's time to remove this `coerceTypes` option from AJV config and implement an ad hoc mapping for this and maybe another coerced types. What do you think? I can do a separate PR for mapping this and other potentially affected types @fmrsabino @moisses89.

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
hectorgomezv: See [previous comment](https://github.com/5afe/safe-client-gateway-nest/pull/169#discussion_r1007778058).

src/domain/safe-apps/entities/safe-app.entity.ts
moisses89: Understood @hectorgomezv I'd say that we should do but not sure if now. I'd do it in a different PR for sure. :)

.vscode/launch.json
```diff
@@ -44,7 +43,23 @@
...
+      "request": "launch",
+      "name": "Debug current E2E test file",
+      "skipFiles": [
+          "<node_internals>/**"
+      ],
+      "runtimeArgs": [
+        "--inspect-brk",
+        "${workspaceRoot}/node_modules/.bin/jest",
+        "--runInBand",
+        "--config=test/jest-ci.json",
```
fmrsabino, line 62: What is the context around this change?

src/domain/safe-apps/entities/__tests__/safe-app.factory.ts
```diff
@@ -0,0 +1,30 @@
...
+  return <SafeApp>{
+    id: id ?? faker.datatype.number(),
+    url: url ?? faker.internet.url(),
+    name: name ?? faker.random.word(),
+    iconUrl: iconUrl ?? faker.internet.url(),
+    description: description ?? faker.random.word(),
+    chainIds: chainIds ?? [faker.random.numeric(), faker.random.numeric()],
+    provider: provider ?? safeAppProviderFactory(),
+    accessControl: accessControl ?? safeAppAccessControlFactory(),
+    tags: tags ?? false,
```
moisses89, line 28: Should be an array of string.

.vscode/launch.json
hectorgomezv: It's just a launch configuration for Visual Studio Code. We have another one to run a test file with an attached debugger, but it doesn't work with E2E since the jest configuration is different, so I needed to pass `jest-ci.json` configuration to this one. I find this really useful to debug a complete request flow in an E2E test but I can keep it local if you prefer 🙂

src/routes/safe-apps/__tests__/get-safe-apps.e2e-spec.ts
```diff
@@ -0,0 +1,94 @@
...
+});
+
+async function redisClientFactory(): Promise<RedisClientType> {
+  const { REDIS_HOST = 'localhost', REDIS_PORT = 6379 } = process.env;
+  const client: RedisClientType = createClient({
+    url: `redis://${REDIS_HOST}:${REDIS_PORT}`,
+  });
+  client.connect();
+  return client;
+}
```
moisses89, line 94: I remember other e2e with the same `redisClientFactory`, could be this a re-utilizable function btw e2e?

src/routes/safe-apps/entities/safe-app.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+  url: string;
+  @ApiProperty()
+  name: string;
+  @ApiProperty()
+  iconUrl: string;
+  @ApiProperty()
+  description: string;
+  @ApiProperty()
+  chainIds: string[];
+  @ApiProperty()
```
moisses89, line 19: Optional `ApiProperty`

moisses89 (changes requested):
Some minor changes

.vscode/launch.json
fmrsabino: I was just trying to understand the context of this change because it might have been pushed by mistake.

In general I find it preferable to have this in a separate PR so that we have more context in the future why this was introduced. The context of this changes are around `/safe-apps` (given the PR title + description).

hectorgomezv: I totally agree, my bad. I'll move this change to another PR 👍🏻

src/domain/safe-apps/entities/__tests__/safe-app.factory.ts
hectorgomezv: Yes, good catch!

src/domain/safe-apps/entities/safe-app.entity.ts
fmrsabino: > Correct, this is an array of integers in the data source (Config Service), but the current implementation of the client gateway converts the integers to strings (see [this](https://safe-client.staging.5afe.dev//v1/chains/5/safe-apps)).

We have two entities in the Safe Client Gateway. The domain entity should follow the spec of the data source. See [here](https://github.com/safe-global/safe-client-gateway/blob/37372e8ad8ecc6cd56860d7a22ac63291ce6b72d/src/common/models/backend/safe_apps.rs#L11).

We then have the route entity where it becomes a `String`.

The Schema validation was done to validate the schema of responses returned by other datasources. In this case it should be `number`.

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
```diff
@@ -0,0 +1,49 @@
+import { JSONSchemaType } from 'ajv';
+import { SafeApp } from '../../../../routes/safe-apps/entities/safe-app.entity';
+import { SafeAppAccessControl } from '../safe-app-access-control.entity';
+import { SafeAppProvider } from '../safe-app-provider.entity';
+
+export const safeAppProviderSchema: JSONSchemaType<SafeAppProvider> = {
```
fmrsabino, line 6: Out of curiosity, is it possible to move the `nullable` to where we use `$ref`?

Thinking about the use-case where we want to re-use the schema in a non-nullable way.
Usually I try to make the root schema as non-nullable and apply the nullable at the property level.

Wdyt?

src/datasources/config-api/config-api.service.ts
```diff
@@ -38,10 +39,29 @@ export class ConfigApi implements IConfigApi {
...
+    clientUrl?: string,
+    url?: string,
+  ): Promise<SafeApp[]> {
+    try {
+      const cacheKey = `${chainId}_safe_apps`;
+      const field = `${clientUrl}_${url}`;
+      return await this.dataSource.get(
+        cacheKey,
+        field,
+        `${this.baseUri}/api/v1/safe-apps/?chainId=${chainId}`,
```
fmrsabino, line 62: Why not have `chainId` in the `params`?

src/datasources/config-api/config-api.service.ts
```diff
@@ -38,10 +39,29 @@ export class ConfigApi implements IConfigApi {
...
       const url = `${this.baseUri}/api/v1/chains/${chainId}`;
-      return await this.dataSource.get(key, field, url);
+      return await this.dataSource.get(cacheKey, field, url);
+    } catch (error) {
+      throw this.httpErrorFactory.from(error);
+    }
+  }
+
+  async getSafeApps(
+    chainId: string,
```
fmrsabino, line 52: This should not be required – it should be possible to retrieve the collection of Safe Apps for every chain.

Even though it is required at the route level I think we should match the datasource with what is possible to get from it. Wdyt?

src/domain/safe-apps/entities/safe-app.entity.ts
hectorgomezv: Yes, I agree @fmrsabino , I tried to express the same in my previous message. This type-coercion solution was included in the project first steps when serialisation wasn't a concern. In fact I just tried to remove the `coerceTypes: true` property from AJV configuration and several controller tests fail as result: chains, balances, and this new safe-apps.

I think it's better to solve this in a separate PR to stay single-purpose. I can do it today if you agree 🙂

fmrsabino (changes requested):

.vscode/launch.json
hectorgomezv: Removed in [21ff50f](https://github.com/5afe/safe-client-gateway-nest/commit/21ff50f7bc77e3cb11bcbaa2c1cf0322c60cd08c)

src/domain/safe-apps/entities/__tests__/safe-app.factory.ts
hectorgomezv: Changed in [21ff50f](https://github.com/5afe/safe-client-gateway-nest/commit/21ff50f7bc77e3cb11bcbaa2c1cf0322c60cd08c)

src/routes/safe-apps/entities/safe-app.entity.ts
hectorgomezv: Changed in [21ff50f](https://github.com/5afe/safe-client-gateway-nest/commit/21ff50f7bc77e3cb11bcbaa2c1cf0322c60cd08c)

src/domain/safe-apps/entities/safe-app.entity.ts
fmrsabino: Yeah I agree. Let's open a PR with this change and see what is affected by it 👍

src/datasources/config-api/config-api.service.ts
hectorgomezv: Yes, I think it's better to match the datasource at the datasource level, and let the param as required at the route level. I'll change it, thanks 👍🏻

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
hectorgomezv: Yes, it's better, I agree, thanks!

moisses89 (dismissed):

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
hectorgomezv: Hmm, I tried it and got this error:
```
Error: "nullable" cannot be used without "type"
```

Now I recall that moving `nullable` to the non-root schema was to in order to keep using `$ref` and `JSONSchemaType` (TS type-aware) instead of simply `Schema` (untyped). I couldn't find another way except by moving it, I could let the `nullable` in the root schema like this:
```
provider: { anyOf: [{ type: 'null' }, { $ref: 'safeAppProviderSchema' }] },
```
But that would make us to use `Schema` instead of `JSONSchemaType`. At leat I couldn't find any workaround.

src/datasources/config-api/config-api.service.ts
hectorgomezv: Changed in [32a5f1eb7e](https://github.com/5afe/safe-client-gateway-nest/commit/32a5f1eb7eb7e1e2b02d5e785aba5335a72d3cac)

src/datasources/config-api/config-api.service.ts
hectorgomezv: Changed in [32a5f1eb7e](https://github.com/5afe/safe-client-gateway-nest/commit/32a5f1eb7eb7e1e2b02d5e785aba5335a72d3cac)

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
hectorgomezv: I left it as `Schema` and moved the nullable to the root entity in [32a5f1eb7e](https://github.com/5afe/safe-client-gateway-nest/commit/32a5f1eb7eb7e1e2b02d5e785aba5335a72d3cac), please let me know if you think there's a better way 🙂

fmrsabino: I think this is ok. `JSONSchemaType` doesn't seem to cover some cases that make our schemas more readable. Recently I had an issue with using `discriminator` and `$merge` together.

For these cases I'd say that we can just use `Schema` as long as we correctly validate the datasources 👍 .

The approach that I'm considering is to always start with `JSONSchemaType`. If there's an incompatibility with it with the schema that we are writing then we fall back to `Schema`. Wdyt?

hectorgomezv: For sure, that's the approach I'm following, start with `JSONSchema` until I find some incompatibility, try to fight it in some way, and finally surrender and use `Schema` 🙂

src/datasources/config-api/config-api.service.ts
```diff
@@ -38,10 +39,29 @@ export class ConfigApi implements IConfigApi {
...
+    }
+  }
+
+  async getSafeApps(
+    chainId?: string,
+    clientUrl?: string,
+    url?: string,
+  ): Promise<SafeApp[]> {
+    try {
+      const cacheKey = `${chainId}_safe_apps`;
```
fmrsabino, line 57: Since `chainId` is optional what do you think about using this in the field parameters for the cache key. So the key is `safe_apps`. This is more in line with how we approach caching in the other methods.

test/common/cache/redis-client.factory.ts
```diff
@@ -0,0 +1,10 @@
+import { RedisClientType, createClient } from 'redis';
+
+export async function redisClientFactory(): Promise<RedisClientType> {
+  const { REDIS_HOST = 'localhost', REDIS_PORT = 6379 } = process.env;
+  const client: RedisClientType = createClient({
+    url: `redis://${REDIS_HOST}:${REDIS_PORT}`,
+  });
+  client.connect();
+  return client;
+}
```
fmrsabino, line 10: Regarding the location of this file: it is our first `ts` file outside of `src`. We have been putting testing utilities close to where they are used.

What do you think about moving this to `src/__tests__/redis-client.factory.ts`? I would probably skip creating `common/cache`. If we get more utils there we can think about a folder structure.

src/routes/safe-apps/__tests__/get-safe-apps.e2e-spec.ts
```diff
@@ -0,0 +1,86 @@
...
+    );
+    expect(cacheContent).toBeDefined();
+  });
+
+  afterAll(async () => {
+    await app.close();
+    await redisClient.flushAll();
+    await redisClient.quit();
+  });
+});
```
fmrsabino, line 86: Safe Apps can change frequently which makes this test unreliable.

I would not consider this endpoint for end to end tests unless it's with an app that we own and provide long term support.

fmrsabino (changes requested):

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
fmrsabino: > For sure, that's the approach I'm following, start with `JSONSchemaType` until I find some incompatibility, try to fight it in some way, and finally surrender and use `Schema` 🙂

Yes. I agree! 💯

src/datasources/config-api/config-api.service.ts
hectorgomezv: Yes, totally. Thanks for pointing it out 🙂

src/routes/safe-apps/__tests__/get-safe-apps.e2e-spec.ts
hectorgomezv: Yes, that's something I also considered, I thought that maybe staging shouldn't change so much but you're right, having a fixed JSON to match against changing data it isn't reliable. What do you think about just filtering by Transaction Builder (an app we own as you said)? Another option could be just validate the response structure (an array of objects with a id, url, name...)

Do you see any of this options feasible? Otherwise I could remove the test completely 🤔

fmrsabino: I would say that the test itself is valuable and one option is mocking the networking part so you cover the route.

I will leave this up to you. I think testing the Transaction Builder should be fine but I'm not 100% confident that it will work long term as the staging environment can be used for testing different configurations.

One thing is blockchain data which is the same between our environments. But metadata related with the blockchain (contract names, chain data, safe apps) are more subject to changes.

test/common/cache/redis-client.factory.ts
hectorgomezv: Changed in [aba0246](https://github.com/5afe/safe-client-gateway-nest/commit/aba0246a22e5a542dacc772f2ff0f7673a8b1ac7)

src/datasources/config-api/config-api.service.ts
hectorgomezv: Changed in [aba0246](https://github.com/5afe/safe-client-gateway-nest/commit/aba0246a22e5a542dacc772f2ff0f7673a8b1ac7)

src/routes/safe-apps/__tests__/get-safe-apps.e2e-spec.ts
hectorgomezv: Changed in [d9d1356](https://github.com/5afe/safe-client-gateway-nest/commit/d9d1356e98c2b05721a2e309e27194d2c784757f), I removed the JSON fixed payloads I'm just verifying the responses contain the expected data structure, and also, in the case of the Transaction Builder Safe App, the response contains just 1 element with the expected name and url (that shouldn't change a lot I assume). Does it sound good?

fmrsabino: The test `GET /chains/<chainId>/safe-apps?url=${transactionBuilderUrl}` fails for me. Which configuration are you using to test it? I believe this might be related with the config service url.

The configuration should probably be set by the test instead of relying on the local environment.

hectorgomezv: I added default values for e2e/ci tests execution in [a36e858](https://github.com/5afe/safe-client-gateway-nest/pull/169/commits/a36e858bd88b6d1cf91311cfdf364fa59e7e711f)

fmrsabino (approved):

PR #170, opened 2022-10-28
Add support to get all safe transactions

fmrsabino:
Closes https://github.com/5afe/safe-client-gateway-nest/issues/127

- Adds support to get all transactions for a safe in `ITransactionApi`
- Adds support to get the transaction history for a safe in `ISafeRepository`
- Adds the following entities:
  * `EthereumTransaction`
  * `ModuleTransaction`
- Additionally adds the entities that just have a type on top of the previous ones:
  * `EthereumTransactionType` (based on `EthereumTransaction`)
  * `ModuleTransactionType` (based on `ModuleTransaction`)
  * `MultisigTransactionType` (based on `MultisigTransaction`)
  * `TransactionType` which is a Union of the above
- Extracts `Operation` to its own entity file
- Adds `discriminator` functionality to `Ajv` – this is particularly useful when we have a collection of multiple types which have a "type" label to identify which schema to expect – see https://ajv.js.org/json-schema.html#discriminator

------

You might notice that some schemas are almost identical (if you compare the `*-type` with the "non-type" schemas). Initially I tried using `$patch` and `$merge` but these keywords do not work well with `discriminator`.

Since `$patch` and `$merge` require an additional dependency I've decided to stick with the `discriminator` feature so that's why the schema is 90% similar between those classes. More info here: https://github.com/ajv-validator/ajv-merge-patch


src/domain/safe/entities/schemas/multisig-transaction-type.schema.ts
```diff
@@ -0,0 +1,50 @@
...
+
+export const multisigTransactionTypeSchema: Schema = {
+  type: 'object',
+  properties: {
+    txType: { type: 'string', const: 'MULTISIG_TRANSACTION' },
+    safe: { type: 'string' },
+    to: { type: 'string' },
+    value: { type: 'string', nullable: true },
+    data: { type: 'string', nullable: true },
+    dataDecoded: { type: 'object', nullable: true },
```
hectorgomezv, line 12: Should we use the existent schema? 🤔 
```javascript
dataDecoded: dataDecodedSchema,
```

src/domain/safe/entities/multisig-transaction.entity.ts
```diff
@@ -10,7 +8,7 @@ export interface Confirmation {
   signature?: string;
 }
 
-export interface MultisigTransaction {
+export type MultisigTransaction = {
```
hectorgomezv, line 11: Not strictly related to the PR, but could be a good point to change this?
```
dataDecoded?: any;
```
vs
```
dataDecoded?: DataDecoded;
```

hectorgomezv (commented):
Great job!

OOC: is there any particular reason to choose the name `TransactionType` (entity, schema, validator) instead of simply `Transaction`? 🙂

src/domain/safe/entities/schemas/multisig-transaction-type.schema.ts
fmrsabino: Yes we should 👍

src/domain/safe/entities/multisig-transaction.entity.ts
fmrsabino: I would tackle it in a different PR. There were already quite some schemas introduced and I didn't want to add a new one here.

fmrsabino:
> Great job!
> 
> OOC: is there any particular reason to choose the name `TransactionType` (entity, schema, validator) instead of simply `Transaction`? 🙂

@hectorgomezv I named this way just because it is a union of the types `EthereumTransactionType`, `ModuleTransactionType` and `MultisigTransactionType`. We can see how in this end this related with other possible transaction entities.

It was a bit tricky to name these ones since `EthereumTransactionType` is a `EthereumTransaction` with the type definition.

The ideal solution in my opinion would actually to have the type as required in the schema but not required for the type which the schema represents. If this is possible then I'd consider removing the `*Type` types.

src/domain/safe/entities/schemas/multisig-transaction-type.schema.ts
fmrsabino: [1a34acc](https://github.com/5afe/safe-client-gateway-nest/pull/170/commits/1a34acc0a7af949761b2b34a22ac35b73c951e41)

hectorgomezv (approved):

PR #173, opened 2022-10-31
Split E2E tests execution in CI

hectorgomezv:
This PR splits up the E2E tests execution from the unit tests execution.
- E2E tests now have a timeout of 20 seconds.
- Github Actions runs both execution jobs (unit tests and end to end tests) in parallel.
- The Action fails fast if one job fails.
- Coverage gets aggregated when both executions are finished.
- An additional task and configuration (`test:all`) is added to the project, in order to ease the development process having a way to launch all the tests in one run. This executes both unit and e2e tests, and is intended for development (not CI) purposes.


package.json
```diff
@@ -37,8 +37,8 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json",
-    "test:ci": "jest --coverage --config ./test/jest-ci.json",
+    "test:e2e": "jest --coverage --config ./test/jest-e2e.json",
+    "test:all": "jest --coverage --config ./test/jest-all.json",
```
fmrsabino, line 41: My suggestion above applies here to. What do you think about having `test:all` and `test:all:cov`?

test/jest-e2e.json
```diff
@@ -1,10 +1,12 @@
 {
   "moduleFileExtensions": ["js", "json", "ts"],
-  "rootDir": ".",
-  "roots": ["../src"],
+  "rootDir": "../",
```
fmrsabino, line 3: This file didn't move so what is the reason for the reference to the `rootDir` changing?

.github/workflows/ci.yml
```diff
@@ -29,6 +29,10 @@ jobs:
 
   tests:
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: true
```
fmrsabino, line 33: Imagine the following scenario:

- `test:cov` is successful but executes in 10 seconds
- `test:e2e` is NOT successful but executes in 5 seconds

Given this I think the outcome for both would be a failed execution right? So in the end those are not entirely split.

I think it's important that we get independent feedback from both tasks and allow each task to be re-executed in case of failure.

fmrsabino (changes requested):

.github/workflows/ci.yml
hectorgomezv: In that scenario, acording to the docs (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast), `test:e2e` would fail, and `test:cov` would be cancelled by the CI. We would still having independent feedback for both jobs. In fact, by reading the docs I realised this is the default behavior. Maybe we could set this explicitly to `false` in order to be able to re-run just one job. What do you think?

.github/workflows/ci.yml
```diff
@@ -47,14 +51,26 @@ jobs:
...
         env:
           EXCHANGE_API_BASE_URI: http://api.exchangeratesapi.io/v1
           EXCHANGE_API_KEY: ${{ secrets.EXCHANGE_API_KEY }}
           REDIS_HOST: localhost
           REDIS_PORT: 6379
           SAFE_CONFIG_BASE_URI: ${{ secrets.SAFE_CONFIG_BASE_URI }}
-      - name: Upload coverage
-        uses: coverallsapp/github-action@1.1.3
+      - name: Coveralls Parallel
+        uses: coverallsapp/github-action@master
```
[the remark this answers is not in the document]
hectorgomezv: Thanks, I just copied it from the Coveralls docs, but I agree it's better to avoid pulling `master` 👍🏻

test/jest-e2e.json
hectorgomezv: After several tests, I concluded `rootDir` has to be set to the root folder in order to correctly generate the coverage report (otherwise Jest generates an empty report). I had to change it because before this change coverage wasn't generated, so the tests ran correctly, but it seems it has to be changed to generate the coverage.

.github/workflows/ci.yml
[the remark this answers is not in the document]
hectorgomezv: Changed in [ac37ec1](https://github.com/5afe/safe-client-gateway-nest/pull/173/commits/ac37ec1366ece0316d011ad1e4c0ed4d860e705f)

.github/workflows/ci.yml
```diff
@@ -47,14 +51,26 @@ jobs:
...
           github-token: ${{ secrets.GITHUB_TOKEN }}
+          flag-name: run-${{ matrix.task }}
+          parallel: true
+
+  finish:
+    needs: tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Coveralls Finished
+        uses: coverallsapp/github-action@master
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [ac37ec1](https://github.com/5afe/safe-client-gateway-nest/pull/173/commits/ac37ec1366ece0316d011ad1e4c0ed4d860e705f)

package.json
```diff
@@ -37,8 +37,8 @@
     "test:watch": "jest --watch",
     "test:cov": "jest --coverage",
     "test:debug": "node --inspect-brk -r tsconfig-paths/register -r ts-node/register node_modules/.bin/jest --runInBand",
-    "test:e2e": "jest --config ./test/jest-e2e.json",
-    "test:ci": "jest --coverage --config ./test/jest-ci.json",
+    "test:e2e": "jest --coverage --config ./test/jest-e2e.json",
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [ac37ec1](https://github.com/5afe/safe-client-gateway-nest/pull/173/commits/ac37ec1366ece0316d011ad1e4c0ed4d860e705f)

package.json
hectorgomezv: [ac37ec1](https://github.com/5afe/safe-client-gateway-nest/pull/173/commits/ac37ec1366ece0316d011ad1e4c0ed4d860e705f)

.github/workflows/ci.yml
hectorgomezv: I removed the fail-fast configuration option in [ac37ec1](https://github.com/5afe/safe-client-gateway-nest/pull/173/commits/ac37ec1366ece0316d011ad1e4c0ed4d860e705f). As I said, I'm ok with having an explicit `fail-fast: false` if you find it useful, just let me know 🙂

fmrsabino: > In that scenario, acording to the docs (https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast), `test:e2e` would fail, and `test:cov` would be cancelled by the CI. We would still having independent feedback for both jobs.

But then it's not really independent as one would cancel the other right? 🤔

hectorgomezv: That's right, the first that fails cancels all the other running test executions. I think it's valuable to have them totally independent from each other as you pointed out, changed in [a47ebca](https://github.com/5afe/safe-client-gateway-nest/pull/173/commits/a47ebca22cf7867f0712207413aa025b7a0956e7) 🙂

test/jest-e2e.json
fmrsabino: Interesting! Thanks for the explanation!

fmrsabino (approved):

fmrsabino (approved):
Nice! I will check the required status in order to have this mergeable. 

Update: done ✅ 

You need to rebase https://github.com/5afe/safe-client-gateway-nest/pull/176 in order for the checks to work

Issue #177, opened 2022-11-04
Implement tokens cache warm-up job


hectorgomezv:
**Context**
The service needs to retrieve tokens info in a reliable and fast way. After the implementation of #166, the service will be able to query Transaction service and store each token info in a Redis hash. 

**Problem**
This retrieval and store mechanism is not performant due to the amount of non-tokens and random addresses the service receives. This causes the service to make requests to the Transaction Service for those addresses, which forces Transaction Service to retrieve the data from the blockchain at least one time for each.

Additional [context](https://github.com/safe-global/safe-client-gateway/pull/259) (Rust Client Gateway implementation).

**Proposed solution**
Warm up the cache using a scheduled job which will iterate over the [paginated token list](https://safe-transaction-mainnet.safe.global/api/v1/tokens/) returned by the Transaction Service. This job should be executed every 10 minutes. 

Keep in mind the service could be deployed in a HA setup, so several instances of the service could be running in a given point in time. Due to this, a lock mechanism is necessary. 
- This lock will be set in Redis, so both `lastExecution` (date) and `locked` (boolean) should be managed by the service.
- When a instance executes the job, it should write `lastExecution` and set `locked` to `true`.
- If an instance sees `(now() - lastExecution) < 10 minutes` and `locked == false`  the job should be started.
- After a successful run, `locked` should be set to `false`.
- A random jitter should be added to the task execution to prevent all the instances trying to execute with the same cadence.

Note: we probably need a TTL for `locked` in Redis, to prevent the scenario where a instance doesn't finish the job and never releases the lock.

hectorgomezv:
@fmrsabino @iamacook I'm thinking of closing this one as not planned. We have been retrieving tokens one by one for a while with no issues, and we have been doing some work recently to mitigate corner cases (high amount of nested transfers, multiple parallel tokens metadata requests, etc.). So I guess this can be seen as not needed at least by now. Any concern?

iamacook:
> @fmrsabino @iamacook I'm thinking of closing this one as not planned. We have been retrieving tokens one by one for a while with no issues, and we have been doing some work recently to mitigate corner cases (high amount of nested transfers, multiple parallel tokens metadata requests, etc.). So I guess this can be seen as not needed at least by now. Do you have any concerns regarding closing this?

I see no issue. We can revisit it if necessary.

PR #178, opened 2022-11-04
Add support for tokens retrieval from Transaction Service datasource

### LINKED ISSUE #166: Add support for /v1/tokens/
Support should be added to `ITransactionApi`

hectorgomezv:
Closes #166 

This PR adds support for tokens retrieval from Transaction Service datasource. 

It implements both per-address and paginated retrieval. Depending on the outcome of #177 and related discussions regarding the tokens strategy, one of this functions could be deleted and the repository code will be written accordingly when implementing the route.


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -284,4 +285,33 @@ export class TransactionApi implements ITransactionApi {
       throw this.httpErrorFactory.from(error);
     }
   }
+
+  async getTokenInfo(address: string): Promise<Page<TokenInfo>> {
```
fmrsabino, line 289: `/tokens/<address>` doesn't return a Paginated response

fmrsabino, line 289: nitpick: `getToken` seems to fit better. Any reason for having `Info` there?

(the same applies to the entity and `getTokensInfo`)

src/domain/tokens/entities/token-info.entity.ts
```diff
@@ -0,0 +1,15 @@
+export enum TokenType {
+  Erc721 = 'ERC721',
+  Erc20 = 'ERC20',
+  NativeToken = 'NATIVE_TOKEN',
+  Unknown = 'UNKNOWN',
```
fmrsabino, line 5: The Safe Transaction Service does not have an `UNKNOWN` type for `TokenType`.

In the Rust client it's under `#[serde(other)]` which is the fallback in case matching with the other types fail.

fmrsabino (changes requested):

src/datasources/transaction-api/transaction-api.service.ts
hectorgomezv: True 😅

hectorgomezv: I changed it in [16eda59](https://github.com/5afe/safe-client-gateway-nest/pull/178/commits/16eda598ca12e2c19b144794bd4a8812dae7d5db)

I also changed `TokenInfo` to `Token` as we discussed.

hectorgomezv: Fixed in [16eda59](https://github.com/5afe/safe-client-gateway-nest/pull/178/commits/16eda598ca12e2c19b144794bd4a8812dae7d5db)

src/domain/tokens/entities/token-info.entity.ts
hectorgomezv: Fixed in [16eda59](https://github.com/5afe/safe-client-gateway-nest/pull/178/commits/16eda598ca12e2c19b144794bd4a8812dae7d5db)

src/domain/tokens/entities/schemas/token-info.schema.ts
```diff
@@ -0,0 +1,15 @@
+import { JSONSchemaType } from 'ajv';
+import { Token, TokenType } from '../token.entity';
+
+export const tokenSchema: JSONSchemaType<Token> = {
+  type: 'object',
+  properties: {
+    address: { type: 'string' },
+    decimals: { type: 'number' },
+    logoUri: { type: 'string' },
```
fmrsabino, line 9: Since we know that this is a `uri` we can do something like:

```typescript
logoUri: { type: 'string', format: 'uri' }
```

fmrsabino (changes requested):
LGTM! Just one small comment!

src/domain/tokens/entities/schemas/token-info.schema.ts
hectorgomezv: Fixed in [134ec39](https://github.com/5afe/safe-client-gateway-nest/pull/178/commits/134ec39812f27abc6b233638507d224fe415e9a1)

fmrsabino (approved):

PR #182, opened 2022-11-07
Add custom isDate AJV keyword

### LINKED ISSUE #180: Add Date format validation
- The date format is commonly used in our payloads
- We should add validation for these fields to our schema

fmrsabino:
Closes #180

- Adds a `isDate` – a custom keyword for AJV which validates date types. If a date is valid then a `Date` instance is created on the target property where `isDate` was applied
- `isDate` can be applied to any property. For the date validation and serialization to happen its value should be set to `true`: `{ isDate: true }`

### DIFF
--- a/src/domain/schema/keywords/is-date.keyword.spec.ts
+++ b/src/domain/schema/keywords/is-date.keyword.spec.ts
@@ -1,0 +1,90 @@
+import Ajv, { Schema, ValidateFunction } from 'ajv';
+import { addIsDate } from './is-date.keyword';
+import { faker } from '@faker-js/faker';
+
+describe('AJV Keyword – isDate', () => {
+  let ajv: Ajv;
+
+  beforeEach(() => {
+    ajv = new Ajv();
+    addIsDate(ajv);
+  });
+
+  describe('isDate is true', () => {
+    const schema: Schema = {
+      type: 'object',
+      properties: {
+        date: { type: 'string', isDate: true, nullable: true },
+      },
+    };
+
+    let validateFunction: ValidateFunction;
+
+    beforeEach(() => {
+      validateFunction = ajv.compile(schema);
+    });
+
+    it('not valid if date is empty string', async () => {
+      const actual = validateFunction({
+        date: '',
+      });
+
+      expect(actual).toBe(false);
+    });
+
+    it('valid if date is ISO-8601', async () => {
+      const actual = validateFunction({
+        date: faker.date.past().toISOString(),
+      });
+
+      expect(actual).toBe(true);
+    });
+
+    it('valid if date is null', async () => {
+      const actual = validateFunction({
+        date: null,
+      });
+
+      expect(actual).toBe(true);
+    });
+  });
+
+  describe('isDate is false', () => {
+    const schema: Schema = {
+      type: 'object',
+      properties: {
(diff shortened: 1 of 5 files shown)


src/domain/schema/keywords/is-date.keyword.ts
```diff
@@ -0,0 +1,37 @@
...
+ *
+ * Null date values are supported – we assume that a null date is a valid
+ * date.
+ *
+ * @param ajv - the AJV instance to which the isDate keyword should be registered
+ */
+export function addIsDate(ajv: Ajv) {
+  ajv.addKeyword({
+    keyword: 'isDate',
+    compile: (schema) => (value, obj) => {
```
hectorgomezv, line 20: I'd give `obj` a more meaningful name. I think it refers to the parent schema, so maybe `parentSchema` is a good name, as it is [in the docs](https://ajv.js.org/keywords.html#define-keyword-with-compile-function) 🙂 Wdyt?

src/domain/schema/keywords/is-date.keyword.ts
```diff
@@ -0,0 +1,37 @@
...
+      // isDate: false – no need to validate
+      if (schema === false) return true;
+      // Nullability support – if a date is null we skip validation
+      if (value == null) return true;
```
moisses89, line 24: For the following schema:
```
date: { type: 'string', isDate: true, nullable: false }
```
if date is null `isDate` returns `true` but then evaluate `nullable` and returns false-invalid schema?

fmrsabino: The `isDate` does not affect the `nullable` check. We just consider that `isDate` returns `true` if it is `null`.

So it should return invalid schema because the field `date` is `null` when the field is declared as `nullable: false`

src/domain/schema/keywords/is-date.keyword.ts
fmrsabino: Maybe `data` and `dataContext`?

See: https://github.com/ajv-validator/ajv/blob/948ffbf07345eeca286e50cc233bedc3f72e8f03/lib/types/index.ts#L51

src/domain/schema/keywords/is-date.keyword.spec.ts
```diff
@@ -0,0 +1,90 @@
...
+    });
+
+    it('not valid if date is empty string', async () => {
+      const actual = validateFunction({
+        date: '',
+      });
+
+      expect(actual).toBe(false);
+    });
+
```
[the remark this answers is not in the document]
fmrsabino: Yes I think we should test that too 👍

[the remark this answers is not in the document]
fmrsabino: [d82fd92](https://github.com/5afe/safe-client-gateway-nest/pull/182/commits/d82fd92bd4d5776069af9f72d9831d4a3ce52773)

src/domain/schema/keywords/is-date.keyword.ts
hectorgomezv: Sure, I'm ok with it 🙂

hectorgomezv (approved):
👏🏻 great job!

PR #184, opened 2022-11-09
Add MultisigTransaction retrieval by safeTxHash

### LINKED ISSUE #164: Add support for /v1/multisig-transactions/{}/
Support should be added to `ITransactionApi`

hectorgomezv:
Closes #164

This PR adds the datasource retrieval logic for pulling one Multisig Transaction from Tx Service by `SafeTransactionHash`.

### DIFF
--- a/src/domain/safe/safe.repository.interface.ts
+++ b/src/domain/safe/safe.repository.interface.ts
@@ -16,6 +16,18 @@ export interface ISafeRepository {
     offset?: number,
   ): Promise<Page<Transfer>>;
 
+  getIncomingTransfers(
+    chainId: string,
+    safeAddress: string,
+    executionDateGte?: string,
+    executionDateLte?: string,
+    to?: string,
+    value?: string,
+    tokenAddress?: string,
+    limit?: number,
+    offset?: number,
+  ): Promise<Page<Transfer>>;
+
   getQueuedTransactions(
     chainId: string,
     safeAddress: string,
--- a/src/domain/safe/safe.repository.interface.ts
+++ b/src/domain/safe/safe.repository.interface.ts
@@ -27,4 +27,9 @@ export interface ISafeRepository {
     chainId: string,
     safeAddress: string,
   ): Promise<Page<TransactionType>>;
+
+  getMultiSigTransaction(
+    chainId: string,
+    safeTransactionHash: string,
+  ): Promise<MultisigTransaction>;
 }
--- a/src/datasources/transaction-api/transaction-api.service.ts
+++ b/src/datasources/transaction-api/transaction-api.service.ts
@@ -232,6 +232,36 @@ export class TransactionApi implements ITransactionApi {
     }
   }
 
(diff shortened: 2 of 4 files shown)


src/domain/safe/safe.repository.interface.ts
```diff
@@ -27,4 +27,9 @@ export interface ISafeRepository {
     chainId: string,
     safeAddress: string,
   ): Promise<Page<TransactionType>>;
+
```
hectorgomezv, line 30: @fmrsabino @moisses89 I included this function in `SafeRepository`, but I think it's not totally related to a Safe. I tried to figure out a better place, so I ended up with the following options:
- Include it in `ChainsRepository` (importing `MultisigTransaction` entity from another _domain_).
- Add a new ad-hoc repository.
- Include it here, in `SafeRepository`.

I took the decision of including it here, because I think in the future, some Safe-related routes could need this function (I see in the Rust implementation `/v1/chains/<chain_id>/transactions/<safe_address>/propose` uses `get_transactions_details` which uses this function.

What do you think, do you see a more appropriated place?

moisses89: I think that some route uses some function/datasource shouldn't determine the place of the function.
Maybe we could put all related with transactions in a new repository called for example `History`.

fmrsabino: I think it makes sense to be in the `SafeRepository`.

When working with the domain we should consider things from the route/presentation perspective. If on a route level we want to get the multisig transactions for a specific Safe I would probably think it would be under the repository that handles data for that Safe.

As the application grows we can think if other repositories make sense but for now I think this functionality fits here.

fmrsabino (approved):

src/domain/safe/safe.repository.interface.ts
hectorgomezv: Yes, both are valid perspectives, I agree with @moisses89, in most cases the routes doesn't have to determine the repository shape. I agree also with @fmrsabino that a feature-driven design it's good for this service, since it's a gateway at the end.

This function is like an outlier because it's used to both serve the Safe feature and independently, I'm more in favor of letting it here in `SafeRepository` and in the future we can think in a different grouping if we found a lot of outliers which could be grouped together, does it sound ok, @moisses89?

moisses89 (approved):

src/domain/safe/safe.repository.interface.ts
fmrsabino: I don't see the `SafeRepository` as a repository only to get "Safe info" but to get anything under a safe. Transactions are part of the Safe in this case so that's why I think it makes sense to be here.

However a more general concept or pattern might emerge and we can shift things around (eg.: Transaction repository). But I still think that the Safe is the "central" entity around which the transactions are made.

We can always propose a different structure/alternatives! The project structure allows us to do that easily 🙂

PR #188, opened 2022-11-10
Add /health endpoint

### LINKED ISSUE #160: Add /health route

hectorgomezv:
Closes #160 

This route is implemented to match the exact behaviour of the current Rust client gateway.

### DIFF
--- a/src/routes/health/health.controller.ts
+++ b/src/routes/health/health.controller.ts
@@ -1,0 +1,14 @@
+import { Controller, Get, Res } from '@nestjs/common';
+import { ApiOkResponse, ApiTags } from '@nestjs/swagger';
+import { Response } from 'express';
+
+@ApiTags('health')
+@Controller({ path: 'health' })
+export class HealthController {
+  @ApiOkResponse({ type: String })
+  @Get()
+  getHealth(@Res() res: Response): void {
+    res.setHeader('Content-Type', 'application/json');
+    res.end('""');
+  }
+}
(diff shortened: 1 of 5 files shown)


src/routes/health/health.controller.ts
```diff
@@ -0,0 +1,14 @@
...
+@ApiTags('health')
+@Controller({ path: 'health' })
+export class HealthController {
+  @ApiOkResponse({ type: String })
+  @Get()
+  getHealth(@Res() res: Response): void {
+    res.setHeader('Content-Type', 'application/json');
+    res.end('""');
+  }
+}
```
hectorgomezv, line 14: @fmrsabino @moisses89 As I said in the description this implementation mimics the behaviour of the Rust client gateway, which returns `""` with a header `Content-Type: application/json`. This feels a bit odd, I don't know this is totally required. If we could return an object instead, the implementation/response would feel more natural in my opinion:

```javascript
  getHealth() {
    return { status: 'OK' };
  }
```

Uxio0: I like your suggestion @hectorgomezv , and that shouldn't break anything as in liveness scripts we are only checking the `200` status code

hectorgomezv: Great, that's what I thought, thanks @Uxio0!

fmrsabino: Yeah for internal related routes we can go about these changes. I don't see any issue with the solution that you propose.

PR #189, opened 2022-11-10
Extend test timeout for development setup

hectorgomezv:
This PR extends the Jest timeout for `test:all` task. Some e2e tests need more than 5 seconds (cold start) to finish due to app initialisation and network delays. This configuration is already present for `test:e2e`, but it's also useful for running `test:all`, which is intended to use during the development process.

### DIFF
--- a/test/jest-all.json
+++ b/test/jest-all.json
@@ -5,6 +5,7 @@
   "setupFiles": ["<rootDir>/test/e2e-setup.ts"],
   "testEnvironment": "node",
   "testRegex": [".*\\.spec\\.ts$", ".*\\e2e-spec\\.ts$"],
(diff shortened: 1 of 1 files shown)


hectorgomezv:
> I agree, this is going to increase the testTimeout to 20 seconds per test? Do you think that this could happen a lot?

It shouldn't, but I'm seeing the first e2e test executed in 6-7 seconds in my local setup, mostly because network latencies and app startup times. This is only for development, this `test:all` task is not executed by the CI right now.

20 seconds it's a bit defensive, 10 seconds should be enough but I think it doesn't matter very much, test are run in parallel so it's not 20 seconds per test, all the tests should run in 20 seconds total if they fail.

PR #190, opened 2022-11-11
Add /about route

### LINKED ISSUE #151: Add GET /about route

hectorgomezv:
Closes #151 

This PR adds the `/about` endpoint to the service. The data is retrieved from `package.json` and Github Actions environment variables (as the Rust implementation does).


src/main.ts
```diff
@@ -17,7 +20,7 @@ async function bootstrap() {
 
   const config = new DocumentBuilder()
     .setTitle('Safe Client Gateway')
-    .setVersion(process.env.npm_package_version ?? '')
+    .setVersion(configService.get('about.version') ?? '')
```
hectorgomezv, line 23: I fixed this all well, as it is closely related to the purpose of the PR. Note this wasn't working since we abandoned `npm` as build tool in favor of `yarn`. More context [here](https://github.com/yarnpkg/berry/issues/1659) 

Another besides reading `package.json` directly could be to use [this library](https://github.com/Ayc0/yarn-plugin-envs), but I think it doesn't adds a lot of value for our purposes.

Dockerfile
```diff
@@ -16,6 +16,7 @@ RUN yarn run build
 #
 FROM node:16-alpine as production
 USER node
+COPY --chown=node:node --from=base /app/package.json ./package.json
```
hectorgomezv, line 19: This is necessary as we need to access `package.json` inside the container to gather the data. I tested it by building the container in my local setup.

hectorgomezv:
I'm drafting this PR since there's an ongoing discussion about the matter [here](https://www.notion.so/gnosis-safe/Config-Service-UI-Banner-feature-discussion-abfe48ccd366444298421313191524ea).

moisses89 (dismissed):
Looks good to me!

src/config/entities/configuration.ts
```diff
@@ -1,4 +1,15 @@
+import { readFileSync } from 'fs';
+
+const packageJson = JSON.parse(
+  readFileSync('package.json', { encoding: 'utf-8' }),
```
fmrsabino, line 4: I have some concerns regarding this approach. This basically makes the whole `package.json` available to the running instance – I think `package.json` should be seens a build/publishing tool and therefore we should avoid using it directly in the code (also ties the service to the build/publishing tool being used).

My suggestion is to find a more independent alternative to it. The version was previously retrieved with an environment variable – `npm_package_version` – which is only available while using `npm`. We can have a configuration script that sets the environment variable inside `pakage.json` and use that environment variable here.

fmrsabino (changes requested):

src/main.ts
fmrsabino: > I fixed this also, as it is closely related to the purpose of the PR. Note this wasn't working since we abandoned `npm` as build tool in favor of `yarn`. More context [here](https://github.com/yarnpkg/berry/issues/1659)

For me it seems to be working with `yarn` 🤔 

> Another option, besides reading `package.json` directly, could be to use [this library](https://github.com/Ayc0/yarn-plugin-envs), but I think it doesn't add a lot of value for our purposes.

I would not go with this approach. It's an external dependency (and we are trying to keep those to a minimum) and does more things that we currently require (this should not be complex).

src/config/entities/configuration.ts
hectorgomezv: I agree the optimal solution would be setting the data into a environment variable. 

Maybe we can output the version and the name during the building phase and pass that variable to the production image. I'll try to do it with `yarn` or even with `npm`, both should be available in the build container. What do you think @fmrsabino ?

fmrsabino: > I agree the optimal solution would be setting the data into a environment variable.
> 
> Maybe we can output the version and the name during the building phase and pass that variable to the production image. I'll try to do it with `yarn` or even with `npm`, both should be available in the build container. What do you think @fmrsabino ?

Yes I prefer this approach. Maybe a script that just sets these environment variables in the current session. Then the service wouldn't depend on the structure of the `package.json` but only on the local environment.

.github/workflows/ci.yml
```diff
@@ -49,12 +49,15 @@ jobs:
         with:
           node-version: '16.x'
           cache: 'yarn'
+      - run: |
+          echo "CGW_VERSION=`npm pkg get version`" >> $GITHUB_ENV
```
fmrsabino, line 53: This is not validated in our tests right? If not I'd consider removing it until we need it.

Dockerfile
fmrsabino, line 19: This is not required in the current implementation right?

.github/workflows/ci.yml
hectorgomezv: Yes, removed :) [9c26798](https://github.com/5afe/safe-client-gateway-nest/pull/190/commits/9c2679898364e7474e96937607672fd3b810cacb)

Dockerfile
hectorgomezv: Changed in [9c26798](https://github.com/5afe/safe-client-gateway-nest/pull/190/commits/9c2679898364e7474e96937607672fd3b810cacb)

src/config/entities/configuration.ts
hectorgomezv: >An environment variable set on the build phase and also on Docker itself should be simpler.

Totally agree 👍 , but based on my findings I think I won't be easy, unless we maintain the version number outside `package.json` and set it as environment variable in the container.

> My suggestion is to focus only on the controller/route part for this.

You mean just returning an `null`/`` for the version in the controller for now? 

> Also I'd say that this doesn't need to be tackled right now and probably a separate issue for the versioning would be better.

Yes, my bad, I thought I had found a simpler solution but yes, let's tackle this in a separate PR when we reach a consensus about the approach.

PR #191, opened 2022-11-11
Add AddressInfoHelper

fmrsabino:
- The goal of AddressInfoHelper is to provide common mapping functionality between retrieving a contract address and mapping it to a `AddressInfo`
- `get` retrieves the address information for the contract OR `null` if the address is `0x0000000000000000000000000000000000000000` OR error if the information could not be retrieved
- `getOrdDefault` is similar to `get` but returns a default `AddressInfo` if the information could not be retrieved
- `getCollection` is similar to `getOrDefault` but works on with a collection of addresses


fmrsabino:
Missing the route entity which I didn't cherry-pick from my branch

fmrsabino:
> Missing the route entity which I didn't cherry-pick from my branch

Fixed

hectorgomezv (commented):
Great job! 🚀

PR #196, opened 2022-12-01
Add factory functions. Improve transfer schemas

fmrsabino:
- Sorts properties lexicographically
- Adds factory functions for `ERC20Transfer`, `ERC721Transfer`, `EthereumTransaction`, `ModuleTransaction`, `MultisigTransaction` and `NativeTokenTransfer`
- `ERC20Transfer.executionDate`, `ERC721Transfer.executionDate`, `ModuleTransaction.executionDate` are now represented by `Date`
- Removes `type` from `Transfer` – this was only used to know which schema to use. Therefore, we can specify the schema in the schemas alone
- Removes `EthereumTransactionType`, `ModuleTransactionType` and `MultisigTransactionType` – these types only contained an additional field which was only used for serialization therefore we can have them specified in the respective schemas only.


hectorgomezv (approved):
LGTM, great job! 🚀

PR #197, opened 2022-12-05
Add GET /safes/<safe-address>/multisig-transactions route

### LINKED ISSUE #159: Add GET /safes/{safe_address}/multisig/transactions route

- [x] `/v1/safes/{}/multisig-transactions/` https://github.com/5afe/safe-client-gateway-nest/issues/167
- [x] `/v1/safes/{}/` https://github.com/5afe/safe-client-gateway-nest/pull/116
- [x] `/v1/safe-apps/` https://github.com/5afe/safe-client-gateway-nest/issues/165

hectorgomezv:
Closes #159 

Add GET /safes/<safe-address>/multisig-transactions route

- [X] Caching support
- [X] Unit/controller tests
- [X] End-to-end tests
- [X] OpenApi specification


src/routes/transactions/transactions.service.ts
```diff
@@ -0,0 +1,59 @@
...
+    safeAddress: string,
+    executionDateGte?: string,
+    executionDateLte?: string,
+    to?: string,
+    value?: string,
+    nonce?: string,
+    executed?: boolean,
+    limit?: number,
+    offset?: number,
+  ): Promise<Partial<Page<MultisigTransaction>>> {
```
hectorgomezv, line 26: This uses `Partial<T>` to comply with the current implementation, which does not return the `count` attribute in this route.

src/routes/transactions/transactions.module.ts
```diff
@@ -0,0 +1,15 @@
+import { Module } from '@nestjs/common';
+import { AddressInfoHelper } from '../common/address-info/address-info.helper';
+import { MultisigTransactionMapper } from './mappers/multisig-transaction.mapper';
+import { TransactionsController } from './transactions.controller';
+import { TransactionsService } from './transactions.service';
+
+@Module({
+  controllers: [TransactionsController],
+  providers: [
+    AddressInfoHelper,
```
fmrsabino, line 10: Here the `AddressInfoModule` should be imported

src/routes/transactions/transactions.service.ts
```diff
@@ -0,0 +1,59 @@
...
+      value,
+      nonce,
+      executed,
+      limit,
+      offset,
+    );
+
+    const safeInfo = await this.safeRepository.getSafe(chainId, safeAddress);
+    const results = await Promise.all(
+      transactions.results.map(async (multiSignTransaction) => ({
```
fmrsabino, line 42: typo: `multiSignTransaction`

src/routes/transactions/transactions.service.ts
```diff
@@ -0,0 +1,59 @@
...
+          safeInfo,
+        ),
+        conflictType: 'None',
+      })),
+    );
+
+    return {
+      next: transactions.next, // TODO: map URL
+      previous: transactions.previous, // TODO: map URL
+      results,
```
fmrsabino, line 56: We have the tools in place to map this easily.

You can check the `RouteUrlDecorator`, `PaginationDataDecorator` and the functions under `pagination.data.ts`

src/domain/tokens/entities/schemas/token.schema.ts
```diff
@@ -5,11 +5,11 @@ export const tokenSchema: JSONSchemaType<Token> = {
   type: 'object',
   properties: {
     address: { type: 'string' },
-    decimals: { type: 'number' },
+    decimals: { type: 'number', nullable: true },
```
fmrsabino, line 8: Good catch. This only applies to `ERC721` but maybe we should split this schema into multiple ones and use a `discriminator` for the `type`.

Something that can be tackled separately. 👍

src/domain/tokens/__tests__/token.factory.ts
```diff
@@ -0,0 +1,20 @@
+import { faker } from '@faker-js/faker';
+import { Token, TokenType } from '../entities/token.entity';
+
+export default function (
+  address?: string,
+  decimals?: number,
```
fmrsabino, line 6: If `decimals` is null we should allow that here too.

src/routes/transactions/entities/multisig-transaction.entity.ts
```diff
@@ -0,0 +1,30 @@
...
+  confirmationsRequired: number;
+  confirmationsSubmitted: number;
+  missingSigners?: string[];
+}
+
+export class TransactionInfo {
+  type: string;
+}
+
+export class TransactionSummary {
```
fmrsabino, line 15: OOC: why don't we have Swagger decorators on some of the classes on this route?

src/routes/transactions/entities/transfer-transaction.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+  logoUri?: string;
+}
+
+export interface NativeCoinTransferInfo {
+  value: string;
+}
+
+export interface TransferTransaction extends TransactionInfo {
+  sender: unknown;
+  recipient: unknown;
```
fmrsabino, line 32: Why are these `unkown`? Shouldn't they be `AddressInfo`?

fmrsabino (changes requested):
Still need to review some changes. This is more for early feedback/clarifications.

src/routes/transactions/transactions.module.ts
hectorgomezv: Changed in [6d58c66](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/6d58c665e46105677aa5f746b9ac461b4e0c8ee3)

src/routes/transactions/transactions.service.ts
hectorgomezv: Changed in [6d58c66](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/6d58c665e46105677aa5f746b9ac461b4e0c8ee3)

src/routes/transactions/transactions.service.ts
hectorgomezv: Changed in [6d58c66](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/6d58c665e46105677aa5f746b9ac461b4e0c8ee3)

src/domain/tokens/__tests__/token.factory.ts
hectorgomezv: Changed in [6d58c66](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/6d58c665e46105677aa5f746b9ac461b4e0c8ee3)

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+        TestCacheModule,
+        TestConfigurationModule,
+        TestNetworkModule,
+      ],
+    }).compile();
+
+    app = moduleFixture.createNestApplication();
+    app.useGlobalFilters(new DataSourceErrorFilter());
+
+    await app.init();
```
usame-algan, line 53: Do we need to init the app before running each test or could it work with only doing it once? When I move this part into `beforeAll`, some tests randomly fail with a 503 error. In the [nestjs documentation](https://docs.nestjs.com/fundamentals/testing) they are also calling `init` in `beforeAll`.

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+    it('Failure: data validation fails', async () => {
+      const chainId = faker.random.numeric();
+      const safeAddress = faker.finance.ethereumAddress();
+      const chainResponse = chainFactory(chainId);
+      mockNetworkService.get.mockResolvedValueOnce({ data: chainResponse });
+      mockNetworkService.get.mockResolvedValueOnce({
+        data: { results: ['invalidData'] },
+      });
+
+      await request(app.getHttpServer())
```
usame-algan, line 126: Do we need to await here if there is nothing after this call?

src/routes/transactions/transactions.controller.spec.ts
hectorgomezv: Yes, good catch, I also tried to init the app once in the `beforeAll` hook, and noted that. I'm not sure why this is happening, maybe it's something regarding the app configuration. I didn't spend so much time on it honestly but could be interesting to tackle it in a separate issue since it's not specific to this PR. Thanks!

src/routes/transactions/transactions.controller.spec.ts
hectorgomezv: I tried it in [9590b4e](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/9590b4edc09b406d88e253fcdf98e6725d60da84) and didn't work. I think not awaiting the Promise to resolve is letting Jest execute the `afterAll` hook and therefore closing down the app, I think that's the reason I was getting a 503 when removing the `await` instruction 🙂

src/routes/transactions/entities/multisig-transaction.entity.ts
```diff
@@ -0,0 +1,41 @@
...
+  @ApiProperty()
+  type: string;
+  @ApiProperty()
+  nonce: number;
+  @ApiProperty()
+  confirmationsRequired: number;
+  @ApiProperty()
+  confirmationsSubmitted: number;
+  @ApiPropertyOptional()
+  missingSigners?: string[];
```
moisses89, line 13: `missingSigners` are not really an `array` of strings, https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/transactions/models/summary.rs#L155

src/routes/transactions/entities/multisig-transaction.entity.ts
```diff
@@ -0,0 +1,41 @@
...
+  @ApiProperty()
+  id: string;
+  @ApiPropertyOptional()
+  timestamp?: number;
+  @ApiProperty()
+  txStatus: string;
+  @ApiProperty()
+  txInfo: TransactionInfo;
+  @ApiProperty()
+  executionInfo: ExecutionInfo;
```
moisses89, line 31: I think that this is optional and there is other one that is `SafeAppInfo`.

moisses89 (changes requested):
Good work!
I think that is a very hard schema with a lot of entities related, then is normal make mistakes :)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+
+  private async getToEtherTransfer(
+    chainId: string,
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): Promise<TransferTransaction> {
+    const recipient =
+      transaction.safe !== transaction.to
+        ? await this.addressInfoHelper.getOrDefault(chainId, transaction.to)
+        : { value: transaction.to };
```
fmrsabino, line 508: Why do we have this fallback?

fmrsabino (changes requested):
Some questions. Focused more on the mapper functionality. Still reviewing other parts.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -293,21 +293,31 @@ export class TransactionApi implements ITransactionApi {
     ordering?: string,
     executed?: boolean,
     trusted?: boolean,
+    executionDateGte?: string,
+    executionDateLte?: string,
+    to?: string,
+    value?: string,
+    nonce?: string,
```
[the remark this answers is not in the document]
hectorgomezv: Hmm, not sure what you mean, do you mean we should convert them to `number` internally? The input type is `string` (see https://safe-client.safe.global/index.html) and those values are sent using a query param to the transaction service. In fact, I just tried to use them for filtering and it works as expected 🤔

src/routes/transactions/__tests__/get-multisig-transactions.e2e-spec.ts
```diff
@@ -0,0 +1,123 @@
...
+    );
+
+    await request(app.getHttpServer())
+      .get(
+        `/chains/${chainId}/safes/${safeAddress}/multisig-transactions?execution_date__gte=${executionDateGte}&execution_date__lte=${executionDateLte}`,
+      )
+      .expect(200)
+      .then(({ body }) => {
+        expect(sortTransactionsByTimestamp(body)).toEqual(
+          sortTransactionsByTimestamp(expectedResponse),
```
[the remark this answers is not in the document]
hectorgomezv: Yes, it should, I experienced divergences because of different `modified` timestamps between Goerli staging and production services, but it shouldn't be necessary anymore 👍 I'll remove the sorting.

src/domain/interfaces/transaction-api.interface.ts
```diff
@@ -100,6 +100,11 @@ export interface ITransactionApi {
     ordering?: string,
     executed?: boolean,
     trusted?: boolean,
+    executionDateGte?: string,
+    executionDateLte?: string,
+    to?: string,
+    value?: string,
+    nonce?: string,
     limit?: number,
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [85720bd](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/85720bdf33a94eaab2d886ffe53d1e0f35df0059)

src/routes/transactions/__tests__/get-multisig-transactions.e2e-spec.ts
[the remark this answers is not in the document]
hectorgomezv: Changed in [85720bd](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/85720bdf33a94eaab2d886ffe53d1e0f35df0059)

src/routes/transactions/entities/custom-transaction.entity.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class CustomTransactionInfo extends TransactionInfo {
+  @ApiProperty()
+  to: AddressInfo | null;
+  @ApiProperty()
+  dataSize: string;
+  @ApiProperty()
+  value: string;
+  @ApiProperty()
+  methodName: string;
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [85720bd](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/85720bdf33a94eaab2d886ffe53d1e0f35df0059)

src/routes/transactions/entities/multisig-transaction.entity.ts
hectorgomezv: Changed in [85720bd](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/85720bdf33a94eaab2d886ffe53d1e0f35df0059)

src/routes/transactions/entities/multisig-transaction.entity.ts
hectorgomezv: True, I forgot to include it 😟 maybe because `SafeAppInfo` wasn't included in any of the transactions I used to test the functionality. But I think this deserves special attention because it seems there's some retrieval logic (see [safe_app_info.rs](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/transactions/converters/safe_app_info.rs)) and I'm not sure we need to change something in the domain because `origin` seems to be a JSON field. 

I included the entities in [ea6fec6](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/ea6fec62a715b307096688b11b1a4da1b39de46f) and I think it's better to let this as a TODO and tackle it in a separate PR, wdyt? (Any help from @fmrsabino is really appreciated since he probably has more context on this)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+
+  private getConfirmationsCount(transaction: MultisigTransaction): number {
+    return transaction.confirmations?.length || 0;
+  }
+
+  private getConfirmationsRequired(
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): number {
+    return transaction.confirmationsRequired ?? safe.threshold;
```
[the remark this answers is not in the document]
hectorgomezv: Good question. Maybe when the transaction was created the number of required signers is recorded in the `confirmationsRequired` field, and It could be different to `safe.threshold` because the Safe configuration could change after the transaction was created. Is that possible?

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+            ),
+          };
+        }
+        return <DeleteGuard>{
+          type: 'DELETE_GUARD',
+        };
+    }
+  }
+
+  private mapTxStatus(transaction: MultisigTransaction, safe: Safe): string {
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [20294e3](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/20294e3902127f7a18dd407d5d72ad9f5cd97e86)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+
+  private readonly SET_FALLBACK_HANDLER = 'setFallbackHandler';
+  private readonly ADD_OWNER_WITH_THRESHOLD = 'addOwnerWithThreshold';
+  private readonly REMOVE_OWNER = 'removeOwner';
+  private readonly SWAP_OWNER = 'swapOwner';
+  private readonly CHANGE_THRESHOLD = 'changeThreshold';
+  private readonly CHANGE_MASTER_COPY = 'changeMasterCopy';
+  private readonly ENABLE_MODULE = 'enableModule';
+  private readonly DISABLE_MODULE = 'disableModule';
+  private readonly SET_GUARD = 'setGuard';
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [20294e3](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/20294e3902127f7a18dd407d5d72ad9f5cd97e86)

src/domain/tokens/__tests__/token.factory.ts
```diff
@@ -0,0 +1,20 @@
...
+  symbol?: string,
+  type?: TokenType,
+): Token {
+  return <Token>{
+    address: address ?? faker.finance.ethereumAddress(),
+    decimals: decimals === undefined ? faker.datatype.number() : decimals,
+    logoUri: logoUri ?? faker.internet.url(),
+    name: name ?? faker.random.word(),
+    symbol: symbol ?? faker.finance.currencySymbol(),
+    type: type ?? TokenType.Erc20,
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [20294e3](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/20294e3902127f7a18dd407d5d72ad9f5cd97e86)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+    value: number,
+    dataSize: number,
+    chainId: string,
+  ): Promise<CustomTransactionInfo> {
+    const toAddressInfo = await this.addressInfoHelper.getOrDefault(
+      chainId,
+      transaction.to,
+    );
+    return {
+      type: 'Custom',
```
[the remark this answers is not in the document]
hectorgomezv: Do you mean extracting the multisig transaction types to an enum? Or using a constructor/builder with a fixed property value for `type`? Could you elaborate, please? :)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+  }
+
+  private async mapTransactionInfo(
+    chainId: string,
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): Promise<TransactionInfo> {
+    const value = Number(transaction?.value) || 0;
+    const dataSize = transaction.data
+      ? (Buffer.byteLength(transaction.data) - 2) / 2
```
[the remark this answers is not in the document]
hectorgomezv: Do you mean if `Buffer.byteLength(transaction.data)` is `0`? 

`transaction.data` is a string

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+      this.checkSenderOrReceiver(transaction)
+    );
+  }
+
+  private async mapTransactionInfo(
+    chainId: string,
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): Promise<TransactionInfo> {
+    const value = Number(transaction?.value) || 0;
```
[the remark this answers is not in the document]
hectorgomezv: The transaction service returns a string representing the number. Do you mean changing it on the Transaction Service side, or maybe coercing it before reaching this point?

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+    return this.mapCustomTransaction(transaction, value, dataSize, chainId); // TODO: default case
+  }
+
+  private filterAddressInfo(addressInfo: AddressInfo): AddressInfo {
+    return {
+      ...{ value: addressInfo.value },
+      ...(addressInfo.name !== '' && { name: addressInfo.name }),
+      ...(addressInfo.logoUri && { logoUri: addressInfo.logoUri }),
+    };
+  }
```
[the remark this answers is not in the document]
hectorgomezv: This is intended to remove `null` values from the returned `AddressInfo`. I mean, the Rust implementation removes properties holding a `null` value so at the end if `name` or `logoUri` are `null`, it returns an object just holding the `value` property. This is intended to mimic that behavior in JS, maybe there's a better/clearer way to do it. I'm open to suggestions 🙂

src/routes/transactions/entities/multisig-transaction.entity.ts
moisses89: I would like to include it in this PR because is part of the `multisigtransaction` endpoint, if we forget include it we could break things on the client side. 
I was searching an example but while the tx-service returns the `origin`, the client-gateway is not including the `safe_app_info`.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
hectorgomezv: I just rewrote it from the Rust implementation https://github.com/safe-global/safe-client-gateway/blob/8855fec20cb1ac2cb9783802dad35ba1071578bc/src/common/converters/mod.rs#L25

Maybe I'm missing something and the fallback is not necessary, do you think we could just use `transaction.to` value directly here?

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+      case this.SAFE_TRANSFER_FROM_METHOD:
+        return typeof dataDecoded.parameters[1]?.value === 'string'
+          ? dataDecoded.parameters[1]?.value
+          : fallback;
+      default:
+        return fallback;
+    }
+  }
+
+  private checkSenderOrReceiver(transaction: MultisigTransaction): boolean {
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [c42b4c3](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/c42b4c3a1e16fe4d11c566fcd7d75d553c157e34)

src/routes/transactions/entities/multisig-transaction.entity.ts
hectorgomezv: Ok, I was suggesting it to avoid doing too many things in this PR, it is complex enough I think. Closing this PR doesn't mean this feature/route is ready for production if we create an additional issue to tackle it. But I'm fine doing it here if you prefer 🙂 I'd like to have some inputs from @fmrsabino regarding this if possible before implementing the `origin` mapping.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
fmrsabino: With caching in place for both endpoints it's hard to say. I'd say to leave a `TODO` here so we can come back to this later.

But It'd be better to rely only on one source.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
fmrsabino: You return `CustomTransactionInfo`. Does `CustomTransactionInfo` have `Custom` always as the value of `type`? If yes we don't need to parameterise it no?

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
fmrsabino: Can't this be solved on the entity side with Swagger? I think this is quite verbose and on the CGW it's solved on the entity side – if we do it like this then we need to always apply this logic everywhere else.

Better yet, I'd allow `null`. There should be no distinction in the spec between `null` values and non-present values

src/routes/transactions/entities/multisig-transaction.entity.ts
hectorgomezv: Done 🙂 https://github.com/5afe/safe-client-gateway-nest/issues/201

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: In that case `transaction.data` would be a falsy value so the ternary operation would return `0`. Maybe there's a corner case if `transaction.data` is `1`, because `dataSize` becomes negative in that case, but reviewing the code it seems it won't cause an error either. I don't have all the context here, I'm just trying to mimic the Rust implementation, do you have any concern about it?

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: That's right 👍🏻 But `CustomTransactionInfo` is a JS class, if we want to have a fixed value for one of its properties we'd need a constructor. Maybe this could be a good candidate to use a builder pattern, wdyt?

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: I do agree, I think allowing `null` shouldn't break anything. I'll do it that way 👍🏻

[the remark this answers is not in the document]
hectorgomezv: Changed in [4827c02](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/4827c02120e9672707846c48d0552b30dd1952cb)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: Added a TODO in [4827c02](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/4827c02120e9672707846c48d0552b30dd1952cb)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
fmrsabino: Ok I see what you mean with `transaction.data` being `0`. In my opinion I think it's harder to read if we treat `""` as being the same as `undefined`.

Probably something like the following what do you think?

```typescript
const dataByteLength = transaction.data
  ? Buffer.byteLength(transaction.data)
  : 0;

const dataSize = dataByteLength >= 2 ? (dataByteLength - 2) / 2 : 0;
```

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
fmrsabino: Yes I think a constructor would be a good option here:

```typescript
export class CustomTransactionInfo extends TransactionInfo {
  @ApiProperty()
  type = 'Custom';

  ...

  constructor(
    to: AddressInfo | null,
    dataSize: string,
    value: string,
    methodName: string,
    actionCount: number | null,
    isCancellation: boolean,
  ) {
    super();
    this.to = to;
    this.dataSize = dataSize;
    this.value = value;
    this.methodName = methodName;
    this.actionCount = actionCount;
    this.isCancellation = isCancellation;
  }
}
```

Also I think that instead of having `?` in the route properties we should have `T | null`. At this point we should know if the value is present or not and I think it's more semantically correct. Wdyt?

src/routes/transactions/mappers/multisig-transaction.mapper.ts
fmrsabino: I think this is related with some other comments that I made: https://github.com/5afe/safe-client-gateway-nest/pull/197#discussion_r1045949407 and https://github.com/5afe/safe-client-gateway-nest/pull/197#discussion_r1045599136

We should just return here whatever the `addressInfoHelper` returns. Calling `addressInfoHelper.getOrDefault` and then having another default seems a bit strange 🤔

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,599 @@
...
+  private async mapSettingsInfo(
+    chainId: string,
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): Promise<SettingsInfo | undefined> {
+    const { dataDecoded } = transaction;
+    switch (dataDecoded.method) {
+      case MultisigTransactionMapper.SET_FALLBACK_HANDLER:
+        return <SetFallbackHandler>{
+          type: 'SET_FALLBACK_HANDLER',
```
fmrsabino, line 114: I think we should set the `type` as a property in the class itself. Allowing `SetFallbackHandler` + separately setting the `label` would involve more work.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,599 @@
...
+      executionInfo: this.mapExecutionInfo(transaction, safe, txStatus),
+      // TODO: include safeAppInfo retrieval logic where needed
+    };
+  }
+
+  private async mapSettingsInfo(
+    chainId: string,
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): Promise<SettingsInfo | undefined> {
```
fmrsabino, line 109: This function doesn't have a lot of tests to check if the decoding works as expected 😬

fmrsabino (changes requested):
I think we are lacking some tests and I've also suggested how we can rely less on setting the types on the route entities.

Let me know if you want to tackle this in a different way.

My suggestion is to Unit test the mapper. There are a lot of conditions that can be cumbersome to write. So I'd isolate the mapper and test the different conditions there.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
fmrsabino: If the transactions service returns it as a `String` but we convert it to `value` we should probably review if we should coerce it.

For now I think this is ok given that we are only directly using it here.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
hectorgomezv: I was trying to mimic the exact response from the Rust CGW implementation, even for `nulls`, that's the reason for that defaults, but I agree maybe it doesn't make much sense. I removed all the fallbacks to rely on `AddressInfoHelper`. It's changed in [a70656a](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/a70656aa38e4762de4339cba81c6ec7ff0d4fd13)

hectorgomezv:
> I think we are lacking some tests and I've also suggested how we can rely less on setting the types on the route entities.
> 
> Let me know if you want to tackle this in a different way.
> 
> My suggestion is to Unit test the mapper. There are a lot of conditions that can be cumbersome to write. So I'd isolate the mapper and test the different conditions there.

Yes, I agree, I think the best way to test all the conditions is to unit-test the mapper. I'll do it 👍🏻

I'm converting the PR to draft in the meanwhile. I'll be addressing your comments as well.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+      nonce: transaction.nonce,
+      confirmationsRequired: this.getConfirmationsRequired(transaction, safe),
+      confirmationsSubmitted: this.getConfirmationsCount(transaction),
+    };
+
+    if (txStatus === 'AWAITING_CONFIRMATIONS') {
+      return {
+        ...executionInfo,
+        missingSigners: this.getMissingSigners(transaction, safe, txStatus),
+      };
```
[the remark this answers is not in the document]
fmrsabino: What I mean is that every time you need to return an `ExecutionInfo` you need to use the spread operator for it to be "correct" right?

I think it's preferable to return the actual instance of `ExecutionInfo` than an implementation (which can be extended of `ExecutionInfo`).

If in order to instantiate a `ExecutionInfo` we need to think about what to include or not include then I think a constructor solution is more simple and also more elegant (easier to read compared to the spread operator).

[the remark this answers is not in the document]
fmrsabino: Example:

Given the following route entities:

```typescript
export abstract class ExecutionInfo {
  @ApiProperty()
  type: string;

  protected constructor(type: string) {
    this.type = type;
  }
}

export class MultisigExecutionInfo extends ExecutionInfo {
  @ApiProperty()
  nonce: number;
  @ApiProperty()
  @ApiProperty()
  confirmationsSubmitted: number;
  @ApiPropertyOptional()
  missingSigners: AddressInfo[] | null;

  constructor(
    nonce: number,
    confirmationsRequired: number,
    confirmationsSubmitted: number,
    missingSigners: AddressInfo[] | null,
  ) {
    super('MULTISIG');
    this.nonce = nonce;
    this.confirmationsRequired = confirmationsRequired;
    this.confirmationsSubmitted = confirmationsSubmitted;
    this.missingSigners = missingSigners;
  }
}
```

You can do the following

```typescript
private mapExecutionInfo(
    transaction: MultisigTransaction,
    safe: Safe,
    txStatus: string,
  ): ExecutionInfo {
    const missingSigners =
      txStatus === 'AWAITING_CONFIRMATIONS'
        ? this.getMissingSigners(transaction, safe)
        : null;

    return new MultisigExecutionInfo(
      transaction.nonce,
      this.getConfirmationsRequired(transaction, safe),
      this.getConfirmationsCount(transaction),
      missingSigners,
    );
  }
```

This accomplishes the following:
- You no longer need to specify a `type`. This could result in issues if other strings are applied but we want to enforce a specific type and not any `string`.
- By no longer relying on the spread operator we moved the logic on how to build this entity to the entity itself – which in my opinion avoids mistakes elsewhere.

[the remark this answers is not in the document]
fmrsabino: Alternatively you could set the type to `type: "MULTISIG"`.

However when Module transactions are introduced you'd need to have `type: "MULTISIG" | "MODULE"`. But the schema for both is quite different (module transactions only have one property: `address`.

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,599 @@
...
+    safe: Safe,
+  ): Promise<TransactionSummary> {
+    const txStatus = this.mapTransactionStatus(transaction, safe);
+    const txInfo = await this.mapTransactionInfo(chainId, transaction, safe);
+
+    return {
+      id: `multisig_${transaction.safe}_${transaction.safeTxHash}`,
+      timestamp:
+        transaction?.executionDate?.getTime() ??
+        transaction?.submissionDate?.getTime(),
```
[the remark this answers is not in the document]
hectorgomezv: Yes, `executionDate` should be always available according to the Transaction Service specification. Maybe the fallback was a legacy thing? I've changed it in [1f6827a](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/1f6827af4f3e7f07a085d74cc91686ad41a0385b)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: Yes, good catch. I changed it in [912132b](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/912132beed905de946a60c305496cd05313a51ba), thanks!

src/routes/transactions/mappers/multisig-transaction.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: I agree, changed in [de4181b](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/de4181bab770581f3e869e7eadce9fd43d16e5a9)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,592 @@
...
+    return safe.owners.filter((owner) => !confirmedOwners.includes(owner));
+  }
+
+  private mapExecutionInfo(
+    transaction: MultisigTransaction,
+    safe: Safe,
+    txStatus: string,
+  ): ExecutionInfo {
+    const executionInfo = {
+      type: 'MULTISIG',
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [32a999f](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/32a999fbc2d03262535493692ec03f34f8ff83ff) 👍🏻

src/routes/transactions/mappers/multisig-transaction.mapper.ts
hectorgomezv: Changed in [c464a50](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/c464a504861327618faa18ddc5c0662aace9d5a0)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
hectorgomezv: Tests added in [c9783bc](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/c9783bcef32930fa60fd6c12e8ef339fbdc46b07)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,599 @@
...
+        return <DeleteGuard>{
+          type: 'DELETE_GUARD',
+        };
+    }
+  }
+
+  private mapTransactionStatus(
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): TransactionStatus {
```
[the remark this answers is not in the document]
hectorgomezv: Tests added in [c9783bc](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/c9783bcef32930fa60fd6c12e8ef339fbdc46b07)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,599 @@
...
+    txStatus: string,
+  ): ExecutionInfo {
+    const executionInfo = {
+      type: 'MULTISIG',
+      nonce: transaction.nonce,
+      confirmationsRequired: this.getConfirmationsRequired(transaction, safe),
+      confirmationsSubmitted: this.getConfirmationsCount(transaction),
+    };
+
+    if (txStatus === 'AWAITING_CONFIRMATIONS') {
```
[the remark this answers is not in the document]
hectorgomezv: Tests added in [c9783bc](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/c9783bcef32930fa60fd6c12e8ef339fbdc46b07)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
hectorgomezv: Explicit constructors were introduced in [c464a50](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/c464a504861327618faa18ddc5c0662aace9d5a0)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,599 @@
...
+      this.getConfirmationsCount(transaction) <
+      this.getConfirmationsRequired(transaction, safe)
+    ) {
+      return TransactionStatus.AwaitingConfirmations;
+    }
+    return TransactionStatus.AwaitingExecution;
+  }
+
+  private getConfirmationsCount(transaction: MultisigTransaction): number {
+    return transaction.confirmations?.length || 0;
```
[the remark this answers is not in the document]
hectorgomezv: Covered in [9b9d1d7](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/9b9d1d7ef7f597aa9f44d919320a43efde273b59)

src/routes/transactions/mappers/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,599 @@
...
+  }
+
+  private getConfirmationsRequired(
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): number {
+    return transaction.confirmationsRequired ?? safe.threshold; // TODO: review if this fallback still required
+  }
+
+  private getMissingSigners(
```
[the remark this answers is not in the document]
hectorgomezv: Covered in [9b9d1d7](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/9b9d1d7ef7f597aa9f44d919320a43efde273b59) also

src/routes/transactions/entities/transaction.entity.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export class Transaction {
+  @ApiProperty()
+  id: string;
+  @ApiProperty()
+  timestamp: number;
+  @ApiProperty()
+  txStatus: string;
+  @ApiProperty()
+  txInfo: TransactionInfo;
```
fmrsabino, line 14: `@nestjs/swagger` requires a little bit more info here to generate the possible schemas for `TransactionInfo`

You can do something like:

```typescript
@ApiProperty({
  oneOf: [
    { $ref: getSchemaPath(CustomTransactionInfo) },
    { $ref: getSchemaPath(SettingsChangeTransaction) },
    { $ref: getSchemaPath(TransferTransactionInfo) },
  ],
})
txInfo: TransactionInfo;
```

The Transaction class however needs to include these models too:

```typescript
@ApiExtraModels(
  CustomTransactionInfo,
  SettingsChangeTransaction,
  TransferTransactionInfo,
)
export class Transaction {
  ...
}
```

This way the correct schemas will be available via the Swagger interface. The same can be done for `ExecutionInfo` in this class.

If you prefer to tackle this in a separate PR it's also ok 👍

src/routes/transactions/mappers/multisig-transactions/transaction-info/custom-transaction.mapper.ts
```diff
@@ -0,0 +1,75 @@
...
+      value.toString(),
+      transaction?.dataDecoded?.method ?? null,
+      this.getActionCount(transaction),
+      this.isCancellation(transaction, dataSize),
+    );
+  }
+
+  private getActionCount(transaction: MultisigTransaction): number | null {
+    const { dataDecoded } = transaction;
+    if (dataDecoded?.method === 'multiSend') {
```
fmrsabino, line 36: nitpick: extract `'multiSend'` (static)

src/routes/transactions/mappers/multisig-transactions/transaction-info/custom-transaction.mapper.ts
```diff
@@ -0,0 +1,75 @@
...
+      const parameter = dataDecoded.parameters?.find(
+        (parameter) => parameter.name === 'transactions',
```
fmrsabino, line 38: nitpick: extract `'transactions'` (static)

src/routes/transactions/mappers/multisig-transactions/transaction-info/custom-transaction.mapper.ts
```diff
@@ -0,0 +1,75 @@
...
+      to === safe &&
+      dataSize === 0 &&
+      (!value || Number(value) === 0) &&
+      operation === 0 &&
+      (!baseGas || Number(baseGas) === 0) &&
+      (!gasPrice || Number(gasPrice) === 0) &&
+      (!gasToken || gasToken === CustomTransactionMapper.NULL_ADDRESS) &&
+      (!refundReceiver ||
+        refundReceiver === CustomTransactionMapper.NULL_ADDRESS) &&
+      (!safeTxGas || safeTxGas === 60)
```
fmrsabino, line 72: Is `60` a typo?

src/routes/transactions/mappers/multisig-transactions/transaction-info/erc721-transfer.mapper.ts
```diff
@@ -0,0 +1,62 @@
...
+import { AddressInfoHelper } from '../../../../common/address-info/address-info.helper';
+import { TransferTransactionInfo } from '../../../entities/transfer-transaction-info.entity';
+import { Erc721Transfer } from '../../../entities/transfers/erc721-transfer.entity';
+import { DataDecodedParamHelper } from './data-decoded-param.helper';
+import { TransferDirectionHelper } from './transfer-direction.helper';
+
+@Injectable()
+export class Erc721TransferMapper {
+  private static readonly NULL_ADDRESS =
+    '0x0000000000000000000000000000000000000000';
```
fmrsabino, line 13: We should extract this. It's already being used in a couple places.

src/routes/transactions/mappers/multisig-transactions/multisig-transaction-execution-info.mapper.ts
```diff
@@ -0,0 +1,40 @@
...
+import { AddressInfo } from '../../../common/entities/address-info.entity';
+import { ExecutionInfo } from '../../entities/execution-info.entity';
+import { MultisigExecutionInfo } from '../../entities/multisig-execution-info.entity';
+
+@Injectable()
+export class MultisigTransactionExecutionInfoMapper {
+  mapExecutionInfo(
+    transaction: MultisigTransaction,
+    safe: Safe,
+    txStatus: string,
```
fmrsabino, line 13: Why not pass the mapped `txStatus`? Then we could compare against an enum and not against a `string`

src/routes/transactions/entities/multisig-execution-info.entity.ts
```diff
@@ -0,0 +1,27 @@
...
+import { ExecutionInfo } from './execution-info.entity';
+
+export class MultisigExecutionInfo extends ExecutionInfo {
+  @ApiProperty()
+  nonce: number;
+  @ApiProperty()
+  confirmationsRequired: number;
+  @ApiProperty()
+  confirmationsSubmitted: number;
+  @ApiPropertyOptional({ type: Array<AddressInfo>, nullable: true })
```
fmrsabino, line 12: I don't think the generic can be applied here (it's expecting an expression) 🤔

src/routes/transactions/mappers/multisig-transactions/multisig-transaction.mapper.ts
```diff
@@ -0,0 +1,43 @@
...
+    transaction: MultisigTransaction,
+    safe: Safe,
+  ): Promise<Transaction> {
+    const txStatus = this.statusMapper.mapTransactionStatus(transaction, safe);
+    const txInfo = await this.transactionInfoMapper.mapTransactionInfo(
+      chainId,
+      transaction,
+      safe,
+    );
+    const executionInfo = await this.executionInfoMapper.mapExecutionInfo(
```
fmrsabino, line 28: `await` shouldn't be required here.

fmrsabino (changes requested):
I think the new changes are a big improvement!

I have some additional comments but all of them should be a small thing.

src/routes/common/entities/address-info.entity.ts
```diff
@@ -3,14 +3,14 @@ import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
...
-  @ApiPropertyOptional()
-  readonly name?: string;
-  @ApiPropertyOptional()
-  readonly logoUri?: string;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  readonly name: string | null;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  readonly logoUri: string | null;
 
   constructor(value: string, name?: string, logoUri?: string) {
```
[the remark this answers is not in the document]
hectorgomezv: Yes, I thought the same, I think we can tackle that in a separate PR 👍

src/routes/transactions/entities/transaction.entity.ts
hectorgomezv: You're right, good catch! Included here [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306). Thanks for the suggestion!

src/routes/transactions/mappers/multisig-transactions/transaction-info/custom-transaction.mapper.ts
hectorgomezv: Changed in [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306)

src/routes/transactions/mappers/multisig-transactions/transaction-info/custom-transaction.mapper.ts
hectorgomezv: Changed in [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306)

src/routes/transactions/mappers/multisig-transactions/transaction-info/custom-transaction.mapper.ts
hectorgomezv: Oh yes, thanks!

Fixed in [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306)

src/routes/transactions/mappers/multisig-transactions/multisig-transaction-execution-info.mapper.ts
hectorgomezv: Agreed, changed in [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306)

src/routes/transactions/entities/multisig-execution-info.entity.ts
hectorgomezv: True, fixed in [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306)

src/routes/transactions/mappers/multisig-transactions/multisig-transaction.mapper.ts
hectorgomezv: Changed in [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306)

src/routes/transactions/mappers/multisig-transactions/transaction-info/erc721-transfer.mapper.ts
hectorgomezv: Included in [d985d6c](https://github.com/5afe/safe-client-gateway-nest/pull/197/commits/d985d6cb5043fb4c54881c2cdab35c7007f20306) 👍🏻

fmrsabino (dismissed):
🚀 🚀 🚀

hectorgomezv:
> 🚀 🚀 🚀

I just merged your ChainBuilder change in a test class, now it should be ready for approving again, sorry for the inconveniences 🙂

fmrsabino (approved):

moisses89 (approved):

PR #202, opened 2022-12-13
Use null instead of undefined for domain entities

fmrsabino:
- Replaces `undefined` properties in our domain models with `null`
- This was done for semantic reasons:
  * `undefined` should be seen as not knowing if the property is present (ie: the property is declared but absent)
  * `null` should be seen as the result that we know that the property does not have any value set
- Route entities that implemented the domain entities were also updated accordingly
  * However, `ApiPropertyOptional` does not work with union types with `null`
  * Therefore we need explicitly set in the options the respective `type` and `nullability`
  * See https://github.com/nestjs/swagger/issues/2129
- Use default: null for nullable values
  * There's no coercion between `undefined` values and `null` values – since we want to use `null` to specify that the value is absent, `default: null` is used effectively modifying the object being parsed if no value is present (see https://ajv.js.org/guide/modifying-data.html#assigning-defaults)


fmrsabino:
It looks like `null` is not assigned automatically.

With the following interface:
```typescript
interface X {
  a: string;
  b: string | null;
}
```

And JSON payload:
```
{
  "a": "foo"
}
```

the resulting struct would be:

```typescript
<X>{
  "a": "foo"
}
```

So in addition to not requiring the field to be present, my suggestion is to add `default: null` for those fields.

hectorgomezv:
LGTM, just a minor comment:

With this change, it seems we lost some schema details in the OpenApi specification. Using `collectibles` endpoint as example, we went from this:
Screenshot 2022-12-13 at 16 05 49
To this:
Screenshot 2022-12-13 at 16 06 58
Not sure if we could specify the expected types by using something like `@ApiPropertyOptional({ enum: ['string', 'null']})` into the route entities. It's a bit harder to maintain and I'm not sure the benefit pays off, just commenting it out.

hectorgomezv (dismissed):

fmrsabino:
> LGTM, just a minor comment:
> 
> With this change, it seems we lost some schema details in the OpenApi specification. 
> 
> Not sure if we could specify the expected types by using something like `@ApiPropertyOptional({ enum: ['string', 'null']})` into the route entities. It's a bit harder to maintain and I'm not sure the benefit pays off, just commenting it out.

Good catch. Seems to be a bug. See: https://github.com/nestjs/swagger/issues/2129

I will fix it with

```typescript
```

hectorgomezv (approved):

PR #206, opened 2022-12-21
Add GET /safes/incoming-transfers route

### LINKED ISSUE #157: Add GET /safes/{safe_address}/incoming-transfers route

- [x] `/v1/safes/{}/incoming-transfers/` https://github.com/5afe/safe-client-gateway-nest/issues/168
- [x] `/v1/tokens/` https://github.com/5afe/safe-client-gateway-nest/issues/166
- [x] `/v1/contracts/{}/` https://github.com/5afe/safe-client-gateway-nest/issues/100

hectorgomezv:
Closes #157 

Adds `GET /safes/incoming-transfers` route.

- [X] Caching support
- [X] Unit/controller tests
- [X] End-to-end tests
- [X] OpenApi specification


src/routes/transactions/transactions.controller.ts
```diff
@@ -50,4 +52,36 @@ export class TransactionsController {
       paginationData,
     );
   }
+
+  @ApiOkResponse({ type: IncomingTransferPage })
+  @Get('chains/:chainId/safes/:safeAddress/incoming-transfers')
```
hectorgomezv, line 57: We could wrap all the route params into a DTO here for the sake of brevity/clarity. Since this could also be done for other routes, probably it's better to tackle it in a separate issue, wdyt?

src/routes/transactions/transactions.service.ts
```diff
@@ -70,4 +73,54 @@ export class TransactionsService {
...
+      ),
+    );
+
+    const nextURL = cursorUrlFromLimitAndOffset(routeUrl, transfers.next);
+    const previousURL = cursorUrlFromLimitAndOffset(
+      routeUrl,
+      transfers.previous,
+    );
+
+    return {
```
hectorgomezv, line 120: For several routes we are omitting `next` and `previous` (when next/previous pages aren't needed and therefore their values are `undefined`). Probably for coherence we want to explicitly set them to `null`. Since this affects several routes also, I think we can create a separate issue as well, wdyt?

src/domain/safe/entities/transfer.entity.ts
```diff
@@ -1,4 +1,5 @@
 export interface Transfer {
+  type: string;
```
fmrsabino, line 2: Why was this added back? See https://github.com/5afe/safe-client-gateway-nest/pull/196

fmrsabino (changes requested):

src/domain/safe/entities/__tests__/erc20-transfer.factory.ts
```diff
@@ -11,6 +11,7 @@ export default function (
   value?: string,
 ): ERC20Transfer {
   return <ERC20Transfer>{
+    type: 'ERC20_TRANSFER',
```
fmrsabino, line 14: I believe this is only part of the list endpoint and not returned when a a single transfer is requested.

Even though the types match we should keep this in mind.

src/domain/tokens/entities/schemas/token.schema.ts
```diff
@@ -6,7 +6,7 @@ export const tokenSchema: JSONSchemaType<Token> = {
   properties: {
     address: { type: 'string' },
     decimals: { type: 'number', nullable: true },
-    logoUri: { type: 'string', format: 'uri' },
```
fmrsabino, line 9: Why was this removed? Can it not have the `uri` format?

We should document this (in the PR for example).

src/routes/transactions/transactions.service.ts
fmrsabino: Yes I agree. We should set them to `null` explicitly.

But we can also do it here too. `cursorUrlFromLimitAndOffset` already returns either a value or `null` so it's up to the route to map it `undefined` (which we shouldn't). No strong opinion on how you want to do this so I'll leave it up to you.

fmrsabino (changes requested):

src/routes/transactions/mappers/transfers/transfer-info.mapper.ts
```diff
@@ -0,0 +1,105 @@
...
+    );
+
+    switch (type) {
+      case TransferInfoMapper.ERC20_TRANSFER: {
+        const { tokenAddress, value } = transfer as ERC20Transfer;
+        const token = await this.getToken(chainId, tokenAddress);
+        return new TransferTransactionInfo(
+          sender,
+          recipient,
+          direction,
```
fmrsabino, line 52: Given that we always build a `TransferTransactionInfo` with the same `sender`, `recipient` and `direction` maybe we can return it in the end and use the switch/case just to build the different transfers wdyt?

src/routes/transactions/transactions.service.ts
hectorgomezv: Great, I'll fix this route pagination values here, and I've also created a separated PR to explicitly set these values to null when applicable into other routes 🙂 https://github.com/5afe/safe-client-gateway-nest/pull/208

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: Yes, I didn't notice that, probably was added back on a conflict resolution on my side. But, having that said, if we remove this field, we lose the transfer type coming from the Transaction Service, and how we could map the transfers without that information? I mean, each transfer type has specific rules/fields, and also the expected response includes a field with:
```javascript
"transferInfo": {
    "type": ...,
}
```

Maybe I'm missing something here 🤔

src/domain/safe/entities/__tests__/erc20-transfer.factory.ts
hectorgomezv: Oh, I see, so do you think removing this hardcoded value from the factory and passing the type as an argument would be enough?

src/domain/tokens/entities/schemas/token.schema.ts
hectorgomezv: I was getting validation errors because of `logoUri` fields not starting with `http:://`. I had an example but it's not available right now due to staging service interruption 🤔 Maybe we should assume the `logoUri` is always following the URI format?

src/domain/safe/entities/transfer.entity.ts
fmrsabino: The goal was that the `type` would only be used to specify to which object the payload would be serialised to. So AJV uses the `type` to check if the object should be checked against a `ERC20Transfer`, `ERC721Transfer` or `NativeTokenTransfer`.

Given that you then have this validation in place you wouldn't need a `type` on top of the different types available. We'd use something like type predicated to distinguish between the different types above: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

src/routes/transactions/transactions.service.ts
hectorgomezv: Great, I'll fix this route pagination values here, and I've also created a separated PR to explicitly set these values to null when applicable into other routes 🙂 https://github.com/5afe/safe-client-gateway-nest/pull/208

src/domain/safe/entities/__tests__/erc20-transfer.factory.ts
fmrsabino: Maybe this is something that we need to align on. I think these factories can be seen as "Domain" factories and maybe it's better to see them as "json payload" factories because that's how we have been using them.

So in this example `type: 'ERC20_TRANSFER'` would be part of the JSON payload but would not necessarily be part of the Domain object that results from that.

Not sure if this was super clear 😅

src/domain/tokens/entities/schemas/token.schema.ts
fmrsabino: That's a good question. We should check on this on the Transaction Service side. We can leave a comment either in the PR or in the codebase as to why it doesn't follow that format (depending on our findings).

src/domain/safe/entities/__tests__/erc20-transfer.factory.ts
hectorgomezv: Yes, I get your point, these factories are being used to reproduce the responses gathered from external services (transaction service, config service, currency conversion APIs, etc.), but until now we don't have any separation between our _domain objects_ and these responses. I mean, we were designing our _domain object_ shapes just to represent the payload coming from these external services, and I thought it was intentional and for the sake of simplicity.

I'm not against doing a separation between the _JSON payload_ coming from external services and our domain layer entities, but as you said that requires an alignment. 

In the current architecture, we are doing mappings/transformations between _domain objects_ and _route objects_ in the services classes, which belong to the routes layer. In the past, I commented that maybe that transformation/mapping could be done in the domain layer, and the mapping between domain objects and route objects could be trivial, just 1:1 mappings. That could resolve the problem of not having a 1:1 mapping between _JSON payloads_ and _domain objects_. (And also the -in the case of this application- minor design problem like having transformation logic coupled to the HTTP routes layer).

Another option to solve this _problem_ is to maintain the current _domain-to-route object_ transformations and add another transformation layer between JSON payloads and the domain. But honestly, I think this is a bit too much (maintaining two transformation layers) for this kind of application.

Of course this deserves its own discussion if you think it's better to change the design. IMHO we could just follow the current architecture at least until the service v1 is finally released 😅

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: Maybe I'm missing something, but your proposal is to delete the type information from the transaction service response, and then recalculate it using another mechanism? 🤔

src/domain/tokens/entities/schemas/token.schema.ts
hectorgomezv: Screenshot 2022-12-30 at 12 25 00

This is the response from the transaction service. We could investigate why the domain part missing in the `logoUri` URL. 

It may be related to the backup restoration these days, but in terms of the gateway it doesn't matter a lot I think, the fact it's it could not follow the URI format in some scenarios. IMHO we don't want to return a 500 from the gateway because the validation doesn't pass in those scenarios, what do you think? Do you think we need to error in the gateway response in that case?

src/routes/transactions/mappers/transfers/transfer-info.mapper.ts
hectorgomezv: I agree, I think it's more clear that way. I refactored the main function in the class to have just one exit point and also moved the `Transfer` mapping to a separate function `getTransferByType` 🙂 Let me know if you find it more readable this way.

hectorgomezv: Refactored in [2f6d0c5](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/2f6d0c508eace06c134039e921fe51865fc195eb)

src/routes/transactions/transactions.service.ts
hectorgomezv: Changed in [2f6d0c5](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/2f6d0c508eace06c134039e921fe51865fc195eb)

src/domain/safe/entities/__tests__/erc20-transfer.factory.ts
fmrsabino: > In the current architecture, we are doing mappings/transformations between domain objects and route objects in the services classes, which belong to the routes layer. In the past, I commented that maybe that transformation/mapping could be done in the domain layer, and the mapping between domain objects and route objects could be trivial, just 1:1 mappings.

The decision to do this was mostly because route entities are rarely shared between routes. So you have the domain which corresponds to the Transaction Service and Config Service domains and then each route maps that domain to a specific entity. Or do you see that the route entities are mostly shared and therefore it makes sense for them to be our domain instead of the current one?

> Of course this deserves its own discussion if you think it's better to change the design. IMHO we could just follow the current architecture at least until the service v1 is finally released

I'm not sure if I understand the design change that you mention. We have:

- Domain entities
- Factories for those domain entities

Factories are used in the tests to mock the return of the HTTP layer. However if AJV serializes data in a different way (eg.: date serialisation to `Date` instead of `String`) then the factory might not match 1:1 the domain entity and that's ok! So the factory would have the date as a `String` but the resulting domain entity would have it as `Date` because it was then parsed by AJV.

src/domain/safe/entities/transfer.entity.ts
fmrsabino: > Maybe I'm missing something, but your proposal is to delete the type information from the transaction service response, and then recalculate it using another mechanism? 🤔

We are not deleting the type information. The type is in the interface itself. AJV uses the type to check which interface the payload should "serialised" to.

If you have:

```typescript
interface X {
  type: "X"
}

interface Y {
  type: "Y"
}

interface Z {
  type: "Z"
}
```

What would be the value of having the `type` as a `String` in addition to it being encapsulated in an interface that represents that same type?

src/domain/tokens/entities/schemas/token.schema.ts
fmrsabino: The current CGW does not return an error if the `logoUri` is not a a `uri` so let's remove the `format` as you did 👍

src/routes/transactions/mappers/common/transfer-direction.helper.ts
```diff
@@ -3,13 +3,17 @@ import { TransferDirection } from '../../entities/transfer-transaction-info.enti
 
 @Injectable()
 export class TransferDirectionHelper {
```
fmrsabino, line 5: This class doesn't seem to provide a lot of value given that the only functionality that it has is represented by a function.

What do you think about just having the function (and no class)?

fmrsabino (changes requested):

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: I tried to remove the `type: string;` line from the Transfer interface but I still not getting how you can implement something like the `getTransferByType` function ([this one](https://github.com/5afe/safe-client-gateway-nest/pull/206/files?diff=split&w=0#diff-e71537626c00f304d6f5a6e18f9d4fe64a39cc739a29deeeb36b8bd7df2693e4R54)) without having the type information. Do you propose using type predicates to infer the type of the transfer? I'm not sure how we can use the interfaces directly here (they are a TS construct witch is not available at runtime as you know) so the only way I see is to use type predicates to rebuild a piece of information which is already in the object 🤔 

Could you provide a concrete example to build the `Transfer` in `getTransferByType`?

src/domain/safe/entities/__tests__/erc20-transfer.factory.ts
hectorgomezv: Ok, so if I understand correctly, and based on the discussion we had today, this factory should return a plain object instead of a domain object, but should also maintain `type: 'ERC20_TRANSFER'` in the payload, because that field is in the JSON payload coming from the transaction service 🙂

fmrsabino: > Ok, so if I understand correctly, and based on the discussion we had today, this factory should return a plain object instead of a domain object, but should also maintain `type: 'ERC20_TRANSFER'` in the payload, because that field is in the JSON payload coming from the transaction service 🙂

Yes exactly 👍 AJV would use that field to know which schema to use for validation!

src/routes/transactions/mappers/common/transfer-direction.helper.ts
hectorgomezv: Yes, good point. We can remove a lot of boilerplate code and injection logic that way, thanks!

Changed in [b53dc23](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/b53dc2327a87d90e6c7c67f3e0b8f7fa41ad7167)

src/routes/transactions/transactions.controller.ts
```diff
@@ -52,7 +54,6 @@ export class TransactionsController {
       paginationData,
     );
   }
-
```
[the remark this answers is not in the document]
hectorgomezv: Added in [b53dc23](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/b53dc2327a87d90e6c7c67f3e0b8f7fa41ad7167) 👍🏻 thanks

src/domain/safe/entities/transfer.entity.ts
fmrsabino: In `src/domain/safe/entities/transfer.entity.ts` you can add the following Type Predicates:

```typescript
export function isERC20Transfer(transfer: Transfer): transfer is ERC20Transfer {
  const erc20Transfer = transfer as ERC20Transfer;
  return (
    erc20Transfer.tokenAddress !== undefined &&
    erc20Transfer.value !== undefined
  );
}

export function isERC721Transfer(
  transfer: Transfer,
): transfer is ERC721Transfer {
  const erc721Transfer = transfer as ERC721Transfer;
  return (
    erc721Transfer.tokenAddress !== undefined &&
    erc721Transfer.tokenId !== undefined
  );
}

export function isNativeTokenTransfer(
  transfer: Transfer,
): transfer is NativeTokenTransfer {
  const nativeTokenTransfer = transfer as NativeTokenTransfer;
  return nativeTokenTransfer.value !== undefined;
}
```

And then `getTransferByType` can be written as follows:

```typescript
private async getTransferByType(
  chainId: string,
  domainTransfer: DomainTransfer,
): Promise<Transfer> {
  if (isERC20Transfer(domainTransfer)) {
    const { tokenAddress, value } = domainTransfer;
    const token = await this.getToken(chainId, tokenAddress);
    return new Erc20Transfer(
      token.address,
      value,
      token.name,
      token.symbol,
      token.logoUri,
      token.decimals,
    );
  } else if (isERC721Transfer(domainTransfer)) {
    const { tokenAddress, tokenId } = domainTransfer;
    const token = await this.getToken(chainId, tokenAddress);
    return new Erc721Transfer(
      token.address,
      tokenId,
      token.name,
      token.symbol,
      token.logoUri,
    );
  } else if (isNativeTokenTransfer(domainTransfer)) {
    return new NativeCoinTransfer(domainTransfer.value);
  } else {
    throw Error('Unknown transfer type');
  }
}
```

src/routes/transactions/transactions.controller.ts
fmrsabino: No strong opinion on this one. I think for now this works as it also reduces the coupling to the `common` package (which would contain those DTOs)

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: Thanks for the example code 🙌🏻

I thought about that approach as well, but I saw some cons regarding it:
- We would be implementing/testing/maintaining that type predicates functions.
- The service would be doing extra computations to calculate something (the transfer type) that in fact, it's already present in the data.
- What happens if one of the conditions which define a type changes? (if `tokenId` is not `undefined` for the ERC721 in the future, for any reason) This would force us to implement the change in the transaction service and also here. It could lead to bugs for a change that happens outside of this service _domain_.

The obvious pro of using type predicates is we would have a cleaner/more elegant domain entity definition. What do you think?

src/domain/safe/entities/__tests__/erc20-transfer.factory.ts
hectorgomezv: Changed [here]([39c06d7](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/39c06d7085e75619177e496b3664338ad4a550c3))

src/domain/safe/entities/transfer.entity.ts
fmrsabino: > Thanks for the example code 🙌🏻
> 
> I thought about that approach as well, but I saw some cons regarding it:
> 
> * We would be implementing/testing/maintaining that type predicates functions. (Using the provided `type` field comes for free.)

Correct. But I wouldn't say that `type` field coming from free is "good". We use the type to match with the schema but once we know the schema we shouldn't need the type. The `type` comes in the response to make it easier for the consumers of the API to know which schema to expect – once you know that, keeping the `type` is a code smell in my opinion.

When we design an interface that can have 3 interfaces implementing it (non-json related) we wouldn't add a type to each one or? 

> * The service would be doing extra computations to calculate something (the transfer type) that in fact, it's already present in the data.

This is assuming that checking the presence of some fields is computationally more expensive than doing a string comparison (to check the type) – I would have to check this.

> * What happens if one of the conditions which define a type changes? (if `tokenId` is not `undefined` for the ERC721 in the future, for any reason) This would force us to implement the change in the transaction service and also here. It could lead to bugs for a change that happens outside of this service _domain_.

This is contract breaking change. Every field in the domain should either have a value or be `null`. An `undefined` field means that it was removed from the payload. AJV would then have `null` in its place. If a nullable field becomes nullable we consider that it's an API breaking change.

If one of the conditions that define a type change wouldn't you also need to change AJV?

> The obvious pro of using type predicates is we would have a cleaner/more elegant domain entity definition. But I have some doubts about if it pays the price. What do you think?

I think you listed some good points but I'd say that for point 2) is not clear how it performs extra computations (even if it does are we optimising at this level) and for point 3) changing the API contract means that we need to check AJV. Also if the domain changes suddenly of course it can lead to bugs. But the domain is our contract with the backend.

hectorgomezv: Ok, I get your point and I agree the domain should be clean. I tried to use the approach you proposed based on type predicates but I'm getting some mismatches between transfer types, because for some fields we are checking they are not `undefined` but in fact the transaction service is returning `null`. For example, this is a payload coming from the transaction service:

```
{
    "count": 2,
    "next": null,
    "previous": null,
    "results": [
        {
            "type": "ETHER_TRANSFER",
            "executionDate": "2022-08-04T12:44:22Z",
            "blockNumber": 7345420,
            "transactionHash": "0x5009500e71ff0dc12ddce3cbe64fd346df9bce9f2413f4dd4c4e82877313112b",
            "to": "0x4127839cdf4F73d9fC9a2C2861d8d1799e9DF40C",
            "value": "513839314563540",
            "tokenId": null,
            "tokenAddress": null,
            "tokenInfo": null,
            "from": "0x283Af0B28c62C092C9727F1Ee09c02CA627EB7F5"
        },
        {
            "type": "ERC721_TRANSFER",
            "executionDate": "2022-08-04T12:44:22Z",
            "blockNumber": 7345420,
            "transactionHash": "0x5009500e71ff0dc12ddce3cbe64fd346df9bce9f2413f4dd4c4e82877313112b",
            "to": "0x4127839cdf4F73d9fC9a2C2861d8d1799e9DF40C",
            "value": null,
            "tokenId": "110380171670080793190909999014722746036502744273864122623133899039475607835745",
            "tokenAddress": "0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85",
            "tokenInfo": {
                "type": "ERC721",
                "address": "0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85",
                "name": "Ethereum Name Service",
                "symbol": "ENS",
                "decimals": null,
                "logoUri": "https://safe-transaction-assets.safe.global/tokens/logos/ENS.png"
            },
            "from": "0x283Af0B28c62C092C9727F1Ee09c02CA627EB7F5"
        }
    ]
}
```

So to solve this, we should use something like:

```javascript
export function isERC20Transfer(transfer: Transfer): transfer is ERC20Transfer {
  const erc20Transfer = transfer as ERC20Transfer;
  return (
    erc20Transfer.tokenAddress !== undefined &&
    erc20Transfer.tokenAddress !== null &&
    erc20Transfer.value !== undefined &&
    erc20Transfer.value !== null
  );
}

export function isERC721Transfer(
  transfer: Transfer,
): transfer is ERC721Transfer {
  const erc721Transfer = transfer as ERC721Transfer;
  return (
    erc721Transfer.tokenAddress !== undefined &&
    erc721Transfer.tokenAddress !== null &&
    erc721Transfer.tokenId !== undefined &&
    erc721Transfer.tokenId !== null
  );
}

export function isNativeTokenTransfer(
  transfer: Transfer,
): transfer is NativeTokenTransfer {
  const nativeTokenTransfer = transfer as NativeTokenTransfer;
  return (
    nativeTokenTransfer.value !== undefined &&
    nativeTokenTransfer.value !== null
  );
}
```

What do you think about using nulls here? Not sure if we are adding too much complexity here 🤔 

Apart from that, maybe you already notice, but these type predicates are not mutually exclusive. I mean, a transfer with non-null `tokenAddress` and non-null `value` satisfies both `isERC20Transfer` and `isNativeTokenTransfer`, so we have to keep this in mind when building the `if-else` statements in `getTransferByType` or in other places: the order of the conditions matters. We wouldn't have this complexity if we just rely on the `type` field coming from the source.

I don't have a strong opinion here, just checking we are considering all the implications of this 🙂

fmrsabino: As I stated before I believe that having a type in a domain entity is strange. You'd need to explicitly cast depending on the type and then access the respective properties.

Type predicates do also perform a cast (implicit cast instead of an explicit one) but they work with the structure of the object to determine the type and not with a string value that defines it in a JSON payload.

If you believe that having type predicates adds complexity in this case and you'd see that having a type as a string is better then I'd say to go for it. I'll leave this decision to you.

hectorgomezv: I do see value in using type predicates, having implicit casting, and having a clearer domain. I think the predicates are a great tool. I was just pointing out some concerns about their use in this particular use case. But I think we can go that way and see how it evolves, I took your suggested approach in [697dda9](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/697dda95b51318e373bcf339345ead59abb5e27d). Sorry for the long discussion, I just wanted to align on some cons it could have. Thanks for reviewing! 🙂

fmrsabino: I've addressed some of the concerns but it was my understanding that you see that it'd be more complex to maintain and write the type predicates. Therefore I'd approve whatever you decide and think it's easier to maintain.

If the having the type as a property is easier for maintaining and writing business logic around it then I'd say to go for it!

src/domain/safe/entities/transfer.entity.ts
```diff
@@ -12,11 +11,43 @@ export interface ERC20Transfer extends Transfer {
   value: string;
 }
 
+export function isERC20Transfer(transfer: Transfer): transfer is ERC20Transfer {
+  const erc20Transfer = transfer as ERC20Transfer;
+  return (
+    erc20Transfer.tokenAddress !== undefined &&
+    erc20Transfer.tokenAddress !== null &&
+    erc20Transfer.value !== undefined &&
+    erc20Transfer.value !== null
```
fmrsabino, line 20: Should be equivalent to

```typescript
erc20Transfer.value != null
```

src/domain/safe/entities/transfer.entity.ts
```diff
@@ -12,11 +11,43 @@ export interface ERC20Transfer extends Transfer {
...
   tokenId: string;
 }
 
+export function isERC721Transfer(
+  transfer: Transfer,
+): transfer is ERC721Transfer {
+  const erc721Transfer = transfer as ERC721Transfer;
+  return (
+    erc721Transfer.tokenAddress !== undefined &&
+    erc721Transfer.tokenAddress !== null &&
```
fmrsabino, line 35: Similar to the above ☝️ 

I think our model is incorrect. `tokenAddress` cannot be `null`.

After changing that you can write it like:

```typescript
erc721Transfer.tokenAddress != null
```

src/domain/safe/entities/transfer.entity.ts
```diff
@@ -12,11 +11,43 @@ export interface ERC20Transfer extends Transfer {
...
+    erc721Transfer.tokenId !== undefined &&
+    erc721Transfer.tokenId !== null
```
fmrsabino, line 37: Should be equivalent to 

```typescript
erc721Transfer.tokenId != null
```

src/domain/safe/entities/transfer.entity.ts
```diff
@@ -12,11 +11,43 @@ export interface ERC20Transfer extends Transfer {
...
 }
+
+export function isNativeTokenTransfer(
+  transfer: Transfer,
+): transfer is NativeTokenTransfer {
+  const nativeTokenTransfer = transfer as NativeTokenTransfer;
+  return (
+    nativeTokenTransfer.value !== undefined &&
+    nativeTokenTransfer.value !== null
+  );
```
fmrsabino, line 52: As you mentioned the following payload:

```typescript
{
  value: 123,
  tokenAddress: "0x00"
}
```

can be considered both a `NativeTokenTransfer` and a `ERC20Transfer`. My suggestion is to rewrite it as:

```typescript
export function isNativeTokenTransfer(
  transfer: Transfer,
): transfer is NativeTokenTransfer {
  const nativeTokenTransfer = transfer as NativeTokenTransfer;
  return (
    (transfer as any).tokenAddress == null && nativeTokenTransfer.value != null
  );
}
```

Therefore we only consider a transfer to be a `NativeTokenTransfer` IFF `tokenAddress` is `null` or `undefined` AND `value` is not `null` or `undefined`.

Wdyt?

fmrsabino (changes requested):

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: Yes, I think that's a more solid definition. Thanks, changed in [a41f860](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/a41f8600dc9fbe237de992a74c51dce6a6703668)

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: [a41f860](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/a41f8600dc9fbe237de992a74c51dce6a6703668)

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: [a41f860](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/a41f8600dc9fbe237de992a74c51dce6a6703668)

src/domain/safe/entities/transfer.entity.ts
hectorgomezv: [a41f860](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/a41f8600dc9fbe237de992a74c51dce6a6703668)

src/domain/safe/entities/transfer.entity.ts
```diff
@@ -12,11 +11,43 @@ export interface ERC20Transfer extends Transfer {
   value: string;
 }
 
+export function isERC20Transfer(transfer: Transfer): transfer is ERC20Transfer {
+  const erc20Transfer = transfer as ERC20Transfer;
+  return (
+    erc20Transfer.tokenAddress !== undefined &&
+    erc20Transfer.tokenAddress !== null &&
```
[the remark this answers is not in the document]
hectorgomezv: [a41f860](https://github.com/5afe/safe-client-gateway-nest/pull/206/commits/a41f8600dc9fbe237de992a74c51dce6a6703668)

fmrsabino (approved):
🚀 🚀 🚀

PR #207, opened 2022-12-23
Add ConflictType enum

hectorgomezv:
This PR adds a `ConflictType` enum, instead of relying on a simple string. This will be used in routes that need to be implemented, i.e. `QueuedTransactions` retrieval.

### DIFF
--- a/src/routes/transactions/entities/conflict-type.entity.ts
+++ b/src/routes/transactions/entities/conflict-type.entity.ts
@@ -1,0 +1,5 @@
+export enum ConflictType {
+  None = 'None',
+  HasNext = 'HasNext',
+  End = 'End',
+}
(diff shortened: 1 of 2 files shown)


moisses89 (commented):
This is tested in any place? I mean in the e2e test for multisigtransactions for example.

hectorgomezv:
> This is tested in any place? I mean in the e2e test for multisigtransactions for example.

Yes, every time a new `MultisigTransaction` is created using the constructor, the enum is used, so it's used in the E2E tests. More precisely, is used [here](https://github.com/5afe/safe-client-gateway-nest/blob/b9a5cacc5d4299e550802075236d9a3f72ad3c85/src/routes/transactions/transactions.service.ts#L49) 🙂

moisses89 (approved):

PR #208, opened 2022-12-23
Explicitly return null for empty pagination data

hectorgomezv:
(Related to https://github.com/5afe/safe-client-gateway-nest/pull/206#discussion_r1056314317)

This PR adds explicit null values when pagination data is empty.


moisses89 (approved):
Looks good to me!

PR #209, opened 2022-12-28
Fix end to end tests

hectorgomezv:
Due to the recent database restore, some transactions have different attributes (last modified timestamp mostly), so this causes the transactions ordering in the response from Transaction Service is different. This PR changes the expected response JSON to make end-to-end tests work again.


moisses89 (approved):
Looks good to me, but we should research the reason of why the order of response changed because a database restore.

hectorgomezv:
> Looks good to me, but we should research the reason of why the order of response changed because a database restore.

Yes, we can research the transaction service for these particular transactions. Still, the result seems correct because in the restored backup the `modified` dates are different from what we had In staging.  👍🏻

PR #212, opened 2023-01-02
Revert date change for ModuleTransaction

fmrsabino:
- Reverts the change from `Date` to `string` made in dbe756d7355c3bbd787934336f6f29a1afc5fc28
- `Date` was introduced on the domain level so that any business logic around dates can use the Date API directly
- See b25d3077686426eb0313cfb18bb800691da541a8 – which introduced support for `Date` on the domain level (AJV support)


moisses89 (approved):

hectorgomezv (approved):

fmrsabino:
> We should change this conversion https://github.com/5afe/safe-client-gateway-nest/blob/main/src/routes/transactions/mappers/module-transactions/module-transaction.mapper.ts#L32

Done in [4a972b3](https://github.com/5afe/safe-client-gateway-nest/pull/212/commits/4a972b344f317b69c69bd7cb38c707ec9036fdba)

moisses89 (approved):

PR #215, opened 2023-01-04
Add GET /transactions/queued endpoint

### LINKED ISSUE #156: Add GET /safes/{safe_address}/transactions/queued route
- [x] `/v1/safes/{}/` https://github.com/5afe/safe-client-gateway-nest/pull/116
- [x] `/v1/safes/{}/multisig-transactions/` https://github.com/5afe/safe-client-gateway-nest/issues/167
- [x] `/v1/safe-apps/` https://github.com/5afe/safe-client-gateway-nest/issues/165

hectorgomezv:
Closes #156 

Implements `/chains/:chainId/safes/:safeAddress/transactions/queued` route.


src/routes/common/pagination/pagination.data.ts
```diff
@@ -3,7 +3,7 @@ const QUERY_PARAM_LIMIT = 'limit';
 const QUERY_PARAM_OFFSET = 'offset';
 
 export class PaginationData {
-  private constructor(readonly limit?: number, readonly offset?: number) {}
+  constructor(readonly limit?: number, readonly offset?: number) {}
```
hectorgomezv, line 6: I'm setting this constructor as public since I'm using it to create adjusted paginations [here](https://github.com/5afe/safe-client-gateway-nest/pull/215/files#diff-45bd14f413961db261e493c23aa045bdf6f69f6103f6fc34122688642e8116acR238). I can't see any downside on this, but another option would be using a builder pattern and keep the constructor private, feel free to comment 🙂

src/domain/safe/safe.repository.ts
```diff
@@ -119,7 +119,7 @@ export class SafeRepository implements ISafeRepository {
     const page: Page<MultisigTransaction> =
       await transactionService.getMultisigTransactions(
         safeAddress,
-        '-modified',
+        'nonce',
```
fmrsabino, line 122: Why was this changed?

fmrsabino (changes requested):

src/domain/safe/safe.repository.ts
hectorgomezv: Just copying the Rust implementation, which is [ordering by nonce](https://github.com/safe-global/safe-client-gateway/blob/2534c54ca58878fcf0da9188896b0643505fddc7/src/routes/transactions/handlers/queued.rs#L36). I assumed this was a bug, wasn't it?

fmrsabino: See https://github.com/safe-global/safe-client-gateway/blob/2534c54ca58878fcf0da9188896b0643505fddc7/src/routes/safes/handlers/safes.rs#L122-L130

hectorgomezv: Not sure I'm following 🙂 [That function](https://github.com/safe-global/safe-client-gateway/blob/2534c54ca58878fcf0da9188896b0643505fddc7/src/routes/safes/handlers/safes.rs#L122-L130) gets the last queued transaction, which if I'm not wrong will be used to populate the payload of the Safe Info endpoint. The function I modified gets a list of queued transactions. 

So if I'm not wrong the function I modified needs to call `transactionService.getMultisigTransactions` with `nonce` as `ordering`, and for function you pointed we'd need to implement another function which need to call `transactionService.getMultisigTransactions` with `-modified` as `ordering`. Am I missing something?

hectorgomezv: Another option would be parametrizing the call and just passing the `ordering` as parameter. Not sure which is the clearer solution, what do you think?

hectorgomezv: Yes, I agree that's the best option. So do you think using this `getQueuedTransactions` function to retrieve the transactions ordered by `nonce` is ok, or should we split it into two functions in this PR? 🙂

src/routes/transactions/entities/queued-items/transaction-queued-item.entity.ts
```diff
@@ -0,0 +1,17 @@
...
+import { QueuedItem } from '../queued-item.entity';
+import { Transaction } from '../transaction.entity';
+
+export class TransactionQueuedItem extends QueuedItem {
+  @ApiProperty()
+  transaction: Transaction;
+  @ApiProperty()
+  conflictType: string;
+
+  constructor(transaction: Transaction, conflictType: ConflictType) {
```
moisses89, line 12: `module-transaction.entity`  and `multisig-transaction.entity` and this one are equals, maybe we can refactor it in one entity.

hectorgomezv: Yes, in fact, I had that in a TODO note to propose it in a separate PR 😄. 

In the Rust implementation, I think this is named simply `Transaction`, and Nest's `Transaction` is a `TransactionSummary`in Rust impl. What do you think about following that names? Separate PR or do you prefer to do it here? @moisses89 @fmrsabino

moisses89: Separate PR and follow the Rust names are ok for me.

src/domain/safe/safe.repository.ts
fmrsabino: We should split into two functions in this PR. `getTransactionQueue` with the modified parameter is already being used by a local change of mine.

src/routes/transactions/__tests__/resources/incoming-transfers/e2e/erc721-expected-response.json
```diff
@@ -58,7 +58,7 @@
             "tokenId": "110380171670080793190909999014722746036502744273864122623133899039475607835745",
             "tokenName": "Ethereum Name Service",
             "tokenSymbol": "ENS",
-            "logoUri": "/media/tokens/logos/ENS.png"
+            "logoUri": "/media/tokens/logos/0x57f1887a8BF19b14fC0dF6Fd9B2acc9Af147eA85.png"
```
fmrsabino, line 61: See https://github.com/5afe/safe-client-gateway-nest/pull/218

If this was done to fix an unrelated E2E test it'd be better to have it in a separate PR.

src/routes/transactions/transactions.service.ts
```diff
@@ -175,4 +181,124 @@ export class TransactionsService {
...
+  /**
+   * Adjusts the pagination data to return extra items in both "edges" of the current page:
+   * - If no pagination data info, then return the original pagination data.
+   * - If it is the first page (offset 0), then return offset: 0, limit: limit + 1.
+   * - If it is not the first page, then return offset: offset - 1, limit: limit + 2.
+   * @param paginationData pagination data to adjust.
+   * @returns pagination data adjusted.
+   */
+  private getAdjustedPagination(
+    paginationData?: PaginationData,
```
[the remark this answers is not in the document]
fmrsabino: Fixed here: https://github.com/5afe/safe-client-gateway-nest/pull/219

You can remove the `undefined` parameter from the route added in this PR 👍

fmrsabino (changes requested):

src/domain/safe/safe.repository.ts
hectorgomezv: Perfect, but for what purpose are you using this function in your local changes? I mean, in the Rust implementation I see two purposes: 

- Get the last transaction of the queue.
- Get the list of queued transactions.

Which names do you propose?  Do you plan to maintain `getTransactionQueue` for just picking the last transaction in the queue? 🙂

src/routes/transactions/__tests__/resources/incoming-transfers/e2e/erc721-expected-response.json
hectorgomezv: Yes, after rebasing there's no diff anymore, thanks! 👍🏻

src/domain/safe/safe.repository.ts
fmrsabino: My suggestion would be the following:

```typescript
async getTransactionQueueByModified( // new function for `ISafeRepository`
  chainId: string,
  safeAddress: string,
  limit?: number,
  offset?: number,
): Promise<Page<MultisigTransaction>> {
  return this.getTransactionQueue(
    chainId,
    safeAddress,
    '-modified',
    limit,
    offset,
  );
}

async getTransactionQueueByNonce( // new function for `ISafeRepository`
  chainId: string,
  safeAddress: string,
  limit?: number,
  offset?: number,
): Promise<Page<MultisigTransaction>> {
  return this.getTransactionQueue(
    chainId,
    safeAddress,
    'nonce',
    limit,
    offset,
  );
}

private async getTransactionQueue( // this becomes private. Remove from `ISafeRepository`
  chainId: string,
  safeAddress: string,
  ordering: string, // new ordering parameter
  limit?: number,
  offset?: number,
): Promise<Page<MultisigTransaction>> {
  ...
}
```

Wdyt?

hectorgomezv: Yes, that's a great solution 👍🏻 I was thinking of having something like `getLastQueuedTransaction`, but maybe `getTransactionQueueByModified` makes it more reusable.

I will make the change, thanks.

fmrsabino: > getLastQueuedTransaction

In my opinion, `getLastQueuedTransaction` is so specific that I left it in the route level (local changes) 😉

src/routes/transactions/transactions.controller.ts
```diff
@@ -54,7 +56,6 @@ export class TransactionsController {
       paginationData,
     );
   }
-
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [1ca115c](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/1ca115cafd135af939cb823dff24f26b7e56151a)

src/domain/safe/safe.repository.ts
hectorgomezv: Changed in [1ca115c](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/1ca115cafd135af939cb823dff24f26b7e56151a)

src/routes/transactions/transactions.service.ts
```diff
@@ -175,4 +181,125 @@ export class TransactionsService {
...
+        paginationData.limit + 1,
+        paginationData.offset,
+      );
+    } else {
+      return new PaginationData(
+        paginationData.limit + 2,
+        paginationData.offset - 1,
+      );
+    }
+  }
```
moisses89, line 256: @hectorgomezv why this function is necessary?

fmrsabino (changes requested):

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -612,6 +612,109 @@ describe('Transactions Controller (Unit)', () => {
...
+        const getChainUrl = `${safeConfigApiUrl}/api/v1/chains/${chainId}`;
+        const getMultisigTransactionsUrl = `${chainResponse.transactionService}/api/v1/safes/${safeAddress}/multisig-transactions/`;
+        const getSafeUrl = `${chainResponse.transactionService}/api/v1/safes/${safeAddress}`;
+        const getContractUrlPattern = `${chainResponse.transactionService}/api/v1/contracts/`;
+        if (url === getChainUrl) {
+          return Promise.resolve({ data: chainResponse });
+        }
+        if (url === getMultisigTransactionsUrl) {
+          return Promise.resolve({
+            data: getJsonResource(
```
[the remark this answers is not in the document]
hectorgomezv: Yes, using factories here seems more appropriate, thanks, I will change it. 👍🏻

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -612,6 +612,109 @@ describe('Transactions Controller (Unit)', () => {
...
+        }
+        if (url.includes(getContractUrlPattern)) {
+          return Promise.resolve({
+            data: getJsonResource(
+              'queued-items/single-page/contract-source-data.json',
+            ),
+          });
+        }
+        return Promise.reject(new Error(`Could not match ${url}`));
+      });
```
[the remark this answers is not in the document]
hectorgomezv: Other conditions don't get tested because the first matching condition will exit the function calling `return`. But I'm not against using the `if else` notation if you prefer, or if you find it clearer 🙂

src/routes/transactions/mappers/queued-items/queued-items.mapper.ts
```diff
@@ -0,0 +1,137 @@
...
+        ...result,
+        [executionInfo['nonce']]: [
+          ...(result[executionInfo['nonce']] || []),
+          transaction,
+        ],
+      };
+    }, {});
+
+    return Object.entries(groups).map((group) => ({
+      nonce: Number(group[0]),
```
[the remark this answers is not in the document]
hectorgomezv: Every group should have exactly two elements (by the grouping implementation we are using).

src/routes/transactions/mappers/queued-items/queued-items.mapper.ts
```diff
@@ -0,0 +1,137 @@
...
+  private getMappedTransactionGroup(
+    hasConflicts: boolean,
+    conflictFromPreviousPage: boolean,
+    isEdgeGroup: boolean,
+    transactionGroup: TransactionGroup,
+  ): TransactionQueuedItem[] {
+    const firstItem = this.getAsFirstTransaction(
+      hasConflicts,
+      conflictFromPreviousPage,
+      transactionGroup.transactions[0],
```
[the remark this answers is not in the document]
hectorgomezv: Every group should have at least one transaction (because of the grouping implementation), it's related to [this comment](https://github.com/5afe/safe-client-gateway-nest/pull/215#discussion_r1062685266).

src/routes/transactions/transactions.service.ts
```diff
@@ -175,4 +181,124 @@ export class TransactionsService {
...
+  private getFirstTransactionNonce(
+    page: Page<DomainMultisigTransaction>,
+  ): number {
+    return page.results[0].nonce;
+  }
+
+  private getLastTransactionNonce(
+    page: Page<DomainMultisigTransaction>,
+  ): number {
+    return page.results[page.results.length - 1].nonce;
```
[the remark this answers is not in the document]
hectorgomezv: Thanks, changed (and hopefully simplified) in [d85741c](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/d85741c13deac6ecd3bc235116e439894a2ca16a)

src/routes/transactions/transactions.service.ts
```diff
@@ -175,4 +181,124 @@ export class TransactionsService {
...
+  /**
+   * Checks the if page contains a next cursor.
+   */
+  private hasNextPage(page: Page<DomainMultisigTransaction>): boolean {
+    return page.next !== null;
+  }
+  private getFirstTransactionNonce(
+    page: Page<DomainMultisigTransaction>,
+  ): number {
+    return page.results[0].nonce;
```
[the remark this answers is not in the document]
hectorgomezv: Change in [d85741c](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/d85741c13deac6ecd3bc235116e439894a2ca16a) 👍🏻

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
fmrsabino: Ah yes you are right. I missed the `return` statement 😬

src/routes/transactions/mappers/queued-items/queued-items.mapper.ts
```diff
@@ -114,24 +114,16 @@ export class QueuedItemsMapper {
...
-    return Object.entries(groups).map((group) => ({
-      nonce: Number(group[0]),
-      transactions: group[1] as Transaction[],
-    }));
+    return Object.entries(groupBy(transactions, 'executionInfo.nonce')).map(
+      (group) => ({
+        nonce: Number(group[0]),
+        transactions: group[1] as Transaction[],
+      }),
+    );
```
fmrsabino, line 127: I think now it's a bit more clear what the `group` represents with `groupBy`. But I think we can make it even more clear:

```typescript
private groupByNonce(transactions: Transaction[]): TransactionGroup[] {
  return Object.entries(groupBy(transactions, 'executionInfo.nonce')).map(
    ([nonce, transactions]) =>
      <TransactionGroup>{
        nonce: Number(nonce),
        transactions: transactions,
      },
  );
}
```

src/routes/transactions/mappers/queued-items/queued-items.mapper.ts
```diff
@@ -114,24 +114,16 @@ export class QueuedItemsMapper {
...
-          transaction,
-        ],
-      };
-    }, {});
-
-    return Object.entries(groups).map((group) => ({
-      nonce: Number(group[0]),
-      transactions: group[1] as Transaction[],
-    }));
+    return Object.entries(groupBy(transactions, 'executionInfo.nonce')).map(
```
fmrsabino, line 122: I think working with the route entity at this level has some downsides. You loose some domain related types that can be useful when mapping types.

In this example `nonce` is a `string` even though we end up mapping it to a `number`.

Why not work with the domain entity and map to the final route entity at the end?

If going with the domain option we can explicitly say that we expect a `MultisigTransaction` and not any other transaction. As it stands now `executionInfo.nonce` is not available for every  `Transaction`.

src/routes/transactions/transactions.service.ts
moisses89: > Or what's the purpose of adjusting the pagination values?  
The second one :)

hectorgomezv: I just followed the Rust implementation, in which, an extra item is needed on both the previous and the next page of the page being requested. This is necessary because some logic is in place to help the clients draw the view. The client shows transaction groups and needs to know whether the first/last element is part of a group of the previous/next page and other labeling-related stuff. You can find more details in the Rust service Swagger (and in the code itself of course).

Having that said, since one extra item is requested on each _side_ of the page, an adjustment is needed to remove the extra items from the response. (But again, here I'm just following the logic implemented for the Rust service, IMHO all this logic eventually could be placed in the client side, to have a more agnostic backend response 🤔)

src/routes/transactions/mappers/queued-items/queued-items.mapper.ts
```diff
@@ -0,0 +1,137 @@
...
+
+      items = [
+        ...items,
+        ...this.getMappedTransactionGroup(
+          hasConflicts,
+          conflictFromPreviousPage,
+          isEdgeGroup,
+          transactionGroup,
+        ),
+      ];
```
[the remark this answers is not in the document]
hectorgomezv: Yes, I agree, changed in [06716c5](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/06716c571741022b78adee62a90c5c2ae6b1de3d)

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
hectorgomezv: @fmrsabino I pushed [06716c5](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/06716c571741022b78adee62a90c5c2ae6b1de3d) using some builders just as sneak peak. I see the utility of using builders but I also feel we are hardcoding a little too much data in each test and making them a bit harder to read. Wdyt? Maybe building the objects in a separate file could be a intermediate solution?

src/domain/safe/entities/multisig-transaction.entity.ts
```diff
@@ -16,7 +16,7 @@ export type MultisigTransaction = {
   data: string | null; // TODO will be added under https://github.com/5afe/safe-client-gateway-nest/pull/132
   dataDecoded: any | null;
   ethGasPrice: string | null;
-  executionDate: Date;
+  executionDate: Date | null;
```
hectorgomezv, line 19: This was changed since the transactions in the queue have a `null` `executionDate`. @fmrsabino I could move this to a separate PR if you prefer.

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
hectorgomezv: I've set the PR as a draft because I feel the change in [06716c5](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/06716c571741022b78adee62a90c5c2ae6b1de3d) could be improved (extracting common data, etc.) but I'm not sure this time-consuming change adds a lot of value right now, and I think we have to be conscious of time at this point. Should we remove the usage of `getJsonResource` completely from this file? This would imply also composing the expected responses using builders, and also refactoring all the tests in the file. Of course, we can do it in a separate PR, but I have doubts about doing this work at this point since it's time-consuming. What do you think? @fmrsabino @moisses89  🤔

src/routes/transactions/mappers/queued-items/queued-items.mapper.ts
hectorgomezv: Yes, it's clearer that way, thanks. Changed in  [f187823](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/f187823c26df72b65449d359c96646014fdc0b66)

src/routes/transactions/mappers/queued-items/queued-items.mapper.ts
hectorgomezv: I changed this to use `MultisigTransaction` into `QueuedItemsMapper`. It also gave me the opportunity to refactor the code a bit and merge `getAsFirstTransaction` and `getAsNonFirstTransactions` into a hopefully simpler `getConflictType` which is called for each transaction. I think it's easier to follow the code this way.

hectorgomezv:
I've changed the approach (related to [this discussion](https://github.com/5afe/safe-client-gateway-nest/pull/215#discussion_r1062503121)) in the controller tests to use builders in [96d459c](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/96d459c17a134cedbfc58f7e9a151de95f2d2656). I've removed all the JSON files related to these feature tests. Hopefully, this approach is easier to follow. I also think the use of matchers (i.e. `objectContaining`) makes easier to focus in the parts being tested (in this case ordering, labels, headers, and conflict types) instead of in the transactions mapping itseft (covered by other tests). Any feedback is welcomed @fmrsabino @moisses89 🙂

src/routes/transactions/transactions.service.ts
```diff
@@ -175,4 +182,114 @@ export class TransactionsService {
...
+    const safeInfo = await this.safeRepository.getSafe(chainId, safeAddress);
+    const transactions = await this.safeRepository.getTransactionQueueByNonce(
+      chainId,
+      safeAddress,
+      pagination.limit,
+      pagination.offset,
+    );
+
+    const { next, previous } = transactions;
+    const nextURL = cursorUrlFromLimitAndOffset(routeUrl, next);
```
moisses89, line 202: I'm working in something similar. I think that here you should consider the user request offset and limit instead the adjusted pagination, the tx-service response is calculated from the modified values.

hectorgomezv: Yes, you're right, the offset and limit are not being properly calculated due to this. We need to calculate them with the original values instead of relying on the transaction service response, I'll fix it, thanks!

src/routes/transactions/entities/queued-items/transaction-group.entity.ts
```diff
@@ -0,0 +1,6 @@
+import { MultisigTransaction } from '../../../../domain/safe/entities/multisig-transaction.entity';
+
+export class TransactionGroup {
```
fmrsabino, line 3: This is used only in the `QueuedItemsMapper`.

I would just have the class there and not export it to increase encapsulation.

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -16,14 +16,20 @@ import {
...
 import { chainBuilder } from '../../domain/chains/entities/__tests__/chain.builder';
+import { contractBuilder } from '../../domain/contracts/entities/__tests__/contract.builder';
+import { MultisigTransaction } from '../../domain/safe/entities/multisig-transaction.entity';
 import {
   moduleTransactionBuilder,
   toJson,
 } from '../../domain/safe/entities/__tests__/module-transaction.builder';
+import {
+  multisigTransactionBuilder,
+  toJson as multiSignToJson,
```
fmrsabino, line 28: `multisigToJson` instead of of `multiSignToJson`

src/routes/transactions/transactions.controller.ts
```diff
@@ -109,4 +111,49 @@ export class TransactionsController {
       paginationData,
     );
   }
+
+  @ApiOkResponse({
+    type: QueuedItemPage,
+    description: `
+  This endpoint returns all the transactions that are still
```
fmrsabino, line 118: I've commented on this here: https://github.com/5afe/safe-client-gateway-nest/pull/215#discussion_r1062476876

fmrsabino (changes requested):

src/routes/transactions/transactions.controller.spec.ts
hectorgomezv: Fixed in [d1821d2](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/d1821d25460a597f844c1773d312a01f0621a0d1), thanks!

src/routes/transactions/transactions.controller.ts
hectorgomezv: I deleted the description in [d1821d2](https://github.com/5afe/safe-client-gateway-nest/pull/215/commits/d1821d25460a597f844c1773d312a01f0621a0d1). I understood we could leave this description for now and review it later, but yes, maybe it's better to remove it at this point, and document all the endpoints later 🙂

fmrsabino (dismissed):

fmrsabino (approved):

PR #219, opened 2023-01-05
Fix PaginationDataDecorator return type

fmrsabino:
- `PaginationData.fromCursor` should always return a `PaginationData` even if the necessary data to create one is not provided
- Changes the routes which used `@PaginationDataDecorator()` – the respective parameters are now non-undefined/non-null
- The domain interfaces were not changed – `offset` and `limit` can still be undefined – this allows the route layer to still request the specified data while ignoring pagination parameters


hectorgomezv (approved):
🚀

PR #220, opened 2023-01-05
Add ConfirmationBuilder

fmrsabino:
- Adds a `ConfirmationBuilder` – this builder can be used to create `Confirmation` instances in a testing environment
- Uses the builder as a default value for `MultisigTransactionBuilder.confirmations`
- Uses `dataDecodedFactory` instead of `faker.datatype.json()` in `MultisigTransactionBuilder.dataDecoded`

### DIFF
--- a/src/domain/safe/entities/__tests__/multisig-transaction.factory.ts
+++ b/src/domain/safe/entities/__tests__/multisig-transaction.factory.ts
@@ -4,8 +4,9 @@ import {
 } from '../multisig-transaction.entity';
 import { Operation } from '../operation.entity';
 import { faker } from '@faker-js/faker';
-import multisigTransactionConfirmationFactory from './multisig-transaction-confirmation.factory';
 import { Builder } from '../../../common/__tests__/builder';
+import { ConfirmationBuilder } from './multisig-transaction-confirmation.factory';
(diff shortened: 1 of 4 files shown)


src/domain/safe/entities/__tests__/multisig-transaction.factory.ts
```diff
@@ -211,9 +212,17 @@ export class MultisigTransactionBuilder
 
   toJson(): unknown {
     const entity = this.build();
+    // TODO: this is exactly the same as the one in ConfirmationBuilder
+    const confirmations = entity.confirmations?.map((confirmation) => {
+      return {
+        ...confirmation,
+        submissionDate: confirmation.submissionDate.toISOString(),
+      };
+    });
```
fmrsabino, line 221: This should be discussed as the `.toJson` is tied to the `Builder` interface.

We need to assume a situation where `MultisigTransactionBuilder` has some actual `Confirmation` set. From the `Confirmation` we cannot call `toJson` since it's not part of that class.

fmrsabino:
Moving this to a draft. If https://github.com/5afe/safe-client-gateway-nest/pull/222 is merged this is no longer needed.

fmrsabino:
Closed in favour of https://github.com/5afe/safe-client-gateway-nest/pull/222

PR #221, opened 2023-01-06
Add ERC20TransferBuilder

fmrsabino:
- Adds a `ERC20TransferBuilder` which can be used to build instances of `ERC20Transfer`
- Removes a `TransferInfoMapper` test that checks the result if the transfer type is unknown – this condition should never happen as the mapper only deals with domain entities which do not contain the type. This validation is done by AJV instead, and it should fail at that level.


moisses89 (approved):
Looks good to me!

fmrsabino:
Closed in favour of https://github.com/5afe/safe-client-gateway-nest/pull/222

PR #222, opened 2023-01-06
Add generic builder

fmrsabino:
- Adds a generic builder class: `Builder<T>`
- This `Builder` builds upon a `Partial` type of `T`. Properties of `T` can be set via `with` (which is constrained to the properties of `T`)
- Replaces every domain factory with this `Builder` pattern.
- Some entities require the Json payload to be generated – this request didn't target a common generic solution to this but since the old `Builder` implementation had a `toJson` function, this was adapted to the new solution – a `toJson` function is now exported when required i.e. it's no longer tied to the `Builder` interface.


src/domain/collectibles/entities/__tests__/collectible.builder.ts
```diff
@@ -0,0 +1,17 @@
...
+    .with('address', faker.finance.ethereumAddress())
+    .with('tokenName', faker.company.name())
+    .with('tokenSymbol', faker.finance.currencySymbol())
+    .with('logoUri', faker.internet.url())
+    .with('id', faker.datatype.uuid())
+    .with('uri', faker.internet.url())
+    .with('name', faker.company.name())
+    .with('description', faker.random.words())
+    .with('imageUri', faker.internet.url())
+    .with('metadata', {});
```
moisses89, line 16: Maybe is better to fake a random json.

src/domain/safe/entities/__tests__/multisig-transaction.builder.ts
```diff
@@ -0,0 +1,50 @@
...
+  return Builder.new<MultisigTransaction>()
+    .with('baseGas', faker.datatype.number())
+    .with('blockNumber', faker.datatype.number())
+    .with('confirmations', [confirmationBuilder().build()])
+    .with('confirmationsRequired', faker.datatype.number())
+    .with('data', faker.datatype.hexadecimal())
+    .with('dataDecoded', faker.datatype.json())
+    .with('ethGasPrice', faker.datatype.hexadecimal())
+    .with('executionDate', faker.date.recent())
+    .with('fee', faker.datatype.hexadecimal())
```
moisses89, line 20: executer wasn't included https://github.com/5afe/safe-client-gateway-nest/pull/222/files#diff-272990ad7b76ae9ca394cb353d1b93b87c6713630a48c8d1d5dd0b9a67590234L23

src/domain/safe/entities/__tests__/multisig-transaction.builder.ts
```diff
@@ -0,0 +1,50 @@
...
+    .with('gasPrice', faker.datatype.hexadecimal())
+    .with('gasToken', faker.finance.ethereumAddress())
+    .with('gasUsed', faker.datatype.number())
+    .with('isExecuted', faker.datatype.boolean())
+    .with('isSuccessful', faker.datatype.boolean())
```
moisses89, line 25: modified wasn't included https://github.com/5afe/safe-client-gateway-nest/pull/222/files#diff-272990ad7b76ae9ca394cb353d1b93b87c6713630a48c8d1d5dd0b9a67590234L30

moisses89 (commented):
Awesome work! a big improvement for our tests :)

src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -31,9 +31,12 @@ describe('ExchangeApi', () => {
   });
 
   it('Should return the fiatCodes', async () => {
-    const expectedFiatCodes = exchangeFiatCodesFactory(true, {
-      USD: 'Dollar',
-    });
+    const expectedFiatCodes = exchangeFiatCodesBuilder()
+      .with('success', true)
+      .with('symbols', {
```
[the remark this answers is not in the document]
fmrsabino: Do you mean set in the `Builder` (we don't have a constructor for `ExchangeFiatCodes`)?

I just converted the way the tests were setting this fields to use the builder. Maybe it's relevant to show in the test that `success` is `true`. But I didn't go as far as reviewing each test (I just converted the factory usage to the builder one)

src/domain/safe/entities/__tests__/multisig-transaction.builder.ts
fmrsabino: Fixed in [b793a9d](https://github.com/5afe/safe-client-gateway-nest/pull/222/commits/b793a9d39a95f02911e8b2276a615103919fac24)

hectorgomezv:
👏🏻

PR #225, opened 2023-01-16
Add web and nginx images to docker compose

fmrsabino:
- Adds a `web` image (the NestJS service) to docker compose
  * This web image reads the `APPLICATION_PORT` and the `EXCHANGE_API_KEY` from the local environment and uses these values for configuring the service
  * This is mostly to showcase how the web image works behind the Nginx reverse proxy (no other environment variables/configuration will be read)
- Adds a Nginx image as a reverse proxy in docker compose
  * The container behind this image is accessible to the host via `NGINX_HOST_PORT` (defaults to `8080`)
  * This container is responsible to redirect the incoming requests to the underlying web application

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,5 +2,27 @@ services:
   redis:
     image: redis/redis-stack:7.0.2-RC1
     ports:
-      - "${REDIS_PORT-6379}:${REDIS_PORT-6379}"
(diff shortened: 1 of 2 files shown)


fmrsabino:
@luarx this configuration should give us an initial template on how we can deploy this to our environment.

luarx (approved):

PR #226, opened 2023-01-17
Add features property to Safe App

fmrsabino:
- With the recent addition of `SafeApp.features` in the Safe Config Service, and given that the schema has `additionalProperties` set to `false` we need to add `features` to the schema to make it valid again
- Adds the `features` property to the `SafeApp` entity and builder
- See https://github.com/safe-global/safe-config-service/commit/814b8eb5faaabf43809f4852e9cac330e0501335

### DIFF
--- a/src/domain/safe-apps/entities/__tests__/safe-app.builder.ts
+++ b/src/domain/safe-apps/entities/__tests__/safe-app.builder.ts
@@ -14,5 +14,6 @@ export function safeAppBuilder(): IBuilder<SafeApp> {
     .with('chainIds', [faker.datatype.number(), faker.datatype.number()])
     .with('provider', safeAppProviderBuilder().build())
     .with('accessControl', safeAppAccessControlBuilder().build())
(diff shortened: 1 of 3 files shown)


hectorgomezv (approved):

hectorgomezv:
🚀

Btw, we need to include these `features` property in the `SafeApp` entity and builder as well I think (we could do it in a separate PR) 🙂

fmrsabino:
> 🚀
> 
> Btw, we need to include these `features` property in the `SafeApp` entity and builder as well I think (we could do it in a separate PR) 🙂

I think it makes sense to address in this PR too. Added here [9312d51](https://github.com/5afe/safe-client-gateway-nest/pull/226/commits/9312d51e1f4c6046e3876c51e337901c90034459)

hectorgomezv (dismissed):

moisses89 (approved):
Looks good to me!

hectorgomezv (approved):

PR #228, opened 2023-01-17
Revert "Disable test:e2e:cov task (#217)"

fmrsabino:
This reverts commit 6aa81d72e60b3a3183ac4c855b26bb25396d8dd8.

We consider that the E2E tests were fixed so we can reenable the task again.


fmrsabino:
@5afe/safe-services Rebased and ready to be reviewed! ✅

PR #234, opened 2023-01-19
Avoid floating point numbers for transactions dataSize value

hectorgomezv:
`dataSize` should be an integer value, (see [the Rust implementation of the function](https://github.com/safe-global/safe-client-gateway/blob/501371f64c13ee0a0585eaf5a0909ded7299ad7f/src/routes/transactions/converters/mod.rs#L295) which calculates the `dataSize` value) the current implementation was returning floating-point numbers.


hectorgomezv:
🚀

PR #236, opened 2023-01-20
Add SafeAppInfo field to MultisigTransactions response

### LINKED ISSUE #201: Map Safe Apps "origin" field for Multisig Transactions
Safe Apps `origin` field should be mapped when retrieving multisig transactions. This work is extracted from #197. It seems [the Rust implementation](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/transactions/converters/safe_app_info.rs) contains special logic to map this field, so it deserves some special attention.

hectorgomezv:
Closes #201 

Added `SafeAppInfoMapper` which takes the responsibility for mapping `SafeAppInfo`.

- When a transaction with a non-null origin URL field is received, its value is searched against the data returned by `/safe-apps` endpoint of the Config Service.
- If no coincidence is found for the URL within the list of Safe Apps, `null` is returned (to mimic the behavior of the current Rust implementation), but additionally a log warn line is generated for easier monitoring.
- Replacement of IPFS URLs are kept (from `ipfs.io` to `cloudflare-ipfs.com`) as we have in the Rust implementation.


src/routes/transactions/mappers/common/safe-app-info.mapper.ts
```diff
@@ -0,0 +1,52 @@
...
+    const originUrl = this.getOriginUrl(transaction);
+    if (!originUrl) return null;
+
+    const [safeApp] = await this.safeAppsRepository.getSafeApps(
+      chainId,
+      undefined,
+      originUrl,
+    );
+    if (!safeApp) {
+      this.logger.error(
```
fmrsabino, line 31: This should not be an error in my opinion.

Maybe `info` or `warning`. Not finding a Safe App is part of what we consider to be "normal".

src/routes/transactions/mappers/common/safe-app-info.mapper.ts
```diff
@@ -0,0 +1,52 @@
...
+      ),
+      safeApp.iconUrl,
+    );
+  }
+
+  private getOriginUrl(transaction: MultisigTransaction): string | null {
+    return transaction.origin
+      ? JSON.parse(transaction.origin).url ?? null
+      : null;
+  }
```
fmrsabino, line 51: If we expect this to be a JSON payload with a `url` why not solve this at the AJV level?

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -250,14 +251,28 @@ describe('Transactions Controller (Unit)', () => {
...
+      const safeAppsResponse = [
+        safeAppBuilder()
+          .with('url', 'https://apps.gnosis-safe.io/safe-claiming-app')
+          .with(
+            'iconUrl',
+            'https://safe-apps.dev.5afe.dev/safe-claiming-app/logo.svg',
+          )
+          .with('name', '$SAFE Claiming App')
+          .build(),
+      ];
```
fmrsabino, line 263: Did you consider using fake data here?

I don't really see the benefit right now of reading JSON files in the tests when we can compare the payloads with a TS/JS object here. Generated fake data would also provide different inputs on each run which can be valuable.

src/routes/transactions/mappers/common/safe-app-info.mapper.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+import { multisigTransactionBuilder } from '../../../../domain/safe/entities/__tests__/multisig-transaction.builder';
+import { SafeAppInfo } from '../../entities/safe-app-info.entity';
+import { SafeAppInfoMapper } from './safe-app-info.mapper';
+
+describe('SafeAppInfo mapper (Unit)', () => {
+  const safeAppsRepositoryMock = jest.mocked({
+    getSafeApps: jest.fn(),
+  } as unknown as SafeAppsRepository);
+
+  const mapper = new SafeAppInfoMapper(safeAppsRepositoryMock);
```
fmrsabino, line 13: We can face issues with this in the future if an internal state is introduced in the mapper.

I would build the mapper before each test.

fmrsabino (changes requested):

src/routes/transactions/transactions.controller.spec.ts
hectorgomezv: I've rewritten this test using builders instead of JSON files in [d9abbce](https://github.com/5afe/safe-client-gateway-nest/pull/236/commits/d9abbce9dd91fd30e603198f8ea696748569eb1a). For the expected response I've just written the expected object in the test. An alternative approach would be using [Jest matchers](https://jestjs.io/docs/expect#expectobjectcontainingobject) to assert the parts of the response in which we are interested are correct. So, to summarise, I see several options to assert the response objects:
- Keep using JSON files for them.
- Use plain JS objects in each test.
- Use Jest matchers to check the interesting parts of the response.
- Use builders also for responses (i.e. use builder for route entities).

What do you think? 🙂

src/routes/transactions/mappers/common/safe-app-info.mapper.ts
hectorgomezv: Hmm, `transaction.origin` is actually a `string` which contains a JSON. By solving it at AJV level you mean using a custom keyword as we did for `isDate`? I'm saying this because I think we would need to parse the JSON content explicitly to validate it 🤔

src/routes/transactions/mappers/common/safe-app-info.mapper.ts
hectorgomezv: I'm changing it to `warn` in [d9abbce](https://github.com/5afe/safe-client-gateway-nest/pull/236/commits/d9abbce9dd91fd30e603198f8ea696748569eb1a), I thought not finding a Safe App could be considered an error. Anyway, ping me if you feel `info` is more appropriate here.

src/routes/transactions/mappers/common/safe-app-info.mapper.spec.ts
hectorgomezv: I'd say the mappers should be stateless but yes, I agree it's a best practice, thanks! Changed in [ec85300](https://github.com/5afe/safe-client-gateway-nest/pull/236/commits/ec85300efc26bd20ca0b69b3e6606d6336138dd1)

src/routes/transactions/mappers/common/safe-app-info.mapper.ts
fmrsabino: I meant if we expect it to be of a JSON format always then we can try to solve it at the parsing
/validation level but I get your point. Let's leave it like this for now 👍

src/routes/transactions/transactions.controller.spec.ts
fmrsabino: I would say it's fine to have the play JS object as the expected object in each test.

I would try to avoid using internal tools (builders or jest) to represent the expected contract at this stage. We can reconsider this if we don't see value in having the complete payload in the expected response.

fmrsabino (changes requested):

src/routes/transactions/transactions.controller.spec.ts
hectorgomezv: Perfect, I agree we can use just plain JS objects for now. I think in other cases as I commented out [here](https://github.com/5afe/safe-client-gateway-nest/pull/215#issuecomment-1401752850) maybe it's better to use some matchers because the focus of the tests is not the mapping. But for this PR I think plain JS objects are the way to go since we're interested in the explicit mappings of transactions 👍

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -249,34 +260,91 @@ describe('Transactions Controller (Unit)', () => {
...
+        confirmationBuilder().build(),
+        confirmationBuilder().build(),
+      ];
+      const safeAppsResponse = [
+        safeAppBuilder()
+          .with('url', safeAppUrl)
+          .with('iconUrl', safeAppLogoUri)
+          .with('name', safeAppName)
+          .build(),
+      ];
```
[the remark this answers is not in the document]
hectorgomezv: I've removed some intermediate variables in [f714450](https://github.com/5afe/safe-client-gateway-nest/pull/236/commits/f71445053424d8dfd6d54d73e5b21c2517fbc5a5), but I'm not sure I can use the `MultisigTransaction` built directly on the response assertion, since the builder creates a domain object and the response is a route object 🤔 Maybe I'm missing something.

src/routes/transactions/mappers/common/safe-app-info.mapper.spec.ts
```diff
@@ -0,0 +1,113 @@
...
+  const safeAppsRepositoryMock = jest.mocked({
+    getSafeApps: jest.fn(),
+  } as unknown as SafeAppsRepository);
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+  });
+
+  it('should get a null SafeAppInfo for a transaction with no origin', async () => {
+    const mapper = new SafeAppInfoMapper(safeAppsRepositoryMock);
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [f714450](https://github.com/5afe/safe-client-gateway-nest/pull/236/commits/f71445053424d8dfd6d54d73e5b21c2517fbc5a5)

src/routes/transactions/mappers/common/safe-app-info.mapper.spec.ts
```diff
@@ -10,12 +10,15 @@ describe('SafeAppInfo mapper (Unit)', () => {
...
+  let mapper: SafeAppInfoMapper;
+
   beforeEach(async () => {
     jest.clearAllMocks();
+    mapper = new SafeAppInfoMapper(safeAppsRepositoryMock);
   });
 
   it('should get a null SafeAppInfo for a transaction with no origin', async () => {
-    const mapper = new SafeAppInfoMapper(safeAppsRepositoryMock);
+    mapper = new SafeAppInfoMapper(safeAppsRepositoryMock);
```
fmrsabino, line 21: You don't need this line. You can assume that you have a `mapper` at this stage (due to `beforeEach`)

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
fmrsabino: Not sure if I understand. The `domainTransaction` in https://github.com/5afe/safe-client-gateway-nest/pull/236/files#diff-0d3b6a85b1cf9923c19fdfdc2297d006077468cf7450dea8d7cb8e4eacf572a6R278-R307

it's being used by just setting fake values. Why not use `multisigTransactionBuilder().build()`?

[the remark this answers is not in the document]
hectorgomezv: Hmm, I think that some parts of `domainTransaction` are as you said just random data, but other are important to have them set to concrete values, since we need to assert in the response those values are mapped correctly. I mean, for example, we need to set a concrete `dataDecoded` value to assert `actionCount` in the response, we need to set both concrete `isExecuted` and `isSuccessful` to assert `txStatus` contains the expected value, etc. 

I think I've set the minimum amount of non-random values for `domainTransaction` in . Please let me know if you see this could be handled in a different way 🙂

Also, going through this, I think `MultisigTransactionBuilder` could be improved a little, because for example:
```javascript
.with('origin', faker.internet.url())
```
is not correct: `origin` should be a string containing a JSON. 

And also:
```javascript
.with('confirmations', [confirmationBuilder().build()])
``` 
could generate a random number (between 0 and 5 for example) of confirmations instead of just 1, this will make tests stronger. What do you think? I can create a new PR with these minor improvements 🙂

src/routes/transactions/mappers/common/safe-app-info.mapper.spec.ts
hectorgomezv: Oh yes, sorry these lines weren't removed by mistake, I should have seen this in the review 🤦 Thanks! 🙂 Fixed in [a9d7769](https://github.com/5afe/safe-client-gateway-nest/pull/236/commits/a9d776911ee96604a21921ca6f63720492e3dee2)

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
fmrsabino: > Hmm, I think that some parts of `domainTransaction` are as you said just random data, but other are important to have them set to concrete values, since we need to assert in the response those values are mapped correctly. I mean, for example, we need to set a concrete `dataDecoded` value to assert `actionCount` in the response, we need to set both concrete `isExecuted` and `isSuccessful` to assert `txStatus` contains the expected value, etc.
> 
> I think I've set the minimum amount of non-random values for `domainTransaction` in [a9d7769](https://github.com/5afe/safe-client-gateway-nest/pull/236/commits/a9d776911ee96604a21921ca6f63720492e3dee2). Please let me know if you see this could be handled in a different way 🙂

I do think that now it's better because it seems that the data that it's being generated is more targeted to what is required for the test 👍

> 
> Also, going through this, I think `MultisigTransactionBuilder` could be improved a little, because for example:
> could generate a random number (between 0 and 5 for example) of confirmations instead of just 1, this will make tests stronger. What do you think? I can create a new PR with these minor improvements 🙂

For this I'd suggest to create a PR as you mentioned 👍

fmrsabino (approved):

hectorgomezv:
🎉🚀

PR #240, opened 2023-01-23
Implement GET /chains/:chain_id/about

hectorgomezv:
- Adds `/v1/chains/:chain_id/about` route to the service.
- The payload returned is formed by aggregating Config Service data (`transactionServiceBaseUri`) and some of the service's internal configuration data (`name`, `version`, `buildNumber`).
- Controller tests and OpenAPI specifications are included.
- No E2E was included since the expected data for this endpoint is volatile.

### DIFF
--- a/src/routes/chains/chains.controller.spec.ts
+++ b/src/routes/chains/chains.controller.spec.ts
@@ -405,8 +405,7 @@ describe('Chains Controller (Unit)', () => {
 
   describe('GET /:chainId/about', () => {
     it('Success', async () => {
-      const chainId = faker.random.numeric();
-      const chainDomain = chainBuilder().with('chainId', chainId).build();
+      const chainDomain = chainBuilder().build();
       const name = faker.random.words();
       const version = faker.system.semver();
       const buildNumber = faker.random.numeric();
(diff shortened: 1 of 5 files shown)


moisses89 (approved):

src/routes/chains/chains.controller.spec.ts
```diff
@@ -402,4 +402,60 @@ describe('Chains Controller (Unit)', () => {
         });
     });
   });
+
+  describe('GET /:chainId/about', () => {
+    it('Success', async () => {
+      const chainId = faker.random.numeric();
```
fmrsabino, line 408: nitpick: you can get the `chainId` from the `chainBuilder()` itself. This can be useful if you don't want to set any chain id specifically as it generates one for you.

fmrsabino (approved):

src/routes/chains/chains.controller.spec.ts
hectorgomezv: True, thanks! Changed in [ad8a8a3](https://github.com/5afe/safe-client-gateway-nest/pull/240/commits/ad8a8a37a0635cfebfe73331756645bab851417c)

fmrsabino (approved):

PR #242, opened 2023-01-25
Get history transactions endpoint

### LINKED ISSUE #155: Add GET /safes/{safe_address}/transactions/history

moisses89:
Closes #155 
- Add get transactions history endpoint
- Add mapper for generic transaction (multisigTransaction, moduleTransaction, ethereumtransaction)
- Add DateLabel entity
- Add safeCreationInfo entity (fill txInfo of CreationTransaction)
- Add safeCreation mapper. 
- Add CreationTransaction to Transaction domain


src/routes/transactions/entities/transaction-history.entity.ts
```diff
@@ -0,0 +1,18 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { ConflictType } from './conflict-type.entity';
+import { Transaction } from './transaction.entity';
+
+export class TransactionHistory {
```
fmrsabino, line 5: I'd use a different name here because it doesn't need to be tied to "history".

Something like `TransactionItem`.

src/routes/transactions/transactions.service.ts
```diff
@@ -175,4 +185,119 @@ export class TransactionsService {
...
+  private getDayInMillis(timestamp: number, timezoneOffset?: string): number {
+    const date = new Date(timestamp);
+    if (timezoneOffset !== undefined) {
+      date.setUTCSeconds(parseInt(timezoneOffset) || 0);
+    }
+    return Date.UTC(date.getFullYear(), date.getMonth(), date.getDate());
+  }
+
+  private groupByDay(
+    transactions: TransactionHistory[],
```
fmrsabino, line 198: Why not use the domain entity?

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,63 @@
...
+    private readonly moduleTransactionMapper: ModuleTransactionMapper,
+    private readonly incomingTransferMapper: IncomingTransferMapper,
+  ) {}
+  async mapTransaction(
+    chainId: string,
+    transactionsDomain: TransactionDomain[],
+    safe: Safe,
+  ): Promise<TransactionHistory[]> {
+    const results = await transactionsDomain.map(async (transaction) => {
+      if ('isExecuted' in transaction) {
```
fmrsabino, line 26: We should work with domain entities at this level and use type predicates instead.

src/routes/transactions/transactions.service.ts
```diff
@@ -175,4 +185,119 @@ export class TransactionsService {
...
+    offset?: number,
+  ): PaginationData {
+    if (offset !== undefined && offset > 0) {
+      if (limit === undefined) {
+        limit = 20; //default limit
+      }
+      return new PaginationData(limit + 1, offset - 1);
+    }
+    return new PaginationData(limit, offset);
+  }
```
fmrsabino, line 225: You always have a default in the route. No need to set the defaults here.

fmrsabino (changes requested):

src/routes/transactions/transactions.service.ts
moisses89: @fmrsabino default pagination is coming here https://github.com/5afe/safe-client-gateway-nest/pull/263 if you want to take a look.

fmrsabino: > @fmrsabino default pagination is coming here #263 if you want to take a look.

You already have a default on the route for `PaginationData`. https://github.com/5afe/safe-client-gateway-nest/blob/main/src/routes/common/decorators/pagination.data.decorator.ts

src/routes/transactions/mappers/common/transaction.mapper.ts
fmrsabino: This should be helpful to your PR. I'm fine extracting this to a different PR.

https://github.com/5afe/safe-client-gateway-nest/pull/247/files#diff-b5a64dd573f3ba03d25e2ac0a50dd52bc7935a0f217c0c2f2984a375a35c6579R10-R26

moisses89: Yes it is, but I guess that your PR is going to be merged soon, I can wait and see if there is any place more that we can be aligned.

src/routes/transactions/transactions.service.ts
moisses89: Because I need map first the transaction domain to `TransactionHistory`. For example each Transfer in `EthereumTransaction` is a `TransactionHistory` with timestamp and this timestamps should be considered each one to be grouped by day.

src/routes/transactions/entities/transaction-history-group.entity.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ModuleTransaction } from '../../../domain/safe/entities/module-transaction.entity';
+import { EthereumTransaction } from '../../../domain/safe/entities/ethereum-transaction.entity';
+
+export class TransactionDomainGroup {
+  timestamp: number;
+  transactions:
+    | MultisigTransaction[]
+    | ModuleTransaction[]
+    | EthereumTransaction[];
+}
```
moisses89, line 11: I need this to group the domain transactions, maybe should move it to domain folder. :thinking:

src/routes/transactions/transactions.service.ts
moisses89: Done https://github.com/5afe/safe-client-gateway-nest/pull/242/commits/d4188008ab148ab6a23fbd76f30d96e2ba14e495 :smiley:

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +224,72 @@ export class TransactionsService {
...
+      );
+    }
+  }
+
+  async getTransactionHistory(
+    chainId: string,
+    routeUrl: Readonly<URL>,
+    safeAddress: string,
+    timezoneOffset?: string,
+    paginationData?: PaginationData,
```
fmrsabino, line 249: This should not be `undefined`. You should always have a `PaginationData`

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +224,72 @@ export class TransactionsService {
     };
   }
 
+  private adjustTransactionHistoryPage(
+    limit?: number,
+    offset?: number,
```
fmrsabino, line 229: This should not be `undefined`

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +224,72 @@ export class TransactionsService {
...
+    }
+  }
+
+  async getTransactionHistory(
+    chainId: string,
+    routeUrl: Readonly<URL>,
+    safeAddress: string,
+    timezoneOffset?: string,
+    paginationData?: PaginationData,
+  ): Promise<Partial<TransactionItemPage>> {
```
fmrsabino, line 250: Why does this function return a `Partial` type?

src/routes/transactions/entities/transaction-history-group.entity.ts
fmrsabino: This is only used in `TransactionMapper`. Why not have it there?

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,142 @@
...
+import { DateLabel } from '../../entities/date-label.entity';
+
+@Injectable()
+export class TransactionMapper {
+  constructor(
+    private readonly multisigTransactionMapper: MultisigTransactionMapper,
+    private readonly moduleTransactionMapper: ModuleTransactionMapper,
+    private readonly incomingTransferMapper: IncomingTransferMapper,
+  ) {}
+  async mapTransaction(
```
fmrsabino, line 27: This function is called `mapTransaction` but it takes a collection of transactions.

I'd call it `mapTransactions` as it is more explicit. But maybe a better approach would be to have a `mapTransaction` where it takes a single transaction. Acting on a collection can be done in a simple way using the language features.

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,142 @@
...
+    chainId: string,
+    transactionsDomain: TransactionDomain[],
+    safe: Safe,
+    timezoneOffset?: string,
+    offset?: number,
+  ): Promise<[TransactionItem | DateLabel]> {
+    const transactionList: any[] = [];
```
fmrsabino, line 34: Why `any`? We should know which types to expect or?

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,142 @@
...
+        (
+          transactionsDomain[0] as MultisigTransaction
+        ).submissionDate?.getTime();
+      if (timestamp !== undefined) {
+        prev_page_timestamp = this.getDayInMillis(timestamp, timezoneOffset);
+      } else {
+        throw Error('ExecutionDate cannot be null');
+      }
+      transactionsDomain = transactionsDomain.slice(1);
+    }
```
fmrsabino, line 50: Honestly, I find this section quite confusing.

The conditional branches should have some documentation to follow along. The nesting also doesn't help in my opinion.

It seems that we are trying to get the previous page timestamp therefore I'd probably encapsulate this in a function (suggestion).

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,142 @@
...
+    const transactionsDomainGroups = this.groupByDay(
+      transactionsDomain,
+      timezoneOffset,
+    );
+    transactionsDomainGroups.forEach((transactionGroup) => {
```
fmrsabino, line 55: There's a lot of nesting in this section too which makes it more difficult to read and follow the code.

My suggestion is to restructure this part. It seems that we are tackling different types so that's a good starting point for splitting these branches.

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,142 @@
...
+      groupBy(transactions, (transaction) => {
+        let transaction_timestamp;
+        if (isMultisigTransaction(transaction)) {
+          transaction_timestamp =
+            transaction.executionDate ?? transaction.submissionDate;
+        } else {
+          transaction_timestamp = transaction.executionDate;
+        }
+        return this.getDayInMillis(
+          transaction_timestamp?.getTime(),
```
fmrsabino, line 122: We are converting here to a UNIX timestamp only to create a date again inside `getDayInMillis`.

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,142 @@
...
+    ).map(
+      ([timestamp, transactions]) =>
+        <TransactionDomainGroup>{
+          timestamp: Number(timestamp),
+          transactions: transactions,
+        },
+    );
+  }
+
+  private getDayInMillis(timestamp: number, timezoneOffset?: string): number {
```
fmrsabino, line 135: nitpick: for me this seems more of a function that "resets" the hours of the day to 00:00.

You can for example use `date.setHours(0, 0, 0, 0)` or similar.

`getDayInMillis` can be confusing. This function should probably take a `Date` and return a `Date` too.

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
```diff
@@ -0,0 +1,50 @@
...
+      : null;
+    const factory = transaction.factoryAddress
+      ? await this.addressInfoHelper.getOrDefault(
+          chainId,
+          transaction.factoryAddress,
+        )
+      : null;
+    const txInfo = new CreationTransactionInfo(
+      creator,
+      transactionHash,
```
fmrsabino, line 37: we can just take from `transaction.transactionHash`

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
```diff
@@ -0,0 +1,50 @@
...
+      transaction.creator,
+    );
+    const transactionHash = transaction.transactionHash;
+    const implementation = transaction.masterCopy
+      ? await this.addressInfoHelper.getOrDefault(
+          chainId,
+          transaction.masterCopy,
+        )
+      : null;
+    const factory = transaction.factoryAddress
```
fmrsabino, line 29: Doesn't `transaction.factoryAddress` always have a value?

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
```diff
@@ -0,0 +1,50 @@
...
+        )
+      : null;
+    const txInfo = new CreationTransactionInfo(
+      creator,
+      transactionHash,
+      implementation,
+      factory,
+    );
+    return new Transaction(
+      `creation_${safe.address}`,
```
fmrsabino, line 42: `creation` should be extracted somewhere else. Having the raw string value here does not provide any context on what it means.

fmrsabino (changes requested):

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -1075,6 +1088,371 @@ describe('Transactions Controller (Unit)', () => {
         });
     });
   });
+  describe('GET transactions history', () => {
```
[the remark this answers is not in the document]
moisses89: yes, makes sense

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: Shouldn't be any, got it.

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: I'll move it to a function.

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -1075,6 +1088,371 @@ describe('Transactions Controller (Unit)', () => {
...
+          return Promise.resolve({
+            data: transactionHistoryBuilder,
+          });
+        }
+        if (url === getSafeUrl) {
+          return Promise.resolve({
+            data: safe,
+          });
+        }
+        if (url.includes(gettokenUrlPattern)) {
```
hectorgomezv, line 1297: Nitpick: `getTokenUrlPattern` 🙂

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -1075,6 +1088,371 @@ describe('Transactions Controller (Unit)', () => {
...
+      const chainId = faker.random.numeric();
+      const safeAddress = faker.finance.ethereumAddress();
+      const chainResponse = chainBuilder().with('chainId', chainId).build();
+      const limit = 5;
+      const offset = 5;
+      const moduleTransaction = moduleTransactionTojson(
+        moduleTransactionBuilder().with('dataDecoded', null).build(),
+      );
+      const safe = safeBuilder().build();
+      const client_next_cursor = `cursor=limit%3D${limit}%26offset%3D10`;
```
hectorgomezv, line 1413: Nitpick: we should use camelCase to name variables I think 🙂

src/routes/transactions/mappers/common/transaction.mapper.ts
```diff
@@ -0,0 +1,142 @@
...
+    );
+  }
+
+  private groupByDay(
+    transactions: TransactionDomain[],
+    timezoneOffset?: string,
+  ): TransactionDomainGroup[] {
+    return Object.entries(
+      groupBy(transactions, (transaction) => {
+        let transaction_timestamp;
```
hectorgomezv, line 114: Nitpick: snake_case variable name here 🙂

src/routes/transactions/mappers/common/transaction.mapper.ts
hectorgomezv, line 27: Opinionated comment: I'd split this function code into smaller private functions for better readability and maintainabily (for example, extracting the timestamp calculation, and also having smaller functions with map each `TransactionDomain` type to a `TransactionItem`)

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -1075,6 +1088,371 @@ describe('Transactions Controller (Unit)', () => {
...
+      const transfer: any = nativeTokenTransferToJson(
+        nativeTokenTransferBuilder()
+          .with('executionDate', new Date('2022-12-31T17:09:36Z'))
+          .build(),
+      );
+      const incomingTransaction = ethereumTransactionToJson(
+        ethereumTransactionBuilder()
+          .with('executionDate', new Date('2022-12-31T17:09:36Z'))
+          .with('transfers', [transfer])
+          .build(),
```
hectorgomezv, line 1177: Is it possible to use faker/builder defaults to set the dates here? If we need to set the dates to a specific day we could use `faker.date.between` and reference the expected values in the assertions.

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +224,72 @@ export class TransactionsService {
...
+    const paginationData_adjusted = this.adjustTransactionHistoryPage(
```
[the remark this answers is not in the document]
moisses89: Agree

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: yes, makes sense, I'll split this mapper :)

src/domain/safe/entities/__tests__/creation-transaction.builder.ts
```diff
@@ -13,3 +13,11 @@ export function creationTransactionBuilder(): IBuilder<CreationTransaction> {
     .with('setupData', faker.datatype.hexadecimal())
     .with('dataDecoded', dataDecodedBuilder().build());
 }
+
+export function toJson(creationTransaction: CreationTransaction): unknown {
+  return {
+    ...creationTransaction,
+    created: creationTransaction.created.toISOString(),
+    executionDate: creationTransaction.created.toISOString(),
```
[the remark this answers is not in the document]
moisses89: It was a mistake.

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -1075,6 +1088,371 @@ describe('Transactions Controller (Unit)', () => {
...
+
+      expect(mockNetworkService.get).toBeCalledTimes(1);
+      expect(mockNetworkService.get).toBeCalledWith(
+        `${safeConfigApiUrl}/api/v1/chains/${chainId}`,
+        undefined,
+      );
+    });
+
+    it('Failure: Transaction API fails', async () => {
+      const chainId = faker.random.numeric();
```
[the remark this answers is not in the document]
moisses89: Because count is not optional and this endpoint is not returning the count field.

src/routes/transactions/transactions.service.ts
moisses89: Because count is not optional and this endpoint is not returning the count field.

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: I would be more explicit using UTC function, I agree to return Date.

src/routes/transactions/entities/transaction-history-group.entity.ts
moisses89: Done 354fbed6e78c83e90cf797088b4d3f5fc3f9bfc4

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
moisses89: Done d270a0390d810add2a7c263fb46bb57e7bbf9090

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
moisses89: Done 1dbe018c3d842b3c088d028599474b93d558b8cb

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -1075,6 +1088,371 @@ describe('Transactions Controller (Unit)', () => {
...
+
+      await request(app.getHttpServer())
+        .get(`/chains/${chainId}/safes/${safeAddress}/transactions/history/`)
+        .expect(200)
+        .then(({ body }) => {
+          expect(body.results).toHaveLength(
+            6, //3 date labels and 3 transactions
+          );
+          expect(body.results[0].timestamp).toEqual(
+            new Date('2022-12-06T00:00:00Z').getTime(),
```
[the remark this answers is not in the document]
moisses89: Done 986baae4b045d33048d9d5c01375019470534632

src/routes/transactions/transactions.service.ts
moisses89: Done 3ed0214c5e8c5f8593e35c3295694f62ae20da5e

src/routes/transactions/transactions.service.ts
moisses89: Done 3ed0214c5e8c5f8593e35c3295694f62ae20da5e

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: Done 83b9483823373e2c0541a968e0412c38f6d2c766

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: Done 3ed0214c5e8c5f8593e35c3295694f62ae20da5e

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: Done 83b9483823373e2c0541a968e0412c38f6d2c766

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: Done 986baae4b045d33048d9d5c01375019470534632

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: Done 986baae4b045d33048d9d5c01375019470534632

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
moisses89: Done 986baae4b045d33048d9d5c01375019470534632

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
moisses89: Done 986baae4b045d33048d9d5c01375019470534632

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
moisses89: Done 986baae4b045d33048d9d5c01375019470534632 , my solution was create a constant type to be more clear.

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +224,72 @@ export class TransactionsService {
     };
   }
 
+  private adjustTransactionHistoryPage(
```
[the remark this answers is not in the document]
moisses89: Done 9bff7de84cb59f256094b48f1d8da232c1e29ac2

src/routes/transactions/transactions.service.ts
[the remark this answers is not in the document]
moisses89: Done 9bff7de84cb59f256094b48f1d8da232c1e29ac2

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +224,72 @@ export class TransactionsService {
...
+  async getTransactionHistory(
+    chainId: string,
+    routeUrl: Readonly<URL>,
+    safeAddress: string,
+    timezoneOffset?: string,
+    paginationData?: PaginationData,
+  ): Promise<Partial<TransactionItemPage>> {
+    const paginationData_adjusted = this.adjustTransactionHistoryPage(
+      paginationData?.limit,
+      paginationData?.offset,
```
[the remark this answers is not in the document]
moisses89: Done 3ed0214c5e8c5f8593e35c3295694f62ae20da5e

src/routes/transactions/transactions.controller.spec.ts
moisses89: yes I did it here 9bff7de84cb59f256094b48f1d8da232c1e29ac2, also for the other similar comments

src/routes/transactions/mappers/common/transaction.mapper.ts
moisses89: Take a look 83b9483823373e2c0541a968e0412c38f6d2c766

src/routes/transactions/transactions.controller.spec.ts
moisses89: Done 00777ace656c4400dd865c87065449d4c2a379b3

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +224,65 @@ export class TransactionsService {
...
+    paginationData: PaginationData,
+    timezoneOffset?: string,
+  ): Promise<Partial<TransactionItemPage>> {
+    const paginationDataAdjusted =
+      this.getAdjustedPaginationForHistory(paginationData);
+    const domainTransactions = await this.safeRepository.getTransactionHistory(
+      chainId,
+      safeAddress,
+      paginationDataAdjusted?.limit,
+      paginationDataAdjusted?.offset,
```
hectorgomezv, line 252: Nitpick: these values can't be null/undefined anymore, I think the optional-chaining is not required here.

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+    const transactionList = await Promise.all(
+      transactionsDomainGroups.map(async (transactionGroup) => {
+        const transactions: (TransactionItem | DateLabel)[] = [];
+        if (transactionGroup.timestamp != prevPageTimestamp) {
+          transactions.push(new DateLabel(transactionGroup.timestamp));
+        }
+        transactions.push(
+          ...(await this.mapGroupTransactions(transactionGroup, chainId, safe)),
+        );
+        return await Promise.all(transactions.flat());
```
hectorgomezv, line 65: I think all the promises should be resolved at this point (mappings are awaited in the line 63), so I think this could be simplified to `return transactions.flat()`

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -1075,6 +1088,371 @@ describe('Transactions Controller (Unit)', () => {
...
+          );
+        });
+    });
+
+    it('Should return correctly each transaction', async () => {
+      const chainId = faker.random.numeric();
+      const safeAddress = faker.finance.ethereumAddress();
+      const chainResponse = chainBuilder().with('chainId', chainId).build();
+      const moduleTransaction: any = getJsonResource(
+        'module-transactions/module-transaction-domain.json',
```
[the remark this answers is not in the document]
fmrsabino: > Because the purpose of this test is check that the result is correct for each field, if I use random fields we should check every field one by one, and in my opinion this is going to convert this test in a hardest test to read.

Not sure if I follow. You can check the entire payload at once instead of checking each field 🤔

src/routes/transactions/transactions.service.ts
moisses89: Done 21343d26f082811dfac619628436595613a509eb

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: Done 21343d26f082811dfac619628436595613a509eb

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+  private getPreviousDayTimestamp(
+    transactions: TransactionDomain[],
+    offset: number,
+    timezoneOffset?: string,
+  ): Date | undefined {
+    if (offset > 0) {
+      // Get previous page label
+      const timestamp =
+        transactions[0].executionDate ??
+        (transactions[0] as MultisigTransaction).submissionDate;
```
fmrsabino, line 81: I would consider using type predicates instead of explicit casting. Also we can reduce the nesting on this function.

Suggestion:

```typescript
private getPreviousPageDate(
  transactions: TransactionDomain[],
  offset: number,
  timezoneOffset?: string,
): Date | undefined {
  if (offset <= 0) return undefined;

  const previousPageTransaction = transactions[0];
  let timestamp: Date;
  if (isMultisigTransaction(previousPageTransaction)) {
    const executionDate = previousPageTransaction.executionDate;
    timestamp = executionDate ?? previousPageTransaction.submissionDate;
  } else if (isEthereumTransaction(previousPageTransaction)) {
    timestamp = previousPageTransaction.executionDate;
  } else if (isModuleTransaction(previousPageTransaction)) {
    timestamp = previousPageTransaction.executionDate;
  } else {
    throw Error('Unknown transaction type');
  }

  return this.getDayDate(timestamp, timezoneOffset);
}
```

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+  ): TransactionDomainGroup[] {
+    return Object.entries(
+      groupBy(transactions, (transaction) => {
+        let transactionTimestamp;
+        if (isMultisigTransaction(transaction)) {
+          transactionTimestamp =
+            transaction.executionDate ?? transaction.submissionDate;
+        } else {
+          transactionTimestamp = transaction.executionDate;
+        }
```
fmrsabino, line 102: Given the repeated pattern that I see regarding the type predicates I would extract this to a different function. Something like `getTransactionTimestamp(transaction: TransactionDomain): Date`

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+    ).map(
+      ([timestamp, transactions]) =>
+        <TransactionDomainGroup>{
+          timestamp: Number(timestamp),
+          transactions: transactions,
+        },
+    );
+  }
+
+  private getDayDate(timestamp: Date, timezoneOffset?: string): Date {
```
fmrsabino, line 114: Shouldn't the `timezoneOffset` be a `number | undefined`? We can tackle this at the route level. If we always try to convert it to a number we should let the route deal with that parsing (coming from the clients)

fmrsabino, line 114: nitpick: I think `getDayFromDate` represents better what the function does.

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+    return new Date(
+      Date.UTC(
+        timestamp.getFullYear(),
+        timestamp.getMonth(),
+        timestamp.getDate(),
+      ),
+    );
+  }
+
+  private mapEthereumTransfer(
```
fmrsabino, line 127: nitpick: `mapTransfer`. It's shorter and we don't use `Ethereum` to distinguish between other transfers.

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+              this.mapEthereumTransfer(transfers, chainId, safe),
+            );
+          }
+        } else {
+          // This should never happen as AJV would not allow an unknown transaction to get to this stage
+          throw Error('Unrecognized transaction type');
+        }
+      },
+    );
+    return Promise.all(transactions);
```
fmrsabino, line 180: Why do we wrap the `transactions` with a `Promise`? They should all be resolved or? 🤔

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
```diff
@@ -19,27 +19,25 @@ export class CreationTransactionMapper {
...
+      transaction.factoryAddress,
+    );
     const txInfo = new CreationTransactionInfo(
       creator,
-      transactionHash,
+      transaction.transactionHash,
       implementation,
       factory,
     );
+    const type = 'creation';
```
fmrsabino, line 38: This can be static.

src/routes/transactions/transactions.service.ts
fmrsabino: But this might be related with the `Page` entity itself. I don't think returning a `Partial` type to get around the fact that it has `count` is the way to go. We are adding technical debt this way.

The issue is that we should not have any `count` at all when serialising the `Page` to the clients. This should be solved separately and not by adding `Partial` to this route.

fmrsabino (changes requested):

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
moisses89: What you mean?

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+import { TransactionItem } from '../../entities/transaction-item.entity';
+import { DateLabel } from '../../entities/date-label.entity';
+import { Transfer } from '../../../../domain/safe/entities/transfer.entity';
+
+class TransactionDomainGroup {
+  timestamp: number;
+  transactions:
+    | MultisigTransaction[]
+    | ModuleTransaction[]
+    | EthereumTransaction[];
```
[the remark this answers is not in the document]
moisses89: Done 0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+    private readonly moduleTransactionMapper: ModuleTransactionMapper,
+    private readonly incomingTransferMapper: IncomingTransferMapper,
+  ) {}
+  async mapTransactions(
+    chainId: string,
+    transactionsDomain: TransactionDomain[],
+    safe: Safe,
+    offset: number,
+    timezoneOffset?: string,
+  ): Promise<[TransactionItem | DateLabel]> {
```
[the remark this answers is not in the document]
moisses89: Yes, it was my intention, done here 0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,182 @@
...
+          ...(await this.mapGroupTransactions(transactionGroup, chainId, safe)),
+        );
+        return await Promise.all(transactions.flat());
+      }),
+    );
+
+    return <[TransactionItem | DateLabel]>transactionList.flat();
+  }
+
+  private getPreviousDayTimestamp(
```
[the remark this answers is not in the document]
moisses89: Done 0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: Done 0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: Done 0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: Done 0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: I changed it to the map function, for some reason is not resolved, take a look 0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/transactions.service.ts
moisses89: Ok, I removed the Partial and add count. 
0dac6dc15d88b563a60ab7a8d7d8e6367360f35f

src/routes/transactions/transactions.controller.spec.ts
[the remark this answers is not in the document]
moisses89: If I create random domain transactions, the response for the service should be this transactions mapped to the route entities, then I can't check the response with the builder at once, or maybe I'm missing something.

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -168,15 +173,15 @@ export class TransactionsMapper {
...
           }
         } else {
           // This should never happen as AJV would not allow an unknown transaction to get to this stage
           throw Error('Unrecognized transaction type');
         }
-      },
+      }),
     );
-    return Promise.all(transactions);
+    return transactions;
```
fmrsabino, line 185: We don't need this function to be `async` we are just returning a promise:

```typescript
private mapGroupTransactions(
    transactionGroup,
    chainId,
    safe,
  ): Promise<TransactionItem[]> {
    return Promise.all(
      transactionGroup.transactions.map(async (transaction) => {
        ...
      }),
    );
```

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
fmrsabino: Right now this constant is part of the function stack. There's no need for it. We use it as global constant (static)

src/routes/transactions/mappers/common/transactions.mapper.ts
fmrsabino: Why does it need to be string from the clients?

moisses89: I mean on current client-gateway is optional string https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/transactions/routes.rs#L137

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -62,45 +63,53 @@ export class TransactionsMapper {
...
-        transactions[0].executionDate ??
-        (transactions[0] as MultisigTransaction).submissionDate;
-      if (timestamp !== null) {
-        return this.getDayDate(timestamp, timezoneOffset);
-      } else {
-        throw Error('ExecutionDate cannot be null');
-      }
-    }
+    if (offset <= 0) return undefined;
+    const previousPageTransaction = transactions[0];
```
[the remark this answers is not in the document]
moisses89: Yes, there was an issue with this case. Changed here aabf4def48fec1635bfd7765b9198795027e52d2

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: Done 070b3eaec76c5674f35d518d36a85f9af4e888cd

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,205 @@
...
+    ).map(
+      ([timestamp, transactions]) =>
+        <TransactionDomainGroup>{
+          timestamp: Number(timestamp),
+          transactions: transactions,
+        },
+    );
+  }
+
+  private getDayFromDate(timestamp: Date, timezoneOffset: number): Date {
```
hectorgomezv, line 134: I'd add documentation to this function. I think it's not trivial to know its expected output and the role of `timezoneOffset` on it 🙂

src/routes/transactions/mappers/common/transactions.mapper.ts
fmrsabino: That's not my point. Do we use it as a string? We need to evaluate these things when we rewrite the service. It shouldn't be a 1:1 copy besides respecting the API and internal business logic.

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
```diff
@@ -0,0 +1,48 @@
...
+import { Injectable } from '@nestjs/common';
+import { Safe } from '../../../../domain/safe/entities/safe.entity';
+import { Transaction } from '../../entities/transaction.entity';
+import { CreationTransaction } from '../../../../domain/safe/entities/creation-transaction.entity';
+import { TransactionStatus } from '../../entities/transaction-status.entity';
+import { CreationTransactionInfo } from '../../entities/creation-transaction-info.entity';
+
+@Injectable()
+export class CreationTransactionMapper {
+  static type = 'creation';
```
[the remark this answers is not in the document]
fmrsabino: I've suggested extracting here: https://github.com/5afe/safe-client-gateway-nest/pull/242#discussion_r1098529626

Creates more meaning towards it in my opinion – it's used as a prefix.

@moisses89 what is the reason for creating a type here?

src/routes/transactions/transactions.service.ts
```diff
@@ -218,6 +221,57 @@ export class TransactionsService {
...
+        await this.safeRepository.getCreationTransaction(chainId, safeAddress);
+      domainTransactions.results.push(creationTransaction);
+    }
+    const safeInfo = await this.safeRepository.getSafe(chainId, safeAddress);
+    const results = await this.transactionMapper.mapTransactions(
+      chainId,
+      domainTransactions.results,
+      safeInfo,
+      paginationData.offset,
+      parseInt(timezoneOffset ?? '0'),
```
hectorgomezv, line 264: In this case, instead of doing this parsing manually, I'd suggest parsing this at the controller level. I think the domain layer should receive a `number`. I mean, at the controller level you can parse by just defining the type as `number` or using Nest built-in `ParseIntPipe` (with a default value if needed), it's really simple to add: https://docs.nestjs.com/pipes
```javascript
@Query('timezone_offset', new DefaultValuePipe(0), ParseIntPipe) timezoneOffset: number,
```

hectorgomezv (changes requested):

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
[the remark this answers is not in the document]
moisses89: Do you mean the type name? Maybe could be better TAG or PREFIX.

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: Done 14bf055fda1872f7fabc8b04cf92820621f778f3

src/routes/transactions/transactions.service.ts
moisses89: Done

src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts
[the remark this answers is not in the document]
fmrsabino: It's not about the type name.

> what is the reason for creating a type here?

You added a type.

```
static type = 'creation';
       ^^^^
       ||||
       here
```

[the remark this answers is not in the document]
moisses89: Oh I see, this was changed here https://github.com/5afe/safe-client-gateway-nest/pull/242/files#diff-14ec1b4e83766d6a67bd7aaaa5fd07c0f54a0039e5e4e0874fada1abad5611d1R11. 
I didn't noticed that type is a reserved word, my mistake. This change is ok for this PR and it clarifies that `creation` is the transaction_type, but we should do the same in other PR for the other mappers.

src/routes/transactions/mappers/common/transactions.mapper.ts
```diff
@@ -0,0 +1,219 @@
...
+  transactions: (
+    | MultisigTransaction
+    | ModuleTransaction
+    | EthereumTransaction
+    | CreationTransaction
+  )[];
+}
+
+@Injectable()
+export class TransactionsMapper {
```
hectorgomezv, line 34: I'd say this mapper is not so generic, it contains specific logic coupled to the transactions history endpoint (`DateLabels`), so I'm thinking maybe this should be named something like `HistoryItemMapper`, `TransactionsHistoryItemMapper`, `TransactionsHistoryMapper`... and maybe it doesn't make a lot of sense to have it in the `common` folder. What do you think @fmrsabino @moisses89?

hectorgomezv (changes requested):

src/routes/transactions/mappers/common/transactions.mapper.ts
moisses89: Agree with the third one :) and move to mappers root.

moisses89: @hectorgomezv 6251e27dc680bce987ff0e6688054b33644c5805 what do you think?

hectorgomezv: Awesome, thanks 🚀

hectorgomezv (approved):

PR #243, opened 2023-01-25
Add get safe creation transaction

### LINKED ISSUE #241: Domain get safe creation transaction
# Description
Get transactions history endpoint add at the end the creation transaction for the safe because `all-transactions` endpoint on tx-service doesn't include this transaction. 
See the current implementation of the [client-gateway](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/transactions/handlers/history.rs#L62) 

# Proposal
Add safe domain method to return this transaction and add it at the end of the transactions history.

moisses89:
Closes #241 
- Add domain get safe creation transaction

# Additional notes
Couldn't keep alphabetic sort in [domain.module](https://github.com/5afe/safe-client-gateway-nest/compare/add_safe_creation_transaction?expand=1#diff-7d821304f2e0af7a603b971363f1bb70cb50afde8e2254719eb910536255ed27R71) because `creationTransactionValidator` contains a field that needs `dataDecodeValidator`.

### DIFF
--- a/src/domain/safe/entities/creation-transaction.entity.ts
+++ b/src/domain/safe/entities/creation-transaction.entity.ts
@@ -4,7 +4,7 @@ export interface CreationTransaction {
   created: Date;
   creator: string;
   transactionHash: string;
-  factoryAddress: string | null;
+  factoryAddress: string;
   masterCopy: string | null;
   setupData: string | null;
   dataDecoded: DataDecoded | null;
--- a/src/domain/safe/entities/creation-transaction.entity.ts
+++ b/src/domain/safe/entities/creation-transaction.entity.ts
@@ -1,0 +1,11 @@
+import { DataDecoded } from '../../data-decoder/entities/data-decoded.entity';
+
+export interface CreationTransaction {
+  created: Date;
+  creator: string;
+  transactionHash: string;
+  factoryAddress: string | null;
+  masterCopy: string | null;
+  setupData: string | null;
+  dataDecoded: DataDecoded | null;
+}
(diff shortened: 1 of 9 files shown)


src/domain/safe/entities/creation-transaction.entity.ts
```diff
@@ -0,0 +1,11 @@
+import { DataDecoded } from '../../data-decoder/entities/data-decoded.entity';
+
+export interface CreationTransaction {
+  created: Date;
+  creator: string;
+  transactionHash: string;
+  factoryAddress: string | null;
```
hectorgomezv, line 7: Is `factoryAddress` nullable? In Swagger it seems it's required (also `masterCopy` and `setupData`, but maybe that's a mistake), and [in the code](https://github.com/safe-global/safe-transaction-service/blob/d1a8aaa9e30902d1028998d109bff5d44be5218c/safe_transaction_service/history/services/safe_service.py#L40) it seems it's not optional 🤔

moisses89: You are right, I take from client-gateway https://github.com/safe-global/safe-client-gateway/blob/main/src/common/models/backend/transactions.rs#L97

moisses89: Done here https://github.com/5afe/safe-client-gateway-nest/pull/243/commits/9e42be09bd60b88af2d1e6a28398bc06f2360b00

hectorgomezv (approved):
🚀

PR #245, opened 2023-01-25
Use dataDecoded schema for MultisigTransaction

fmrsabino:
- `MultisigTransaction` now uses `DataDecoded` in its schema validation instead of `any`
- `SettingsChangeMapper` now uses the respective `DataDecoded` domain entities (instead of the route ones)
- `SettingsChangeMapper` now returns `null` instead of raising an error – this matches with the expected outcome of the mapper in the current Safe Client Gateway
- Fix `SettingsChangeMapper` issue where a `CHANGE_MASTER_COPY` would always return the value of the safe address as the master copy – consequentially, `safe` is no longer required as an argument for `mapSettingsChange`
- Sets the `SettingsChangeTransaction.settingsInfo` as an optional property (it can be `null`)
- Sets `CustomTransactionInfo.to` as required
- Adds support for nullable `DataDecoded` in `DataDecodedParamHelper`

### DIFF
--- a/src/routes/transactions/mappers/common/settings-change.mapper.ts
+++ b/src/routes/transactions/mappers/common/settings-change.mapper.ts
@@ -1,7 +1,6 @@
 import { Injectable } from '@nestjs/common';
 import { ModuleTransaction } from '../../../../domain/safe/entities/module-transaction.entity';
 import { MultisigTransaction } from '../../../../domain/safe/entities/multisig-transaction.entity';
-import { Safe } from '../../../../domain/safe/entities/safe.entity';
 import { AddressInfoHelper } from '../../../common/address-info/address-info.helper';
 import { NULL_ADDRESS } from '../../../common/constants';
 import { AddressInfo } from '../../../common/entities/address-info.entity';
@@ -9,7 +8,6 @@ import { AddOwner } from '../../entities/settings-changes/add-owner.entity';
 import { ChangeMasterCopy } from '../../entities/settings-changes/change-master-copy.entity';
 import { ChangeThreshold } from '../../entities/settings-changes/change-threshold.entity';
 import { DeleteGuard } from '../../entities/settings-changes/delete-guard';
-import { DisableModule } from '../../entities/settings-changes/disable-module.entity';
 import { EnableModule } from '../../entities/settings-changes/enable-module.entity';
 import { RemoveOwner } from '../../entities/settings-changes/remove-owner.entity';
 import { SetFallbackHandler } from '../../entities/settings-changes/set-fallback-handler.entity';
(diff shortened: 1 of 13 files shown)


fmrsabino:
Please review https://github.com/5afe/safe-client-gateway-nest/pull/246 first as it fixes the E2E tests.

src/routes/transactions/mappers/common/settings-change.mapper.ts
```diff
@@ -47,86 +47,183 @@ export class SettingsChangeMapper {
...
-          return new DeleteGuard();
-        }
-      }
+        return this.handleChangeThreshold(dataDecoded);
+      case SettingsChangeMapper.SET_GUARD:
+        return this.handleSetGuard(chainId, dataDecoded);
     }
-    throw new Error('Unknown setting');
+
+    return null;
```
hectorgomezv, line 227: This scenario shouldn't be possible, right? I mean, I'm assuming the `method` should be always contained in `SETTINGS_CHANGE_METHODS`, so if that assumption is right, if we reach this code something weird happened. Do you think it's better to return `null` in that case instead of explicitly throwing an error? What about at least including a message in the logs?

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -132,13 +151,17 @@ export class MultisigTransactionInfoMapper {
     value: number,
     dataSize: number,
   ): boolean {
+    const isSettingsChange: boolean = transaction.dataDecoded
```
hectorgomezv, line 154: nitpick: what about renaming this `isSettingsChangeMethod` to avoid colliding the containing function name?

hectorgomezv (approved):
Great job! I just left minor some comments 🙂 🚀

src/routes/transactions/mappers/common/settings-change.mapper.ts
fmrsabino: Yes this should not be possible if we consider that we check that the method is a settings change beforehand.

The current CGW implementation actually returns `null` here but I think that can hide some issues. Will revert this change.

fmrsabino: But it does return `null` if the method name is known and matches but the arguments don't match.

fmrsabino: Changed in [7ed250e](https://github.com/5afe/safe-client-gateway-nest/pull/245/commits/7ed250ebe5aefdbd2a26d74cf769acd303da7e77)

src/routes/transactions/mappers/common/transaction-info.mapper.ts
fmrsabino: Changed in [7ed250e](https://github.com/5afe/safe-client-gateway-nest/pull/245/commits/7ed250ebe5aefdbd2a26d74cf769acd303da7e77)

hectorgomezv (dismissed):
🚀

hectorgomezv (approved):

PR #246, opened 2023-01-25
Fix DataDecodedParameter property name

fmrsabino:
- `DataDecodedParameter.paramType` is now `DataDecodedParameter.type` as it should be according to the specification


src/routes/data-decode/data-decoded.service.ts
```diff
@@ -31,9 +32,16 @@ export class DataDecodedService {
       data,
       to,
     );
-    const parameters = dataDecoded.parameters?.map(
-      ({ type: paramType, ...rest }) => ({ ...rest, paramType }),
-    );
-    return new DataDecoded(dataDecoded.method, parameters ?? null);
```
fmrsabino, line 37: ⚠️ FYI for how the spread operator left this issue go unnoticed until now.

hectorgomezv (approved):

PR #263, opened 2023-01-27
Add default values to PaginationData

hectorgomezv:
This PR aims to establish default values for pagination data.

Previously, the service was returning `undefined` when no valid values were passed to `PaginationData`. The current Rust implementation does use default values for this use case since default values for `next` URLs are returned to the clients when no valid cursor is sent: see [the Rust implementation](https://github.com/safe-global/safe-client-gateway/blob/ad6e9f132e804fbacbbecbc878cb8b7efe4d2dc1/src/common/converters/page_metadata.rs).

Some tests were adjusted to match the default `limit` and `offset` values.

### DIFF
--- a/src/routes/common/pagination/pagination.data.ts
+++ b/src/routes/common/pagination/pagination.data.ts
@@ -8,7 +8,7 @@ export class PaginationData {
    * These values are intended to apply if no pagination data is present in the client request.
    * They limit the amount of items retrieved from provider services, and abstracts the clients
    * from pagination data management.
-   * 
+   *
    * Relying on pagination values returned by provider services is not feasible, since the
    * provider pagination data could have been customized depending on the use case.
    */
--- a/src/routes/common/pagination/pagination.data.ts
+++ b/src/routes/common/pagination/pagination.data.ts
@@ -3,6 +3,15 @@ const QUERY_PARAM_LIMIT = 'limit';
 const QUERY_PARAM_OFFSET = 'offset';
 
 export class PaginationData {
+  /**
+   * Default pagination (limit and offset) values.
+   * These values are intended to apply if no pagination data is present in the client request.
+   * They limit the amount of items retrieved from provider services, and abstracts the clients
+   * from pagination data management.
+   * 
+   * Relying on pagination values returned by provider services is not feasible, since the
+   * provider pagination data could have been customized depending on the use case.
+   */
   public static readonly DEFAULT_LIMIT = 20;
   public static readonly DEFAULT_OFFSET = 0;
 
--- a/src/routes/common/pagination/pagination.data.ts
+++ b/src/routes/common/pagination/pagination.data.ts
@@ -3,7 +3,13 @@ const QUERY_PARAM_LIMIT = 'limit';
 const QUERY_PARAM_OFFSET = 'offset';
 
 export class PaginationData {
-  constructor(readonly limit?: number, readonly offset?: number) {}
+  public static readonly DEFAULT_LIMIT = 20;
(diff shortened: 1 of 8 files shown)


moisses89 (dismissed):
Looks good for me!

src/routes/common/pagination/pagination.data.ts
```diff
@@ -3,7 +3,13 @@ const QUERY_PARAM_LIMIT = 'limit';
 const QUERY_PARAM_OFFSET = 'offset';
 
 export class PaginationData {
-  constructor(readonly limit?: number, readonly offset?: number) {}
+  public static readonly DEFAULT_LIMIT = 20;
+  public static readonly DEFAULT_OFFSET = 0;
```
fmrsabino, line 7: I'm a bit wary on adding this. This would set all of the pagination to these values. If the default is adjusted on any other service (eg.: like the Safe Transaction Service) then we'd not use the default of the target service but instead would use the CGW default.

What are the use cases where you need a default set?

fmrsabino (changes requested):

src/routes/common/pagination/pagination.data.ts
fmrsabino: I've just synced with Richard regarding this decision. I'll write an update here soon.

hectorgomezv: I share your concerns about having a _decentralised_ configuration between this service and the Transaction Service, in fact I commented with @moisses89 (since his PR is also affected by this) that concern. 

The reason for defining a default pagination it's motivated because we need to build _next_ attribute in the response even when the client is not sending any _cursor_, and as I commented before we can't rely on the Transaction Service values since the page limits requested to that service has an __adjusted_/_shifted_ pagination to build the view the clients need. 

So if this duplicated configuration param is not acceptable, I've some other options:
- Re-adjust the response pagination values coming from the Transaction Service (I think we'd need to do this per-endpoint, so it could be a bit error-prone and harder to maintain).
- Share the configuration values somewhere (maybe we could just take the Transaction Service values as the source of truth and try to sync on the service startup, or even use the Config Service for that).

> I've just synced with Richard regarding this decision. I'll write an update here soon.

Great, thanks! 🙂

fmrsabino: So this decision was made so that we always limit the number of returned items from the other services as one item can be multiplexed into multiple requests (eg.: 1 transaction is multiplexed into balance + fiat requests).

Therefore the default is used so:
1) we limit the number of items that we want to retrieve from the other services
2) the clients shouldn't be concerned about pagination in general – that's why the cursor is "encoded" the way it is. The contract with the clients is that the cursor should not be adjusted at all so every request starts with the default pagination and then we build subsequent ones with `next`/`previous`

I have concerns regarding this design decision but this is the _contract_ that we have in place.

So the PR is good to go 🚀 I'll just request one thing: can we create a comment where the defaults are with the main reasons why we have defaults? (this was not done before, it should be done now).

fmrsabino (approved):

src/routes/common/pagination/pagination.data.ts
hectorgomezv: @fmrsabino @moisses89 I added a documentation comment to the default values in [a7466d1](https://github.com/5afe/safe-client-gateway-nest/pull/263/commits/a7466d16ef912c723606bba145d14ab9825cfc78). Please ping me if you feel it's not clear enough or if you want to add something 🙂

fmrsabino (approved):
🚀

moisses89 (approved):

PR #265, opened 2023-02-01
Add ids to AJV schemas

### LINKED ISSUE #264: Improve cross-schema reference in AJV
We should improve on how to address schemas from different schemas. One potential improvement is to associate an `id` with each schema with a known format.

This `id` can then be address by schemas on different features. One advantage is that we can check if the schema was already added before compiling and adding it again which should result in better performance.

fmrsabino:
Closes #264 

- Adds an `$id` to each AJV schema – the id allows combining schemas more easily using the reference that was set
  * The `id` should be unique – it is resolved as a URI reference therefore the base url was set to `https://safe-client.safe.global/schemas` – this can be any valid URI.
  * The URI does not mean that the schemas are available under that path – it is just used for addressing different schemas using a hierarchy set in the URI path
- Adds  `getSchema` to `JsonSchemaService`
  * It takes a reference key (`id` for the schema) and tries to retrieve it. If the schema is not available it gets compiled and added to the internal AJV instance cache
  * The benefit of this approach is that common schemas (like `DataDecoded`) are now compiled only once and will get reused across the application when required.
  * With this addition, `addSchema` and `compile` were removed from the `JsonSchemaService` – consumers of this API should only be concerned about getting the validation function (and not checking if a schema compilation is required beforehand)
- Removes tests associated with the validators
   * These tests are already covered by the controllers and upper layers. An invalid schema change should be caught by either AJV or the upper layers


hectorgomezv:
Great job 🚀 I think this simplifies our code and makes `JsonSchemaService` API consumers more abstracted from its internals. I was also reviewing if there is a way to resolve a given schema's dependent schemas (`$ref`) in an automated way, without forcing the client to explicitly call `getSchema` for each dependency. That would make the API even easier to use, but I couldn't find a simple way.

hectorgomezv (approved):

fmrsabino:
> Great job 🚀 I think this simplifies our code and makes `JsonSchemaService` API consumers more abstracted from its internals. I was also reviewing if there is a way to resolve a given schema's dependent schemas (`$ref`) in an automated way, without forcing the client to explicitly call `getSchema` for each dependency. That would make the API even easier to use, but I couldn't find a simple way.

I thought about creating an abstraction in the format of a tuple like `(id, schema)` – with this you could create a tree of dependent schemas and navigate this tree so that each schema is added automatically. I think this was going to add complexity and decided not to do it for this PR. The goal was more to improve how we reference each schema.

An alternative is to create a validator per schema and list in the constructor the validators required (eg.: collection of validators). Then we can have a for loop which would go through each validator and would "add" the schema. In other words you'd have a 1:1 mapping between schema and validator. You'd still need to list the validators on which you'd depend though...

hectorgomezv:
> > Great job 🚀 I think this simplifies our code and makes `JsonSchemaService` API consumers more abstracted from its internals. I was also reviewing if there is a way to resolve a given schema's dependent schemas (`$ref`) in an automated way, without forcing the client to explicitly call `getSchema` for each dependency. That would make the API even easier to use, but I couldn't find a simple way.
> 
> I thought about creating an abstraction in the format of a tuple like `(id, schema)` – with this you could create a tree of dependent schemas and navigate this tree so that each schema is added automatically. I think this was going to add complexity and decided not to do it for this PR. The goal was more to improve how we reference each schema.
> 
> An alternative is to create a validator per schema and list in the constructor the validators required (eg.: collection of validators). Then we can have a for loop which would go through each validator and would "add" the schema. In other words you'd have a 1:1 mapping between schema and validator. You'd still need to list the validators on which you'd depend though...

Yes, I think the dependency tree could be a great solution if we had a lot of sub-schemas on several levels, but as you said for the current number of schemas it may add more complexity instead of removing it. 👍

PR #267, opened 2023-02-02
Implement POST v1/register/notifications

### LINKED ISSUE #250: Add POST /v1/register/notifications

hectorgomezv:
Closes #250 

- Add the `POST` endpoint `v1/register/notifications`.
- This route contains some custom logic regarding error handling, which was taken from the current Client Gateway Rust implementation and has been documented in `NotificationsService`.
- A datasource method was implemented to satisfy the endpoint.
- Some entities and testing object builders were introduced (`Device`, `SafeRegistration`, etc.)
- Unit (controller) tests were added.


src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+
+    app = moduleFixture.createNestApplication();
+    app.useGlobalFilters(new DataSourceErrorFilter());
+
+    await app.init();
+  });
+
+  describe('POST /register/notifications', () => {
+    it('Success', async () => {
+      const chains: Chain[] = range(4).map((idx) =>
```
fmrsabino, line 61: nitpick: `chainId` instead of `idx`

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+      );
+      const safeRegistrations: SafeRegistration[] = chains.map((chain) =>
+        safeRegistrationBuilder().with('chain_id', chain.chainId).build(),
+      );
+      const registerDeviceDto = registerDeviceDtoBuilder()
+        .with('safeRegistration', safeRegistrations)
+        .build();
+      mockNetworkService.get.mockImplementation((url) => {
+        const getChainUrlPattern = `${safeConfigUrl}/api/v1/chains/`;
+        if (url.includes(getChainUrlPattern)) {
```
fmrsabino, line 72: Why don't we match with the expected url?

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+          return Promise.resolve({ data: sample(chains) });
+        }
+        return Promise.reject(`No matching rule for url: ${url}`);
+      });
+      mockNetworkService.post.mockImplementation(() => Promise.resolve());
```
fmrsabino, line 77: We should match against a URL so that we don't consider "any" `post` call.

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+
+      await request(app.getHttpServer())
+        .post('/register/notifications')
+        .send(registerDeviceDto)
+        .expect(200);
```
fmrsabino, line 82: Should the body be empty?

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+        faker.datatype.number({ min: 400, max: 499 }),
+      );
+      mockNetworkService.post
+        .mockImplementationOnce(() => Promise.reject(error))
+        .mockImplementation(() => Promise.resolve());
+
+      await request(app.getHttpServer())
+        .post('/register/notifications')
+        .send(registerDeviceDto)
+        .expect(400);
```
fmrsabino, line 113: What's the expectation of the body of the message in this case?

(same for the other tests)

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -0,0 +1,243 @@
...
+          return Promise.resolve({ data: sample(chains) });
+        }
+        return Promise.reject(`No matching rule for url: ${url}`);
+      });
+      // return redirection and generic errors for two providers
+      const redirection = new NetworkResponseError(
+        faker.datatype.number({ min: 300, max: 399 }),
+      );
+      mockNetworkService.post
+        .mockImplementationOnce(() => Promise.reject(redirection))
```
fmrsabino, line 233: Why is a redirection an error?

src/routes/notifications/notifications.service.ts
```diff
@@ -0,0 +1,88 @@
...
+import { SafeRegistration } from '../../domain/notifications/entities/safe-registration.entity';
+import { NotificationsRepository } from '../../domain/notifications/notifications.repository';
+import { INotificationsRepository } from '../../domain/notifications/notifications.repository.interface';
+import { RegisterDeviceDto } from './entities/register-device.dto.entity';
+
+@Injectable()
+export class NotificationsService {
+  constructor(
+    @Inject(INotificationsRepository)
+    private readonly notificationsRepository: NotificationsRepository,
```
fmrsabino, line 18: Use the interface here: `INotificationsRepository`

fmrsabino (changes requested):
Nice! Some small comments.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -1,24 +1,25 @@
...
+import { Safe } from '../../domain/safe/entities/safe.entity';
 import { Transaction } from '../../domain/safe/entities/transaction.entity';
+import { Transfer } from '../../domain/safe/entities/transfer.entity';
 import { Token } from '../../domain/tokens/entities/token.entity';
-import { ModuleTransaction } from '../../domain/safe/entities/module-transaction.entity';
-import { CreationTransaction } from '../../domain/safe/entities/creation-transaction.entity';
+import { CacheFirstDataSource } from '../cache/cache.first.data.source';
+import { ICacheService } from '../cache/cache.service.interface';
+import { HttpErrorFactory } from '../errors/http-error-factory';
+import { INetworkService } from '../network/network.service.interface';
```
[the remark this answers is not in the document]
hectorgomezv: This was changed automatically by the IDE when optimizing imports. I agree this shouldn't be part of the PR, I will revert it 👍

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: According to the Rust implementation, it should be empty when the device registration is successful. But when errors from the provider are caught, it should be indicated in the body response, I will change it 👍

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: True, the body in case of error is missing, I will change this part, thanks!

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: True, in this case, it should be indicated in the client response, I will change this as well 👍

src/datasources/transaction-api/transaction-api.service.ts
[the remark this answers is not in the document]
fmrsabino: My point was more that we should have some deterministic way to sort the imports. I need to check which rules the formatter supports.

src/routes/notifications/notifications.controller.spec.ts
fmrsabino: > According to the Rust implementation, it should be empty when the device registration is successful. But when errors from the provider are caught, it should be indicated in the body response, I will change it 👍

Not sure if I understood: what will you change given that the status code is 200?

My suggestion was more leaning towards validating the body in the response.

src/datasources/transaction-api/transaction-api.service.ts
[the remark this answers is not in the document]
hectorgomezv: Yes, in my case I think it's sorting alphabetically by the file route, but of course, we need to unify this and configure that explicitly if possible in the linter.

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: Sorry I explained it poorly, I was meant I'm going to change it in the case of errors. In this case the body should be empty so it's OK I guess.

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: In this case it's expected to make a call against the Config Service for each chain in `chains`, something like:
```
mockNetworkService.get.mockImplementation((url) => {
        if (
          chains.some(
            (chain) =>
              url === `${safeConfigUrl}/api/v1/chains/${chain.chainId}`,
          )
        ) {
          return Promise.resolve({ data: sample(chains) });
        }
        return Promise.reject(`No matching rule for url: ${url}`);
      });
```

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: In this case we also have to check the URL matches against any of the `transactionService` properties of the `chains`. Something like:
```javascript
mockNetworkService.post.mockImplementation((url) => {
  if (
    chains.some(
      (chain) =>
        url ===
        `${chain.transactionService}/api/v1/notifications/devices/`,
    )
  ) {
    return Promise.resolve();
  }
  return Promise.reject(`No matching rule for url: ${url}`);
});
```

Please ping me if you see a simpler solution 🙂

src/routes/notifications/notifications.controller.spec.ts
fmrsabino: > Sorry I explained it poorly, I was meant I'm going to change it in the case of errors. In this case the body should be empty so it's OK I guess.

If the implementation returns a body (which is not expected) then this would still pass right?

src/datasources/transaction-api/transaction-api.service.ts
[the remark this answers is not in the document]
hectorgomezv: Changed in [5bf56a7](https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/5bf56a7f8f6e978eaa8428dc0bf8ac3df335c6cb)

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: Changed in [5bf56a7](https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/5bf56a7f8f6e978eaa8428dc0bf8ac3df335c6cb)

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: Changed the approach in https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/0728521aa558fa73c90498d9c13798e313936c93

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/0728521aa558fa73c90498d9c13798e313936c93

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: I changed the approach in https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/0728521aa558fa73c90498d9c13798e313936c93 

Now the test checks the body is empty in this scenario.

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: I changed the approach in https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/0728521aa558fa73c90498d9c13798e313936c93 

Now the expected body is checked. From [the current Rust implementation](https://github.com/safe-global/safe-client-gateway/blob/ad6e9f132e804fbacbbecbc878cb8b7efe4d2dc1/src/routes/notifications/handlers.rs#L90), I took the `message` only. Do you think the `code`, `arguments`, and `debug` parts are necessary as well? I remember this being discussed to not expose this `debug` data in the service responses, what do you think @fmrsabino?

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: I removed the redirection from the error scenarios in [0728521](https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/0728521aa558fa73c90498d9c13798e313936c93)

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -68,23 +68,39 @@ describe('Notifications Controller (Unit)', () => {
...
       mockNetworkService.get.mockImplementation((url) => {
-        const getChainUrlPattern = `${safeConfigUrl}/api/v1/chains/`;
-        if (url.includes(getChainUrlPattern)) {
+        if (
+          chains.some(
+            (chain) =>
+              url === `${safeConfigUrl}/api/v1/chains/${chain.chainId}`,
+          )
+        ) {
           return Promise.resolve({ data: sample(chains) });
```
fmrsabino, line 77: I think using `sample` can potentially cause issues right? We might not match the chain returned. 

You have a different implementation using `find` in this PR which I think is better wdyt?

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -68,23 +68,39 @@ describe('Notifications Controller (Unit)', () => {
...
+        }
+        return Promise.reject(`No matching rule for url: ${url}`);
+      });
 
       await request(app.getHttpServer())
         .post('/register/notifications')
         .send(registerDeviceDto)
-        .expect(200);
+        .expect(200)
+        .then(({ body }) => expect(body).toEqual({}));
```
fmrsabino, line 98: You can call `expect` since it takes a `body` directly.

```suggestion
        .expect({});
```

same for the other tests.

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -93,29 +109,47 @@ describe('Notifications Controller (Unit)', () => {
...
+          expect(body).toMatchObject({
+            message: `Push notification registration failed for chain IDs: ${errorChainIds}`,
+          }),
+        );
     });
 
     it('Server errors returned from provider', async () => {
-      const chains: Chain[] = range(4).map((idx) =>
-        chainBuilder().with('chainId', idx.toString()).build(),
+      const chains: Chain[] = range(21, 30).map((chainId) =>
```
fmrsabino, line 151: OOC: Why do we increase the range on each test?

src/routes/notifications/notifications.controller.spec.ts
```diff
@@ -124,29 +158,47 @@ describe('Notifications Controller (Unit)', () => {
...
         .build();
       mockNetworkService.get.mockImplementation((url) => {
-        const getChainUrlPattern = `${safeConfigUrl}/api/v1/chains/`;
-        if (url.includes(getChainUrlPattern)) {
-          return Promise.resolve({ data: sample(chains) });
+        const targetChain = chains.find(
+          (chain) => url === `${safeConfigUrl}/api/v1/chains/${chain.chainId}`,
+        );
+        if (targetChain) {
+          return Promise.resolve({ data: targetChain });
```
fmrsabino, line 165: I think that this might be overcomplicating the tests a little bit. Doing it once to get a random number of chains is fine if that should be really be part of the test.

For most tests you can have a set number of chains (eg.: 2 or 3) and map the chain id directly with the expected payload.

You can leave a test that leaves some randomness to the amount of items in the array but I don't think it's necessary and makes the test harder to read.

src/routes/notifications/notifications.controller.spec.ts
fmrsabino: See my comment [here](https://github.com/5afe/safe-client-gateway-nest/pull/267#discussion_r1098651611) instead

src/routes/notifications/notifications.controller.spec.ts
fmrsabino: Let's keep it as you have right now. A change at this level would also impact other error payloads.

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: Changed in [48a179c](https://github.com/5afe/safe-client-gateway-nest/pull/267/commits/48a179c6ee1ea8bd2c04c00e567f9a2b97b71aa0)

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: The problem I've been facing with this is that `TransactionApiManager` is not stateless, because it has this internal state:
```javascript
private transactionApiMap: Record<string, TransactionApi> = {};
```
So not every test having N chains will use `mockNetworkService.get` N times, because if the internal state of `TransactionApiManager` already contains that chain, the mock function doesn't get called. So this makes the value of `chain.transactionService` not controlled and therefore the URL for the subsequent POST call wouldn't match.

This problem is a bit tricky to bypass and that's something that it's overcomplicating the tests as you noted on [this comment](https://github.com/5afe/safe-client-gateway-nest/pull/267#discussion_r1098651611). I see some mitigations but I'm open to suggestions:

- Move the internal state of `TransactionApiManager` from memory to redis: this solution could be useful but is out of the scope of this PR I guess.
- Implement a function to clear the internal state of `TransactionApiManager` and call that function on `beforeEach` hook: not a fan of this one since we would be changing production code due to testing implementation requirements.
- Remove the numerical ranges and use uuids for `chainIds` in these tests: probably the easier but tests will remain verbose and pretty difficult to read and follow.
- Make an exception and rely on patterns instead of strict URL matching when implementing mock functions: I think we could go this way.

What do you think @fmrsabino?

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: Please refer to [this comment](https://github.com/5afe/safe-client-gateway-nest/pull/267#discussion_r1098784500).

src/routes/notifications/notifications.controller.spec.ts
hectorgomezv: I'll give this a second thought and I'll try to come up with a simpler solution @fmrsabino.

fmrsabino: But the `beforeEach` should give you a clean state of the application or that is not happening?

hectorgomezv: Ok, I found the reason, the app was correctly initialized on each test run, the problem I had was the test wasn't clearing the cache, I forgot to include:
```javascript
fakeCacheService.clear();
```
in the `beforeEach` hook 🤦🏻‍♂️

Now the previous implementation with the strict URL matching and without the need to use different ranges for `chainId` in each test works. But I'm pretty happy with the current implementation using URL patterns TBH. What do you think?

fmrsabino (approved):

src/routes/notifications/notifications.controller.spec.ts
fmrsabino: > Ok, I found the reason, the app was correctly initialized on each test run, the problem I had was the test wasn't clearing the cache, I forgot to include:
> 
> ```js
> fakeCacheService.clear();
> ```
> 
> in the `beforeEach` hook 🤦🏻‍♂️

That's strange because, as you mentioned before, if the `TransactionApiManager` was the root cause of the issue, it doesn't use Redis. It's just a local `Record`.

The `fakeCacheService` replaces Redis (as a fake implementation).

> Now the previous implementation with the strict URL matching and without the need to use different ranges for `chainId` in each test works. But I'm pretty happy with the current implementation using URL patterns TBH. What do you think?

I think it's a nice improvement and the tests are more readable in my opinion! In this case it works – for some other tests it could also work but some of them test that the path is called with some specific parameters (safe address, transaction hash, etc.) – there's also value in knowing and having the test fail if the URL does not match.

hectorgomezv: > That's strange because, as you mentioned before, if the `TransactionApiManager` was the root cause of the issue, it doesn't use Redis. It's just a local `Record`.

I assumed the problem was in `TransactionApiManager` but it wasn't, my fault, the problem was in the deeper layer because the cache was not being cleared 😄

fmrsabino: > > That's strange because, as you mentioned before, if the `TransactionApiManager` was the root cause of the issue, it doesn't use Redis. It's just a local `Record`.
> 
> I assumed the problem was in `TransactionApiManager` but it wasn't, my fault, the problem was in the deeper layer because the cache was not being cleared 😄

Got it 🙌

PR #270, opened 2023-02-10
Implement DELETE /v1/chains/:chainId/safes/:safeAddress/delegates/:delegateAddress

### LINKED ISSUE #260: Add DELETE /v1/chains/<chain_id>/safes/<safe_address>/delegates/<delegate_address>

hectorgomezv:
Closes #260 

Add the DELETE endpoint `/v1/chains/:chainId/safes/:safeAddress/delegates/:delegateAddress`

- All unit (controller) tests in `src/routes/delegates/delegates.controller.spec.ts` were refactored to use strict URL matching.
- Both  `create-delegate.dto.factory.ts` and `delete-delegate.dto.factory.ts` were replaced by builders: `create-delegate-dto.builder.ts`, `delete-delegate-dto.builder.ts`.
- Unit (controller) tests were added.


src/routes/delegates/delegates.controller.ts
```diff
@@ -91,4 +92,22 @@ export class DelegatesController {
...
+  @Delete('chains/:chainId/safes/:safeAddress/delegates/:delegateAddress')
+  async deleteSafeDelegate(
+    @Param('chainId') chainId: string,
+    @Param('safeAddress') safeAddress: string,
+    @Param('delegateAddress') delegateAddress: string,
+    @Body() deleteSafeDelegateRequest: DeleteSafeDelegateDto,
+  ): Promise<unknown> {
+    if (safeAddress !== deleteSafeDelegateRequest.safe) {
+      throw new BadRequestException();
+    }
```
fmrsabino, line 105: Is this a new check (compared to the old CGW)?

src/routes/delegates/delegates.service.ts
```diff
@@ -87,4 +94,21 @@ export class DelegatesService {
...
   }
+
+  async deleteSafeDelegate(
+    chainId: string,
+    delegateAddress: string,
+    deleteSafeDelegateRequest: DeleteSafeDelegateDto,
+  ): Promise<unknown> {
+    if (delegateAddress !== deleteSafeDelegateRequest.delegate) {
+      throw new BadRequestException();
+    }
```
fmrsabino, line 105: Why do we repeat the check here?

src/routes/delegates/entities/__tests__/delete-delegate-dto.builder.ts
```diff
@@ -0,0 +1,10 @@
+import { faker } from '@faker-js/faker';
+import { Builder, IBuilder } from '../../../../__tests__/builder';
+import { CreateDelegateDto } from '../create-delegate.entity';
+
+export function deleteDelegateDtoBuilder(): IBuilder<CreateDelegateDto> {
```
fmrsabino, line 5: Shouldn't this be a builder for `DeleteDelegateDto`?

src/routes/delegates/delegates.service.ts
hectorgomezv: This one is for checking the delegate address, the one in the controller is checking the safe address, but anyway, it makes no sense to have validation in both places, I'll change it 👍

src/routes/delegates/delegates.controller.ts
hectorgomezv: Yes, this is a new check, sorry, I forgot to comment on this explicitly in the PR.

In the current CGW specification, both the Safe address and the delegate address are present in two places: in the URL and in the body request. Since we are sending them also in two places (see [the network call in TransactionApiService](https://github.com/5afe/safe-client-gateway-nest/blob/2953e1e713197878f94a45bc13efc0590fe68cfd/src/datasources/transaction-api/transaction-api.service.ts#L224)), I think it could be a good practice to validate these values are equal between URL and request body. Do you think this validation check is excessive? (I can't foresee a scenario where the clients may want to have different values between the URL and request body)

src/routes/delegates/entities/__tests__/delete-delegate-dto.builder.ts
hectorgomezv: Of course! This was a mistake (properties of both classes overlap so the compiler didn't complaint 😄). I will fix it!

src/routes/delegates/delegates.controller.ts
fmrsabino: My concern is that this might change what the clients are expecting. Did you check with the clients this behaviour?

My suggestion is to keep it out for now but keep it in mind (or even have an issue) to evaluate adding these checks. It might be worth reviewing in other routes too with the clients.

hectorgomezv: Ok, maybe it's better to remove these validation checks. I didn't check with the clients yet but having an issue reviewing this sounds good. Then we can include these kinds of checks, or even check if having the same data in several parts of the requests make sense at all.

fmrsabino (commented):
You might need to adjust `src/routes/delegates/delegates.controller.spec.ts` with the introduction of the following: https://github.com/5afe/safe-client-gateway-nest/commit/abdf01cfe2ca7e494617f8f300adacc66a3e48a1

src/routes/delegates/entities/__tests__/delete-delegate-dto.builder.ts
hectorgomezv: Fixed in [74dde61](https://github.com/5afe/safe-client-gateway-nest/pull/270/commits/74dde610ba8bbbdcb48c03380f2c23a1a6c9f9fc)

src/routes/delegates/delegates.service.ts
hectorgomezv: Removed in [74dde61](https://github.com/5afe/safe-client-gateway-nest/pull/270/commits/74dde610ba8bbbdcb48c03380f2c23a1a6c9f9fc)

src/routes/delegates/delegates.controller.ts
hectorgomezv: Removed in [74dde61](https://github.com/5afe/safe-client-gateway-nest/pull/270/commits/74dde610ba8bbbdcb48c03380f2c23a1a6c9f9fc)

fmrsabino (approved):
🚀

PR #271, opened 2023-02-14
Validate delete delegate route body

hectorgomezv:
This PR implements DTO validation using [Nest Pipes](https://docs.nestjs.com/pipes).

Failing-fast, when a malformed request is sent, has some benefits, the most obvious is to avoid calling backend services (Config Service, Transaction Service) in those situations, as well as saving resources in the Client Gateway itself. This is augmented by the fact that failed requests are not cached by default, so a client sending recurrent malformed bodies could be computationally expensive for the whole architecture.

To tackle this, elemental request body validation could be implemented in the Client gateway following several strategies:
- Using [auto-validation](https://docs.nestjs.com/techniques/validation#auto-validation) by attaching a `ValidationPipe` directly to the `app` object: `app.useGlobalPipes(new ValidationPipe());`: this will make all the `@Body()` annotations automatically validated, by using `class-validator` package and mandatory annotations in all DTOs.
- Use custom a `ValidationPipe` for each DTO (as we did previously [here](https://github.com/5afe/safe-client-gateway-nest/blob/2aa248b875a4a8fe1b6ebe189d70a01cd0fe83e3/src/routes/cache-hooks/cache-hooks.controller.ts#L20)). 

This PR goes through the second approach. For this, on [Nest docs](https://docs.nestjs.com/pipes#object-schema-validation) Joi is used, but we have AJV in place so we can reuse some code by using AJV (as it's used for [EventValidationPipe](https://github.com/5afe/safe-client-gateway-nest/blob/2aa248b875a4a8fe1b6ebe189d70a01cd0fe83e3/src/routes/cache-hooks/pipes/event-validation.pipe.ts#L12))

### DIFF
--- a/src/routes/delegates/delegates.module.ts
+++ b/src/routes/delegates/delegates.module.ts
@@ -1,9 +1,17 @@
 import { Module } from '@nestjs/common';
+import { GenericValidator } from '../../domain/schema/generic.validator';
+import { JsonSchemaService } from '../../domain/schema/json-schema.service';
+import { ValidationErrorFactory } from '../../domain/schema/validation-error-factory';
 import { DelegatesController } from './delegates.controller';
 import { DelegatesService } from './delegates.service';
 
 @Module({
   controllers: [DelegatesController],
-  providers: [DelegatesService],
+  providers: [
+    JsonSchemaService,
+    GenericValidator,
+    ValidationErrorFactory,
+    DelegatesService,
+  ],
 })
 export class DelegatesModule {}
--- a/src/routes/delegates/delegates.module.ts
+++ b/src/routes/delegates/delegates.module.ts
@@ -1,17 +1,11 @@
 import { Module } from '@nestjs/common';
-import { GenericValidator } from '../../domain/schema/generic.validator';
-import { JsonSchemaService } from '../../domain/schema/json-schema.service';
-import { ValidationErrorFactory } from '../../domain/schema/validation-error-factory';
+import { ValidationModule } from '../../validation.module';
 import { DelegatesController } from './delegates.controller';
 import { DelegatesService } from './delegates.service';
 
 @Module({
   controllers: [DelegatesController],
-  providers: [
-    JsonSchemaService,
-    GenericValidator,
-    ValidationErrorFactory,
-    DelegatesService,
-  ],
+  providers: [DelegatesService],
+  imports: [ValidationModule],
 })
 export class DelegatesModule {}
--- a/src/routes/delegates/pipes/delete-delegate-dto-validation.pipe.ts
+++ b/src/routes/delegates/pipes/delete-delegate-dto-validation.pipe.ts
@@ -1,0 +1,30 @@
+import { HttpStatus, Injectable, PipeTransform } from '@nestjs/common';
+import { ValidateFunction } from 'ajv';
+import { GenericValidator } from '../../../domain/schema/generic.validator';
+import { JsonSchemaService } from '../../../domain/schema/json-schema.service';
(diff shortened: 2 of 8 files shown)


src/routes/delegates/delegates.module.ts
```diff
@@ -1,9 +1,17 @@
...
 import { DelegatesService } from './delegates.service';
 
 @Module({
   controllers: [DelegatesController],
-  providers: [DelegatesService],
+  providers: [
+    JsonSchemaService,
+    GenericValidator,
+    ValidationErrorFactory,
+    DelegatesService,
```
fmrsabino, line 14: Maybe it makes more sense to export these in the `DomainModule`. But since this is being used for the routes maybe we should create a new module that bundles: `JsonSchemaService`, `GenericValidator` and `ValidationErrorFactory`. What do you think?

src/routes/delegates/pipes/delete-delegate-dto-validation.pipe.ts
```diff
@@ -0,0 +1,30 @@
...
+    this.isValid = this.jsonSchemaService.getSchema(
+      'https://safe-client.safe.global/schemas/delegates/delete-delegate.json',
+      deleteDelegateDtoSchema,
+    );
+  }
+  transform(data: any): DeleteDelegateDto {
+    return this.genericValidator.validate(
+      this.isValid,
+      data,
+      HttpStatus.BAD_REQUEST,
```
fmrsabino, line 27: I understand the need to parameterise this but I think every route would eventually map the error to a `BAD_REQUEST`.

For now I think it's ok but the interface around validation is getting a bit convoluted (i.e. deals with schema validation _AND_ http).

Having an exception thrown and caught on the route that would map to a specific code seems to be a more scalable approach.

src/routes/delegates/delegates.module.ts
hectorgomezv: Yes, I think it's a good solution to this. What do you think about naming it `ValidationModule`? I've created it [04628c0](https://github.com/5afe/safe-client-gateway-nest/pull/271/commits/04628c01ebcec0eb0cd6129196bfb51ffdfc99c3). It's grouping the validation-related stuff, and located at the same level that `domain.module.ts`, do you think it makes sense?

src/routes/delegates/pipes/delete-delegate-dto-validation.pipe.ts
hectorgomezv: Yes, I agree maybe this was introducing too much complexity into the validation logic. I've reverted the parametrization, so with this approach, we're just mapping the status code in the pipe directly: [7287b89](https://github.com/5afe/safe-client-gateway-nest/pull/271/commits/7287b895b0e15e316a3b2d81986efa8dc4e0aeb7) What do you think?

src/validation.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { GenericValidator } from './domain/schema/generic.validator';
+import { JsonSchemaService } from './domain/schema/json-schema.service';
+import { ValidationErrorFactory } from './domain/schema/validation-error-factory';
+
+@Module({
+  providers: [JsonSchemaService, GenericValidator, ValidationErrorFactory],
+  exports: [JsonSchemaService, GenericValidator, ValidationErrorFactory],
+})
+export class ValidationModule {}
```
fmrsabino, line 10: This module can now be used in the `AppModule`. If you prefer to do it in a separate PR or here either way it's fine 👍

fmrsabino (approved):

PR #272, opened 2023-02-14
Share configuration setup between app and tests

fmrsabino:
- Introduces `AppProvider` – its main goal is to provide a `INestApplication`.
- This class was introduced because the application created for the service and the applications created for a testing scenario should be mostly the same (besides some differences like the module being targeted).
- This also allows to configure applications in a more centralized manner given that in most scenarios, they should have the same configuration
- In a testing scenario `TestAppProvider` should be used as it supports instances of `TestingModule`.
- Given that API versioning is now part of the testing application, the tests were adjusted to include the version in the path

### DIFF
--- a/src/app.provider.ts
+++ b/src/app.provider.ts
@@ -28,6 +28,13 @@ function configureSwagger(app: INestApplication) {
   SwaggerModule.setup('', app, document);
 }
 
+const DEFAULT_CONFIGURATION: ((app: INestApplication) => void)[] = [
+  configureVersioning,
+  configureShutdownHooks,
+  configureFilters,
+  configureSwagger,
+];
+
 /**
  * The main goal of {@link AppProvider} is to provide
  * a {@link INestApplication}.
@@ -35,20 +42,17 @@ function configureSwagger(app: INestApplication) {
  * Extensions of this class should return the application in
  * {@link getApp}.
  *
- * By default the {@link AppProvider} has a configuration collection ({@link setup})
- * that can be changed by each extension of the class
+ * Each provider should have a {@link configuration} which specifies
+ * the steps taken to configure the application
  */
 export abstract class AppProvider {
-  protected readonly setup: Array<(app: INestApplication) => void> = [
-    configureVersioning,
-    configureShutdownHooks,
-    configureFilters,
-    configureSwagger,
-  ];
+  protected abstract readonly configuration: Array<
+    (app: INestApplication) => void
+  >;
 
   public async provide(module: any): Promise<INestApplication> {
     const app = await this.getApp(module);
-    this.setup.forEach((f) => f(app));
+    this.configuration.forEach((f) => f(app));
     return app;
   }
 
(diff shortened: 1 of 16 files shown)


src/app.provider.ts
```diff
@@ -0,0 +1,98 @@
...
+    super();
+    if (process.env.NODE_ENV !== 'test') {
+      throw Error('TestAppProvider used outside of a testing environment');
+    }
+
+    // Disables shutdown hooks for tests (they are not required)
+    // Enabling this in the tests might result in a MaxListenersExceededWarning
+    // as the number of listeners that this adds exceed the default
+    const index = this.setup.indexOf(configureShutdownHooks);
+    if (index > -1) this.setup.splice(index, 1);
```
hectorgomezv, line 88: What about defining the array of setup functions explicitly for each class extending `AppProvider`? In my opinion, is a bit clearer than modifying the parent class array, but I leave this to your preference 🙂

I was thinking in something like:
```javascript
export abstract class AppProvider {
  protected setup: Array<(app: INestApplication) => void>;
  ...
```
And then: 
```javascript
export class DefaultAppProvider extends AppProvider {
  constructor() {
    super();
    this.setup = [configureVersioning, configureShutdownHooks, configureFilters, configureSwagger];
    ...
```
```javascript
export class TestAppProvider extends AppProvider {
  constructor() {
    super();
    this.setup = [configureVersioning, configureFilters, configureSwagger];
    ...
```

hectorgomezv (approved):
Great job! Looks good to me, I've just left a small comment/suggestion. 🚀

src/app.provider.ts
fmrsabino: > What about defining the array of setup functions explicitly for each class extending `AppProvider`? In my opinion, is a bit clearer than modifying the parent class array, but I leave this to your preference 🙂
> 
> I was thinking in something like:
> 
> ```js
> export abstract class AppProvider {
>   protected setup: Array<(app: INestApplication) => void>;
>   ...
> ```
> 
> And then:
> 
> ```js
> export class DefaultAppProvider extends AppProvider {
>   constructor() {
>     super();
>     this.setup = [configureVersioning, configureShutdownHooks, configureFilters, configureSwagger];
>     ...
> ```
> 
> ```js
> export class TestAppProvider extends AppProvider {
>   constructor() {
>     super();
>     this.setup = [configureVersioning, configureFilters, configureSwagger];
>     ...
> ```

My concern with this approach is that you stop sharing the same configuration. Ideally the `TestAppProvider` should be exactly the same as the `DefaultAppProvider`. 

If we add more configurations I see it as adding to the abstract `AppProvider` and that's it (unless it breaks the tests). Also ideally there wouldn't be a `TestAppProvider` and we would the _normal_ `DefaultAppProvider`.

Unfortunately how you create a test application is different than how you create a normal application:

```typescript
testApp = module.createNestApplication()
// vs
normalApp = NestFactory.create(module)
```

I think we can consider your approach in the future. I'm adding a new logging configuration which needs to be on both the testing and the default providers. We can see how it develops over time.

hectorgomezv (approved):

PR #275, opened 2023-02-16
Implement GET /v1/chains/:chainId/messages/:messageHash

### LINKED ISSUE #257: Add GET /v1/chains/<chain_id>/messages/<message_hash>

hectorgomezv:
Closes #257

Adds the GET endpoint `/v1/chains/:chainId/messages/:messageHash`

- Mapping logic for messages was extracted to `MessageMapper` since this mapper will probably be reused for other endpoints (retrieve messages by Safe address by example).
- Unit (controller) tests were added for the route. Additional unit tests for `MessageMapper` were introduced.


src/routes/messages/entities/message.entity.ts
```diff
@@ -0,0 +1,63 @@
...
+  NeedsConfirmation = 'NEEDS_CONFIRMATION',
+  Confirmed = 'CONFIRMED',
+}
+
+export class Message {
+  @ApiProperty()
+  messageHash: string;
+  @ApiProperty()
+  status: MessageStatus;
+  @ApiPropertyOptional()
```
fmrsabino, line 15: Don't we need to specify the nullable type in this annotation too?

src/routes/messages/mappers/message-mapper.ts
```diff
@@ -0,0 +1,75 @@
...
+      domainMessageConfirmations.length >= safe.threshold
+        ? MessageStatus.Confirmed
+        : MessageStatus.NeedsConfirmation;
+    const proposedBy = await this.addressInfoHelper.getOrDefault(
+      chainId,
+      message.proposedBy,
+    );
+    const confirmations = await this.mapConfirmations(
+      chainId,
+      message.confirmations ?? [],
```
fmrsabino, line 38: We should always have a non-nullable array for confirmations right?

src/routes/messages/mappers/message-mapper.spec.ts
```diff
@@ -0,0 +1,210 @@
...
+import { safeAppBuilder } from '../../../domain/safe-apps/entities/__tests__/safe-app.builder';
+import { SafeAppsRepository } from '../../../domain/safe-apps/safe-apps.repository';
+import { safeBuilder } from '../../../domain/safe/entities/__tests__/safe.builder';
+import { AddressInfoHelper } from '../../common/address-info/address-info.helper';
+import { AddressInfo } from '../../common/entities/address-info.entity';
+import { MessageConfirmation } from '../entities/message-confirmation.entity';
+import { Message, MessageStatus } from '../entities/message.entity';
+import { MessageMapper } from './message-mapper';
+
+describe('Message mapper (Unit)', () => {
```
fmrsabino, line 15: In my opinion most of these cases should be covered by the controller test right?

One disadvantage of this test (in comparison to the controller) is that it relies more heavily on the underlying implementation.

This type of mapping, in my opinion, is the core part of the route itself. What do you think about merged the scenarios described here with the controller tests?

src/routes/messages/entities/message.entity.ts
hectorgomezv: Yes, good catch, it's needed to make OpenAPI specification to show `nullable: true` flag. I've fixed it in [7f7e62e](https://github.com/5afe/safe-client-gateway-nest/pull/275/commits/7f7e62e3e71639a85bf9640f861913cb68f50e10)

We need to fix this as well for other routes ([example 1](https://github.com/5afe/safe-client-gateway-nest/blob/982d7b7392eae06d4b91f21bb82c40d0ba3651f4/src/routes/safes/entities/safe-info.entity.ts#L29), [example 2](https://github.com/5afe/safe-client-gateway-nest/blob/982d7b7392eae06d4b91f21bb82c40d0ba3651f4/src/routes/transactions/entities/settings-change-transaction.entity.ts#L9)). I will do it in a separate PR to avoid adding noise to this one 🙂

src/routes/messages/mappers/message-mapper.ts
hectorgomezv: According to the [Transaction Service specification](https://safe-transaction-mainnet.safe.global/), `confirmations` in `SafeMessageResponse` is optional, so I assumed this could be null (I defined this field as `nullable` in `message.schema.ts` as well), maybe I'm missing something here? 🙂

src/routes/messages/mappers/message-mapper.spec.ts
hectorgomezv: Yes, it's a good idea, I'll move these tests to the controller tests 👍

hectorgomezv: I've moved the tests cases which were not previously covered to the controller in [1ddeeae](https://github.com/5afe/safe-client-gateway-nest/pull/275/commits/1ddeeaecf40f2b46d16314890d294c129580cd8f)

src/routes/messages/mappers/message-mapper.ts
fmrsabino: > According to the [Transaction Service specification](https://safe-transaction-mainnet.safe.global/), `confirmations` in `SafeMessageResponse` is optional, so I assumed this could be null (I defined this field as `nullable` in `message.schema.ts` as well), maybe I'm missing something here? 🙂

The schema is automatically generated and it's not always correct. It says that `confirmations` is a `string` too but it's actually an array.

See https://github.com/safe-global/safe-transaction-service/blob/4e0d981c20c59f6524aa3a3032f0b07ff7b991f4/safe_transaction_service/safe_messages/serializers.py#L163-L172 (serializer uses `many=True` and is a required field)

Also the current implementation of the CGW: https://github.com/safe-global/safe-client-gateway/blob/12c8e08370218030a974613ecb733d8ca1398f33/src/routes/messages/backend_models.rs#L34

fmrsabino (commented):
Only thing remaining is the domain entity change 👍

src/routes/messages/mappers/message-mapper.ts
hectorgomezv: Got it, fixed in [dc207bc](https://github.com/5afe/safe-client-gateway-nest/pull/275/commits/dc207bcc65410f71d4e6713e9bff1bd77f6d3b83)

src/domain/messages/entities/__tests__/message.builder.ts
```diff
@@ -0,0 +1,35 @@
...
+    )
+    .with('preparedSignature', faker.datatype.hexadecimal(32));
+}
+
+export function toJson(message: Message): unknown {
+  return {
+    ...message,
+    created: message.created.toISOString(),
+    modified: message.modified.toISOString(),
+    confirmations: message?.confirmations?.map((confirmation) =>
```
fmrsabino, line 31:
```suggestion
    confirmations: message?.confirmations.map((confirmation) =>
```

fmrsabino (changes requested):

src/domain/messages/entities/__tests__/message.builder.ts
hectorgomezv: I forgot this one, thanks!

fmrsabino (approved):
🚀 🚀 🚀

PR #278, opened 2023-02-20
Use ValidationModule to validate input DTOs

### LINKED ISSUE #277: Use ValidationModule to validate input HTTP requests
We want to extend the work made in [this PR](https://github.com/5afe/safe-client-gateway-nest/pull/271) to validate the payloads of several endpoints by using Nest Pipes and `ValidationModule`. Some requirements will be fulfilled by this work:

- Use `ValidationModule` on `AppModule` directly.
- Evaluate whether `ValidationModule` should be a `Global` module in the project (remove `ValidationModule` components from `DomainModule` in that case).
- Implement `ValidationPipes` in each body sent by the API clients (review all the current HTTP controllers of the service)

hectorgomezv:
Closes #277 

Besides filling the requirements specified in #277, this PR also simplifies some stuff related to DTOs validation:
- Unifies input payload names by using the `Dto` suffix. For example: `EstimationRequest` -> `GetEstimationDto`.
- Adds the `ValidationModule` dependency to many test classes, so a lot of files are modified with this minor change.


src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
```diff
@@ -55,6 +56,40 @@ describe('Data decode e2e tests', () => {
       });
   });
 
+  it('POST /data-decoder should throw a validation error', async () => {
```
hectorgomezv, line 59: Moving the validation checks here since these are performed at controller/validation pipe level now.

src/routes/data-decode/__tests__/get-data-decoded.dto.factory.ts
```diff
@@ -0,0 +1,9 @@
+import { faker } from '@faker-js/faker';
+import { GetDataDecodedDto } from '../entities/get-data-decoded.dto.entity';
+
+export default function (data?: string, to?: string): GetDataDecodedDto {
```
hectorgomezv, line 4: Renamed `CreateDataDecodedDto` to `GetDataDecodedDto`. I think it makes more sense from the clients perspective.

src/routes/data-decode/entities/get-data-decoded.dto.entity.ts
```diff
@@ -1,7 +1,6 @@
 import { ApiProperty } from '@nestjs/swagger';
-import { isHex } from '../../common/utils/utils';
 
-export class CreateDataDecodedDto {
+export class GetDataDecodedDto {
```
hectorgomezv, line 3: Moving `isCreateDataDecodeDto` guard/validation function from this entity class to the proper ValidationPipe.

src/routes/data-decode/data-decoded.service.spec.ts
```diff
@@ -32,30 +32,6 @@ describe('DataDecoded Service', () => {
...
-    const createDataDecodedDto = createDataDecodedDtoFactory(
-      faker.random.words(),
-      faker.random.word(),
-    );
-
-    await expect(
-      service.getDataDecoded(chainId, createDataDecodedDto),
-    ).rejects.toThrow('Invalid payload');
-  });
 });
```
hectorgomezv, line 35: Moving validation checks to the e2e tests since this validation step is also moved from the service to the controller/validation pipe.

src/routes/delegates/entities/schemas/get-delegate.dto.schema.ts
```diff
@@ -0,0 +1,19 @@
...
+export const getDelegateDtoSchema: JSONSchemaType<GetDelegateDto> = {
+  $id: 'https://safe-client.safe.global/schemas/delegates/get-delegate.dto.json',
+  type: 'object',
+  properties: {
+    safe: { type: 'string', nullable: true },
+    delegate: { type: 'string', nullable: true },
+    delegator: { type: 'string', nullable: true },
+    label: { type: 'string', nullable: true },
+  },
+  oneOf: [
```
hectorgomezv, line 13: This assures at least one of the properties is present (business requirement).

src/routes/data-decode/__tests__/get-data-decoded.dto.factory.ts
fmrsabino: Why don't we use the `Builder` class?

src/routes/data-decode/data-decoded.service.spec.ts
fmrsabino: Do you think this test class should exist now? I would say that it is a single test that should be already covered by the controller (and this one is tied to the implementation)

src/validation.module.ts
```diff
@@ -1,8 +1,9 @@
-import { Module } from '@nestjs/common';
+import { Global, Module } from '@nestjs/common';
 import { GenericValidator } from './domain/schema/generic.validator';
 import { JsonSchemaService } from './domain/schema/json-schema.service';
 import { ValidationErrorFactory } from './domain/schema/validation-error-factory';
 
+@Global()
```
fmrsabino, line 6: I would create a folder for this class under `src` and for the `providers` that this class depends.

If this is a Global module I think it makes sense to not have a direct dependency to the `domain`

src/routes/delegates/pipes/create-delegate.dto.validation.pipe.ts
```diff
@@ -0,0 +1,31 @@
...
+    );
+  }
+  transform(data: any): CreateDelegateDto {
+    try {
+      return this.genericValidator.validate(this.isValid, data);
+    } catch (err) {
+      err.status = HttpStatus.BAD_REQUEST;
+      throw err;
+    }
+  }
```
fmrsabino, line 30: Have you considered something like the `DataSourceErrorFilter`?

Basically we can work only with exceptions at this level and ignore the HTTP body/status. The Error Filter would then, depending on the exception caught assign the correct response payload.

fmrsabino (changes requested):

src/routes/data-decode/data-decoded.service.spec.ts
hectorgomezv: Yes, I agree, this should be covered already, I've deleted it in [3a1de96](https://github.com/5afe/safe-client-gateway-nest/pull/278/commits/3a1de96b74e40438d80e3a094f6e635d7330182d)

src/routes/data-decode/__tests__/get-data-decoded.dto.factory.ts
hectorgomezv: I deleted this file in [3a1de96](https://github.com/5afe/safe-client-gateway-nest/pull/278/commits/3a1de96b74e40438d80e3a094f6e635d7330182d) since it was used by a redundant test class only (see https://github.com/5afe/safe-client-gateway-nest/pull/278#discussion_r1115971367)

src/routes/chains/chains.controller.spec.ts
```diff
@@ -170,11 +172,10 @@ describe('Chains Controller (Unit)', () => {
...
 
-      await request(app.getHttpServer()).get('/v1/chains').expect(500).expect({
-        message: 'Validation failed',
-        code: 42,
-        arguments: [],
-      });
+      await request(app.getHttpServer())
+        .get('/v1/chains')
+        .expect(500)
+        .expect({ message: 'Validation failed', code: 42, arguments: [] });
```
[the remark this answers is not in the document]
hectorgomezv: I was dealing with validation and I rewrote this assertion in a more compact way so I changed it in other parts of the code as well. But agree with you that maybe changing formatting in multiple places makes PRs harder to review.  I've reverted these format-related changes in [3a1de96](https://github.com/5afe/safe-client-gateway-nest/pull/278/commits/3a1de96b74e40438d80e3a094f6e635d7330182d)

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -469,15 +469,15 @@ export class TransactionApi implements ITransactionApi {
   }
   async postEstimation(
     address: string,
-    estimationRequest: EstimationRequest,
+    getEstimationDto: GetEstimationDto,
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [3a1de96](https://github.com/5afe/safe-client-gateway-nest/pull/278/commits/3a1de96b74e40438d80e3a094f6e635d7330182d)

src/validation.module.ts
hectorgomezv: I agree. I've created the folder `src/validation` and another folder `src/validation/providers/` with the contents of `src/domain/schema` 👍🏻 [2d26fea](https://github.com/5afe/safe-client-gateway-nest/pull/278/commits/2d26feaa7e29bf2e4f70bf72c0b82468211ebe3b)

src/routes/delegates/pipes/create-delegate.dto.validation.pipe.ts
hectorgomezv: Yes, I agree in general terms, but do you think this could be done in a separate PR? I think it could be useful to define a `ValidationErrorFilter` similar to `DataSourceErrorFilter` and attach it to the application, but I also think we should differentiate client validation errors from errors when validating provider responses (Transaction Service, Config Service, exchanges, etc.). 

If that's the case we would need to modify a lot of tests as well, and add a couple of classes, so maybe a new PR fits better, what do you think?

src/routes/data-decode/data-decoded.service.spec.ts
hectorgomezv: I've created another PR to delete some other test classes whose test cases should be already covered by controller layer unit tests: https://github.com/5afe/safe-client-gateway-nest/pull/284 🙂

src/routes/delegates/pipes/create-delegate.dto.validation.pipe.ts
fmrsabino: > Yes, I agree in general terms, but do you think this could be done in a separate PR? I think it could be useful to define a `ValidationErrorFilter` similar to `DataSourceErrorFilter` and attach it to the application, but I also think we should differentiate client validation errors from errors when validating provider responses (Transaction Service, Config Service, exchanges, etc.).
> 
> If that's the case we would need to modify a lot of tests as well, and add a couple of classes, so maybe a new PR fits better, what do you think?

Can be done in a different PR. And doing something like the `DataSourceErrorFilter` doesn't mean that you need to output the same error as with the current `DataSourceErrorFilter`. I was only talking about the mechanism surrounding catching types of errors to return a specific response.

fmrsabino (approved):

PR #279, opened 2023-02-23
Implement GET /v1/chains/:chainId/safes/:safeAddress/messages

hectorgomezv:


src/routes/messages/mappers/message-mapper.ts
```diff
@@ -16,6 +17,32 @@ export class MessageMapper {
...
+          message.name,
+          message.message,
+          message.creationTimestamp,
+          message.modifiedTimestamp,
+          message.confirmationsSubmitted,
+          message.confirmationsRequired,
+          message.proposedBy,
+          message.confirmations,
+          message.preparedSignature,
+        );
```
hectorgomezv, line 41: This could be simplified if we merge both `MessageItem` and `Message` entities. See: https://github.com/5afe/safe-client-gateway-nest/pull/279/files#r1118446395 🙂

src/domain/interfaces/transaction-api.interface.ts
```diff
@@ -151,4 +151,11 @@ export interface ITransactionApi {
   ): Promise<Estimation>;
 
   getMessageByHash(messageHash: string): Promise<Message>;
+
+  getMessagesBySafe(
+    safeAddress: string,
+    ordering?: string,
```
fmrsabino, line 157: I don't think we use the `ordering` parameter at all for this feature. I think it's better to remove it to keep the number of parameters smaller.

src/routes/messages/messages.controller.spec.ts
```diff
@@ -425,4 +426,181 @@ describe('Messages controller', () => {
...
+      await request(app.getHttpServer())
+        .get(`/v1/chains/${chain.chainId}/safes/${safe.address}/messages`)
+        .expect(200)
+        .expect(({ body }) => {
+          expect(body).toEqual(
+            pageBuilder()
+              .with('next', null)
+              .with('previous', null)
+              .with('count', messages.length)
+              .with('results', [
```
fmrsabino, line 472: nitpick: I like the way this test is written – it tries to go down to what is expected but not trying to match everything.

However maybe it's due to the size of the `results` but I think testing the entire schema returned is valuable. We can do that with a single message and then use partial-matching for bigger payloads.

src/routes/messages/messages.service.ts
```diff
@@ -27,4 +36,70 @@ export class MessagesService {
...
+      new Date(
+        Date.UTC(
+          m.created.getFullYear(),
+          m.created.getMonth(),
+          m.created.getDate(),
+        ),
+      ).getTime(),
+    );
+
+    return orderBy(Object.entries(groups), '0').map(([timestamp, messages]) => [
```
fmrsabino, line 100: What does this '0' represent here?

src/routes/messages/entities/message-item.entity.ts
```diff
@@ -0,0 +1,12 @@
...
+import { Message } from './message.entity';
+
+export class MessageItem extends Message {
+  @ApiProperty()
+  type: string;
+
+  constructor(...args: ConstructorParameters<typeof Message>) {
+    super(...args);
+    this.type = 'MESSAGE';
+  }
```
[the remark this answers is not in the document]
fmrsabino: I would leave the type when we have a collection of different types (eg.: collection of dates and messages).

Adding a `type` in other message related endpoints is not an API breaking change but semantically it might not make sense. If you hit a resource where you are expected to only get messages why is a `MESSAGE` type returned?

This should be dealt within the service in my opinion. We should be able to reuse entities and show/hide properties when required.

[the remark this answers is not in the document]
hectorgomezv: Yes, I agree, having the `type: MESSAGE` property in other related endpoint responses it's maybe doubtful. So I think we could keep this as it is and use this `MessageItem` class to map the expected format. What do you think?

Another option would be adding the `type: MESSAGE` property in the service and using `Required<Message>` (https://www.typescriptlang.org/docs/handbook/utility-types.html#requiredtype) but it's maybe a bit overkill for this use case.

src/routes/messages/messages.controller.spec.ts
hectorgomezv: Yes I agree, I've added another test that checks the validity of a response consisting in a single `DateLabel` and `MessageItem` in [7293f5f](https://github.com/5afe/safe-client-gateway-nest/pull/279/commits/7293f5fdc5270406df511c6be56bcb0a5c38a728)

fmrsabino (approved):
🚀

PR #285, opened 2023-03-02
Extract cache directions calculation

hectorgomezv:
**(Please review https://github.com/5afe/safe-client-gateway-nest/pull/284 before)**

This PR is preliminary work for #255. The `/v2/flush` endpoint needs to invalidate cache keys on demand, but the calculation of a concrete key/field is currently bonded to the client API functions (`transaction-api.service`, `config-api.service` classes), so this PR:

- Extracts the cache key/field calculation from the API client class.
- Centralises this calculation in a utility class with static functions (since state is not needed and the calculation is deterministic): `CacheRouter`.
- Adds the `CacheDir` interface which encapsulates the key/field for a given cache address.

### DIFF
--- a/src/datasources/cache/cache.first.data.source.ts
+++ b/src/datasources/cache/cache.first.data.source.ts
@@ -1,11 +1,12 @@
+import { Inject, Injectable } from '@nestjs/common';
+import * as winston from 'winston';
 import { NetworkRequest } from '../network/entities/network.request.entity';
-import { CacheService, ICacheService } from './cache.service.interface';
 import {
   INetworkService,
   NetworkService,
 } from '../network/network.service.interface';
-import { Inject, Injectable } from '@nestjs/common';
-import * as winston from 'winston';
+import { CacheService, ICacheService } from './cache.service.interface';
+import { CacheDir } from './entities/cache-dir.entity';
 
 /**
  * A data source which tries to retrieve values from cache using
@@ -24,34 +25,36 @@ export class CacheFirstDataSource {
   ) {}
 
   /**
-   * Gets the cached value behind the {@link field} of {@link key}.
+   * Gets the cached value behind the {@link CacheDir}.
    * If the value is not present, it tries to get the respective JSON
    * payload from {@link url}.
    * Errors are not cached.
    *
-   * @param key - the key to be used to retrieve from cache
-   * @param field - the field to get from {@link key}
+   * @param cacheDir - {@link CacheDir} containing the key and field to be used to retrieve from cache
    * @param url - the HTTP endpoint to retrieve the JSON payload
    * @param params - the parameters to be used for the HTTP request
    * @param expireTimeSeconds - the time to live in seconds for the payload
    * behind {@link key}
    */
   async get<T>(
-    key: string,
-    field: string,
+    cacheDir: CacheDir,
     url: string,
     params?: NetworkRequest,
     expireTimeSeconds?: number,
   ): Promise<T> {
-    const cached = await this.cacheService.get(key, field);
+    const cached = await this.cacheService.get(cacheDir);
     if (cached != null) {
-      winston.debug(`[Cache] Cache hit: ${key}`);
+      winston.debug(`[Cache] Cache hit: ${cacheDir.key}`);
       return JSON.parse(cached);
     }
-    winston.debug(`[Cache] Cache miss: ${key}`);
+    winston.debug(`[Cache] Cache miss: ${cacheDir.key}`);
     const { data } = await this.networkService.get(url, params);
     const rawJson = JSON.stringify(data);
-    await this.cacheService.set(key, field, rawJson, expireTimeSeconds);
+    await this.cacheService.set(
+      new CacheDir(cacheDir.key, cacheDir.field),
+      rawJson,
+      expireTimeSeconds,
+    );
     return data;
   }
 }
--- a/src/datasources/cache/cache.first.data.source.ts
+++ b/src/datasources/cache/cache.first.data.source.ts
@@ -1,11 +1,11 @@
-import { Inject, Injectable } from '@nestjs/common';
-import * as winston from 'winston';
 import { NetworkRequest } from '../network/entities/network.request.entity';
+import { CacheService, ICacheService } from './cache.service.interface';
 import {
   INetworkService,
   NetworkService,
 } from '../network/network.service.interface';
-import { CacheService, ICacheService } from './cache.service.interface';
+import { Inject, Injectable } from '@nestjs/common';
+import * as winston from 'winston';
 import { CacheDir } from './entities/cache-dir.entity';
 
 /**
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -24,13 +24,6 @@ export class CacheRouter {
   private static readonly TOKENS_KEY = 'tokens';
   private static readonly TRANSFERS_KEY = 'transfers';
 
-  static balanceCacheKey(chainId: string, safeAddress: string): CacheDir {
-    return new CacheDir(
-      `${chainId}_${safeAddress}_${CacheRouter.BALANCES_KEY}`,
-      '',
-    );
-  }
-
   static getBalanceCacheDir(
     chainId: string,
     safeAddress: string,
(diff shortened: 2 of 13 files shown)


src/datasources/cache/entities/cache-dir.entity.ts
```diff
@@ -0,0 +1,9 @@
+export class CacheDir {
```
hectorgomezv, line 1: I chose the name `CacheDir` but I'm open to suggestions. I also considered `CacheKey` but this kind of hides the fact the objects hold both `key` and `field`, also I thought about `CacheAddress` but the `address` concept is more heavily used in blockchain for other stuff.

hectorgomezv: Another option, instead of having an ad-hoc interface, could be just returning a tuple `[key, field]` but I find it a bit more convoluted, less explicit, and harder to manage.

src/datasources/cache/cache.router.ts
```diff
@@ -0,0 +1,249 @@
...
+  private static readonly MESSAGE_KEY = 'message';
+  private static readonly MODULE_TRANSACTIONS_KEY = 'module_transactions';
+  private static readonly MULTISIG_TRANSACTION_KEY = 'multisig_transaction';
+  private static readonly MULTISIG_TRANSACTIONS_KEY = 'multisig_transactions';
+  private static readonly OWNERS_SAFE_KEY = 'owner_safes';
+  private static readonly SAFE_APPS_KEY = 'safe_apps';
+  private static readonly SAFE_KEY = 'safe';
+  private static readonly TOKEN_KEY = 'token';
+  private static readonly TOKENS_KEY = 'tokens';
+  private static readonly TRANSFERS_KEY = 'transfers';
```
hectorgomezv, line 24: I just copied the keys from the current functions. I think at this point we're free to change these keys and maybe normalize them (`-` vs `_` usage for example).

src/datasources/cache/cache.router.ts
```diff
@@ -0,0 +1,265 @@
+import { CacheDir } from './entities/cache-dir.entity';
+
+export class CacheRouter {
```
fmrsabino, line 3: My biggest concern with the `CacheRouter` is that we introduce a component that is now aware of all the features on what we consider to be the caching layer.

Previously this data was provided by each feature, and on a feature basis. I'll try to think of an alternative solution but I do agree that we need a mechanism to easily retrieve/delete/update these entries.

src/datasources/cache/cache.router.ts
fmrsabino: Yeah we can change these keys freely (in the current state of the project). We can do it in a separate PR.

src/datasources/cache/entities/cache-dir.entity.ts
```diff
@@ -0,0 +1,9 @@
...
+  key: string;
+  field: string;
```
fmrsabino, line 3:
```suggestion
  readonly key: string;
  readonly field: string;
```

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -1,11 +1,12 @@
...
 import { NetworkRequest } from '../network/entities/network.request.entity';
-import { CacheService, ICacheService } from './cache.service.interface';
 import {
   INetworkService,
   NetworkService,
 } from '../network/network.service.interface';
-import { Inject, Injectable } from '@nestjs/common';
-import * as winston from 'winston';
+import { CacheService, ICacheService } from './cache.service.interface';
+import { CacheDir } from './entities/cache-dir.entity';
```
fmrsabino, line 9: We need to double check the import rules else the import order constantly changes between PRs.

src/datasources/config-api/config-api.service.ts
```diff
@@ -23,26 +24,29 @@ export class ConfigApi implements IConfigApi {
...
   }
 
   async getChain(chainId: string): Promise<Chain> {
     try {
-      const cacheKey = `${chainId}_chain`;
-      const field = '';
       const url = `${this.baseUri}/api/v1/chains/${chainId}`;
-      return await this.dataSource.get(cacheKey, field, url);
+      return await this.dataSource.get(
+        CacheRouter.getChainCacheDir(chainId),
```
fmrsabino, line 47: nitpick: for readability we could extract the `cacheDir` to a const.

```typescript
const url = `${this.baseUri}/api/v1/chains/${chainId}`;
const cacheDir = CacheRouter.getChainCacheDir(chainId);
return await this.dataSource.get(cacheDir, url);
```

src/datasources/config-api/config-api.service.ts
```diff
@@ -54,11 +58,8 @@ export class ConfigApi implements IConfigApi {
     url?: string,
   ): Promise<SafeApp[]> {
     try {
-      const cacheKey = `safe_apps`;
-      const field = `${chainId}_${clientUrl}_${url}`;
       return await this.dataSource.get(
-        cacheKey,
-        field,
+        CacheRouter.getSafeAppsCacheDir(chainId, clientUrl, url),
```
fmrsabino, line 62: nitpick: see my comment regarding extracting the `cacheDir` parameter.

Additionally, to extract the URL and params here.

src/datasources/cache/entities/cache-dir.entity.ts
fmrsabino: Let's go with `CacheDir` for now. Another alternative could be `CachePath` (I see it as the "cache" path to the object that we want to retrieve).

src/datasources/cache/cache.first.data.source.ts
fmrsabino: I'm checking solution to this.

src/datasources/cache/cache.router.ts
hectorgomezv: I agree, this could be more polished by decoupling the data to create the keys from this class. Thank you for your thoughts on this. I'd go with the current solution to speed up the delivery as you said, but we could decouple this in the future by another mechanism (maybe a register as we have for AJV schemas? That way each feature could register its `CacheDirs` on the application startup).

src/datasources/config-api/config-api.service.ts
hectorgomezv: Changed in [76724b7](https://github.com/5afe/safe-client-gateway-nest/pull/285/commits/76724b74b6eeed8fb5ab522bf72045f6ffe0f730)

src/datasources/config-api/config-api.service.ts
hectorgomezv: Changed in [76724b7](https://github.com/5afe/safe-client-gateway-nest/pull/285/commits/76724b74b6eeed8fb5ab522bf72045f6ffe0f730)

src/datasources/cache/cache.service.interface.ts
```diff
@@ -1,14 +1,15 @@
...
-    field: string,
+    cacheDir: CacheDir,
     value: string,
     expireTimeSeconds?: number,
   ): Promise<void>;
 
-  get(key: string, field: string): Promise<string | undefined>;
+  get(cacheDir: CacheDir): Promise<string | undefined>;
 
   delete(key: string): Promise<number>;
```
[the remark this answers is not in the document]
hectorgomezv: Yes! Thanks, changed in [76724b7](https://github.com/5afe/safe-client-gateway-nest/pull/285/commits/76724b74b6eeed8fb5ab522bf72045f6ffe0f730)

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Yes, sorry, I reverted the imports order changes in [393ea7b](https://github.com/5afe/safe-client-gateway-nest/pull/285/commits/393ea7b48995c3432938360953ca1ee336732664). This is an automatic feature of my IDE (VS Code), but we could delegate this utility into a prettier/eslint plugin (I found [organize-imports](https://www.npmjs.com/package/prettier-plugin-organize-imports) and [eslint-plugin-import](https://github.com/import-js/eslint-plugin-import)) to have a uniform behavior between all the developers environments.

fmrsabino (approved):

PR #288, opened 2023-03-22
Implement POST /v2/flush

hectorgomezv:


src/domain/flush/flush.repository.ts
```diff
@@ -0,0 +1,73 @@
...
+  private async invalidateContracts(): Promise<void> {
+    const chains = await this.configApi.getChains();
+    await Promise.all(
+      chains.results.map(async (chain) =>
+        this.cacheService.deleteByPattern(
+          CacheRouter.getContractsCachePattern(chain.chainId),
+        ),
+      ),
+    );
+  }
```
hectorgomezv, line 55: Here we're invalidating the contracts of each chain one by one. Probably we could want to invalidate contracts for a specific chain selectively in the future, so I included the `chainId` into the cache pattern. What do you think @fmrsabino? If you don't see any value here, we can just invalidate all the contracts at one by using the pattern `*_contract`.

src/datasources/cache/__tests__/fake.cache.service.spec.ts
```diff
@@ -49,4 +50,52 @@ describe('FakeCacheService', () => {
 
     expect(target.keyCount()).toBe(0);
   });
+
+  it('deletes keys by pattern', async () => {
+    const prefix = faker.datatype.uuid();
+
+    // insert 5 items matching the pattern
+    await Promise.all(
+      range(5).map(async () =>
```
fmrsabino, line 59: I don't think that you need the mapper function to be `async`. The mapper returns an async function which is then consumed by the `Promise.all`

src/datasources/cache/__tests__/fake.cache.service.spec.ts
```diff
@@ -49,4 +50,52 @@ describe('FakeCacheService', () => {
...
+        target.set(
+          new CacheDir(`${prefix}_${faker.datatype.uuid()}_${suffix}`, ''),
+          '',
+        ),
+      ),
+    );
+
```
fmrsabino, line 89: nitpick: for the test setup we have been following the following "standard":

```
// setup \n
\n
// action \n
\n
// assertions
```

In this case I'd say it's fine to remove the newlines as it's part of the "setup"

src/routes/flush/flush.controller.ts
```diff
@@ -0,0 +1,24 @@
...
+  constructor(private readonly flushService: FlushService) {}
+
+  @ApiOkResponse()
+  @HttpCode(200)
+  @Post('flush')
+  async getContract(
+    @Body(InvalidationPatternDtoValidationPipe)
+    invalidationPatternDto: InvalidationPatternDto,
+  ): Promise<void> {
+    await this.flushService.flush(invalidationPatternDto);
```
fmrsabino, line 22: We can just return the promise here

```suggestion
    return this.flushService.flush(invalidationPatternDto);
```

fmrsabino (commented):
Nice 👏 

Some small comments (still reviewing) 👍

src/datasources/cache/__tests__/fake.cache.service.spec.ts
```diff
@@ -49,4 +50,52 @@ describe('FakeCacheService', () => {
 
     expect(target.keyCount()).toBe(0);
   });
+
+  it('deletes keys by pattern', async () => {
+    const prefix = faker.datatype.uuid();
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [54cf021](https://github.com/5afe/safe-client-gateway-nest/pull/288/commits/54cf021cea7b64d7e4246ae066d3c671209e3dde)

src/datasources/cache/__tests__/fake.cache.service.spec.ts
hectorgomezv: Changed in [54cf021](https://github.com/5afe/safe-client-gateway-nest/pull/288/commits/54cf021cea7b64d7e4246ae066d3c671209e3dde)

src/datasources/cache/__tests__/fake.cache.service.spec.ts
hectorgomezv: Changed in [54cf021](https://github.com/5afe/safe-client-gateway-nest/pull/288/commits/54cf021cea7b64d7e4246ae066d3c671209e3dde)

src/datasources/cache/redis.cache.service.ts
```diff
@@ -46,6 +46,12 @@ export class RedisCacheService implements ICacheService, OnModuleDestroy {
     return await this.client.unlink(cacheDir.key);
   }
 
+  async deleteByPattern(pattern: string): Promise<void> {
+    for await (const key of this.client.scanIterator({ MATCH: pattern })) {
+      await this.client.unlink(key);
+    }
+  }
```
fmrsabino, line 53: We should really avoid performing a scan over a pattern. One of the biggest selling features of the rewrite is that the way we interact with Redis should be more deterministic. Therefore if we need to delete all the "values" for a specific feature we should think of a solution that groups those values.

Therefore we should strive to call `UNLINK` to the key that holds these values. If we need to change the key structure for that then we should evaluate that first.

fmrsabino (changes requested):

src/datasources/cache/redis.cache.service.ts
hectorgomezv: Good point! We could change the key in the case of tokens and contracts (which are the entities being deleted by pattern in this branch) and use the Redis hash to store just the reference to the chain in the _key_ part, and the concrete token/contract address in the _field_ part. So that will allow us to delete all the contracts/tokens on a chain in one call to Redis. Do you see any downside to this approach?

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -44,10 +44,10 @@ export class CacheFirstDataSource {
   ): Promise<T> {
     const cached = await this.cacheService.get(cacheDir);
     if (cached != null) {
-      winston.debug(`[Cache] Cache hit: ${cacheDir.key}`);
+      winston.debug(`[Cache] Cache hit: [${cacheDir.key}, ${cacheDir.field}]`);
```
hectorgomezv, line 47: Just adding the field key part to the logs output for better visibility. We are using the field to store addresses in this PR so I think it makes sense to include that in the logs now.

src/domain/flush/flush.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+      case InvalidationTarget.Contracts:
+        return this.invalidateContracts();
+      case InvalidationTarget.Tokens:
+        return this.invalidateTokens(pattern);
+    }
+  }
+
+  private async invalidateChains(): Promise<void> {
+    const chains = await this.configApi.getChains();
+    await this.cacheService.delete(CacheRouter.getChainsCacheDir());
```
hectorgomezv, line 36: @fmrsabino what do you think about renaming this `ICacheService.delete` function to `ICacheService.deleteByKey`? I mean, the inner implementation of the function ignores the `field` part of the `CacheDir` passed in and it just deletes all the contents pointed by the `key` part, so I think it would be clearer for the consumers of ICacheService API 🤔

src/routes/flush/flush.controller.spec.ts
```diff
@@ -0,0 +1,293 @@
...
+      await request(app.getHttpServer()).get(
+        `/v1/chains/${chain.chainId}/contracts/${contracts[1].address}`,
+      );
+
+      // check the cache is filled
+      await Promise.all(
+        contracts.map(async (contract) =>
+          expect(
+            await fakeCacheService.get(
+              CacheRouter.getContractCacheDir(chain.chainId, contract.address),
```
fmrsabino, line 164: nitpick: I would use the expected string value here instead of going through the `CacheRouter`. Reason being that if we refactor `CacheRouter` in an unintended way this test would catch that.

Or maybe we should have some tests to specifically target this 🤔

src/routes/flush/pipes/invalidation-pattern.dto.validation.pipe.ts
```diff
@@ -0,0 +1,39 @@
...
+      invalidationPatternDetailSchema,
+    );
+    this.isValid = this.jsonSchemaService.getSchema(
+      'https://safe-client.safe.global/schemas/delegates/invalidation-pattern.dto.json',
+      invalidationPatternDtoSchema,
+    );
+  }
+  transform(data: any): InvalidationPatternDto {
+    try {
+      data.invalidate = data.invalidate?.toUpperCase();
```
fmrsabino, line 32: Why do we change it to upper case?

fmrsabino (changes requested):

src/domain/flush/flush.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+        return this.invalidateChains();
+      case InvalidationTarget.Contracts:
+        return this.invalidateContracts();
+      case InvalidationTarget.Tokens:
+        return this.invalidateTokens(pattern);
+    }
+  }
+
+  private async invalidateChains(): Promise<void> {
+    const chains = await this.configApi.getChains();
```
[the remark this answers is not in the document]
hectorgomezv: I agree, this also seemed odd to me, in fact, I had to invert the order between lines 35 and 36 when I realized that call was filling up the cache after the deletion 😅 Now that we are restructuring the cache hashes we can do this in a more straightforward way 👍 I'll change it.

src/routes/flush/flush.controller.spec.ts
hectorgomezv: Yes I agree, I think having the expected string here is better 🙂

src/domain/flush/flush.repository.ts
hectorgomezv: Hmm, I'm not against keeping the current signature, but imagine a client of the API calling `delete(new CacheDir('foo', 'bar'))`. As a client, I think it's reasonable to expect that the field `bar` in within the key `foo` gets deleted, but not the whole key `foo`. Maybe we can just document that if a field is passed within the `CacheDir` it gets ignored and the function just deletes all the contents pointed by the key 🤔

src/domain/flush/flush.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  private async invalidateContracts(): Promise<void> {
+    const chains = await this.configApi.getChains();
+    await Promise.all(
+      chains.results.map((chain) =>
+        this.cacheService.delete(
+          CacheRouter.getContractCacheDir(chain.chainId, ''),
+        ),
+      ),
+    );
```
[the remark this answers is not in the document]
fmrsabino: My point was more that we can restructure the keys and unlink the "root" one easily but if from a design perspective the key structure becomes confusing we can also think that the `SCAN` will work with `N` chains where `N` is quite small and doesn't change that often.

If `SCAN` is used in a potential large collection then it'd be more impactful for performance.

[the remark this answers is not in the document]
hectorgomezv: Oh, now I think I get it, you mean keeping the `chainId` in the key part and using `SCAN` for deletion by iterating the keys (using a pattern like `${chainId}_contracts`) and executing `UNLINK` against the found keys, right? Yes, I think it's more intuitive than having just a "root" with the feature name (`contracts` in this case), and opens the door to a future use case where the client wants to delete the contracts for a specific chain only.

I'd go with the `SCAN` solution then, but in that case, we need an additional function to retrieve/delete the keys by pattern, as [this PR had previously](https://github.com/5afe/safe-client-gateway-nest/pull/288#discussion_r1162811965), right?

src/routes/flush/entities/invalidation-pattern.dto.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+
+export class InvalidationPatternDetails {
+  @ApiPropertyOptional({ type: 'string', nullable: true })
+  chain_id: string | null;
```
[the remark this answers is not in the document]
hectorgomezv: I took this from here: https://www.notion.so/safe-global/How-to-invalidate-the-cache-e573a9f39a0c40c99f45a7868dd459c0

src/routes/flush/pipes/invalidation-pattern.dto.validation.pipe.ts
hectorgomezv: To cover the case where the client sends an all-lowercase string in the payload, like `{ "invalidate": "chains" }`. Maybe this could fit better in the repository, where the actual matching against a `InvalidationTarget` happens, what do you think?

fmrsabino: But the current CGW doesn't support upper case types or?

Asking this because I'm also considering the value that the mapping in the enum provides. For example maybe we can just use:

```typescript
export enum InvalidationTarget {
  Chains,
  Contracts,
  Tokens,
}
```

And as an example:

```typescript
console.log(InvalidationTarget[InvalidationTarget.Chains]) // prints 'Chains'
```

src/domain/flush/flush.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+    const chainId = pattern?.patternDetails?.chainId;
+    if (!chainId) {
+      throw new UnprocessableEntityException(
+        `Unprocessable cache invalidation pattern detail: ${JSON.stringify(
+          pattern.patternDetails,
+        )}`,
+      );
+    }
+    await this.cacheService.delete(CacheRouter.getTokensCacheDir(chainId));
+    await this.cacheService.delete(CacheRouter.getTokenCacheDir(chainId, ''));
```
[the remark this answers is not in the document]
fmrsabino: > Yes, I think this is derived from passing a `CacheDir` to the delete method (see [this discussion](https://github.com/5afe/safe-client-gateway-nest/pull/285#discussion_r1159723868)), maybe just passing a string representing the key instead of a full `CacheDir` makes more sense after all. This is also related with [this comment](https://github.com/5afe/safe-client-gateway-nest/pull/288#discussion_r1163904212).
> 
> My suggestion here is to just pass the key part to the `delete` function (because we are always dealing by key and ignoring the field part) and rename it: `ICacheService.deleteByKey(key: string): Promise<void>`, what do you think?

For now I think this is ok but I still think we should improve on this solution. if `delete` receives a key then we leak the implementation of what `CacheDir` contains. This is more prominent when you have `set(CacheDir)` and `get(CacheDIr)`.

But I would just move forward with your suggestion since we won't be tackling this in this PR 👍 Just providing food for thought now!

src/domain/flush/flush.repository.ts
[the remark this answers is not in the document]
fmrsabino: > Oh, now I think I get it, you mean keeping the `chainId` in the key part and using `SCAN` for deletion by iterating the keys (using a pattern like `${chainId}_contracts`) and executing `UNLINK` against the found keys, right? Yes, I think it's more intuitive than having just a "root" with the feature name (`contracts` in this case), and opens the door to a future use case where the client wants to delete the contracts for a specific chain only.
> 
> I'd go with the `SCAN` solution then, but in that case, we need an additional function to retrieve/delete the keys by pattern, as [this PR had previously](https://github.com/5afe/safe-client-gateway-nest/pull/288#discussion_r1162811965), right?

Let's go with the `SCAN` solution for now. If you prefer to implement the `SCAN` solution separately that's also ok. If you think that this specific feature can be under a non-parameterised key (eg.: no chain id and no addresses) for this PR it's also ok 👍

src/domain/flush/flush.repository.ts
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+
+  private async invalidateTokens(
+    pattern: InvalidationPatternDto,
+  ): Promise<void> {
+    const chainId = pattern?.patternDetails?.chainId;
+    if (!chainId) {
+      throw new UnprocessableEntityException(
+        `Unprocessable cache invalidation pattern detail: ${JSON.stringify(
```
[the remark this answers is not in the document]
hectorgomezv: Yes, seems unnecessary to stringify the payload, just changed it in [7975518](https://github.com/5afe/safe-client-gateway-nest/pull/288/commits/79755184f53fb1d187f427143d3132b251d5b265)

src/routes/flush/pipes/invalidation-pattern.dto.validation.pipe.ts
hectorgomezv: My idea was to make this `InvalidationTarget` case insensitive, but I tested the current CGW and it is case sensitive (i.e. it requires a capitalized word like `Contracts`, `Tokens`...) so maybe it makes sense to remove the `toUpperCase()` hack and just use the enum values as you commented 👍🏻

src/routes/flush/flush.controller.spec.ts
hectorgomezv: Changed in [04c3cf8](https://github.com/5afe/safe-client-gateway-nest/pull/288/commits/04c3cf8e534137d209634a8f98ee83cd4e636576)

src/routes/flush/pipes/invalidation-pattern.dto.validation.pipe.ts
hectorgomezv: Changed in [c8b41c8](https://github.com/5afe/safe-client-gateway-nest/pull/288/commits/c8b41c8d1c87bd4e25d55da26973411044cfafdd)

src/domain/flush/flush.repository.ts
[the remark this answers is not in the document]
hectorgomezv: I created [this issue](https://github.com/5afe/safe-client-gateway-nest/issues/297) to tackle those concerns, if you're OK with it, I'll work on a clearer solution there so we can move forward with this PR 🙂

src/domain/flush/flush.repository.ts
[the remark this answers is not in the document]
hectorgomezv: Cool, let's tackle this in [this issue](https://github.com/5afe/safe-client-gateway-nest/issues/297) if you're OK with it 🙂

src/domain/flush/flush.repository.ts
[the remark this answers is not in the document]
hectorgomezv: This will be addressed in [this issue](https://github.com/5afe/safe-client-gateway-nest/issues/297) also 🙂

src/domain/flush/flush.repository.ts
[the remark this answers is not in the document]
fmrsabino: > Cool, let's tackle this in [this issue](https://github.com/5afe/safe-client-gateway-nest/issues/297) if you're OK with it 🙂

Yeah! Ok from my end! 👍

src/domain/flush/flush.repository.ts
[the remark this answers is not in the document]
fmrsabino: > I created [this issue](https://github.com/5afe/safe-client-gateway-nest/issues/297) to tackle those concerns, if you're OK with it, I'll work on a clearer solution there so we can move forward with this PR 🙂

Totally ok with tackling in a separate PR/discussion! 👍

src/routes/flush/pipes/invalidation-pattern.dto.validation.pipe.ts
fmrsabino: > My idea was to make this `InvalidationTarget` case insensitive, but I tested the current CGW and it is case sensitive (i.e. it requires a capitalized word like `Contracts`, `Tokens`...) so maybe it makes sense to remove the `toUpperCase()` hack and just use the enum values as you commented 👍🏻

For now let's remove it but the `flush` endpoint is mostly used internally so there are no strict requirements to match the previous API (i.e. this is not used by the Safe Transaction Service).

For this one I don't think we should spend a lot of time reworking it but it's something that we can think about in the future. Maybe we end up not needing the endpoint at all in the future (e.g. by having an authenticated redis web interface to perform these type of tasks).

fmrsabino (approved):
🚀 🚀 🚀

PR #290, opened 2023-03-24
Add authorization check to privileged endpoints

hectorgomezv:
This PR adds authentication checks to some privileged endpoints (i.e. cache hooks and cache flushes). To accomplish this task, [the Nest.js recommended approach](https://docs.nestjs.com/security/authentication) was followed. This includes the addition of the `passport` library, which is recommended by Nest.js docs, and provides a pluggable and strategy pattern-based approach to authentication management. 

I've used a custom strategy (`passport-custom`) since I'm assuming we want to keep compatibility with the Rust implementation (which expects a `Basic <token>` header). 

One of the advantages this strategy-based approach gives is that we could use different strategies for different parts of the service API (JWT for example), without too much effort (just plugging an additional strategy subclass and a Nest.js guard).


src/routes/common/auth/basic-auth.guard.ts
```diff
@@ -0,0 +1,5 @@
+import { Injectable } from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+
+@Injectable()
+export class BasicAuthGuard extends AuthGuard('basic-auth') {}
```
fmrsabino, line 5: Have you looked into the `CanActivate` guard?

Maybe this ends up boiling down to the discussion of Authorisation vs Authentication.

It seems that this implementation tries to use an authentication framework (`passport`) to protect the endpoints. However maybe just a simple authorisation guard might be enough? (since we do not have a concept of a user)

Example:

```typescript
@Injectable()
export class BasicAuthGuard implements CanActivate {
  constructor(
  ) {}

  canActivate(context: ExecutionContext): boolean | Promise<boolean> {
    const request = context.switchToHttp().getRequest();
    const token = this.configurationService.getOrThrow('auth.token');
    return request.headers['Authorization'] === `Basic ${token}`;
  }
}
```

```typescript
@UseGuards(BasicAuthGuard)
@Post('/chains/:chainId/hooks/events')
@HttpCode(200)
async postEvent(
  ...
)
```

hectorgomezv: Good point, I also prefer to keep the number of dependencies at a minimum. I didn't explore `CanActivate` since it's more oriented to Authorization as you said, but maybe it makes sense in this use case. I'll try this solution and share the outcome, thanks.

.env.sample
```diff
@@ -22,3 +22,7 @@
 # The default cache expiration time in seconds if none is set
 # (default=60)
 #EXPIRATION_TIME_DEFAULT_SECONDS
+
+# Authorization token to use privileged endpoints.
+# The AUTH_TOKEN should always be set
```
fmrsabino, line 27: If `AUTH_TOKEN` is not setup, we will only get the error once an authorised route is triggered. We should consider using something like `onApplicationBootstrap()` lifecycle events to check if the setup is as expected.

No need to tackle this in this PR 👍

fmrsabino (approved):
🚀 🚀 🚀

.env.sample
hectorgomezv: Yeah, good point, I've just created a new issue to tackle this: https://github.com/5afe/safe-client-gateway-nest/issues/299

PR #291, opened 2023-03-29
Implement POST /preview

hectorgomezv:
This PR adds `POST /v1/chains/:chainId/transactions/:safeAddress/preview`

It's implemented following the requirements for the Rust implementation ([see the original PR](https://github.com/safe-global/safe-client-gateway/pull/924)), but omitting the feature flag.

### DIFF
--- a/src/routes/common/constants.ts
+++ b/src/routes/common/constants.ts
@@ -1,4 +1,1 @@
 export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
-export const MULTI_SEND_METHOD_NAME = 'multiSend';
-export const TRANSACTIONS_PARAMETER_NAME = 'transactions';
-export const ADDRESS_PARAMETER_TYPE = 'address';
--- a/src/routes/common/constants.ts
+++ b/src/routes/common/constants.ts
@@ -1,1 +1,4 @@
 export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
+export const MULTI_SEND_METHOD_NAME = 'multiSend';
+export const TRANSACTIONS_PARAMETER_NAME = 'transactions';
+export const ADDRESS_PARAMETER_TYPE = 'address';
--- a/src/routes/transactions/entities/preview-transaction.dto.entity.ts
+++ b/src/routes/transactions/entities/preview-transaction.dto.entity.ts
@@ -1,12 +1,11 @@
-import { ApiProperty } from '@nestjs/swagger';
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 import { Operation } from '../../../domain/safe/entities/operation.entity';
-import { PreviewTransactionDto as DomainPreviewTransactionDto } from '../../../domain/transactions/entities/preview-transaction.entity';
 
-export class PreviewTransactionDto implements DomainPreviewTransactionDto {
+export class PreviewTransactionDto {
   @ApiProperty()
   to: string;
-  @ApiProperty()
-  data: string;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  data: string | null;
   @ApiProperty()
   value: string;
   @ApiProperty()
--- a/src/routes/transactions/entities/preview-transaction.dto.entity.ts
+++ b/src/routes/transactions/entities/preview-transaction.dto.entity.ts
@@ -1,0 +1,14 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { Operation } from '../../../domain/safe/entities/operation.entity';
+import { PreviewTransactionDto as DomainPreviewTransactionDto } from '../../../domain/transactions/entities/preview-transaction.entity';
+
+export class PreviewTransactionDto implements DomainPreviewTransactionDto {
+  @ApiProperty()
+  to: string;
+  @ApiProperty()
+  data: string;
+  @ApiProperty()
+  value: string;
+  @ApiProperty()
+  operation: Operation;
+}
--- a/src/routes/transactions/entities/transaction-data.entity.ts
+++ b/src/routes/transactions/entities/transaction-data.entity.ts
@@ -1,0 +1,40 @@
+import { ApiExtraModels, ApiProperty } from '@nestjs/swagger';
+import { Operation } from '../../../domain/safe/entities/operation.entity';
+import { AddressInfo } from '../../common/entities/address-info.entity';
+import { DataDecoded } from '../../data-decode/entities/data-decoded.entity';
+
+@ApiExtraModels(AddressInfo, DataDecoded)
+export class TransactionData {
+  @ApiProperty()
+  hexData: string;
+  @ApiProperty()
+  dataDecoded: DataDecoded;
+  @ApiProperty()
+  to: AddressInfo;
+  @ApiProperty()
+  value: string;
+  @ApiProperty()
+  operation: Operation;
+  @ApiProperty()
+  isTrustedDelegateCall: boolean | null;
+  @ApiProperty()
+  addressInfoIndex: Record<string, AddressInfo> | null;
+
+  constructor(
+    hexData: string,
+    dataDecoded: DataDecoded,
+    to: AddressInfo,
+    value: string,
+    operation: Operation,
+    isTrustedDelegateCall: boolean | null,
+    addressInfoIndex: Record<string, AddressInfo> | null,
+  ) {
+    this.hexData = hexData;
+    this.dataDecoded = dataDecoded;
+    this.to = to;
+    this.value = value;
+    this.operation = operation;
+    this.isTrustedDelegateCall = isTrustedDelegateCall;
+    this.addressInfoIndex = addressInfoIndex;
+  }
+}
--- a/src/routes/transactions/entities/transaction-data.entity.ts
+++ b/src/routes/transactions/entities/transaction-data.entity.ts
@@ -1,32 +1,36 @@
-import { ApiExtraModels, ApiProperty } from '@nestjs/swagger';
+import {
+  ApiExtraModels,
+  ApiProperty,
+  ApiPropertyOptional,
+} from '@nestjs/swagger';
 import { Operation } from '../../../domain/safe/entities/operation.entity';
 import { AddressInfo } from '../../common/entities/address-info.entity';
 import { DataDecoded } from '../../data-decode/entities/data-decoded.entity';
 
 @ApiExtraModels(AddressInfo, DataDecoded)
 export class TransactionData {
-  @ApiProperty()
-  hexData: string;
-  @ApiProperty()
-  dataDecoded: DataDecoded;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  hexData: string | null;
+  @ApiPropertyOptional({ type: DataDecoded, nullable: true })
+  dataDecoded: DataDecoded | null;
   @ApiProperty()
   to: AddressInfo;
-  @ApiProperty()
-  value: string;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  value: string | null;
   @ApiProperty()
   operation: Operation;
-  @ApiProperty()
-  isTrustedDelegateCall: boolean | null;
-  @ApiProperty()
+  @ApiPropertyOptional({ type: Boolean, nullable: true })
+  trustedDelegateCallTarget: boolean | null;
+  @ApiPropertyOptional({ type: Object, nullable: true })
   addressInfoIndex: Record<string, AddressInfo> | null;
 
   constructor(
-    hexData: string,
-    dataDecoded: DataDecoded,
+    hexData: string | null,
+    dataDecoded: DataDecoded | null,
     to: AddressInfo,
     value: string,
     operation: Operation,
-    isTrustedDelegateCall: boolean | null,
+    trustedDelegateCallTarget: boolean | null,
     addressInfoIndex: Record<string, AddressInfo> | null,
   ) {
     this.hexData = hexData;
@@ -34,7 +38,7 @@ export class TransactionData {
     this.to = to;
     this.value = value;
     this.operation = operation;
-    this.isTrustedDelegateCall = isTrustedDelegateCall;
+    this.trustedDelegateCallTarget = trustedDelegateCallTarget;
     this.addressInfoIndex = addressInfoIndex;
   }
 }
--- a/src/routes/transactions/entities/transaction-preview.entity.ts
+++ b/src/routes/transactions/entities/transaction-preview.entity.ts
@@ -1,9 +1,26 @@
-import { ApiProperty } from '@nestjs/swagger';
+import { ApiExtraModels, ApiProperty, getSchemaPath } from '@nestjs/swagger';
+import { CreationTransactionInfo } from './creation-transaction-info.entity';
+import { CustomTransactionInfo } from './custom-transaction.entity';
+import { SettingsChangeTransaction } from './settings-change-transaction.entity';
 import { TransactionData } from './transaction-data.entity';
 import { TransactionInfo } from './transaction-info.entity';
+import { TransferTransactionInfo } from './transfer-transaction-info.entity';
 
+@ApiExtraModels(
+  CreationTransactionInfo,
+  CustomTransactionInfo,
+  SettingsChangeTransaction,
+  TransferTransactionInfo,
+)
 export class TransactionPreview {
-  @ApiProperty()
+  @ApiProperty({
+    oneOf: [
+      { $ref: getSchemaPath(CreationTransactionInfo) },
+      { $ref: getSchemaPath(CustomTransactionInfo) },
+      { $ref: getSchemaPath(SettingsChangeTransaction) },
+      { $ref: getSchemaPath(TransferTransactionInfo) },
+    ],
+  })
   txInfo: TransactionInfo;
   @ApiProperty()
   txData: TransactionData;
--- a/src/routes/transactions/entities/transaction-preview.entity.ts
+++ b/src/routes/transactions/entities/transaction-preview.entity.ts
@@ -1,0 +1,15 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { TransactionData } from './transaction-data.entity';
+import { TransactionInfo } from './transaction-info.entity';
+
+export class TransactionPreview {
+  @ApiProperty()
+  txInfo: TransactionInfo;
+  @ApiProperty()
+  txData: TransactionData;
+
+  constructor(txInfo: TransactionInfo, txData: TransactionData) {
+    this.txInfo = txInfo;
+    this.txData = txData;
+  }
+}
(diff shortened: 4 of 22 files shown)


src/routes/common/constants.ts
```diff
@@ -1 +1,4 @@
 export const NULL_ADDRESS = '0x0000000000000000000000000000000000000000';
+export const MULTI_SEND_METHOD_NAME = 'multiSend';
+export const TRANSACTIONS_PARAMETER_NAME = 'transactions';
+export const ADDRESS_PARAMETER_TYPE = 'address';
```
hectorgomezv, line 4: Just moving constants here since they are used in several places now.

src/routes/transactions/transactions.module.ts
```diff
@@ -38,9 +41,11 @@ import { CreationTransactionMapper } from './mappers/creation-transaction/creati
     MultisigTransactionMapper,
     MultisigTransactionStatusMapper,
     NativeCoinTransferMapper,
-    SafeAppInfoMapper,
     QueuedItemsMapper,
+    SafeAppInfoMapper,
```
hectorgomezv, line 45: Just ordering items alphabetically here.

src/routes/transactions/mappers/common/data-decoded-param.helper.spec.ts
```diff
@@ -400,4 +400,223 @@ describe('DataDecoded param helper (Unit)', () => {
       expect(fromParam).toBe('fallback');
     });
   });
+
+  describe('hasNestedDelegate function tests', () => {
```
fmrsabino, line 404: nitpick: remove "function tests" from the name. It's too specific to the "implementation" in my opinion. Can be just `hasNestedDelegate` or `hasNestedDelegate()`

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -700,6 +701,289 @@ describe('Transactions Controller (Unit)', () => {
...
+          return Promise.resolve({ data: dataDecodedResponse });
+        }
+        return Promise.reject(new Error(`Could not match ${url}`));
+      });
+      await request(app.getHttpServer())
+        .post(`/v1/chains/${chainId}/transactions/${safeAddress}/preview`)
+        .send(previewTransactionDto)
+        .expect(200)
+        .expect(({ body }) => {
+          expect(body).toMatchObject({
```
fmrsabino, line 752: Why do we use `toMatchObject` here? It seems that it's missing just two properties but is it because of that?

fmrsabino (changes requested):

src/routes/transactions/entities/transaction-data.entity.ts
```diff
@@ -0,0 +1,40 @@
...
+@ApiExtraModels(AddressInfo, DataDecoded)
+export class TransactionData {
+  @ApiProperty()
+  hexData: string;
+  @ApiProperty()
+  dataDecoded: DataDecoded;
+  @ApiProperty()
+  to: AddressInfo;
+  @ApiProperty()
+  value: string;
```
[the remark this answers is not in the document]
hectorgomezv: You mean the `hexData` field, right? This contains the value sent by the client into the `value` field of the payload, which is mandatory. The nullable field is `data`, which gets mapped to `hexData`.

src/routes/transactions/mappers/transaction-preview.mapper.ts
```diff
@@ -0,0 +1,59 @@
...
+      dataDecoded = await this.dataDecodedService.getDataDecoded(
+        chainId,
+        new GetDataDecodedDto(
+          previewTransactionDto.data,
+          previewTransactionDto.to,
+        ),
+      );
+    } catch (error) {
+      winston.warn(
+        `Error trying to decode the input data: ${JSON.stringify(error)}`,
```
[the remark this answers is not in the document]
hectorgomezv: If there's an error, it should be a `DataSourceError`. I gave this a second thought and maybe we don't need to catch the error here since it would go through the error pipeline to the logs as any other error. This will simplify this function as well.

src/routes/transactions/entities/transaction-data.entity.ts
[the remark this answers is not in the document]
fmrsabino: I mean the `value`: https://github.com/safe-global/safe-client-gateway/blob/c49c4446de49301dcbefe0f0b33378dc79b769f2/src/routes/transactions/models/details.rs#L292

If the entity that this payload represents is used elsewhere I would try to reuse it. The concepts should be mostly the same.

src/routes/transactions/entities/transaction-data.entity.ts
```diff
@@ -0,0 +1,40 @@
...
+  @ApiProperty()
+  operation: Operation;
+  @ApiProperty()
```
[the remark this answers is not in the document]
hectorgomezv: No, we need to use `@ApiPropertyOptional({ type: Boolean, nullable: true })` here, thanks for noticing, I'll change it.

src/routes/transactions/mappers/common/data-decoded-param.helper.ts
```diff
@@ -65,4 +66,13 @@ export class DataDecodedParamHelper {
     if (!dataDecoded || !dataDecoded.parameters?.length) return null;
     return dataDecoded.parameters[position]?.value ?? null;
   }
+
+  hasNestedDelegate(dataDecoded: DataDecoded): boolean {
+    if (!dataDecoded.parameters) return false;
+    return dataDecoded.parameters?.some(
+      (param) =>
+        isArray(param.valueDecoded) &&
+        param.valueDecoded.some((innerParam) => innerParam?.operation === 1),
```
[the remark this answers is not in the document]
hectorgomezv: Agree, the problem with that is `operation.entity.ts` contains Typescript types that can't be used as values at runtime if I'm not wrong. We would need to refactor that or create additional constants which could be used later in the types. We could solve this by exporting real (runtime-accessible) JS constants like these:

```javascript
export const CALL_VALUE = 0;
export const DELEGATE_VALUE = 1;

export type Call = typeof CALL_VALUE;
export type Delegate = typeof DELEGATE_VALUE;
export type Operation = Call | Delegate;

```

src/routes/transactions/entities/preview-transaction.dto.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { Operation } from '../../../domain/safe/entities/operation.entity';
+import { PreviewTransactionDto as DomainPreviewTransactionDto } from '../../../domain/transactions/entities/preview-transaction.entity';
+
+export class PreviewTransactionDto implements DomainPreviewTransactionDto {
+  @ApiProperty()
+  to: string;
+  @ApiProperty()
+  data: string;
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

src/routes/common/constants.ts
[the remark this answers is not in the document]
hectorgomezv: I've moved them to `src/routes/transactions/constants.ts` in [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d), let me know if you think there's a better place 🙂

src/routes/transactions/mappers/transaction-preview.mapper.ts
```diff
@@ -0,0 +1,59 @@
...
+    private readonly transactionDataMapper: TransactionDataMapper,
+    private readonly dataDecodedService: DataDecodedService,
+  ) {}
+
+  async mapTransactionPreview(
+    chainId: string,
+    safe: Safe,
+    previewTransactionDto: PreviewTransactionDto,
+  ): Promise<TransactionPreview> {
+    let dataDecoded;
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

src/routes/transactions/mappers/transaction-preview.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: Changed in [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

src/routes/transactions/entities/transaction-data.entity.ts
```diff
@@ -0,0 +1,40 @@
...
+import { Operation } from '../../../domain/safe/entities/operation.entity';
+import { AddressInfo } from '../../common/entities/address-info.entity';
+import { DataDecoded } from '../../data-decode/entities/data-decoded.entity';
+
+@ApiExtraModels(AddressInfo, DataDecoded)
+export class TransactionData {
+  @ApiProperty()
+  hexData: string;
+  @ApiProperty()
+  dataDecoded: DataDecoded;
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

src/routes/transactions/entities/transaction-data.entity.ts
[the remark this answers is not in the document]
hectorgomezv: Changed in [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

src/routes/transactions/mappers/common/data-decoded-param.helper.ts
```diff
@@ -65,4 +66,13 @@ export class DataDecodedParamHelper {
     if (!dataDecoded || !dataDecoded.parameters?.length) return null;
     return dataDecoded.parameters[position]?.value ?? null;
   }
+
+  hasNestedDelegate(dataDecoded: DataDecoded): boolean {
+    if (!dataDecoded.parameters) return false;
+    return dataDecoded.parameters?.some(
```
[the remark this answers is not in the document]
hectorgomezv: [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

src/routes/transactions/mappers/common/data-decoded-param.helper.spec.ts
hectorgomezv: [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

src/routes/transactions/entities/transaction-preview.entity.ts
```diff
@@ -0,0 +1,15 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { TransactionData } from './transaction-data.entity';
+import { TransactionInfo } from './transaction-info.entity';
+
+export class TransactionPreview {
+  @ApiProperty()
+  txInfo: TransactionInfo;
```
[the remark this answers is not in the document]
hectorgomezv: Good point, in fact, in `transaction.entity.ts` we're missing `CreationTransactionInfo` as well as extension type of `TransactionInfo`. I've changed `transaction-preview.entity.ts` in [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d) and I'm going to fix the other one in a separate PR if you're OK with that 🙂

src/routes/transactions/mappers/transaction-preview.mapper.ts
```diff
@@ -0,0 +1,59 @@
...
+import { TransactionPreview } from '../entities/transaction-preview.entity';
+import { TransactionDataMapper } from './common/transaction-data.mapper';
+import { MultisigTransactionInfoMapper } from './common/transaction-info.mapper';
+
+@Injectable()
+export class TransactionPreviewMapper {
+  constructor(
+    private readonly transactionInfoMapper: MultisigTransactionInfoMapper,
+    private readonly transactionDataMapper: TransactionDataMapper,
+    private readonly dataDecodedService: DataDecodedService,
```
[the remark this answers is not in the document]
hectorgomezv: Yes, I agree, I'll move the shared logic from `DataDecodedService` to the domain (to `DataDecodedRepository`).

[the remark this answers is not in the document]
hectorgomezv: In fact, I've removed the additional logic that `DataDecodedService` had, since it was just redundant mapping logic, the domain and route objects map 1:1 in this case: [e1b9fb3](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/e1b9fb3c1c746a4879855e3b7ad35002eb9dd716)

src/routes/transactions/mappers/common/transaction-data.mapper.spec.ts
```diff
@@ -0,0 +1,525 @@
...
+const contractsRepository = jest.mocked({
+  getContract: jest.fn(),
+} as unknown as ContractsRepository);
+
+const dataDecodedParamHelper = jest.mocked({
+  hasNestedDelegate: jest.fn(),
+} as unknown as DataDecodedParamHelper);
+
+describe('Transaction Data Mapper (Unit)', () => {
+  const mapper = new TransactionDataMapper(
```
fmrsabino, line 28: We should move this to `beforeEach` so that we can avoid potential state sharing between the tests in the future

src/routes/transactions/mappers/common/transaction-data.mapper.ts
```diff
@@ -0,0 +1,170 @@
...
+   * is trusted, and the {@link DataDecoded} received contains a nested
+   * DELEGATE operation, then true is returned.
+   * Otherwise the function will return false.
+   */
+  async isTrustedDelegateCall(
+    chainId: string,
+    operation: Operation,
+    to: string,
+    dataDecoded: DataDecoded | null,
+  ): Promise<boolean | null> {
```
fmrsabino, line 82: We should reevaluate the return of this function. `isX` hints at a boolean return type or a future which provides a boolean. However we have 3 states being represented here: `true | false | null`. If we find that we really need 3 states I'd suggest representing those with an enum or similar.

This doesn't need to be tackled in this PR.

src/routes/transactions/mappers/common/transaction-data.mapper.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (dataDecoded === null || !isArray(dataDecoded.parameters)) return {};
+
+    const addressInfos = await Promise.all(
+      this.getAddressParametersFromDataDecoded(dataDecoded).map((param) =>
+        this.getIfValidAddress(chainId, param),
+      ),
+    );
+
+    return addressInfos
+      .filter((addressInfo) => !isNull(addressInfo))
```
fmrsabino, line 119: Do we need to `filter` here? In the `reduce` function we are also tackling the nullability of `addressInfo` 🤔

src/routes/transactions/transactions.controller.spec.ts
```diff
@@ -700,6 +705,296 @@ describe('Transactions Controller (Unit)', () => {
...
+          return Promise.reject({ error: 'Data cannot be decoded' });
+        }
+        return Promise.reject(new Error(`Could not match ${url}`));
+      });
+      await request(app.getHttpServer())
+        .post(`/v1/chains/${chainId}/transactions/${safeAddress}/preview`)
+        .send(previewTransactionDto)
+        .expect(200)
+        .expect(({ body }) => {
+          expect(body).toEqual({
```
fmrsabino, line 887: `toEqual` doesn't seem required. We can test the `body` directly right?

fmrsabino (changes requested):
Nice 👏 Just some smaller comments!

src/routes/transactions/transactions.controller.spec.ts
fmrsabino: > Yes, it was used to abstract from those two property values in the tests which are not strictly related to them. But I've changed the tests to have the complete expected payload (setting more values into `previewTransactionDtoBuilder` call) so the tests are a bit more strict: [a158c2a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/a158c2a02ceed6346e496ef948aac0185e20225d)

I think the way it was before was also ok! Just clarifying if that was the case. I think we should use `toMatchObject` if we already test the complete payload in some other test (this reduces the size of the test even though not by a lot in this case) 👍

src/routes/transactions/transactions.controller.spec.ts
hectorgomezv: Right! Thanks. Changed in [9565c0a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/9565c0ad8b6c6533745567f36f005f4915859bf4)

src/routes/transactions/mappers/common/transaction-data.mapper.ts
hectorgomezv: True, we were double-checking it, thanks. Changed in [9565c0a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/9565c0ad8b6c6533745567f36f005f4915859bf4)

src/routes/transactions/mappers/common/transaction-data.mapper.spec.ts
hectorgomezv: Yes, I agree. In this case, the mapper instance is built just using mocks, so they are reset for each test, but it's a best practice and also covers the case the mapper actually contains state in the future 👍🏻 I've changed it in [9565c0a](https://github.com/5afe/safe-client-gateway-nest/pull/291/commits/9565c0ad8b6c6533745567f36f005f4915859bf4)

This could be applied to other parts of the code, I'm going to address those in a separate PR.

src/routes/transactions/mappers/common/transaction-data.mapper.ts
fmrsabino: Yeah exactly 👍 Even if the clients use `true | false | null` we can map this on the route level and work with a more meaningful domain representation.

src/routes/transactions/mappers/common/transaction-data.mapper.spec.ts
fmrsabino: Let's do it! 👍 

This is just to avoid breaking the tests if any internal state is introduced. It's can be trickier to debug those cases then.

fmrsabino (approved):
🚀 🚀 🚀

src/routes/transactions/mappers/common/transaction-data.mapper.spec.ts
hectorgomezv: https://github.com/5afe/safe-client-gateway-nest/pull/301 🙂

PR #294, opened 2023-04-10
Unify cache keys

hectorgomezv:
This PR contains just a simple change in a cache name, but the intention is to discuss the keys we want to use for the different features of the service.

I have some doubts about whether `all_transactions` and `owner_safes` are descriptive names, what do you think @fmrsabino?

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -11,7 +11,7 @@ export class CacheRouter {
   private static readonly CREATION_TRANSACTION_KEY = 'creation_transaction';
   private static readonly DELEGATES_KEY = 'delegates';
   private static readonly INCOMING_TRANSFERS_KEY = 'incoming_transfers';
-  private static readonly MASTER_COPIES_KEY = 'master-copies';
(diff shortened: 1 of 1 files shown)


fmrsabino (approved):
🚀

fmrsabino:
> I have some doubts about whether `all_transactions` and `owner_safes` are descriptive names, what do you think @fmrsabino?

Regarding `all_transactions` I think it's descriptive: `<chainId>_<safeAddress>_all_transactions`.

`owner_safes` does sound a little bit more cryptic: `<chainId>_<ownerAddress>_owner_safes`. An alternative can be `<chainId>_<ownerAddress>_safes` but I don't have a strong opinion on this one.

hectorgomezv:
> owner_safes does sound a little bit more cryptic: <chainId>_<ownerAddress>_owner_safes. An alternative can be <chainId>_<ownerAddress>_safes but I don't have a strong opinion on this one.

Yes, I agree but I'd keep it for now, `*_safes` sounds good but maybe it's a bit generic as well, we have an issue to refactor the cache directions so we could rethink it there also.

PR #295, opened 2023-04-10
Cache 404 errors

hectorgomezv:
**Motivation**
This PR tries to avoid the following scenario:

```
debug: [Cache] Cache miss: 5_0x710085cB90e4Ce09F80E354Cf579a32C9790A140_contract
debug: [Cache] Cache miss: 5_0x710085cB90e4Ce09F80E354Cf579a32C9790A140_contract
debug: [Cache] Cache miss: 5_0x40A2aCCbd92BCA938b02010E17A5b8929b49130D_contract
debug: [Cache] Cache miss: 5_0xf10E2042ec19747401E5EA174EfB63A0058265E6_contract
debug: [Cache] Cache miss: 5_0x7f3774E7D0e410444b6f19a10D50Cd34857Bb236_contract
debug: [Cache] Cache miss: 5_0x7f3774E7D0e410444b6f19a10D50Cd34857Bb236_contract
debug: [Cache] Cache miss: 5_0x7f3774E7D0e410444b6f19a10D50Cd34857Bb236_contract
...
error: An error occurred {"code":404}
error: An error occurred {"code":404}
error: An error occurred {"code":404}
error: An error occurred {"code":404}
error: An error occurred {"code":404}
error: An error occurred {"code":404}
error: An error occurred {"code":404}
...
```

i.e: Avoid making successive calls to the Transaction Service asking for the same item.

This is common in the current flow since a list of transactions could contain references to the same contract for each transaction. The current behavior is to repeat the call to the Transaction Service N times for N contracts, whether the contract is found or it doesn't.

So this PR:
- Modifies the `get` function of the `CacheFirstDataSource` class to cache 404 errors. The errors are cached with a fixed TTL of 30 seconds in this first approach. Optionally in the future, we could want to externalize this configuration and place this TTL value behind an environment variable.


src/domain/chains/entities/__tests__/chain.builder.ts
```diff
@@ -10,7 +10,7 @@ import { rpcUriBuilder } from './rpc-uri.builder';
 
 export function chainBuilder(): IBuilder<Chain> {
   return Builder.new<Chain>()
-    .with('chainId', faker.random.numeric())
+    .with('chainId', faker.random.numeric(6))
```
hectorgomezv, line 13: This change is motivated because I was getting random errors (flaky tests) due to the cache contents being different from the expected ones. This was caused because `faker.random.numeric` only produces one digit by default, so sometimes two sequential calls to `chainBuilder()` produce the same digit, and the cache direction gets repeated. So only one chain was being cached (the same cache identifier was produced twice).

To avoid this situation, the simplest fix is increasing the entropy and making `faker.random.numeric` to produce 6 digits. Collisions are very rare that way, but we could also think on producing an uuid here, even if this would make our `chainBuilder` differ from the production behaviour, where chainIds are integers.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -144,7 +144,7 @@ export class TransactionApi implements ITransactionApi {
         contractAddress,
       );
       const url = `${this.baseUrl}/api/v1/contracts/${contractAddress}`;
-      return await this.dataSource.get(cacheDir, url);
+      return await this.dataSource.getAndCacheError(cacheDir, url);
```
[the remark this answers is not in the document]
fmrsabino: I think the added methods add complexity to the relatively simple API that we currently have. I think we should clarify if we want to cache 404s and if we want to cache 404s would it be for every single endpoint?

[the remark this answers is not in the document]
hectorgomezv: I unified the condition of both functions in [bf97b75](https://github.com/safe-global/safe-client-gateway-nest/pull/295/commits/bf97b75b8dc54807875a63f8b17f9c3884f9ce12)

I agree, it doesn't make sense to have different conditions here. I chose `cached != null` because this covers the case the cached data is both `null` or `undefined`, but it discards the case the cached data is any other _falsy_ value.

hectorgomezv:
@fmrsabino I've made the changes we discussed to only cache 404 errors with a 30 secs TTL. I've also adapted the PR description.

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -45,16 +47,42 @@ export class CacheFirstDataSource {
...
-    await this.cacheService.set(
-      new CacheDir(cacheDir.key, cacheDir.field),
-      rawJson,
-      expireTimeSeconds,
-    );
-    return data;
+  }
+
+  private isNotFoundError(data: unknown) {
+    const error = data as NetworkResponseError;
```
fmrsabino, line 71: Out of curiosity, what is your opinion on trying to get the expected field without the cast?

Since `NetworkResponseError` is not the only error that can be thrown maybe we should try to get the expected field without specifying the class itself?

fmrsabino (changes requested):
Nice! 👏 

I'd leave the `contracts.controller.spec.ts` out of this set of changes. These tests can be added separately and I don't think they are directly related with caching 404s. They might be affected by this change but in this PR we added the entire test suite for the Contracts.

Testing that the `CacheFirstDataSource` works as intended (TTL and if it caches only with 404s is a bigger priority)

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -45,16 +47,42 @@ export class CacheFirstDataSource {
...
-    const rawJson = JSON.stringify(data);
-    await this.cacheService.set(
-      new CacheDir(cacheDir.key, cacheDir.field),
-      rawJson,
-      expireTimeSeconds,
-    );
-    return data;
+  }
+
+  private isNotFoundError(data: unknown) {
```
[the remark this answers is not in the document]
hectorgomezv: What type do you think you could use as return type in the type predicate? I mean, this function should verify the error data (`status` is 404), not the type, right? Not sure I'm following you here.

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -19,6 +20,8 @@ import { CacheDir } from './entities/cache-dir.entity';
  */
 @Injectable()
 export class CacheFirstDataSource {
+  private readonly ERROR_TTL_SECONDS = 30;
```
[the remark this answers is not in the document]
hectorgomezv: I included a `cache.first.data.source.spec.ts` class testing this in [8733756](https://github.com/safe-global/safe-client-gateway-nest/pull/295/commits/873375691e18ef8fd3589d301ebb2775bebd8b7e)

hectorgomezv:
> I'd leave the contracts.controller.spec.ts out of this set of changes. These tests can be added separately and I don't think they are directly related with caching 404s. They might be affected by this change but in this PR we added the entire test suite for the Contracts.

Yes, good point, I've removed them from this PR in [8733756](https://github.com/safe-global/safe-client-gateway-nest/pull/295/commits/873375691e18ef8fd3589d301ebb2775bebd8b7e), I'll open a new PR including with those tests 🙂

src/datasources/cache/cache.first.data.source.ts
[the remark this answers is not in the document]
fmrsabino: With the new changes I don't think this is currently necessary but I was thinking along the following lines:

```typescript
private isNotFoundError(data: unknown): data is { status: 404 } {
  return get(data, 'status') === 404;
}
```

Using this in an `if` expression should make `status` available and with the value of `404`

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+import { ICacheService } from './cache.service.interface';
+import { CacheDir } from './entities/cache-dir.entity';
+import { fakeCacheService } from './__tests__/test.cache.module';
+
+describe('CacheFirstDataSource', () => {
+  let cacheFirstDataSource: CacheFirstDataSource;
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    fakeCacheService.clear();
```
fmrsabino, line 14: Why not create a `FakeCacheService` here instead of using a global one?

Actually this can lead to bugs and we should change all the instances in the future. Access to global components should be done via the module and not via an exported const.

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    expect(mockNetworkService.get).toHaveBeenCalledTimes(1);
+    expect(fakeCacheService.keyCount()).toBe(1);
+  });
+
+  it('should cache errors with the expected TTL', async () => {
+    const mockCache = jest.mocked({
+      get: jest.fn(),
+      set: jest.fn(),
+      delete: jest.fn(),
+    } as unknown as ICacheService);
```
fmrsabino, line 108: Suggestion: maybe we can have a `getTTL` on the `FakeCacheService` so you wouldn't need to mock another object just for this test.

Let me know what you think!

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -45,16 +48,36 @@ export class CacheFirstDataSource {
     const cached = await this.cacheService.get(cacheDir);
     if (cached != null) {
       winston.debug(`[Cache] Cache hit: [${cacheDir.key}, ${cacheDir.field}]`);
-      return JSON.parse(cached);
+      const data = JSON.parse(cached);
+      if (this.isNotFoundError(data)) {
+        throw new NetworkResponseError(404);
```
fmrsabino, line 53: Something to consider: if we cache 404 errors this way we just store the status code right? We loose any additional message that might come with the payload.

fmrsabino (changes requested):
Some smaller comments! Not all need to be tackled here.

The addition of `cache.first.data.source.spec.ts` is really valuable! 🙌

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: Do you mean having something like getting the TTL given a key (similar to redis [TTL function](https://redis.io/commands/ttl/))? That would be nice, I had been thinking about something like that also, but I suspect we would need to change the data representation in `FakeCacheService` to save the TTL somewhere, right? So in that case `Record<string, Record<string, any>>` doesn't fit anymore and we would need something like `Record<Record<string, number>, Record<string, any>>`, or use tuples, to store the TTL there. And this would affect the current `FakeCacheService` functions as well, so I decided to simply use the mock approach in the test class. Do you think is it better to go through the other path?

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: Yes, I agree having a common instance exported by a constant to multiple places is not very convenient. This affects many test classes, I think this could be tackled separately if you agree 🙂

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: I've changed it to store any additional data coming from the upstream service in [670d495](https://github.com/safe-global/safe-client-gateway-nest/pull/295/commits/670d49548487aa23e218fb2c3def0f8e70f0593a)

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: Yes, I agree having a common instance exported by a constant to multiple places is not very convenient. This affects many test classes, I think this could be tackled separately if you agree 🙂

src/datasources/cache/cache.first.data.source.ts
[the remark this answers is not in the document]
hectorgomezv: Yes, maybe it's better that way, so we can get rid of hardcoding the 404 twice. Changed in [670d495](https://github.com/safe-global/safe-client-gateway-nest/pull/295/commits/670d49548487aa23e218fb2c3def0f8e70f0593a) 👍🏻

src/datasources/cache/cache.first.data.source.spec.ts
fmrsabino: > Yes, I agree having a common instance exported by a constant to multiple places is not very convenient. This affects many test classes, I think this could be tackled separately if you agree 🙂

There are two scenarios in most of our tests:
- Reference the `fakeCacheService` via a global reference _because_ that's one (not so good) way to get the reference to the cache that will be used by the app module
- Reference to the `fakeCacheService` directly – use the reference to build the target class being tested – this is the use case of this test and we should address it here. 

If we start getting the cache instance via the module it won't change this test because there are no modules involved.

src/datasources/cache/cache.first.data.source.spec.ts
fmrsabino: > 

Ok since we have a single test that checks the TTL at the moment I would say that this change is not needed right now and we can tackle it later if we write more tests around the TTL part 👍

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: Makes sense, thanks, changing it in [7e82cff](https://github.com/safe-global/safe-client-gateway-nest/pull/295/commits/7e82cff13ee7845879c4cb33aee1317a83082372)

fmrsabino (approved):
🚀

PR #305, opened 2023-04-26
Refactor cache directions, implement pattern invalidations

### LINKED ISSUE #297: Refactor cache directions to partition by chain
Please refer to https://github.com/5afe/safe-client-gateway-nest/pull/288#discussion_r1164138082 for context

**Goals**
- Partition first-level cache keys by feature and chain.
- Use `SCAN` to allow selective deletions on the first level (by chain) without harming the performance.
- Clarify the usage of `CacheDir` for deletions.

hectorgomezv:
Closes #297 

This PR:
- Changes the cache data storage directions format. This PR unifies the _key_ format to `$chainId_$feature_$additional-identifier(s)`.
- Derived from the previous point, `SCAN` can be used to delete _by feature_ (for example, to delete all the transactions/contracts/tokens/etc. belonging to one/all chains) without the need of knowing the configured chains beforehand.

### DIFF
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -12,9 +12,19 @@ export class FakeCacheService implements ICacheService {
     this.cache = {};
   }
 
-  delete(cacheDir: CacheDir): Promise<number> {
-    delete this.cache[cacheDir.key];
+  deleteByKey(key: string): Promise<number> {
+    delete this.cache[key];
     return Promise.resolve(1);
+  }
+
+  deleteByKeyPattern(pattern: string): Promise<void> {
+    const patternRegex = RegExp(pattern.replace('*', '.*'));
+    for (const key in this.cache) {
+      if (patternRegex.test(key)) {
+        delete this.cache[key];
+      }
+    }
+    return Promise.resolve();
   }
 
   get(cacheDir: CacheDir): Promise<string | undefined> {
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -24,6 +24,10 @@ export class CacheRouter {
   private static readonly TOKENS_KEY = 'tokens';
   private static readonly TRANSFERS_KEY = 'transfers';
 
+  static getBalancesCacheKey(chainId: string, safeAddress: string): string {
+    return `${chainId}_${CacheRouter.BALANCES_KEY}_${safeAddress}`;
+  }
+
   static getBalanceCacheDir(
     chainId: string,
     safeAddress: string,
@@ -31,14 +35,14 @@ export class CacheRouter {
     excludeSpam?: boolean,
   ): CacheDir {
     return new CacheDir(
-      `${chainId}_${safeAddress}_${CacheRouter.BALANCES_KEY}`,
+      CacheRouter.getBalancesCacheKey(chainId, safeAddress),
       `${trusted}_${excludeSpam}`,
     );
   }
 
   static getSafeCacheDir(chainId: string, safeAddress: string): CacheDir {
     return new CacheDir(
-      `${chainId}_${safeAddress}_${CacheRouter.SAFE_KEY}`,
+      `${chainId}_${CacheRouter.SAFE_KEY}_${safeAddress}`,
       '',
     );
   }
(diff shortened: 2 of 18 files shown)


src/datasources/cache/cache.service.interface.ts
```diff
@@ -11,5 +11,7 @@ export interface ICacheService {
 
   get(cacheDir: CacheDir): Promise<string | undefined>;
 
-  delete(cacheDir: CacheDir): Promise<number>;
+  deleteByKey(key: string): Promise<number>;
+
+  deleteByKeyPattern(pattern: string): Promise<void>;
```
hectorgomezv, line 16: I changed `delete` function name to `deleteByKey` to be more explicit on the expected argument and the purpose of the function. Previously this `delete` function was accepting a `CacheDir` instead of a `string`. Here I think there is no perfect solution:
- If a `CacheDir` is accepted, then the client is forced to send a `CacheDir` which will have `key` and `field`. The `delete` function will delete by `key` and ignore the `field`, so the purpose of the `delete` function could be obscure to the client using the class API (will be forced to set an empty `field`). Anything could be placed in the `field` since it will get ignored by the `delete` function which is very weird from the client's point of view.
- If a `string` is accepted, the purpose is clearer, but it makes the class API a bit odd (some functions accept `CacheDir` and others accept `string`), and the internal structure of `CacheDir` is exposed.
I sightly prefer the second approach based on its clarity, but of course, this is an open matter.

src/datasources/cache/cache.router.ts
```diff
@@ -24,41 +24,46 @@ export class CacheRouter {
   private static readonly TOKENS_KEY = 'tokens';
   private static readonly TRANSFERS_KEY = 'transfers';
 
+  static getBalanceCacheKey(chainId: string, safeAddress: string): string {
+    return `${CacheRouter.BALANCES_KEY}_${chainId}_${safeAddress}`;
```
hectorgomezv, line 28: @fmrsabino I tried to maintain a common format for first-level cache keys (`$feature_$chainId`), but this is an exception to that rule. The reason is that according to `cache-hooks.service.ts`, it should be possible to invalidate the balances **by address**, and redis [does not implement sub-key expirations](https://github.com/redis/redis/issues/167), so the options I see here are:
- Use `balances_$chainId` as key and Implement a custom logic to search through the fields using `HSCAN` to retrieve and delete fields matching `$safeAddress_*`.
- Use `balances_$chainId` as key and just delete all the chain balances at once.
- Make an exception and use `balances_$chainId_$address` as key.

I took the last option, but of course is open for discussion. Or if you see more options they will be welcomed 🙂

src/datasources/cache/cache.service.interface.ts
fmrsabino: Let's do as you suggested (2nd approach). Any potential improvements would probably require some work on `CacheDir` itself.

hectorgomezv:
> What do you think?

Yes, I agree, having independent TTLs is the priority, in fact, as we discussed, we already have this problem on _main_ branch now, so I will modify this PR to re-introduce these identity discriminators back to the `key` part of all the `CacheDirs`, and just use `deleteByPattern` as a contingency solution to implement the `/flush` endpoint while we work on the definitive solution 👍🏻

I'll ping you back when it is changed, thanks 🙂

src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -12,11 +12,18 @@ export class FakeCacheService implements ICacheService {
...
 
-  delete(cacheDir: CacheDir): Promise<number> {
-    delete this.cache[cacheDir.key];
+  deleteByKey(key: string): Promise<number> {
+    delete this.cache[key];
     return Promise.resolve(1);
   }
 
+  deleteByKeyPattern(pattern: string): Promise<void> {
+    this.cache = Object.entries(this.cache)
```
fmrsabino, line 21: I don't think we should consider immutability at this level (even on the real cache). We can consider the cache to be mutable (there's a slight performance cost if we consider it otherwise).

Something like the following seems clearer what do you think?

```typescript
for (const key in Object.keys(this.cache)) {
  if (patternRegex.test(key)) {
    delete this.cache[key];
  }
}
```

fmrsabino (commented):
Nice! Just one small comment.

src/datasources/cache/__tests__/fake.cache.service.ts
hectorgomezv: Thanks, yes, I think it's better to keep it simple here and consider it mutable. I've changed the code to meet your suggestion, with a subtle change:
```javascript
for (const key in this.cache) {
  if (patternRegex.test(key)) {
    delete this.cache[key];
  }
}
```
Because `for...in` already iterates over the keys of the object, so `Object.keys()` is not necessary in this case (it would be on a `for...of` loop) 🙂

fmrsabino (approved):
🚀

PR #307, opened 2023-04-27
Validate mandatory environment at startup

hectorgomezv:
This PR aims to check the required environment is present at the service startup time, preventing the application from unexpected failures at runtime.

The official Nest.js documentation talks about this use case [here](https://docs.nestjs.com/techniques/configuration#schema-validation). The approach taken was heavily inspired by this documentation but AJV was used (which is already a service dependency) instead of Joi for configuration validation.

### DIFF
--- a/src/config/configuration.validator.spec.ts
+++ b/src/config/configuration.validator.spec.ts
@@ -1,0 +1,33 @@
+import { faker } from '@faker-js/faker';
+import { validate } from './configuration.validator';
+
+describe('Configuration validator', () => {
+  const { NODE_ENV } = process.env;
+
+  afterAll(() => {
+    process.env.NODE_ENV = NODE_ENV;
+  });
+
+  it('should bypass this validation on tests', () => {
+    process.env.NODE_ENV = 'test';
+    const expected = JSON.parse(faker.datatype.json());
+    const validated = validate(expected);
+    expect(validated).toBe(expected);
+  });
+
+  it('should detect a malformed configuration in production environment', () => {
+    process.env.NODE_ENV = 'production';
+    expect(() => validate(JSON.parse(faker.datatype.json()))).toThrow();
+  });
+
+  it('should return the input configuration if validated in production environment', () => {
+    process.env.NODE_ENV = 'production';
+    const expected = {
+      ...JSON.parse(faker.datatype.json()),
+      AUTH_TOKEN: 'foo',
+      EXCHANGE_API_KEY: 'bar',
+    };
+    const validated = validate(expected);
+    expect(validated).toBe(expected);
+  });
+});
--- a/src/config/configuration.validator.spec.ts
+++ b/src/config/configuration.validator.spec.ts
@@ -2,12 +2,6 @@ import { faker } from '@faker-js/faker';
 import { validate } from './configuration.validator';
 
 describe('Configuration validator', () => {
-  const { NODE_ENV } = process.env;
-
-  afterAll(() => {
-    process.env.NODE_ENV = NODE_ENV;
-  });
-
   it('should bypass this validation on tests', () => {
     process.env.NODE_ENV = 'test';
     const expected = JSON.parse(faker.datatype.json());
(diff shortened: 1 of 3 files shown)


hectorgomezv:
In the Nest.js documentation, the validation is executed against **the environment variables**, by passing a custom `validate` function to the `ConfigModule` from  `@nestjs/config`. I had some options here:
- Check the environment variables that don't have default values (current approach).
- Check all the environment variables, even if we currently have a default value hardcoded (maybe a more strict solution).
- Check the resulting object generated by `configuration.ts`, by creating a new class that inherits from `OnApplicationBootstrap` (mentioned in [the original issue](https://github.com/safe-global/safe-client-gateway-nest/issues/299))

I took the first option based on its simplicity and in order to stay aligned with the practices the Nest.js documentation contains, but I'm open to change if we find the third option better for some reason.

src/config/configuration.validator.ts
```diff
@@ -0,0 +1,23 @@
...
+    EXCHANGE_API_KEY: { type: 'string' },
+  },
+  required: ['AUTH_TOKEN', 'EXCHANGE_API_KEY'],
+};
+
+export function validate(
+  configuration: Record<string, unknown>,
+): Record<string, unknown> {
+  if (process.env.NODE_ENV !== 'test') {
+    const ajv = new Ajv({ allErrors: true });
```
hectorgomezv, line 16: Here we need to export a function, so we can't use DI to inject the global AJV instances and the common validation utilities we have in place. So I opted to stay simple a just create a new AJV instance. This function will be executed only once during the application bootstrap, and the instance will be immediately garbage-collected, so I tried to keep it simple in this case.

src/config/configuration.validator.ts
```diff
@@ -0,0 +1,23 @@
...
+    AUTH_TOKEN: { type: 'string' },
+    EXCHANGE_API_KEY: { type: 'string' },
+  },
+  required: ['AUTH_TOKEN', 'EXCHANGE_API_KEY'],
+};
+
+export function validate(
+  configuration: Record<string, unknown>,
+): Record<string, unknown> {
+  if (process.env.NODE_ENV !== 'test') {
```
hectorgomezv, line 15: This check prevents this environment variables validation to run while executing the tests. Having the full environment defined shouldn't be mandatory to run the application tests IMO.

src/config/configuration.validator.ts
```diff
@@ -0,0 +1,23 @@
...
+    const ajv = new Ajv({ allErrors: true });
+    if (!ajv.validate(configurationSchema, configuration)) {
+      const errors = ajv.errors?.reduce((acc, e) => [...acc, e.message], []);
```
fmrsabino, line 18: This is using the reduce function but the accumulator is an array that is being built with the spread operator. Why not have the initial value as an empty array and add the error message to it?

src/config/configuration.validator.ts
fmrsabino: I would say that not having this check is also ok. When we start a test application it should be almost identical to the real one (minus some testing modules). But for now this is ok!

src/config/configuration.validator.spec.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  it('should bypass this validation on tests', () => {
+    process.env.NODE_ENV = 'test';
+    const expected = JSON.parse(faker.datatype.json());
+    const validated = validate(expected);
+    expect(validated).toBe(expected);
+  });
+
+  it('should detect a malformed configuration in production environment', () => {
+    process.env.NODE_ENV = 'production';
```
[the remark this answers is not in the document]
hectorgomezv: We are running tests in serial right now, to avoid cache collisions on the execution of the tests (this probably deserves a review also), so this test class takes care of restoring the original value of `process.env.NODE_ENV` (which should be `test`). But even if we switch back to running tests in parallel in the future, modifying the value of the keys in  `process.env` shouldn't be a cross-test problem since Jest provides a sandboxed environment for each test suite. 

(From [Jest docs](https://jestjs.io/docs/configuration#testenvironment-string))
```testEnvironment is sandboxed. Each test suite will trigger setup/teardown in their own TestEnvironment.```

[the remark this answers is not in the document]
hectorgomezv: In fact, I think we could even remove the restoration for the same reason 🙂 I'll check that.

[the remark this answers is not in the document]
hectorgomezv: I've removed the `process.env` restoration, Jest provides the encapsulated environment for each test. Let me know if you have any concern regarding this.

src/config/configuration.validator.ts
hectorgomezv: > Why not have the initial value as an empty array and add the error message to it?

Yes, that was the purpose of the reduce function, but in this case I think the code can be simplified by just using a map function. I've changed it in [56eed68](https://github.com/safe-global/safe-client-gateway-nest/pull/307/commits/56eed689b2b7473d50ecf3b5f2dec8db6e3d0463), I hope now the purpose is clearer.

fmrsabino (approved):
🚀

PR #310, opened 2023-05-05
Use JSON for logging

hectorgomezv:
This PR is heavily inspired by https://github.com/safe-global/safe-gelato-relay-service/pull/72

- Uses JSON as the default logging format – `winston.format.json()` is now the main formatter
- The `ILoggingService` interface changed to accommodate the new format – a `message` can now be an unknown payload format or a `string` – regardless of the type, this value will be logged under `{ message: <value> }`
- All the logging-related code is found under `src/logging` which makes it independent from the other service layers. The idea is to maintain the logging utilities as a shared service that can be injected on demand in other parts of the service.
- The logging mechanism also adds `nestjs-cls` as a dependency to store request-wide data (like request ids).
- Some redundant error messages were removed (`http-error.factory.ts` and `transaction-preview.mapper.ts` for example) since they would be logged twice with this new setup.

The base format of the logs is as follows:

```javascript
{
  message: <string | object>, // string value or json payload
  request_id: <string>, // id tied to the request
  timestamp: <string>, // ISO 8601 Date string
}
```

The `message` value when logging route requests is as follows:

```javascript
{
  ...
  message: {
      client_ip: <optional string>, // retrieved from the header X-Real-IP. null if not available
      method: <string>, // GET, POST, etc..
      response_time_ms: <number>, // response time in milliseconds
      route: <optional string>, // route which handled the request. Null if no route handled this request (eg.: 404 requests)
      path: <string>, // path of the resource requested
      status_code: <number>, // resulting status code of the request
      detail: <optional string>, // optional message regarding the result of the request
  }
}
```


src/datasources/transaction-api/transaction-api.manager.ts
```diff
@@ -25,13 +24,9 @@ export class TransactionApiManager implements ITransactionApiManager {
   ) {}
 
   async getTransactionApi(chainId: string): Promise<TransactionApi> {
```
hectorgomezv, line 26: I took the opportunity to remove some logging here, I think let me know if you think it was valuable enough or if you want to tackle this in a separate PR 🙂

src/logging/logging.module.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const LoggerTransports = Symbol('LoggerTransports');
+
+/**
+ * Factory which provides a collection of transports to be used by the
+ * logger instance
+ */
+function winstonTransportsFactory(): Transport[] | Transport {
+  return new winston.transports.Console({
+    level: 'debug',
```
hectorgomezv, line 24: I took this from the Gelato Relay Service, but I think this is a clear candidate for externalization. We could find it useful to have different logging levels between environments, so I think this could be tackled in a separate issue/PR if you are OK with it.

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -47,15 +51,19 @@ export class CacheFirstDataSource {
   ): Promise<T> {
     const cached = await this.cacheService.get(cacheDir);
     if (cached != null) {
-      winston.debug(`[Cache] Cache hit: [${cacheDir.key}, ${cacheDir.field}]`);
+      this.loggingService.debug(
+        `[Cache] Cache hit: [${cacheDir.key}, ${cacheDir.field}]`,
```
fmrsabino, line 55: Something to consider: we can restructure log too. Maybe having something like:

```javascript
{
  "type": "cache_hit",
  "key": <key>,
  "field": <field>,
}
```

(no need to tackle this in this PR)

src/datasources/cache/redis.cache.service.ts
```diff
@@ -58,22 +62,22 @@ export class RedisCacheService implements ICacheService, OnModuleDestroy {
...
 
   /**
    * Forces the closing of the Redis connection associated with this service.
    */
   private async forceQuit() {
-    winston.verbose('Forcing Redis connection close');
+    this.loggingService.warn('Forcing Redis connection close');
     await this.client.disconnect();
-    winston.verbose('Redis connection closed');
+    this.loggingService.warn('Redis connection closed');
```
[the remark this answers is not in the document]
fmrsabino: I think that having `info` for the normal lifecycle and `warn` for the fallback one is appropriate. So all good from my side.

src/datasources/transaction-api/transaction-api.manager.ts
fmrsabino: I think it's ok to remove them. This was mostly done in the initial stages of the project and right now I think we have a setup to understand the flow and what might have gone wrong if something happens!

src/logging/logging.module.ts
fmrsabino: Yes that's a good idea. We can set it in the application configuration (we can keep the default as `debug` if none is set).

src/logging/logging.module.ts
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * Provides the RequestScopedLoggingService which logs the current time and request ID with every message.
+ */
+@Global()
+@Module({
+  providers: [
+    { provide: LoggingService, useClass: RequestScopedLoggingService },
+    { provide: LoggerTransports, useFactory: winstonTransportsFactory },
+    {
+      // Using a symbol or a string as a provider key doesn't work when using factories
```
fmrsabino, line 40: This comment is not 100% correct but I have more context since I wrote it 😅 

`string` works as a value if provided directly as you see it in the `provide:`. However if you extract the `string` to a `const` or `let` it stops working. I didn't have time to debug this (maybe the `const` needs to be exported due to how the modules are linked and built).

src/logging/logging.service.spec.ts
```diff
@@ -0,0 +1,91 @@
+import { faker } from '@faker-js/faker';
+import { ClsService } from 'nestjs-cls';
+import * as winston from 'winston';
+import { RequestScopedLoggingService } from './logging.service';
+
+const mockClsService = jest.mocked({
```
fmrsabino, line 6: Why do we use `jest.mocked` here but not on `mockLogger`? (or the other way around)

src/validation/providers/validation-error-factory.ts
```diff
@@ -25,7 +32,7 @@ export class ValidationErrorFactory {
       message,
     }));
 
-    winston.error({ ...errPayload, detail });
+    this.loggingService.info({ ...errPayload, detail });
```
fmrsabino, line 35: We should probably keep this as an `error` for now. If somehow the contract with the other services breaks this means that we won't be able to serve the specified resource (either it's an issue on our side or an issue on the services that we consume but should be tackled as soon as possible).

This is a different scenario than clients sending malformed payloads. For the malformed client payloads we can debate if we should consider it an error (or assign it a different level).

fmrsabino (commented):
Nice 👏 Just some small comments!

src/logging/logging.module.ts
hectorgomezv: Perfect, I created an issue for this: https://github.com/safe-global/safe-client-gateway-nest/issues/313

src/validation/providers/validation-error-factory.ts
hectorgomezv: Agree 100%, I'll change this, it shouldn't happen in a typical scenario, which is totally different from clients sending malformed payloads as you said (in that case I'd support not logging them as `error`).

src/logging/logging.service.spec.ts
hectorgomezv: For mockLogger it's not strictly necessary because we're not mocking the implementation of its members. If we don't use `jest.mocked` for `mockClsService` we would get the error:
```
error TS2339: Property 'mockReturnValue' does not exist on type
```

when calling:
```
mockClsService.getId.mockReturnValue(requestId);
```

We could use `jest.mocked` for both mocks even if it's not strictly necessary for the sake of consistency, no strong opinion here 🙂

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Yes, good point, changed in [e71d09d](https://github.com/safe-global/safe-client-gateway-nest/pull/310/commits/e71d09d04b82581fa7504f3f5361f920c54b8dd3)

src/validation/providers/validation-error-factory.ts
hectorgomezv: Fixed in [125db74](https://github.com/safe-global/safe-client-gateway-nest/pull/310/commits/125db74a23085f0e92d517551d45c34d48ed4186)

src/logging/logging.module.ts
hectorgomezv: I managed to use `Symbol` by exporting it from a separate file (as [indicated in the docs](https://docs.nestjs.com/fundamentals/dynamic-modules#module-configuration)) and importing here and into `LoggingService`. Not sure I'm happy with having a separate file just exporting a `Symbol`, do you find this setup valuable? Otherwise we can get back to using a plain string.

Here is the change: [c73623d](https://github.com/safe-global/safe-client-gateway-nest/pull/310/commits/c73623dec1bd6d2c7798b2608897c25e75a3c3a9)

fmrsabino (approved):
🚀

PR #312, opened 2023-05-08
Add ids into domain models and schemas

hectorgomezv:
Related to https://github.com/safe-global/safe-transaction-service/pull/1365

This PR adds the new fields (`transferId`, `moduleTransactionId`) for Transfers and Module Transactions into the associated entities, validation schemas, and test instance builders.

### DIFF
--- a/src/domain/safe/entities/__tests__/erc20-transfer.builder.ts
+++ b/src/domain/safe/entities/__tests__/erc20-transfer.builder.ts
@@ -10,7 +10,8 @@ export function erc20TransferBuilder(): IBuilder<ERC20Transfer> {
     .with('to', faker.finance.ethereumAddress())
     .with('transactionHash', faker.datatype.hexadecimal())
     .with('tokenAddress', faker.finance.ethereumAddress())
-    .with('value', faker.datatype.hexadecimal());
+    .with('value', faker.datatype.hexadecimal())
+    .with('transferId', faker.datatype.string());
 }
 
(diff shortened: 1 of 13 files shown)


src/domain/safe/entities/__tests__/erc20-transfer.builder.ts
```diff
@@ -10,7 +10,8 @@ export function erc20TransferBuilder(): IBuilder<ERC20Transfer> {
     .with('to', faker.finance.ethereumAddress())
     .with('transactionHash', faker.datatype.hexadecimal())
     .with('tokenAddress', faker.finance.ethereumAddress())
-    .with('value', faker.datatype.hexadecimal());
+    .with('value', faker.datatype.hexadecimal())
+    .with('transferId', faker.datatype.string());
```
fmrsabino, line 14: nitpick: `string` seems to not be really representative of the values that `transferId` returns.

This is an non-issue since we just forward these to the clients but you can also consider the following (ids seems to be hex):

```typescript
faker.datatype.hexadecimal({ prefix: '' }) // to remove the 0x
```

fmrsabino (approved):

src/domain/safe/entities/__tests__/erc20-transfer.builder.ts
fmrsabino: Let's leave it as it is then! No need to change it. It was just to get more representative data.

PR #316, opened 2023-05-09
Use CacheFirstDataSource in ExchangeApi

### LINKED ISSUE #314: Use `CacheFirstDataSource` in `ExchangeApi`
In order to avoid fetching the exchange rates every time they are needed, we should use `CacheFirstDataSource` in `ExchangeApi`.

We can set a default value of 12h for the exchange rates as we do in other deployments. We can adjust this value as we see fit.

fmrsabino:
Closes #314

- `ExchangeApi` now uses `CacheFirstDataSource` in order to avoid calling external fiat rate exchange endpoints more often than required

### DIFF
--- a/src/datasources/exchange-api/exchange-api.module.ts
+++ b/src/datasources/exchange-api/exchange-api.module.ts
@@ -6,11 +6,11 @@ import { CacheFirstDataSourceModule } from '../cache/cache.first.data.source.mod
 
 @Global()
 @Module({
-  exports: [IExchangeApi],
   imports: [CacheFirstDataSourceModule],
   providers: [
     HttpErrorFactory,
     { provide: IExchangeApi, useClass: ExchangeApi },
   ],
+  exports: [IExchangeApi],
 })
 export class ExchangeApiModule {}
--- a/src/datasources/exchange-api/exchange-api.module.ts
+++ b/src/datasources/exchange-api/exchange-api.module.ts
@@ -2,13 +2,15 @@ import { Global, Module } from '@nestjs/common';
 import { HttpErrorFactory } from '../errors/http-error-factory';
 import { ExchangeApi } from './exchange-api.service';
 import { IExchangeApi } from '../../domain/interfaces/exchange-api.interface';
+import { CacheFirstDataSourceModule } from '../cache/cache.first.data.source.module';
 
 @Global()
 @Module({
+  exports: [IExchangeApi],
+  imports: [CacheFirstDataSourceModule],
   providers: [
     HttpErrorFactory,
     { provide: IExchangeApi, useClass: ExchangeApi },
   ],
-  exports: [IExchangeApi],
 })
 export class ExchangeApiModule {}
--- a/src/datasources/exchange-api/exchange-api.service.spec.ts
+++ b/src/datasources/exchange-api/exchange-api.service.spec.ts
@@ -1,32 +1,39 @@
 import { ExchangeApi } from './exchange-api.service';
 import { FakeConfigurationService } from '../../config/__tests__/fake.configuration.service';
-import { mockNetworkService } from '../network/__tests__/test.network.module';
 import { faker } from '@faker-js/faker';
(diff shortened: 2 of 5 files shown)


src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -37,23 +44,98 @@ describe('ExchangeApi', () => {
...
+
+    await target.getFiatCodes();
+
+    expect(mockCacheFirstDataSource.get).toBeCalledWith(
+      new CacheDir('exchange_fiat_codes', ''),
+      `${exchangeBaseUri}/symbols`,
+      { params: { access_key: exchangeApiKey } },
+      ttl, // 60 seconds
+    );
+  });
```
fmrsabino, line 91: The way this test is setup makes it less clear on why I'm creating a new instance of `ExchangeApi` here. The reason is because we set environment variables that are read in its constructor. Changing the configuration just for the test wouldn't work without creating a new instance.

I'm more in favour of removing the `beforeEach` in this case and do a setup on each test (the test `should error if configuration is not defined` was already doing something similar so I decided not to change the pattern right now).

src/datasources/exchange-api/exchange-api.service.ts
```diff
@@ -1,51 +1,60 @@
...
+import { CacheFirstDataSource } from '../cache/cache.first.data.source';
+import { CacheRouter } from '../cache/cache.router';
 
 @Injectable()
 export class ExchangeApi implements IExchangeApi {
   private readonly baseUrl: string;
   private readonly apiKey: string;
+  private readonly cacheTtlSeconds: number;
+
+  static readonly DEFAULT_CACHE_TTL_SECONDS = 60 * 60 * 12;
```
hectorgomezv, line 16: Should we set an access modifier here? I mean, this should be `private` to increase encapsulation, right? 

And related to that: why do you prefer to have this parameter's default value defined within its class and the `baseUrl` in `configuration.ts`? Should we move both to the same place?

hectorgomezv (changes requested):
🚀 Great job, just some minor comments.

src/datasources/exchange-api/exchange-api.module.ts
```diff
@@ -2,13 +2,15 @@ import { Global, Module } from '@nestjs/common';
...
+import { CacheFirstDataSourceModule } from '../cache/cache.first.data.source.module';
 
 @Global()
 @Module({
+  exports: [IExchangeApi],
+  imports: [CacheFirstDataSourceModule],
   providers: [
     HttpErrorFactory,
     { provide: IExchangeApi, useClass: ExchangeApi },
   ],
```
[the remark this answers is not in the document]
fmrsabino: It was sorted alphabetically. I will change it back as I don't have a strong opinion on this either. 👍

src/datasources/exchange-api/exchange-api.service.ts
fmrsabino: > Should we set an access modifier here? I mean, this should be `private` to increase encapsulation, right?

Yes! 👍 

> And related to that: why do you prefer to have this parameter's default value defined within its class and the `baseUrl` in `configuration.ts`? Should we move both to the same place?

Actually I started with that solution (doing it in `configuration.ts`). But doing that there would break a lot of tests because then you need to setup the Fake Configuration in every test with this values (because it's a new value missing).

I'd actually prefer to have it in the configuration itself but that means adjusting all the tests. On the other hand we should probably start thinking about having a fake configuration by default (provided).

hectorgomezv (approved):

src/datasources/exchange-api/exchange-api.service.ts
fmrsabino: Created a followup issue for this: https://github.com/safe-global/safe-client-gateway-nest/issues/318

hectorgomezv: Great, thanks @fmrsabino 🚀

PR #319, opened 2023-05-10
Add header support to NetworkRequest

fmrsabino:
- Adds header support to `NetworkRequest`. Headers are a record of a `string` to `string | number | boolean` which matches Axios interface
- Renamed the parameter `config` in `INetworkService` to `networkRequest` as it is more representative of what is expected by the function


hectorgomezv (approved):
🚀

PR #322, opened 2023-05-11
Fix Token property name in balances endpoint

### LINKED ISSUE #311: Get balances endpoint, previously type now tokenType
When fetching safe balances we are using `type` on the old backend but this has changes to `tokenType` on the revamped client gateway. We need to check if this can cause some compatibility issues.

New client gateway:

`curl --location 'https://safe-client-nest.staging.5afe.dev/v1/chains/1/safes/0xB88F61E6FbdA83fbfffAbE364112137480398018/balances/eur'`

Old client gateway

`curl --location 'https://safe-client.staging.5afe.dev/v1/chains/1/safes/0xB88F61E6FbdA83fbfffAbE364112137480398018/balances/eur'`

Also when specifying the new client gateway when setting the project locally the /assets page crashes:

Screenshot 2023-05-08 at 09 25 28

fmrsabino:
Closes #311

The property `tokenType` in `Token` (mapped in the balances route) should be named `type` instead to match the current API

### DIFF
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -97,7 +97,7 @@ describe('Balances Controller (Unit)', () => {
           items: [
             {
               tokenInfo: {
-                tokenType: 'ERC20',
+                type: 'ERC20',
                 address: expectedBalance.tokenAddress,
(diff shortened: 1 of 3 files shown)


hectorgomezv (approved):
🚀

PR #323, opened 2023-05-11
Set headers when requesting Fiat Codes

fmrsabino:
`ExchangeApi.getFiatCodes` was incorrectly using query parameters to set the API key (to match the requirements of the older API version).


src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -60,7 +60,7 @@ describe('ExchangeApi', () => {
     expect(mockCacheFirstDataSource.get).toBeCalledWith(
       new CacheDir('exchange_fiat_codes', ''),
       `${exchangeBaseUri}/symbols`,
-      { params: { access_key: exchangeApiKey } },
+      { headers: { apikey: exchangeApiKey } },
```
fmrsabino, line 63: Unfortunately even the test was testing against the older API version (so the test was wrong too)

hectorgomezv (approved):

PR #330, opened 2023-05-12
Fix native token mapping

### LINKED ISSUE #325: [{safe}/balances endpoint] Different payload for NATIVE_TOKEN
When fetching safe balances, the payload for the native currency is different between the old and new client gateway.

New client gateway:

`curl --location 'https://safe-client-nest.staging.5afe.dev/v1/chains/1/safes/0xB88F61E6FbdA83fbfffAbE364112137480398018/balances/eur'`

Old client gateway:

`curl --location 'https://safe-client.staging.5afe.dev/v1/chains/1/safes/0xB88F61E6FbdA83fbfffAbE364112137480398018/balances/eur'`

Payload between nest cg (left) legacy cg (right):

Screenshot 2023-05-11 at 17 45 13

fmrsabino:
Closes #325

- Fixes native token mapping in `GET /v1/chains/:chainId/safes/:safeAddress/balances/:fiatCode`
- The token type is now correctly assigned – it was comparing to `undefined` with a strict comparison (the value can be `null` but not `undefined`)
- Uses the Native Token metadata from the chain correctly – this is used for retrieving respective decimals, symbol, name and logo URI

### DIFF
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -63,7 +63,7 @@ describe('Balances Controller (Unit)', () => {
     await app.close();
   });
 
-  describe('Maps ERC20 token correctly', () => {
+  describe('GET /balances maps ERC20 token correctly', () => {
     it(`Success`, async () => {
       const chainId = '1';
       const safeAddress = '0x0000000000000000000000000000000000000001';
@@ -129,7 +129,7 @@ describe('Balances Controller (Unit)', () => {
       );
     });
 
-    it(`Maps native token correctly`, async () => {
+    it(`GET /balances maps native token correctly`, async () => {
       const safeAddress = faker.finance.ethereumAddress();
       const transactionApiBalancesResponse = [
         balanceBuilder().with('tokenAddress', null).build(),
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -63,8 +63,8 @@ describe('Balances Controller (Unit)', () => {
     await app.close();
   });
 
-  describe('GET /balances maps ERC20 token correctly', () => {
-    it(`Success`, async () => {
+  describe('GET /balances', () => {
+    it(`maps ERC20 token correctly`, async () => {
       const chainId = '1';
       const safeAddress = '0x0000000000000000000000000000000000000001';
       const transactionApiBalancesResponse = [balanceBuilder().build()];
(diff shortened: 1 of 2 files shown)


fmrsabino:
@safe-global/safe-client-gateway-maintainers Adding a test to validate this change.

src/routes/balances/balances.controller.spec.ts
```diff
@@ -127,6 +129,57 @@ describe('Balances Controller (Unit)', () => {
...
+      const transactionApiBalancesResponse = [
+        balanceBuilder().with('tokenAddress', null).build(),
+      ];
+      const exchangeApiResponse = exchangeRatesBuilder()
+        .with('success', true)
+        .with('rates', { USD: 2.0 })
+        .build();
+      const chain = chainBuilder().build();
+      mockNetworkService.get.mockImplementation((url) => {
+        if (url == `https://test.safe.config/api/v1/chains/${chain.chainId}`) {
```
fmrsabino, line 143: I kept the url (without using faker) because I didn't want to change the overall test setup in this PR.

src/routes/balances/balances.controller.spec.ts
```diff
@@ -61,7 +63,7 @@ describe('Balances Controller (Unit)', () => {
     await app.close();
   });
 
-  describe('GET /balances', () => {
+  describe('Maps ERC20 token correctly', () => {
```
hectorgomezv, line 66: Maybe this title should remain the same (`GET /balances`)? 🤔

hectorgomezv (changes requested):
🚀 Great job, I've left a minor doubt/comment only.

src/routes/balances/balances.controller.spec.ts
fmrsabino: > Maybe this title should remain the same (`GET /balances`)? 🤔

I wanted to draw some comparison with the new test which also tests `GET /balances` in a way. Do you have other suggestions?

hectorgomezv: Maybe I'm not getting it, but you are changing the `describe` which contains two tests. The first one could have `it('Maps ERC20 token correctly', ...)` and the second one `it('Maps native token correctly', ...)`, but the `describe` block should remain the same, right? 🙂

fmrsabino: [f9fb1a6](https://github.com/safe-global/safe-client-gateway-nest/pull/330/commits/f9fb1a68f747ee51909a330d04e89a7bf8968cda)

hectorgomezv (approved):
🚀

PR #345, opened 2023-05-19
Map DataSourceErrors in RouteLoggerInterceptor

### LINKED ISSUE #343: RouteLoggerInterceptor does not retrieve the status code from DataSourceErrors
The `RouteLoggerInterceptor` correctly intercepts the error type `DataSourceError`. However, currently this interceptor assumes that all errors will be instances of `HttpException` (from NestJS) meaning that the status code is not retrieved correctly for `DataSourceError` resulting in a status code of `500` even if it's any other error code.

We should handle `DataSourceError` error types or even consider if the NestJS type would be a better fit (to be thrown) instead of the `DataSourceError`.

fmrsabino:
Closes #343 

`RouteLoggerInterceptor` now maps `DataSourceErrors` correctly meaning that the correct status code is retrieved for those errors.

Previously, only `HttpException` were handled. Since `DataSourceErrors` are not `HttpException` then an Internal Server Error would be returned by default.

### DIFF
--- a/src/routes/common/interceptors/route-logger.interceptor.spec.ts
+++ b/src/routes/common/interceptors/route-logger.interceptor.spec.ts
@@ -9,6 +9,7 @@ import {
   INestApplication,
 } from '@nestjs/common';
 import * as request from 'supertest';
+import { DataSourceError } from '../../../domain/errors/data-source.error';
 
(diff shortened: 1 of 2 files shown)


hectorgomezv (approved):
👏🏻

PR #346, opened 2023-05-19
Log external requests on error

fmrsabino:


src/datasources/network/axios.network.service.ts
```diff
@@ -32,26 +41,29 @@ export class AxiosNetworkService implements INetworkService {
...
   async delete<T = any, R = NetworkResponse<T>>(
     url: string,
     data?: object,
   ): Promise<R> {
+    const startTimeMs: number = performance.now();
     try {
       return await this.client.delete(url, { data: data });
     } catch (error) {
-      this.handleError(error);
+      this.handleError(error, startTimeMs);
```
hectorgomezv, line 60: Nitpick: what do you think about calculating the response time at this point, instead of passing down `startTimeMs` to several private functions? I mean, maybe calculating the response time is not a responsibility for those functions, and I'd say it's also more accurate to calculate a `responseTimeMs` value just after the request is executed instead of a couple of function calls later (even if the resulting number would be pretty close).

hectorgomezv (dismissed):
Good job, I just left a comment for a possible improvement but I'm also OK with this version. 🚀

src/datasources/network/axios.network.service.ts
fmrsabino: I agree. Changed here [ecbd821](https://github.com/safe-global/safe-client-gateway-nest/pull/346/commits/ecbd82194bc9f9b28de77aed3b526aec3010bfc7)

hectorgomezv (approved):

PR #348, opened 2023-05-22
Do not return cancelled transactions as queued

### LINKED ISSUE #335: [Get /queued] Queued endpoint returns already executed txs
**Request**: Get https://safe-client-nest.staging.5afe.dev/v1/chains/137/safes/0xb412684F4F0B5d27cC4A4D287F42595aB3ae124D/transactions/queued 

**The current safe nonce** : 84

**Response:**  The tx starting from the nonce 18 is returned. It seems that all cancelled tx( replaced by another tx) are returned in the response

Steps from the web wallet side:
1. go to the https://test_new_cgw_url--walletweb.review-wallet-web.5afe.dev/transactions/queue?safe=matic:0xb412684F4F0B5d27cC4A4D287F42595aB3ae124D
Expected result : https://safe-client.staging.5afe.dev/v1/chains/137/safes/0xb412684F4F0B5d27cC4A4D287F42595aB3ae124D/transactions/queued 
The safe contain only 2 txs with the same nonce 84 in the queued list

fmrsabino:
Closes #335

- The method `SafeRepository.getTransactionQueueByNonce` was requesting all non-executed transaction sorted by nonce. The issue with returning non-executed transactions is that is also returns cancelled transactions.
- `nonceGte` should be used instead so that we can return all the transactions with nonce greater than the current Safe one.
- `getTransactionQueueByModified` was updated to consider `nonceGte`
- `getTransactionQueueByNonce` was renamed to `getTransactionQueue` – this should be considered the standard "operation" when requesting the queue

### DIFF
--- a/src/domain/safe/safe.repository.interface.ts
+++ b/src/domain/safe/safe.repository.interface.ts
@@ -48,16 +48,37 @@ export interface ISafeRepository {
     offset?: number,
   ): Promise<Page<ModuleTransaction>>;
 
-  getTransactionQueueByModified(
+  /**
+   * Returns the Safe Transaction Queue in ascending order.
+   *
+   * The ascending order is first considered for the nonce.
+   * If multiple transactions have the same nonce, then they
+   * will be sorted according to the submission date
+   *
+   * @param chainId - the chain id from where to retrieve the transactions from
+   * @param safe - the target safe from where to get the transaction from
+   * @param limit - the maximum number of transactions to be returned
+   * @param offset - the starting point from which to start the pagination
+   */
+  getTransactionQueue(
     chainId: string,
-    safeAddress: string,
+    safe: Safe,
     limit?: number,
     offset?: number,
   ): Promise<Page<MultisigTransaction>>;
 
-  getTransactionQueueByNonce(
+  /**
+   * Returns the Safe Transaction Queue sorted by modified status.
+   * (from most recently modified to last)
+   *
+   * @param chainId - the chain id from where to retrieve the transactions from
+   * @param safe - the target safe from where to get the transaction from
+   * @param limit - the maximum number of transactions to be returned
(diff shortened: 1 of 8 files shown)


src/domain/safe/safe.repository.interface.ts
```diff
@@ -48,16 +48,37 @@ export interface ISafeRepository {
...
+  /**
+   * Returns the Safe Transaction Queue sorted by modified status.
+   * (from most recently modified to last)
+   *
+   * @param chainId - the chain id from where to retrieve the transactions from
+   * @param safe - the target safe from where to get the transaction from
+   * @param limit - the maximum number of transactions to be returned
+   * @param offset - the starting point from which to start the pagination
+   */
+  getTransactionQueueByModified(
```
hectorgomezv, line 79: I agree this simplifies the repository API, but I'm also a bit concerned about the fact that changing the primitive input parameters (`safeAddress: string, nonce: number`) by a domain object (`safe: Safe`) forces the client of the repository to retrieve the `Safe` beforehand. I think it's acceptable in this context, which is limited to a concrete feature, just commenting as a side note 🙂

hectorgomezv (approved):
👏🏻

src/domain/safe/safe.repository.interface.ts
fmrsabino: For the use cases that I have seen we always have the `SafeInfo`. In my opinion by using the domain object itself makes it more robust (not only in the function signature by not using `string` but also to use other safe properties that are available if needed).

Adding the `nonce` on top of the `safeAddress` seemed a bit odd in this context because we actually want the `nonce` of the `safeAddress`.

hectorgomezv: Yes, after giving it a second thought, it makes sense 👍

PR #349, opened 2023-05-22
Handle non-JSON payloads for tx origin

### LINKED ISSUE #340: [Get /history] 500 internal server error for the 1.0.0 safe
**Request** : https://safe-client-nest.staging.5afe.dev/v1/chains/1/safes/0xcd2E72aEBe2A203b84f46DEEC948E6465dB51c75/transactions/history
**Response** : {statusCode: 500, message: "Internal server error"}
message: "Internal server error"
statusCode: 500

fmrsabino:
Closes #340 

- If a transaction had an `origin` which was not a JSON payload the service would throw an exception. The expected behaviour however, is to return `null` for non-JSON payloads

### DIFF
--- a/src/routes/transactions/mappers/common/safe-app-info.mapper.ts
+++ b/src/routes/transactions/mappers/common/safe-app-info.mapper.ts
@@ -50,12 +50,8 @@ export class SafeAppInfoMapper {
   }
 
   private getOriginUrl(transaction: MultisigTransaction): string | null {
-    if (!transaction.origin) {
-      return null;
-    }
-
     try {
-      return JSON.parse(transaction.origin).url ?? null;
+      return transaction.origin ? JSON.parse(transaction.origin).url : null;
     } catch (e) {
       this.loggingService.debug(
         `Safe TX Hash ${transaction.safeTxHash} origin is not valid JSON. origin=${transaction.origin}`,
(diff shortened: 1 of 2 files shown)


src/routes/transactions/mappers/common/safe-app-info.mapper.ts
```diff
@@ -50,8 +50,17 @@ export class SafeAppInfoMapper {
...
+    }
+
+    try {
+      return JSON.parse(transaction.origin).url ?? null;
+    } catch (e) {
+      this.loggingService.debug(
+        `Safe TX Hash ${transaction.safeTxHash} origin is not valid JSON. origin=${transaction.origin}`,
+      );
+      return null;
+    }
```
hectorgomezv, line 64: If I'm not wrong you still could use the ternary operator:
```
try {
  return transaction.origin ? JSON.parse(transaction.origin).url : null;
} catch (e) {
  this.loggingService.debug(
    `Safe TX Hash ${transaction.safeTxHash} origin is not valid JSON. origin=${transaction.origin}`,
  );
  return null;
}
```
But I'm OK with the option you find more readable 👍

hectorgomezv (dismissed):

src/routes/transactions/mappers/common/safe-app-info.mapper.ts
fmrsabino: Pushed with the ternary operator [5569d8c](https://github.com/safe-global/safe-client-gateway-nest/pull/349/commits/5569d8cc0ce13374c5bd4af4c356401d9f22e379)

hectorgomezv (approved):

PR #350, opened 2023-05-22
Make Multisig Transaction submissionDate non-nullable

hectorgomezv:
This PR aims to make `submissionDate` non-nullable for multisig transactions. 

- This is a mandatory field in the Transaction Service code, since it's derived from built-in `created` field: https://github.com/safe-global/safe-transaction-service/blob/d00ad8bd891e6ab991b0c32493dadda1ab0b0c60/safe_transaction_service/history/serializers.py#L516 
- Reference from the Rust Client Gateway, where the property is not-optional: https://github.com/safe-global/safe-client-gateway/blob/e0d69faabd0a05d858eec81f235acc0945ac0993/src/common/models/backend/transactions.rs#L42


src/domain/safe/entities/multisig-transaction.entity.ts
```diff
@@ -14,7 +14,7 @@ export type MultisigTransaction = {
   blockNumber: number | null;
   confirmations: Confirmation[] | null;
   confirmationsRequired: number;
-  data: string | null; // TODO will be added under https://github.com/5afe/safe-client-gateway-nest/pull/132
```
fmrsabino, line 17: Wasn't this comment about replacing `string | null` with the new entity `DataDecoded`?

fmrsabino (approved):

src/domain/safe/entities/multisig-transaction.entity.ts
hectorgomezv: You're right! I changed the TODO for a more explicit message in [b20649a](https://github.com/safe-global/safe-client-gateway-nest/pull/350/commits/b20649a386e71b7bddaf347cadd861d8a85435f7)

fmrsabino (approved):
🚀

PR #357, opened 2023-05-24
fix: centralise `exchange.cacheTtlSeconds` fallback

### LINKED ISSUE #318: Set the default TTL for exchange rates in `configuration.ts`
Currently the `exchange.cacheTtlSeconds` does not have a fallback set in `configuration.ts`. The fallback value is set in the `ExchangeApi` itself.

This makes it harder to evaluate the fallback values just by looking at `configuration.ts`. 

Another issue is that if another component reads `exchange.cacheTtlSeconds`, that same component won't have the fallback value set.

iamacook:
Resolves #318

This centralises the `exchange.cacheTtlSeconds` fallback value by moving it from the `ExchangeApi` to the `configuration.ts`.


iamacook:
Closing in favour of an upcoming refactor from @fmrsabino.

PR #359, opened 2023-05-24
fix: forward balances params

### LINKED ISSUE #355: [{{safe}}/balances endpoint] trusted key
The trusted key is currently used on the web interface to filter and only show the tokens that are marked as trusted/default on the tx-service via this selector:

Screenshot 2023-05-23 at 12 16 27

**Steps to reproduce:**

* On tx-service django console, mark a token as trusted

* On web UI go to /balances page (for a token that was trusted and the safe owns) and filter by default tokens.

**Actual result on nest cg:**

Trusted key is "ignored" and token marked as trusted is not fetched, returning all the tokens for any given safe.

`curl --location 'https://safe-client-nest.staging.5afe.dev/v1/chains/5/safes/0x4127839cdf4F73d9fC9a2C2861d8d1799e9DF40C/balances/EUR?trusted=true' \
--header 'Accept: application/json'`

**Expected result on nest cg:**

Only show trusted tokens like in og cg

`curl --location 'https://safe-client.staging.5afe.dev/v1/chains/5/safes/0x4127839cdf4F73d9fC9a2C2861d8d1799e9DF40C/balances/EUR?trusted=true' \
--header 'Accept: application/json'`

iamacook:
Resolves #355

The `trusted` and `exclude_spam` parameters are now forwarded to the Transaction Service when requesting balances.

Example request to compare against:
`curl --location 'https://safe-client.staging.5afe.dev/v1/chains/5/safes/0x4127839cdf4F73d9fC9a2C2861d8d1799e9DF40C/balances/EUR?trusted=true`

### DIFF
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -133,11 +133,8 @@ describe('Balances Controller (Unit)', () => {
       const chainId = '1';
       const safeAddress = '0x0000000000000000000000000000000000000001';
       const transactionApiBalancesResponse = [balanceBuilder().build()];
-      const exchangeApiResponse = exchangeRatesBuilder()
-        .with('success', true)
-        .with('rates', { USD: 2.0 })
-        .build();
-      const chainResponse = chainBuilder().with('chainId', chainId).build();
+      const exchangeApiResponse = exchangeRatesBuilder().build();
+      const chainResponse = chainBuilder().build();
       const excludeSpam = true;
       const trusted = true;
 
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -129,6 +129,50 @@ describe('Balances Controller (Unit)', () => {
       );
     });
 
+    it(`excludeSpam and trusted params are forwarded to tx service`, async () => {
+      const chainId = '1';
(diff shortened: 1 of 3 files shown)


src/routes/balances/balances.controller.spec.ts
```diff
@@ -129,6 +129,50 @@ describe('Balances Controller (Unit)', () => {
...
 
+    it(`excludeSpam and trusted params are forwarded to tx service`, async () => {
+      const chainId = '1';
+      const safeAddress = '0x0000000000000000000000000000000000000001';
+      const transactionApiBalancesResponse = [balanceBuilder().build()];
+      const exchangeApiResponse = exchangeRatesBuilder()
+        .with('success', true)
+        .with('rates', { USD: 2.0 })
+        .build();
+      const chainResponse = chainBuilder().with('chainId', chainId).build();
```
hectorgomezv, line 140: Nitpick: I think this additional builders setup is not necessary for this test, we can rely on _random_ entities and just use the builders output:
```javascript
const exchangeApiResponse = exchangeRatesBuilder().build();
const chainResponse = chainBuilder().build();
```

iamacook: I agree. I've adjusted the two entities accordingly.

hectorgomezv (approved):
Great job! 👏🏻

PR #363, opened 2023-05-24
Fix returned messages order

### LINKED ISSUE #353: [Get /messages] Wrong messages order in the response
request: Get https://safe-client-nest.staging.5afe.dev/v1/chains/5/safes/0xC642bEc0e4f69815e89723C09caA8DAa524cAc9c/messages

Current result: The latest messages are at the bottom, as a result we have the wrong sorting on the UI

Old request: https://safe-client.staging.5afe.dev/v1/chains/5/safes/0xC642bEc0e4f69815e89723C09caA8DAa524cAc9c/messages - The latest messages were at the top

fmrsabino:
Closes #353 

The messages returned by `GET chains/:chainId/safes/:safeAddress/messages` were being returned in ascending order (regarding their creation date). Descending order should be used instead.

Additionally, the results were being ~mapped~ pushed to an external array inside a `Promise.all` therefore breaking the deterministic sorting rules required.

### DIFF
--- a/src/routes/messages/messages.service.ts
+++ b/src/routes/messages/messages.service.ts
@@ -56,22 +56,23 @@ export class MessagesService {
     const previousURL = cursorUrlFromLimitAndOffset(routeUrl, page.previous);
     const groups = this.getOrderedGroups(page.results);
 
-    const results: (DateLabel | MessageItem)[] = [];
-    for (const [timestamp, messages] of groups) {
-      const messageItems = await this.messageMapper.mapMessageItems(
-        chainId,
-        messages,
-        safe,
-      );
-      results.push(new DateLabel(timestamp));
-      results.push(...messageItems);
-    }
+    const labelledGroups = await Promise.all(
+      groups.map(async ([timestamp, messages]) => {
+        const messageItems = await this.messageMapper.mapMessageItems(
+          chainId,
+          messages,
+          safe,
+        );
+
+        return [new DateLabel(timestamp), ...messageItems];
+      }),
+    );
 
     return <Page<DateLabel | MessageItem>>{
       count: page.count,
       next: nextURL?.toString() ?? null,
       previous: previousURL?.toString() ?? null,
-      results,
+      results: labelledGroups.flat(),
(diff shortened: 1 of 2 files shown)


src/routes/messages/messages.service.ts
```diff
@@ -99,8 +97,24 @@ export class MessagesService {
...
+            Number(groupKey),
+            orderBy(
+              groups[groupKey],
+              (message) => {
+                message.created.getTime();
+              },
+              ['desc'],
+            ),
+          ];
+        })
```
iamacook, line 117: `orderBy` isn't changing anything here (nothing is being returned in [the iteratee](https://lodash.com/docs/4.17.15#orderBy)). It can be removed without breaking the tests.

```suggestion
        .map((groupKey) => {
          return [
            Number(groupKey),
            groups[groupKey]
          ];
        })
```

src/routes/messages/messages.service.ts
```diff
@@ -57,17 +57,15 @@ export class MessagesService {
...
-    );
+    for (const [timestamp, messages] of groups) {
+      const messageItems = await this.messageMapper.mapMessageItems(
+        chainId,
+        messages,
+        safe,
+      );
+      results.push(new DateLabel(timestamp));
+      results.push(...messageItems);
+    }
```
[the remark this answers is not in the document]
fmrsabino: Yes I do believe so! At least we leverage the Promise API better in a way! 👍 Will change it.

src/routes/messages/messages.service.ts
fmrsabino: You are right. It was not returning from the function and it only worked because the `messages` were in the correct order.

I've added the return and adjusted the tests to include different hours (for the messages on the same day) [0a7753d](https://github.com/safe-global/safe-client-gateway-nest/pull/363/commits/0a7753d8541a2a45eb8d72fa29fb92af0e0efb93)

iamacook (dismissed):

iamacook (approved):

PR #369, opened 2023-05-25
fix: return `null` if guard is empty address

### LINKED ISSUE #347: [Get /safes/{safeAddress}] No message Tag and 0x000 guard is processed incorrectly

Issues:

iamacook:
Resolves #347

`SafeState['guard']` now returns `null` if the Safe's guard is an empty address (`0x0000000000000000000000000000000000000000`). It was previously defaulting to an `AddressInfo`. Test coverage for this has been added.

Note: a chain of promises were also combined into one `Promise.all` to improve performance.

### DIFF
--- a/src/routes/common/address-info/address-info.helper.ts
+++ b/src/routes/common/address-info/address-info.helper.ts
@@ -4,6 +4,7 @@ import { IContractsRepository } from '../../../domain/contracts/contracts.reposi
 import { TokenRepository } from '../../../domain/tokens/token.repository';
 import { ITokenRepository } from '../../../domain/tokens/token.repository.interface';
 import { AddressInfo } from '../entities/address-info.entity';
+import { NULL_ADDRESS } from '../constants';
 
 export type Source = 'CONTRACT' | 'TOKEN';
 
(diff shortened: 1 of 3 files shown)


src/routes/common/address-info/address-info.helper.ts
```diff
@@ -35,8 +36,7 @@ export class AddressInfoHelper {
     address: string,
     source: Source = 'CONTRACT',
   ): Promise<AddressInfo | null> {
-    if (address == '0x0000000000000000000000000000000000000000')
-      return Promise.resolve(null);
+    if (address == NULL_ADDRESS) return Promise.resolve(null);
```
fmrsabino, line 39: After reviewing the usages of this fallback I think it is not worth it to have at this level (i.e. the `AddressInfoHelper` level).

The usages on the Safe Client Gateway of this fallback are only in Safe Info related logic: https://github.com/safe-global/safe-client-gateway/blob/eb481938a49c29b979fe98a5577f719f42a2b2eb/src/routes/safes/converters.rs#L49-L54

Therefore I would suggest removing this fallback. This can potentially be an issue to other endpoints that we have which should return the `NULL_ADDRESS` instead of `null`.

What do you think @hectorgomezv ?

fmrsabino: Pushed the suggested change here: https://github.com/safe-global/safe-client-gateway-nest/pull/380

This would then affect the changes on this PR which is more specific to Safe Info data.

hectorgomezv: Yeah, I think it could be confusing, especially for a caller sending `NULL_ADDRESS` as you said, and it's not providing any benefit ATM. I'd vote for removing the fallback 👍

iamacook: Thanks for the thorough reviews and associated adjustments.

As the aforementioned change affects this PR, I will close this and we can open a separate one instead.

PR #375, opened 2023-05-26
Add fallback when a token cannot be retrieved

fmrsabino:
Depends on https://github.com/safe-global/safe-client-gateway-nest/pull/378

Adds a fallback to `MultisigTransactionInfoMapper` and `TransferInfoMapper` if the information regarding a specific token cannot be retrieved.

Previously if a token was not retrieved successfully (e.g. 404) – this error would propagate to the route and be returned to the clients.

### DIFF
--- a/src/routes/transactions/mappers/common/transaction-info.mapper.ts
+++ b/src/routes/transactions/mappers/common/transaction-info.mapper.ts
@@ -103,35 +103,40 @@ export class MultisigTransactionInfoMapper {
       );
     }
 
-    if (this.isValidTokenTransfer(transaction)) {
-      const token = await this.tokenRepository.getToken(
+    if (!this.isValidTokenTransfer(transaction)) {
+      return this.customTransactionMapper.mapCustomTransaction(
+        transaction,
+        value,
+        dataSize,
         chainId,
-        transaction.to,
       );
+    }
 
-      if (token.type === TokenType.Erc20) {
+    const token = await this.tokenRepository
+      .getToken(chainId, transaction.to)
+      .catch(() => null);
+
+    switch (token?.type) {
+      case TokenType.Erc20:
         return this.erc20TransferMapper.mapErc20Transfer(
           token,
           chainId,
           transaction,
         );
-      }
-
-      if (token.type === TokenType.Erc721) {
+      case TokenType.Erc721:
         return this.erc721TransferMapper.mapErc721Transfer(
           token,
           chainId,
           transaction,
         );
-      }
+      default:
+        return this.customTransactionMapper.mapCustomTransaction(
+          transaction,
+          value,
+          dataSize,
+          chainId,
+        );
     }
-
-    return this.customTransactionMapper.mapCustomTransaction(
-      transaction,
-      value,
-      dataSize,
-      chainId,
-    );
   }
 
   private isCustomTransaction(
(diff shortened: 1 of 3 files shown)


src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -103,35 +103,40 @@ export class MultisigTransactionInfoMapper {
...
           transaction,
         );
-      }
+      default:
+        return this.customTransactionMapper.mapCustomTransaction(
+          transaction,
+          value,
+          dataSize,
+          chainId,
+        );
```
hectorgomezv, line 138: Nitpick: IMO instead of inverting the `isValidTokenTransfer(...)` condition, and redirecting to `mapCustomTransaction` in two places, it could be simpler to write it this way (and the flow would be the same):

```javascript
if (this.isValidTokenTransfer(transaction)) {
  const token = await this.tokenRepository
    .getToken(chainId, transaction.to)
    .catch(() => null);

  switch (token?.type) {
    case TokenType.Erc20:
      return this.erc20TransferMapper.mapErc20Transfer(
        token,
        chainId,
        transaction,
      );
    case TokenType.Erc721:
      return this.erc721TransferMapper.mapErc721Transfer(
        token,
        chainId,
        transaction,
      );
  }
}

return this.customTransactionMapper.mapCustomTransaction(
  transaction,
  value,
  dataSize,
  chainId,
);
```

What do you think, do you find it more readable?

src/routes/transactions/mappers/transfers/transfer-info.mapper.ts
```diff
@@ -52,24 +52,29 @@ export class TransferInfoMapper {
   ): Promise<Transfer> {
     if (isERC20Transfer(domainTransfer)) {
       const { tokenAddress, value } = domainTransfer;
-      const token = await this.getToken(chainId, tokenAddress);
+      const token: Token | null = await this.getToken(
+        chainId,
+        tokenAddress,
+      ).catch(() => null);
```
hectorgomezv, line 58: Nitpick: Just FYI, I think there are two main reasons why we couldn't get the token data: validation error, and datasource retrieval error. In the first case, we are logging the error at `ValidationErrorFactory`, but in the second case we wouldn't notice and this would just return `null` silently if I'm not wrong.

I'm OK with this behavior, but I'm interested in whether you think we could also log errors on `HttpErrorFactory` as well, or log them selectively depending on the use case (in this new `catch` block for example).

fmrsabino: It wouldn't be completely silent. We are logging network responses [here](https://github.com/safe-global/safe-client-gateway-nest/blob/a65d5d9d0dcd695a5d6d200b9cf2db34872c10cf/src/datasources/network/axios.network.service.ts#L78C1-L88).

However, it is true that the logging happens on the network part instead of the Datasource abstraction.

hectorgomezv: Oh, yes, it's fine then. Of course we can think if that's the best spot for the logging but if it's not silent I'm OK with this, we can improve further 👍

src/routes/transactions/mappers/common/transaction-info.mapper.ts
fmrsabino: I do! Pushed the suggestion [7decefc](https://github.com/safe-global/safe-client-gateway-nest/pull/375/commits/7decefc0b871d524f1bf79f177fa23aa56180fff)

hectorgomezv (approved):
🚀

PR #378, opened 2023-05-26
Remove transfer/transactions E2E tests

fmrsabino:
`get-multisig-transaction.e2e-spec.ts`, `get-module-transaction.e2e-spec.ts` and `get-incoming-transfers.e2e-spec.ts` rely on Safe data that can change and break these tests.

Therefore, these tests were removed so that they would not block any other contributions to the project.


hectorgomezv (approved):
🚀🚀🚀

PR #381, opened 2023-05-29
Fix addressInfoIndex

hectorgomezv:
Closes https://github.com/safe-global/safe-client-gateway-nest/issues/362 and https://github.com/safe-global/safe-client-gateway-nest/issues/367

This PR:
- Fixes the legacy Rust implementation differences by avoiding traversing the `dataDecoded` with an arbitrary depth size. This fix makes the service only get results from the first level (as the Rust implementation does).
- For the items contained in `addressInfoIndex`, the associated `AddressInfo` is retrieved from `Token`, and from `Contract` as a fallback. Then, `AddressInfo` with no `name` property are filtered out. (to meet the legacy behavior)
- Unifies the `txData.addressInfoIndex` response as a non-empty object, or `null`.

### DIFF
--- a/src/routes/transactions/mappers/common/transaction-data.mapper.ts
+++ b/src/routes/transactions/mappers/common/transaction-data.mapper.ts
@@ -1,5 +1,5 @@
 import { Inject, Injectable } from '@nestjs/common';
-import { isArray } from 'lodash';
+import { isArray, isEmpty } from 'lodash';
 import { ContractsRepository } from '../../../../domain/contracts/contracts.repository';
 import { IContractsRepository } from '../../../../domain/contracts/contracts.repository.interface';
 import {
@@ -11,7 +11,6 @@ import { NULL_ADDRESS } from '../../../common/constants';
 import { AddressInfo } from '../../../common/entities/address-info.entity';
 import { isHex } from '../../../common/utils/utils';
 import { Contract } from '../../../contracts/entities/contract.entity';
-import { DataDecodedParameter } from '../../../data-decode/entities/data-decoded-parameter.entity';
 import { DataDecoded } from '../../../data-decode/entities/data-decoded.entity';
 import {
   ADDRESS_PARAMETER_TYPE,
@@ -58,7 +57,7 @@ export class TransactionDataMapper {
       previewTransactionDto.value,
       previewTransactionDto.operation,
       isTrustedDelegateCall ?? null,
-      addressInfoIndex,
+      isEmpty(addressInfoIndex) ? null : addressInfoIndex,
     );
   }
 
@@ -99,7 +98,7 @@ export class TransactionDataMapper {
    * Builds a {@link Record<string, AddressInfo>} which contains all the addresses
    * extracted from {@link DataDecoded} as keys, and their related {@link AddressInfo}
    * as value.
-   * @param chainId chain id
+   * @param chainId - chain id to use
    * @param dataDecoded data decoded to use
    * @returns {@link Record<string, AddressInfo>}
    */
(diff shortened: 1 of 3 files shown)


src/routes/transactions/mappers/common/transaction-data.mapper.ts
```diff
@@ -108,61 +107,83 @@ export class TransactionDataMapper {
...
+          promises.push(this._getFromValueDecoded(chainId, valueDecoded));
+        }
+      }
+    } else {
+      for (const parameter of parameters) {
+        if (parameter.type === ADDRESS_PARAMETER_TYPE) {
+          promises.push(this._getIfValid(chainId, parameter.value));
+        }
+      }
+    }
```
fmrsabino, line 126: nitpick: I think it's more concise to have the above as follows:

```typescript
for (const parameter of parameters) {
  if (
    method === MULTI_SEND_METHOD_NAME &&
    parameter.name === TRANSACTIONS_PARAMETER_NAME &&
    parameter.valueDecoded
  ) {
    promises.push(
      this._getFromValueDecoded(chainId, parameter.valueDecoded),
    );
  } else if (parameter.type === ADDRESS_PARAMETER_TYPE) {
    promises.push(this._getIfValid(chainId, parameter.value));
  }
}
```

src/routes/transactions/mappers/common/transaction-data.mapper.ts
```diff
@@ -108,61 +107,83 @@ export class TransactionDataMapper {
...
-  ): DataDecodedParameter[] {
+  ): Promise<(AddressInfo | null)[]> {
     if (!isArray(valueDecoded)) return [];
-    return valueDecoded.flatMap((operation) =>
-      this.getAddressParametersFromDataDecoded(operation.dataDecoded),
-    );
+    const promises: Promise<AddressInfo | null>[] = [];
+
+    for (const operation of valueDecoded) {
+      if (operation.to) {
```
fmrsabino, line 152: At this point we should know that the transaction should have a `to`. This validation is done on the Rust side with `serde`. It should fail if we don't have a `to` but here we are silently dealing with transactions that do not have a `to`.

src/routes/transactions/mappers/common/transaction-data.mapper.ts
```diff
@@ -108,61 +107,83 @@ export class TransactionDataMapper {
...
     valueDecoded: unknown,
-  ): DataDecodedParameter[] {
+  ): Promise<(AddressInfo | null)[]> {
     if (!isArray(valueDecoded)) return [];
-    return valueDecoded.flatMap((operation) =>
-      this.getAddressParametersFromDataDecoded(operation.dataDecoded),
-    );
+    const promises: Promise<AddressInfo | null>[] = [];
+
+    for (const operation of valueDecoded) {
```
fmrsabino, line 151: nitpick: use `transaction` instead of `operation`. At this stage this should be considered a Multisend.

src/routes/transactions/mappers/common/transaction-data.mapper.ts
hectorgomezv: I agree, changed in [d154939](https://github.com/safe-global/safe-client-gateway-nest/pull/381/commits/d1549397cc4351b05e528e841f2e5a849bc90835)

src/routes/transactions/mappers/common/transaction-data.mapper.ts
hectorgomezv: Changed in [d154939](https://github.com/safe-global/safe-client-gateway-nest/pull/381/commits/d1549397cc4351b05e528e841f2e5a849bc90835)

src/routes/transactions/mappers/common/transaction-data.mapper.ts
hectorgomezv: Great, I agree it's simpler that way. Thanks, I've committed your suggestion 🚀

src/routes/transactions/mappers/common/transaction-data.mapper.ts
```diff
@@ -108,61 +106,75 @@ export class TransactionDataMapper {
...
-      isHex(value) &&
-      value !== NULL_ADDRESS
-      ? this.addressInfoHelper.getOrDefault(chainId, value)
-      : null;
+    if (typeof value === 'string' && value !== NULL_ADDRESS) {
+      const addressInfo = await this.addressInfoHelper
+        .get(chainId, value, 'TOKEN')
+        .catch(() => this.addressInfoHelper.get(chainId, value, 'CONTRACT'))
+        .catch(() => null);
+      return addressInfo?.name ? addressInfo : null;
```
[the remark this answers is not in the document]
fmrsabino: Given the difference that you mentioned let's keep as you have it.

However we should not change the `AddressInfoHelper` as it's generic enough and used by most features. Adding this there would change the output of every feature and this seems to be something that is specific to this feature or?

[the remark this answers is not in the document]
hectorgomezv: I´d say it affects all callers of `address_ex_from_contracts` in Rust implementation, but other flows are using the `or_default` fallback, so I think this is something specific to this feature 👍🏻 I agree we can leave `AddressInfoHelper` as a generic service class.

PR #388, opened 2023-06-01
Parameterize service configuration

fmrsabino:
- The service configuration is now parameterised meaning that the `ConfigurationModule` now takes as an argument the configuration to load (effectively making it a `DynamicModule`)
- Adds a test configuration that relies on `faker` for most properties
- Adjusts tests so that we now read values from the configuration that was loaded instead of a global property

This change allows tests to not specify required environment variables in order to run the service as long as a `ConfigurationModule` is provided with a valid `configuration` (production or fake one).


hectorgomezv (commented):
🚀🚀🚀

hectorgomezv (approved):

PR #389, opened 2023-06-02
Move exchange.cacheTtlSeconds default to configuration

### LINKED ISSUE #318: Set the default TTL for exchange rates in `configuration.ts`

fmrsabino:
Closes #318 

Instead of providing a default in `ExchangeApi`, the default value is now provided in the `configuration` itself.

### DIFF
--- a/src/datasources/exchange-api/exchange-api.service.spec.ts
+++ b/src/datasources/exchange-api/exchange-api.service.spec.ts
@@ -15,11 +15,16 @@ describe('ExchangeApi', () => {
   let fakeConfigurationService: FakeConfigurationService;
   const exchangeBaseUri = faker.internet.url();
   const exchangeApiKey = faker.random.alphaNumeric();
+  const exchangeCacheTtlSeconds = faker.datatype.number();
 
   beforeAll(async () => {
     fakeConfigurationService = new FakeConfigurationService();
     fakeConfigurationService.set('exchange.baseUri', exchangeBaseUri);
     fakeConfigurationService.set('exchange.apiKey', exchangeApiKey);
+    fakeConfigurationService.set(
+      'exchange.cacheTtlSeconds',
+      exchangeCacheTtlSeconds,
+    );
   });
 
   beforeEach(async () => {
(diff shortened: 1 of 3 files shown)


src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -51,20 +56,6 @@ describe('ExchangeApi', () => {
...
-
-    await service.getFiatCodes();
-
-    expect(mockCacheFirstDataSource.get).toBeCalledWith(
-      new CacheDir('exchange_fiat_codes', ''),
-      `${exchangeBaseUri}/symbols`,
-      { headers: { apikey: exchangeApiKey } },
-      60 * 60 * 12, // 12h in seconds
-    );
-  });
```
fmrsabino, line 66: This test is no longer needed as it was testing the default set in the class. However, the default is not the class' responsibility anymore.

src/datasources/exchange-api/exchange-api.service.spec.ts
```diff
@@ -103,17 +94,6 @@ describe('ExchangeApi', () => {
...
-  it('exchangeRates uses default cache TTL (12h) if none is set', async () => {
-    await service.getRates();
-
-    expect(mockCacheFirstDataSource.get).toBeCalledWith(
-      new CacheDir('exchange_rates', ''),
-      `${exchangeBaseUri}/latest`,
-      { headers: { apikey: exchangeApiKey } },
-      60 * 60 * 12, // 12h in seconds
-    );
-  });
```
fmrsabino, line 115: Same as https://github.com/safe-global/safe-client-gateway-nest/pull/389#discussion_r1214163038

hectorgomezv (approved):
👏🏻

PR #419, opened 2023-06-12
fix: return `null` if guard is a "null address"

### LINKED ISSUE #347: [Get /safes/{safeAddress}] No message Tag and 0x000 guard is processed incorrectly

iamacook:
Resolves #347

`SafeState['guard']` now returns `null` if the Safe's guard is a "null address" (`0x0000000000000000000000000000000000000000`). It was previously serialising it instead. Test coverage for this has been added.

Note: promise chains were also combined to improve performance.


fmrsabino (approved):
🚀 🚀 🚀

PR #424, opened 2023-06-12
feat: add `SafeState['messagesTag']`

### LINKED ISSUE #347: [Get /safes/{safeAddress}] No message Tag and 0x000 guard is processed incorrectly

iamacook:
Partially resolves #347

This adds `SafeState['messagesTag']`, returning the most recent message `modified` timestamp. Test coverage for this has been added.

### DIFF
--- a/src/routes/safes/safes.service.ts
+++ b/src/routes/safes/safes.service.ts
@@ -171,7 +171,7 @@ export class SafesService {
     }
 
     const sortedMessages = messages.results.sort((m1, m2) => {
-      return m2.modified > m1.modified ? 1 : -1;
+      return m2.modified.getTime() - m1.modified.getTime();
     });
 
     return sortedMessages[0].modified;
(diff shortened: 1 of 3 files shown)


src/routes/safes/safes.service.ts
```diff
@@ -150,6 +157,26 @@ export class SafesService {
...
+      chainId,
+      safeAddress,
+    );
+
+    if (messages.results.length === 0) {
+      return null;
+    }
+
+    const sortedMessages = messages.results.sort((m1, m2) => {
+      return m2.modified > m1.modified ? 1 : -1;
```
fmrsabino, line 174: nitpick: we can set the `compareFn` as the difference between the dates.

```typescript
(m1, m2) => {
  return m2.modified.getTime() - m1.modified.getTime();
}
```

It should return 0 if the dates are equal [^1] (although it works for this specific scenario).

[^1]: https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/array/sort#syntax

fmrsabino (approved):

fmrsabino (approved):

src/routes/safes/safes.service.ts
iamacook: Much cleaner! I've adjusted it accordingly.

PR #491, opened 2023-06-21
Fix bad request and response in TransactionApi.postDelegate

### LINKED ISSUE #457: [Post /delegates] 500 error on mobile
{"code":500,"message":"An error occurred"}

Endpoint : /v1/chains/1/delegates/
Check the body in the slack nest js channel

hectorgomezv:
Closes #457 

This PR fixes two things related with the `TransactionApi.postDelegate` function:

- The function was returning the whole `NetworkResponse` object, which was generating a 500 status error while trying to serialize the whole object to JSON before sending the response to the client. In fact, this endpoint is expected to return no data so I just removed the return statement.
- The function was also sending `safe: undefined` in the payload, which is correct, but it seems the Transaction Service has a bug in my opinion related to this, since passing `undefined` generates a 500 status error in the Transaction Service, but passing `null` doesn't. So `safe: safeAddress ?? null` is now used as a workaround. A TODO comment is added to change this back when the Transaction Service bug gets fixed (I have opened a ticket here: https://github.com/safe-global/safe-transaction-service/issues/1521)

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.ts
+++ b/src/datasources/transaction-api/transaction-api.service.ts
@@ -251,11 +251,11 @@ export class TransactionApi implements ITransactionApi {
     delegator?: string,
     signature?: string,
     label?: string,
-  ): Promise<unknown> {
+  ): Promise<void> {
     try {
       const url = `${this.baseUrl}/api/v1/delegates/`;
-      return await this.networkService.post(url, {
-        safe: safeAddress,
+      await this.networkService.post(url, {
+        safe: safeAddress ?? null, // TODO: this is a workaround while https://github.com/safe-global/safe-transaction-service/issues/1521 is not fixed.
         delegate: delegate,
         delegator: delegator,
         signature: signature,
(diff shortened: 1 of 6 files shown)


src/domain/delegate/delegate.repository.interface.ts
```diff
@@ -21,7 +21,7 @@ export interface IDelegateRepository {
     delegator?: string,
     signature?: string,
     label?: string,
-  ): Promise<unknown>;
+  ): Promise<void>;
```
hectorgomezv, line 24: This endpoint is expected to return no data, just the 200 status code, so I changed the return type of the related functions in the interfaces, in order to make this fact explicit.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -251,11 +251,11 @@ export class TransactionApi implements ITransactionApi {
...
     signature?: string,
     label?: string,
-  ): Promise<unknown> {
+  ): Promise<void> {
     try {
       const url = `${this.baseUrl}/api/v1/delegates/`;
-      return await this.networkService.post(url, {
-        safe: safeAddress,
+      await this.networkService.post(url, {
+        safe: safeAddress ?? null, // TODO: this is a workaround while https://github.com/safe-global/safe-transaction-service/issues/1521 is not fixed.
```
hectorgomezv, line 258: This is the main fix I'm talking about in the PR comments.

iamacook (approved):

PR #497, opened 2023-06-26
Remove chain id from the hook route path

### LINKED ISSUE #495: Don't use `chainId` on the cache invalidation urls
- Currently, for invalidating cache `/v1/chains/:chainId/hooks/events` must be called
- All the webhook objects contain the `chainId`, so it's redundant
- We propose to use instead a `/v1/hooks/events` url so it's not required to configure one url per chain

fmrsabino:
Closes #495 

- Removes the `chainId` from the route path i.e. `/chains/:chainId/hooks/events` is now `/hooks/events`.
- The chain id is available on every webhook payload that we currently handle, so we can use it instead (making the route chain agnostic).


Uxio0 (approved):

Uxio0:
Looks good, thanks for the quick implementation @fmrsabino !

hectorgomezv (approved):
🚀🚀🚀

PR #514, opened 2023-06-27
Use latest git tag for service version

fmrsabino:
Depends on #513 

Spawns a sub process to read the available git tags and sets it as the service version. If git is not available or an error is returned, the service version is set to null.

### DIFF
--- a/src/routes/about/__tests__/get-about.e2e-spec.ts
+++ b/src/routes/about/__tests__/get-about.e2e-spec.ts
@@ -25,7 +25,7 @@ describe('Get about e2e test', () => {
         expect(body).toEqual(
           expect.objectContaining({
             name: expect.any(String),
-            version: expect.any(String),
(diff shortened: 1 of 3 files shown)


iamacook (approved):

src/routes/about/__tests__/get-about.e2e-spec.ts
```diff
@@ -2,6 +2,8 @@ import { INestApplication } from '@nestjs/common';
 import { Test } from '@nestjs/testing';
 import * as request from 'supertest';
 import { AppModule } from '../../../app.module';
+import { expect } from '@jest/globals';
+import '../../../__tests__/matchers/to-be-string-or-null';
```
iamacook, line 6: As per our discussion, let's try to move this to the Jest configuration so it need not be imported directly.

iamacook (approved):

PR #516, opened 2023-06-28
Add LOG_LEVEL environment variable

### LINKED ISSUE #313: Add log level environment variable
Context: https://github.com/safe-global/safe-client-gateway-nest/pull/310#discussion_r1186072133

- Add `LOG_LEVEL` environment variable, with `debug ` default value.

hectorgomezv:
Closes #313 

This PR:
- Adds the `LOG_LEVEL` environment variable to the expected configuration.
- Adds this `log.level` parameter to the configuration entity (and to the test entity). 
- Adds the `LOG_LEVEL` environment variable to the `.env.template` file.
- Injects the log level value into the `Logger` build configuration.

### DIFF
--- a/src/config/configuration.validator.spec.ts
+++ b/src/config/configuration.validator.spec.ts
@@ -18,7 +18,6 @@ describe('Configuration validator', () => {
   });
 
   it('should an invalid LOG_LEVEL configuration in production environment', () => {
-    process.env.NODE_ENV = 'production';
     expect(() =>
       validate({
         ...JSON.parse(fakeJson()),
--- a/src/config/configuration.validator.spec.ts
+++ b/src/config/configuration.validator.spec.ts
@@ -10,11 +10,23 @@ describe('Configuration validator', () => {
     expect(validated).toBe(expected);
   });
 
-  it('should detect a malformed configuration in production environment', () => {
+  it('should detect missing mandatory configuration in production environment', () => {
     process.env.NODE_ENV = 'production';
     expect(() => validate(JSON.parse(fakeJson()))).toThrow(
-      /Mandatory configuration is missing: .*AUTH_TOKEN.*EXCHANGE_API_KEY/,
+      /must have required property 'AUTH_TOKEN'.*must have required property 'EXCHANGE_API_KEY'/,
     );
+  });
+
+  it('should an invalid LOG_LEVEL configuration in production environment', () => {
+    process.env.NODE_ENV = 'production';
+    expect(() =>
+      validate({
+        ...JSON.parse(fakeJson()),
+        AUTH_TOKEN: faker.string.uuid(),
+        EXCHANGE_API_KEY: faker.string.uuid(),
+        LOG_LEVEL: faker.word.words(),
+      }),
+    ).toThrow(/LOG_LEVEL must be equal to one of the allowed values/);
   });
 
   it('should return the input configuration if validated in production environment', () => {
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -1,5 +1,4 @@
 import { faker } from '@faker-js/faker';
-import { sample } from 'lodash';
 import configuration from '../../entities/configuration';
 
 export default (): ReturnType<typeof configuration> => ({
@@ -21,9 +20,7 @@ export default (): ReturnType<typeof configuration> => ({
     default: faker.number.int(),
   },
   httpClient: { requestTimeout: faker.number.int() },
-  log: {
-    level: sample(['error', 'warn', 'info', 'debug']) ?? 'debug',
-  },
+  log: { level: 'debug' },
   redis: {
     host: faker.internet.domainName(),
     port: faker.internet.port().toString(),
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -1,4 +1,5 @@
 import { faker } from '@faker-js/faker';
+import { sample } from 'lodash';
 import configuration from '../../entities/configuration';
 
 export default (): ReturnType<typeof configuration> => ({
@@ -20,6 +21,9 @@ export default (): ReturnType<typeof configuration> => ({
     default: faker.number.int(),
   },
   httpClient: { requestTimeout: faker.number.int() },
+  log: {
+    level: sample(['error', 'warn', 'info', 'debug']) ?? 'debug',
+  },
   redis: {
     host: faker.internet.domainName(),
     port: faker.internet.port().toString(),
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -61,6 +61,9 @@ export default () => ({
       process.env.HTTP_CLIENT_REQUEST_TIMEOUT_MILLISECONDS ?? `${5_000}`,
     ),
   },
+  log: {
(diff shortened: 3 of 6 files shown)


src/config/entities/__tests__/configuration.ts
```diff
@@ -16,18 +17,21 @@ export default (): ReturnType<typeof configuration> => ({
...
     port: faker.internet.port().toString(),
   },
-  expirationTimeInSeconds: {
-    default: faker.number.int(),
+  safeConfig: {
+    baseUri: faker.internet.url({ appendSlash: false }),
+  },
+  safeTransaction: {
+    useVpcUrl: false,
   },
```
hectorgomezv, line 36: Just ordering alphabetically here (besides adding the `log` property`).

src/config/entities/configuration.ts
```diff
@@ -72,4 +61,18 @@ export default () => ({
...
+    host: process.env.REDIS_HOST || 'localhost',
+    port: process.env.REDIS_PORT || '6379',
+  },
+  safeConfig: {
+    baseUri:
+      process.env.SAFE_CONFIG_BASE_URI || 'https://safe-config.safe.global/',
+  },
+  safeTransaction: {
+    useVpcUrl: process.env.USE_TX_SERVICE_VPC_URL?.toLowerCase() === 'true',
+  },
```
hectorgomezv, line 77: Just ordering alphabetically here too (besides adding the `log` property`).

src/config/entities/__tests__/configuration.ts
fmrsabino: Can you cherry pick this to a separate PR? Rolling back the `log` should not roll back this sorting in my opinion.

hectorgomezv: Done! https://github.com/safe-global/safe-client-gateway-nest/pull/519

src/config/entities/__tests__/configuration.ts
```diff
@@ -20,6 +21,9 @@ export default (): ReturnType<typeof configuration> => ({
     default: faker.number.int(),
   },
   httpClient: { requestTimeout: faker.number.int() },
+  log: {
+    level: sample(['error', 'warn', 'info', 'debug']) ?? 'debug',
```
fmrsabino, line 25: What use case did you consider to sample from different levels for logging in the tests?

src/config/entities/configuration.ts
```diff
@@ -61,6 +61,9 @@ export default () => ({
       process.env.HTTP_CLIENT_REQUEST_TIMEOUT_MILLISECONDS ?? `${5_000}`,
     ),
   },
+  log: {
+    level: process.env.LOG_LEVEL || 'debug',
```
fmrsabino, line 65: What happens if `LOG_LEVEL` is a random string (e.g. `foo`)?

hectorgomezv: In that case, the service won't crash but it wouldn't log anything 🤔 Maybe it is better to include the log level validation in [its schema](https://github.com/safe-global/safe-client-gateway-nest/blob/c4e872915755574a83a18efeee91172713dc6167/src/config/configuration.validator.ts#L3) to avoid that scenario and fail fast, what do you think?

src/config/entities/__tests__/configuration.ts
hectorgomezv: I took inspiration from `npm` logging levels described [here](https://github.com/winstonjs/winston#logging-levels) but I didn't take all of them for the sake of brevity. Maybe it's better to sample from all the _allowed_ values (related to https://github.com/safe-global/safe-client-gateway-nest/pull/516#discussion_r1245332342)

fmrsabino: > I took inspiration from `npm` logging levels described [here](https://github.com/winstonjs/winston#logging-levels) but I didn't take all of them for the sake of brevity. Maybe it's better to sample from all the _allowed_ values (related to [#516 (comment)](https://github.com/safe-global/safe-client-gateway-nest/pull/516#discussion_r1245332342))

My question was not about what different logging levels exist but why do we need to sample between logging levels. What use case did you consider?

src/config/entities/configuration.ts
fmrsabino: > In that case, the service won't crash but it wouldn't log anything 🤔 Maybe it is better to include the log level validation in [its schema](https://github.com/safe-global/safe-client-gateway-nest/blob/c4e872915755574a83a18efeee91172713dc6167/src/config/configuration.validator.ts#L3) to avoid that scenario and fail fast, what do you think?

Yes I think this is a good solution 👍 We should fail fast if the log level is set but is not recognised as a valid one.

src/config/entities/__tests__/configuration.ts
hectorgomezv: My initial thought was to provide realistic random values to test classes as we do with the rest of the test configuration properties, but maybe this doesn't make a lot of sense and we could have a fixed value here. In any case, I think this value wouldn't be used in any place since the mocked `ILoggingService` we use in tests [redirects to the console](https://github.com/safe-global/safe-client-gateway-nest/blob/c4e872915755574a83a18efeee91172713dc6167/src/logging/__tests__/test.logging.module.ts#L4). Correct me if I'm wrong 🙂

src/config/entities/configuration.ts
hectorgomezv: Validation added in [d5b53b6](https://github.com/safe-global/safe-client-gateway-nest/pull/516/commits/d5b53b6ada8810891f3f206cc89af60d0f020b83)

src/config/entities/__tests__/configuration.ts
fmrsabino: I think we should set a default value (`debug` for example) else it can be confusing when you run your tests multiple times.

Logs that are generated by tests can also be useful for debugging. If the logging level is random, then we generate different logs on different runs for the same codebase.

hectorgomezv: Agree, changed in [317288e](https://github.com/safe-global/safe-client-gateway-nest/pull/516/commits/317288eac9dc6f63ee3f853a6c7dab6169e50db9)

src/config/configuration.validator.spec.ts
```diff
@@ -10,13 +10,25 @@ describe('Configuration validator', () => {
...
+  it('should detect missing mandatory configuration in production environment', () => {
     process.env.NODE_ENV = 'production';
     expect(() => validate(JSON.parse(fakeJson()))).toThrow(
-      /Mandatory configuration is missing: .*AUTH_TOKEN.*EXCHANGE_API_KEY/,
+      /must have required property 'AUTH_TOKEN'.*must have required property 'EXCHANGE_API_KEY'/,
     );
   });
 
+  it('should an invalid LOG_LEVEL configuration in production environment', () => {
+    process.env.NODE_ENV = 'production';
```
fmrsabino, line 21: Why do we need to set the environment to `production`? The validation should be the same in a testing and production scenarios right?

hectorgomezv: Yes, true. It's not necessary for this test, thanks. Changed in [cde8296](https://github.com/safe-global/safe-client-gateway-nest/pull/516/commits/cde8296f1a982e7facc15879d09d5021dc8099df)

fmrsabino (approved):
Nice! 🚀

PR #517, opened 2023-06-28
Read service version and build number from env

fmrsabino:
- Reads the service version `APPLICATION_VERSION` and build number `APPLICATION_BUILD_NUMBER` from the environment.
- The Docker image now takes as an argument a `VERSION` and a `BUILD_NUMBER`. These arguments are mapped to `APPLICATION_VERSION` and `APPLICATION_BUILD_NUMBER` respectively.
- The `About` entity no longer requires a `version` and `buildNumber` to have a value. It was previously set to an empty-string but `null` is a better representation for the absence of a value.

**Note regarding the value of `VERSION`**

In our CI it reads from `github.ref_name` which is the branch name in case of a PR or the tag name in case of a tagged release. Examples:

```javascript
// For staging deployments
{
  ...
  "version": "main"
}
``` 

```javascript
// For a v1.0.0 tag
{
  ...
  "version": "v1.0.0"
}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - run: |
+          BUILD_NUMBER=${{ github.sha }}
+          echo "BUILD_NUMBER=${BUILD_NUMBER::7}" >> "$GITHUB_ENV"
       - uses: docker/setup-qemu-action@v2.2.0
         with:
           platforms: arm64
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -82,6 +82,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
+      - run: |
+          BUILD_NUMBER=${{ github.sha }}
+          echo "BUILD_NUMBER=${BUILD_NUMBER::7}" >> "$GITHUB_ENV"
```
fmrsabino, line 87: Takes the first 7 characters of `BUILD_NUMBER`. This is not possible to do directly with `${{ github.sha }}`

hectorgomezv (approved):
👏🏻

PR #520, opened 2023-06-29
Add support for disabling logging

fmrsabino:
- Adds a new configuration `log.silent` which, if set to `true` disables logging (default is `false`).
- This configuration can be set using the environment variable `LOG_SILENT`.
- Sets CI `LOG_SILENT` value to `true` (no test logs by default)


hectorgomezv (approved):
👏🏻

PR #522, opened 2023-06-30
Mount cache in Docker image

### LINKED ISSUE #308: Mount cache in Docker image
- Change the `Dockerfile` definition to avoid Docker cache rewrites in each build, and decrease the build time in the CI pipeline. Please refer to https://github.com/safe-global/safe-gelato-relay-service/pull/89 for details.

hectorgomezv:
Closes #308 
Depends on https://github.com/safe-global/safe-client-gateway-nest/pull/526

This PR: 
 - Uses a production-only yarn workspace to build the production image.
 - Mounts a cache in CI to speed up package installs.


package.json
```diff
@@ -21,16 +21,17 @@
     "test:e2e": "jest --config ./test/jest-e2e.json --runInBand",
     "test:e2e:cov": "jest --coverage --config ./test/jest-e2e.json --runInBand",
     "test:all": "jest --config ./test/jest-all.json --runInBand",
-    "test:all:cov": "jest --coverage --config ./test/jest-all.json --runInBand",
-    "postinstall": "husky install"
+    "test:all:cov": "jest --coverage --config ./test/jest-all.json --runInBand"
   },
   "dependencies": {
+    "@nestjs/cli": "^10.0.5",
```
hectorgomezv, line 27: This is included as a production dependency since `nest build` is used when building the image. This could be changed by using `tsc` directly in the future, but I left this task out of the scope of this PR.

package.json
```diff
@@ -21,16 +21,17 @@
...
     "@nestjs/common": "^10.0.3",
     "@nestjs/config": "^3.0.0",
     "@nestjs/core": "^10.0.3",
     "@nestjs/platform-express": "^10.0.3",
     "@nestjs/swagger": "^7.0.7",
     "@node-redis/json": "^1.0.2",
+    "@types/lodash": "^4.14.195",
```
hectorgomezv, line 34: After some research, I finally found that not having this concrete dependency on the production image building process, makes the compilation fail because of the type checks. If you take a look to [this previous commit ](https://github.com/safe-global/safe-client-gateway-nest/pull/522/commits/7c457dd5db7f8f44b825cc9ec4c8ddf84ca8e889) you'll see I initially tried to solve these errors, blaming a `yarn workspaces focus --production` for installing a different typescript version or so. But the real root cause for those compilation errors was the compiler couldn't infer some variables were non-null after calling lodash's `isArray`.

So the final approach I took is to maintain `@types/lodash` as a production dependency, but I'm not against an alternative approach like removing these lodash-driven type guards and using vanilla TS for it. That would make us able to move this dependency to `devDependencies` again.

In any case, I'd do any follow-up work regarding this in a separate PR 🙂

src/app.provider.ts
```diff
@@ -96,7 +95,7 @@ export class TestAppProvider extends AppProvider {
   }
 
   protected getApp(module: any): Promise<INestApplication> {
-    if (!(module instanceof TestingModule))
+    if (module.constructor.name !== 'TestingModule')
```
hectorgomezv, line 98: I took this approach because `TestingModule` belongs to `@nestjs/testing`, which is not installed as a production dependency. I thought this was the simpler way to remove `TestingModule` import from this file, but alternatives like including `@nestjs/testing` as a production dependency, or even moving `TestAppProvider` class to a separate file under a `__tests__` folder (so it would get ignored when building the production image), could be also valid approaches.

As I said I think I took the simpler alternative but I'm open to suggestions.

fmrsabino: In my opinion I'd do something like it was done [here](https://github.com/safe-global/safe-gelato-relay-service/pull/99).

I'd say that it can be a separate PR on which this depends on. Isolating testing logic shouldn't be directly linked to Cache/Docker.

package.json
```diff
@@ -21,16 +21,17 @@
     "test:e2e": "jest --config ./test/jest-e2e.json --runInBand",
     "test:e2e:cov": "jest --coverage --config ./test/jest-e2e.json --runInBand",
     "test:all": "jest --config ./test/jest-all.json --runInBand",
-    "test:all:cov": "jest --coverage --config ./test/jest-all.json --runInBand",
-    "postinstall": "husky install"
```
fmrsabino, line 25: We should mention that git hooks are no longer installed automatically after project installation.

My suggestion is to write something like [this section](https://github.com/safe-global/safe-gelato-relay-service/blob/43f49bc377dbf658033bb2eb984e04f431e7b4cb/README.md?plain=1#L86-L95)

src/app.provider.ts
hectorgomezv: > I'd say that it can be a separate PR on which this depends on.

I agree, I will tackle this in a separate PR, and leave this one as a draft.

hectorgomezv: Done https://github.com/safe-global/safe-client-gateway-nest/pull/526 🙂

package.json
hectorgomezv: Added in [700b4a2](https://github.com/safe-global/safe-client-gateway-nest/pull/522/commits/700b4a242a29ead048e1ccee36cd4207ccc8ec9a), thanks!!

PR #525, opened 2023-06-30
Add new tag Github action

### LINKED ISSUE #521: Create tagged release job
A Github Action job should be triggered once a tag is created in order to create and upload a docker image with the respective tag.

hectorgomezv:
Closes #521 

This PR adds a new action which:
- Is meant to be executed only when a new release is pushed to the repository.
- Takes the build cache data from the `staging` built from the `main` branch atm.
- Pushes the release-tagged image to the docker images repository.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,13 +115,14 @@ jobs:
     needs: [ prettier, es-lint, tests-finish ]
     runs-on: ubuntu-latest
     env:
+      IMAGE_CACHE_SOURCE: staging
       IMAGE_REGISTRY: safeglobal
       IMAGE_NAME: safe-client-gateway-nest
     steps:
       - uses: actions/checkout@v3
       - name: Extract commit metadata
         id: meta
-        uses: docker/metadata-action@v3
+        uses: docker/metadata-action@v4
         with:
           images: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}
       - run: |
@@ -143,7 +144,7 @@ jobs:
             BUILD_NUMBER=${{ env.BUILD_NUMBER }}
             VERSION=${{ github.ref_name }}
           tags: ${{ steps.meta.outputs.tags }}
-          cache-from: type=registry,ref=${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}:latest
+          cache-from: type=registry,ref=${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}:${{ env.IMAGE_CACHE_SOURCE }}
           cache-to: type=inline
 
   autodeploy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   push:
-    tags:
-      - '*'
   pull_request:
+  release:
+    types: [ released ]
 
 jobs:
   prettier:
@@ -79,7 +79,7 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           parallel-finished: true
 
-  docker-publish:
+  docker-publish-staging:
     if: (github.event_name == 'push' && github.ref == 'refs/heads/main')
     needs: [ prettier, es-lint, tests-finish ]
     runs-on: ubuntu-latest
@@ -110,7 +110,7 @@ jobs:
           cache-from: type=registry,ref=${{ env.DOCKER_IMAGE_TAG }}
           cache-to: type=inline
 
-  tag-publish:
+  docker-publish-release:
     if: (github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/'))
     needs: [ prettier, es-lint, tests-finish ]
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -108,6 +110,43 @@ jobs:
...
+        with:
+          platforms: arm64
+      - uses: docker/setup-buildx-action@v2
+      - uses: docker/login-action@v2
+        with:
+          username: ${{ secrets.DOCKER_USER }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - uses: docker/build-push-action@v4
+        with:
+          platforms: linux/amd64,linux/arm64
```
hectorgomezv, line 141: I took this from the `staging` image build process. AFAIK we could omit `linux/amd64` and rely on `linux/arm64` only, but I think the build is made in parallel removing `linux/amd64` won't decrease the build time. I left both architectures as targets but no strong opinion on this subject.

.github/workflows/ci.yml
```diff
@@ -2,6 +2,8 @@ name: CI
 
 on:
   push:
+    tags:
+      - '*'
```
fmrsabino, line 6: Have you thought about using the [release event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release)?

We do it also in other projects.

.github/workflows/ci.yml
```diff
@@ -108,6 +110,43 @@ jobs:
...
+      IMAGE_CACHE_SOURCE: staging
+      IMAGE_REGISTRY: safeglobal
+      IMAGE_NAME: safe-client-gateway-nest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Extract commit metadata
+        id: meta
+        uses: docker/metadata-action@v4
+        with:
+          images: ${{ env.IMAGE_REGISTRY }}/${{ env.IMAGE_NAME }}
```
fmrsabino, line 127: Do we need this action?

.github/workflows/ci.yml
hectorgomezv: I was under the impression the issue was targeting creating an image on each tag, not only releases. But maybe I've misinterpreted it, so we could use [a trigger like this one](https://github.com/safe-global/safe-transaction-service/blob/cd330d0c0c81bdf2cbd233ec7564f4b5a03ce2c9/.github/workflows/python.yml#L8) and just generate images on release events, what is your opinion on this?

.github/workflows/ci.yml
fmrsabino: > If we restrict the target to releases I think we could just use `github.event.release.tag_name` and remove this action 👍

You don't need the release event or this action to get the tag name. You can get it via `github.ref_name`. https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

This PR should accomplish the same thing (but for the staging action): https://github.com/safe-global/safe-client-gateway-nest/pull/517

.github/workflows/ci.yml
fmrsabino: > I was under the impression the issue was targeting creating an image on each tag, not only releases. But maybe I've misinterpreted it, so we could use [a trigger like this one](https://github.com/safe-global/safe-transaction-service/blob/cd330d0c0c81bdf2cbd233ec7564f4b5a03ce2c9/.github/workflows/python.yml#L8) and just generate images on release events, what is your opinion on this?

Something like that works. Regarding the issue, maybe it was not clear. Not that this doesn't work but we should compare different solutions. [The Safe Config Service also uses the release event](https://github.com/safe-global/safe-config-service/blob/3877fdc19a9e3109b0f7515025d9c3cecf099c58/.github/workflows/test.yml#L11-L12)

.github/workflows/ci.yml
hectorgomezv: Changed in [494a482](https://github.com/safe-global/safe-client-gateway-nest/pull/525/commits/494a482d3dfb77fdcf3eed5ca9ab5a3e4b7ba808)

.github/workflows/ci.yml
hectorgomezv: I've took the Config Service as model to unify both services pipeline (also renaming `docker-publish` action to `docker-publish-staging`) in [494a482](https://github.com/safe-global/safe-client-gateway-nest/pull/525/commits/494a482d3dfb77fdcf3eed5ca9ab5a3e4b7ba808)

hectorgomezv:
I've also updated the PR description to conform to the latest changes 🙂

.github/workflows/ci.yml
```diff
@@ -108,9 +110,43 @@ jobs:
           cache-from: type=registry,ref=${{ env.DOCKER_IMAGE_TAG }}
           cache-to: type=inline
 
+  docker-publish-release:
+    if: (github.event_name == 'push' && startsWith(github.event.ref, 'refs/tags/'))
```
fmrsabino, line 114: From the Safe Config Service:

```
if: (github.event_name == 'release' && github.event.action == 'released')
```

It's a better match in my opinion given the event that triggered it.

https://github.com/safe-global/safe-config-service/blob/3877fdc19a9e3109b0f7515025d9c3cecf099c58/.github/workflows/test.yml#L190C5-L190C78

fmrsabino (changes requested):

.github/workflows/ci.yml
hectorgomezv: Absolutely! Thanks.

Changed in[374283a](https://github.com/safe-global/safe-client-gateway-nest/pull/525/commits/374283a78f1744be50dad3197fcaa262d6539a09)

fmrsabino (approved):
🚀 🚀 🚀

PR #527, opened 2023-07-03
Use overrideModule to replace modules

fmrsabino:
- Uses `overrideModule` to override the production modules with testing ones.
- Extracts the dynamic module which was built in place in `AppModule` so that the module can be replaced in a testing scenario using its direct reference. Context: https://github.com/nestjs/nest/issues/11967
- Uses `AppModule` throughout the tests instead of the modules for each feature. This way the setup between the tests can be shared, and all the required modules should be available by default.

`overrideModule` was introduced in NestJS 10 – https://github.com/nestjs/nest/releases/tag/v10.0.0

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -35,6 +35,9 @@ import { RouteLoggerInterceptor } from './routes/common/interceptors/route-logge
 import { NotFoundLoggerMiddleware } from './middleware/not-found-logger.middleware';
 import configuration from './config/entities/configuration';
 
+// See https://github.com/nestjs/nest/issues/11967
+export const configurationModule = ConfigurationModule.register(configuration);
+
 @Module({
   imports: [
     // features
@@ -65,7 +68,7 @@ import configuration from './config/entities/configuration';
         idGenerator: () => uuidv4(),
       },
     }),
-    ConfigurationModule.register(configuration),
+    configurationModule,
     DomainModule,
     NetworkModule,
     RequestScopedLoggingModule,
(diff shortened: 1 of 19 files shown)


hectorgomezv:
> What's your opinion on using the AppModule directly instead of each feature?

I think your summary is accurate. I've tested it and the running time for unit tests increased slightly, but I'm happy with that for the sake of having a simpler/understandable setup in the tests. I mean, I think overriding the `AppModule` is far more intuitive and easy for a newcomer than trying to understand why the test class has to import for example`DomainModule` or `ValidationModule` (i.e. knowing how the inner dependencies of the project relate to each other). 

So I'm happy with the change and I see it as a good addition 👏🏻

hectorgomezv (approved):
🚀 🚀 🚀

fmrsabino:
> > What's your opinion on using the AppModule directly instead of each feature?
> 
> I think your summary is accurate. I've tested it and the running time for unit tests increased slightly, but I'm happy with that for the sake of having a simpler/understandable setup in the tests. I mean, I think overriding the `AppModule` is far more intuitive and easy for a newcomer than trying to understand why the test class has to import for example`DomainModule` or `ValidationModule` (i.e. knowing how the inner dependencies of the project relate to each other).
> 
> So I'm happy with the change and I see it as a good addition 👏🏻

I agree. I think it's worth the additional time it takes to setup all the modules via the `AppModule`. I'll keep these changes in then! 🙌

PR #528, opened 2023-07-03
Add ExceptionFilter to log uncaught errors

fmrsabino:
Depends on #527 

- Adds an `ExceptionFilter` which logs all errors that do not have a filter associated with it.
- Errors were already being logged to the console (default NestJS logger behavior). This changes that default to use our own `ILoggingService`.
- Moves `DataSourceErrorFilter` from `AppProvider` into `AppModule` – the new exception filter requires dependency injection to be in place (which cannot be done via `app.useGlobalFilters()` and, in order to have a standard way of adding exception filters, both of them are now provided by `AppModule`

### DIFF
--- a/src/routes/common/filters/global-error.filter.spec.ts
+++ b/src/routes/common/filters/global-error.filter.spec.ts
@@ -1,0 +1,64 @@
+import { Controller, Get, HttpException, HttpStatus } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import { TestAppProvider } from '../../../__tests__/test-app.provider';
+import * as request from 'supertest';
+import { APP_FILTER } from '@nestjs/core';
+import { GlobalErrorFilter } from './global-error.filter';
+import { TestLoggingModule } from '../../../logging/__tests__/test.logging.module';
+import { ConfigurationModule } from '../../../config/configuration.module';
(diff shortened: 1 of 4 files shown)


fmrsabino:
Moved to draft to add a test.

fmrsabino:
@safe-global/safe-client-gateway-maintainers ready for review

src/routes/common/filters/global-error.filter.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  @Get('http-exception')
+  async httpException() {
+    throw new HttpException(
+      { message: 'Some http exception' },
+      HttpStatus.BAD_GATEWAY,
+    );
+  }
+
+  @Get('non-http-exception')
+  async bar() {
```
hectorgomezv, line 22:
```suggestion
  async nonHttpException() {
```

hectorgomezv (changes requested):
Great job 🚀  just requested a small change.

src/routes/common/filters/global-error.filter.spec.ts
fmrsabino: Ups! 😅 Thank you! Added.

hectorgomezv (approved):

PR #534, opened 2023-07-04
Validate datasource pages metadata

### LINKED ISSUE #71: Validate paginated data coming from datasources
More context [here](https://github.com/5afe/safe-client-gateway-nest/pull/66#discussion_r959683933).

Currently we are validating data coming in a paginated fashion from Safe Transaction Service and Safe Config Service, but we aren't validating the page format itself. Instead of iterating the array of elements and we can define a `Page<T>` schema for each `T` and validate the complete payload using composed JSONSchemas.

hectorgomezv:
Closes #71 

This PR aims to implement `Page` entity validation for use in several endpoints. Currently, the application is not validating the metadata (`count`, `previous`, `next`) coming from external data sources when requesting a page of data items. So this PR:

- Creates an `IPageValidator` interface, which can be optionally implemented by each validator.
- This `IPageValidator` contains one function only, `validatePage`, which will be responsible for validating the metadata and returning a `Page<T>`, being `T` the content (`results`) of the `Page`.
- Removes validation by loops (`page.results.map(item => validator.validate(item))`) in favor of validating all the items at once, which should increase the performance.
- Adds unit tests validating malformed `Page` objects would trigger validation errors.

### DIFF
--- a/src/domain/entities/schemas/page.schema.factory.ts
+++ b/src/domain/entities/schemas/page.schema.factory.ts
@@ -1,0 +1,17 @@
+import { Schema, SchemaObject } from 'ajv';
+
+export function buildPageSchema(
+  keyRef: string,
+  itemSchema: SchemaObject,
+): Schema {
+  return <SchemaObject>{
+    $id: keyRef,
+    type: 'object',
+    properties: {
+      count: { type: 'number' },
+      next: { type: 'string', nullable: true },
+      previous: { type: 'string', nullable: true },
+      results: { type: 'array', items: { $ref: itemSchema.$id } },
+    },
+  };
+}
--- a/src/domain/messages/entities/schemas/message.schema.ts
+++ b/src/domain/messages/entities/schemas/message.schema.ts
@@ -1,4 +1,5 @@
 import { Schema } from 'ajv';
+import { buildPageSchema } from '../../../entities/schemas/page.schema.factory';
 
 export const messageConfirmationSchema: Schema = {
   $id: 'https://safe-client.safe.global/schemas/messages/message-confirmation.json',
@@ -40,3 +41,8 @@ export const messageSchema: Schema = {
     'confirmations',
   ],
 };
+
+export const messagePageSchema: Schema = buildPageSchema(
+  'https://safe-client.safe.global/schemas/messages/message-page.json',
+  messageSchema,
+);
--- a/src/domain/safe/entities/schemas/transaction-type.schema.ts
+++ b/src/domain/safe/entities/schemas/transaction-type.schema.ts
@@ -1,4 +1,5 @@
 import { Schema } from 'ajv';
+import { buildPageSchema } from '../../../entities/schemas/page.schema.factory';
 
 export const transactionTypeSchema: Schema = {
   $id: 'https://safe-client.safe.global/schemas/safe/transaction-type.json',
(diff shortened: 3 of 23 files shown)


src/domain/entities/schemas/page.schema.factory.ts
```diff
@@ -0,0 +1,17 @@
...
+  return <SchemaObject>{
+    $id: keyRef,
+    type: 'object',
+    properties: {
+      count: { type: 'number' },
+      next: { type: 'string', nullable: true },
+      previous: { type: 'string', nullable: true },
+      results: { type: 'array', items: { $ref: itemSchema.$id } },
+    },
+  };
```
hectorgomezv, line 16: This function is meant to automatize the creation of `Page<T>` schemas.

I changed the approach while developing the feature: at first, I tried to automatize the `$id` field, by just adding a suffix to the `$id` of the input `itemSchema`, something like:
```
return <SchemaObject>{
  $id: itemSchema.replace('.json', '-page.json'),
  ...
```
Using this approach the function would take just one parameter, because `keyRef` would not be necessary anymore. But later I took the opposite approach and added `keyRef` parameter to be explicitly set by the caller. Since this `$id` will be referenced in several places (by validators when registering the schema) I didn't want to rely on inner logic. A third option consists of exporting a function that calculates the `$id` but it seemed a bit overkill to me.

I'm good with the current approach but also open to feedback of course.

src/domain/safe/entities/schemas/transaction-type.schema.ts
```diff
@@ -1,4 +1,5 @@
 import { Schema } from 'ajv';
+import { buildPageSchema } from '../../../entities/schemas/page.schema.factory';
 
 export const transactionTypeSchema: Schema = {
```
hectorgomezv, line 4: This is out of the context of this PR, but what about renaming `transactionType*` to simply `transaction*`?

This `TransactionTypeValidator` is used only [here](https://github.com/safe-global/safe-client-gateway-nest/blob/efe7832c9bef6446f6e630058aa70f1522c5a6d9/src/domain/safe/safe.repository.ts#L281) to validate `Transaction` items 🤔

src/domain/safe/safe.repository.ts
```diff
@@ -63,9 +63,7 @@ export class SafeRepository implements ISafeRepository {
       limit,
       offset,
     );
-    page.results.map((transfer) => this.transferValidator.validate(transfer));
-
-    return page;
+    return this.transferValidator.validatePage(page);
```
hectorgomezv, line 66: These changes are the main point of the PR.

src/domain/messages/entities/schemas/message.schema.ts
```diff
@@ -40,3 +41,8 @@ export const messageSchema: Schema = {
     'confirmations',
   ],
 };
+
+export const messagePageSchema: Schema = buildPageSchema(
+  'https://safe-client.safe.global/schemas/messages/message-page.json',
```
iamacook, line 46: Out of context of this PR, but what do you think about moving the `keyRef` strings to constants? We're using them in multiple places and they're easily mistyped.

iamacook (approved):

src/domain/messages/entities/schemas/message.schema.ts
hectorgomezv: Thank you @iamacook, after a second thought, I think this constants extraction fits better in a separate PR as you suggested!

PR #536, opened 2023-07-05
Invalidate safe apps cache

fmrsabino:
Invalidates the Safe Apps cache on a `Chains` invalidation hook event.

### DIFF
--- a/src/domain/flush/flush.repository.ts
+++ b/src/domain/flush/flush.repository.ts
@@ -18,7 +18,11 @@ export class FlushRepository implements IFlushRepository {
   async execute(pattern: InvalidationPatternDto): Promise<void> {
     switch (pattern.invalidate) {
       case InvalidationTarget[InvalidationTarget.Chains]:
-        return this.configApi.clearChains();
+        await Promise.allSettled([
+          this.configApi.clearChains(),
+          this.configApi.clearSafeApps(),
+        ]);
+        break;
       default:
         this.loggingService.debug(
           `Unknown flush pattern ${pattern.invalidate}`,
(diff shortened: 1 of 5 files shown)


src/domain/flush/flush.repository.ts
```diff
@@ -18,7 +18,11 @@ export class FlushRepository implements IFlushRepository {
   async execute(pattern: InvalidationPatternDto): Promise<void> {
     switch (pattern.invalidate) {
       case InvalidationTarget[InvalidationTarget.Chains]:
-        return this.configApi.clearChains();
+        await Promise.allSettled([
```
hectorgomezv, line 21: Not sure I get the usage of `allSettled` here: is it the intention to ignore errors when any of the promises in the array rejects?

fmrsabino: `Promise.all` rejects immediately after any rejection whereas `Promise.allSettled` waits for every promise to complete.

Regarding error handling, we can log it for each promise that was rejected and compute the error state from there what do you think?

hectorgomezv: > Regarding error handling, we can log it for each promise that was rejected and compute the error state from there what do you think?

Yes, that was my point, because we are not using the result in any case, so I think both `all` and `allSettled` would have the same effect here (`all` rejects immediately by throwing an exception, but it doesn't cancel the promises in execution). A good use case for `allSettled` would be logging the errored promises as you said.

I'm fine with using `allSettled` here, I asked just for the sake of curiosity. And if we find value in managing the errors, probably we could do it in a separate PR 👍

hectorgomezv (approved):
🚀🚀🚀

src/domain/flush/flush.repository.ts
hectorgomezv: Well, also it's worth noticing that `allSettled` would led us to have silent errors in that block (it would return an object containing the errors, but this output is ignored), while `all` would throw an exception if I'm not wrong 🤔

fmrsabino: Let's go with `Promise.all` then 👍 [ad54ab8](https://github.com/safe-global/safe-client-gateway-nest/pull/536/commits/ad54ab87925eba912c5d5359296a1364d4839e99)

hectorgomezv (approved):

PR #538, opened 2023-07-07
Log as error level only for server side errors

fmrsabino:
Sets the `GlobalErrorFilter` to log as error level only on server error codes. Other codes will be logged as info level.


hectorgomezv (approved):
🚀

PR #548, opened 2023-07-11
Fix bad scale for timezoneOffset query param

### LINKED ISSUE #547: [transactions/history endpoint] Miscalculated timestamp values
**Problem**

The service responds with invalid timestamps for some requests. It seems (to be consolidated) that the `timestamp` calculation is not working properly when the client sends the `timezone_offset` query string.

Example: [request](https://safe-client-nest.staging.5afe.dev/v1/chains/5/safes/0xC642bEc0e4f69815e89723C09caA8DAa524cAc9c/transactions/history?timezone_offset=7200000&cursor=limit%3D20%26offset%3D20).

hectorgomezv:
Closes #547 

This PR:
- Fixes the expected scale for the `timezoneOffset` query param. The clients are sending this value in milliseconds, but the service was expecting to have it in seconds.

### DIFF
--- a/src/routes/transactions/mappers/transactions-history.mapper.ts
+++ b/src/routes/transactions/mappers/transactions-history.mapper.ts
@@ -143,11 +143,11 @@ export class TransactionsHistoryMapper {
    * Returns a day {@link Date } at 00:00:00 from the input timestamp.
    *
    * @param timestamp - date to convert
-   * @param timezoneOffset - Offset of time zone in seconds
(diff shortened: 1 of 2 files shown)


hectorgomezv:
Source in the old CGW: https://github.com/safe-global/safe-client-gateway/blob/eb481938a49c29b979fe98a5577f719f42a2b2eb/src/routes/transactions/handlers/history.rs#L26

fmrsabino (approved):
🚀

liliya-soroka:
fixed on mobile

PR #563, opened 2023-07-13
fix: return transaction value directly

iamacook:
This fixes high values from being returned in scientific notation for custom transactions. The value of the transaction is returned directly instead.

`mapCustomTransaction` was returning the casted number of transaction values. This meant very high values were being returned in scientific notation:

```js
Number('1000000000000000000000000').toString() === '1e+24'
```

Example transaction on the [transaction service](https://safe-transaction-gnosis-chain.safe.global/api/v1/multisig-transactions/0xba31bc6dbda0ad9451e087843b96ba6bccc1d2cbbb8c9ac52df20f105cf727ea/) and on the [gateway](https://safe-client.safe.global/v1/chains/100/transactions/multisig_0x10E4597fF93cbee194F4879f8f1d54a370DB6969_0xba31bc6dbda0ad9451e087843b96ba6bccc1d2cbbb8c9ac52df20f105cf727ea).

### DIFF
--- a/src/routes/transactions/mappers/common/custom-transaction.mapper.spec.ts
+++ b/src/routes/transactions/mappers/common/custom-transaction.mapper.spec.ts
@@ -25,7 +25,6 @@ describe('Multisig Custom Transaction mapper (Unit)', () => {
   it('should build a CustomTransactionInfo with null actionCount', async () => {
     const toAddress = new AddressInfo(faker.finance.ethereumAddress());
     addressInfoHelper.getOrDefault.mockResolvedValue(toAddress);
-    const value = faker.number.int();
     const dataSize = faker.number.int();
     const chainId = faker.string.numeric();
     const transaction = multisigTransactionBuilder()
@@ -34,7 +33,6 @@ describe('Multisig Custom Transaction mapper (Unit)', () => {
 
     const customTransaction = await mapper.mapCustomTransaction(
       transaction,
-      value,
       dataSize,
       chainId,
     );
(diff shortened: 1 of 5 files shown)


src/routes/transactions/mappers/common/native-coin-transfer.mapper.ts
```diff
@@ -26,7 +24,7 @@ export class NativeCoinTransferMapper {
     );
 
     return new TransferTransactionInfo(
-      new AddressInfo(safe.address),
```
iamacook, line 29: I altered this to remove the unnecessary argument.

src/routes/transactions/mappers/common/custom-transaction.mapper.spec.ts
```diff
@@ -127,13 +148,13 @@ describe('Multisig Custom Transaction mapper (Unit)', () => {
     const toAddress = new AddressInfo(faker.finance.ethereumAddress());
     addressInfoHelper.getOrDefault.mockResolvedValue(toAddress);
     const method = faker.word.sample();
-    const value = faker.number.int();
+    const value = '0';
```
fmrsabino, line 151: Was this changed to avoid big values with faker?

iamacook: I think it was wrong. See line 157 below pre-change. A cancellation transaction [has a value of '0'](https://github.com/safe-global/safe-client-gateway-nest/blob/scientific-notation/src/routes/transactions/mappers/common/custom-transaction.mapper.ts#L71).

fmrsabino: Oh I see. Yeah this is about cancellation transactions so it makes sense 👍

fmrsabino (approved):
Some tests need to be fixed but besides that, looks good to me! 🚀

hectorgomezv (approved):

iamacook:
> Some tests need to be fixed but besides that, looks good to me! 🚀

I reverted the change related to the broken tests (removing the unnecessary argument) as it otherwise requires a lot more changes/test updates. I'd sooner do it in another PR instead.

fmrsabino (approved):

PR #567, opened 2023-07-14
refactor: remove unnecessary `Safe` usage

iamacook:
This removes the `safe` argument from `mapNativeCoinTransfer` and updates all relevant other functions and tests accordingly.

### DIFF
--- a/src/routes/transactions/mappers/common/transaction-info.mapper.ts
+++ b/src/routes/transactions/mappers/common/transaction-info.mapper.ts
@@ -1,8 +1,6 @@
 import { Inject, Injectable } from '@nestjs/common';
 import { ModuleTransaction } from '../../../../domain/safe/entities/module-transaction.entity';
 import { MultisigTransaction } from '../../../../domain/safe/entities/multisig-transaction.entity';
-import { Operation } from '../../../../domain/safe/entities/operation.entity';
-import { Safe } from '../../../../domain/safe/entities/safe.entity';
 import { TokenRepository } from '../../../../domain/tokens/token.repository';
 import { ITokenRepository } from '../../../../domain/tokens/token.repository.interface';
 import { TokenType } from '../../../balances/entities/token-type.entity';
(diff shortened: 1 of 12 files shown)


src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -57,19 +54,10 @@ export class MultisigTransactionInfoMapper {
     const dataSize =
       dataByteLength >= 2 ? Math.floor((dataByteLength - 2) / 2) : 0;
 
-    if (this.isCustomTransaction(value, dataSize, transaction.operation)) {
-      return await this.customTransactionMapper.mapCustomTransaction(
-        transaction,
-        dataSize,
-        chainId,
-      );
-    }
```
iamacook, line 66: This is unnecessary as it is the ["default" map of this function](https://github.com/safe-global/safe-client-gateway-nest/blob/refactor-argument/src/routes/transactions/mappers/common/transaction-info.mapper.ts#L114-L118).

hectorgomezv (approved):

hectorgomezv (approved):
Great job 👏🏻

src/routes/transactions/mappers/common/transaction-info.mapper.ts
hectorgomezv: Not sure about this change because I think order matters here. This is the default but the condition to consider a transaction _custom_ is:
```
(value > 0 && dataSize > 0) || operation !== 0
```
But not sure this overlaps with other conditions which are checked later in the code, so I think we need to check the tests are covering all the cases before doing this change. Maybe it's possible to refactor, but I'd add more exhaustive testing here.

hectorgomezv (approved):

src/routes/transactions/mappers/common/transaction-info.mapper.ts
iamacook: Considering the unsurity, I've reverted this. We can revisit it later as increasing the test coverage is a necessity either way.

PR #578, opened 2023-07-19
Add configuration param for contract/token not found cache TTL

### LINKED ISSUE #577: Configure not found error cache TTL independently
**Context**

Currently, we are assuming the same TTL for all kinds of errors happening when retrieving data from data sources. This issue targets the ability to set a specific/independent TTL for `NOT_FOUND` responses.

hectorgomezv:
Closes #577 

This PR:
 - Adds a CONTRACT_NOT_FOUND_ERROR_TTL_SECONDS environment variable to configure the time to live for cached errors regarding the contracts retrieval from the TX Service.
 -  - Adds a TOKEN_NOT_FOUND_ERROR_TTL_SECONDS environment variable to configure the time to live for cached errors regarding the tokens retrieval from the TX Service.
 - Modifies both production and testing `configuration.ts` accordingly.
 - Adds a unit test checking the new configuration option is passed to the  `CacheFirstDataSource` and to the underlying `CacheService` implementations.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -22,6 +22,8 @@
 # The default cache expiration time in seconds if none is set
 # (default=60)
 #EXPIRATION_TIME_DEFAULT_SECONDS
+#CONTRACT_NOT_FOUND_ERROR_TTL_SECONDS
+#TOKEN_NOT_FOUND_ERROR_TTL_SECONDS
 
 # Authorization token to use privileged endpoints.
 # The AUTH_TOKEN should always be set
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -31,6 +31,14 @@ export default () => ({
     level: process.env.LOG_LEVEL || 'debug',
     silent: process.env.LOG_SILENT?.toLowerCase() === 'true',
   },
+  notFoundErrorTTLSeconds: {
+    contract: parseInt(
+      process.env.CONTRACT_NOT_FOUND_ERROR_TTL_SECONDS ?? `${60 * 60}`,
+    ),
+    token: parseInt(
+      process.env.TOKEN_NOT_FOUND_ERROR_TTL_SECONDS ?? `${60 * 60}`,
+    ),
+  },
   redis: {
     host: process.env.REDIS_HOST || 'localhost',
     port: process.env.REDIS_PORT || '6379',
--- a/src/datasources/cache/cache.first.data.source.spec.ts
+++ b/src/datasources/cache/cache.first.data.source.spec.ts
@@ -117,7 +117,7 @@ describe('CacheFirstDataSource', () => {
     expect(fakeCacheService.keyCount()).toBe(1);
   });
 
-  it('should cache errors with the expected TTL', async () => {
+  it('should cache not found errors with the default TTL', async () => {
     const mockCache = jest.mocked({
       get: jest.fn(),
       set: jest.fn(),
@@ -148,5 +148,49 @@ describe('CacheFirstDataSource', () => {
     ).rejects.toThrowError(expectedError);
 
     expect(mockCache.set).toHaveBeenCalledWith(cacheDir, expect.anything(), 30);
+  });
+
+  it('should cache not found errors with a specific TTL', async () => {
+    const mockCache = jest.mocked({
+      get: jest.fn(),
+      set: jest.fn(),
+      delete: jest.fn(),
+    } as unknown as ICacheService);
+
+    cacheFirstDataSource = new CacheFirstDataSource(
+      mockCache,
+      mockNetworkService,
+      mockLoggingService,
+    );
+
+    const targetUrl = faker.internet.url({ appendSlash: false });
+    const cacheDir = new CacheDir(faker.word.sample(), faker.word.sample());
+    const expectedError = new NetworkResponseError(404);
+    const notFoundErrorTTLSeconds = faker.number.int();
+    mockCache.get.mockResolvedValue(undefined);
+    mockNetworkService.get.mockImplementation((url) => {
+      switch (url) {
+        case targetUrl:
+          return Promise.reject(expectedError);
+        default:
+          return Promise.reject(`No matching rule for url: ${url}`);
+      }
+    });
+
+    await expect(
+      cacheFirstDataSource.get(
+        cacheDir,
+        targetUrl,
+        undefined,
+        faker.number.int(),
+        notFoundErrorTTLSeconds,
+      ),
+    ).rejects.toThrowError(expectedError);
+
+    expect(mockCache.set).toHaveBeenCalledWith(
+      cacheDir,
+      expect.anything(),
+      notFoundErrorTTLSeconds,
+    );
   });
 });
--- a/src/datasources/cache/cache.first.data.source.ts
+++ b/src/datasources/cache/cache.first.data.source.ts
@@ -24,7 +24,7 @@ import {
  */
 @Injectable()
 export class CacheFirstDataSource {
-  private readonly ERROR_TTL_SECONDS = 30;
+  private readonly defaultNotFoundErrorTTLSeconds = 30;
 
   constructor(
     @Inject(CacheService) private readonly cacheService: ICacheService,
@@ -36,18 +36,20 @@ export class CacheFirstDataSource {
    * Gets the cached value behind the {@link CacheDir}.
    * If the value is not present, it tries to get the respective JSON
    * payload from {@link url}.
-   * 404 errors are cached with {@link ERROR_TTL_SECONDS} seconds expiration time.
+   * 404 errors are cached with {@link notFoundErrorTTLSeconds} seconds expiration time.
    *
    * @param cacheDir - {@link CacheDir} containing the key and field to be used to retrieve from cache
    * @param url - the HTTP endpoint to retrieve the JSON payload
    * @param networkRequest - the HTTP request to be used if there is a cache miss
    * @param expireTimeSeconds - the time to live in seconds for the payload behind {@link CacheDir}
+   * @param notFoundErrorTTLSeconds - the time to live in seconds for the error when the item is not found
    */
   async get<T>(
     cacheDir: CacheDir,
     url: string,
     networkRequest?: NetworkRequest,
     expireTimeSeconds?: number,
+    notFoundErrorTTLSeconds?: number,
   ): Promise<T> {
     const cached = await this.cacheService.get(cacheDir);
     if (cached != null) {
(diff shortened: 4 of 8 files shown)


src/config/entities/configuration.ts
```diff
@@ -31,6 +31,14 @@ export default () => ({
...
     silent: process.env.LOG_SILENT?.toLowerCase() === 'true',
   },
+  notFoundErrorTTLSeconds: {
+    contract: parseInt(
+      process.env.CONTRACT_NOT_FOUND_ERROR_TTL_SECONDS ?? `${60 * 60}`,
+    ),
+    token: parseInt(
+      process.env.TOKEN_NOT_FOUND_ERROR_TTL_SECONDS ?? `${60 * 60}`,
+    ),
+  },
```
hectorgomezv, line 41: I'm setting 1 hour here as the default for both, but this is open to discussion.

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -36,18 +36,20 @@ export class CacheFirstDataSource {
...
    * If the value is not present, it tries to get the respective JSON
    * payload from {@link url}.
-   * 404 errors are cached with {@link ERROR_TTL_SECONDS} seconds expiration time.
+   * 404 errors are cached with {@link notFoundErrorTTLSeconds} seconds expiration time.
    *
    * @param cacheDir - {@link CacheDir} containing the key and field to be used to retrieve from cache
    * @param url - the HTTP endpoint to retrieve the JSON payload
    * @param networkRequest - the HTTP request to be used if there is a cache miss
    * @param expireTimeSeconds - the time to live in seconds for the payload behind {@link CacheDir}
+   * @param notFoundErrorTTLSeconds - the time to live in seconds for the error when the item is not found
```
hectorgomezv, line 45: I choose `notFoundErrorTTLSeconds` as the name for this parameter. To stay consistent with the one we already have in place it could be `notFoundErrorExpireTimeSeconds` but I thought this was a bit too lengthy, specially adding the default: `defaultNotFoundErrorExpireTimeSeconds`. But this is open for discussion of course.

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -24,7 +24,7 @@ import {
  */
 @Injectable()
 export class CacheFirstDataSource {
-  private readonly ERROR_TTL_SECONDS = 30;
+  private readonly defaultNotFoundErrorTTLSeconds = 30;
```
fmrsabino, line 27: I would not keep this one. Else we are splitting the configuration across multiple stages:
- The config from the envar
- The default if the envar is not set
- The default if it does not apply to token or contracts.

I'd move the default to the config also.

src/datasources/cache/cache.first.data.source.ts
fmrsabino: Suggestion: `notFoundExpireTimeSeconds`. It's still a bit long but more in line with the other one.

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -36,18 +36,20 @@ export class CacheFirstDataSource {
...
    * @param networkRequest - the HTTP request to be used if there is a cache miss
    * @param expireTimeSeconds - the time to live in seconds for the payload behind {@link CacheDir}
+   * @param notFoundErrorTTLSeconds - the time to live in seconds for the error when the item is not found
    */
   async get<T>(
     cacheDir: CacheDir,
     url: string,
     networkRequest?: NetworkRequest,
     expireTimeSeconds?: number,
+    notFoundErrorTTLSeconds?: number,
```
fmrsabino, line 52: If we extract `defaultNotFoundErrorTTLSeconds` then we don't need to have this parameter as `undefined` which I think it's preferable.

.env.sample
```diff
@@ -22,6 +22,8 @@
 # The default cache expiration time in seconds if none is set
 # (default=60)
 #EXPIRATION_TIME_DEFAULT_SECONDS
+#CONTRACT_NOT_FOUND_ERROR_TTL_SECONDS
+#TOKEN_NOT_FOUND_ERROR_TTL_SECONDS
```
fmrsabino, line 26: This is missing what the envar represents.

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -149,4 +149,48 @@ describe('CacheFirstDataSource', () => {
 
     expect(mockCache.set).toHaveBeenCalledWith(cacheDir, expect.anything(), 30);
```
fmrsabino, line 150: We should not use 30 after this change.

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -149,4 +149,48 @@ describe('CacheFirstDataSource', () => {
...
+      get: jest.fn(),
+      set: jest.fn(),
+      delete: jest.fn(),
+    } as unknown as ICacheService);
+
+    cacheFirstDataSource = new CacheFirstDataSource(
+      mockCache,
+      mockNetworkService,
+      mockLoggingService,
+    );
```
fmrsabino, line 164: The test setup here looks a bit strange. On one hand we have the `beforeEach` which already sets up a `CacheFirstDataSource`. But we ignore it here to setup a different one.

fmrsabino (changes requested):

src/config/entities/configuration.ts
hectorgomezv: Agreed, I'll set these defaults to 60 seconds as well then.

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Yes, maybe it's better, thanks. 

I think this will also imply a change in the new environment variables names:
- `CONTRACT_NOT_FOUND_ERROR_TTL_SECONDS` -> `CONTRACT_NOT_FOUND_EXPIRE_TIME_SECONDS`
- `TOKEN_NOT_FOUND_ERROR_TTL_SECONDS` -> `TOKEN_NOT_FOUND_EXPIRE_TIME_SECONDS`

I'm totally fine with it, just commenting 🙂

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Yeah, I agree, initially I tried to not touch this wider default and just extract the contracts and tokens ones, targetting a further refactor in the future. But yes, maybe it's better to tackle this duplication/confusion now.

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Yes, I also thought the same, but I defined this as optional to stay consistent with `expireTimeSeconds`, which is optional but it's never called with an `undefined` value in the code. Maybe we should make both non-optional? What do you think? @fmrsabino

fmrsabino: But `expireTimeSeconds` is not required for some endpoints right?

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Changed in [f64c4da](https://github.com/safe-global/safe-client-gateway-nest/pull/578/commits/f64c4dad0e26f2530a411001a900ee3b255cc940)

src/config/entities/configuration.ts
```diff
@@ -31,6 +31,14 @@ export default () => ({
     level: process.env.LOG_LEVEL || 'debug',
     silent: process.env.LOG_SILENT?.toLowerCase() === 'true',
   },
+  notFoundErrorTTLSeconds: {
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [f64c4da](https://github.com/safe-global/safe-client-gateway-nest/pull/578/commits/f64c4dad0e26f2530a411001a900ee3b255cc940)

.env.sample
hectorgomezv: Added in [f64c4da](https://github.com/safe-global/safe-client-gateway-nest/pull/578/commits/f64c4dad0e26f2530a411001a900ee3b255cc940)

src/config/entities/configuration.ts
hectorgomezv: Changed in [f64c4da](https://github.com/safe-global/safe-client-gateway-nest/pull/578/commits/f64c4dad0e26f2530a411001a900ee3b255cc940)

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Right, that should be optional, but that's not the case for `notFoundExpireTimeSeconds`, which can be mandatory and will be explicitly set by the caller (`TransactionApi`, `ConfigApi`, etc.) 👍

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: I think this is wired that way to allow using a spy as `CacheService` and make assertions on the calls made. I copied the setup from [the test above this one](https://github.com/safe-global/safe-client-gateway-nest/blob/06dfc1a6e2b13ed39abb9a2b18f3b9fe30d4afdc/src/datasources/cache/cache.first.data.source.spec.ts#L120). Maybe we could think of a better way to achieve this, but I'd say it could be in a different PR, wdyt?

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: Changed in [4e30dc6](https://github.com/safe-global/safe-client-gateway-nest/pull/578/commits/4e30dc61d72b12d7b8baa789be526cca80bca083)

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: I've made it mandatory in [4e30dc6](https://github.com/safe-global/safe-client-gateway-nest/pull/578/commits/4e30dc61d72b12d7b8baa789be526cca80bca083)

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: Changed in [4e30dc6](https://github.com/safe-global/safe-client-gateway-nest/pull/578/commits/4e30dc61d72b12d7b8baa789be526cca80bca083)

fmrsabino (approved):

src/datasources/cache/cache.first.data.source.spec.ts
fmrsabino: Ok let's consider rewriting this in a separate PR 👍

PR #579, opened 2023-07-20
Enable keepAlive for http/s connections

fmrsabino:
- Sets the axios network client to keep-alive connections done both with http and https


hectorgomezv (approved):
🚀

PR #596, opened 2023-07-26
Add trusted filter when getting the last transaction to estimate the next nonce

### LINKED ISSUE #595: [POST multisig-transactions/estimations] Bad recommended nonce for untrusted transactions
**Problem**

The transaction queue filters out untrusted transactions, this is the correct behavior. But when the last transaction is a not-trusted one, this shouldn't be taken as the last transaction for the recommended nonce calculation.

**Solution**

The function `getLastTransactionSortedByNonce` should add a filter to not include non `trusted` transactions. See [the legacy implementation](https://github.com/safe-global/safe-client-gateway/blob/eb481938a49c29b979fe98a5577f719f42a2b2eb/src/routes/safes/handlers/estimations.rs#L27).

hectorgomezv:
Closes #595 

- Adds the `trusted` missing filter when calling the TX service to get the last transaction for recommended nonce calculation.

[(Reference in the Rust CGW implementation)](https://github.com/safe-global/safe-client-gateway/blob/eb481938a49c29b979fe98a5577f719f42a2b2eb/src/routes/safes/handlers/estimations.rs#L27)

### DIFF
--- a/src/domain/safe/safe.repository.ts
+++ b/src/domain/safe/safe.repository.ts
@@ -384,7 +384,7 @@ export class SafeRepository implements ISafeRepository {
         safeAddress,
         '-nonce',
         undefined,
-        undefined,
+        true,
         undefined,
         undefined,
         undefined,


fmrsabino (approved):
🚀

PR #610, opened 2023-08-01
Use anonymous interfaces for function args

fmrsabino:
- Replace the arguments of some functions with an interface containing those same arguments.
- The advantage of this approach is that we can now use the anonymous interface names to specify the value of each argument (similar to named arguments).
- Methods that were sharing the same type in multiple arguments were also converted – e.g., having `foo(chainId: string, safeAddress: string)` can cause issues if the order of the parameters change or the callee calls `foo(safeAddress, chainId)` (which is a valid call).
- Similarly, functions which had optional values can be a source of errors by the callee – e.g., having `bar(chainId: string, a?: number, c?: number)` and later introducing a new version of `bar` such as `bar(chainId: string, a?: number, b?: number, c?: number)` would not require any changes by the callers (although `c` is now being used as `b`)

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.ts
+++ b/src/datasources/transaction-api/transaction-api.service.ts
@@ -67,12 +67,10 @@ export class TransactionApi implements ITransactionApi {
     excludeSpam?: boolean;
   }): Promise<Balance[]> {
     try {
-      const cacheDir = CacheRouter.getBalanceCacheDir(
-        this.chainId,
-        args.safeAddress,
-        args.trusted,
-        args.excludeSpam,
-      );
+      const cacheDir = CacheRouter.getBalanceCacheDir({
+        chainId: this.chainId,
+        ...args,
+      });
       const url = `${this.baseUrl}/api/v1/safes/${args.safeAddress}/balances/usd/`;
       return await this.dataSource.get(
         cacheDir,
@@ -92,7 +90,10 @@ export class TransactionApi implements ITransactionApi {
   }
 
   async clearLocalBalances(safeAddress: string): Promise<void> {
-    const cacheKey = CacheRouter.getBalancesCacheKey(this.chainId, safeAddress);
+    const cacheKey = CacheRouter.getBalancesCacheKey({
+      chainId: this.chainId,
+      safeAddress,
+    });
     await this.cacheService.deleteByKey(cacheKey);
   }
 
@@ -120,14 +121,10 @@ export class TransactionApi implements ITransactionApi {
     excludeSpam?: boolean;
   }): Promise<Page<Collectible>> {
     try {
-      const cacheDir = CacheRouter.getCollectiblesCacheDir(
-        this.chainId,
-        args.safeAddress,
-        args.limit,
-        args.offset,
-        args.trusted,
-        args.excludeSpam,
-      );
+      const cacheDir = CacheRouter.getCollectiblesCacheDir({
+        chainId: this.chainId,
+        ...args,
+      });
       const url = `${this.baseUrl}/api/v2/safes/${args.safeAddress}/collectibles/`;
       return await this.dataSource.get(
         cacheDir,
@@ -149,7 +146,10 @@ export class TransactionApi implements ITransactionApi {
   }
 
   clearCollectibles(safeAddress: string): Promise<void> {
-    const key = CacheRouter.getCollectiblesKey(this.chainId, safeAddress);
+    const key = CacheRouter.getCollectiblesKey({
+      chainId: this.chainId,
+      safeAddress,
+    });
     return this.cacheService.deleteByKey(key).then(() => {
       return;
     });
@@ -193,7 +193,10 @@ export class TransactionApi implements ITransactionApi {
 
   async getSafe(safeAddress: string): Promise<Safe> {
     try {
-      const cacheDir = CacheRouter.getSafeCacheDir(this.chainId, safeAddress);
+      const cacheDir = CacheRouter.getSafeCacheDir({
+        chainId: this.chainId,
+        safeAddress,
+      });
       const url = `${this.baseUrl}/api/v1/safes/${safeAddress}`;
       return await this.dataSource.get(
         cacheDir,
@@ -208,7 +211,10 @@ export class TransactionApi implements ITransactionApi {
   }
 
   async clearSafe(safeAddress: string): Promise<void> {
-    const key = CacheRouter.getSafeCacheKey(this.chainId, safeAddress);
+    const key = CacheRouter.getSafeCacheKey({
+      chainId: this.chainId,
+      safeAddress,
+    });
     return this.cacheService.deleteByKey(key).then(() => {
       return;
     });
@@ -218,10 +224,10 @@ export class TransactionApi implements ITransactionApi {
   // Therefore, this data will live in cache until [defaultExpirationTimeInSeconds]
   async getContract(contractAddress: string): Promise<Contract> {
     try {
-      const cacheDir = CacheRouter.getContractCacheDir(
-        this.chainId,
+      const cacheDir = CacheRouter.getContractCacheDir({
+        chainId: this.chainId,
         contractAddress,
-      );
+      });
       const url = `${this.baseUrl}/api/v1/contracts/${contractAddress}`;
       return await this.dataSource.get(
         cacheDir,
@@ -244,15 +250,10 @@ export class TransactionApi implements ITransactionApi {
     offset?: number;
   }): Promise<Page<Delegate>> {
     try {
-      const cacheDir = CacheRouter.getDelegatesCacheDir(
-        this.chainId,
-        args.safeAddress,
-        args.delegate,
-        args.delegator,
-        args.label,
-        args.limit,
-        args.offset,
-      );
+      const cacheDir = CacheRouter.getDelegatesCacheDir({
+        chainId: this.chainId,
+        ...args,
+      });
       const url = `${this.baseUrl}/api/v1/delegates/`;
       return await this.dataSource.get(
         cacheDir,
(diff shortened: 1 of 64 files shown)


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -96,46 +96,49 @@ export class TransactionApi implements ITransactionApi {
...
-  ): Promise<Page<Collectible>> {
+  async getCollectibles(args: {
+    safeAddress: string;
+    limit?: number;
+    offset?: number;
+    trusted?: boolean;
+    excludeSpam?: boolean;
+  }): Promise<Page<Collectible>> {
     try {
       const cacheDir = CacheRouter.getCollectiblesCacheDir(
```
iamacook, line 123: All of the `CacheRouter` methods would also benefit from a similar change.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -96,46 +96,49 @@ export class TransactionApi implements ITransactionApi {
     await this.cacheService.deleteByKey(cacheKey);
   }
 
-  async getDataDecoded(data: string, to?: string): Promise<DataDecoded> {
+  async getDataDecoded(args: {
+    data: string;
+    to?: string;
+  }): Promise<DataDecoded> {
```
iamacook, line 102: This is a bit excessive imo. Should we now _always_ use an object for 2-3 arguments? (I won't mark other functions likr this.)

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -96,46 +96,49 @@ export class TransactionApi implements ITransactionApi {
...
     try {
       const url = `${this.baseUrl}/api/v1/data-decoder/`;
       const { data: dataDecoded } = await this.networkService.post(url, {
-        data,
-        to,
+        data: args.data,
+        to: args.to,
```
iamacook, line 107: You can use `data` directly here instead of create a new object.

src/domain/safe/safe.repository.ts
```diff
@@ -34,382 +34,382 @@ export class SafeRepository implements ISafeRepository {
...
+    limit?: number;
+    offset?: number;
+  }): Promise<Page<MultisigTransaction>> {
+    return this._getTransactionQueue({
+      chainId: args.chainId,
+      safe: args.safe,
+      ordering: '-modified',
+      limit: args.limit,
+      offset: args.offset,
+    });
```
iamacook, line 197: What do you think about spreading `args` in these?

src/domain/safe/safe.repository.ts
```diff
@@ -34,382 +34,382 @@ export class SafeRepository implements ISafeRepository {
...
+    limit?: number;
+    offset?: number;
+  }): Promise<Page<Transaction>> {
+    return this.getAllExecutedTransactions({
+      chainId: args.chainId,
+      safeAddress: args.safeAddress,
+      ordering: 'executionDate',
+      limit: args.limit,
+      offset: args.offset,
+    });
```
iamacook, line 246: We could also spread here.

src/domain/safe/safe.repository.ts
```diff
@@ -34,382 +34,382 @@ export class SafeRepository implements ISafeRepository {
...
+    safeAddress: string;
+    limit?: number;
+    offset?: number;
+  }): Promise<Page<Transaction>> {
+    return this.getAllExecutedTransactions({
+      chainId: args.chainId,
+      safeAddress: args.safeAddress,
+      limit: args.limit,
+      offset: args.offset,
+    });
```
iamacook, line 260:
```suggestion
    return this.getAllExecutedTransactions(args);
```

src/routes/balances/balances.service.ts
```diff
@@ -30,12 +30,12 @@ export class BalancesService {
     trusted?: boolean,
     excludeSpam?: boolean,
   ): Promise<Balances> {
```
iamacook, line 32: We should also change the arguments of `getBalances` (and other functions in this service).

src/routes/transactions/transactions.service.ts
```diff
@@ -295,13 +313,16 @@ export class TransactionsService {
     paginationData: PaginationData,
   ): Promise<Page<QueuedItem>> {
```
iamacook, line 314: This would also benefit similar changes (as well as other functions in this service).

src/datasources/transaction-api/transaction-api.service.ts
fmrsabino: I've considered not doing this for an object with 2 or 3 arguments. One reason why I decided to push this for these functions is mostly because of what I wrote in the PR description (bullet point 3).

One of the current challenges that we have is that we do not have strong types on some arguments (most of them are `string` and can be used interchangeably).

With this change we would need to specify the name for each argument. I have no strong opinion on this one 🤔 . I'd much prefer something like:

```typescript
foo(a: A, b: B, c: C)
```

src/datasources/transaction-api/transaction-api.service.ts
fmrsabino: > You can use `data` directly here instead of create a new object.

You meant `args` here right?

fmrsabino: On a second thought I have some concerns regarding using `args` directly. What if there is a new arg that is used for other purposes (e.g. a cache key)? If we use `args` directly we would send data to the Safe Transaction Service that might not be required (and might provide data that is internal to the CGW instance). Wdyt?

iamacook: > You meant args here right?

Sorry, typo.

> What if there is a new arg that is used for other purposes (e.g. a cache key)? If we use `args` directly we would send data to the Safe Transaction Service that might not be required (and might provide data that is internal to the CGW instance). Wdyt?

If we were to later add new arguments, it would then be tested. However, it is also more explicit as is. I have no strong opinion.

src/datasources/transaction-api/transaction-api.service.ts
fmrsabino: I've tried to keep the instances where even with 2 or more arguments the types are different.

Examples: 

```typescript
getSafeAppById(chainId: string, id: number): Promise<SafeApp | null>;
```

```typescript
registerDevice(device: Device, safeRegistration: SafeRegistration): Promise<SafeRegistration>;
```

What do you think about these examples?

hectorgomezv: In my opinion, it's OK to use interfaces even for 2-3 argument functions on the repositories/datasources layer as the PR does. These are not directly covered by unit tests, so make it harder to introduce a bug.

Going through the path of implementing something like:
```
foo(a: A, b: B, c: C)
```
If I understand it correctly, we'd need to wrap primitives into something like _DDD value objects_, or even group arguments together into more complex structs that we could reuse. Not sure this is what are you proposing exactly, but depending on how we use it, I think it might be a bit overkill for the current purpose/nature of the project.

src/datasources/transaction-api/transaction-api.service.ts
fmrsabino: > > You meant args here right?
> 
> Sorry, typo.
> 
> > What if there is a new arg that is used for other purposes (e.g. a cache key)? If we use `args` directly we would send data to the Safe Transaction Service that might not be required (and might provide data that is internal to the CGW instance). Wdyt?
> 
> If we were to later add new arguments, it would then be tested. However, it is also more explicit as is. I have no strong opinion.

I'd say that we can use the spread operator or the args bundle throughout the internal functions of the CGW but when interacting with external services, to be more explicit on what we are sending (to avoid potential leaks).

src/routes/transactions/transactions.service.ts
fmrsabino: [9db8374](https://github.com/safe-global/safe-client-gateway-nest/pull/610/commits/9db8374f600be9d8fbe58d5c91761aef0ea8b5cf)

src/routes/balances/balances.service.ts
fmrsabino: [9db8374](https://github.com/safe-global/safe-client-gateway-nest/pull/610/commits/9db8374f600be9d8fbe58d5c91761aef0ea8b5cf)

src/domain/safe/safe.repository.ts
fmrsabino: [9db8374](https://github.com/safe-global/safe-client-gateway-nest/pull/610/commits/9db8374f600be9d8fbe58d5c91761aef0ea8b5cf)

src/domain/safe/safe.repository.ts
fmrsabino: [9db8374](https://github.com/safe-global/safe-client-gateway-nest/pull/610/commits/9db8374f600be9d8fbe58d5c91761aef0ea8b5cf)

src/datasources/transaction-api/transaction-api.service.ts
fmrsabino: [9db8374](https://github.com/safe-global/safe-client-gateway-nest/pull/610/commits/9db8374f600be9d8fbe58d5c91761aef0ea8b5cf)

iamacook (approved):

src/datasources/transaction-api/transaction-api.service.ts
hectorgomezv: > I'd say that we can use the spread operator or the args bundle throughout the internal functions of the CGW but when interacting with external services, to be more explicit on what we are sending (to avoid potential leaks).

Yeah, I totally support this 💯

hectorgomezv (approved):
Awesome work! 🚀

PR #611, opened 2023-08-01
Remove Promise.then() mapping calls

fmrsabino:
Removes `Promise.then()` calls that were used in order to map the result of a promise to `void`. Instead, we await for the promise and return from the function (this is more consistent with the way that we handle promises in other functions).

### DIFF
--- a/src/datasources/config-api/config-api.service.ts
+++ b/src/datasources/config-api/config-api.service.ts
@@ -60,7 +60,6 @@ export class ConfigApi implements IConfigApi {
       this.cacheService.deleteByKey(CacheRouter.getChainsCacheKey()),
       this.cacheService.deleteByKeyPattern(pattern),
     ]);
-    return;
   }
 
   async getChain(chainId: string): Promise<Chain> {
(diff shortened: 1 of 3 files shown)


src/datasources/config-api/config-api.service.ts
```diff
@@ -51,15 +51,13 @@ export class ConfigApi implements IConfigApi {
...
-
-    return Promise.all([
+    await Promise.all([
       this.cacheService.deleteByKey(CacheRouter.getChainsCacheKey()),
       this.cacheService.deleteByKeyPattern(pattern),
-    ]).then(() => {
-      return;
-    });
+    ]);
+    return;
```
iamacook, line 60: I don't think we need to `return` anymore.
```suggestion
```

There are too many instances of this to mark, so I'll just comment here.

fmrsabino: Ups you are right! 🤦 Automatic suggestion by the IDE which also removed the `.then()`

fmrsabino: [5ea015b](https://github.com/safe-global/safe-client-gateway-nest/pull/611/commits/5ea015b998adf08743fbadb397c815e8f74b6a3c)

iamacook (approved):

fmrsabino:
@iamacook had to rebase due to the recent merge on `main` 🙇

hectorgomezv (approved):

iamacook (approved):

PR #614, opened 2023-08-02
fix: prevent scientific notation in large balances

### LINKED ISSUE #613: fiatTotal has exponential syntax in JSON
The `/balances` endpoint returns JSON with values in the exponential format:

```
{
    "fiatTotal": "6.269762685290865e+43",
    "items": [...]
}
```

E.g. https://safe-client.safe.global/v1/chains/137/safes/0x828d55AaF0fE916879C7dCf0E3278D1E8dbEDDd3/balances/usd?trusted=true

This causes parsing errors in the web app that crash the whole site.

iamacook:
Resolves #613

This prevents large balances from being returned in scientific notation.

An [example Safe on the Transaction Service](https://safe-transaction-polygon.safe.global/api/v1/safes/0x828d55AaF0fE916879C7dCf0E3278D1E8dbEDDd3/balances/usd/) should no longer return the `fiatTotal` or individual `fiatBalance`/`fiatConversion` with exponents, but the full number as a string: `/v1/chains/137/safes/0x828d55AaF0fE916879C7dCf0E3278D1E8dbEDDd3/balances/usd`.

### DIFF
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -221,6 +221,84 @@ describe('Balances Controller (Unit)', () => {
         });
     });
 
+    it('returns large numbers as is (not in scientific notation)', async () => {
+      const chainId = '1';
+      const safeAddress = '0x0000000000000000000000000000000000000001';
+      const transactionApiBalancesResponse = [
+        balanceBuilder()
+          .with('balance', '266279793958307969327868')
+          // The Transaction Service can return scientific notation for large numbers
+          .with('fiatBalance', '6.25164198388829e+43')
+          .with('fiatConversion', '2.347771827128244e+38')
+          .build(),
+      ];
+      const exchangeApiResponse = exchangeRatesBuilder()
+        .with('success', true)
+        .with('rates', { USD: 1.0 })
+        .build();
+      const chainResponse = chainBuilder().with('chainId', chainId).build();
+      networkService.get.mockImplementation((url) => {
+        if (url == `${safeConfigUrl}/api/v1/chains/${chainId}`) {
+          return Promise.resolve({ data: chainResponse });
+        } else if (
+          url ==
+          `${chainResponse.transactionService}/api/v1/safes/${safeAddress}/balances/usd/`
+        ) {
+          return Promise.resolve({
+            data: transactionApiBalancesResponse,
+          });
+        } else if (
+          url == `${exchangeUrl}/latest?access_key=${exchangeApiKey}`
+        ) {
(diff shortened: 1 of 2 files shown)


katspaugh (commented):
Aren't we parsing these numbers to floats on the client-side? What's the point formatting them to a localizes string here?

src/routes/balances/balances.service.ts
```diff
@@ -23,6 +23,13 @@ export class BalancesService {
     private readonly exchangeRepository: IExchangeRepository,
   ) {}
 
+  getNumberString(value: number): string {
+    // Prevent scientific notation
+    return value.toLocaleString('fullwide', {
```
katspaugh, line 28: IMHO the wrong tool for the job...

katspaugh (approved):
After talking with Aaron:
* The number that comes from tx-service is a huge number that cannot be natively stringified by JSON to anything other than the exponential notation
* CGW then multiplies it which results in another huge number
* It then needs to send it to the client as a string (because apparently it cannot be safely represented as a number)

In this regard, I suggest we accept this as a quick fix but I would suggest later reverting it and dealing with the scientific notation on the client side instead, using proper big number APIs.

src/routes/balances/balances.controller.spec.ts
```diff
@@ -221,6 +221,84 @@ describe('Balances Controller (Unit)', () => {
         });
     });
 
+    it('returns large numbers as is (not in scientific notation)', async () => {
+      const chainId = '1';
+      const safeAddress = '0x0000000000000000000000000000000000000001';
```
fmrsabino, line 226: nitpick: we can use `faker` here.

src/routes/balances/balances.controller.spec.ts
```diff
@@ -221,6 +221,84 @@ describe('Balances Controller (Unit)', () => {
...
+      const transactionApiBalancesResponse = [
+        balanceBuilder()
+          .with('balance', '266279793958307969327868')
+          // The Transaction Service can return scientific notation for large numbers
+          .with('fiatBalance', '6.25164198388829e+43')
+          .with('fiatConversion', '2.347771827128244e+38')
```
fmrsabino, line 232: Would be interesting to know which options faker provides here. We can generate numbers with a minimum value (potentially also with scientific notation).

fmrsabino (approved):

src/routes/balances/balances.controller.spec.ts
iamacook: As agree, let's tackle this after the release of the hotifx.

PR #626, opened 2023-08-08
Fix: pass Safe Apps origin as url, not clientUrl

katspaugh:
Tx-service sends an origin for each Safe App transaction, which contains the URL of the Safe App that created a tx. This URL needs to be passed to the config-service in the URL field, not clientUrl.

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -9,6 +9,9 @@ async function bootstrap() {
     app.get<IConfigurationService>(IConfigurationService);
   const applicationPort: string =
     configurationService.getOrThrow('applicationPort');
+
+  app.enableCors();
+
   await app.listen(applicationPort);
 }
 
--- a/src/main.ts
+++ b/src/main.ts
@@ -10,8 +10,6 @@ async function bootstrap() {
   const applicationPort: string =
     configurationService.getOrThrow('applicationPort');
 
-  app.enableCors();
-
   await app.listen(applicationPort);
 }
 
--- a/src/routes/transactions/mappers/common/safe-app-info.mapper.ts
+++ b/src/routes/transactions/mappers/common/safe-app-info.mapper.ts
@@ -29,7 +29,7 @@ export class SafeAppInfoMapper {
 
     const [safeApp] = await this.safeAppsRepository.getSafeApps({
       chainId,
-      clientUrl: originUrl,
+      url: originUrl,
     });
     if (!safeApp) {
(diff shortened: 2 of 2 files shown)


src/main.ts
```diff
@@ -9,6 +9,9 @@ async function bootstrap() {
     app.get<IConfigurationService>(IConfigurationService);
   const applicationPort: string =
     configurationService.getOrThrow('applicationPort');
+
+  app.enableCors();
```
iamacook, line 13: Why is this change necessary? I'd suggest we move it to a separate PR as it is unrelated.

src/routes/transactions/mappers/common/safe-app-info.mapper.ts
```diff
@@ -29,7 +29,7 @@ export class SafeAppInfoMapper {
 
     const [safeApp] = await this.safeAppsRepository.getSafeApps({
       chainId,
-      clientUrl: originUrl,
+      url: originUrl,
```
iamacook, line 32: Let's add test coverage for this.

katspaugh:
I'm a little confused because it logs
```
{"level":"info","message":"No Safe Apps matching the origin url https://apps-portal.safe.global/safe-claiming-app (safeTxHash: 0xdf0e307ef9013523629e9f3922b70753e1e247834d628c857ddfb18ea92b3c5b)","request_id":"c95a6cff-3cbf-46c5-9e26-66b025e5b518","timestamp":"2023-08-08T07:23:15.965Z"}
```
but still displays the correct Safe App icon and name for the transaction.

src/main.ts
katspaugh, line 13: This is unrelated but it's needed when testing the service in combination with a locally running frontend.

iamacook: I'd suggest we make this change in a separate PR.

src/routes/transactions/mappers/common/safe-app-info.mapper.ts
katspaugh: There's already a test for this, but it's bogus as it was returning false positives. I don't have time right now to fix it. Feel free to make a separate PR. This fix needs to go live ASAP.

iamacook: Understood. I'll create a separate PR later on.

iamacook (approved):

PR #631, opened 2023-08-09
Add human readable descriptions

usame-algan:
## What it solves

- Includes a human readable description with the Transaction info if it exists for a given function call.

## ToDos

- [x] Write tests for `parseMessages`
- [x] Write tests for `human-description.mapper`
- [x] Extend tests for `data-decode`
- [x] Add feature flag
- [x] Fix type casting
- [x] Extend all tests where we return `humanDescription: null`

### DIFF
--- a/src/routes/transactions/mappers/common/human-description.mapper.ts
+++ b/src/routes/transactions/mappers/common/human-description.mapper.ts
@@ -40,11 +40,8 @@ export class HumanDescriptionMapper {
       return null;
     }
 
-    const dataStart = transaction.data.slice(
-      0,
-      HumanDescriptionMapper.SIG_HASH_INDEX,
-    );
-    const sigHash = isHex(dataStart) ? dataStart : null;
+    const sigHash = this.getSigHash(transaction.data);
+
     if (!sigHash) return null;
 
     let token: Token | null = null;
@@ -93,7 +90,6 @@ export class HumanDescriptionMapper {
       .map((fragment) => {
         switch (fragment.type) {
           case ValueType.TokenValue:
-            console.log(token);
             if (!token?.decimals) return fragment.value.amount;
 
             // Unlimited approval
@@ -125,5 +121,11 @@ export class HumanDescriptionMapper {
     }
 
     return `${address.slice(0, length + 2)}...${address.slice(-length)}`;
+  }
+
+  private getSigHash(data: Hex): Hex | null {
+    const dataStart = data.slice(0, HumanDescriptionMapper.SIG_HASH_INDEX);
+
+    return isHex(dataStart) ? dataStart : null;
   }
 }


src/routes/transactions/entities/readable-descriptions/parsed-messages.ts
```diff
@@ -0,0 +1,103 @@
...
+      const address = to;
+
+      return {
+        amount,
+        address,
+      };
+    case ValueType.Recipient:
+      return params[valueIndex];
+    default:
+      return to;
```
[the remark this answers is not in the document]
usame-algan: Makes sense!

src/routes/transactions/entities/readable-descriptions/parsed-messages.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    contractMessages[callSignature] = {
+      render: (to: string, params: unknown[]) => {
+        const fragments: HumanReadableFragment[] = [];
+
+        /** This matches 2 patterns:
+         *  1. Match double curly braces consisting of 2 groups separated by a space
+         *  2. Match any non-whitespace character i.e. simple words
+         */
+        const regex = /{{(.*?)\s(.*?)}}|(\S+)/g;
```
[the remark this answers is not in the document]
usame-algan: I've moved it outside the function but still in the same file and also gave it a more descriptive name.

src/routes/transactions/entities/readable-descriptions/parsed-messages.ts
```diff
@@ -0,0 +1,103 @@
...
+          if (fullMatch && !valueType && !valueIndex) {
+            fragments.push({
+              type: ValueType.Word,
+              value: fullMatch,
+            });
+            continue;
+          }
+
+          const parsedExpression = parseExpression(
+            valueType as ValueType,
```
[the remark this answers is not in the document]
usame-algan: Done!

src/main.ts
```diff
@@ -5,6 +5,8 @@ import { DefaultAppProvider } from './app.provider';
 async function bootstrap() {
   const app = await new DefaultAppProvider().provide(AppModule);
 
+  app.enableCors();
```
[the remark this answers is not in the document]
usame-algan: As far as I understand we only need this to test locally so I would remove it before merging the PR. Will add a comment accordingly.

src/routes/transactions/entities/readable-descriptions/parsed-messages.ts
```diff
@@ -0,0 +1,103 @@
...
+
+        /** This matches 2 patterns:
+         *  1. Match double curly braces consisting of 2 groups separated by a space
+         *  2. Match any non-whitespace character i.e. simple words
+         */
+        const regex = /{{(.*?)\s(.*?)}}|(\S+)/g;
+
+        let match: RegExpExecArray | null;
+
+        while ((match = regex.exec(template)) !== null) {
```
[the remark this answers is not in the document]
usame-algan: Done!

src/routes/transactions/mappers/common/readable-descriptions.mapper.ts
```diff
@@ -0,0 +1,125 @@
...
+
+      const isHumanReadable = data.startsWith(sigHash);
+
+      if (isHumanReadable) {
+        const token = await this.tokenRepository
+          .getToken({ chainId, address: to })
+          .catch(() => null);
+
+        const abi = parseAbi([callSignature]);
+        const { args } = decodeFunctionData({ abi, data: data as Hex });
```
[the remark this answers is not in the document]
usame-algan: Thanks, I've added the check further up

src/routes/transactions/entities/readable-descriptions/parsed-messages.ts
```diff
@@ -0,0 +1,117 @@
...
+};
+
+const isValueType = (type: unknown): type is ValueType => {
+  return (
+    type === ValueType.TokenValue ||
+    type === ValueType.Word ||
+    type === ValueType.Recipient ||
+    type === ValueType.Decimals ||
+    type === ValueType.Identifier
+  );
```
iamacook, line 19:
```suggestion
  return Object.values(ValueType).includes(type);
```

src/routes/transactions/entities/readable-descriptions/erc20-descriptions.json
```diff
@@ -0,0 +1,5 @@
+{
+  "function transfer(address, uint256)": "Send {{tokenValue 1}} to {{address 0}}",
+  "function approve(address, uint256)": "Approve {{tokenValue 1}} to {{address 0}}",
+  "function withdraw(uint256)": "Withdraw {{tokenValue 0}}"
+}
```
katspaugh, line 5: I would suggest to keep all of these templates in one file.

src/routes/transactions/entities/readable-descriptions/parsed-messages.ts
```diff
@@ -0,0 +1,111 @@
...
+  for (const callSignature in messages) {
+    const template = messages[callSignature];
+
+    contractMessages[callSignature] = {
+      render: (to: string, params: unknown[]) => {
+        const fragments: HumanReadableFragment[] = [];
+
+        let match;
+
+        while ((match = TEMPLATE_REGEX.exec(template)) !== null) {
```
katspaugh, line 41: You can use `String.replace(regex, (s, s1, s1) => { ... }` instead of a loop.

src/routes/transactions/mappers/common/readable-descriptions.mapper.ts
```diff
@@ -0,0 +1,130 @@
...
+          .catch(() => null);
+
+        const abi = parseAbi([callSignature]);
+        try {
+          const { args } = decodeFunctionData({ abi, data });
+          const messageBlocks = message.render(to, args);
+
+          return this.createMessage(messageBlocks, token);
+        } catch (e) {
+          console.log(e);
```
katspaugh, line 97: Isn't there a more sophisticated error handler in this project?

src/routes/transactions/entities/readable-descriptions/parsed-messages.ts
usame-algan: String.replace would return a new string but we want to build an array of object depending on what is inside that string.

src/routes/transactions/entities/readable-descriptions/erc20-descriptions.json
usame-algan: Done, I've moved all of them into a single json.

src/datasources/human-description-api/json/contract-descriptions.json
```diff
@@ -0,0 +1,13 @@
...
+  "function approve(address, uint256)": "Approve {{tokenValue 1}} to {{address 0}}",
+  "function withdraw(uint256)": "Withdraw {{tokenValue 0}}",
+  "function changeThreshold(uint256)": "Change threshold to {{decimals 0}}",
+  "function addOwnerWithThreshold(address,uint256)": "Add new owner {{address 0}} with threshold {{decimals 1}}",
+  "function removeOwner(address,address,uint256)": "Remove owner {{address 1}} with threshold {{decimals 2}}",
+  "function swapOwner(address,address,address)": "Swap owner {{address 1}} with {{address 2}}",
+  "function setGuard(address)": "Set guard {{address 0}}",
+  "function enableModule(address)": "Enable module {{address 0}}",
+  "function setFallbackHandler(address)": "Set fallback handler {{address 0}}",
+  "function approveHash(bytes32)": "Approve hash {{identifier 0}}"
```
katspaugh, line 12: As I already mentioned before, `0` and `1` don't scream "positional arguments" to me. They look like arguments to the preceding symbol. E.g. `decimals 0` reads like `format number with zero decimal points`.
`$0` and `$1` would be a more intuitive notation.

usame-algan: Thats a good point. I will adjust it accordingly.

src/datasources/human-description-api/json/contract-descriptions.json
```diff
@@ -0,0 +1,15 @@
...
+  "function changeThreshold(uint256)": "Change threshold to {{decimals $0}}",
+  "function addOwnerWithThreshold(address,uint256)": "Add new owner {{address $0}} with threshold {{decimals $1}}",
+  "function removeOwner(address,address,uint256)": "Remove owner {{address $1}} with threshold {{decimals $2}}",
+  "function swapOwner(address,address,address)": "Swap owner {{address $1}} with {{address $2}}",
+  "function setGuard(address)": "Set guard {{address $0}}",
+  "function enableModule(address)": "Enable module {{address $0}}",
+  "function setFallbackHandler(address)": "Set fallback handler {{address $0}}",
+  "function setDelegate(bytes32, address)": "Set delegate {{address $1}}",
+  "function multiSend(bytes)": "Interaction with MultiSend",
+  "function setAllowance(address,address,uint96,uint16,uint32)": "Set allowance for {{address $0}}"
```
iamacook, line 14: Do we not want to support ERC-721?

It's not clear which of these are ERC-20/Safe. I understand that it was requested to merge all methods into one JSON, but perhaps creating a pre-build script to do so is a cleaner approach. What do you think?

src/main.ts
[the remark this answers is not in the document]
iamacook: Let's not forget this.

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+    const toAddress = new AddressInfo(faker.finance.ethereumAddress());
+    const chainId = faker.string.numeric();
+    const token = tokenBuilder()
+      .with('decimals', 18)
+      .with('name', 'Test Token')
+      .with('symbol', 'TST')
+      .build();
+    tokenRepository.getToken.mockResolvedValue(token);
+    const data =
+      '0xa9059cbb0000000000000000000000007a9af6ef9197041a5841e84cb27873bebd3486e2000000000000000000000000000000000000000000000001236efcbcbb340000';
```
iamacook, line 77: Let's dynamically generate this from the ABI to make the tests more robust. [Here's how we do it on the relay service](https://github.com/safe-global/safe-gelato-relay-service/blob/main/src/__mocks__/transaction-calldata.mock.ts#L190-L199).

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,110 @@
...
+  it('should return undefined for corrupt data', async () => {
+    const toAddress = new AddressInfo(faker.finance.ethereumAddress());
+    const chainId = faker.string.numeric();
+    const token = tokenBuilder()
+      .with('decimals', 18)
+      .with('name', 'Test Token')
+      .with('symbol', 'TST')
+      .build();
+    tokenRepository.getToken.mockResolvedValue(token);
+    const data =
```
iamacook, line 98: As above.

src/routes/transactions/mappers/common/human-descriptions.mapper.ts
```diff
@@ -0,0 +1,106 @@
...
+} from '../../../../logging/logging.interface';
+import { shortenAddress } from '../../../common/utils/utils';
+import { SafeAppInfo } from '../../../transactions/entities/safe-app-info.entity';
+import { IHumanDescriptionApi } from '../../../../domain/interfaces/human-description-api.interface';
+import {
+  HumanReadableFragment,
+  ValueType,
+} from '../../../../datasources/human-description-api/entities/human-description.entity';
+
+// TODO: Write tests for this mapper
```
iamacook, line 25: 👀

src/routes/transactions/mappers/common/human-descriptions.mapper.ts
```diff
@@ -0,0 +1,106 @@
...
+        const abi = parseAbi([callSignature]);
+
+        try {
+          const { args } = decodeFunctionData({ abi, data });
+          const messageFragments = template.process(to, args);
+
+          const message = this.createMessage(messageFragments, token);
+
+          return safeAppInfo
+            ? message.concat(' via ', safeAppInfo.name)
```
iamacook, line 66: Nit: we aren't using `concat` elsewhere

```suggestion
            ? `${message} via ${safeAppInfo.name}`
```

src/routes/transactions/mappers/common/human-descriptions.mapper.ts
```diff
@@ -0,0 +1,106 @@
...
+
+    for (const [callSignature, template] of parsedMessages) {
+      const sigHash = getFunctionSelector(callSignature);
+
+      const isHumanReadable = data.startsWith(sigHash);
+
+      if (isHumanReadable) {
+        const token = await this.tokenRepository
+          .getToken({ chainId, address: to })
+          .catch(() => null);
```
iamacook, line 55: We should add a test case for this.

src/datasources/human-description-api/human-description-api.service.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+      'Send {{tokenValue 1}} to {{address 0}}',
+  };
+
+  it('should return parsed messages from json', () => {
+    const humanDescriptionApi = new HumanDescriptionApi(contractDescriptions);
+
+    const parsedMessages = humanDescriptionApi.getParsedMessages();
+
+    expect(parsedMessages['function transfer(address, uint256)']).toEqual({
+      process: expect.anything(),
```
iamacook, line 15: Let's make this more robust.

```suggestion
      process: expect.any(Function),
```

iamacook (commented):
Looks great! I think we should also extend the existing tests to cover the added logic as well.

src/datasources/human-description-api/json/contract-descriptions.json
usame-algan: I would also imagine if this was to be extracted into an external repository it would be nicer to have those messages split by protocol. Wdyt @katspaugh ?

katspaugh: Perhaps a JS module like this:

```js
export default {
    // ERC20
   "function transfer(address, uint256)": "Send {{tokenValue $1}} to {{address $0}}",
   "function approve(address, uint256)": "Approve {{tokenValue $1}} to {{address $0}}",
   
   // Safe
   "function changeThreshold(uint256)": "Change threshold to {{decimals $0}}",
   ...
}
```

katspaugh: Btw, why is threshold a "decimals"?

src/routes/transactions/mappers/common/human-descriptions.mapper.ts
usame-algan: After adding the test I noticed we never catch a throw here so I wrapped it in a try catch instead. We have this pattern in 3 other places but I am not sure why it is not catching it. Maybe it has to do with how we mock the function call i.e. `tokenRepository.getToken.mockImplementation(() => {throw Error()})`.

src/datasources/human-description-api/json/contract-descriptions.json
usame-algan: Its the type we defined for any number that is not a tokenValue. We could also call it `number`. Wdyt?

katspaugh: In this case it would be specifically an integer since it's not divided by any "decimals".

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+  .with('name', 'Test Token')
+  .with('symbol', 'TST')
+  .build();
+const abi = parseAbi(['function transfer(address, uint256)']);
+const mockTransferData = encodeFunctionData({
+  abi,
+  functionName: 'transfer',
+  args: [
+    '0x7a9af6Ef9197041A5841e84cB27873bEBd3486E2',
+    BigInt('21000000000000000000'),
```
iamacook, line 39: Let's generate a random value for this as well.

```suggestion
    `faker.number.bigInt()`,
```

iamacook (commented):
There are still some unresolved comments above.

src/datasources/human-description-api/human-description-api.service.ts
```diff
@@ -0,0 +1,120 @@
...
+   *  1. Double curly braces consisting of 2 groups separated by a space
+   *  2. Any non-whitespace character i.e. simple words
+   */
+  private readonly templateRegex = /{{(.*?)\s(\$.*?)}}|(\S+)/g;
+
+  constructor(
+    @Inject('ContractDescriptions')
+    private readonly contractDescriptions: Expression,
+  ) {
+    this.parsedMessages = this.parseMessages(this.contractDescriptions);
```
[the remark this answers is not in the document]
hectorgomezv: Not sure I understand the question, do you mean creating an additional class to handle the parsing? It could be nice, but no strong opinion on that from my side.

src/datasources/human-description-api/human-description-api.service.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  getParsedMessages(): MessageTemplates {
+    return this.parsedMessages;
+  }
+
+  parseMessages(messages: Expression): MessageTemplates {
```
hectorgomezv, line 36: Is there a way to reduce the cyclomatic complexity of this function (break it in smaller functions, etc.) to make it more readable? It contains several loops and nested conditionals, it took me a while to think whether the `while` loop really has confiable exit point and it won't let to infinite loops no matter the input 🙂

src/config/entities/__tests__/configuration.ts
```diff
@@ -39,4 +39,7 @@ export default (): ReturnType<typeof configuration> => ({
   safeTransaction: {
     useVpcUrl: false,
   },
+  features: {
+    humanDescription: true,
```
hectorgomezv, line 43: I'd let this as false by default and enable it on-demand in each related test, wdyt?

hectorgomezv (changes requested):
Congrats on the hard work! 👏🏻 I've just left some change requests but they are mostly nitpicks.

src/routes/common/utils/utils.ts
```diff
@@ -1,3 +1,11 @@
 export function isHex(value: unknown): boolean {
   return typeof value === 'string' && value.startsWith('0x');
 }
+
+export const shortenAddress = (address: string, length = 4): string => {
+  if (!address) {
+    return '';
+  }
+
+  return `${address.slice(0, length + 2)}...${address.slice(-length)}`;
```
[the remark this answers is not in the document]
usame-algan: Makes sense, thanks!

src/routes/data-decode/entities/data-decoded.entity.ts
```diff
@@ -6,9 +6,16 @@ export class DataDecoded {
   method: string;
   @ApiPropertyOptional({ type: [DataDecodedParameter], nullable: true })
   parameters: DataDecodedParameter[] | null;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  humanDescription?: string;
```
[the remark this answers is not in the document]
usame-algan: I've had it as null before but had to add the property to a lot of mappers, entities and tests. If we don't mind a larger diff from this PR I can change it back to null.

src/config/entities/configuration.ts
```diff
@@ -53,4 +53,7 @@ export default () => ({
   safeTransaction: {
     useVpcUrl: process.env.USE_TX_SERVICE_VPC_URL?.toLowerCase() === 'true',
   },
+  features: {
+    humanDescription: process.env.FF_HUMAN_DESCRIPTION === 'true',
```
fmrsabino, line 57: To keep it consistent, can we use `?.toLowerCase()` here also?

src/datasources/human-description-api/json/contract-descriptions.json
```diff
@@ -0,0 +1,14 @@
...
+  "function withdraw(uint256)": "Withdraw {{tokenValue $0}}",
+  "function changeThreshold(uint256)": "Change threshold to {{decimals $0}}",
+  "function addOwnerWithThreshold(address,uint256)": "Add new owner {{address $0}} with threshold {{decimals $1}}",
+  "function removeOwner(address,address,uint256)": "Remove owner {{address $1}} with threshold {{decimals $2}}",
+  "function swapOwner(address,address,address)": "Swap owner {{address $1}} with {{address $2}}",
+  "function setGuard(address)": "Set guard {{address $0}}",
+  "function enableModule(address)": "Enable module {{address $0}}",
+  "function setFallbackHandler(address)": "Set fallback handler {{address $0}}",
+  "function setDelegate(bytes32, address)": "Set delegate {{address $1}}",
+  "function setAllowance(address,address,uint96,uint16,uint32)": "Set allowance for {{address $0}}"
```
fmrsabino, line 13: Why do we have a space between the arguments for `setDelegate` but not for `setAllowance`?

src/domain/interfaces/human-description-api.interface.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  Expression,
+  HumanReadableFragment,
+  MessageTemplates,
+  ValueType,
+} from '../../datasources/human-description-api/entities/human-description.entity';
+
+export const IHumanDescriptionApi = Symbol('IHumanDescriptionApi');
+
+export interface IHumanDescriptionApi {
```
fmrsabino, line 10: This interface is mostly for the data layer to know which methods it needs to implement. It makes it easier to replace the json implementation that you added with a different one.

I would still create something like a `Repository` which handles business logic regarding the human descriptions – this would be the `domain` class and the one to be used on the `routes` layer (instead of this interface directly).

fmrsabino (changes requested):
Providing some early feedback (still reviewing the PR)

src/routes/common/utils/utils.ts
```diff
@@ -1,3 +1,17 @@
 export function isHex(value: unknown): boolean {
   return typeof value === 'string' && value.startsWith('0x');
 }
+
+export const shortenAddress = (address: string, length = 4): string => {
+  if (!address) {
+    return '';
+  }
+
+  const visibleCharactersLength = length + 2 * 2;
```
fmrsabino, line 10: What's the reason behind the `+2` and the `*2`?

src/routes/common/utils/utils.ts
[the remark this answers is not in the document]
fmrsabino: I'd make it more restrictive – addresses in Ethereum are 42 characters long (including the `0x` prefix). I would not accept other lengths in this function (can throw an error).

src/datasources/human-description-api/entities/human-description.entity.ts
```diff
@@ -0,0 +1,54 @@
...
+  | TokenValueFragment
+  | IdentifierFragment
+  | AddressFragment
+  | DecimalsFragment;
+
+type MessageTemplate = {
+  process: (to: string, params: readonly unknown[]) => HumanReadableFragment[];
+};
+
+export type MessageTemplates = Record<string, MessageTemplate>;
```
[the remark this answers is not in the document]
usame-algan: Lets use `HumanDescription` instead i.e. `HumanDescriptionTemplates`, `parseHumanDescriptions` etc.

src/domain/data-decoder/entities/data-decoded.entity.ts
```diff
@@ -8,4 +8,5 @@ export interface DataDecodedParameter {
 export interface DataDecoded {
   method: string;
   parameters: DataDecodedParameter[] | null;
+  humanDescription?: string;
```
fmrsabino, line 11: For this project, we have decided to use `null` to be more explicit on the absence of value.

`undefined` was considered but ultimately for domain classes we wanted to be more explicit that the property has no value (and not that it was not "set").

src/routes/data-decode/entities/data-decoded.entity.ts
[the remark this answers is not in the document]
fmrsabino: We should also not expose the property if the feature is not enabled.

(For the tests, by using the builders we can cover most of the default)

.env.sample
```diff
@@ -35,3 +35,6 @@
 
 # Log level for the service.
 #LOG_LEVEL=
+
+# Enable human description feature
+#HUMAN_DESCRIPTION=
```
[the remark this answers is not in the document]
fmrsabino: Either that or `FEATURE_*`. No strong opinion on this one!

src/datasources/human-description-api/human-description-api.service.ts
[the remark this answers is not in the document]
fmrsabino: I agree @iamacook . I think the parsed messages can be provided via DI in a ready-to-consume manner to `HumanDescriptionApi`. You can use a Nest Factory to provide those.

src/datasources/human-description-api/human-description-api.service.ts
```diff
@@ -0,0 +1,124 @@
...
+@Injectable()
+export class HumanDescriptionApi implements IHumanDescriptionApi {
+  private readonly parsedMessages: MessageTemplates;
+
+  /**
+   *  Regex Template that matches two patterns
+   *  1. Double curly braces consisting of 2 groups separated by a space
+   *  2. Any non-whitespace character i.e. simple words
+   */
+  private readonly TEMPLATE_REGEX = /{{(.*?)\s(\$.*?)}}|(\S+)/g;
```
fmrsabino, line 23: Can be `static` as I don't see this as a property of the instance itself.

src/datasources/human-description-api/entities/human-description.entity.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export type Expression = {
+  [key: string]: string;
+};
```
fmrsabino, line 54: What is the value of this type? We can use `Record<string, string>` directly to accomplish the same constraints that we have for reading JSON.

fmrsabino (changes requested):

src/datasources/human-description-api/json/contract-descriptions.json
usame-algan: Nice catch, I think it doesn't matter for getting the function signature so I will add a space to all of them for better readability.

src/routes/common/utils/utils.ts
usame-algan: The result address is 0x-prefixed so we need to count that in and multiply length by 2 because thats the number of characters visible from the start and end of the address i.e. `0x${length characters}...{length characters}`. It should actually be `length * 2 + 2` to work. Will fix it and also write some tests.

fmrsabino: I'd also suggest adding docs to this function so that it's more clear what it does 🙂

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,108 @@
...
+    safeAppInfo: SafeAppInfo | null,
+  ): Promise<string | undefined> {
+    if (!data || !isHex(data) || !to) return;
+
+    const parsedMessages = Object.entries(
+      this.humanDescriptionApiService.getParsedMessages(),
+    );
+
+    for (const [callSignature, template] of parsedMessages) {
+      const sigHash = getFunctionSelector(callSignature);
```
fmrsabino, line 47: This is called for every transaction that has data which can impact the performance as `getFunctionSelector` relies on hashing and string manipulation.

For readability I understand why we have the functions as a key in JSON format. But we should do some lazy evaluation for each one once we know the hash for each one.

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,108 @@
...
+
+      const isHumanReadable = data.startsWith(sigHash);
+
+      if (isHumanReadable) {
```
fmrsabino, line 51: You can return early instead. Make it more readable in my opinion.

```typescript
if (!isHumanReadable) continue;
```

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,108 @@
...
+        let token: Token | null = null;
+        try {
+          token = await this.tokenRepository.getToken({ chainId, address: to });
+        } catch (error) {
+          this.loggingService.info(
+            `Error trying to get token: ${error.message}`,
+          );
+        }
+
+        const abi = parseAbi([callSignature]);
```
fmrsabino, line 61: We are decoding known ABIs (at build time) every time we map a transaction. This has a performance impact. We should be able to parse the ABIs lazily. This should also not be done by the presentation layer – seems to be more related with the business logic of the service.

fmrsabino (changes requested):

src/datasources/human-description-api/json/contract-descriptions.json
fmrsabino: > Nice catch, I think it doesn't matter for getting the function signature so I will add a space to all of them for better readability.

The computed hash can potentially be different. Depends on the implementation of the library. I'll take a look.

fmrsabino: it seems that the `viem` trims the whitespace between the parameters so we should be good either way 👍 

https://github.com/wagmi-dev/viem/blob/169a695edeb7a0d1b6a9885509c3da2683cfe7ba/src/utils/contract/extractFunctionParts.ts#L17

I'd still keep one style to be consistent (preference for without spaces).

src/routes/data-decode/entities/data-decoded.entity.ts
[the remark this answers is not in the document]
usame-algan: I am not sure how we can only expose the `humanDescription` property from the class if the feature is enabled. Do you have an example of this?

src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
```diff
@@ -157,6 +158,49 @@ describe('Data decode e2e tests', () => {
       });
   });
 
+  it('POST /data-decoder should include humanDescription', async () => {
```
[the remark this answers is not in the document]
usame-algan: I tried adding that without success. Any ideas why it doesn't work?

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -55,18 +63,35 @@ export class MultisigTransactionInfoMapper {
     const dataSize =
       dataByteLength >= 2 ? Math.floor((dataByteLength - 2) / 2) : 0;
 
+    const safeAppInfo = isMultisigTransaction(transaction)
```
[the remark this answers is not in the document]
usame-algan: I agree, it is cleaner to have `safeAppInfo` inside `HumanDescriptionMapper`. Just need to figure out how I can adjust the signature of either `mapSafeAppInfo` or `mapHumanDescription` to be compatible. Right now `mapHumanDescription` expects `to` and `data` but `mapSafeAppInfo` expects a `transaction` object.

src/routes/common/utils/utils.ts
```diff
@@ -1,3 +1,17 @@
 export function isHex(value: unknown): boolean {
   return typeof value === 'string' && value.startsWith('0x');
 }
+
+export const shortenAddress = (address: string, length = 4): string => {
```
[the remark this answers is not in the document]
usame-algan: Ah nice catch! Just a copy-paste issue from web-core. Lets change it to a normal function.

src/routes/common/utils/utils.ts
```diff
@@ -1,3 +1,17 @@
...
+  if (!address) {
+    return '';
+  }
```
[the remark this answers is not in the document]
usame-algan: I think this is an issue with copy-pasting it from web-core. We don't have the type-safety for address there so we had to add this at some point. I will remove it here.

src/routes/data-decode/data-decoded.service.ts
```diff
@@ -3,22 +3,40 @@ import { DataDecodedRepository } from '../../domain/data-decoder/data-decoded.re
...
+      }),
+
+      this.configurationService.get('features.humanDescription')
+        ? this.humanDescriptionMapper.mapHumanDescription(
+            args.getDataDecodedDto.to,
+            args.getDataDecodedDto.data,
+            args.chainId,
+            null,
+          )
+        : undefined,
```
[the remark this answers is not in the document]
usame-algan: I've changed it so that `humanDescription` is null instead of undefined everywhere now.

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,108 @@
...
+    @Inject(IHumanDescriptionApi)
+    private readonly humanDescriptionApiService: IHumanDescriptionApi,
+  ) {}
+
+  async mapHumanDescription(
+    to: string | undefined,
+    data: string | null,
+    chainId: string,
+    safeAppInfo: SafeAppInfo | null,
+  ): Promise<string | undefined> {
```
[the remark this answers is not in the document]
usame-algan: done!

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
usame-algan: done

.env.sample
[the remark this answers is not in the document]
usame-algan: Changed it to `FF_*`

src/datasources/human-description-api/entities/human-description.entity.ts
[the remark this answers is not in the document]
usame-algan: I've pushed an update on this.

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,108 @@
...
+          );
+          return;
+        }
+      }
+    }
+
+    return;
+  }
+
+  createMessage(
```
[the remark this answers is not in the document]
usame-algan: I've changed it to `createHumanDescription`.

src/routes/transactions/entities/custom-transaction.entity.ts
```diff
@@ -23,8 +23,9 @@ export class CustomTransactionInfo extends TransactionInfo {
     methodName: string | null,
     actionCount: number | null,
     isCancellation: boolean,
+    humanDescription?: string,
```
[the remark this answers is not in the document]
usame-algan: done

src/datasources/human-description-api/human-description-api.service.ts
```diff
@@ -0,0 +1,127 @@
...
+  ): HumanReadableFragment | null {
+    try {
+      const parsedParam = this.parseParam(valueType, valueIndex, to, params);
+
+      return <HumanReadableFragment>{
+        type: valueType,
+        value: parsedParam,
+      };
+    } catch (error) {
+      return null;
```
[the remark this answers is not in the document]
usame-algan: Right now all templates are local in json so there shouldn't be any issues but eventually I think we should be notified if there was an issue parsing a template so we could also log it additionally.

src/datasources/human-description-api/human-description-api.service.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  parseParam(
+    valueType: ValueType,
+    valueIndex: number,
+    to: string,
+    params: unknown[],
+  ): HumanReadableFragment['value'] {
+    switch (valueType) {
+      case ValueType.TokenValue:
+        const amount = params[valueIndex];
```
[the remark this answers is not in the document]
usame-algan: Makes sense, we are not using those variables for anything else so I removed them.

src/config/entities/configuration.ts
usame-algan: done

src/datasources/human-description-api/human-description-api.service.ts
usame-algan: done

src/datasources/human-description-api/entities/human-description.entity.ts
usame-algan: I like the `Record` syntax so I've changed it. It should be the same as before just so we don't have `any` when parsing the json.

src/routes/data-decode/__tests__/data-decode.e2e-spec.ts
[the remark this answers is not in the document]
usame-algan: It seems to have worked now maybe it was a caching issue in github.

src/routes/transactions/mappers/common/human-description.mapper.ts
usame-algan: I've moved this into `HumanDescriptionRepository` as discussed so we don't have to compute on every call anymore.

src/datasources/human-description-api/human-description-api.service.ts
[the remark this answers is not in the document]
usame-algan: I've simplified `HumanDescriptionApi` a lot now. It only has one get function and essentially returns the json. Once we move it to an external source we can define more properties on the class like `baseUrl` etc.

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,115 @@
...
+
+    const parsedDescriptions =
+      this.humanDescriptionRepository.getDescriptions();
+
+    const dataStart = data.slice(0, 10);
+    const sigHash = isHex(dataStart) ? dataStart : null;
+
+    if (!sigHash) return null;
+
+    const template = parsedDescriptions[sigHash];
```
usame-algan, line 45: This can be undefined but is currently not caught because we don't have `"noUncheckedIndexedAccess": true` in the tsconfig. Activating that rule breaks a lot of other places too so I suggest to change it in a separate PR.

src/routes/transactions/mappers/common/transaction-info.mapper.ts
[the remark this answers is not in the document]
usame-algan: I've changed the `HumanDescriptionMapper` to accept transactions instead of to and data and moved the safeAppInfo call inside the mapper.

src/datasources/human-description-api/json/contract-descriptions.json
```diff
@@ -0,0 +1,14 @@
+{
+  "function transfer(address, uint256)": "Send {{tokenValue $1}} to {{address $0}}",
+  "function approve(address, uint256)": "Approve {{tokenValue $1}}",
```
iamacook, line 3: Did we make a decision whether we will separate these?

src/domain/human-description/human-description.repository.ts
```diff
@@ -0,0 +1,136 @@
...
+    params: unknown[],
+  ): HumanDescriptionFragment | null {
+    try {
+      const parsedParam = this.parseParam(valueType, valueIndex, to, params);
+
+      return <HumanDescriptionFragment>{
+        type: valueType,
+        value: parsedParam,
+      };
+    } catch (error) {
```
iamacook, line 106: Nit: as `error` is not used, let's remove it:

```suggestion
    } catch {
```

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,134 @@
...
+    chainId: string,
+  ): Promise<string | null> {
+    if (!transaction.data || !isHex(transaction.data) || !transaction.to) {
+      return null;
+    }
+
+    const parsedDescriptions =
+      this.humanDescriptionRepository.getDescriptions();
+
+    const dataStart = transaction.data.slice(0, 10);
```
iamacook, line 44: Nit: let's move the `10` to a constant.

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,134 @@
...
+            return fragment.value;
+        }
+      })
+      .join(' ');
+  }
+
+  private shortenAddress(address: Hex, length = 4): string {
+    if (address.length !== 42) {
+      throw Error('Invalid address');
+    }
```
iamacook, line 124: We can use [`viem`'s `isAddress`](https://viem.sh/docs/utilities/isAddress.html) here.

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,134 @@
...
+  createHumanDescription(
+    descriptionFragments: HumanDescriptionFragment[],
+    token: Token | null,
+  ): string {
+    return descriptionFragments
+      .map((fragment) => {
+        switch (fragment.type) {
+          case ValueType.TokenValue:
+            if (!token?.decimals) {
+              console.log(token);
```
iamacook, line 100: 👀

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  .with('decimals', 18)
+  .with('name', 'Test Token')
+  .with('symbol', 'TST')
+  .build();
+const abi = parseAbi(['function transfer(address, uint256)']);
+const mockAmount = faker.number.bigInt();
+const mockTransferData = encodeFunctionData({
+  abi,
+  functionName: 'transfer',
+  args: ['0x7a9af6Ef9197041A5841e84cB27873bEBd3486E2', mockAmount],
```
iamacook, line 47: Let's use `faker` here for the address as well.

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+const humanDescriptionAPI = new HumanDescriptionApi();
+const humanDescriptionRepository = new HumanDescriptionRepository(
+  humanDescriptionAPI,
+);
+const toAddress = new AddressInfo(faker.finance.ethereumAddress());
+const chainId = faker.string.numeric();
+const token = tokenBuilder()
+  .with('decimals', 18)
+  .with('name', 'Test Token')
+  .with('symbol', 'TST')
```
iamacook, line 40: We don't need to specifiy these for the builder as it generates them for us.

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  .build();
+const abi = parseAbi(['function transfer(address, uint256)']);
+const mockAmount = faker.number.bigInt();
+const mockTransferData = encodeFunctionData({
```
iamacook, line 44: If we generate this _inside_ each test, it will test more cases.

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+  });
+
+  it('should return a description for unlimited token approvals', async () => {
+    tokenRepository.getToken.mockImplementation(() => Promise.resolve(token));
+
+    const mockApprovalData = encodeFunctionData({
+      abi: parseAbi(['function approve(address, uint256)']),
+      functionName: 'approve',
+      args: ['0x7a9af6Ef9197041A5841e84cB27873bEBd3486E2', MAX_UINT256],
+    });
```
iamacook, line 143: We should create a generator for this as well and/or "fake" the address.

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -0,0 +1,180 @@
...
+      chainId,
+    );
+
+    expect(humanDescription).toBe('Approve unlimited TST');
+  });
+
+  it('should append the safe app name to the description if it exists', async () => {
+    tokenRepository.getToken.mockImplementation(() => Promise.resolve(token));
+    const mockSafeAppInfo = new SafeAppInfo(
+      'CSV Airdrop',
```
iamacook, line 160: Let's use `faker` here as well.

src/domain/human-description/human-description.repository.interface.ts
```diff
@@ -0,0 +1,29 @@
...
+    params: unknown[],
+  ): HumanDescriptionFragment | null;
+
+  parseParam(
+    valueType: ValueType,
+    valueIndex: number,
+    to: string,
+    params: unknown[],
+  ): HumanDescriptionFragment['value'];
+}
```
fmrsabino, line 29: The interface is exposing functionality that is internal to how the repository works. I will create a branch with some suggestions.

fmrsabino (changes requested):

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
usame-algan: I still have to limit the generated decimals otherwise `formatUnits` which uses `String.padStart` throws a range error if decimals is too big.

src/datasources/human-description-api/json/contract-descriptions.json
usame-algan: I don't think we did. Since the number of functions we support right now is small I suggest to leave this task for later when we add more protocol support and/or extract them to an external source. cc @katspaugh

katspaugh:
Please try to wrap up the review process, we need to release this by the end of the month.

fmrsabino:
> Please try to wrap up the review process, we need to release this by the end of the month.

The stakeholders are aware of the current state of this change. Wrapping up happens when it's ready to be merged.

src/datasources/human-description-api/json/contract-descriptions.json
usame-algan: I've changed it to `number` and also the type from `unknown` to `bigint` because so far we are always using it on `uint256` which viem decodes as a BigInt. We can revisit this later if more types need to be added.

src/routes/transactions/mappers/common/human-descriptions.mapper.ts
usame-algan: This is done now. I've removed the TODO

src/main.ts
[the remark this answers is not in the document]
usame-algan: I've removed it.

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,128 @@
...
+  ): string {
+    return descriptionFragments
+      .map((fragment) => {
+        switch (fragment.type) {
+          case ValueType.TokenValue:
+            if (!token?.decimals) return fragment.value.amount;
+
+            // Unlimited approval
+            if (fragment.value.amount === MAX_UINT256) {
+              return `unlimited ${token.symbol}`;
```
fmrsabino, line 100: What's the reasoning behind considering `MAX_UINT256` as unlimited?

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,128 @@
...
+        descriptionFragments,
+        token,
+      );
+
+      const safeAppInfo = isMultisigTransaction(transaction)
+        ? await this.safeAppInfoMapper.mapSafeAppInfo(chainId, transaction)
+        : null;
+
+      return safeAppInfo
+        ? `${description} via ${safeAppInfo.name}`
```
fmrsabino, line 78: Something to consider for the future: should this be responsibility of the domain layer or the route?

Right now we are including the Safe App Info as part of the description whenever we have it. One can see this as an extension to every template that we have listed and can probably be formatted in a different way (e.g.: to include the safe app logo).

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -15,12 +15,15 @@ import { NativeCoinTransferMapper } from './native-coin-transfer.mapper';
 import { SettingsChangeMapper } from './settings-change.mapper';
 import { DataDecoded } from '../../../data-decode/entities/data-decoded.entity';
 import { DataDecodedParameter } from '../../../data-decode/entities/data-decoded-parameter.entity';
+import { HumanDescriptionMapper } from '../common/human-description.mapper';
```
fmrsabino, line 18: can be shortened:

```typescript
import { HumanDescriptionMapper } from './human-description.mapper';
```

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -35,13 +38,20 @@ export class MultisigTransactionInfoMapper {
...
     private readonly settingsChangeMapper: SettingsChangeMapper,
     private readonly nativeCoinTransferMapper: NativeCoinTransferMapper,
     private readonly erc20TransferMapper: Erc20TransferMapper,
     private readonly erc721TransferMapper: Erc721TransferMapper,
-  ) {}
+    private readonly humanDescriptionMapper: HumanDescriptionMapper,
+  ) {
+    this.HUMAN_DESCRIPTION_ENABLED = this.configurationService.get(
+      'features.humanDescription',
+    );
```
fmrsabino, line 53: Since this configuration should always exist we can use `getOrThrow`.

Even if the environment variable is not set it would have a value of `false`.

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -15,12 +15,15 @@ import { NativeCoinTransferMapper } from './native-coin-transfer.mapper';
...
 import { DataDecodedParameter } from '../../../data-decode/entities/data-decoded-parameter.entity';
+import { HumanDescriptionMapper } from '../common/human-description.mapper';
+import { IConfigurationService } from '../../../../config/configuration.service.interface';
 
 @Injectable()
 export class MultisigTransactionInfoMapper {
   private readonly TRANSFER_METHOD = 'transfer';
   private readonly TRANSFER_FROM_METHOD = 'transferFrom';
   private readonly SAFE_TRANSFER_FROM_METHOD = 'safeTransferFrom';
+  private readonly HUMAN_DESCRIPTION_ENABLED: boolean | undefined;
```
fmrsabino, line 26: I would rename this to something like:

```typescript
isHumanDescriptionFeatureEnabled
```

or

```typescript
isHumanDescriptionEnabled
```

The reason is that from a style perspective all-caps would mean a global variable known at build time.

Even though we don't strictly follow this, it's better explained here: https://google.github.io/styleguide/tsguide.html#identifiers

src/domain.module.ts
```diff
@@ -55,10 +55,18 @@ import { FlushRepository } from './domain/flush/flush.repository';
...
+import { HumanDescriptionRepository } from './domain/human-description/human-description.repository';
 
 @Global()
 @Module({
-  imports: [ConfigApiModule, ExchangeApiModule, TransactionApiModule],
+  imports: [
+    ConfigApiModule,
+    ExchangeApiModule,
+    TransactionApiModule,
+    HumanDescriptionApiModule,
```
fmrsabino, line 68: nitpick: to keep the alphabetical order here.

src/domain.module.ts
```diff
@@ -76,6 +84,10 @@ import { HealthRepository } from './domain/health/health.repository';
     { provide: ISafeAppsRepository, useClass: SafeAppsRepository },
     { provide: ISafeRepository, useClass: SafeRepository },
     { provide: ITokenRepository, useClass: TokenRepository },
+    {
+      provide: IHumanDescriptionRepository,
+      useClass: HumanDescriptionRepository,
+    },
```
fmrsabino, line 90: nitpick: to keep the alphabetical order here.

src/domain.module.ts
```diff
@@ -115,6 +127,7 @@ import { HealthRepository } from './domain/health/health.repository';
     ISafeAppsRepository,
     ISafeRepository,
     ITokenRepository,
+    IHumanDescriptionRepository,
```
fmrsabino, line 130: nitpick: to keep the alphabetical order here.

package.json
```diff
@@ -43,6 +43,7 @@
     "rimraf": "^5.0.1",
     "rxjs": "^7.8.1",
     "semver": "^7.5.4",
+    "viem": "^1.5.3",
```
fmrsabino, line 46: nitpick: If you want to upgrade `viem` here to `1.7.0` since this is the first PR to introduce it. Else we can do it in a separate PR.

fmrsabino (changes requested):
Nice job! Some smaller comments :)

src/routes/transactions/mappers/common/transaction-info.mapper.ts
fmrsabino: This would avoid having an `undefined` value and additional branches to deal with `undefined`.

src/routes/common/utils/utils.ts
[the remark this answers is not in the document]
fmrsabino: > Should we expect corrupt encoded data at this stage?

This is a great question. `string` was mostly used throughout the project which is not the best as we lose any meaningful validation done to it. We had this discussion internally in the past but one thing that was considered was to use an `Address` type not as a simple alias to `string` but one where you could be sure that if you had an `Address` then it was valid. If we do this from the get go you'd receive here an `Address` type and not a `string` which would remove the validation that you just added.

> I've added the `isAddress` check from viem instead.

I think this is a good solution for now! 🚀

src/routes/data-decode/entities/data-decoded.entity.ts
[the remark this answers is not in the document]
fmrsabino: > I am not sure how we can only expose the `humanDescription` property from the class if the feature is enabled. Do you have an example of this?

This is the first time that the CGW has a feature rollout with NestJS so we don't have other examples. I can explore different alternatives on how this can be made.

src/domain/human-description/human-description.repository.interface.ts
fmrsabino: This has been resolved in [7aa7d2e](https://github.com/safe-global/safe-client-gateway-nest/pull/631/commits/7aa7d2e8d8b58ecebedd7289e7efe159a3acb191)

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -0,0 +1,128 @@
...
+    @Inject(IHumanDescriptionRepository)
+    private readonly humanDescriptionRepository: HumanDescriptionRepository,
+    private readonly safeAppInfoMapper: SafeAppInfoMapper,
+  ) {}
+
+  async mapHumanDescription(
+    transaction: MultisigTransaction | ModuleTransaction,
+    chainId: string,
+  ): Promise<string | null> {
+    if (!transaction.data || !isHex(transaction.data) || !transaction.to) {
```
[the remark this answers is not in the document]
fmrsabino: The private function can still be inside the mapper for now. But I agree that it encapsulates better the expected input/output.

src/domain/human-description/human-description.repository.ts
```diff
@@ -0,0 +1,50 @@
...
+  HumanDescriptionFragment,
+} from './entities/human-description.entity';
+import { IHumanDescriptionApi } from '../interfaces/human-description-api.interface';
+import { getFunctionSelector } from 'viem';
+import { HumanDescriptionTemplate } from './entities/human-description-template.entity';
+
+@Injectable()
+export class HumanDescriptionRepository implements IHumanDescriptionRepository {
+  private readonly templates: Record<
+    FunctionSignature,
```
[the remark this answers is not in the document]
usame-algan: Nice catch! I will change it in the appropriate places.

src/routes/transactions/mappers/common/human-description.mapper.ts
fmrsabino: Thank you for clarifying! 🙌

src/routes/transactions/mappers/common/human-description.mapper.ts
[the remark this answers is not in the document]
usame-algan: Makes sense, I moved it to a private function 👍

src/datasources/human-description-api/human-description-api.service.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { HumanDescriptionApi } from './human-description-api.service';
+
+describe('HumanDescriptionAPI', () => {
+  it('should return descriptions from json', () => {
+    const humanDescriptionApi = new HumanDescriptionApi();
+
+    const descriptions = humanDescriptionApi.getDescriptions();
+
+    expect(descriptions['function transfer(address, uint256)']).toEqual(
```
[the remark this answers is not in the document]
usame-algan: In its current state its not really a useful test since it just checks that the json import is returned by the `getDescriptions` call so we might as well change it to something like:

```
import ContractDescriptions from './json'

const descriptions = humanDescriptionApi.getDescriptions();
expect(descriptions).toEqual(ContractDescriptions);
```

Or I can remove it until we extend the api service. Wdyt?

src/datasources/human-description-api/json/contract-descriptions.json
usame-algan: I will move this to the discussion document for v2

src/datasources/human-description-api/human-description-api.service.spec.ts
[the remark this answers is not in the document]
fmrsabino: Since it's still a valid test I'd consider extending/refactoring it after merging the PR 👍 

The scope is already quite big for this one 😅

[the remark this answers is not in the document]
hectorgomezv: Yeah, no need to tackle on the PR, I was just noticing this to see if we are aligned in the case of a future improvement 👍

src/routes/transactions/mappers/common/human-description.mapper.ts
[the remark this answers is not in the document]
hectorgomezv: Thanks!

hectorgomezv (approved):
Great job!! 🚀

iamacook (approved):

PR #632, opened 2023-08-09
Make safe version nullable

### LINKED ISSUE #629: Make Safe version nullable
**Problem**
_Validation errors_ are not being handled in the frontend. So when the service retrieves a Safe with a `null` version from the Transaction Service, the call to `/chains/:chainId/safes/:safeAddress` fails and the frontend is not capable of managing the error.

Failing when an _incorrect_ Safe entity (with `version: null`) is retrieved is the correct behavior, but it's not the expected one right now.

**Short-term solution**
- Make `version` a nullable field for a `Safe`. The web UI already has some logic to manage `null` Safe versions: https://github.com/safe-global/safe-wallet-web/blob/consistent-provider/src/hooks/coreSDK/safeCoreSDK.ts#L56

**Long-term solution**
- Push the implementation of a _validation error view_ in the clients, so they can manage this kind of scenario.

hectorgomezv:
Closes #629 

This PR:
- Adds support for having `null` as Safes `version` attribute, avoiding returning a _validation error_  which crashes the Web UI (nullable `version` field [was the case in the Rust implementation](https://github.com/safe-global/safe-client-gateway/blob/eb481938a49c29b979fe98a5577f719f42a2b2eb/src/providers/info.rs#L64)).
- Modifies the OpenAPI definition accordingly.
- Adapts the code to return `UNKNOWN` for the `implementationVersionState` field when `version` is `null`, [as the Rust implementation does](https://github.com/safe-global/safe-client-gateway/blob/eb481938a49c29b979fe98a5577f719f42a2b2eb/src/routes/safes/converters.rs).

### DIFF
--- a/src/domain/safe/entities/safe.entity.ts
+++ b/src/domain/safe/entities/safe.entity.ts
@@ -7,5 +7,5 @@ export interface Safe {
   nonce: number;
   owners: string[];
   threshold: number;
-  version: string;
+  version: string | null;
 }
--- a/src/domain/safe/entities/schemas/safe.schema.ts
+++ b/src/domain/safe/entities/schemas/safe.schema.ts
@@ -1,10 +1,9 @@
-import { JSONSchemaType } from 'ajv';
-import { Safe } from '../safe.entity';
+import { Schema } from 'ajv';
 
 export const SAFE_SCHEMA_ID =
   'https://safe-client.safe.global/schemas/safe/safe.json';
 
-export const safeSchema: JSONSchemaType<Safe> = {
+export const safeSchema: Schema = {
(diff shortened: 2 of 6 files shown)


hectorgomezv:
Closing the PR after a second thought and after discussing the purpose with @Uxio0. The outcome is it's OK to fail on the CGW side for `null` versions since _incorrect_ contract deployments (L2/nonL2 misalignment) won't be supported. 

For special cases, the Safe Core team would add the contracts on demand. (cc. @iamacook, as you were involved also in the discussions)

Uxio0:
@hectorgomezv thanks for taking care of this

PR #633, opened 2023-08-09
refactor: Use path alias instead of relative imports

usame-algan:
## What it solves

Uses path aliases instead of relative imports for `@domain`, `@data`, `@routes`.

Note: There are still some (short) relative imports which my IDE didn't catch but they are not causing warnings. However in the future we can restrict relative imports even further by adding an eslint rule see https://stackoverflow.com/a/76095340

## How to test

- All tests should still run


src/datasources/network/network.module.spec.ts
```diff
@@ -1,10 +1,10 @@
 import { Test, TestingModule } from '@nestjs/testing';
 import { NetworkModule } from './network.module';
-import { RequestScopedLoggingModule } from '../../logging/logging.module';
+import { RequestScopedLoggingModule } from '@/logging/logging.module';
 import { ClsModule } from 'nestjs-cls';
-import { ConfigurationModule } from '../../config/configuration.module';
+import { ConfigurationModule } from '@/config/configuration.module';
 import configuration from '../../config/entities/__tests__/configuration';
```
iamacook, line 6: I've noticed a few places where the alias is not being used. Many of which are up and down a folder, but what do you think about beyond that?

usame-algan: I assume there is a rule that allows relative imports up to a certain depth but I haven't found a way to auto replace those with the path alias. If I remove the import and import again it will be imported with a path alias though.

iamacook: I agree but we should align on this. @hectorgomezv, what are your thoughts?

hectorgomezv: I've been researching this a bit because I've never used path aliases before. It seems nice, but I'd be in favor of going for an all-or-none strategy, mixing aliases with relative paths seems a bit ugly and error-prone to me. 

I'd wait for @fmrsabino opinion on this as well since is a wider change that would affect the whole team. I'm in fact trying to make my IDE understand it when dealing with auto-imports with no luck yet. As I said, I'm not against the change at all, but in my opinion, we could put this on hold and wait for the whole team to agree.

usame-algan: Agree, we can split it further for each subdirectory. Regarding IDE support I had to configure both VSCode and Webstorm to work with this.

In Webstorm there is an option `Use paths relative to tsconfig.json` that I had to turn on.
In VSCode I had to change `typescript.preferences.importModuleSpecifier` to `non-relative` in the settings.json.

fmrsabino: You have some other cases like

```typescript
import { SafeAppsRepository } from '../../../../domain/safe-apps/safe-apps.repository';
```

https://github.com/safe-global/safe-client-gateway-nest/blob/e176080a51c2b8e17d6fa9ce48e283abf4c5d1f1/src/routes/transactions/mappers/common/safe-app-info.mapper.ts#L2-L4

For these cases, they'd become

```typescript
import { SafeAppsRepository } from `@domain/safe-apps/safe-apps.repository';
```

One disadvantage that I see with this approach though is that the modules are mapped via a regex in the testing configuration which looks like that it can break easily if we move somethings around 🤔

hectorgomezv: Done: https://github.com/safe-global/safe-client-gateway-nest/pull/643 
If we agree to use this, we would need to modify that file 🙂

usame-algan: > One disadvantage that I see with this approach though is that the modules are mapped via a regex in the testing configuration which looks like that it can break easily if we move somethings around 🤔

The mapping we do in the jest config is very similar to the mapping in the tsconfig. Ideally, it would be one source of truth for both mappings but I haven't found a way to do that.

usame-algan: There is a way to disallow relative imports via eslint ("no-restricted-imports") but they would have to be fixed by hand afaik. We could still add an exception for imports from the same directory so if we decide to go with path aliases I can do that.

usame-algan: I've added path aliases for domain, datasources and routes. As discussed with @fmrsabino, further restricting relative imports is something we can add later on.

fmrsabino: > I've added path aliases for domain, datasources and routes. As discussed with @fmrsabino, further restricting relative imports is something we can add later on.

Awesome! Thanks :) I will have a look into the new changes.

package.json
```diff
@@ -89,7 +89,13 @@
       "index.ts",
       ".+\\/__tests__\\/.+\\.factory.(t|j)s"
     ],
-    "testEnvironment": "node"
+    "testEnvironment": "node",
+    "moduleNameMapper": {
+      "^@domain/(.*)$": "<rootDir>/../src/domain/$1",
+      "^@data/(.*)$": "<rootDir>/../src/datasources/$1",
+      "^@routes/(.*)$": "<rootDir>/../src/routes/$1",
```
iamacook: I would sooner suggest `@/` as an alias. There is otherwise overlap with packages, e.g. `@safe-global/safe-deployments`.

iamacook (changes requested):

package.json
usame-algan: Interesting, I haven't thought about that. Afaik the `@` notation refers to scoped packages but I am not sure if there are any restrictions on them i.e. if someone could publish a package with scope `@domain` or `@routes`. Can we somehow test how this could be an issue with an example?

iamacook: You could install a package that begins with one of the specified aliases.

I don't think there should be a differing prefix though. It's [currently possible](https://github.com/safe-global/safe-client-gateway-nest/blob/path-alias/tsconfig.json#L22-L25) to import from `@/routes/xyz` or `@routes/xyz`.

usame-algan: I've tested it with this package: https://github.com/npm/validate-npm-package-name and it seems a package like `@data/something` would be valid so there could indeed be overlaps. Wdyt about only using `@/` @fmrsabino @hectorgomezv ?

fmrsabino: @usame-algan @iamacook do you have any references on what the issue would be if there's a package named `data` and we have a mapping for `@data`?

usame-algan: I don't think its a huge issue because it would require us to install a package with such an overlap. Nevertheless I went ahead an installed a package called `@routes/health` just to see how my IDE would behave. There is no error for existing `@routes` imports. When I try to import something from `@routes/...` it will actually suggest both the external package (first suggestion) and the internal paths (everything else).
Screenshot 2023-08-29 at 16 00 29

Considering this imo we can also keep it as `@domain` etc.

fmrsabino: I'd be interested in knowing how it resolves those conflicts.

While I see the value of just having an `@` I think we get a lot more "semantic" meaning by having the `@domain`, `@data` or `@routes`.

What do you think on, given the above, removing the `@` rule. Meaning that anything that is not under domain, data or routes we should consider adding another "module" (no need to tackle it in this PR though).

iamacook: > While I see the value of just having an @ I think we get a lot more "semantic" meaning by having the @domain, @data or @routes.

I think the inclusion of `/` gives a clear distinction between packages (as they cannot have `/` in their name) vs. aliases but I understand that it is personal preference.

> What do you think on, given the above, removing the @ rule. Meaning that anything that is not under domain, data or routes we should consider adding another "module" (no need to tackle it in this PR though).

I see no issue in having a "global" alias which will encompass this and we'd then not need make any future changes.

fmrsabino (approved):
Great work! 🚀 

I think we should explore solutions that centralize the configuration a bit more. Right now we have it `package.json`, `jest-all.json`, `jest-e2e.json` and `tsconfig.json`. I understand the need for each one but is bound to issues if we need to change one path in the future.

iamacook (approved):

PR #634, opened 2023-08-10
Decrease the usage of all-transactions endpoint

### LINKED ISSUE #630: Avoid the usage of Transaction Service's all-transactions endpoint
The usages of the `all-transactions` endpoint could be replaced by individual calls to get Multisig, Module, and Transfer transactions.

Currently, the CGW is using that endpoint for two things:
- Getting the paginated transactions history.
- Getting the last executed transaction.

The first use case usage is hard to replace since the CGW does not know on beforehand the number of transactions that need to be fetched of each type to compose a page of results. The second one could be replaced by several individual calls.

**Solution**
- Replace the `all-transactions` calls, when getting the last executed transaction, by individual calls for each transaction type.
- Compare the execution date for the last transaction of each kind.
- Return the most recent transaction, based on the comparison.

hectorgomezv:
Closes #630 

This PR aims to decrease the usage of the Transaction Service `all-transactions` endpoint in favor of several calls to `/multisig-transactions`, `/module-transactions`, and `/transfers`. The changes include:

- Remove now unused `SafeRepository` functions, create a new function, and rename other functions.
- Adjust cache invalidations to clear the `MultisigTransactions` and `Transfers` when one of `[INCOMING_ETHER, OUTGOING_ETHER, INCOMING_TOKEN, OUTGOING_TOKEN]` events happen. This was previously covered by `clearAllExecutedTransactions` but that's not the case anymore since we are using other endpoints.
- Adjusts the tests in `safes.controller.spec.ts` accordingly.
- Adjusts the function which calculates the `txHistoryTag` to use the three endpoints.

### DIFF
--- a/src/domain/safe/safe.repository.interface.ts
+++ b/src/domain/safe/safe.repository.interface.ts
@@ -23,10 +23,7 @@ export interface ISafeRepository {
     offset?: number;
   }): Promise<Page<Transfer>>;
 
-  clearCollectibleTransfers(args: {
-    chainId: string;
-    safeAddress: string;
-  }): Promise<void>;
+  clearTransfers(args: { chainId: string; safeAddress: string }): Promise<void>;
 
   getIncomingTransfers(args: {
     chainId: string;
--- a/src/domain/safe/safe.repository.interface.ts
+++ b/src/domain/safe/safe.repository.interface.ts
@@ -95,13 +95,6 @@ export interface ISafeRepository {
     offset?: number;
   }): Promise<Page<MultisigTransaction>>;
 
-  getTransactionHistoryByExecutionDate(args: {
-    chainId: string;
-    safeAddress: string;
-    limit?: number;
-    offset?: number;
-  }): Promise<Page<Transaction>>;
-
   getCreationTransaction(args: {
     chainId: string;
     safeAddress: string;
@@ -148,6 +141,12 @@ export interface ISafeRepository {
   }): Promise<Page<MultisigTransaction>>;
 
   getTransfer(args: { chainId: string; transferId: string }): Promise<Transfer>;
+
+  getTransfers(args: {
+    chainId: string;
+    safeAddress: string;
+    limit?: number;
+  }): Promise<Page<Transfer>>;
 
   getSafesByOwner(args: {
     chainId: string;
--- a/src/domain/safe/safe.repository.ts
+++ b/src/domain/safe/safe.repository.ts
@@ -63,7 +63,7 @@ export class SafeRepository implements ISafeRepository {
     return this.transferValidator.validatePage(page);
   }
 
-  async clearCollectibleTransfers(args: {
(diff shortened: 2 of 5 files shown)


hectorgomezv:
Thanks to @Uxio0 and @moisses89, who inspired this change with their findings.

src/domain/safe/safe.repository.interface.ts
```diff
@@ -23,10 +23,7 @@ export interface ISafeRepository {
     offset?: number;
   }): Promise<Page<Transfer>>;
 
-  clearCollectibleTransfers(args: {
-    chainId: string;
-    safeAddress: string;
-  }): Promise<void>;
+  clearTransfers(args: { chainId: string; safeAddress: string }): Promise<void>;
```
hectorgomezv, line 26: Just renaming here. This function calls `TransactionApi.clearTransfers` which clears all the `transfers`, not only _collectibles_ (ERC-721).

src/domain/safe/safe.repository.ts
```diff
@@ -203,18 +203,6 @@ export class SafeRepository implements ISafeRepository {
...
-    safeAddress: string;
-    limit?: number;
-    offset?: number;
-  }): Promise<Page<Transaction>> {
-    return this.getAllExecutedTransactions({
-      ...args,
-      ordering: 'executionDate',
-    });
-  }
-
```
hectorgomezv, line 217: This wouldn't be used anymore after this PR.

src/routes/cache-hooks/cache-hooks.service.ts
```diff
@@ -130,6 +130,14 @@ export class CacheHooksService {
...
             safeAddress: event.address,
           }),
+          this.safeRepository.clearMultisigTransactions({
+            chainId: event.chainId,
+            safeAddress: event.address,
+          }),
+          this.safeRepository.clearTransfers({
+            chainId: event.chainId,
+            safeAddress: event.address,
+          }),
```
hectorgomezv, line 140: Clearing both `MultisigTransactions` and `Transfers` in several hooks defined in this class, since the PR changes expect to have these endpoints cache clearer when an event happens in order to get the last timestamp.

src/routes/safes/safes.service.ts
```diff
@@ -141,33 +137,48 @@ export class SafesService {
...
+      }),
+      this.safeRepository.getTransfers({
+        chainId,
+        safeAddress,
+        limit: 1,
+      }),
+    ]);
+
+    const dates = txPages
+      .flatMap(({ results }) => [...results])
```
iamacook, line 175: I had to fiddle with the types a bit, but this satisfies TypeScript, as well as removes the unnecessary spread:

```suggestion
      .flatMap(({ results }): (MultisigTransaction | ModuleTransaction | Transfer)[] =>
        results
      )
```

iamacook (approved):

PR #644, opened 2023-08-18
Return current timestamp for messagesTag

fmrsabino:
Since there are no hooks implemented for `messages`, the service is always fetching the `messages` endpoint for Safe Info. We can avoid fetching these messages by returning the current timestamp instead and let the clients fetch the messages directly when needed.


fmrsabino:
We can consider reverting this change once web hooks are implemented for messages. This way we would not fetch the `messages` if there are no changes.

hectorgomezv (approved):

fmrsabino:
Closing in favour of https://github.com/safe-global/safe-client-gateway-nest/pull/654

PR #654, opened 2023-08-23
Manage caching for messages

### LINKED ISSUE #653: Manage caching for messages
# Background
Since a webhook was implemented in the Transaction Service for message-related actions (https://github.com/safe-global/safe-transaction-service/pull/1629), now it is possible for the CGW to manage a `Message` cache. See https://github.com/safe-global/safe-transaction-service/issues/1240 to gather more context on the specification.

# Problem
- Until now, messages were not being cached in the CGW, due to the lack of a proper webhook.

# Solution 
- Store the messages for a given Safe address in the cache.
- Implement a handler that will clear the related cache fields when the associated webhook is received.

hectorgomezv:
Closes #653 

This PR implements the cache storing/clearing for the `messages` coming from the Transaction Service. 

The cache hooks implemented in https://github.com/safe-global/safe-transaction-service/pull/1629 is now handled by the CGW so it's possible to clear the messages cache by sending those webhooks.

Additional changes: 
- `MessagesService.getMessagesBySafe` was optimized by grouping a couple of non-dependent promises to be executed in parallel.

### DIFF
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -26,7 +26,7 @@ export default (): ReturnType<typeof configuration> => ({
   },
   features: {
     humanDescription: true,
-    messagesCache: false,
+    messagesCache: true,
   },
   httpClient: { requestTimeout: faker.number.int() },
   log: {


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -831,6 +831,7 @@ export class TransactionApi implements ITransactionApi {
             offset: args.offset,
           },
         },
+        this.defaultExpirationTimeInSeconds,
```
hectorgomezv, line 834: Context: sending a value here makes the underlying Redis command `EXPIRE` actually set a TTL for the cache registry. We were sending `undefined` to the Redis driver, which made the driver to immediately delete the registry, (from Redis docs):
```
Note that calling EXPIRE/[PEXPIRE](https://redis.io/commands/pexpire) with a non-positive timeout or [EXPIREAT](https://redis.io/commands/expireat)/[PEXPIREAT](https://redis.io/commands/pexpireat) with a time in the past will result in the key being [deleted](https://redis.io/commands/del) rather than expired
```

So this is the main change to store the `messages` in the cache.

src/routes/cache-hooks/entities/new-message-confirmation.entity.ts
```diff
@@ -0,0 +1,6 @@
+import { EventPayload, EventType } from './event-payload.entity';
+
+export interface NewMessageConfirmation
```
hectorgomezv, line 3: I chose this `NewMessageConfirmation` name because both `MessageConfirmation` and `NewConfirmation` are already used in the CGW. Instead of using `New*` I also thought of a `*Event` suffix, but it would be the only event using that suffix, so I decided on `NewMessageConfirmation`. 

If you have a better alternative here, suggestions are welcome. 🙂

src/routes/messages/messages.service.ts
```diff
@@ -45,16 +45,18 @@ export class MessagesService {
...
-      chainId: args.chainId,
-      address: args.safeAddress,
-    });
-    const page = await this.messagesRepository.getMessagesBySafe({
-      chainId: args.chainId,
-      safeAddress: args.safeAddress,
-      limit: args.paginationData.limit,
-      offset: args.paginationData.offset,
-    });
+    const [safe, page] = await Promise.all([
```
hectorgomezv, line 48: Not strictly related, but I saw this performance increase opportunity while reviewing the messages retrieval flow. We can do it separately if you prefer.

src/routes/cache-hooks/entities/new-message-confirmation.entity.ts
fmrsabino: This is ok from my side! 👍

We can consider `Event` as a suffix. The location of the class should also help when importing it (even when we have duplicate names) but I agree that it's easier to make the mistake of importing the wrong type if we use the same names.

src/domain/interfaces/transaction-api.interface.ts
```diff
@@ -207,4 +207,8 @@ export interface ITransactionApi {
     messageHash: string;
     signature: string;
   }): Promise<unknown>;
+
+  clearMessagesBySafe(args: { safeAddress: string });
+
+  clearMessagesByHash(args: { messageHash: string });
```
fmrsabino, line 213: Missing explicit return type here.

src/routes/messages/messages.service.ts
fmrsabino: In general I prefer to leave the changes within the scope of the PR. You can add this change to the commit message/PR description 👍

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -804,6 +804,8 @@ export class TransactionApi implements ITransactionApi {
         cacheDir,
         url,
         this.defaultNotFoundExpirationTimeSeconds,
+        undefined,
+        this.defaultExpirationTimeInSeconds,
```
fmrsabino, line 808: We need a feature flag for this one for one reason: if we deploy a production version of the CGW before the hooks are on production on the Transaction Service then the users will not see new messages until the cache entry expires.

fmrsabino (changes requested):
Nice! There's one concern that I have regarding syncing releases.

src/datasources/transaction-api/transaction-api.service.ts
fmrsabino: Unrelated but I think we'd benefit by having an anonymous interface for the arguments as with the number of `undefined` involved this can cause bugs in the future.

hectorgomezv: 💯 I'd add an anonymous interface to `CacheFirstDataSource` in a separate PR if you agree, since a lot of file would be affected 🙂

src/routes/cache-hooks/entities/new-message-confirmation.entity.ts
hectorgomezv: Yeah, no strong opinion on this. We could think of adding `*Event` suffix to all webhook events in a separate issue, especially if the list keeps growing.

src/routes/messages/messages.service.ts
hectorgomezv: Done, added to the PR description, thanks!

src/datasources/transaction-api/transaction-api.service.ts
hectorgomezv: Good point, I was thinking of releasing this after the release in the Transaction Service, but maybe is more solid if we deliver this behind an FF, even if there's some work for cleaning after that, to cover the scenario where the Transaction Service needs to rollback for some reason after its deployment. 

I'll implement the FF for this 👍

src/domain/interfaces/transaction-api.interface.ts
hectorgomezv: Added in [add1cee](https://github.com/safe-global/safe-client-gateway-nest/pull/654/commits/add1cee38699a4159183b097465a5e79e2889962)

src/config/entities/configuration.ts
```diff
@@ -31,6 +31,11 @@ export default () => ({
       ),
     },
   },
+  features: {
```
hectorgomezv, line 34: Just moving the attribute placement to follow the alphabetical order 🙂

src/config/entities/__tests__/configuration.ts
```diff
@@ -24,6 +24,10 @@ export default (): ReturnType<typeof configuration> => ({
       token: faker.number.int(),
     },
   },
+  features: {
+    humanDescription: true,
+    messagesCache: false,
```
fmrsabino, line 29: OOC: why is this disabled in the tests?

src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -47,6 +47,7 @@ describe('TransactionApi', () => {
 
     defaultExpirationTimeInSeconds = faker.number.int();
     notFoundExpireTimeSeconds = faker.number.int();
+    const messagesCache = faker.datatype.boolean();
```
fmrsabino, line 50: This test class needs a review as it stands right now, it seems incomplete and is probably covered by other tests.

fmrsabino (approved):

src/config/entities/__tests__/configuration.ts
hectorgomezv: This is just set because the test configuration needs to follow the `configuration.ts` interface, but it's not used by unit tests (besides the ones which specifically set it, like `transaction-api.service.spec.ts`, so I picked a boolean value. Maybe it's more accurate if we use `faker.datatype.boolean()` also here, wdyt?

Note: It could have an influence on e2e tests, but we don't have any for messages. In that case, it should be set in `e2e-setup.ts` if I'm not wrong.

fmrsabino: This file represents the test configuration and since this is a feature flag that is bound to be deleted once the releases are synced I think it makes sense for it to be `true`.

If tests are written that target messages we should assume tests that handle caching on the messages side.

I don't think `faker.datatype.boolean()` is a good choice for this feature flag in this case as each test should assume this new standard. I think it'd make sense to have it `false` if the service assumed no caching for the messages AND we'd have tests targeting when the feature is enabled (there we'd set it to `true`).

hectorgomezv: Yes, makes perfect sense, I agree.

I've updated it in [e624337](https://github.com/safe-global/safe-client-gateway-nest/pull/654/commits/e6243371ad9238adaf8ed44d678134365d6a3608), thanks!

fmrsabino (approved):
🚀

PR #671, opened 2023-08-31
feat: Human readable v2

usame-algan:
## What it solves

Part of https://github.com/safe-global/safe-wallet-web/issues/2449

- Returns human descriptions as structured data instead of a string
- Adds `RichHumanDescriptionFragments` on top of `HumanDescriptionFragments` to hold information that is meant to be used by clients
- Caches the regex match for each `HumanDescriptionTemplate`

### DIFF
--- a/src/domain/human-description/entities/human-description-template.entity.ts
+++ b/src/domain/human-description/entities/human-description-template.entity.ts
@@ -23,11 +23,20 @@ export class HumanDescriptionTemplate {
   private static readonly REGEX =
     /{{(?<typeToken>\w+) \$(?<paramIndex>\d+)}}|(?<wordToken>\w+)/g;
 
+  /**
+   * Store the regex matches as an array instead of an iterable so that it can be restarted
+   * @private
+   */
+  private readonly templateMatches: RegExpMatchArray[];
+
   constructor(
     functionSignature: string,
     private readonly template: string,
   ) {
     this.functionAbi = parseAbi([functionSignature]);
+    this.templateMatches = Array.from(
+      template.matchAll(HumanDescriptionTemplate.REGEX),
+    );
   }
 
   /**
@@ -46,9 +55,8 @@ export class HumanDescriptionTemplate {
     });
 
     const fragments: HumanDescriptionFragment[] = [];
-    const matches = this.template.matchAll(HumanDescriptionTemplate.REGEX);
 
-    for (const match of matches) {
+    for (const match of this.templateMatches) {
       if (!match.groups) throw Error(`Error parsing template ${this.template}`);
 
       if ('wordToken' in match.groups && match.groups.wordToken !== undefined) {
--- a/src/routes/transactions/entities/human-description.entity.ts
+++ b/src/routes/transactions/entities/human-description.entity.ts
@@ -1,0 +1,25 @@
+import { Token } from '@/domain/tokens/entities/token.entity';
+import {
+  AddressFragment,
+  NumberFragment,
+  ValueType,
+  WordFragment,
+} from '@/domain/human-description/entities/human-description.entity';
+
+export interface RichTokenValueFragment {
+  type: ValueType.TokenValue;
+  value: {
+    amount: string;
+    token: Token | null;
+  };
+}
+
+export interface RichWordFragment extends WordFragment {}
+export interface RichAddressFragment extends AddressFragment {}
+export interface RichNumberFragment extends NumberFragment {}
+
+export type RichHumanDescriptionFragment =
+  | RichWordFragment
+  | RichTokenValueFragment
+  | RichAddressFragment
+  | RichNumberFragment;
--- a/src/routes/transactions/entities/transaction-info.entity.ts
+++ b/src/routes/transactions/entities/transaction-info.entity.ts
@@ -1,12 +1,16 @@
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { RichHumanDescriptionFragment } from '@/routes/transactions/entities/human-description.entity';
 
 export class TransactionInfo {
   @ApiProperty()
   type: string;
   @ApiPropertyOptional({ type: String, nullable: true })
-  humanDescription: string | null;
+  humanDescription: RichHumanDescriptionFragment[] | null;
 
-  protected constructor(type: string, humanDescription: string | null) {
+  protected constructor(
+    type: string,
+    humanDescription: RichHumanDescriptionFragment[] | null,
+  ) {
     this.type = type;
     this.humanDescription = humanDescription;
   }
--- a/src/routes/transactions/mappers/common/custom-transaction.mapper.spec.ts
+++ b/src/routes/transactions/mappers/common/custom-transaction.mapper.spec.ts
@@ -192,7 +192,7 @@ describe('Multisig Custom Transaction mapper (Unit)', () => {
     const dataSize = 0;
     const chainId = faker.string.numeric();
     const transaction = multisigTransactionBuilder().build();
-    const humanDescription = faker.word.words();
+    const humanDescription = [];
 
     const customTransaction = await mapper.mapCustomTransaction(
       transaction,
--- a/src/routes/transactions/mappers/common/human-description.mapper.ts
+++ b/src/routes/transactions/mappers/common/human-description.mapper.ts
@@ -1,21 +1,26 @@
 import { Hex } from 'viem/src/types/misc';
 import { Inject, Injectable } from '@nestjs/common';
-import { formatUnits, isAddress, isHex } from 'viem';
+import { formatUnits, isHex } from 'viem';
 import { ITokenRepository } from '@/domain/tokens/token.repository.interface';
 import { TokenRepository } from '@/domain/tokens/token.repository';
-import { Token } from '@/domain/tokens/entities/token.entity';
 import { MAX_UINT256 } from '../../constants';
 import { ILoggingService, LoggingService } from '@/logging/logging.interface';
 import { IHumanDescriptionRepository } from '@/domain/human-description/human-description.repository.interface';
 import { HumanDescriptionRepository } from '@/domain/human-description/human-description.repository';
 import {
   HumanDescriptionFragment,
+  TokenValueFragment,
   ValueType,
 } from '@/domain/human-description/entities/human-description.entity';
 import { MultisigTransaction } from '@/domain/safe/entities/multisig-transaction.entity';
 import { ModuleTransaction } from '@/domain/safe/entities/module-transaction.entity';
 import { isMultisigTransaction } from '@/domain/safe/entities/transaction.entity';
 import { SafeAppInfoMapper } from './safe-app-info.mapper';
+import {
+  RichHumanDescriptionFragment,
+  RichTokenValueFragment,
+  RichWordFragment,
+} from '@/routes/transactions/entities/human-description.entity';
 
 @Injectable()
 export class HumanDescriptionMapper {
@@ -32,7 +37,7 @@ export class HumanDescriptionMapper {
   async mapHumanDescription(
     transaction: MultisigTransaction | ModuleTransaction,
     chainId: string,
-  ): Promise<string | null> {
+  ): Promise<RichHumanDescriptionFragment[] | null> {
     if (!transaction.data || !isHex(transaction.data) || !transaction.to) {
       return null;
     }
@@ -41,36 +46,25 @@ export class HumanDescriptionMapper {
 
     if (!sigHash) return null;
 
-    let token: Token | null = null;
     try {
-      token = await this.tokenRepository.getToken({
-        chainId,
-        address: transaction.to,
-      });
-    } catch (error) {
-      this.loggingService.debug(`Error trying to get token: ${error.message}`);
-    }
-
-    try {
-      const descriptionFragments =
+      const humanDescriptionFragments =
         this.humanDescriptionRepository.getHumanDescription({
           functionSignatureHash: sigHash,
           to: transaction.to,
           data: transaction.data,
         });
 
-      const description = this.createHumanDescription(
-        descriptionFragments,
-        token,
+      const humanDescriptionComponents = await this.enrichFragments(
+        humanDescriptionFragments,
+        transaction,
+        chainId,
       );
 
-      const safeAppInfo = isMultisigTransaction(transaction)
-        ? await this.safeAppInfoMapper.mapSafeAppInfo(chainId, transaction)
-        : null;
-
-      return safeAppInfo
-        ? `${description} via ${safeAppInfo.name}`
-        : description;
+      return this.enrichSafeAppInfo(
+        humanDescriptionComponents,
+        transaction,
+        chainId,
+      );
     } catch (error) {
       this.loggingService.debug(
         `Error trying to decode the input data: ${error.message}`,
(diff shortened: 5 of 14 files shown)


src/routes/transactions/entities/transaction-info.entity.ts
```diff
@@ -1,12 +1,16 @@
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import { RichHumanDescriptionFragment } from '@/routes/transactions/entities/human-description.entity';
 
 export class TransactionInfo {
   @ApiProperty()
   type: string;
   @ApiPropertyOptional({ type: String, nullable: true })
-  humanDescription: string | null;
+  humanDescription: RichHumanDescriptionFragment[] | null;
```
fmrsabino, line 8: ⚠️ This is API breaking – merging this would break clients that are currently expecting a `string`.

fmrsabino, line 8: ❓ What is the difference between `null` and an empty array for `humanDescription`?

fmrsabino (changes requested):

src/domain/human-description/entities/human-description-template.entity.ts
```diff
@@ -46,9 +55,8 @@ export class HumanDescriptionTemplate {
     });
 
     const fragments: HumanDescriptionFragment[] = [];
-    const matches = this.template.matchAll(HumanDescriptionTemplate.REGEX);
 
-    for (const match of matches) {
+    for (const match of this.templateMatches) {
       if (!match.groups) throw Error(`Error parsing template ${this.template}`);
```
fmrsabino, line 60: This error can be moved to the constructor since we do the `matchAll` there. At that point we already know if this template is valid or not.

src/routes/transactions/entities/human-description.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+  NumberFragment,
+  ValueType,
+  WordFragment,
+} from '@/domain/human-description/entities/human-description.entity';
+
+export interface RichTokenValueFragment {
+  type: ValueType.TokenValue;
+  value: {
+    amount: string;
+    token: Token | null;
```
fmrsabino, line 13: I would not expose the domain entity directly here to the clients. The main reason is if we refactor this domain entity, we would break the clients.

By having what the clients expect separate from the domain entity we know that by changing the domain entity we also need to change the mapping.

In order to reduce nesting I would consider for now only what the clients for rendering. Suggestion:

```typescript
export interface RichTokenValueFragment {
  type: ValueType.TokenValue;
  amount: string;
  symbol: string;
  logoUri: string;
}
```

src/routes/transactions/mappers/common/custom-transaction.mapper.spec.ts
```diff
@@ -192,7 +192,7 @@ describe('Multisig Custom Transaction mapper (Unit)', () => {
     const dataSize = 0;
     const chainId = faker.string.numeric();
     const transaction = multisigTransactionBuilder().build();
-    const humanDescription = faker.word.words();
+    const humanDescription = [];
```
fmrsabino, line 195: We should adjust the test to generate human descriptions that are ready to be consumed by the clients.

We can leverage the builder pattern to make it easier to write those tests 😉

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -79,45 +73,75 @@ export class HumanDescriptionMapper {
...
+      return <RichTokenValueFragment>{
+        type: ValueType.TokenValue,
+        value: {
+          token,
+          amount:
+            fragment.value.amount === MAX_UINT256
+              ? 'unlimited'
+              : token.decimals
+              ? formatUnits(fragment.value.amount, token.decimals)
+              : fragment.value.amount.toString(),
```
fmrsabino, line 114: I think the chain of conditions for the `amount` can be confusing for the reader.

Suggestion (extract the conditions):

```typescript
let amountLabel: string;
if (fragment.value.amount === MAX_UINT256) {
  amountLabel = 'unlimited';
} else if (token && token.decimals) {
  amountLabel = formatUnits(fragment.value.amount, token.decimals);
} else {
  amountLabel = fragment.value.amount.toString();
}
```

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -79,45 +73,75 @@ export class HumanDescriptionMapper {
...
+  async enrichTokenValue(
+    fragment: TokenValueFragment,
+    transaction: MultisigTransaction | ModuleTransaction,
+    chainId: string,
+  ): Promise<RichTokenValueFragment> {
+    try {
+      const token = await this.tokenRepository.getToken({
+        chainId,
+        address: transaction.to,
+      });
```
fmrsabino, line 103: Here we can ignore the result of the token if we don't get any. Therefore token can be `Token | null`:

```typescript
const token: Token | null = await this.tokenRepository
  .getToken({
    chainId,
    address: transaction.to,
  })
  .catch(() => null);

// taken from my comment below 👇
let amountLabel: string;
if (fragment.value.amount === MAX_UINT256) {
  amountLabel = 'unlimited';
} else if (token && token.decimals) {
  amountLabel = formatUnits(fragment.value.amount, token.decimals);
} else {
  amountLabel = fragment.value.amount.toString();
}

return <RichTokenValueFragment>{
  type: ValueType.TokenValue,
  value: {
    token,
    amount: amountLabel,
  },
};
```

fmrsabino (changes requested):

src/routes/transactions/mappers/common/human-description.mapper.ts
usame-algan: This was my first approach as well but somehow the `catch` clause here is not catching an error when throwing in the test so I changed it to use `try catch` which works.

src/routes/transactions/entities/transaction-info.entity.ts
usame-algan: How do we deal with that? Practically, this should only be in use by the web client so far but there is no explicit dependency so increasing the version number by a breaking change wouldn't help. On the other hand adding the new change in a backwards compatible way seems excessive.

fmrsabino: When working with the API which is exposed to the any client we should always consider on how the contract between the clients and the service breaks.

That's one of the reasons why I've also suggested to use a feature flag so that the API can evolve without breaking the client-side.

This should always be considered when working with an API contract as we might have 3rd parties to the service that expect `X` but get `Y` after some time.

What happens if this gets merged as-is on the current state of the Safe Web App?

usame-algan: I don't think there is a semantic difference but if we want it to be clear I suggest we go with `null` otherwise clients would have to check the length of the array in order to be sure that it exists which is not as explicit as `null`.

src/domain/human-description/entities/human-description-template.entity.ts
usame-algan: Makes sense! I've moved the check into the constructor.

src/routes/transactions/entities/human-description.entity.ts
usame-algan: With domain entity do you mean the `ValueType` enum or that this type has the same "structure" as the fragment from the domain entity? I went with a flat structure now and only return what is needed as you suggested.

src/routes/transactions/mappers/common/custom-transaction.mapper.spec.ts
usame-algan: Good idea, I've created a new builder for human description fragments and used it in this test.

katspaugh:
@usame-algan let's discuss and stabilize the API specs before reviewing this.

usame-algan:
> @usame-algan let's discuss and stabilize the API specs before reviewing this.

I've added the proposed changes from the API Spec.

src/domain/human-description/entities/human-description-template.entity.ts
```diff
@@ -5,9 +5,17 @@ import {
...
+  TextFragment,
 } from './human-description.entity';
 
+type SafeRegExpMatchArray = RegExpMatchArray & {
+  groups: NonNullable<RegExpMatchArray['groups']>;
+};
+
+function hasGroups(match: RegExpMatchArray): match is SafeRegExpMatchArray {
+  return !!match.groups;
+}
```
fmrsabino, line 17: I don't think this is clear. Do we want to check if it's not undefined? How about:

```typescript
return match.groups !== undefined
```

src/domain/human-description/entities/human-description-template.entity.ts
```diff
@@ -23,11 +31,29 @@ export class HumanDescriptionTemplate {
...
+      template.matchAll(HumanDescriptionTemplate.REGEX),
+    )
+      .map((match) => {
+        if (!match.groups) {
+          throw Error(`Error parsing template ${this.template}`);
+        }
+
+        return match;
+      })
+      .filter(hasGroups);
```
fmrsabino, line 56: The `map` above checks if the matches have groups but then we call `hasGroups` again here.

Shouldn't the filter tackle this for us beforehand?

fmrsabino (changes requested):

src/domain/human-description/entities/human-description.entity.ts
```diff
@@ -1,37 +1,44 @@
 import { Hex } from 'viem/src/types/misc';
 
 export enum ValueType {
-  Word = 'word',
+  Text = 'text',
```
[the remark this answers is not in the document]
usame-algan: Afaik it should be changed everywhere now. Can you point me to where it needs to be updated still?

src/routes/transactions/entities/__tests__/human-description.builder.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+const humanDescriptionBuilders: Array<() => IBuilder<RichInfoFragment>> = [
+  richTokenValueFragmentBuilder,
+  richWordFragmentBuilder,
+  richAddressFragmentBuilder,
+  richNumberFragmentBuilder,
+];
+
+export function buildHumanDescription(): RichInfoFragment[] {
```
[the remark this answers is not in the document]
usame-algan: In the case where I am using this function the mapper expects rich fragments. However we could create a Builder for fragments that we can use to refactor the human-description.mapper tests.

src/domain/human-description/entities/human-description-template.entity.ts
usame-algan: This is an issue with Typescript not being able to filter out undefined values so we have to use a type-guard. Alternatively, we could also cast the value but I think its less clean.

src/routes/transactions/entities/human-description.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+  };
+}
+
+export interface RichTextFragment extends RichInfoFragment {}
+export interface RichAddressFragment extends RichInfoFragment {}
+export interface RichNumberFragment extends RichInfoFragment {}
+
+export type RichInfo = {
+  fragments: RichInfoFragment[];
+};
```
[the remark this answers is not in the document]
usame-algan: It will be useful if we decide to add another property like the one proposed in the API spec for multisend support. In that case we would only have to change it in one place.

src/domain/human-description/entities/human-description.entity.ts
```diff
@@ -1,37 +1,44 @@
...
   TokenValue = 'tokenValue',
   Address = 'address',
   Number = 'number',
 }
 
-export interface TokenValueFragment {
+export interface InfoFragment {
+  type: ValueType;
+  value: string | bigint | `0x${string}` | TokenValue;
```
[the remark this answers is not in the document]
usame-algan: No big difference afaik. `HumanDescriptionFragment` was an explicit union type of all fragments while `InfoFragment` defines properties that exist for all fragments that extend from it so its a single source of truth and potentially requires fewer type-guards. Wdyt @iamacook?

src/routes/transactions/entities/__tests__/human-description.builder.ts
```diff
@@ -0,0 +1,51 @@
...
+function richWordFragmentBuilder(): IBuilder<RichTextFragment> {
+  return Builder.new<RichTextFragment>()
+    .with('type', ValueType.Text)
+    .with('value', faker.word.adverb());
+}
+
+function richAddressFragmentBuilder(): IBuilder<RichAddressFragment> {
+  return Builder.new<RichAddressFragment>()
+    .with('type', ValueType.Address)
+    .with('value', faker.finance.ethereumAddress() as `0x${string}`);
```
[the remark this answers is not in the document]
usame-algan: After removing the generic union type we have to cast here again.

src/routes/transactions/entities/human-description.entity.ts
```diff
@@ -0,0 +1,25 @@
+import {
+  InfoFragment,
+  ValueType,
+} from '@/domain/human-description/entities/human-description.entity';
+
+export interface RichInfoFragment extends InfoFragment {
+  richData: Record<string, unknown> | null;
+}
```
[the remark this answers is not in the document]
usame-algan: Removed it

src/routes/transactions/entities/human-description.entity.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export interface RichTokenValueFragment extends RichInfoFragment {
+  type: ValueType.TokenValue;
+  value: string;
+  richData: {
+    symbol: string | null;
+    logoUri: string | null;
+  };
```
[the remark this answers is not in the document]
usame-algan: As we are not using extending from `RichInfoFragment` anymore I would also be in favour of flattening this since at the moment this is only being used by one fragment.

fmrsabino (changes requested):

src/domain/human-description/entities/human-description.entity.ts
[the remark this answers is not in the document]
usame-algan: Ah thanks for the explanation. Now I understand. I agree in this case it works but for other value types it would also require changes to the JSON which we don't want to do. Also good find with the regex. Lets rename it there to `textToken` as well.

src/domain/human-description/entities/human-description.entity.ts
```diff
@@ -1,22 +1,24 @@
...
+  Text = 'text',
   TokenValue = 'tokenValue',
   Address = 'address',
   Number = 'number',
 }
 
+type TokenValue = {
+  amount: bigint;
+  address: string;
+};
```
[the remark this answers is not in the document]
usame-algan: True, it is a leftover of the previous change so I will revert it back as it is not being exported or used elsewhere.

src/routes/transactions/entities/__tests__/human-description.builder.ts
```diff
@@ -0,0 +1,42 @@
...
+  RichTextFragment,
+  RichDecodedInfoFragment,
+  RichFragmentType,
+} from '@/routes/transactions/entities/human-description.entity';
+import { faker } from '@faker-js/faker';
+
+function richTokenValueFragmentBuilder(): IBuilder<RichTokenValueFragment> {
+  return Builder.new<RichTokenValueFragment>()
+    .with('type', RichFragmentType.TokenValue)
+    .with('value', faker.number.int().toString())
```
fmrsabino, line 14:
```suggestion
    .with('value', faker.string.numeric())
```

See https://fakerjs.dev/api/string.html#numeric

src/routes/transactions/entities/__tests__/human-description.builder.ts
```diff
@@ -0,0 +1,42 @@
...
+    .with('symbol', faker.finance.currencySymbol())
+    .with('logoUri', faker.internet.avatar());
+}
+
+function richWordFragmentBuilder(): IBuilder<RichTextFragment> {
+  return Builder.new<RichTextFragment>()
+    .with('type', RichFragmentType.Text)
+    .with('value', faker.word.adverb());
```
fmrsabino, line 22: nitpick: semantics only (no need to change) but you also have the following functions:

`faker.lorem.word()` – https://fakerjs.dev/api/lorem.html#word
`faker.word.sample()` – https://fakerjs.dev/api/word.html#sample

src/routes/transactions/entities/human-description.entity.ts
```diff
@@ -0,0 +1,68 @@
...
+    super(RichFragmentType.Text, value);
+  }
+}
+
+export class RichAddressFragment extends RichDecodedInfoFragment {
+  constructor(value: `0x${string}`) {
+    super(RichFragmentType.Address, value);
+  }
+}
+@ApiExtraModels(RichTokenValueFragment, RichTextFragment, RichAddressFragment)
```
fmrsabino, line 51: nitpick: new line between the classes (should probably be a rule set by prettier)

```suggestion
}

@ApiExtraModels(RichTokenValueFragment, RichTextFragment, RichAddressFragment)
```

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -41,36 +49,29 @@ export class HumanDescriptionMapper {
...
+        transaction,
+        chainId,
+      );
 
-      return safeAppInfo
-        ? `${description} via ${safeAppInfo.name}`
-        : description;
+      return {
+        fragments: richDecodedInfo,
+      };
```
fmrsabino, line 74: Nitpick: since we have a class for this entity in the route I would return the respective instance. Wdyt?

```suggestion
      return new RichDecodedInfo(richDecodedInfo);
```

src/routes/transactions/mappers/common/human-description.mapper.ts
```diff
@@ -79,34 +80,95 @@ export class HumanDescriptionMapper {
...
+        type: RichFragmentType.Text,
+        value: `via ${safeAppInfo.name}`,
+      });
+    }
+
+    return fragments;
   }
 
-  shortenAddress(address: Hex, length = 4): string {
+  private shortenAddress(address: string, length = 4): string {
```
fmrsabino, line 171: Why was this one changed to `string`?

You can use `isHex` in the callers and keep the parameter as `Hex`.

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
```diff
@@ -135,18 +142,28 @@ describe('Human descriptions mapper (Unit)', () => {
...
+    const expectedResult = [
+      { type: ValueType.Text, value: 'Send' },
+      {
+        type: ValueType.TokenValue,
+        value: mockAmount.toString(),
+        symbol: null,
+        logoUri: null,
+      },
+      { type: ValueType.Text, value: 'to' },
+      { type: ValueType.Address, value: mockAddress },
```
fmrsabino, line 163: nitpick: use the raw values ("string") as the expected outcome. In doing so this test would fail if `ValueType.Text` (or any other) is changed accidentally.

src/routes/transactions/mappers/common/human-description.mapper.ts
usame-algan: Nice find! Makes sense

src/routes/transactions/mappers/common/human-description.mapper.ts
usame-algan: The `isAddress` call we do converts it to `Hex` thats why I simplified it here.
Screenshot 2023-09-05 at 17 06 49

src/routes/transactions/mappers/common/human-descriptions.mapper.spec.ts
usame-algan: Fair point! I actually changed it to use the enum because of that reason but it also makes sense to have it detect accidental changes to the enum.

src/routes/transactions/entities/__tests__/human-description.builder.ts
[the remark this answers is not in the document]
usame-algan: True, that is surprising since the class is expecting `0x${string}` and the faker call returns `string`. I've removed the casting nonetheless.

src/routes/transactions/mappers/common/human-description.mapper.ts
fmrsabino: Ok I was thinking of constraining the arguments instead of validating it inside but this is also ok from my side 👍

src/routes/transactions/entities/transaction-info.entity.ts
fmrsabino: We have discussed this internally and the feedback was already addressed in the PR.

PR #680, opened 2023-09-06
Add Safe Relay domain

fmrsabino:
- Adds initial version of the Safe Gelato Relay domain
- No changes to the route layer or to the data layer
- Adds dependency to `@safe-global/safe-deployments` – this dependency is used to retrieve the ABI for the Safe contracts
- Adds a `ContractHelper` – the goal of this class is to provide smart-contract decoding functionality to `domain` layer. This class is mostly a wrapper on `viem` functionality but specialises on specific ABI functions (the ABI being the state of the class)
- Adds two implementations of the `ContractHelper`:
  * `Erc20ContractHelper` – uses the ERC20 contract ABI to provide decoding functionality for the `transfer` function
  * `SafeContractHelper` – uses the Safe contract ABI to provide decoding functionality for the `execTransaction` function.

What is out of scope for this Pull Request:
- Safe Relay routes – no routes were implemented. This PR mostly targets the business logic on the domain side
- `data` layer – no datasources were added. This will be where the API which targets the Gelato Relay Service will be added
- Handling of Multisend payloads
- Handling of Create Proxy with Nonce payloads

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,6 +32,8 @@
     "@nestjs/serve-static": "^4.0.0",
     "@nestjs/swagger": "^7.1.7",
     "@node-redis/json": "^1.0.2",
+    "@openzeppelin/contracts": "^4.9.3",
+    "@safe-global/safe-deployments": "^1.26.0",
     "@types/lodash": "^4.14.197",
     "ajv": "^8.12.0",
     "ajv-formats": "^2.1.1",
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "@nestjs/serve-static": "^4.0.0",
     "@nestjs/swagger": "^7.1.7",
     "@node-redis/json": "^1.0.2",
-    "@openzeppelin/contracts": "^4.9.3",
     "@safe-global/safe-deployments": "^1.26.0",
     "@types/lodash": "^4.14.197",
     "ajv": "^8.12.0",
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -51,6 +51,9 @@ export default () => ({
     host: process.env.REDIS_HOST || 'localhost',
     port: process.env.REDIS_PORT || '6379',
   },
+  relay: {
+    limit: parseInt(process.env.RELAY_THROTTLE_LIMIT ?? `${5}`),
+  },
   safeConfig: {
     baseUri:
       process.env.SAFE_CONFIG_BASE_URI || 'https://safe-config.safe.global/',
--- a/src/domain/relay/contracts/erc20-contract.helper.spec.ts
+++ b/src/domain/relay/contracts/erc20-contract.helper.spec.ts
@@ -5,15 +5,20 @@ import { concatHex, getAddress, pad, toHex } from 'viem';
 describe('ERC20 Contract Helper Tests', () => {
   let target: Erc20ContractHelper;
 
+  const transferFunctionSignature = `0xa9059cbb`;
+
   beforeEach(() => {
     target = new Erc20ContractHelper();
   });
 
   it('decodes a transfer correctly', () => {
-    const functionSignature = `0xa9059cbb`;
     const to = getAddress(faker.finance.ethereumAddress());
     const value = toHex(faker.number.bigInt());
-    const callData = concatHex([functionSignature, pad(to), pad(value)]);
+    const callData = concatHex([
+      transferFunctionSignature,
+      pad(to),
+      pad(value),
+    ]);
 
     const actual = target.decode(Erc20ContractHelper.TRANSFER, callData);
 
@@ -34,11 +39,10 @@ describe('ERC20 Contract Helper Tests', () => {
   });
 
   it('isCall returns true for a transfer call', () => {
-    const functionSignature = `0xa9059cbb`;
     const toAddress = getAddress(faker.finance.ethereumAddress());
     const toArg = pad(toAddress);
     const valueArg = pad(toHex(faker.number.hex()));
-    const callData = concatHex([functionSignature, toArg, valueArg]);
+    const callData = concatHex([transferFunctionSignature, toArg, valueArg]);
 
     const actual = target.isCall(callData);
 
--- a/src/domain/relay/contracts/erc20-contract.helper.spec.ts
+++ b/src/domain/relay/contracts/erc20-contract.helper.spec.ts
@@ -1,0 +1,60 @@
+import { Erc20ContractHelper } from '@/domain/relay/contracts/erc20-contract.helper';
+import { faker } from '@faker-js/faker';
+import { concatHex, getAddress, pad, toHex } from 'viem';
+
+describe('ERC20 Contract Helper Tests', () => {
+  let target: Erc20ContractHelper;
+
+  beforeEach(() => {
+    target = new Erc20ContractHelper();
+  });
+
+  it('decodes a transfer correctly', () => {
+    const functionSignature = `0xa9059cbb`;
+    const to = getAddress(faker.finance.ethereumAddress());
+    const value = toHex(faker.number.bigInt());
+    const callData = concatHex([functionSignature, pad(to), pad(value)]);
+
+    const actual = target.decode(Erc20ContractHelper.TRANSFER, callData);
+
+    expect(actual).toEqual({ to });
+  });
+
+  it('decoding a non transfer call throws', () => {
+    const functionSignature = faker.string.hexadecimal({
+      length: 8,
+    }) as `0x${string}`;
+    const arg1 = pad(getAddress(faker.finance.ethereumAddress()));
+    const arg2 = pad(getAddress(faker.finance.ethereumAddress()));
+    const callData = concatHex([functionSignature, arg1, arg2]);
+
+    expect(() => {
+      target.decode(Erc20ContractHelper.TRANSFER, callData);
+    }).toThrow();
+  });
+
+  it('isCall returns true for a transfer call', () => {
+    const functionSignature = `0xa9059cbb`;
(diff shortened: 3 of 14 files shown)


package.json
```diff
@@ -32,6 +32,8 @@
     "@nestjs/serve-static": "^4.0.0",
     "@nestjs/swagger": "^7.1.7",
     "@node-redis/json": "^1.0.2",
+    "@openzeppelin/contracts": "^4.9.3",
```
fmrsabino, line 35: I'm still considering using the `transfer` ABI item for ERC20 contracts. Including a whole collection of contracts in the project does not seem necessary. 2Mb are added to the docker image with this dependency.

iamacook: As the contracts are immutable, I see no issue in not installing this package but we should analyse the bundle to see if it's "tree shakeable" first.

src/config/entities/configuration.ts
```diff
@@ -51,6 +51,9 @@ export default () => ({
     host: process.env.REDIS_HOST || 'localhost',
     port: process.env.REDIS_PORT || '6379',
   },
+  relay: {
+    limit: parseInt(process.env.RELAY_THROTTLE_LIMIT ?? `${5}`),
```
iamacook, line 55: Nit:

```suggestion
    limit: parseInt(process.env.RELAY_THROTTLE_LIMIT ?? '5'),
```

src/domain/relay/contracts/safe-contract.helper.ts
```diff
@@ -0,0 +1,35 @@
...
+    to: Hex;
+    value: bigint;
+  }> = {
+    name: 'execTransaction',
+    mapper(args: readonly unknown[]): { data: Hex; to: Hex; value: bigint } {
+      if (!isHex(args[2]) || !isHex(args[0]) || typeof args[1] !== 'bigint')
+        throw Error('Unexpected type for execTransaction args');
+      return { data: args[2], to: args[0], value: args[1] };
+    },
+  };
```
iamacook, line 25: I know it's likely a TODO, but this should support every contract method for the relayer.

src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -0,0 +1,74 @@
...
+      const erc20DecodedData = this.erc20Contract.decode(
+        Erc20ContractHelper.TRANSFER,
+        execTransaction.data,
+      );
+      // If the ERC20 transfer targets 'self' (the Safe), we consider it to be invalid
+      return erc20DecodedData.to !== to;
+    } catch {
+      this.loggingService.debug(
+        'execTransaction data is not an ERC20 transfer',
+      );
```
iamacook, line 57: I think this log is excessive. Any Safe method that isn't `execTransaction` will trigger this.

src/domain/relay/contracts/safe-contract.helper.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+  const operation = CALL_OPERATION;
+  const safeTxGas = faker.number.bigInt();
+  const baseGas = faker.number.bigInt();
+  const gasPrice = faker.number.bigInt();
+  const gasToken = getAddress(faker.finance.ethereumAddress());
+  const refundReceiver = getAddress(faker.finance.ethereumAddress());
+  const signatures = faker.string.hexadecimal({ length: 64 * 2 });
+
+  return encodeFunctionData({
+    abi: [ABI_ITEM],
```
iamacook, line 139: Instead of manually adding the ABI here, we could call the deployment getter inside this function.

src/domain/relay/contracts/erc20-contract.helper.spec.ts
```diff
@@ -0,0 +1,60 @@
...
+
+describe('ERC20 Contract Helper Tests', () => {
+  let target: Erc20ContractHelper;
+
+  beforeEach(() => {
+    target = new Erc20ContractHelper();
+  });
+
+  it('decodes a transfer correctly', () => {
+    const functionSignature = `0xa9059cbb`;
```
iamacook, line 13: Nit: let's move this to a const as it's used multiple times here.

src/domain/relay/entities/contract-function.entity.ts
```diff
@@ -0,0 +1,56 @@
...
+): B {
+  const { functionName, args } = decodeFunctionData({
+    abi,
+    data,
+  });
+
+  if (contractFunction.name != functionName || !args)
+    throw Error(`Provided data is not a call to ${contractFunction.name}`);
+
+  return contractFunction.mapper(args);
```
iamacook, line 36: Mapping the arguments for every contract method will make each `ContractHelper` very "boilerplatey".

I think generalising `decode` to accept the method name and throw/return the `args` accordingly would suffice. It would simplify `ContractHelper`s for those which we need to support multiple methods, e.g. the Safe.

Let's discuss this on a call when we get the chance.

src/config/entities/configuration.ts
fmrsabino: One reason why I wrote with `${5}` was that it allows you to express numeric types using TS (one example from this file is that we can have `${5_000}` which would not work if it was a string (`5_000`).

I kept it this way also for consistency but I have no strong opinion on it and I can use a simple string as suggested 👍

src/domain/relay/contracts/safe-contract.helper.ts
fmrsabino: From the current implementation of the Safe Gelato Relay, we decode the `execTransaction` call but not the others. Support for the other functions is limited to checking if it's a call to the Safe Contract (from the `data` of the `execTransaction` call). This is already part of the PR: https://github.com/safe-global/safe-client-gateway/pull/680/files#diff-0fbe9bd32fe04ec35fc1fab18a5bf2db84a4c21877551de1311be99dc07b258fR72

iamacook: I see, oversight on my behalf - apologies. Will resolve this!

src/domain/relay/contracts/safe-contract.helper.spec.ts
```diff
@@ -0,0 +1,154 @@
...
+    }).toThrow();
+  });
+
+  it('isCall returns true for a execTransaction call', () => {
+    const callData = getExecTransactionCallData();
+
+    const actual = target.isCall(callData);
+
+    expect(actual).toBe(true);
+  });
```
iamacook, line 107: I think we should add coverage for Safe contract calls as well.

src/domain/relay/entities/contract-function.entity.ts
fmrsabino: One motivation for this approach was that `viem` returns `unknown[]` as a type of the function parameters thus the callers need to know that index 0 represents the argument with a given "name".

The goal of the abstraction was to return a structured object for the arguments for contracts. But I do agree that mapping every contract call can be excessive.

iamacook: Considering the point you made about only requiring `execTransaction` we could admittedly leave it as is for now but it will not "translate" well with interface changes.

Type safety is a valid concern. It doesn't seem like `typechain` supports `viem`, should we want to automatically generate them. It's a tough call - let's discuss it later today.

fmrsabino: > Considering the point you made about only requiring `execTransaction` we could admittedly leave it as is for now but it will not "translate" well with interface changes.
> 
> Type safety is a valid concern. It doesn't seem like `typechain` supports `viem`, should we want to automatically generate them. It's a tough call - let's discuss it later today.

I would say that this is ok: if the interface changes, this should break. The way it is implemented right now makes it compatible with any contract that has `execTransaction` as a function name, `to` as the 1st argument, `value` as the 2nd argument and `data` as the 3rd argument – we ignore the remaining arguments and their position.

The current `safe-contract.helper.spec.ts` instantiates a `SafeContractHelper` so the contracts' SDK will be used in the tests and they will fail if the SDK starts returning a different contract.

src/domain/relay/contracts/safe-contract.helper.spec.ts
fmrsabino: Done in [bdcbda1](https://github.com/safe-global/safe-client-gateway/pull/680/commits/bdcbda1c2d7e8987b4bd3d5e4ccd3cfdb28610dd)

src/domain/relay/limit-addresses.mapper.ts
fmrsabino: Done [4f16ddf](https://github.com/safe-global/safe-client-gateway/pull/680/commits/4f16ddf9e78c9526f8f057773493beebcc2a9c67).

But this log was triggered if an `execTransaction` didn't have an ERC20 transfer 😉

src/domain/relay/contracts/safe-contract.helper.spec.ts
fmrsabino: Done in [831f2e9](https://github.com/safe-global/safe-client-gateway/pull/680/commits/831f2e9947aaecddfaff2fb97d987b3378f36ba3).

It does not cover all the Safe Contract functions but the more commonly used ones though.

package.json
fmrsabino: I removed the dependency since we don't strip the unused assets of this package on the final image. Since the scope is limited to the `transfer` call I decided to limit its scope (i.e.: the ABI only has `transfer`). [532f5ad](https://github.com/safe-global/safe-client-gateway/pull/680/commits/532f5ad1ede95d7d1e9ce70f2334db02ec293e1e)

iamacook (approved):

PR #695, opened 2023-09-18
Use object literal for CacheFirstDataSource.get()

fmrsabino:
Uses an object literal for the arguments of `CacheFirstDataSource.get()`. This makes the value association from the caller perspective more explicit.


hectorgomezv (approved):
Great job! 🚀

PR #706, opened 2023-09-21
Add .prettierignore, apply formatting to project root

hectorgomezv:
**Context**
This is a follow-up to this comment: https://github.com/safe-global/safe-client-gateway/pull/703#discussion_r1330427264

**Motivation**
This PR aims to increase the formatting targets to include `yaml`, `json`, `md`, etc. files to the auto-formating feature provided by Prettier.

Prettier takes `.gitignore` as source for omitting files, so only `./yarn/*` files were included in `.prettierignore` file, in order to avoid parsing  Yarn's `.cjs` and other files that are being tracked by Git.


.prettierignore
```diff
@@ -0,0 +1 @@
+/.yarn
```
fmrsabino, line 1: I think we should add build outputs in addition to the `.yarn` folder. Some of the outputs would be in HTML (coverage). Same for code-related files (`dist`).

```suggestion
/.yarn
assets
coverage
dist
```

hectorgomezv: We can include them also, just to be sure they never get prettified, but as I commented in the description Prettier ignores the files and folders already present in `.gitignore`, so `coverage` and `dist` are covered.

In the case of `assets` I think it gets ignored because the extensions are not supported by Prettier (at least they aren't in the output when running `prettier --check .`)

fmrsabino: After coordinating internally it seems that `--ignore-path <path>` has `.gitignore` and `.prettierignore` as a default so we can use `.prettierignore` right now as an extension to it. We need to keep in mind if we change `.gitgnore` that we need to re-evaluate how the Prettier rules apply to it now.

PR #731, opened 2023-10-05
Log successful transaction proposal requests

fmrsabino:
- Emits an `INFO` log with the type of `propose-transaction-success` when a transaction is successfully proposed to the Safe Transaction Service.
- The log contains the Safe Address and Safe Transaction Hash of said transaction.
- Moves the `safe_app_user_agent` logging from the `RouteLoggerInterceptor` to the `RequestScopedLoggingService`. This means that the `safe_app_user_agent` will now part of the logs generated by the `RequestScopedLoggingService`.

### DIFF
--- a/src/logging/logging.service.spec.ts
+++ b/src/logging/logging.service.spec.ts
@@ -6,6 +6,7 @@ import { IConfigurationService } from '@/config/configuration.service.interface'
 
 const mockClsService = jest.mocked({
   getId: jest.fn(),
+  get: jest.fn(),
 } as unknown as ClsService);
 
 const mockLogger = {
--- a/src/app.module.spec.ts
+++ b/src/app.module.spec.ts
@@ -1,0 +1,55 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { AppModule, configurationModule } from './app.module';
+import * as request from 'supertest';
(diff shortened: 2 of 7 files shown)


src/logging/logging.service.spec.ts
```diff
@@ -6,6 +6,7 @@ import { IConfigurationService } from '@/config/configuration.service.interface'
 
 const mockClsService = jest.mocked({
```
hectorgomezv, line 7: Should we assert the logger gets called with a payload including the `safe_app_user_agent` property in the tests on this file? 

Debugging the tests it seems it's currently being called with an object including `"safe_app_user_agent": undefined`, so we need to mock the call  `mockClsService.get` as it's done for `mockClsService.getId` if we want to test it.

hectorgomezv (commented):
LGTM, just commenting out a detail.

src/logging/logging.service.spec.ts
fmrsabino: Good catch! Done in [dd29079](https://github.com/safe-global/safe-client-gateway/pull/731/commits/dd29079d9d7903a021181e430f2b63f5507216c6)

hectorgomezv (approved):
🚀🚀

PR #732, opened 2023-10-05
Add external prices provider support

hectorgomezv:


src/config/entities/__tests__/configuration.ts
```diff
@@ -33,6 +34,27 @@ export default (): ReturnType<typeof configuration> => ({
     level: 'debug',
     silent: process.env.LOG_SILENT?.toLowerCase() === 'true',
   },
+  prices: {
+    baseUri:
+      process.env.PRICES_PROVIDER_API_BASE_URI ||
+      'https://api.coingecko.com/api/v3',
+    apiKey: process.env.PRICES_PROVIDER_API_KEY,
+    pricesTtlSeconds: parseInt(process.env.PRICES_TTL_SECONDS ?? `${300}`),
+    chains: {
```
hectorgomezv, line 43: This Coingecko-specific chain configuration data could be externalized to the Safe Config Service. However, this can be done in a separate PR, because it also depends on whether we plan to continue to use Coingecko for the long term.

src/config/entities/configuration.ts
```diff
@@ -47,6 +49,27 @@ export default () => ({
     level: process.env.LOG_LEVEL || 'debug',
     silent: process.env.LOG_SILENT?.toLowerCase() === 'true',
   },
+  prices: {
+    baseUri:
+      process.env.PRICES_PROVIDER_API_BASE_URI ||
+      'https://api.coingecko.com/api/v3',
+    apiKey: process.env.PRICES_PROVIDER_API_KEY,
+    pricesTtlSeconds: parseInt(process.env.PRICES_TTL_SECONDS ?? `${300}`),
+    chains: {
```
hectorgomezv, line 58: This Coingecko-specific chain configuration data could be externalized to the Safe Config Service. However, this can be done in a separate PR, because it also depends on whether we plan to continue to use Coingecko for the long term.

src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -139,7 +139,7 @@ describe('TransactionApi', () => {
 
       await service.clearLocalBalances(safeAddress);
 
-      expect(mockCacheService.deleteByKey).toBeCalledTimes(1);
+      expect(mockCacheService.deleteByKey).toBeCalledTimes(2);
```
hectorgomezv, line 142: `clearLocalBalances` now clears both `Balances` (Transaction Service balances including fiat amounts) and `SimpleBalances` (Transaction Service balances not including fiat amounts).

src/domain/prices/prices.repository.ts
```diff
@@ -0,0 +1,63 @@
...
+import {
+  ILoggingService,
+  LoggingService,
+} from '../../logging/logging.interface';
+import { IPricesApi } from '../interfaces/prices-api.interface';
+import { IPricesRepository } from './prices.repository.interface';
+import { isArray } from 'lodash';
+
+@Injectable()
+export class PricesRepository implements IPricesRepository {
```
hectorgomezv, line 11: In this class, I'm assuming we want to fail back to 0 (and log the error) if any error happens when getting the price. Other options considered when failing:
- Returning `null` for the price: Doesn't seem appropriate since the clients may not be handling it.
- Omitting the failed token in the balances response: doesn't seem appropriate either. 
- Throwing an error to the client: doesn't seem appropriate to fail completely when a specific price is not reachable.

src/routes/balances/balances.service.ts
```diff
@@ -106,12 +126,130 @@ export class BalancesService {
...
+      .filter((b) => b.fiatBalance !== null)
+      .reduce((acc, b) => acc + Number(b.fiatBalance), 0);
+
+    return <Balances>{
+      fiatTotal: this.getNumberString(fiatTotal),
+      items: orderBy(balances, 'fiatBalance', 'desc'),
+    };
+  }
+
+  private async _mapSimpleBalance(
```
hectorgomezv, line 155: This function contains code that is mostly a duplicate of the previously existing`mapBalance`, except for minor details. I opted to include an almost-duplicated function instead of merging it with the original one, so after the feature flag cleaning, it can be just deleted/renamed.

src/datasources/prices-api/prices-api.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      );
+    }
+  }
+
+  private _mapProviderError(error: any): string {
+    const errorCode = error?.status?.error_code;
+    return errorCode ? ` [status: ${errorCode}]` : '';
+  }
+
+  private _buildNetworkRequest(): NetworkRequest | undefined {
```
hectorgomezv, line 113: This function only sets the authentication headers when the `apiKey` is present (and therefore a paid subscription is being used). 

I was wondering if we could include a `timeout` parameter also into these Coingecko calls, just to prevent our backend from waiting too much for the response. Waiting for a price for a long time seems unacceptable, what do you think @fmrsabino?

src/domain/interfaces/prices-api.interface.ts
```diff
@@ -0,0 +1,18 @@
+import { AssetPrice } from '../prices/entities/asset-price.entity';
+
+export const IPricesApi = Symbol('IPricesApi');
+
+export interface IPricesApi {
+  getNativeCoinPrice(args: {
+    nativeCoinId: string;
```
fmrsabino, line 7: I would change `nativeCoinId` to be the `chainId` instead. Here we are "leaking" the implementation details on how Coingecko works to the rest of the application.

If Coingecko is not used as a price provider this function might not make sense anymore (in the new provider).

src/datasources/prices-api/prices-api.service.ts
```diff
@@ -0,0 +1,118 @@
...
+        `Error getting Fiat Codes from prices provider${this._mapProviderError(
+          error,
+        )}`,
+      );
+    }
+  }
+
+  private _mapProviderError(error: any): string {
+    const errorCode = error?.status?.error_code;
+    return errorCode ? ` [status: ${errorCode}]` : '';
```
fmrsabino, line 110: What is the main reason for mapping the error code like this? The `DataSourceError` also has a `code` property which is not being used. Have you considered setting it also?

src/datasources/cache/cache.router.ts
```diff
@@ -403,4 +432,30 @@ export class CacheRouter {
   static getExchangeRatesCacheDir(): CacheDir {
     return new CacheDir(CacheRouter.EXCHANGE_RATES_KEY, '');
   }
+
+  static getNativeCoinPriceCacheDir(args: {
+    nativeCoinId: string;
+    fiatCode: string;
+    tokenAddress?: string;
```
fmrsabino, line 439: This argument can be removed.

src/datasources/cache/cache.router.ts
```diff
@@ -403,4 +432,30 @@ export class CacheRouter {
...
+    return new CacheDir(
+      `${args.nativeCoinId}_${CacheRouter.NATIVE_COIN_PRICE_KEY}`,
+      args.fiatCode,
+    );
+  }
+
+  static getTokenPriceCacheDir(args: {
+    chainName: string;
+    fiatCode: string;
+    tokenAddress?: string;
```
fmrsabino, line 450: Can you get a cached token price when `tokenAddress` is `undefined`?

src/domain/prices/prices.repository.ts
```diff
@@ -0,0 +1,63 @@
+import { Inject, Injectable } from '@nestjs/common';
+import {
+  ILoggingService,
+  LoggingService,
+} from '../../logging/logging.interface';
```
fmrsabino, line 5: We can shorten this import.

src/domain/prices/entities/asset-price.entity.ts
```diff
@@ -0,0 +1,3 @@
+export interface AssetPrice {
+  [x: string]: Record<string, number>;
```
fmrsabino, line 2: I'd give a more descriptive name than `x`. E.g.: `assetName`.

src/domain/prices/prices.repository.ts
```diff
@@ -0,0 +1,63 @@
...
+  async getNativeCoinPrice(args: {
+    nativeCoinId: string;
+    fiatCode: string;
+  }): Promise<number> {
+    try {
+      const result = await this.pricesApi.getNativeCoinPrice(args);
+      const { nativeCoinId, fiatCode } = args;
+      const nativeCoinPrice = result?.[nativeCoinId]?.[fiatCode];
+      if (!nativeCoinPrice) {
+        this.loggingService.error(`Got an invalid ${nativeCoinId} price`);
```
fmrsabino, line 26: This error does not seem correct. Would our Fiat Price provider return an "invalid" price for a token/native coin if it exists?

Or is it more that the token/native coin does not exist, or the conversion to that specific code does not exist?

src/domain/prices/prices.repository.ts
```diff
@@ -0,0 +1,63 @@
...
+        return 0;
```
fmrsabino, line 27: I would throw an error in these cases. `0` should be seen as a valid value from the domain's perspective.

Mapping the error to `0` can be done on the route level because there we have the contract with the clients (API). But this is not the case on the domain layer and it's better to build a more meaningful result at this level.

src/domain/prices/prices.repository.ts
```diff
@@ -0,0 +1,63 @@
...
+      return tokenPrice;
+    } catch (err) {
+      this.loggingService.error(err.message);
+      return 0;
+    }
+  }
+
+  async getFiatCodes(): Promise<string[]> {
+    const fiatCodes = await this.pricesApi.getFiatCodes();
+    if (!isArray(fiatCodes)) {
```
fmrsabino, line 58: Wouldn't AJV be useful to tackle the validation of these entities?

src/domain/prices/entities/asset-price.entity.ts
```diff
@@ -0,0 +1,3 @@
+export interface AssetPrice {
```
fmrsabino, line 1: What is the reason to not have an AJV schema for the `AssetPrice`?

src/domain/balances/balances.repository.interface.ts
```diff
@@ -14,6 +16,17 @@ export interface IBalancesRepository {
     excludeSpam?: boolean;
   }): Promise<Balance[]>;
 
+  /**
+   * Gets the collection of {@link Balance} associated with {@link safeAddress}
```
fmrsabino, line 20:
```
Gets the collection of {@link SimpleBalance}
```

fmrsabino (changes requested):
Initial review. I've started with the `data` and `domain` layers.

src/datasources/prices-api/prices-api.service.ts
fmrsabino: We should set a timeout for these requests. For now I'd use `httpClient.requestTimeout` from the configuration.

src/domain/prices/prices.repository.ts
fmrsabino: Since we are in the `domain` layer I don't agree with this exactly. I've expanded more on why in https://github.com/safe-global/safe-client-gateway/pull/732#discussion_r1350075295

src/domain/prices/prices.repository.ts
hectorgomezv: It could be both cases. The provider is not helping here because the status codes and the payloads are not very helpful. The status code is always 200, no matter the kind of absence (token doesn't exist, or fiat code is not supported). Examples:
- Token doesn't exist (made up address): https://api.coingecko.com/api/v3/simple/token_price/ethereum?contract_addresses=0xa0b86991c6218b36c1d19d4b2e9eb0ce3606eb48&vs_currencies=eur
- Native coin doesn't exist: https://api.coingecko.com/api/v3/simple/price?ids=foo&vs_currencies=eur
- Token exists but fiat code is not supported: https://api.coingecko.com/api/v3/simple/token_price/ethereum?contract_addresses=0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48&vs_currencies=foo

So for the sake of simplicity, I'm just inspecting the response and throwing an error if it doesn't contain the price. Of course, we could implement some logic to distinguish between those scenarios and throw ad-hoc errors. That could make the code a bit more complex but I'd be happy to add this if you think it's valuable.

src/domain/prices/prices.repository.ts
hectorgomezv: Yes, good point. This should be just a string array, but for consistency we should declare a schema and use it. Thanks!

src/domain/interfaces/prices-api.interface.ts
hectorgomezv: I agree, good point. I'll move the `chainId-to-nativeCoinId` mapping (i.e. get the `nativeCoinId` from the app configuration) to the Coingecko-specific classes 👍

src/datasources/prices-api/prices-api.service.ts
hectorgomezv: Initially I thought `error_code` was only available in the response payload object:
```
{
    "status": {
        "error_code": 429,
        "error_message": "You've exceeded the Rate Limit. Please visit https://www.coingecko.com/en/api/pricing to subscribe to our API plans for higher rate limits."
    }
}
```

But it seems the status code is available as well, so I'm going to try to get it from there, thanks 👍

src/datasources/cache/cache.router.ts
hectorgomezv: Good point. It shouldn't be possible. I'll make it mandatory here 👍

src/domain/prices/prices.repository.ts
hectorgomezv: It could be both cases. The provider is not helping here because the status codes and the payloads are not very helpful. The status code is always 200, no matter the kind of absence (token doesn't exist, or fiat code is not supported). Examples:
- Token doesn't exist (made up address): https://api.coingecko.com/api/v3/simple/token_price/ethereum?contract_addresses=0xa0b86991c6218b36c1d19d4b2e9eb0ce3606eb48&vs_currencies=eur
- Native coin doesn't exist: https://api.coingecko.com/api/v3/simple/price?ids=foo&vs_currencies=eur
- Token exists but fiat code is not supported: https://api.coingecko.com/api/v3/simple/token_price/ethereum?contract_addresses=0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48&vs_currencies=foo

So for the sake of simplicity, I'm just inspecting the response and throwing an error if it doesn't contain the price. Of course, we could implement some logic to distinguish between those scenarios and throw ad-hoc errors. That could make the code a bit more complex but I'd be happy to add this if you think it's valuable.

src/domain/prices/entities/asset-price.entity.ts
hectorgomezv: Yeah, we can include one, thanks.

src/datasources/prices-api/prices-api.service.ts
fmrsabino: It's also fine if the error code doesn't make sense to be set in the `DataSourceError`. The main goal of this error is to abstract away all the `datasource` related errors. If this error is not caught in the service then the `DataSourceErrorFilter` will return the following to the clients:

```javascript
{
  "code": <int>,
  "message": <string>
}
```

It's important to be aware that we should not leak the internals on why the request failed – in this case the Rate Limit exceeded on the CGW side should not be returned to the clients.

src/domain/prices/prices.repository.ts
fmrsabino: In that case and to start simple I would log the conditions on which this error happened. You can use a structure to help you in this case.

Something like:
```typescript
this.loggingService.error({
  type: 'invalid_native_coin_price',
  native_coin_id: nativeCoinId,
  fiat_code: fiatCode,
});
```

You can do the same for `getTokenPrice`

Also probably to use the `warn` log level here as an indication that something unexpected happened (as you have it in `getTokenPrice`). Or did you want to distinguish these two?

src/domain/balances/balances.repository.ts
```diff
@@ -26,7 +32,26 @@ export class BalancesRepository implements IBalancesRepository {
       trusted: args.trusted,
       excludeSpam: args.excludeSpam,
     });
-    return balances.map((balance) => this.validator.validate(balance));
+    return balances.map((balance) => this.balancesValidator.validate(balance));
+  }
+
+  async getSimpleBalances(args: {
```
fmrsabino, line 38: Adding new functionality to `IBalancesRepository` for the migration does not seem to be the best way forward.

From the domain perspective, the `BalancesRepository` was already returning balances and fiat prices. Therefore I would keep the same assumption on the `domain` layer resulting in no changes in the `routes` layer.

What do you think?

fmrsabino (changes requested):

src/domain/balances/balances.repository.ts
hectorgomezv: Not sure I follow, do you mean returning balances + prices (`Balance` entities) from `BalancesRepository` instead of having these two functions (one returning `Balance` entities and the other returning `SimpleBalance` entities)?

fmrsabino: Exactly 👍 The routes should not change because we get the pricing info from a different provider now. For the routes the `Balance` interface should still be followed.

src/datasources/prices-api/prices-api.module.ts
```diff
@@ -0,0 +1,13 @@
+import { Global, Module } from '@nestjs/common';
+import { CacheFirstDataSourceModule } from '../cache/cache.first.data.source.module';
+import { HttpErrorFactory } from '../errors/http-error-factory';
+import { PricesApi } from './prices-api.service';
+import { IPricesApi } from '@/domain/interfaces/prices-api.interface';
+
+@Global()
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/datasources/prices-api/prices-api.service.ts
```diff
@@ -0,0 +1,118 @@
+import { Inject, Injectable } from '@nestjs/common';
+import { IConfigurationService } from '../../config/configuration.service.interface';
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [d4cf6711](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/datasources/prices-api/prices-api.service.ts
```diff
@@ -0,0 +1,118 @@
...
+import { IPricesApi } from '../../domain/interfaces/prices-api.interface';
+import { AssetPrice } from '../../domain/prices/entities/asset-price.entity';
+import { CacheFirstDataSource } from '../cache/cache.first.data.source';
+import { CacheRouter } from '../cache/cache.router';
+import { DataSourceError } from '../../domain/errors/data-source.error';
+import { NetworkRequest } from '../network/entities/network.request.entity';
+
+@Injectable()
+export class PricesApi implements IPricesApi {
+  private static readonly pricesProviderHeader: string = 'x-cg-pro-api-key';
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [d4cf6711](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/datasources/prices-api/prices-api.service.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+  async getTokenPrice(args: {
+    chainName: string;
+    tokenAddress: string;
+    fiatCode: string;
+  }): Promise<AssetPrice> {
+    try {
+      const cacheDir = CacheRouter.getTokenPriceCacheDir(args);
+      const url = `${this.baseUrl}/simple/token_price/${args.chainName}?contract_addresses=${args.tokenAddress}&vs_currencies=${args.fiatCode}`;
```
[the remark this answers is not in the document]
hectorgomezv: Done in [d4cf6711](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/datasources/prices-api/prices-api.service.ts
```diff
@@ -0,0 +1,118 @@
...
+      const cacheDir = CacheRouter.getPriceFiatCodesCacheDir();
+      const url = `${this.baseUrl}/simple/supported_vs_currencies`;
+      const result: string[] = await this.dataSource.get({
+        cacheDir,
+        url,
+        networkRequest: this._buildNetworkRequest(),
+        notFoundExpireTimeSeconds: this.defaultNotFoundExpirationTimeSeconds,
+        expireTimeSeconds: this.defaultExpirationTimeInSeconds,
+      });
+      return result;
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [d4cf6711](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/datasources/cache/cache.router.ts
hectorgomezv: Fixed in [d4cf6711](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/datasources/cache/cache.router.ts
hectorgomezv: Fixed in [d4cf6711](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/domain/balances/balances.repository.interface.ts
hectorgomezv: Fixed in [d4cf6711](https://github.com/safe-global/safe-client-gateway/commit/d4cf6711d2b6f399acd83fd4be5b8ff10e1b778d)

src/domain/balances/balances.repository.ts
hectorgomezv: I agree, but that would imply calculating ' fiatBalance` and `fiatConversion` into the repository, right? So we would need to inject `PriceRepository` into `BalancesRepository`, or remove `PriceRepository` and integrate its functions into `BalancesRepository`, therefore calling `CoinegeckoApi` from there. 

I'm not against it. In fact, I think it makes perfect sense, I'm just aligning expectations.

hectorgomezv: Yes, totally get your point. I was structuring this more like a migration (this feature is behind a FF), so in the end the routes would expect to have balances without fiat, and keep doing calculations on the routes layer.

But I agree it makes more sense to move these calculations to the domain layer (to the repository in this case), and by doing so, we can also abstract the routes from the balances composition, so we would benefit from that also. I will change this, thanks!

src/domain/prices/prices.repository.ts
hectorgomezv: My initial attempt was to distinguish between token price absence and native coin price absence and assign a `warn` level to the former and an `error` level to the latter, which seems more critical. But let's start simple as you said, and assign a `warn` level to both.

I'm going to set prices to`null` for each item whose price cannot be retrieved, so the routes layer can map to `0` to fulfill the contract with the clients.

hectorgomezv: Changed in [17040f3](https://github.com/safe-global/safe-client-gateway/pull/732/commits/17040f3f07ad63a3682ce6b665dc451dfae545a2)

src/domain/prices/prices.repository.ts
hectorgomezv: I've changed this to return `null` from the domain layer when an error happens in [17040f3](https://github.com/safe-global/safe-client-gateway/pull/732/commits/17040f3f07ad63a3682ce6b665dc451dfae545a2)

src/domain/prices/prices.repository.ts
hectorgomezv: Done [17040f3](https://github.com/safe-global/safe-client-gateway/pull/732/commits/17040f3f07ad63a3682ce6b665dc451dfae545a2)

src/domain/prices/entities/asset-price.entity.ts
hectorgomezv: Done [17040f3](https://github.com/safe-global/safe-client-gateway/pull/732/commits/17040f3f07ad63a3682ce6b665dc451dfae545a2)

src/domain/balances/balances.repository.ts
hectorgomezv: I've changed the approach in [17040f3](https://github.com/safe-global/safe-client-gateway/pull/732/commits/17040f3f07ad63a3682ce6b665dc451dfae545a2) based on your suggestion.

Now the domain layer only exposes a function to get `Balances`:
```javascript
  getBalances(args: {
    chainId: string;
    safeAddress: string;
    trusted?: boolean;
    excludeSpam?: boolean;
  }): Promise<Balance[]>;
``` 
Some _ code pseudo-duplication_ in the routes layer service `BalancesService` (between `getBalances/_getFromSimpleBalances` and `mapBalance/_mapSimpleBalance`) still exists. I tried to not move completely the existing implementation to `BalancesRepository` (bringing the `IExchangeRepository` injection there and so on). This is intended to simplify the PR. I'm also considering this state (having both Coingecko and the Exchange API) should be transitory. 

If you think we should move the Exchange API to the repository also, to simplify the routes layer further (even if this should be a transitory state), I'd suggest we do it separately in a different PR.

src/routes/balances/balances.service.ts
```diff
@@ -37,6 +43,10 @@ export class BalancesService {
     trusted: boolean;
     excludeSpam: boolean;
   }): Promise<Balances> {
+    if (this.pricesProviderChainIds.includes(args.chainId)) {
+      return this._getFromSimpleBalances(args);
+    }
```
fmrsabino, line 48: What is the main reason for having a decision point on the `routes` layer? I don't think this addressed directly the concern regarding how the balances are abstracted 🤔 

If the domain entities (we still get a `Balance` object) why did the logic on the routes had to change?

The lines below this statement should still be valid right? More specifically:
- Mapping the balances to the Route entities
- Getting the total fiat from the mapped balances
- Sorting the balances

hectorgomezv: I tried to address this in my previous comment (https://github.com/safe-global/safe-client-gateway/pull/732#discussion_r1350906338): the decision point still in the routes layer because otherwise we would need to introduce some changes in the previously existing code:
- `IExchangeRepository` injection would be moved to `BalancesRepository`, so we can retrieve/apply fiat conversion rates there.
- (derived from above point) Some logic that is now part of the mapping function (`usdToFiatRate` multiplication, for example), should be moved to the repository too, because it depends on whether we are using the Exchange API or Coingecko API.

So I’d say it’s not possible to keep all the lines you mentioned. I opted for the strategy of having separated functions and an if condition for routing, because the feature flag should be fully applied soon and the Exchange-related function would be removed easily. So perhaps this strategy results in redundant code for a while, but would also be easy to clean later, after the FF is fully applied.

Having said that, I’m open to move all the logic to `BalancesRepository` as you suggested, even if it implies modifying existing code. I leave this up to you, what do you think? Would you push all the logic into the repository in this PR? :slightly_smiling_face: (

hectorgomezv: After a quick sync, I've applied your suggestion in [35029ea](https://github.com/safe-global/safe-client-gateway/pull/732/commits/35029ea40c3cf5243b32eeeec1f1b296b9db2e6a), thanks!

src/domain/prices/prices.repository.ts
```diff
@@ -8,7 +7,6 @@ import { FiatCodesValidator } from './fiat-codes.validator';
 @Injectable()
 export class PricesRepository implements IPricesRepository {
   constructor(
-    @Inject(LoggingService) private readonly loggingService: ILoggingService,
```
hectorgomezv: Just removing an unused dependency here.

fmrsabino (approved):

PR #735, opened 2023-10-09
Set default values for trusted and exclude_spam params

### LINKED ISSUE #734: Adjust getBalances default values for trusted and excludeSpam query parameters
Endpoint: `chains/:chainId/safes/:safeAddress/balances/:fiatCode`

**Current behavior** 
- Both `trusted` and `excludeSpam` have not default values associated.

**Expected behavior**
- `trusted` should default to `false`.
- `excludeSpam` should default to `true`.

Reference: https://github.com/5afe/safe-client-gateway/blob/9a016db82fe7038db73fb41981dbd15bc89770f5/src/routes/balances/routes.rs#L51

fmrsabino:
Closes #734 

- Changes the default values of the `trusted` and `exclude_spam` query parameters if they are not provided in `GET /v1/chains/:chainId/safes/:safeAddress/balances/:fiatCode`. This corresponds to the expected API spec from the Rust client.
- The `trusted` query parameter should have a default value of `false` if none is provided.
- The `exclude_spam` query parameter should have a value of `true` if none is provided.

### DIFF
--- a/src/routes/balances/balances.controller.ts
+++ b/src/routes/balances/balances.controller.ts
@@ -1,4 +1,10 @@
-import { Controller, Get, Param, Query } from '@nestjs/common';
+import {
+  Controller,
+  DefaultValuePipe,
+  Get,
+  Param,
+  Query,
+} from '@nestjs/common';
 import { BalancesService } from './balances.service';
 import { Balances } from './entities/balances.entity';
 import { ApiOkResponse, ApiTags } from '@nestjs/swagger';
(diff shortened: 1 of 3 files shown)


src/routes/balances/balances.controller.ts
```diff
@@ -17,8 +23,8 @@ export class BalancesController {
     @Param('chainId') chainId: string,
     @Param('safeAddress') safeAddress: string,
     @Param('fiatCode') fiatCode: string,
-    @Query('trusted') trusted?: boolean,
-    @Query('exclude_spam') excludeSpam?: boolean,
+    @Query('trusted', new DefaultValuePipe(false)) trusted: boolean,
```
hectorgomezv, line 26: I think this change is totally OK, but I was wondering... should we enforce parsing to boolean at this level, as it's [done here](https://github.com/safe-global/safe-client-gateway/blob/8c4dd24d1ac33e12da8deeff49426cd7a74ff66b/src/routes/transactions/transactions.controller.ts#L194) by using `ParseBoolPipe`? 

It's also present in the example for providing default values in [the Nest docs](https://docs.nestjs.com/pipes#providing-defaults).

This is probably something we could address in a separate PR if we think it's needed, as it affects other controllers also. What do you think?

hectorgomezv (approved):

src/routes/balances/balances.controller.ts
fmrsabino: Good point. I think we can add the parsing across the controllers that we have 👍

PR #736, opened 2023-10-09
Set default query param values in CollectiblesController

fmrsabino:
- Changes the default values of the `trusted` and `exclude_spam` query parameters if they are not provided in `GET /v2/chains/:chainId/safes/:safeAddress/collectibles`. This corresponds to the expected API spec from the Rust client.
- The `trusted` query parameter should have a default value of `false` if none is provided.
- The `exclude_spam` query parameter should have a value of `true` if none is provided.


hectorgomezv (approved):
🚀

PR #747, opened 2023-10-11
Use builtin parsers for boolean query params

hectorgomezv:
- Adds `ParseBoolPipe` to validate boolean query params at controllers.

### DIFF
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -162,8 +162,8 @@ describe('Balances Controller (Unit)', () => {
       // trusted and exclude_spam params are passed
       expect(networkService.get.mock.calls[1][1]).toStrictEqual({
         params: {
-          trusted: trusted.toString(),
-          exclude_spam: excludeSpam.toString(),
+          trusted,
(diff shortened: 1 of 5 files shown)


fmrsabino (commented):
We can also change it in https://github.com/safe-global/safe-client-gateway/blob/b38435dd6851e8c332e8bd12309ec70d0351c137/src/routes/transactions/transactions.controller.ts#L74

To something like:

```typescript
@Query('executed', new ParseBoolPipe({ optional: true })) executed?: boolean
```

hectorgomezv:
> We can also change it in
> 
> https://github.com/safe-global/safe-client-gateway/blob/b38435dd6851e8c332e8bd12309ec70d0351c137/src/routes/transactions/transactions.controller.ts#L74
> 
> To something like:
> 
> ```ts
> @Query('executed', new ParseBoolPipe({ optional: true })) executed?: boolean
> ```

Thanks! I haven't seen the `optional` option, good catch. Changed in [cb37b99](https://github.com/safe-global/safe-client-gateway/pull/747/commits/cb37b99e3806c188421fd177930bd840ebc0ff72)

fmrsabino (approved):

PR #749, opened 2023-10-11
Change relative import paths to non-relative path aliases

hectorgomezv:
**Related to** #748 

- Changes relative import to non-relative imports using path aliases.


fmrsabino (commented):
Have you considered changing the relative imports to the actual folder?

For example having:

```typescript
import { ConfigurationModule } from '@/config/configuration.module';
```

instead of

```typescript
import { ConfigurationModule } from './configuration.module';
```

hectorgomezv:
> Have you considered changing the relative imports to the actual folder?
> 
> For example having:
> 
> ```ts
> import { ConfigurationModule } from '@/config/configuration.module';
> ```
> 
> instead of
> 
> ```ts
> import { ConfigurationModule } from './configuration.module';
> ```

I thought about it, but I opted for allowing relative imports for files that live in the same folder for brevity. I have no opinion on that, we could use non-relative imports everywhere for consistency, what do you think? Any preference?

fmrsabino:
> I thought about it, but I opted for allowing relative imports for files that live in the same folder for brevity. I have no opinion on that, we could use non-relative imports everywhere for consistency, what do you think? Any preference?

For consistency I would do it also for the non-relative imports everywhere. Seems to be a clearer rule to follow.

hectorgomezv:
> > I thought about it, but I opted for allowing relative imports for files that live in the same folder for brevity. I have no opinion on that, we could use non-relative imports everywhere for consistency, what do you think? Any preference?
> 
> For consistency I would do it also for the non-relative imports everywhere. Seems to be a clearer rule to follow.

Done in [f3dfbe0](https://github.com/safe-global/safe-client-gateway/pull/749/commits/f3dfbe0fdc1f21fb5c367a4f7051fa6872cfac17)

.vscode/settings.json
```diff
@@ -1,6 +1,6 @@
 {
   "editor.defaultFormatter": "esbenp.prettier-vscode",
   "[typescript]": {
-    "typescript.preferences.importModuleSpecifier": "relative"
+    "typescript.preferences.importModuleSpecifier": "non-relative"
```
fmrsabino, line 4: Is this change needed? It was added in https://github.com/safe-global/safe-client-gateway/commit/c21e2cf8774803e1aee915ef7a10ca0a5672c974

fmrsabino (approved):
Looks good. My suggestion is to do this by folder/layer next time to make it easier for review.

After squashing and merging I'd add the squashed commit to `.git-blame-ignore-revs` https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt

.vscode/settings.json
hectorgomezv: Sorry, a rebase from `main` left. 

Fixed, thanks!

PR #751, opened 2023-10-13
Add .git-blame-ignore-revs file

hectorgomezv:
- Adds `.git-blame-ignore-revs` containing the commit hash for the change in https://github.com/safe-global/safe-client-gateway/pull/749

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Change import statement paths to path aliases
+4729440761cfe608ab8935d782238ae7fe55343b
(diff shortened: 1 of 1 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Change import statement paths to path aliases
```
fmrsabino, line 1: To follow the recommendation by Github – https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

```suggestion
# .git-blame-ignore-revs
# Change import statement paths to path aliases
```

fmrsabino (approved):
🚀

fmrsabino (approved):

PR #752, opened 2023-10-16
Adjust chains native coins configuration

hectorgomezv:
### Changes
- Adjusts native coin IDs for some networks.
- Adds Goerli configuration.

### Context
`nativeCoin` references the coin associated with the address `0x0000000000000000000000000000000000000000` for each network.

Depending on the network, this address may reference `Ether`, or a network native token. This configuration was incorrect for some networks, that have `Ether` as their native coin. 

For this change, [the Transaction Service code](https://github.com/safe-global/safe-transaction-service/blob/f9e18f16d01c67beedc17b96de77f6444fe9a3d8/safe_transaction_service/tokens/services/price_service.py#L219) was revisited, as it contains logic to determine the links between each network and its native coin.


fmrsabino (approved):
I would say that having a validation on this configuration might also make sense given that these values should not change (part of ane external spec).

PR #753, opened 2023-10-16
Fix bad Coingecko API query param

hectorgomezv:
- Fixes the Coingecko query param to hold the ERC20 contract addresses when retrieving prices (bug introduced in [this commit](https://github.com/safe-global/safe-client-gateway/pull/732/commits/a916ab768d1f4d952fe8e94ee69b4a2b1cba5f90#diff-e0b76c0f8409c4ea42ea003122e6c67d26f42543119fb2bbd37ec90de07f5491))

### DIFF
--- a/src/datasources/prices-api/coingecko-api.service.ts
+++ b/src/datasources/prices-api/coingecko-api.service.ts
@@ -86,7 +86,7 @@ export class CoingeckoApi implements IPricesApi {
         networkRequest: {
           params: {
             vs_currencies: args.fiatCode,
-            contract_address: args.tokenAddress,
+            contract_addresses: args.tokenAddress,
           },
           ...(this.apiKey && {
             headers: {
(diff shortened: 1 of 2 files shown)


src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -86,7 +86,7 @@ export class CoingeckoApi implements IPricesApi {
         networkRequest: {
           params: {
             vs_currencies: args.fiatCode,
-            contract_address: args.tokenAddress,
+            contract_addresses: args.tokenAddress,
```
fmrsabino, line 89: The title of the PR is not correct in this case, right? The name of the query param was incorrectly set (not the header).

fmrsabino (approved):

src/datasources/prices-api/coingecko-api.service.ts
hectorgomezv: Oh, yes, right, I'm changing it, thanks!

PR #754, opened 2023-10-16
Add RootController with SwaggerUI redirection

hectorgomezv:
- Adds `RootController`, which is used to redirect the root path to the SwaggerUI webapp.

### DIFF
--- a/src/routes/root/root.controller.ts
+++ b/src/routes/root/root.controller.ts
@@ -1,0 +1,9 @@
+import { Controller, Get, Res } from '@nestjs/common';
+
+@Controller()
+export class RootController {
+  @Get()
+  getIndex(@Res() res): void {
+    return res.redirect('/index.html');
+  }
+}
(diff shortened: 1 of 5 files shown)


src/routes/root/root.controller.ts
```diff
@@ -0,0 +1,9 @@
+import { Controller, Get, Res } from '@nestjs/common';
+
+@Controller()
+export class RootController {
+  @Get()
+  getIndex(@Res() res): void {
+    return res.redirect('/index.html');
+  }
+}
```
fmrsabino, line 9: We should test this controller. More specifically that the redirect works. Something like:

```typescript
  .get(`/`)
  .expect(302)
  .expect((res) => {
    expect(res.get('location')).toBe('/index.html');
  });
```

fmrsabino (changes requested):

src/routes/root/root.controller.ts
hectorgomezv: Thanks! Added in [c868631](https://github.com/safe-global/safe-client-gateway/pull/754/commits/c8686310290608c2a5a5d1d747c7cf767d231e9d)

fmrsabino (approved):

PR #760, opened 2023-10-17
Add Postgres database integration

fmrsabino:
- Adds support for connecting to a Postgres database.
- This support is provided via `PostgresDatabaseModule`, which exposes an instance of `postgres.Sql` which represents a connection that can be used for performing queries.
- The `PostgresDatabaseHook` executed whenever the `PostgresDatabaseModule` is destroyed – it cancels any ongoing queries while giving a timeout of 5 seconds for pending queries to complete.
- Adds the `db` container setup in docker-compose.yml. The database contents are mapped to `<PROJ-DIR>/data/db` to allow data persistence across runs.
- The following environment variables can be set to change the configuration of the database instance:
  * `POSTGRES_DB` – the name of the default database. Defaults to `safe-client-gateway` if not set.
  * `POSTGRES_USER` – the database superuser name. Defaults to `postgres`.
  * `POSTGRES_PASSWORD` – the database superuser password. Defaults to `postgres`.
  * `POSTGRES_PORT` – the port where the database instance is running. Defaults to `5432`.

❗️The Database connection is currently optional. It will only result in an error if a query is made and there's no database server running.

## Scope of this Pull Request

The following changes are currently outside of the scope of this Pull Request:
- Initial schemas and tables (feature scope)
- Forward-only migration mechanism
- Query execution abstraction – the library being used to execute the SQL queries uses [Tagged Templates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#tagged_templates). Creating an abstraction around how these templates are evaluated was out of scope for this Pull Request and I believe that we can expose the SQL client directly on the `Datasource` layer (i.e.: not create an interface around this client for now).
- Github Action Postgres service – since there are no tests/use-cases currently depending on the database, it is not required.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,6 @@ lerna-debug.log*
 
 # env
 .env
+
+# Database mounted volume
+data
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -5,6 +5,18 @@ services:
       - ${REDIS_PORT-6379}:${REDIS_PORT-6379}
       - 8001:8001
 
+  db:
+    image: postgres:14.8-alpine
+    volumes:
+      - ./data/db:/var/lib/postgresql/data
+    environment:
+      POSTGRES_DB: ${POSTGRES_DB:-safe-client-gateway}
+      POSTGRES_USER: ${POSTGRES_USER:-postgres}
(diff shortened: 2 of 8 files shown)


hectorgomezv (approved):
Great job! 🚀

PR #764, opened 2023-10-18
Set safe version nullable

### LINKED ISSUE #629: Make Safe version nullable

hectorgomezv:
Closes #629 

This PR:
- Adds support for having `null` as Safes `version` attribute.
- Modifies the OpenAPI definition accordingly.
- Adapts the code to return `UNKNOWN` for the `implementationVersionState` field when `version` is `null`.


src/routes/safes/entities/safe-info.entity.ts
```diff
@@ -20,13 +20,13 @@ export class SafeState {
...
   @ApiProperty()
   readonly implementation: AddressInfo;
-  @ApiPropertyOptional()
+  @ApiPropertyOptional({ type: AddressInfo, nullable: true })
   readonly modules: AddressInfo[] | null;
-  @ApiPropertyOptional()
+  @ApiPropertyOptional({ type: AddressInfo, nullable: true })
   readonly fallbackHandler: AddressInfo | null;
-  @ApiPropertyOptional()
+  @ApiPropertyOptional({ type: AddressInfo, nullable: true })
```
fmrsabino, line 27: nitpick: I would create a separate commit for this and not squash it (if done in the same PR) or a separate PR. Consider atomic changes that we should revert if required.

fmrsabino (approved):

src/routes/safes/entities/safe-info.entity.ts
hectorgomezv: I agree, good point! I've left all the changes directly related to this issue in [2a001a3](https://github.com/safe-global/safe-client-gateway/pull/764/commits/2a001a35688205a3e7712d909525a01b25c04778), and I've created another PR for the rest: https://github.com/safe-global/safe-client-gateway/pull/765

src/routes/safes/safes.controller.ts
```diff
@@ -10,6 +10,7 @@ import { SafesService } from '@/routes/safes/safes.service';
 export class SafesController {
   constructor(private readonly service: SafesService) {}
 
+  @ApiOkResponse({ type: SafeState })
```
fmrsabino, line 13: I might have missed this 😬 

I added it in https://github.com/safe-global/safe-client-gateway/commit/1ada6bdd298179987f4fb0b8dfb65a809d6a6be5

Issue #767, opened 2023-10-19
[GET /balances/:fiatCode] Fiat code is not case-insensitive


hectorgomezv:
### Problem
Calls to `chains/:chainId/safes/:safeAddress/balances/:fiatCode` with an upper-case `fiatCode` produce incorrect balance amounts.

```
https://safe-client.staging.5afe.dev/v1/chains/1/safes/0x84443F61efc60D10DA9F9a2398980CD5748394BB/balances/USD
```
![Screenshot 2023-10-19 at 11 24 29](https://github.com/safe-global/safe-client-gateway/assets/21182069/6a6de246-3eba-4165-80c0-3245f32e8c90)

### Expected behavior
The balances endpoint should accept both upper-case and lower-case `fiatCode`.

PR #768, opened 2023-10-19
Make fiatCode case-insensitive on /balances endpoint

hectorgomezv:
(Context: https://github.com/safe-global/safe-client-gateway/issues/767)
- Parses input `fiatCode`, converting it to lower case, so clients can use both upper or lower cases.

### DIFF
--- a/src/domain/prices/prices.repository.ts
+++ b/src/domain/prices/prices.repository.ts
@@ -21,7 +21,7 @@ export class PricesRepository implements IPricesRepository {
       nativeCoinId: args.nativeCoinId,
       fiatCode: lowerCaseFiatCode,
     });
-    const assetPrice = await this.assetPriceValidator.validate(result);
+    const assetPrice = this.assetPriceValidator.validate(result);
     return assetPrice?.[args.nativeCoinId]?.[lowerCaseFiatCode];
   }
 
@@ -37,7 +37,7 @@ export class PricesRepository implements IPricesRepository {
       tokenAddress: lowerCaseTokenAddress,
       fiatCode: lowerCaseFiatCode,
     });
-    const assetPrice = await this.assetPriceValidator.validate(result);
+    const assetPrice = this.assetPriceValidator.validate(result);
     return assetPrice?.[lowerCaseTokenAddress]?.[lowerCaseFiatCode];
   }
 
--- a/src/domain/prices/prices.repository.ts
+++ b/src/domain/prices/prices.repository.ts
@@ -16,10 +16,13 @@ export class PricesRepository implements IPricesRepository {
     nativeCoinId: string;
     fiatCode: string;
   }): Promise<number> {
-    const result = await this.coingeckoApi.getNativeCoinPrice(args);
+    const lowerCaseFiatCode = args.fiatCode.toLowerCase();
+    const result = await this.coingeckoApi.getNativeCoinPrice({
+      nativeCoinId: args.nativeCoinId,
+      fiatCode: lowerCaseFiatCode,
+    });
     const assetPrice = await this.assetPriceValidator.validate(result);
-    const { nativeCoinId, fiatCode } = args;
-    return assetPrice?.[nativeCoinId]?.[fiatCode];
+    return assetPrice?.[args.nativeCoinId]?.[lowerCaseFiatCode];
   }
 
   async getTokenPrice(args: {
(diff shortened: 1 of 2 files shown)


src/domain/prices/prices.repository.ts
```diff
@@ -16,21 +16,29 @@ export class PricesRepository implements IPricesRepository {
     nativeCoinId: string;
     fiatCode: string;
   }): Promise<number> {
-    const result = await this.coingeckoApi.getNativeCoinPrice(args);
+    const lowerCaseFiatCode = args.fiatCode.toLowerCase();
+    const result = await this.coingeckoApi.getNativeCoinPrice({
+      nativeCoinId: args.nativeCoinId,
+      fiatCode: lowerCaseFiatCode,
+    });
     const assetPrice = await this.assetPriceValidator.validate(result);
```
fmrsabino, line 24: No need for `await` here.

src/domain/prices/prices.repository.ts
```diff
@@ -16,21 +16,29 @@ export class PricesRepository implements IPricesRepository {
...
   }): Promise<number> {
-    const result = await this.coingeckoApi.getTokenPrice(args);
+    const lowerCaseFiatCode = args.fiatCode.toLowerCase();
+    const lowerCaseTokenAddress = args.tokenAddress.toLowerCase();
+    const result = await this.coingeckoApi.getTokenPrice({
+      chainName: args.chainName,
+      tokenAddress: lowerCaseTokenAddress,
+      fiatCode: lowerCaseFiatCode,
+    });
     const assetPrice = await this.assetPriceValidator.validate(result);
```
fmrsabino, line 40: No need for `await` here.

src/domain/prices/prices.repository.ts
```diff
@@ -16,21 +16,29 @@ export class PricesRepository implements IPricesRepository {
...
   }
 
   async getTokenPrice(args: {
     chainName: string;
     tokenAddress: string;
     fiatCode: string;
   }): Promise<number> {
-    const result = await this.coingeckoApi.getTokenPrice(args);
+    const lowerCaseFiatCode = args.fiatCode.toLowerCase();
+    const lowerCaseTokenAddress = args.tokenAddress.toLowerCase();
```
fmrsabino, line 34: Why are we converting the token address to lower case?

src/routes/balances/balances.controller.spec.ts
```diff
@@ -352,9 +352,7 @@ describe('Balances Controller (Unit)', () => {
         });
 
         await request(app.getHttpServer())
-          .get(
-            `/v1/chains/${chain.chainId}/safes/${safeAddress}/balances/${currency}`,
-          )
+          .get(`/v1/chains/${chain.chainId}/safes/${safeAddress}/balances/EUR`)
```
fmrsabino, line 355: Consider using `faker.finance.currencyCode()`

fmrsabino (changes requested):

src/domain/prices/prices.repository.ts
hectorgomezv: For context: Coingecko converts the token address to lowercase internally, so even if we send a _randomcase_ address, we need to check for a lowercase key in the response. This was already taken into consideration before this change:

```
return assetPrice?.[tokenAddress.toLowerCase()]?.[fiatCode];
```

So having said that it's fair to point out that the lowercase conversion before calling `CoingeckoApi` is not needed, but for consistency, given the API will return lowercase addresses anyway, I took the decision to send them in lowercase also. This also covers us in the (improbable) case that Coingecko starts refusing non-lowercase addresses.

But not strong opinion on this, if you think it's better to just call `.toLowerCase()` when checking the response I'm also OK with it.

src/domain/prices/prices.repository.ts
hectorgomezv: Changed in [b69541f](https://github.com/safe-global/safe-client-gateway/pull/768/commits/b69541f62eedaf3a1a759dab61d4a1ad90206006)

src/domain/prices/prices.repository.ts
hectorgomezv: Changed in [b69541f](https://github.com/safe-global/safe-client-gateway/pull/768/commits/b69541f62eedaf3a1a759dab61d4a1ad90206006)

src/routes/balances/balances.controller.spec.ts
hectorgomezv: Changed in [b69541f](https://github.com/safe-global/safe-client-gateway/pull/768/commits/b69541f62eedaf3a1a759dab61d4a1ad90206006)

src/domain/prices/prices.repository.ts
fmrsabino: No strong opinion on this either. I would leave a comment though on why we change the `tokenAddress` to lower case.

Also the `PricesRepository` is in the domain and we are tying an implementation detail of Coingecko to the domain instead of the `IPricesApi` (Coingecko specific). Making it lowercase, and setting/transforming the respective keys seems to be more appropriate.

fmrsabino (approved):

src/domain/prices/prices.repository.ts
hectorgomezv: > Also the `PricesRepository` is in the domain and we are tying an implementation detail of Coingecko to the domain instead of the `IPricesApi` (Coingecko specific). Making it lowercase, and setting/transforming the respective keys seems to be more appropriate.

I totally agree with you on this. In fact, I had thought about it as well while writing this PR. This is more a technical design issue that was present before this PR, as the lowercase parsing was already in place for the `PricesApi` response. I also think this repository contains logic that should be moved to the Coingecko-specific classes, so I'm going to address that in a separate PR and merge this to address the bug it solves as soon as possible. Thanks!

PR #770, opened 2023-10-20
Change nested transfers mapping

hectorgomezv:
- Changes the calls to `incomingTransferMapper.mapTransfer` inside the function `mapTransfers`: instead of returning an array of promises, the function awaits these calls sequentially and returns resolved values. This avoids requesting the same token data N times in parallel (and therefore having N cache misses). Now the first call sets the token data in the cache, so subsequent calls produce cache hits.
- Adds a new parameter to `TransferInfoMapper.mapTransferInfo`: `resolveAddressInfo`. This is used to indicate the function whether the addresses in the transfer (`from`, `to`, `tokenAddress` fields) should be resolved to an `AddressInfo`. This can be used to avoid resolving an arbitrary number of addresses when the nested transfers for a given transaction surpasses a threshold (`MAX_NESTED_TXS`).

### DIFF
--- a/src/routes/transactions/mappers/transactions-history.mapper.ts
+++ b/src/routes/transactions/mappers/transactions-history.mapper.ts
@@ -32,6 +32,8 @@ class TransactionDomainGroup {
 
 @Injectable()
 export class TransactionsHistoryMapper {
+  private static readonly MAX_NESTED_TXS = 10;
+
   constructor(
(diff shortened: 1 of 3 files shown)


hectorgomezv:
Closed as we decided we'll work on https://github.com/safe-global/safe-client-gateway/pull/771

PR #771, opened 2023-10-20
Slice nested transfers

hectorgomezv:
- Cuts the number of nested transfers in the history to a max value by slicing the `transfers` array within `TransactionsHistoryMapper.mapTransfer`.

### DIFF
--- a/src/routes/transactions/mappers/transactions-history.mapper.ts
+++ b/src/routes/transactions/mappers/transactions-history.mapper.ts
@@ -32,6 +32,8 @@ class TransactionDomainGroup {
 
 @Injectable()
 export class TransactionsHistoryMapper {
+  private static readonly MAX_NESTED_TXS = 10;
+
   constructor(
     private readonly multisigTransactionMapper: MultisigTransactionMapper,
     private readonly moduleTransactionMapper: ModuleTransactionMapper,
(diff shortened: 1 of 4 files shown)


iamacook (approved):
LGTM!

We should probably consider whether this is necessary for the queue as well.

src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -162,16 +171,18 @@ export class TransactionsHistoryMapper {
...
   private mapTransfer(transfers: Transfer[], chainId: string, safe: Safe) {
-    return transfers.map(
-      async (transfer) =>
-        new TransactionItem(
-          await this.incomingTransferMapper.mapTransfer(
-            chainId,
-            transfer,
-            safe,
+    return transfers
+      .slice(0, this.maxNestedTransfers)
```
fmrsabino, line 175: We should be careful regarding how this variable is tied to the `slice` function as negative numbers are allowed.

For example `slice(0, -2)` would give the last two transfers. While this is part of the configuration we should bear in mind the values that we allow.

fmrsabino (approved):

src/routes/transactions/mappers/transactions-history.mapper.ts
hectorgomezv: Yes, good point 👍

PR #772, opened 2023-10-23
fix: remove redundant `isHex` function

iamacook:
With the previous installation of `viem`, we are now using `isHex` from there. As such, our own `isHex` is no longer used. This removes it.


hectorgomezv (approved):

fmrsabino (approved):
🚀

PR #773, opened 2023-10-23
Tag docker image with BUILD_NUMBER

fmrsabino:
Instead of relying on a semver approach for production releases and on a `staging` label for staging releases, these two flows are now merged, and the `BUILD_NUMBER` is instead used for the tag. The `BUILD_NUMBER` corresponds to the shorthand git commit hash.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,8 +3,6 @@ name: CI
 on:
   push:
   pull_request:
-  release:
-    types: [released]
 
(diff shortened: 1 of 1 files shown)


fmrsabino:
Pipeline failing due to b786e81c06db426dd72dab162e8b9e047eb3c597

fmrsabino:
Converting to draft as this proposal is not final and discussions are ongoing.

PR #775, opened 2023-10-23
fix: load Safe info if transaction tags error

### LINKED ISSUE #690: Safe doesn't load on goerli after specific tx
## Bug description

Safe doesn't load anymore.

## Environment
 - Browser: Chrome
 - Wallet: -
 - Chain: Goerli

## Steps to reproduce
1. Open https://app.safe.global/balances?safe=gor:0x9a3227589F7dBDb9f006bACE1383e5e4939dA6CC

## Expected result

Safe is opened

## Obtained result

Not possible to open. Client gateway returns a [500](https://safe-client.safe.global/v1/chains/5/safes/0x9a3227589F7dBDb9f006bACE1383e5e4939dA6CC
) (Should this bug be transferred to the CGW repo then?) 

## Screenshots

image

## Details

This bug was reported at DappCon. It's reproducible for them. They always create 1 Safe, make 1 tx and then the Safe doesn't load anymore. 

The "1 tx" for the Safe above was: https://goerli.etherscan.io/tx/0x9338f8d95210f7325f6bf06cc555dfd017e5cc4819a73bd32ef69b22abe02bae

I have a contact via TG, in case we need more details.

iamacook:
Resolves #690

Generation of the `collectiblesTag`, `txHistoryTag` and `txQueuedTag` rely on multiple endpoints: `/transfers`, `/multisig-transactions` and `/module-transactions`. If one of the latter throws, the Safe could not be loaded (e.g. `/v1/chains/5/safes/0x9a3227589F7dBDb9f006bACE1383e5e4939dA6CC`).

This adjusts the tag generation to otherwise return "now" (the default for Safes with no transactions) if one of the aforementioned endpoints throws. Full test coverage for both the fix and missing cases has also been added.

### DIFF
--- a/src/routes/safes/safes.controller.spec.ts
+++ b/src/routes/safes/safes.controller.spec.ts
@@ -471,6 +471,353 @@ describe('Safes Controller (Unit)', () => {
       );
   });
 
+  it('txQueuedTag is computed from Multisig transaction with modified date', async () => {
+    const chain = chainBuilder().build();
+    const masterCopies = [masterCopyBuilder().build()];
+    const masterCopyInfo = contractBuilder().build();
+    const safeInfo = safeBuilder()
+      .with('masterCopy', masterCopyInfo.address)
+      .build();
+    const fallbackHandlerInfo = contractBuilder()
+      .with('address', safeInfo.fallbackHandler)
+      .build();
+    const guardInfo = contractBuilder().with('address', safeInfo.guard).build();
+    const collectibleTransfers = pageBuilder().build();
+    const moduleTransactions = pageBuilder().build();
+    const messages = pageBuilder().build();
+
+    networkService.get.mockImplementation((url) => {
+      switch (url) {
+        case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
+          return Promise.resolve({ data: chain });
+        case `${chain.transactionService}/api/v1/safes/${safeInfo.address}`:
+          return Promise.resolve({ data: safeInfo });
+        case `${chain.transactionService}/api/v1/about/master-copies/`:
+          return Promise.resolve({ data: masterCopies });
+        case `${chain.transactionService}/api/v1/contracts/${masterCopyInfo.address}`:
+          return Promise.resolve({ data: masterCopyInfo });
+        case `${chain.transactionService}/api/v1/contracts/${fallbackHandlerInfo.address}`:
+          return Promise.resolve({ data: fallbackHandlerInfo });
+        case `${chain.transactionService}/api/v1/contracts/${guardInfo.address}`:
+          return Promise.resolve({ data: guardInfo });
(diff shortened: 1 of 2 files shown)


src/routes/safes/safes.controller.spec.ts
```diff
@@ -471,6 +471,353 @@ describe('Safes Controller (Unit)', () => {
...
+      }
+      return Promise.reject(`No matching rule for url: ${url}`);
+    });
+
+    await request(app.getHttpServer())
+      .get(`/v1/chains/${chain.chainId}/safes/${safeInfo.address}`)
+      .expect(200)
+      .expect((response) =>
+        expect(response.body).toMatchObject({
+          txQueuedTag: Math.floor(Date.now() / 1000).toString(),
```
fmrsabino, line 592: This can potentially fail and make the tests less robust (as it might fail intermittently).

If the difference between the `Date` that is computed in the response and the `Date` that is computed in the `expect` is bigger than `1000` then the test would fail.

I would suggest the following:

```typescript
{
  jest.useFakeTimers();
  jest.setSystemTime(); // set a fake known date here (can be with `faker`)

  ... // test. Instead of Date.now() you can now use the faked date. Date.now should return the faked date also though

  jest.useRealTimers();
}
```

fmrsabino (changes requested):
Looks good! One concern though regarding how we test with timers!

src/routes/safes/safes.controller.spec.ts
iamacook: Good point. I've included the suggestion in ea765edd9135643752f12e879b97fd30afecdc0e.

src/routes/safes/safes.controller.spec.ts
```diff
@@ -759,6 +1134,241 @@ describe('Safes Controller (Unit)', () => {
...
+      }
+      return Promise.reject(`No matching rule for url: ${url}`);
+    });
+
+    await request(app.getHttpServer())
+      .get(`/v1/chains/${chain.chainId}/safes/${safeInfo.address}`)
+      .expect(200)
+      .expect((response) =>
+        expect(response.body).toMatchObject({
+          txHistoryTag: Math.floor(Date.now() / 1000).toString(),
```
fmrsabino, line 1365: nitpick: in the other tests we are accessing the date via `date.getTime()`.

However `Date.now()` still works since we are setting the system time to the fake date.

fmrsabino (approved):

src/routes/safes/safes.controller.spec.ts
iamacook: I overlooked that - thanks! I've changed it to `date.getTime()` in 69ef3acd19c0d1b4e866292043efc35ccdf7c4ab.

fmrsabino (approved):
🚀 🚀 🚀

PR #787, opened 2023-10-24
feat: add alerts (Tenderly) datasource

iamacook:
This adds a new `alerts` datasouce, leveraging [the Tenderly API](https://docs-api.tenderly.co/contract-management) in anticipation of Tenderly Alerts for the forthcoming recovery implementation:

- Tenderly-specific configuration for: URL, API key, account and project.
- `IAlertsApi` with `addContracts` and `removeContracts` methods.
- Test-covered `TenderlyApi`, which implements `IAlertsApi`.
- `AlertsModule`, providing the `TenderlyApi`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,6 +57,9 @@ jobs:
         env:
           EXCHANGE_API_BASE_URI: http://api.exchangeratesapi.io/v1
           EXCHANGE_API_KEY: ${{ secrets.EXCHANGE_API_KEY }}
+          ALERTS_PROVIDER_API_KEY: ${{ secrets.ALERTS_PROVIDER_API_KEY }}
+          ALERTS_PROVIDER_ACCOUNT: ${{ secrets.ALERTS_PROVIDER_ACCOUNT }}
+          ALERTS_PROVIDER_PROJECT: ${{ secrets.ALERTS_PROVIDER_PROJECT }}
           REDIS_HOST: localhost
           REDIS_PORT: 6379
           SAFE_CONFIG_BASE_URI: ${{ secrets.SAFE_CONFIG_BASE_URI }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,9 +57,6 @@ jobs:
         env:
           EXCHANGE_API_BASE_URI: http://api.exchangeratesapi.io/v1
           EXCHANGE_API_KEY: ${{ secrets.EXCHANGE_API_KEY }}
-          ALERTS_PROVIDER_API_KEY: ${{ secrets.ALERTS_PROVIDER_API_KEY }}
-          ALERTS_PROVIDER_ACCOUNT: ${{ secrets.ALERTS_PROVIDER_ACCOUNT }}
-          ALERTS_PROVIDER_PROJECT: ${{ secrets.ALERTS_PROVIDER_PROJECT }}
           REDIS_HOST: localhost
           REDIS_PORT: 6379
           SAFE_CONFIG_BASE_URI: ${{ secrets.SAFE_CONFIG_BASE_URI }}
--- a/src/datasources/alerts-api/tenderly-api.service.spec.ts
+++ b/src/datasources/alerts-api/tenderly-api.service.spec.ts
@@ -1,0 +1,119 @@
+import { faker } from '@faker-js/faker';
+import { FakeConfigurationService } from '@/config/__tests__/fake.configuration.service';
+import { TenderlyApi } from '@/datasources/alerts-api/tenderly-api.service';
+import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
+import { AxiosNetworkService } from '@/datasources/network/axios.network.service';
+import { Contract, ContractId } from '@/domain/alerts/entities/alerts.entity';
+
+const mockNetworkService = jest.mocked({
+  post: jest.fn(),
+  delete: jest.fn(),
+}) as unknown as AxiosNetworkService;
+
+const mockHttpErrorFactory = jest.mocked({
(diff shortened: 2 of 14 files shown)


iamacook:
Marking as draft whilst I investigate the broken test.

src/routes/root/root.controller.spec.ts
```diff
@@ -15,6 +17,8 @@ describe('Root Controller tests', () => {
     })
       .overrideModule(CacheModule)
       .useModule(TestCacheModule)
+      .overrideModule(configurationModule)
+      .useModule(ConfigurationModule.register(configuration))
```
iamacook, line 21: The Tenderly API is required and if we don't mock the configuration here, this test does not pass.

hectorgomezv:
The code looks good, great job! 🚀

But I think the e2e tests are failing because of the lack of some mandatory configuration (`alerts` key in `configuration.ts`). Also, as this configuration is mandatory at the moment, this absence is preventing the service from starting if the variables are not set in the environment. 

I'd suggest either stating this `alerts` configuration is mandatory and including it in `configuration.validator.ts`, `.env.sample`, `docker-compose.yml`, `ci.yml`, etc., or making this configuration optional and using `configurationService.get` instead of `configurationService.getOrThrow`, checking the configuration only when `TenderlyApi` functions are called.

iamacook:
> But I think the e2e tests are failing because of the lack of some mandatory configuration (`alerts` key in `configuration.ts`). Also, as this configuration is mandatory at the moment, this absence is preventing the service from starting if the variables are not set in the environment.
> 
> I'd suggest either stating this `alerts` configuration is mandatory and including it in `configuration.validator.ts`, `.env.sample`, `docker-compose.yml`, `ci.yml`, etc., or making this configuration optional and using `configurationService.get` instead of `configurationService.getOrThrow`, checking the configuration only when `TenderlyApi` functions are called.

Thank you for the review! I will convert this to a draft whilst I implement your suggested fix.

src/domain/alerts/entities/alerts.entity.ts
```diff
@@ -0,0 +1,7 @@
+export type Contract = {
+  address: string;
+  networkId: string;
```
fmrsabino, line 3: From a domain perspective, we have been addressing this field as `chainId` instead of `networkId`.

For consistency, I'd keep using `chainId` on the domain side and map it to `networkId` in the Tenderly API.

src/domain.module.ts
```diff
@@ -71,6 +72,7 @@ import { FiatCodesValidator } from '@/domain/prices/fiat-codes.validator';
     ConfigApiModule,
     ExchangeApiModule,
     PricesApiModule,
+    AlertsApiModule,
```
fmrsabino, line 75: nitpick: I'd leave this in lexicographical order.

(It's already not in the correct order though... so we can either tackle it here or in a different PR).

src/datasources/alerts-api/tenderly-api.service.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+import { TenderlyApi } from '@/datasources/alerts-api/tenderly-api.service';
+import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
+import { AxiosNetworkService } from '@/datasources/network/axios.network.service';
+import { Contract, ContractId } from '@/domain/alerts/entities/alerts.entity';
+import { DataSourceError } from '@/domain/errors/data-source.error';
+
+const networkService = {
+  post: jest.fn(),
+  delete: jest.fn(),
+} as unknown as AxiosNetworkService;
```
fmrsabino, line 12: We should use the interface type here instead of the actual implementation. If we start using a different HTTP client this would mean a new implementation of `INetworkService`.

Using the reference to `INetworkService` instead would make this test compatible with the change that I described.

.github/workflows/ci.yml
```diff
@@ -57,6 +57,9 @@ jobs:
         env:
           EXCHANGE_API_BASE_URI: http://api.exchangeratesapi.io/v1
           EXCHANGE_API_KEY: ${{ secrets.EXCHANGE_API_KEY }}
+          ALERTS_PROVIDER_API_KEY: ${{ secrets.ALERTS_PROVIDER_API_KEY }}
+          ALERTS_PROVIDER_ACCOUNT: ${{ secrets.ALERTS_PROVIDER_ACCOUNT }}
+          ALERTS_PROVIDER_PROJECT: ${{ secrets.ALERTS_PROVIDER_PROJECT }}
```
fmrsabino, line 62: Do we need to setup these since we do not have E2E tests for the Tenderly proxy?

fmrsabino (changes requested):
Nice work 👏 

Some small requests.

src/domain/alerts/entities/alerts.entity.ts
iamacook: Good point. I've change this in 36844eeae590fc1a6dd9d1ec0e87d8645e760b2d.

src/domain.module.ts
iamacook: I've moved this particular module to the correct alphabetical position in 36844eeae590fc1a6dd9d1ec0e87d8645e760b2d. I agree that we should tackle the rest in a separate PR.

src/datasources/alerts-api/tenderly-api.service.spec.ts
iamacook: I used a few other tests as reference here. I've adjusted this in 36844eeae590fc1a6dd9d1ec0e87d8645e760b2d and will update the other tests I used as a reference in another PR.

fmrsabino (approved):
🚀 🚀 🚀

hectorgomezv (approved):
🚀

PR #789, opened 2023-10-25
fix: use interface over actual implementation

iamacook:
As per https://github.com/safe-global/safe-client-gateway/pull/787#discussion_r1371457958, the `transaction-api.service` test has been updated to use the interface type (`INetworkService`) instead of the actual network implementation (`AxiosNetworkService`) in case we use a different client.


fmrsabino (approved):
Good catch! 👏 🚀

PR #792, opened 2023-10-25
fix: `get-contract` E2E test

iamacook:
A logo was added for `0x7cbB62EaA69F79e6873cD1ecB2392971036cFAa4` on the Transaction Service which, in turn, broke our `get-contract.e2e-spec` test. This adds the new `logoUri` to the test.

### DIFF
--- a/src/routes/contracts/__tests__/get-contract.e2e-spec.ts
+++ b/src/routes/contracts/__tests__/get-contract.e2e-spec.ts
@@ -31,8 +31,7 @@ describe('Get contract e2e test', () => {
       address: '0x7cbB62EaA69F79e6873cD1ecB2392971036cFAa4',
       name: 'CreateCall',
       displayName: '',
-      logoUri:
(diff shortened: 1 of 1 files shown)


hectorgomezv (approved):
Great job! I'm OK with the fix, but given this is a mutable field, and it's not owned by this team directly, as an option we could also be more permissive:
```
logoUri: expect.any(String),
```

But I left this up to you. I think it's OK to have this hardcoded as we have this data in-house 🙂

iamacook:
> Great job! I'm OK with the fix, but given this is a mutable field, and it's not owned by this team directly, as an option we could also be more permissive

I agree with you. I have changed it to include your suggestion in 3fff61581abb95c360f1456a5c570faf9d01d7ef.

iamacook:
> Great job! I'm OK with the fix, but given this is a mutable field, and it's not owned by this team directly, as an option we could also be more permissive:
> 
> ```
> logoUri: expect.any(String),
> ```
> 
> But I left this up to you. I think it's OK to have this hardcoded as we have this data in-house 🙂

Whilst I agree, I played around a bit and realised that there are other mutable fields present in the expected response. I think we should leave it as is (we know our E2E is working if a value changes).

PR #793, opened 2023-10-25
feat: add `/alerts` route

iamacook:
This adds a new `/alerts` route for receiving Tenderly Alerts:

- `AlertController` with `AlertValidationPipe`
- `AlertsService` which logs incoming `Alert`s.
- Associated `Builder` and basic test coverage


src/routes/alerts/alerts.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { JsonSchemaService } from '@/validation/providers/json-schema.service';
+import { AlertsController } from '@/routes/alerts/alerts.controller';
+import { AlertsService } from '@/routes/alerts/alerts.service';
+
+@Module({
+  providers: [JsonSchemaService, AlertsService],
```
fmrsabino, line 7: `JsonSchemaService` can be removed as it is provided by a global module so it's already available to this module.

src/routes/alerts/entities/alerts.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+  gas_used: string;
+  cumulative_gas_used: string;
+  gas_price: string;
+  gas_tip_cap: string;
+  gas_fee_cap: string;
+};
+
+export enum AlertEventType {
+  ALERT = 'ALERT',
+}
```
fmrsabino, line 28: Personally I find it strange that an `AlertEventType` would have as a type `ALERT`.

I'd probably use `EventType` here wdyt?

src/routes/alerts/alerts.controller.ts
```diff
@@ -0,0 +1,23 @@
...
+})
+@ApiExcludeController()
+export class AlertsController {
+  constructor(private readonly alertsService: AlertsService) {}
+
+  @Post('/alerts')
+  @HttpCode(200)
+  postAlert(
+    @Body(AlertValidationPipe)
+    alertPayload: Alert,
```
fmrsabino, line 19: This should be ok given the current stage of this feature. But the route entities will need to be converted to classes with OpenAPI annotations.

See `src/routes/transactions/entities/add-confirmation.dto.ts` as an example.

src/app.module.ts
```diff
@@ -47,6 +48,7 @@ export const configurationModule = ConfigurationModule.register(configuration);
   imports: [
     // features
     AboutModule,
+    AlertsModule,
```
fmrsabino, line 51: I think this is a good hint that the `AppModule` needs to be a Dynamic Module.

Why: by including a module that includes a controller and respective routes we are exposing this functionality to the end users.

See: https://docs.nestjs.com/fundamentals/dynamic-modules

fmrsabino (changes requested):

src/routes/alerts/alerts.module.ts
iamacook: I've removed it in ffbf4a04e7995592f9eeb14a12cfb84342b54672.

src/routes/alerts/entities/alerts.entity.ts
iamacook: Very true. I've adjusted it in ffbf4a04e7995592f9eeb14a12cfb84342b54672.

src/routes/alerts/alerts.controller.ts
iamacook: I've converted these to classes in ffbf4a04e7995592f9eeb14a12cfb84342b54672 but not added any `ApiProperty`s as this will not be exposed to the user.

src/app.module.ts
fmrsabino: Suggested approach here: https://github.com/safe-global/safe-client-gateway/pull/794

If approved, you can then conditionally enable this via the provided configuration.

iamacook: In anticipation of that, I've removed the `AlertsModule` from the `AppModule` in 2a5abef92a96cd6b43694740b3c3b22ef7d37a50 and disabled the test for now to unblock https://github.com/safe-global/safe-client-gateway/pull/795.

src/routes/alerts/entities/schemas/alerts.schema.ts
```diff
@@ -77,8 +77,8 @@ export const alertSchema: JSONSchemaType<Alert> = {
     id: { type: 'string' },
     event_type: {
       type: 'string',
-      default: AlertEventType.ALERT,
-      enum: Object.values(AlertEventType),
+      default: EventType.ALERT,
```
fmrsabino, line 80: OOC: why do we assume a default for the `EventType`?

hectorgomezv (commented):
Great job! The PR looks good to me, I've just left a nitpick comment.

fmrsabino (approved):
🚀

src/routes/alerts/pipes/alert-validation.pipe.ts
```diff
@@ -0,0 +1,35 @@
...
+    jsonSchemaService.getSchema(
+      ALERT_TRANSACTION_SCHEMA_ID,
+      alertTransactionSchema,
+    );
+
+    this.isAlert = jsonSchemaService.getSchema(ALERT_SCHEMA_ID, alertSchema);
+  }
+
+  transform(value: unknown): Alert {
+    if (this.isAlert(value)) {
```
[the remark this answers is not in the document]
iamacook: I've migrated to the `GenericValidator` in 07d22ecc4f5541ec9dfbde1722174cf7ce469364.

src/routes/alerts/entities/schemas/alerts.schema.ts
iamacook: We shouldn't be - good catch! There is also a `"TEST"` `type`. I've removed the `default` in 07d22ecc4f5541ec9dfbde1722174cf7ce469364.

iamacook: We shouldn't be - good catch! There is also a `"TEST"` `type`. I've removed the `default` in 07d22ecc4f5541ec9dfbde1722174cf7ce469364.

hectorgomezv (approved):
🚀🚀

PR #794, opened 2023-10-26
Make AppModule a Dynamic Module

fmrsabino:
- `AppModule` is now a Dynamic Module – this means that to instantiate it, `AppModule.register` needs to be called.
- This effectively allows conditionally including and excluding from the application. Example: if new routes are being worked on and should be enabled/disabled via a configuration, this can now be done in the register process of the `AppModule`.
- Removes the `configurationModule` – this module was exposed in order to allow overriding it in the tests (access to the respective reference is required). Since the `configuration` of the service is now an argument of `AppModule` (root level) this means that the configuration can be provided directly via the `AppModule` as it will be forwarded to the inner `ConfigurationModule`.

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -1,4 +1,5 @@
 import {
+  DynamicModule,
   MiddlewareConsumer,
   Module,
   NestModule,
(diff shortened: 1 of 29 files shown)


src/app.module.ts
```diff
@@ -110,4 +51,67 @@ export class AppModule implements NestModule {
       .apply(ClsMiddleware, NotFoundLoggerMiddleware)
       .forRoutes({ path: '*', method: RequestMethod.ALL });
   }
+
+  static register(configFactory: ConfigFactory = configuration): DynamicModule {
+    return {
```
fmrsabino, line 56: @iamacook in https://github.com/safe-global/safe-client-gateway/pull/793 you will then be able to conditionally enable/disable the `AlertsModule` via this register function by accessing the `configuration()` object.

hectorgomezv (approved):
Awesome! 🚀🚀

PR #796, opened 2023-10-26
Add recommended nonce endpoint

### LINKED ISSUE #681: Alternative way to retrieve the `recommendedNonce`
Currently the `recommendedNonce` is retrieved by calling `CGW POST chains/:chainId/safes/:address/multisig-transactions/estimations`.

This route calls the following endpoints on the Safe Transaction Service:
1. `TX POST /api/v1/safes/:safeAddress/multisig-transactions/estimations/`
2. `TX GET /api/v1/safes/:safeAddress`
3. `TX GET /api/v1/safes/:safeAddress/multisig-transactions/` – with `ordering=-nonce&trusted=true&limit=1`

However the parameters returned by call 1 (`safeTxGas`) are not used by the frontend clients for 1.3.0 safes. This means that starting from Safes with version `1.3.0`, the recommended nonce is computed using the current Safe Nonce (from call 2) and the most recent nonce from a trusted Multisig contract (call 3).

## Proposal

Since we should keep compatibility with Safes with version lower than 1.3.0, `CGW POST chains/:chainId/safes/:address/multisig-transactions/estimations` should remain with the same functionality.

Given that for Safes with version greater than 1.3.0, the call to `TX POST /api/v1/safes/:safeAddress/multisig-transactions/estimations/` should be avoided, we should introduce a new endpoint or add the `recommendedNonce` to an existing endpoint.

One idea is to introduce `recommendedNonce` to `CGW GET chains/:chainId/safes/:safeAddress` however this is also one of the most requested endpoints and we might be computing data which is not required by the frontend when this endpoint is called.

fmrsabino:
Closes #681 

- Adds new endpoint: `GET chains/:chainId/safes/:safeAddress/recommended-nonce`
- This endpoint returns the recommended nonce based on the current state of the Safe and the current transactions queued or in history.
- The recommended nonce that is returned is the same one as the one returned by `GET chains/:chainId/safes/:address/multisig-transactions/estimations` but without making any external estimate calls thus this new endpoint is preferred over the estimate one if the nonce is the only property needed to be computed.

The endpoint is as follows:

```javascript
GET chains/:chainId/safes/:safeAddress/recommended-nonce

{
  "nonce": <number>
}
```

### DIFF
--- a/src/routes/estimations/estimations.service.ts
+++ b/src/routes/estimations/estimations.service.ts
@@ -34,37 +34,14 @@ export class EstimationsService {
       chainId: args.chainId,
       address: args.address,
     });
-    const recommendedNonce = await this.getEstimationRecommendedNonce({
+    const recommendedNonce = await this.safeRepository.getRecommendedNonce({
       chainId: args.chainId,
       safeAddress: args.address,
-      safeNonce: safe.nonce,
     });
     return new EstimationResponse(
       safe.nonce,
       recommendedNonce,
       estimation.safeTxGas,
     );
-  }
-
-  /**
-   * Gets the maximum between the current Safe nonce and the last transaction nonce plus 1.
-   * If there is no last transaction, the Safe nonce is returned.
-   *
-   * @returns recommended nonce for next transaction.
-   */
-  private async getEstimationRecommendedNonce(args: {
-    chainId: string;
-    safeAddress: string;
-    safeNonce: number;
-  }): Promise<number> {
-    const lastTransaction =
-      await this.safeRepository.getLastTransactionSortedByNonce({
-        chainId: args.chainId,
-        safeAddress: args.safeAddress,
-      });
-
-    return lastTransaction
-      ? Math.max(args.safeNonce, lastTransaction.nonce + 1)
-      : args.safeNonce;
   }
 }
--- a/src/domain/safe/safe.repository.interface.ts
+++ b/src/domain/safe/safe.repository.interface.ts
@@ -160,4 +160,15 @@ export interface ISafeRepository {
     safeAddress: string;
     proposeTransactionDto: ProposeTransactionDto;
   }): Promise<unknown>;
+
+  /**
+   * Gets the maximum between the current Safe nonce and the last transaction nonce plus 1.
+   * If there is no last transaction, the Safe nonce is returned.
+   *
+   * @returns recommended nonce for next transaction.
+   */
+  getRecommendedNonce(args: {
+    chainId: string;
+    safeAddress: string;
+  }): Promise<number>;
 }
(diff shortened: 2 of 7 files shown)


fmrsabino:
Tests are failing due to recent merge conflict. Fixed in https://github.com/safe-global/safe-client-gateway/pull/797

src/routes/estimations/estimations.service.ts
```diff
@@ -34,37 +34,14 @@ export class EstimationsService {
       chainId: args.chainId,
       address: args.address,
     });
-    const recommendedNonce = await this.getEstimationRecommendedNonce({
+    const recommendedNonce = await this.safeRepository.getRecommendedNonce({
```
hectorgomezv, line 37: `EstimationsService.getEstimation` execution would call `SafeRepository.getSafe` twice if I'm not wrong (in line 33, and also inside `SafeRepository.getRecommendedNonce`). I know source data should be cached on the second call, but even considering this, I think it's better to avoid the duplicated call.

It could make sense to have `safe` as an optional argument for `SafeRepository.getRecommendedNonce`, so the caller could provide the `Safe` or not, and  `SafeRepository.getRecommendedNonce` would call `SafeRepository.getSafe` only when needed. What do you think?

hectorgomezv (changes requested):
Great job! I just commented on a minor concern.

src/routes/estimations/estimations.service.ts
fmrsabino: Yes the reason why I decided to use the Safe address and not the Safe instance is because from an interface perspective (specially one exposed to the routes) we have consistency in the functionality that we provide. See https://github.com/safe-global/safe-client-gateway/blob/recommended-nonce-endpoint/src/domain/safe/safe.repository.interface.ts

If `safe` is then an optional argument at the interface level it can be more confusing in my opinion when developing new features – calling this function would probably raise the question if the Safe info should be passed to it or not which deviates from the other functions.

A better alternative in my opinion would be to return a tuple of `[recommendedNonce, SafeInfo]` (and the Estimations Controller would use both). I didn't do this because I would say that there were not so many benefits to it – that use case is specific to the Estimations Controller and given that we are in the `ISafeRepository` we should assume that it'd have access to the respective `SafeInfo` domain entity.

hectorgomezv: Thanks for the detailed explanations. I think it makes sense from an architecture perspective to favor clarity over performance, and I get your point about having access to the `SafeInfo` entity. I also thought about changing the return type to a tuple including the `SafeInfo`, but I think there is no perfect solution as you said, so I think reading from the cache on the second call is acceptable for the sake of simplicity on the API contract 👍

hectorgomezv (approved):
Great job! 🚀

src/routes/estimations/estimations.service.ts
fmrsabino: Performance was also taken into consideration. We do have one more cache-read when calling the estimations endpoint but I think that is acceptable given the scope.

PR #799, opened 2023-10-27
feat: add `TenderlySignatureGuard`

iamacook:
This implements the `TenderlySignatureGuard` for the `AlertsController` which checks the signature of the incoming alert.

The following template from Tenderly was used as a reference:

```js
// server.js
//
// Use this sample code to handle webhook events
// 1) Paste this code into a new file (server.js)
//
// 2) npm init
//
// 3) Install dependencies
//   npm install --save express
//   npm install --save crypto
//
// 4) Run the server on http://localhost:8008
//   npm start (start script in package.json must point to server.js)

const express = require('express');
const crypto = require('crypto');
const app = express();

// replace this with key taken from dashboard for specific webhook
// if you are testing webhook before creation set it on empty string
const signingKey = '';

app.post('/webhook', express.raw({type: 'application/json'}), (request, response) => {
  const signature = request.headers['x-tenderly-signature'];
  const timestamp = request.headers['date'];

  if (!isValidSignature(signature, request.body, timestamp)) {
    response.status(400).send('Webhook signature not valid');
    return;
  }

  let body;
  try {
    body = JSON.parse(request.body.toString());
  } catch(e) {
    response.status(400).send('Webhook error: ', e);
    return;
  }
  const eventType = body.event_type;

  switch (eventType) {
    case 'TEST':
      // Then define and call a function to handle the test event
      break;
    case 'ALERT':
      // Then define and call a function to handle the alert event
      break;
      // ... handle other event types
    default:
      console.log('Unhandled event type ', eventType);
  }

  // Return a 200 response
  response.send();
});

const port = 8008;
app.listen(port, () => console.log('Running on port ', port));

function isValidSignature(signature, body, timestamp) {
  const hmac = crypto.createHmac("sha256", signingKey); // Create a HMAC SHA256 hash using the signing key
  hmac.update(body.toString(), 'utf8'); // Update the hash with the request body using utf8
  hmac.update(timestamp); // Update the hash with the request timestamp
  const digest = hmac.digest("hex");
  return crypto.timingSafeEqual(Buffer.from(signature), Buffer.from(digest));
}
```


src/routes/alerts/entities/__tests__/fakeTenderlySignature.ts
```diff
@@ -0,0 +1,24 @@
+import { Alert } from '@/routes/alerts/entities/alert.dto.entity';
+import * as crypto from 'crypto';
+
+// The `x-tenderly-signature` header contains a cryptographic signature. The webhook request signature is
+// a HMAC SHA256 hash of concatenated signing secret, request payload, and timestamp, in this order.
+// @see https://github.com/Tenderly/tenderly-docs/blob/d836e99fcc22f141a155688128548505b9fcbf9c/alerts/configuring-alert-destinations/configuring-alert-destinations.md?plain=1#L74
+export function fakeTenderlySignature(args: {
```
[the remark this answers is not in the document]
iamacook: Good point. I've localised it in 72ce3f58d963e22d86f8b9c3bbf204c023752885.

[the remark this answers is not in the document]
hectorgomezv: Great, thanks! 🙏

hectorgomezv (approved):

.github/workflows/ci.yml
```diff
@@ -61,6 +61,7 @@ jobs:
         env:
           EXCHANGE_API_BASE_URI: http://api.exchangeratesapi.io/v1
           EXCHANGE_API_KEY: ${{ secrets.EXCHANGE_API_KEY }}
+          ALERTS_PROVIDER_SIGNING_KEY: ${{ secrets.ALERTS_PROVIDER_SIGNING_KEY }}
```
fmrsabino, line 64: Is it possible to generate the signing key on the test itself? That would mean that we wouldn't need to configure it in the actions.

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -41,23 +66,54 @@ describe('Alerts (Unit)', () => {
...
+      alert,
+      timestamp,
+    });
+
+    await request(app.getHttpServer())
+      .post('/alerts')
+      .set('x-tenderly-signature', signature)
+      .set('date', timestamp)
+      .send(alert)
+      .expect(200);
```
fmrsabino, line 83: We should probably assert the body too even if empty. That way if we start returning something this test would fail.

src/routes/alerts/guards/tenderly-signature.guard.ts
```diff
@@ -0,0 +1,76 @@
...
+  private static readonly SIGNATURE_HEADER = 'x-tenderly-signature';
+  private static readonly TIMESTAMP_HEADER = 'date';
+
+  private readonly signingKey: string;
+
+  constructor(
+    @Inject(IConfigurationService)
+    private readonly configurationService: IConfigurationService,
+  ) {
+    // Dashboard key for specific webhook
```
fmrsabino, line 22: I move this description to the environment variable itself (in `.env.sample` for example) as it's the entry point of contact for the environment configuration (for users deploying the service).

src/routes/alerts/alerts.controller.spec.ts
iamacook: I've added the expectation in 40b5b7e3bc64689c32778bf570fe515d9ba972a7.

src/routes/alerts/guards/tenderly-signature.guard.ts
iamacook: As we [already have a description in the .env.sample](https://github.com/safe-global/safe-client-gateway/blob/40b5b7e3bc64689c32778bf570fe515d9ba972a7/.env.sample#L17), I removed this comment in 40b5b7e3bc64689c32778bf570fe515d9ba972a7.

.github/workflows/ci.yml
iamacook: I've moved this to `e2e-setup` in a57a546ec1e5824d6de4f58f09b739af118f2211.

fmrsabino (approved):
🚀 🚀 🚀

fmrsabino (approved):

fmrsabino (approved):

PR #795, opened 2023-10-26
feat: add alerts domain

iamacook:
The adds a new alerts domain that (initially) decodes `TransactionAdded` events containing `addOwnerWithThreshold` `data`:

- `AlertsRepository` implementing `IAlertsRepository`
  - `handleAlerts` to decode events and notify accordingly
- `AbiDecoder` helper with specific implementations for the Delay Modifier `TransactionAdded` event and Safe `addOwnerWithThreshold` function.

Note: the `AbiDecoder` helper has similarities with the relay `ContractHelper` but preserves types. I'd appreciate feedback on whether we should pivot to the `ContractHelper` instead, or alternatively improve the `ContractHelper` when we continue with the relay migration.

PR #801, opened 2023-10-27
feat: add `/recovery` route

iamacook:
This adds a new `/recovery` route for adding addresses to the Tenderly recovery project:

- `RecoveryController` with `AddRecoveryModuleDtoValidationPipe`
- `RecoveryService` which adds module addresses.
- Associated Builder and test coverage.


iamacook:
I'm marking this as a draft for now as it relies on #795.

src/routes/recovery/recovery.controller.ts
```diff
@@ -0,0 +1,28 @@
...
+
+@ApiTags('recovery')
+@Controller({
+  version: '1',
+})
+export class RecoveryController {
+  constructor(private readonly recoveryService: RecoveryService) {}
+
+  @HttpCode(200)
+  @Post('chains/:chainId/safes/:safeAddress/recovery')
```
fmrsabino, line 15: I'd move this as a prefix of the controller as this should be the base path for any recovery feature.

src/routes/recovery/recovery.controller.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+  let alertsAccount: string;
+  let alertsProject: string;
+  let networkService;
+
+  beforeEach(async () => {
+    jest.clearAllMocks();
+
+    const moduleFixture: TestingModule = await Test.createTestingModule({
+      // TODO remove AlertsModule and RecoveryModule once they're integrated in the AppModule
+      imports: [AppModule, AlertsModule, RecoveryModule],
```
fmrsabino, line 34: The `AlertsModule` doesn't need to be included as we are not testing the alerts routes.

src/routes/recovery/recovery.service.ts
```diff
@@ -0,0 +1,26 @@
...
+  async addRecoveryModule(args: {
+    chainId: string;
+    safeAddress: string;
+    addRecoveryModuleDto: AddRecoveryModuleDto;
+  }): Promise<void> {
+    const contract: Contract = {
+      chainId: args.chainId,
+      address: args.addRecoveryModuleDto.moduleAddress,
+      displayName: `${args.chainId}:${args.safeAddress}`,
+    };
```
fmrsabino, line 23: I think having `Contract` for the alert entity can be confusing as we already have a domain entity named `Contract` which is our representation of a `Contract`.

Maybe this can one can be associated with alerts directly and we can call it `AlertsRegistration` or similar? What do you think?

We can tackle it in a different PR as it affects the domain part of Alerts.

src/datasources/alerts-api/tenderly-api.service.ts
```diff
@@ -49,7 +49,7 @@ export class TenderlyApi implements IAlertsApi {
         },
       });
     } catch (error) {
-      this.httpErrorFactory.from(error);
+      throw this.httpErrorFactory.from(error);
```
fmrsabino, line 52: Are the error messages documented on Tenderly? We need to be sure that we are not leaking any sensitive data while mapping the error from the HttpErrorFactory (as it can forward the error message to the other layers).

src/routes/recovery/recovery.controller.spec.ts
```diff
@@ -0,0 +1,138 @@
...
+
+    it('Should fail with An error occurred', async () => {
+      const addRecoveryModuleDto = addRecoveryModuleDtoBuilder().build();
+      const chainId = faker.string.numeric();
+      const safeAddress = faker.finance.ethereumAddress();
+
+      networkService.post.mockImplementation((url) =>
+        url ===
+        `${alertsUrl}/api/v2/accounts/${alertsAccount}/projects/${alertsProject}/contracts`
+          ? Promise.reject({ status: 503 })
```
fmrsabino, line 124: What is the reason behind testing 503 and not other error codes?

(not saying that we should test each error code but I'm curious about the 503 one)

src/datasources/alerts-api/tenderly-api.service.spec.ts
```diff
@@ -100,14 +100,13 @@ describe('TenderlyApi', () => {
   });
 
   it('should forward error', async () => {
-    mockNetworkService.post.mockImplementation(() =>
-      Promise.reject('Unexpected error'),
-    );
-    mockHttpErrorFactory.from.mockImplementation(
-      () => new DataSourceError('Unexpected error'),
+    const expected = new DataSourceError('Unexpected error');
+    mockHttpErrorFactory.from.mockReturnValue(expected);
```
fmrsabino, line 104: Not specifically related with this PR but would affect the changes proposed: why not use the real instance of the `HttpErrorFactory`?

That way we cover more of the actual implementation and we avoid having to set the correct mock value set here.

src/domain/alerts/alerts.repository.interface.ts
```diff
@@ -1,7 +1,10 @@
+import { Contract } from '@/domain/alerts/entities/alerts.entity';
 import { AlertLog } from '@/routes/alerts/entities/alert.dto.entity';
 
 export const IAlertsRepository = Symbol('IAlertsRepository');
 
 export interface IAlertsRepository {
+  addContracts(contracts: Array<Contract>): Promise<void>;
```
fmrsabino, line 7: nitpick: something to consider (can be done at a later stage) – to include some documentation on what is expected by this function.

src/routes/recovery/recovery.service.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  async addRecoveryModule(args: {
+    chainId: string;
+    safeAddress: string;
+    addRecoveryModuleDto: AddRecoveryModuleDto;
+  }): Promise<void> {
+    const contract: Contract = {
+      chainId: args.chainId,
+      address: args.addRecoveryModuleDto.moduleAddress,
+      displayName: `${args.chainId}:${args.safeAddress}`,
```
fmrsabino, line 22: Is this formatting `<chainId>:<safeAddress>` something that we should enforce across our codebase when using this interface?

fmrsabino (changes requested):

src/routes/recovery/recovery.controller.ts
iamacook: I moved it in 193e5cc012a581c84310250f9ebb28e0567a3ced.

src/routes/recovery/recovery.controller.spec.ts
iamacook: I removed it in 193e5cc012a581c84310250f9ebb28e0567a3ced.

src/routes/recovery/recovery.service.ts
iamacook: I agree. I've made a note to change it in a follow-up PR.

src/routes/recovery/recovery.controller.spec.ts
iamacook: I used the [`delegates.controller.spec` as a reference](https://github.com/safe-global/safe-client-gateway/blob/recovery-route/src/routes/delegates/delegates.controller.spec.ts#L239) for this test. Would you suggest a different error code?

src/datasources/alerts-api/tenderly-api.service.spec.ts
iamacook: I've adjusted it in 193e5cc012a581c84310250f9ebb28e0567a3ced but as per [your suggestion](https://github.com/safe-global/safe-client-gateway/pull/801#discussion_r1383071213), maybe this will need to be further changed.

src/domain/alerts/alerts.repository.interface.ts
iamacook: I will address this in a follow-up PR as I have yet to document other areas of recovery.

src/routes/recovery/recovery.service.ts
iamacook: This is predominantly for Tenderly. We can determine which Safe the module is added to. What are your thoughts?

Whether we include this across the codebase, I am impartial.

hectorgomezv: Out of curiosity: what's the purpose of this `displayName`? I understand this is sent to Tenderly, and what's the purpose there? If it's just for presentation purposes in the UI, I'd concatenate the contract address also, to have an unique identifier. 

And in my opinion, if it's not used in more places in the code, I think the inline template string is clearer.

iamacook: > I understand this is sent to Tenderly, and what's the purpose there? If it's just for presentation purposes in the UI, I'd concatenate the contract address also, to have an unique identifier.

It is just to display it in the UI:

image

I moved it back to the payload and added the module address in 890eef9cec76fe8dd7bf39b0bb801e933817beeb.

src/datasources/alerts-api/tenderly-api.service.ts
iamacook: As discussed, this will be changed before released in a follow-up PR to ensure no "leaks".

src/routes/recovery/recovery.controller.spec.ts
fmrsabino: Any client or server error from the external source should work for this test so you can have something like:

```typescript
const statusCode = faker.internet.httpStatusCode({ types: ['clientError', 'serverError'] });
```

and then use this status code in the assertions.

src/datasources/alerts-api/tenderly-api.service.spec.ts
fmrsabino: Yes my sugggestion was to remove the Mock entirely and use the real implementation:

```typescript
it('should forward error', async () => {
  const status = faker.internet.httpStatusCode({ types: ['serverError'] });
  const error = {
    status,
    data: {
      message: 'Unexpected error',
    },
  };
  mockNetworkService.post.mockRejectedValueOnce(error);

  await expect(service.addContracts([])).rejects.toThrowError(
    new DataSourceError('Unexpected error', status),
  );

  expect(mockNetworkService.post).toHaveBeenCalledTimes(1);
});
```

In this example, the only component being mocked is the HTTP component as it's more complex and we want to abstract it away for this test.

This requires you changing the way `TenderlyApi` is constructed:

```typescript
service = new TenderlyApi(
  fakeConfigurationService,
  mockNetworkService,
  new HttpErrorFactory(),
);
```

src/routes/recovery/recovery.service.ts
fmrsabino: With the "enforcement" that I mentioned it was to consider the following:

```typescript
displayName?: `${string}:${string}:${string}`;
```

instead of

```typescript
displayName?: string;
```

What do you think?

src/routes/recovery/recovery.controller.spec.ts
iamacook: I've changed it in 1b7d94a3cade2cdf3d86e5c12f982c2099259eec. I will alter the other tests to use that in a separate PR too.

src/datasources/alerts-api/tenderly-api.service.spec.ts
iamacook: I see, my bad! I have changed to using a real implementation in 1b7d94a3cade2cdf3d86e5c12f982c2099259eec.

src/routes/recovery/recovery.service.ts
iamacook: It's much better! I've made the types scricter in 1b7d94a3cade2cdf3d86e5c12f982c2099259eec.

src/datasources/alerts-api/tenderly-api.service.spec.ts
```diff
@@ -12,14 +12,10 @@ const networkService = {
...
 
-const httpErrorFactory = {
-  from: jest.fn(),
-} as unknown as HttpErrorFactory;
-const mockHttpErrorFactory = jest.mocked(httpErrorFactory);
-
 describe('TenderlyApi', () => {
   let service: TenderlyApi;
   let fakeConfigurationService: FakeConfigurationService;
+  const httpErrorFactory = new HttpErrorFactory();
```
fmrsabino, line 18: I would instantiate this on the `beforeEach` since if this class gets a state of some sort in the future, that state will be shared across each test which might make some of them fail.

src/datasources/alerts-api/tenderly-api.service.spec.ts
```diff
@@ -55,26 +51,35 @@ describe('TenderlyApi', () => {
         new TenderlyApi(
           fakeConfigurationService,
           mockNetworkService,
-          mockHttpErrorFactory,
+          httpErrorFactory,
```
fmrsabino, line 54: You can provide a new instance here also.

src/datasources/alerts-api/tenderly-api.service.spec.ts
```diff
@@ -55,26 +51,35 @@ describe('TenderlyApi', () => {
...
         ),
     ).toThrow();
   });
 
   it('should add contracts', async () => {
+    const httpErrorFactoryFromSpy = jest.spyOn(httpErrorFactory, 'from');
```
fmrsabino, line 60: nitpick: I would say that we don't need to apply the `spyOn` for the scope of these tests.

We can work with the end result that is returned by the HttpErrorFactory.

fmrsabino (changes requested):
Some small requests but looks good overall! 🎉

src/datasources/alerts-api/tenderly-api.service.spec.ts
iamacook: It's now reinstantiated in `beforeEach` as of 8125503817eab4802917485d9c30ac52169a9040.

src/datasources/alerts-api/tenderly-api.service.spec.ts
iamacook: I've removed the spy and expectation in 8125503817eab4802917485d9c30ac52169a9040 as the response would otherwise throw, hence it's not needed.

fmrsabino (approved):
🚀

fmrsabino (approved):

PR #804, opened 2023-10-27
Execute nested transfers mappings sequentially

hectorgomezv:
- Executes `IncomingTransferMapper.mapTransfer` sequentially, so prevents the case where N mappings referring to the same token/contract address are dispatched in parallel, before caching the token/contract.


iamacook (approved):
Other than the conflict, LGTM!

iamacook (approved):

PR #821, opened 2023-11-01
feat: decode `multiSend` alerts

iamacook:
This adds `multiSend` handling to the `AlertsRepository` via a new `MultiSendDecoder` that [decodes `multiSend` transactions accordingly](https://github.com/safe-global/safe-contracts/blob/main/contracts/libraries/MultiSendCallOnly.sol#L13-L19). Associated test coverage has also been added.


src/domain/alerts/__tests__/safe-transactions.encoder.ts
```diff
@@ -0,0 +1,95 @@
...
+  },
+): Hex {
+  const ABI = parseAbi(['function changeThreshold(uint256 _threshold)']);
+
+  return encodeFunctionData({
+    abi: ABI,
+    functionName: 'changeThreshold',
+    args: [_threshold],
+  });
+}
```
iamacook, line 95: I intend to refactor this into a helper similar to `Builder` in a follow up PR.

src/domain/alerts/mappers/multi-send.mapper.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    const multiSend = decodeFunctionData({
+      abi: this.abi,
+      data: multiSendData,
+    });
+    const transactions = multiSend.args[0];
+
+    let i = 0;
+
+    while (i < size(transactions)) {
```
hectorgomezv, line 40: Nitpick: `size(transactions)` can be calculated in advance and saved to a `const`, instead of evaluating it on each iteration.

src/domain/alerts/mappers/multi-send.mapper.ts
```diff
@@ -0,0 +1,58 @@
...
+      // uint8 operation, address to, value uint256, dataLength uint256, bytes data
+      const operation = slice(transactions, i, (i += 1));
+      const to = slice(transactions, i, (i += 20));
+      const value = slice(transactions, i, (i += 32));
+      const dataLength = slice(transactions, i, (i += 32));
```
hectorgomezv, line 45: I'd extract these numeric values to class variables instead, something like:
```
private static readonly OPERATION_END = 1;
private static readonly TO_END = 20;
private static readonly VALUE_END = 32;
private static readonly DATA_LENGTH_END = 32;
```

Wdyt?

src/domain/alerts/mappers/multi-send.mapper.ts
```diff
@@ -0,0 +1,58 @@
...
+      data: Hex;
+    }> = [];
+
+    const multiSend = decodeFunctionData({
+      abi: this.abi,
+      data: multiSendData,
+    });
+    const transactions = multiSend.args[0];
+
+    let i = 0;
```
hectorgomezv, line 38: What do you think about having a more explicit name for this variable, like `cursor`, `pointer`, etc.?

src/domain/alerts/mappers/multi-send.mapper.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    while (i < size(transactions)) {
+      // uint8 operation, address to, value uint256, dataLength uint256, bytes data
+      const operation = slice(transactions, i, (i += 1));
+      const to = slice(transactions, i, (i += 20));
+      const value = slice(transactions, i, (i += 32));
+      const dataLength = slice(transactions, i, (i += 32));
+      const data = slice(transactions, i, (i += hexToNumber(dataLength)));
```
hectorgomezv, line 46: Loop termination is assured here, right? I mean, I guess `hexToNumber(dataLength)` can't be negative under any circumstance (which would lead to a potential infinite loop).

src/domain/alerts/mappers/multi-send.mapper.ts
```diff
@@ -0,0 +1,58 @@
...
+  Hex,
+  hexToBigInt,
+  hexToNumber,
+  parseAbi,
+  size,
+  slice,
+} from 'viem';
+
+@Injectable()
+export class MultiSendMapper {
```
hectorgomezv, line 14: No strong opinion, but we already have _mappers_ in the project with a slightly different purpose than this class: to aggregate/transform domain entities into route/presentation entities. What about using another suffix for this one? I think _decoder_ could be an option. I left this up to you. If you prefer to maintain the _mapper_ suffix, I think it's also fine.

hectorgomezv (dismissed):
Excellent work! I've left some nitpick comments 🙂

src/domain/alerts/mappers/multi-send.mapper.ts
iamacook: I've moved it to a `const` (`transactionsSize`) in f78093192d8c7d5555f13003c7548c368e3a20b0.

src/domain/alerts/mappers/multi-send.mapper.ts
iamacook: It's much more readable. I've moved them in f78093192d8c7d5555f13003c7548c368e3a20b0, opting for `*_SIZE` to be consistent.

src/domain/alerts/mappers/multi-send.mapper.ts
iamacook: I've changed it to `cursor` in f78093192d8c7d5555f13003c7548c368e3a20b0.

src/domain/alerts/mappers/multi-send.mapper.ts
iamacook: Yes, `dataLength` cannot ever be negative.

src/domain/alerts/mappers/multi-send.mapper.ts
iamacook: I've renamed this to `MultiSendDecoder` and moving the files/adjusted the tests accordingly in f78093192d8c7d5555f13003c7548c368e3a20b0.

hectorgomezv (approved):

PR #827, opened 2023-11-03
Add email-sending feature

hectorgomezv:
Domain changes:
- Adds `IEmailApi`, which exposes a function to interact with the underlying API implementation.
- Adds `CreateEmailMessageDto` a DTO entity that groups the required fields to create a new email message.

Datasource changes:
- Adds `PushwooshApi` which implements `IEmailApi` and maps the required fields into an HTTP request according to the Pushwoosh documentation.

Configuration changes:
- Adds email provider-related configuration under `email.*` key.

### DIFF
--- a/src/datasources/email-api/pushwoosh-api.service.spec.ts
+++ b/src/datasources/email-api/pushwoosh-api.service.spec.ts
@@ -5,6 +5,7 @@ import { INetworkService } from '@/datasources/network/network.service.interface
 import { CreateEmailMessageDto } from '@/domain/email/entities/create-email-message.dto.entity';
 import { DataSourceError } from '@/domain/errors/data-source.error';
 import { faker } from '@faker-js/faker';
+import { v4 as uuidv4 } from 'uuid';
 
 const networkService = {
   post: jest.fn(),
@@ -109,7 +110,6 @@ describe('PushwooshApi', () => {
           auth: pushwooshApiKey,
           notifications: [
             {
-              transactionId: expect.any(String),
               send_date: 'now',
               email_template: createEmailMessageDto.template,
               devices: createEmailMessageDto.to,
@@ -117,6 +117,44 @@ describe('PushwooshApi', () => {
               subject: [{ default: createEmailMessageDto.subject }],
               dynamic_content_placeholders: createEmailMessageDto.substitutions,
               from: { name: pushwooshFromName, email: pushwooshFromEmail },
+            },
+          ],
+        },
+      },
+    );
+  });
+
+  it('should create a new unique (using emailMessageId field) message', async () => {
+    const createEmailMessageDto: CreateEmailMessageDto = {
+      to: [faker.internet.email(), faker.internet.email()],
+      template: faker.string.uuid(),
+      subject: faker.string.sample(),
+      substitutions: {
+        [faker.string.sample()]: faker.string.sample(),
+        [faker.string.sample()]: faker.string.sample(),
+      },
+      emailMessageId: uuidv4(),
+    };
+
+    await service.createMessage(createEmailMessageDto);
+
+    expect(networkService.post).toHaveBeenCalledTimes(1);
+    expect(networkService.post).toHaveBeenCalledWith(
+      `${pushwooshBaseUri}/json/1.3/createEmailMessage`,
+      {
+        request: {
+          application: pushwooshApplicationCode,
+          auth: pushwooshApiKey,
+          notifications: [
+            {
+              send_date: 'now',
+              email_template: createEmailMessageDto.template,
+              devices: createEmailMessageDto.to,
+              use_auto_registration: true,
+              subject: [{ default: createEmailMessageDto.subject }],
+              dynamic_content_placeholders: createEmailMessageDto.substitutions,
+              from: { name: pushwooshFromName, email: pushwooshFromEmail },
+              transactionId: createEmailMessageDto.emailMessageId,
             },
           ],
         },
--- a/src/datasources/email-api/pushwoosh-api.service.spec.ts
+++ b/src/datasources/email-api/pushwoosh-api.service.spec.ts
@@ -3,6 +3,7 @@ import { PushwooshApi } from '@/datasources/email-api/pushwoosh-api.service';
 import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
 import { INetworkService } from '@/datasources/network/network.service.interface';
 import { CreateEmailMessageDto } from '@/domain/email/entities/create-email-message.dto.entity';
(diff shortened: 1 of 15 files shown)


src/datasources/email-api/pushwoosh-api.service.spec.ts
```diff
@@ -0,0 +1,129 @@
...
+const networkService = {
+  post: jest.fn(),
+  delete: jest.fn(),
+} as unknown as INetworkService;
+const mockNetworkService = jest.mocked(networkService);
+
+const httpErrorFactory = {
+  from: jest.fn(),
+} as unknown as HttpErrorFactory;
+const mockHttpErrorFactory = jest.mocked(httpErrorFactory);
```
fmrsabino, line 18: I would avoid mocking our own classes at this level as by testing with the real one with are more confident on how the component behaves while deployed.

fmrsabino (changes requested):

src/datasources/email-api/pushwoosh-api.service.ts
```diff
@@ -0,0 +1,66 @@
...
+        request: {
+          application: this.applicationCode,
+          auth: this.apiKey,
+          notifications: [
+            {
+              transactionId: uuidv4(),
+              send_date: 'now',
+              email_template: createEmailMessageDto.template,
+              devices: createEmailMessageDto.to,
+              use_auto_registration: true,
```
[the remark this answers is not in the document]
hectorgomezv: `use_auto_registration` registers the email address into the provider systems and sends the email message to that address in one step. Source: https://docs.pushwoosh.com/platform-docs/api-reference/email-api#createemailmessage

I'm going to include some comments here explaining non-obvious properties 👍🏻

src/domain/email/entities/create-email-message.dto.entity.ts
```diff
@@ -0,0 +1,6 @@
+export class CreateEmailMessageDto {
+  to: string[];
+  template: string;
+  subject: string;
+  substitutions: Record<string, string> | null;
+}
```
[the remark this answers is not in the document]
hectorgomezv: I included that documentation into [another PR](https://github.com/safe-global/safe-client-gateway/pull/871/files) where I added another function to the interface, but I'm going to include it here too.

src/datasources/email-api/pushwoosh-api.service.ts
```diff
@@ -0,0 +1,66 @@
...
+  ): Promise<void> {
+    try {
+      const url = `${this.baseUri}/json/1.3/createEmailMessage`;
+      await this.networkService.post(url, {
+        request: {
+          application: this.applicationCode,
+          auth: this.apiKey,
+          notifications: [
+            {
+              transactionId: uuidv4(),
```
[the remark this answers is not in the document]
hectorgomezv: Providing a unique identifier here prevents sending a message twice, according to the provider docs: https://api.pushwoosh.com/json/1.3/createEmailMessage

I'm going to document this on our side too 👍🏻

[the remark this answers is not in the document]
hectorgomezv: Added some refs/comments in [450b05c](https://github.com/safe-global/safe-client-gateway/pull/827/commits/450b05c653c741ecca3f181d55a27c308d405c12), let me know if you think we should add more detail.

src/domain/email/entities/create-email-message.dto.entity.ts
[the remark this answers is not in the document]
hectorgomezv: Added in [450b05c](https://github.com/safe-global/safe-client-gateway/pull/827/commits/450b05c653c741ecca3f181d55a27c308d405c12)

src/datasources/email-api/pushwoosh-api.service.ts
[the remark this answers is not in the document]
hectorgomezv: Added some comments in [450b05c](https://github.com/safe-global/safe-client-gateway/pull/827/commits/450b05c653c741ecca3f181d55a27c308d405c12), let me know if you think it's not enough.

src/datasources/email-api/pushwoosh-api.service.spec.ts
hectorgomezv: I agree! Changed in [450b05c](https://github.com/safe-global/safe-client-gateway/pull/827/commits/450b05c653c741ecca3f181d55a27c308d405c12)

src/datasources/email-api/pushwoosh-api.service.ts
```diff
@@ -0,0 +1,71 @@
...
+  ): Promise<void> {
+    try {
+      const url = `${this.baseUri}/json/1.3/createEmailMessage`;
+      await this.networkService.post(url, {
+        request: {
+          application: this.applicationCode, // application code, should exist on Pushwoosh
+          auth: this.apiKey,
+          notifications: [
+            {
+              transactionId: uuidv4(), // unique identifier to avoid messages duplication
```
fmrsabino, line 55: I think this should be provided as an argument of `createMessage` or be part of `CreateEmailMessageDto`.

It would be up to the consumers of this API to know what is considered a unique event or not. By generating an id here, every call to `createMessage` will be considered unique.

hectorgomezv: I was considering this internal to the provider, as an implementation detail. From an API consumer point of view, each email message should be meant to be sent once. Do you see a scenario where the consumer wants to allow the provider to send the same email message potentially N times?

hectorgomezv: If I understand the provider documentation correctly, this is more to prevent the provider from sending the same message several times for no reason. I'm not even sure why this is a parameter in their API, seems like a mechanism to circumvent a malfunction on their side, not sure why this is not Pushwoosh internal 🤔

fmrsabino: I don't see it as internal because it's part of the public API. Else this would be generated on their side or?

hectorgomezv: I meant internal to our `PushwooshApi` class, not the provider REST API.

But giving this a second thought, adding this unique identifier as a parameter as you suggested won't hurt, and it adds some degree of flexibility. I'm going to change this 👍

hectorgomezv: I've added a field to give the clients the opportunity to send an unique identifier into `CreateEmailMessageDto` in [bd7cf93](https://github.com/safe-global/safe-client-gateway/pull/827/commits/bd7cf93077705307fe3e01efc758637736e90b31)

fmrsabino (approved):
🚀

PR #828, opened 2023-11-03
Increase TTL for not found token prices

hectorgomezv:
**Motivation**
This PR reduces the amount of calls to our token prices provider by increasing the TTL for the addresses that can't be retrieved. So subsequent retries for these addresses are delayed more than the usual token price refreshes.

**Included changes**
- Adds a function to `ICacheService` to allow setting a TTL for a given cache key.
- Injects `CacheService` into `CoingeckoApi`, to make the latter able to register prices as not found.
- Adds logic to `PricesRepositiory` to mark a token price as not found, and therefore adds an extended TTL.

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -81,6 +81,9 @@ export default () => ({
       'https://api.coingecko.com/api/v3',
     apiKey: process.env.PRICES_PROVIDER_API_KEY,
     pricesTtlSeconds: parseInt(process.env.PRICES_TTL_SECONDS ?? `${300}`),
+    notFoundPriceTtlSeconds: parseInt(
+      process.env.NOT_FOUND_PRICE_TTL_SECONDS ?? `${259200}`,
+    ),
     chains: {
       1: { nativeCoin: 'ethereum', chainName: 'ethereum' },
       10: { nativeCoin: 'ethereum', chainName: 'optimistic-ethereum' },
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -82,7 +82,7 @@ export default () => ({
     apiKey: process.env.PRICES_PROVIDER_API_KEY,
     pricesTtlSeconds: parseInt(process.env.PRICES_TTL_SECONDS ?? `${300}`),
     notFoundPriceTtlSeconds: parseInt(
-      process.env.NOT_FOUND_PRICE_TTL_SECONDS ?? `${259200}`,
+      process.env.NOT_FOUND_PRICE_TTL_SECONDS ?? `${72 * 60 * 60}`,
     ),
     chains: {
       1: { nativeCoin: 'ethereum', chainName: 'ethereum' },
--- a/src/datasources/cache/cache.service.interface.ts
+++ b/src/datasources/cache/cache.service.interface.ts
@@ -15,5 +15,5 @@ export interface ICacheService {
 
   deleteByKeyPattern(pattern: string): Promise<void>;
 
-  expire(key: string, expireTimeSeconds: number);
+  expire(key: string, expireTimeSeconds: number): Promise<void>;
 }
--- a/src/datasources/cache/cache.service.interface.ts
+++ b/src/datasources/cache/cache.service.interface.ts
@@ -14,4 +14,6 @@ export interface ICacheService {
   deleteByKey(key: string): Promise<number>;
 
   deleteByKeyPattern(pattern: string): Promise<void>;
+
+  expire(key: string, expireTimeSeconds: number);
 }
--- a/src/datasources/prices-api/coingecko-api.service.ts
+++ b/src/datasources/prices-api/coingecko-api.service.ts
@@ -5,6 +5,10 @@ import { AssetPrice } from '@/domain/prices/entities/asset-price.entity';
 import { CacheFirstDataSource } from '../cache/cache.first.data.source';
 import { CacheRouter } from '../cache/cache.router';
 import { DataSourceError } from '@/domain/errors/data-source.error';
+import {
+  CacheService,
+  ICacheService,
+} from '@/datasources/cache/cache.service.interface';
 
 @Injectable()
 export class CoingeckoApi implements IPricesApi {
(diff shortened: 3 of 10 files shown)


fmrsabino (changes requested):
Some questions (raised internally) still require some clarification before this solution can be reviewed.

hectorgomezv:
Converting this to draft. Some discussions that may affect this PR are still pending.

src/config/entities/configuration.ts
```diff
@@ -81,6 +81,9 @@ export default () => ({
       'https://api.coingecko.com/api/v3',
     apiKey: process.env.PRICES_PROVIDER_API_KEY,
     pricesTtlSeconds: parseInt(process.env.PRICES_TTL_SECONDS ?? `${300}`),
+    notFoundPriceTtlSeconds: parseInt(
+      process.env.NOT_FOUND_PRICE_TTL_SECONDS ?? `${259200}`,
```
fmrsabino, line 85: I would say that having `${70 * 60 * 60}` is more readable.

src/datasources/cache/cache.service.interface.ts
```diff
@@ -14,4 +14,6 @@ export interface ICacheService {
   deleteByKey(key: string): Promise<number>;
 
   deleteByKeyPattern(pattern: string): Promise<void>;
+
+  expire(key: string, expireTimeSeconds: number);
```
fmrsabino, line 18: nitpick: return type is missing here

fmrsabino, line 18: Since we now have `expire` as a separate function what do you think about removing that functionality from `set` in order to make these operations more granular? (for another PR)

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -132,6 +141,15 @@ export class CoingeckoApi implements IPricesApi {
...
   }
 
+  async registerNotFoundTokenPrice(args: {
+    chainName: string;
+    tokenAddress: string;
+    fiatCode: string;
+  }): Promise<void> {
+    const cacheDir = CacheRouter.getTokenPriceCacheDir(args);
+    return this.cacheService.expire(cacheDir.key, this.notFoundPriceTtlSeconds);
+  }
```
fmrsabino, line 151: What's the status code returned when a token is not found? The `CacheFirstDataSource` has an expiration time that for those cases.

src/domain/interfaces/prices-api.interface.ts
```diff
@@ -15,4 +15,10 @@ export interface IPricesApi {
   }): Promise<AssetPrice>;
 
   getFiatCodes(): Promise<string[]>;
+
+  registerNotFoundTokenPrice(args: {
+    chainName: string;
+    tokenAddress: string;
+    fiatCode: string;
+  }): Promise<void>;
```
fmrsabino, line 23: Is it possible to move this logic so that it's internal to the datasource? I don't think the domain should be the one deciding when to call this.

fmrsabino (changes requested):

src/datasources/cache/cache.service.interface.ts
hectorgomezv: Yes, I think it makes perfect sense, we can do that in another PR, thanks! 👍

src/datasources/prices-api/coingecko-api.service.ts
hectorgomezv: The Coingecko API always returns a 200 status code even if the token is not found. That makes it harder to use the API for this use case of requesting the token prices one by one. It simply returns an empty object as payload in those cases. That's why checking the response actually contains the price is done in the repository.

My best guess around this design decision on the Coingecko side was taken because the endpoint expects several addresses to be sent in each call (as I'm implementing on another PR). So some addresses may be found and others may not, and the response is always a 200 containing an object holding a subset of the addresses requested.

src/domain/interfaces/prices-api.interface.ts
hectorgomezv: Yes, I also thought about that. Finally I left this on the domain because as I commented in another comment, there's no way to check if the response actually contains the price without processing it (the status code is always 200). So moving this to the datasources layer would imply moving the response validation there.

src/datasources/prices-api/coingecko-api.service.ts
fmrsabino: OK makes sense. Thanks for explaining. Since this should not be considered a final solution I would say that it's fine and we can re-work the domain/datasource API on a second iteration!

src/domain/interfaces/prices-api.interface.ts
fmrsabino: Understood! Can you leave a comment on this function explaining why it exists? Compared to the other functions this seems a bit a out of place 😬

src/config/entities/configuration.ts
hectorgomezv: Changed in [7538cd9](https://github.com/safe-global/safe-client-gateway/pull/828/commits/7538cd923d47b7eaefbfde53d1482a36af131894)

src/datasources/cache/cache.service.interface.ts
hectorgomezv: Fixed in [7538cd9](https://github.com/safe-global/safe-client-gateway/pull/828/commits/7538cd923d47b7eaefbfde53d1482a36af131894)

src/domain/interfaces/prices-api.interface.ts
hectorgomezv: Added in [7538cd9](https://github.com/safe-global/safe-client-gateway/pull/828/commits/7538cd923d47b7eaefbfde53d1482a36af131894)

I hope it's understandable 🙂 Thanks!

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -17,6 +17,11 @@ export class CoingeckoApi implements IPricesApi {
    */
   private static readonly pricesProviderHeader: string = 'x-cg-pro-api-key';
 
+  /**
+   * Time range in seconds used to get a random value when calculating a TTL for not-found token prices.
+   */
+  static readonly notFoundTtlRange: number = 60 * 60 * 24;
```
fmrsabino, line 23: nitpick: I would add `Seconds` to the constant name so that we know the time unit that this represents. So `notFoundTtlRangeSeconds`

fmrsabino (approved):
🚀

PR #829, opened 2023-11-06
Add Emails Datasource

fmrsabino:
## Datasource

Adds a `IEmailDataSource` – this component can be used to interact with the underlying data source which stores the email addresses for a given signer, for a given safe address on a specific chain. The following functionality was added:
  * `saveEmail` – saves an email address in the database. Returns a verification code if it was successfully stored.
  * `setVerificationCode` – resets the verification code for a signer. Returns the new verification code if successful.
  * `verifyEmail` – verifies a specific email given the code that was returned on registration. If the code matches then the email is considered verified. There's a configurable expiration time for the code that was introduced for security reasons – the code needs to be verified within a given timeframe to be considered valid. If the code is not verified in this timeframe then it should be reset.

## Database

- Adds one table – `signer_emails` under the `emails` schema.

The `signer_emails` table is represented as follows:

| id | chain_id | email_address | safe_address | signer | verified | verification_code | sent_on |
|--------|--------|--------|--------|--------|--------|--------|--------|
| SERIAL PRIMARY KEY | int NOT NULL | text NOT NULL | character varying(42) NOT NULL | character varying(42) NOT NULL | boolean NOT NULL DEFAULT false | text |  timestamp with time zone

- A unique constraint exists for entries on `signer_emails`: an entry is considered unique when the following fields are considered: `chain_id`, `safe_address`, `signer`.

## Migration Support

Migration support was added. Migration files need to be under the `migrations` folder. Inside it, a migration folder should be created with 5 digits and under that folder, an `index.sql` containing the migration file. 

For example, the initial migration has the following structure:

```
migrations/
└── 00001_initial
    └── index.sql
```

The initial migration creates the `signer_emails` table. The migration is done whenever the `PostgresDatabaseModule` is initialised (via the `PostgresDatabaseMigrationHook`).

**Note on running multiple instances of the service:** an advisory lock is acquired so that only one instance performs the migration. Other instances will wait for the lock to be released (via `pg_advisory_lock`). **Migrations should be idempotent** so that the final state of the database is the same upon re-execution of the migration file. See https://postgresql.org/docs/16/interactive/explicit-locking.html#ADVISORY-LOCKS

## Docker compose support

- Adds a `db-test` container to Docker Compose – this way, given a development environment, two databases can be launched: the `db` one with an explicit mapped volume for development  (`/data` folder) and the `db-test` with no explicit mapped volumes which can be used for testing.
- The `db-test` container has different environment variable to setup its own location and default user. By default `db` will be exposed on port `5432` of the host while `db-test` will be exposed on port `5433`.

## Out of scope

- The code generation should be considered out of scope of this Pull Request – the one introduced here is used merely for demonstration purposes and should not be considered production ready.
- The domain layer is out of scope.
- The routes layer is out of scope.
- Database security/encryption mechanisms.


src/datasources/email/email.datasource.ts
```diff
@@ -0,0 +1,157 @@
...
+        sql,
+        email.id,
+      );
+
+      return <EmailVerificationCode>{
+        emailAddress: email.email_address,
+        verificationCode: verificationStatus.verification_code,
+      };
+    });
+  }
```
iamacook, line 76: I assume that we will later implement the signature checks for these methods?

src/datasources/email/errors/expired-verification-code.error.ts
```diff
@@ -0,0 +1,10 @@
+import { Email } from '@/datasources/email/entities/email.entity';
+import { VerificationStatus } from '@/datasources/email/entities/verification.status';
+
+export class ExpiredVerificationCodeError extends Error {
+  constructor(email: Email, verificationStatus: VerificationStatus) {
+    super(
+      `Verification code for ${email.email_address} expired. Code: ${verificationStatus.verification_code} `,
```
iamacook, line 7:
```suggestion
      `Verification code for ${email.email_address} expired. Code: ${verificationStatus.verification_code}`,
```

src/datasources/email/email.datasource.ts
```diff
@@ -0,0 +1,157 @@
...
+      'email.verificationExpireTimeInSeconds',
+    );
+  }
+
+  private _resetVerificationCode(
+    sql: postgres.Sql,
+    emailId: number,
+  ): postgres.PendingQuery<VerificationStatus[]> {
+    // TODO proper code generation
+    const code = Math.floor(Math.random() * 100_000);
```
iamacook, line 31: Out of scope, but we need to consider how the email provider handles verification as well (if they do).

src/datasources/email/email.datasource.ts
fmrsabino: Yes but at which level is still up for discussion. The functionality in the Email Datasource can be provided without these checks and we assume that the component will be used by one which will validate the signatures. This would decouple the database operations with the authentication part.

src/datasources/db/migration.helper.js
```diff
@@ -0,0 +1,84 @@
+// Taken from https://github.com/porsager/postgres-shift
```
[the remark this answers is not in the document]
fmrsabino: Wasn't aware of `patch`. It does seem that Yarn offers it though: https://yarnpkg.com/cli/patch. I will look into it 👍

.yarn/patches/postgres-shift-npm-0.1.0-9342b5f6f6.patch
```diff
@@ -0,0 +1,38 @@
...
++module.exports = async function({
+   sql,
+   path = join(process.cwd(), 'migrations'),
+   before = null,
+@@ -48,7 +48,7 @@ export default async function({
+   }) {
+     fs.existsSync(join(path, 'index.sql')) && !fs.existsSync(join(path, 'index.js'))
+       ? await sql.file(join(path, 'index.sql'))
+-      : await import(join(path, 'index.js')).then(x => x.default(sql)) // eslint-disable-line
++      : await require(join(path, 'index.js'))(sql) // eslint-disable-line
```
iamacook, line 23: We don't need to `await` here anymore as `require` is syncronous.

```suggestion
+      : require(join(path, 'index.js'))(sql) // eslint-disable-line
```

fmrsabino: Interesting. I've actually used a previous non-ESM version of the library for this change https://github.com/porsager/postgres-shift/blob/c841edd6f1002db6713ce961e8339b298e8946e8/index.js#L51

iamacook: I see. Let's keep it as is then.

iamacook (approved):
There are some small conflicts but this otherwise LGTM 🚀🚀🚀

src/datasources/email/email.datasource.ts
fmrsabino: I agree. However I saw the provider as one to distribute emails only and not for having some kind of registration with the user. This is something that should be explored when the integration with the provider is done @hectorgomezv

src/datasources/email/email.datasource.ts
```diff
@@ -0,0 +1,157 @@
...
+      };
+    });
+  }
+
+  async resetVerificationCode(args: {
+    chainId: string;
+    safeAddress: string;
+    signer: string;
+  }): Promise<{ emailAddress: string; verificationCode: string }> {
+    const [email] = await this.getEmail(args);
```
hectorgomezv, line 83: This is getting the first (ordered by insertion date) `Email` in the database, is this correct? Is this function intended to be called without specifying the email address for the reset?

src/datasources/email/email.datasource.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  async verifyEmail(args: {
+    chainId: string;
+    safeAddress: string;
+    signer: string;
+    code: string;
+  }): Promise<void> {
+    return await this.sql.begin(async (sql) => {
+      const [email] = await this.getEmail(args);
```
hectorgomezv, line 115: Same concern as above: this is getting the first of potentially many rows matching the `args`.

src/datasources/email/email.datasource.spec.ts
```diff
@@ -0,0 +1,271 @@
...
+describe('Email Datasource Tests', () => {
+  let target: EmailDatasource;
+
+  const sql = postgres({
+    host: process.env.POSTGRES_TEST_HOST || 'localhost',
+    port: parseInt(process.env.POSTGRES_TEST_PORT || `${5433}`),
+    db: process.env.POSTGRES_TEST_DB || 'test-db',
+    user: process.env.POSTGRES_TEST_USER || 'postgres',
+    password: process.env.POSTGRES_TEST_PASSWORD || 'postgres',
+  });
```
hectorgomezv, line 23: What about providing this configuration via `__tests__/configuration.ts`? Randomly generated values for `db.postgres` on there won't be used. Plus, moving this configuration there would make it reusable by other database-related test files. No strong opinion on this, just suggesting.

hectorgomezv (changes requested):
Great job! I've left some small comments.

I also have a more general concern: this PR implements the _datasources layer_ part for dealing with emails and verifications. But I think some _business logic_ is being included here, like time-related calculations or conditional error checks. I know the _domain layer_ part is out of scope, so I'm curious about whether is it planned to move those things (and their tests) to the domain layer later or if it's considered part of the _datasources layer_.

src/datasources/email/email.datasource.ts
```diff
@@ -0,0 +1,157 @@
...
+  private getEmail(args: {
+    chainId: string;
+    safeAddress: string;
+    signer: string;
+  }): postgres.PendingQuery<Email[]> {
+    return this.sql<Email[]>`SELECT *
+                             FROM emails.signer_emails
+                             WHERE chain_id = ${args.chainId}
+                               and safe_address = ${args.safeAddress}
+                               and signer = ${args.signer}`;
```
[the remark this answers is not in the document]
fmrsabino: Isn't this more of an issue on how Visual Studio Code is rendering the file? Github seems to be rendering it correctly:

image


And IntelliJ also:

image

migrations/00001_initial/index.sql
```diff
@@ -0,0 +1,22 @@
...
+    safe_address  character varying(42) NOT NULL,
+    signer        character varying(42) NOT NULL,
+    verified      boolean               NOT NULL DEFAULT false,
+    UNIQUE (chain_id, email_address, safe_address, signer)
+);
+
+DROP TABLE IF EXISTS emails.verification CASCADE;
+CREATE table emails.verification
+(
+    id                SERIAL PRIMARY KEY REFERENCES emails.signer_emails ON DELETE CASCADE,
```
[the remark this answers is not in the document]
fmrsabino: > Is it intended to segment the data vertically?

Can you explain what you mean by this?

src/datasources/email/email.datasource.ts
fmrsabino: > This is getting the first (ordered by insertion date) `Email` in the database, is this correct? 

Yes.

> Is this function intended to be called without specifying the email address for the reset?

Yes. Since the combination of `signer`, `chainId` and `safeAddress` should be unique then the email doesn't need to be provided. Also this makes it easier for designing the client API – if an email was required for the code reset the frontend would need to store the email locally in order to provide it in different flows.

src/datasources/email/email.datasource.ts
fmrsabino: There's a `UNIQUE` constraint enforced on the database level. We can have a test that validates that the constraint was applied.

Are you suggesting validating the collection size in addition to that?

src/datasources/email/email.datasource.spec.ts
fmrsabino: It should be ok to move it there I agree. However we would need to call `configuration()` either way on this test since the `ConfigurationModule` is not part of this test.

fmrsabino:
> Great job! I've left some small comments.
> 
> I also have a more general concern: this PR implements the _datasources layer_ part for dealing with emails and verifications. But I think some _business logic_ is being included here, like time-related calculations or conditional error checks. I know the _domain layer_ part is out of scope, so I'm curious about whether is it planned to move those things (and their tests) to the domain layer later or if it's considered part of the _datasources layer_.

I completely agree with this. I got ahead of myself and started doing validations that I had in mind on the DataSource. I will remove those so overall the DataSource will become simples and perform less validations.

Those will be moved to the domain layer instead.

src/datasources/email/email.datasource.ts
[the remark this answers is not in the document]
hectorgomezv: It depends on the screen width, I took the screenshot from the Github side-by-side changes comparison on my laptop. I think an indentation like [this one](https://github.com/safe-global/safe-client-gateway/blob/047d3b7f31351b88b390880f724099eb099fc4c3/src/datasources/email/email.datasource.ts#L61) is preferable for readability, but we don't have established rules for SQL, so we can discuss the indentation later, no problem on my side 👍

migrations/00001_initial/index.sql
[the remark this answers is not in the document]
hectorgomezv: I mean, the only reason I can imagine for having this kind of 1:1 relationship between these two tables, having the primary keys linked, is to implement some kind of vertical partitioning (aka _row splitting_). Not sure this was the reason, just asking.

If it is not, what's the reasoning for having two separate tables? If `verification_code` and `sent_on` are associated with exactly 1 element of `emails.signer_emails`, why are they separated on a database level?

I think at the _domain level_ these could represent different concepts, but in the _datasouces layer_ they seem to be linked, and perhaps having them linked could simplify `EmailDataSource` because it will operate just one table. 

I don't have an strong opinion on this, but just commenting it out for your consideration 🙂

src/datasources/email/email.datasource.ts
hectorgomezv: I get the point regarding the client API, but I don't understand why unicity is guaranteed for these three elements. I see this in the initial migration:
```
UNIQUE (chain_id, email_address, safe_address, signer)
```
So the unique constraint includes the `email_address`, and therefore, we could have several sets of [`signer`, `chainId`, `safeAddress`] with different `email_address`.

src/datasources/email/email.datasource.ts
hectorgomezv: Same as in the above comment, the email address seems to be part of the unique constraint:
```
UNIQUE (chain_id, email_address, safe_address, signer)
```

Maybe I'm missing something 🤔

src/datasources/email/email.datasource.ts
fmrsabino: Ah yes this is an issue. the `email_address` should not be part of that group.

src/datasources/db/postgres-database.migration.hook.ts
```diff
@@ -0,0 +1,25 @@
...
+@Injectable({})
+export class PostgresDatabaseMigrationHook implements OnModuleInit {
+  constructor(
+    @Inject('DB_INSTANCE') private readonly db: postgres.Sql,
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
+  ) {}
+
+  async onModuleInit() {
+    this.loggingService.info('Checking migrations');
+    await shift({ sql: this.db });
```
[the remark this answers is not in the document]
fmrsabino: > This is probably out of the scope of the PR, but have you thought about what would be the proper behavior if an error happens while running the migrations (inside `shift`)? Maybe an error at this level should prevent the service from starting.

Promise should be rejected but I will test this condition. Good point 👍 

> Also an additional concern regarding this (sorry for the delay between comments): should we prevent the case where several instances of the service execute the migrations when starting? (is there a database lock or other mechanism that prevents concurrent execution?)

The migrations are executed inside a transaction but after looking more into it I think we should have a lock acquired on the application level. Advisory locks seem to be a good fit for this – https://postgresql.org/docs/16/interactive/explicit-locking.html#ADVISORY-LOCKS.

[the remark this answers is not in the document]
fmrsabino: I've implemented the lock here: [7fbe2cb](https://github.com/safe-global/safe-client-gateway/pull/829/commits/7fbe2cb56845e474800d93d9952e58c4ebf11e23)

Application also fails to start if the migration is not executed correctly.

One issue though is that if we have two instances and the following happens:
1. Instance 1 acquires lock
2. Instance 2 fails to acquire lock and returns
3. Instance 1 fails migrations and exits
4. Instance 2 is accepting connections

A possible solution would be to have instances always wait on the lock and perform the migration. This should be ok since migrations should check if a given table or operation was already executed. Given a single database instance the end result should be the same if the migrations are idempotent – either all instances fail or all are successful.

What do you think?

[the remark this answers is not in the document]
hectorgomezv: Good point. I guess we can assume the majority of migrations will be idempotent, we could make that a requirement for a migration. If we need to include a non-idempotent migration in the future for some reason, we would need to consider a different way of persisting its changes to the database. 

Regarding the scenario you described, I'd say we need to test carefully in dev environments what happens when a bad migration is executed and causes the service to stop the bootstrap sequence. I mean, thinking in Kubernetes terms, I guess the failing instance 1 gets restarted repeatedly, but instance 2 as you said accepts connections. I think this scenario is acceptable because at least we don't have a service disruption. But it's also true instance 2 executes code that expects the migration to have been applied.

It's a complex problem, so as I said maybe we need to actually test these error conditions in a dev environment to have a better picture of the implications of waiting/exiting/erroring, etc.

src/datasources/db/postgres-database.migration.hook.ts
```diff
@@ -0,0 +1,49 @@
...
+      await shift({ sql: this.sql });
+      this.loggingService.info('Pending migrations executed');
+    } catch (e) {
+      // If there's an error performing a migration, we should throw the error
+      // and prevent the service from starting
+      this.loggingService.error(e);
+      throw e;
+    } finally {
+      // the lock should be released if the migration completed (successfully or not)
+      await this
```
hectorgomezv, line 45: I think it might be a little JS gotcha here because the `finally` block will be executed even if the function returns on line 33. That will cause the lock to be released by an instance that didn't acquire the lock and it's not executing the migration. 

See: 
https://stackoverflow.com/questions/3837994/why-does-a-return-in-finally-override-try
https://stackoverflow.com/questions/345091/will-code-in-a-finally-statement-fire-if-i-return-a-value-in-a-try-block
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch#the_finally_block

src/datasources/db/postgres-database.migration.hook.ts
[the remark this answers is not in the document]
fmrsabino: Right now I'm more leaning towards every instance tries to perform the pending migrations. Since they should be idempotent, after the migration is executed we should see no changes from the other executions and they should terminate successfully.

In this case I would say that it's more on the deployment side that at least one instance should remain alive (previous service version) if the migrations were not completed successfully.

This implies changing from `pg_try_advisory_lock` (no lock wait) to `pg_advisory_lock` (lock wait).

migrations/00001_initial/index.sql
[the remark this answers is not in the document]
hectorgomezv: From my perspective, I think the current schema is valid and serves its purpose. But I also think having all this data in the same table would simplify things in the datasources layer, even though additional work might be necessary in the domain layer to split the data into two entities. Merging the columns in a single table will also benefit performance, as fewer queries would be necessary, but I think performance is not a concern for this use case.  So I left this up to you, I'm good with the current design, and even if we think we can improve it, it could be done later in another PR.

src/datasources/email/email.datasource.spec.ts
hectorgomezv: Ok, no strong opinion regarding this. We can consider it in the future if we find ourselves copypasting this code in another test class 👍🏻

src/datasources/db/postgres-database.migration.hook.ts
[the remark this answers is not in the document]
hectorgomezv: > In this case I would say that it's more on the deployment side that at least one instance should remain alive (previous service version) if the migrations were not completed successfully.

Agree 100%. We need to check this is the case (i.e. the service never responds OK to the readiness checks if the migrations fail).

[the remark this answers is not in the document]
fmrsabino: I've changed it in [877e31f](https://github.com/safe-global/safe-client-gateway/pull/829/commits/877e31fafa70ccbb0b08685188a446522a0782af)

migrations/00001_initial/index.sql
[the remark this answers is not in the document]
fmrsabino: Overall I agree. I was in between both solutions and decided to split into two table. I do think that we should leverage more the domain layer so if this simplifies our schemas I think it's a good change 👍 . I've merged both tables here: [e6503fb](https://github.com/safe-global/safe-client-gateway/pull/829/commits/e6503fbb11790b97147de898e6e5260875853376)

src/datasources/email/email.datasource.spec.ts
fmrsabino: I think it makes sense to tackle it now. Changed here [466eea5](https://github.com/safe-global/safe-client-gateway/pull/829/commits/466eea5dfa0ede5e66a16337a30739e31d0ab6f9).

Thanks for the suggestion! 🙌

fmrsabino:
Converted to draft as there were some core changes made and I will review this more carefully before pinging again for review.

migrations/00001_initial/index.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE table emails.signer_emails
+(
+    id                SERIAL PRIMARY KEY,
+    chain_id          int                   NOT NULL,
+    email_address     text                  NOT NULL,
+    safe_address      character varying(42) NOT NULL,
+    signer            character varying(42) NOT NULL,
+    verified          boolean               NOT NULL DEFAULT false,
+    verification_code text,
+    sent_on           timestamp with time zone       DEFAULT now() NOT NULL,
```
rmeissner, line 14: Personally I find this name confusing, as it is not clear what was sent.

fmrsabino: Yeah good point. Previously it was on the `verification` table so it made more sense due to the context. Will change it to something like `verification_sent_on`

fmrsabino: Changed in [e335038](https://github.com/safe-global/safe-client-gateway/pull/829/commits/e33503803281ae1ad0a48b4fe29f7f2f9d52adc1)

src/datasources/db/postgres-database.migration.hook.ts
fmrsabino: Does this comment still apply? I believe it was referencing the `return` statement inside the `try`.

hectorgomezv: After the last changes it does not apply anymore 👍

src/datasources/email/email.datasource.ts
```diff
@@ -0,0 +1,104 @@
...
+
+  async setVerificationCode(args: {
+    chainId: string;
+    safeAddress: string;
+    signer: string;
+    code: number;
+  }): Promise<EmailVerificationCode> {
+    const [email] = await this.sql<Email[]>`UPDATE emails.signer_emails
+                                            SET verification_code    = ${args.code},
+                                                verification_sent_on = now()
```
hectorgomezv, line 63: Nitpick: There is no need to tackle this on the PR, but we could consider setting this value (`verification_sent_on`) once the message is actually sent to the user's email. That would also provide insights to detect errors (i.e.: a null value here would indicate the email couldn't be sent for some reason). What do you think?

src/datasources/email/email.datasource.ts
```diff
@@ -0,0 +1,104 @@
...
+    if (!email) {
+      throw new EmailAddressDoesNotExistError(
+        args.chainId,
+        args.safeAddress,
+        args.signer,
+      );
+    }
+
+    return email;
+  }
```
hectorgomezv, line 32: This function is not used anymore after the refactor, right?

src/datasources/email/email.datasource.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+  afterEach(async () => {
+    await sql`TRUNCATE emails.signer_emails CASCADE`;
+  });
+
+  afterAll(async () => {
+    await sql.end();
+  });
+
+  it('stores email successfully', async () => {
+    const chainId = faker.number.int({ max: 2147483647 });
```
hectorgomezv, line 40: Nitpick: my guess is this number is chosen because it's the maximum positive value for a 32-bit field, but as it is repeated through several tests and the reasoning about it is not super obvious, what do you think about storing it in a variable (that will also provide a name explaining the choice)?

migrations/00001_initial/index.sql
```diff
@@ -0,0 +1,16 @@
...
+CREATE table emails.signer_emails
+(
+    id                   SERIAL PRIMARY KEY,
+    chain_id             int                      NOT NULL,
+    email_address        text                     NOT NULL,
+    safe_address         character varying(42)    NOT NULL,
+    signer               character varying(42)    NOT NULL,
+    verified             boolean                  NOT NULL DEFAULT false,
+    verification_code    text,
+    verification_sent_on timestamp with time zone NOT NULL DEFAULT now(),
```
hectorgomezv, line 14: As I commented in internal discussions, I'd consider not setting this default at the database level. I'm just leaving a comment here as well as a reminder.

hectorgomezv (changes requested):
Awesome job! 👏🏻👏🏻👏🏻 I've just left some small concerns.

src/datasources/email/email.datasource.spec.ts
fmrsabino: Exactly. An integer in Postgres goes up to this value – https://postgresql.org/docs/16/interactive/datatype-numeric.html#DATATYPE-NUMERIC.

Extracted to a `DB_CHAIN_ID_MAX_VALUE` to avoid copying the same value around. [641cb20](https://github.com/safe-global/safe-client-gateway/pull/829/commits/641cb204025e4ca4b2373dd5904ad99710b7da68)

src/datasources/email/email.datasource.ts
fmrsabino: Yes it's not longer required! Removed in [641cb20](https://github.com/safe-global/safe-client-gateway/pull/829/commits/641cb204025e4ca4b2373dd5904ad99710b7da68)

migrations/00001_initial/index.sql
fmrsabino: I agree. Removed the default for the `verification_sent_on`. This provides more flexibility on the domain layer on when this field should be set.

src/datasources/email/email.datasource.ts
fmrsabino: I've decided to tackle it here. I agree that decoupling the value that is set here and allowing it to be `NULL` would represent better the state of the verifications status 🙌  [641cb20](https://github.com/safe-global/safe-client-gateway/pull/829/commits/641cb204025e4ca4b2373dd5904ad99710b7da68)

hectorgomezv (approved):
🚀🚀🚀

PR #836, opened 2023-11-07
Batch mutiple token prices per request to Coingecko

hectorgomezv:
This PR adds batching to Coingecko API token price requests. Main change:
- `CoingeckoApi` uses an `INetworkService` directly to get the prices, instead of relying on `CacheFirstDataSource`. This way `CoingeckoApi` can attempt to retrieve several token prices from the `ICacheService` first, and request the not cached ones to `INetworkService` at once.

### DIFF
--- a/src/datasources/prices-api/coingecko-api.service.ts
+++ b/src/datasources/prices-api/coingecko-api.service.ts
@@ -24,6 +24,11 @@ export class CoingeckoApi implements IPricesApi {
   private static readonly pricesProviderHeader: string = 'x-cg-pro-api-key';
 
   /**
+   * Coingecko API maximum amount of token addresses being requested in the same call.
+   */
+  private static readonly maxBatchSize: number = 100;
+
+  /**
    * Time range in seconds used to get a random value when calculating a TTL for not-found token prices.
    */
   static readonly notFoundTtlRangeSeconds: number = 60 * 60 * 24;
@@ -99,40 +104,20 @@ export class CoingeckoApi implements IPricesApi {
     chainName: string;
     tokenAddresses: string[];
     fiatCode: string;
-  }): Promise<[string, number | null][]> {
-    const cachedAssetPrices = await this._getAssetPricesFromCache(args);
-    const notCachedAddresses = difference(
+  }): Promise<AssetPrice[]> {
+    const pricesFromCache = await this._getTokenPricesFromCache(args);
+    const notCachedTokenPrices = difference(
       args.tokenAddresses,
-      cachedAssetPrices.map((assetPrice) => assetPrice[0]),
+      pricesFromCache.map((assetPrice) => Object.keys(assetPrice)).flat(),
     );
-    const retrievedAssetPrices = await this._getAssetPricesFromProvider({
-      ...args,
-      notCachedAddresses,
-    });
+    const pricesFromNetwork = notCachedTokenPrices.length
+      ? await this._getTokenPricesFromNetwork({
+          ...args,
+          tokenAddresses: notCachedTokenPrices,
+        })
+      : [];
 
-    const assetPrices = await Promise.all(
-      notCachedAddresses.map(async (notCachedAddress) => {
-        const price =
-          retrievedAssetPrices.data[notCachedAddress]?.[args.fiatCode];
-        const assetPrice: [string, number | null] = [
-          notCachedAddress,
-          price ?? null,
-        ];
-        await this.cacheService.set(
-          CacheRouter.getTokenPriceCacheDir({
-            ...args,
-            tokenAddress: notCachedAddress,
-          }),
-          JSON.stringify(assetPrice),
-          price
-            ? this.pricesTtlSeconds
-            : this._getRandomNotFoundTokenPriceTtl(),
-        );
-        return assetPrice;
-      }),
-    );
-
-    return [cachedAssetPrices, assetPrices].flat();
+    return [pricesFromCache, pricesFromNetwork].flat();
   }
 
   async getFiatCodes(): Promise<string[]> {
@@ -161,28 +146,11 @@ export class CoingeckoApi implements IPricesApi {
     }
   }
 
-  private _mapProviderError(error: any): string {
-    const errorCode = error?.status?.error_code;
-    return errorCode ? ` [status: ${errorCode}]` : '';
-  }
-
-  /**
-   * Gets a random integer value between (notFoundPriceTtlSeconds - notFoundTtlRangeSeconds)
-   * and (notFoundPriceTtlSeconds + notFoundTtlRangeSeconds).
-   */
-  private _getRandomNotFoundTokenPriceTtl() {
-    const min =
-      this.notFoundPriceTtlSeconds - CoingeckoApi.notFoundTtlRangeSeconds;
-    const max =
-      this.notFoundPriceTtlSeconds + CoingeckoApi.notFoundTtlRangeSeconds;
-    return Math.floor(Math.random() * (max - min) + min);
-  }
-
-  private async _getAssetPricesFromCache(args: {
+  private async _getTokenPricesFromCache(args: {
     chainName: string;
     tokenAddresses: string[];
     fiatCode: string;
-  }): Promise<[string, number][]> {
+  }): Promise<AssetPrice[]> {
     const result = await Promise.all(
       args.tokenAddresses.map(async (tokenAddress) => {
         const cacheDir = CacheRouter.getTokenPriceCacheDir({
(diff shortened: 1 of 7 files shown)


src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -17,6 +23,11 @@ export class CoingeckoApi implements IPricesApi {
    */
   private static readonly pricesProviderHeader: string = 'x-cg-pro-api-key';
 
+  /**
+   * Coingecko API maximum amount of token addresses being requested in the same call.
+   */
+  private static readonly maxBatchSize: number = 100;
```
hectorgomezv, line 29: For batched requests to the Coingecko API, this `maxBatchSize` value is provided to avoid `414 - URI too large` style errors. (I've tested that kind of error happens with 150+ addresses, but this is set to `100` to stay conservative).

src/domain/interfaces/prices-api.interface.ts
```diff
@@ -8,23 +8,11 @@ export interface IPricesApi {
...
 
   getFiatCodes(): Promise<string[]>;
-
-  /**
-   * Registers a price reference for token address as not-found.
-   * This function allows the clients of this interface to signal that a price couldn't
-   * be parsed from its {@link AssetPrice}. This allows the underlying implementation
-   * to delay subsequent requests targeting the same token address.
-   */
-  registerNotFoundTokenPrice(args: {
```
hectorgomezv, line 25: This function is not used anymore since the clients of this interface don't need to signal the absence of a price anymore. (This is not encapsulated in the CoingeckoAPI implementation itself).

If we agree on the current state of this PR, `ICacheService.expire` won't be needed anymore as well, but I'm leaving that change to be done in a separated PR based on that outcome.

src/domain/balances/balances.repository.ts
```diff
@@ -83,12 +84,23 @@ export class BalancesRepository implements IBalancesRepository {
...
+
     return await Promise.all(
       simpleBalances.map(async (simpleBalance) => {
+        const tokenAddress = simpleBalance.tokenAddress?.toLowerCase() ?? null;
         const price = await this._getPrice(
           args.chainId,
           args.fiatCode,
-          simpleBalance.tokenAddress,
+          tokenAddress,
+          prices,
```
fmrsabino, line 103: Maybe this is more related with naming and overall structure but I find it a bit confusing that we get Token Prices (`_getTokenPrices`), for all the tokens in the balances, and then we call `_getPrice` for each token in the balances "again". I know that the functionality is different but pointing out that semantically it was a bit confusing.

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -146,19 +155,99 @@ export class CoingeckoApi implements IPricesApi {
...
+
+    return Promise.all(
+      args.tokenAddresses.map(async (tokenAddress) => {
+        const validPrice = prices.data[tokenAddress]?.[args.fiatCode];
+        const price: AssetPrice = validPrice
+          ? { [tokenAddress]: { [args.fiatCode]: validPrice } }
+          : { [tokenAddress]: { [args.fiatCode]: null } };
+        await this.cacheService.set(
+          CacheRouter.getTokenPriceCacheDir({ ...args, tokenAddress }),
+          JSON.stringify(price),
```
fmrsabino, line 213: To clarify: a `price` is a single token price correct?

The `AssetPrice` supports multiple tokens but we have changed that to return an array of `AssetPrice` where each `AssetPrice` is a single token correct?

fmrsabino (changes requested):

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -146,19 +155,99 @@ export class CoingeckoApi implements IPricesApi {
...
+  }): Promise<AssetPrice[]> {
+    const result = await Promise.all(
+      args.tokenAddresses.map(async (tokenAddress) => {
+        const cacheDir = CacheRouter.getTokenPriceCacheDir({
+          ...args,
+          tokenAddress,
+        });
+        const cached = await this.cacheService.get(cacheDir);
+        const { key, field } = cacheDir;
+        if (cached != null) {
```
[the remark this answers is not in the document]
hectorgomezv: I completely agree. It's more robust that way, thanks. 

I'll change the existent checks on `CacheFirstDataSource` ([this](https://github.com/safe-global/safe-client-gateway/blob/e51aa8b70accc2cfa5bb9545644e4843d8b72825/src/datasources/cache/cache.first.data.source.ts#L53)), if you agree, in a separate PR.

src/domain/balances/balances.repository.ts
hectorgomezv: Yes, I think that's a valid point. The first `_getTokenPrices` get the prices for all the tokens in advance, and then the second `_getPrice` _selects_ a price from that list. I agree with you this is a bit confusing semantically, I think we could just rename the second function. But I'm leaning towards just removing the private `_getPrice` function, as it's pretty simple, it has few lines, and its purpose is a bit blurry, so we can integrate it into the caller body.

src/datasources/prices-api/coingecko-api.service.ts
hectorgomezv: Yes, that's correct. I can include a comment in the code to clarify this part. `AssertPrice` can hold multiple tokens (it models the response from CoinGecko). I thought it would be easier for the clients of this class to receive an array of items instead of an object of 0..N keys so that complexity is hidden within the class.

Another option would be having different entities for the datasource part (CoinGecko mappings) and the domain. I thought this would add complexity but I'm open if you think it's better.

src/datasources/prices-api/coingecko-api.service.ts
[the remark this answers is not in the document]
fmrsabino: > I completely agree. It's more robust that way, thanks.
> 
> I'll change the existent checks on `CacheFirstDataSource` ([this](https://github.com/safe-global/safe-client-gateway/blob/e51aa8b70accc2cfa5bb9545644e4843d8b72825/src/datasources/cache/cache.first.data.source.ts#L53)), if you agree, in a separate PR.

Sounds good! 🙌

src/domain/balances/balances.repository.ts
fmrsabino: I will leave this up to you. I'm fine either way as long as the differences between the two "get prices" are understandable.

src/datasources/prices-api/coingecko-api.service.ts
fmrsabino: GIven the nature of this feature and how we are adapting this solution over time I would say that the way you have it is also ok for its purpose 👍

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -86,38 +99,34 @@ export class CoingeckoApi implements IPricesApi {
       );
     }
   }
-  async getTokenPrice(args: {
+
+  /**
+   * Gets prices for a set of token addresses, trying to get them from cache first.
+   * For those not found in the cache, it trying to retrieve them from the Coingecko API.
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [5a27454](https://github.com/safe-global/safe-client-gateway/pull/836/commits/5a274541184ed56b79ed8f5cffefc259092cc027)

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -86,38 +99,34 @@ export class CoingeckoApi implements IPricesApi {
...
+   *
+   * @param chainName Coingecko's name for the chain (see configuration)
+   * @param tokenAddresses Array of token addresses which prices are being retrieved
+   * @param fiatCode
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [5a27454](https://github.com/safe-global/safe-client-gateway/pull/836/commits/5a274541184ed56b79ed8f5cffefc259092cc027)

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -146,19 +155,99 @@ export class CoingeckoApi implements IPricesApi {
...
+        if (cached != null) {
+          this.loggingService.debug({ type: 'cache_hit', key, field });
+          return JSON.parse(cached);
+        }
+        this.loggingService.debug({ type: 'cache_miss', key, field });
+        return null;
+      }),
+    );
+
+    return result.filter((i) => i !== null);
```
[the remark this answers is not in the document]
hectorgomezv: Yes, that is clearer and more performant, changed in [5a27454](https://github.com/safe-global/safe-client-gateway/pull/836/commits/5a274541184ed56b79ed8f5cffefc259092cc027)

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -146,19 +155,99 @@ export class CoingeckoApi implements IPricesApi {
...
+            ? this.pricesTtlSeconds
+            : this._getRandomNotFoundTokenPriceTtl(),
+        );
+        return price;
+      }),
+    );
+  }
+
+  /**
+   * Does the actual price requests to the Coingecko API, using the {@link NetworkService}.
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [5a27454](https://github.com/safe-global/safe-client-gateway/pull/836/commits/5a274541184ed56b79ed8f5cffefc259092cc027)

src/datasources/prices-api/coingecko-api.service.ts
```diff
@@ -146,19 +155,99 @@ export class CoingeckoApi implements IPricesApi {
...
+    );
+
+    return result.filter((i) => i !== null);
   }
 
-  private _mapProviderError(error: any): string {
-    const errorCode = error?.status?.error_code;
-    return errorCode ? ` [status: ${errorCode}]` : '';
+  /**
+   * For an array of token addresses, gets the prices available from the Coingecko API.
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [5a27454](https://github.com/safe-global/safe-client-gateway/pull/836/commits/5a274541184ed56b79ed8f5cffefc259092cc027)

src/domain/balances/balances.repository.ts
hectorgomezv: I've integrated the function logic in its caller in [5a27454](https://github.com/safe-global/safe-client-gateway/pull/836/commits/5a274541184ed56b79ed8f5cffefc259092cc027)

src/datasources/prices-api/coingecko-api.service.ts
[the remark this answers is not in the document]
hectorgomezv: I was debugging this a bit further, and I found that `RedisCacheService` actually returns `null` when key is not found. So I'd say the return type for `ICacheService.get` ([link](https://github.com/safe-global/safe-client-gateway/blob/5a274541184ed56b79ed8f5cffefc259092cc027/src/datasources/cache/cache.service.interface.ts#L12)) is not accurate. 

Due to this, checking `cached !== undefined` won't work, and even `cached !== null` requires some other changes. So if you agree, I'm leaving this as it is (i.e. working), and I will create another PR tackling that change targeting this PR branch 🙂

PR #838, opened 2023-11-07
fix: rename `Contract` entity to `AlertsRegistration`

iamacook:
As per [suggestion](https://github.com/safe-global/safe-client-gateway/pull/801#discussion_r1383065721), this renames the alerts registration entity from `Contract` to `AlertsRegistration`. It could otherwise be confused with the `Contract` domain entity.


fmrsabino (approved):
🚀

PR #849, opened 2023-11-14
feat: add signature validator + demo email route

fmrsabino:


fmrsabino:
@iamacook closing down this Pull Request since this is being tackled via a combination of Nest Guards for the email feature. I would not delete the branch since it has more logic regarding signatures which can be used at a later stage.

PR #850, opened 2023-11-14
Make uuid dependency explicit

hectorgomezv:
This PR makes the `uuid` dependency (used into `app.module` class) explicit. That dependency was not declared before and therefore its version and availability were inherited from the `@nestjs/config` package.


fmrsabino (approved):
Nice catch! 🚀

PR #852, opened 2023-11-15
Add logs to cache hooks events

hectorgomezv:
Changes:
- Adds a log message for each event received by `CacheHooksService` in order to improve traceability.

### DIFF
--- a/src/routes/cache-hooks/cache-hooks.service.ts
+++ b/src/routes/cache-hooks/cache-hooks.service.ts
@@ -22,6 +22,8 @@ import { LoggingService, ILoggingService } from '@/logging/logging.interface';
 
 @Injectable()
 export class CacheHooksService {
+  private static readonly HOOK_TYPE = 'hook';
+
   constructor(
     @Inject(IBalancesRepository)
     private readonly balancesRepository: IBalancesRepository,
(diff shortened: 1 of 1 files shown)


src/routes/cache-hooks/cache-hooks.service.ts
```diff
@@ -275,14 +287,62 @@ export class CacheHooksService {
...
+        this._logEvent(event);
         break;
     }
     return Promise.all(promises);
   }
+
+  private _logSafeTxEvent(
+    event: ExecutedTransaction | NewConfirmation | PendingTransaction,
+  ): void {
+    this.loggingService.info({
```
fmrsabino, line 307: I would move this to `debug` level given the scope of this change.

fmrsabino, line 307: I would add a "type" for this log for grouping and filtering these logs.

fmrsabino (approved):

src/routes/cache-hooks/cache-hooks.service.ts
hectorgomezv: Added in [d628fae](https://github.com/safe-global/safe-client-gateway/pull/852/commits/d628fae78be188e82ba36dc8ad022fabe899e41b)

fmrsabino (approved):

src/routes/cache-hooks/cache-hooks.service.ts
hectorgomezv: Adding these lines to `debug` level, and enabling that level, would generate a lot of logging, and I think this change could be permanent so I'm more leaning towards using `info` if you agree 🙂

PR #853, opened 2023-11-15
Add OnlySafeOwner guard

fmrsabino:
The `OnlySafeOwner` is a guard is a guard that can be used to verify that the `account` provided in a payload is an owner of a Safe.

**This guard does not validate that a message came from said owner.**

To use this guard, the route should have:
- the 'chainId' declared as a parameter
- the 'safeAddress' declared as a parameter
- the 'account' as part of the JSON body (top level)

Additionally, the `ISafeRepository` now has a function `isOwner` which can be used to verify if an address is an owner of the safe.

### DIFF
--- a/src/routes/email/only-safe-owner.guard.spec.ts
+++ b/src/routes/email/only-safe-owner.guard.spec.ts
@@ -1,0 +1,240 @@
+import { Controller, HttpCode, Post, UseGuards } from '@nestjs/common';
+import { OnlySafeOwner } from '@/routes/email/only-safe-owner.guard';
+import { ISafeRepository } from '@/domain/safe/safe.repository.interface';
+import { Test, TestingModule } from '@nestjs/testing';
+import { TestLoggingModule } from '@/logging/__tests__/test.logging.module';
+import { TestAppProvider } from '@/__tests__/test-app.provider';
+import { ConfigurationModule } from '@/config/configuration.module';
+import configuration from '@/config/entities/__tests__/configuration';
+import * as request from 'supertest';
+import { faker } from '@faker-js/faker';
+
+const safeRepository = {
+  isOwner: jest.fn(),
+} as unknown as ISafeRepository;
+
+const safeRepositoryMock = jest.mocked(safeRepository);
+
+@Controller()
+class TestController {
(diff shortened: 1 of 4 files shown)


src/routes/email/only-safe-owner.guard.spec.ts
```diff
@@ -0,0 +1,240 @@
...
+  async invalidRouteWithSafeAddress() {}
+}
+
+describe('OnlySafeOwner guard tests', () => {
+  let app;
+  // The following address, message and signature represent a valid combination of parameters for validation
+  const testAddress = '0x21509ab252a92b180c539e4d84ea1406f0f87fb8';
+  const testMessage = 'This is a test';
+  const testSignature =
+    '0xcf103c6090c344ffdaa7635d1c41078c5702fd8a8c0528cd980f28c10fb7b3275b7bd803cc99316fc42bf803dcaf982e078d94b094afb3f9853f0dd7ce888aea1c';
```
iamacook, line 42: What do you think about [generating a random signature](https://github.com/safe-global/safe-client-gateway/blob/d39f837dda450914273e68c86344a38c31df2fcb/src/routes/emails/emails.controller.spec.ts#L64-L72) instead?

fmrsabino: Much better! Thanks for the suggestion 👍 [3eb1b70](https://github.com/safe-global/safe-client-gateway/pull/853/commits/3eb1b7087358771ae3cf89fe3e49d0b31ac68d80)

hectorgomezv (approved):

fmrsabino:
@safe-global/safe-client-gateway-maintainers I've split the responsibility of the guard into two different guards.

One (this PR) validates only that an `account` is an owner of the Safe.

The other (https://github.com/safe-global/safe-client-gateway/pull/854) validates the message signature for email registration actions.

The value of splitting these two is that we can combine multiple guards to achieve the desired result. Example: by combining `OnlySafeOwner` and `EmailRegistrationGuard` we only accept email registration requests that are signed correctly **AND** come from an active owner of a safe.

iamacook (approved):

hectorgomezv (approved):
🚀

PR #854, opened 2023-11-16
Add EmailRegistration guard

fmrsabino:
The EmailRegistrationGuard is a guard that can be used to authenticate email registrations.

This guard is successful only if the provided signature is valid. The message that will be validated in the signature needs to follow the format:

`email-register-${chainId}-${safe}-${emailAddress}-${account}-${timestamp}`

The guard can be applied to any route that has:

- the 'chainId' declared as a parameter
- the 'safeAddress' declared as a parameter
- the 'emailAddress' as part of the JSON body (top level)
- the 'account' as part of the JSON body (top level)
- the 'signature' as part of the JSON body (top level) - see message format to be signed
- the 'timestamp' as part of the JSON body (top level)


iamacook (approved):

hectorgomezv (approved):
Awesome! 👏🏻

PR #855, opened 2023-11-16
Adjust ICacheService.get return types

hectorgomezv:


hectorgomezv:
**Side note:**
Another option might be to keep the current signature (i.e. `Promise<string | undefined>`) and actually return `undefined` in the Redis implementation. This would require fewer changes to the codebase, but I opted for `null` at first, as it aligns better with the rest of the code, where we tend to prefer returning `null` over `undefined` on many functions.

fmrsabino (approved):
🚀

PR #856, opened 2023-11-16
Add Timestamp guard

fmrsabino:
The `TimestampGuard` can be used to validate that the provided `timestamp` in an HTTP request body is within a certain time frame from the current system time (in UTC).

This timeframe can be set with `maxElapsedTimeMs`. Example:

```typescript
// Route with timestamp guard set to 5000 ms
@Post()
@UseGuards(TimestampGuard(5_000))
async route() {}
```

If the time that passed since the provided `timeframe` and the current system time is greater than `maxElapsedTimeMs`, a 403 is returned and the route is not triggered.


iamacook (approved):

hectorgomezv (approved):
🚀

PR #859, opened 2023-11-17
Pad generated code to six characters

fmrsabino:
- Pads the generated code to six characters.
- The code generated by `code-generator` will be up to six digits long, however, the possible leading zeroes were not being taken into account when storing the code.
- The `saveEmail` and `setVerificationCode` functions from `IEmailDataSource` now accept `code` as a string instead of as a number. This is closer to the actual type declared for the code in the database (`text`).

### DIFF
--- a/src/domain/email/email.repository.ts
+++ b/src/domain/email/email.repository.ts
@@ -38,10 +38,15 @@ export class EmailRepository implements IEmailRepository {
 
     const verificationCode = codeGenerator();
 
+    // Pads the final verification code to 6 characters
+    // The generated code might have less than 6 digits so the version to be
+    // validated against should account with the leading zeroes
+    const paddedVerificationCode = verificationCode.toString().padStart(6, '0');
+
     try {
       await this.emailDataSource.saveEmail({
         chainId: args.chainId,
-        code: verificationCode,
+        code: paddedVerificationCode,
         emailAddress: email.value,
         safeAddress: args.safeAddress,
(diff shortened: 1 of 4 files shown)


iamacook (approved):

src/domain/email/email.repository.ts
```diff
@@ -38,10 +38,15 @@ export class EmailRepository implements IEmailRepository {
 
     const verificationCode = codeGenerator();
 
+    // Pads the final verification code to 6 characters
+    // The generated code might have less than 6 digits so the version to be
+    // validated against should account with the leading zeroes
+    const paddedVerificationCode = verificationCode.toString().padStart(6, '0');
```
hectorgomezv, line 44: Nitpick: should we move this logic and make `code-generator.ts` to return a padded string? I guess the response depends on the foreseen use cases for the code generator. So I left this up to you, just commenting for your consideration.

hectorgomezv (approved):

src/domain/email/email.repository.ts
fmrsabino: The main reason why I didn't add this logic there was mostly to leave `code-generator.ts` to deal with numbers only.

So the padding part could be seen as an extra step if we want to store it in the DB as such thus having this logic in the repository.

But no other reasons behind having this logic here. We can also see how it evolves. One option is also to merge the `code-generator` with the repository.

PR #860, opened 2023-11-20
Add configuration to set JSON body limit

fmrsabino:
- Adds a new configuration entry which allows setting the JSON body limit of the web server.
- If none is set, it uses the default value of 1Mb.

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -53,7 +53,7 @@ export default () => ({
     // specifies the number of bytes; if it is a string, the value is passed to the
     // bytes library for parsing. Defaults to '100kb'.
     // https://expressjs.com/en/resources/middleware/body-parser.html
-    jsonLimit: process.env.EXPRESS_JSON_LIMIT,
+    jsonLimit: process.env.EXPRESS_JSON_LIMIT ?? '1mb',
   },
   features: {
     pricesProviderChainIds:
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -48,6 +48,13 @@ export default () => ({
       ),
     },
   },
+  express: {
+    // Controls the maximum request body size. If this is a number, then the value
+    // specifies the number of bytes; if it is a string, the value is passed to the
+    // bytes library for parsing. Defaults to '100kb'.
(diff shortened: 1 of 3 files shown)


src/config/entities/configuration.ts
```diff
@@ -48,6 +48,13 @@ export default () => ({
       ),
     },
   },
+  express: {
+    // Controls the maximum request body size. If this is a number, then the value
+    // specifies the number of bytes; if it is a string, the value is passed to the
+    // bytes library for parsing. Defaults to '100kb'.
+    // https://expressjs.com/en/resources/middleware/body-parser.html
```
hectorgomezv, line 55: What's the current value of this limit? I mean, without doing the call `app.use(json({ limit: jsonBodySizeLimit }));`? Is it 100kb as well or is it less?

hectorgomezv: I'm asking because if we have a 100kb limit currently, what do you think about augmenting that value via a default value here? Just considering decreasing the amount of configuration passed via env vars.

hectorgomezv (approved):

src/config/entities/configuration.ts
fmrsabino: > What's the current value of this limit?

The limit is the one described in the comment (default value if none is set). Or do you mean something else?

> I'm asking because if we have a 100kb limit currently, what do you think about augmenting that value via a default value here? Just considering decreasing the amount of configuration passed via env vars.

Good point. I will change the value to meet our requirements. I think the configuration is explicit enough if users running the service want to keep the 100kb.

fmrsabino: Added default in [8d2de6a](https://github.com/safe-global/safe-client-gateway/pull/860/commits/8d2de6a82f6249de856aff63c39959c679cee2ad)

hectorgomezv: > Or do you mean something else?

I was just wondering whether without the explicit call to `app.use(json(...` the value could be different. Now it is clear, thanks!

> Added default in [8d2de6a](https://github.com/safe-global/safe-client-gateway/pull/860/commits/8d2de6a82f6249de856aff63c39959c679cee2ad)

Awesome, thanks!

PR #861, opened 2023-11-20
Fix CoingeckoAPI tests corner case

hectorgomezv:
Changes:

- Adds a minimum value for the CoingeckoAPI response, to replicate the actual behavior of the API (no zero-values), and avoid the tests to fail when a 0 value is used.


fmrsabino (approved):
How is a price with value of 0 represented in CoinGecko?

hectorgomezv:
> How is a price with value of 0 represented in CoinGecko?

I haven't seen any 0 value in a price coming from Coingecko. But if it happens we would treat it as a falsy value, i.e. like a not-found token price, which would be mapped to `null`.

PR #862, opened 2023-11-20
Add email registration route

fmrsabino:
Routes:

- Adds an email registration route, allowing owners of a Safe to register an email address (which will be tied to that owner).
- Moves related email guards to `src/routes/email/guards`
- The `EmailController` is set to be excluded from the Swagger interface – it can be shown whenever the feature is considered to be stable.

The new route is available under:

```javascript
POST /v1/chains/:chainId/safes/:safeAddress/emails

{
  "emailAddress": <string>,
  "account": <string>,
  "timestamp": <number>,
  "signature": <string>
}
```

Domain:

- Removes the owner check from the domain layer as this is already covered by the `OnlySafeOwnerGuard`.

Data:

- Adds a `TestEmailDatasourceModule` – this module should be used whenever the real database instance doesn't need to be considered in the tests. This module uses a mocked instance of `IEmailDataSource`.

### Out of scope

- This feature is currently not part of `AppModule`. Adding this feature to `AppModule` requires a rework of the other tests which will attempt to perform a migration and lead to the number of allowed Postgres connections to be exhausted. This will be tackled in a separate PR.


src/routes/email/email.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+  version: '1',
+})
+@ApiExcludeController()
+export class EmailController {
+  constructor(private readonly service: EmailService) {}
+
+  @Post('')
+  @UseGuards(
+    EmailRegistrationGuard,
+    TimestampGuard(5 * 60 * 1000), // 5 minutes
```
fmrsabino, line 21: The `TimestampGuard` takes a number as an argument but since this is resolved on a build stage, passing the `configuration` value here is not possible (because the configuration is not resolved at this stage).

We can potentially change the `TimestampGuard` to accept a `() => number` instead of `number` so that it is lazily evaluated.

src/routes/email/email.controller.spec.ts
```diff
@@ -0,0 +1,215 @@
...
+    const emailAddress = faker.internet.email();
+    const timestamp = jest.now();
+    const privateKey = generatePrivateKey();
+    const account = privateKeyToAccount(privateKey);
+    const accountAddress = account.address;
+    // Signer is owner of safe
+    const safe = safeBuilder()
+      .with('owners', [accountAddress])
+      // Faker generates non-checksum addresses only
+      .with('address', getAddress(faker.finance.ethereumAddress()))
```
fmrsabino, line 120: We can potentially adjust the builder to returned checksummed addresses by default.

src/domain/email/email.repository.ts
```diff
@@ -1,41 +1,23 @@
...
   constructor(
     @Inject(IEmailDataSource)
     private readonly emailDataSource: IEmailDataSource,
-    @Inject(ISafeRepository) private readonly safeRepository: ISafeRepository,
   ) {}
 
   async saveEmail(args: {
     chainId: string;
-    safeAddress: string;
+    safe: string;
```
hectorgomezv, line 16: Nitpick: why is this renamed to `safe` internally in this repository? I mean, in the public API (controller) `safeAddress` is used, and it is named `safeAddress` in `EmailDataSource` also, so it feels a bit more straightfoward to keep that name within the flow. And also I would tend to think `safe` variable contains a `Safe` entity.

No strong opinion on this, it's a minor detail 🙂

hectorgomezv (approved):
👏🏻 🚀

src/domain/email/email.repository.ts
fmrsabino: I agree 👍 Renamed in [b141e7b](https://github.com/safe-global/safe-client-gateway/pull/862/commits/b141e7b9baad3c5528d2ed74cf5f2e98cdb0bbb4)

No strong reason for the rename. It was to keep the names short but in the domain it does not make sense since we have a `Safe` entity which I expect to be associated with the name `safe` as you mentioned.

fmrsabino: I agree 👍 Renamed in [b141e7b](https://github.com/safe-global/safe-client-gateway/pull/862/commits/b141e7b9baad3c5528d2ed74cf5f2e98cdb0bbb4)

No strong reason for the rename. It was to keep the names short but in the domain it does not make sense since we have a `Safe` entity which I expect to be associated with the name `safe` as you mentioned.

hectorgomezv (approved):

PR #869, opened 2023-11-21
Use TestEmailDatasourceModule in unit tests

fmrsabino:
- Uses `TestEmailDatasourceModule` instead of `EmailDataSourceModule` to avoid connecting to the database if a connection is not required/not part of the test suite.
- This change applies to Unit tests only. E2E tests still use the database integration.


hectorgomezv (approved):
🚀

PR #871, opened 2023-11-21
Add Pushwoosh email addresses deletion

hectorgomezv:
Domain changes:
- Adds `IEmailApi.deleteEmailAddress`, which triggers an email address deletion on the provider side.

Datasource changes:
- Implements `PushwooshApi.deleteEmailAddress`, which implements `IEmailApi.deleteEmailAddress` and triggers the email address deletion on the Pushwoosh email platform.
- Adds related unit tests.


src/datasources/email-api/pushwoosh-api.service.ts
```diff
@@ -63,4 +63,18 @@ export class PushwooshApi implements IEmailApi {
       throw this.httpErrorFactory.from(error);
     }
   }
+
+  async deleteEmailAddress(args: { emailAddress: string }): Promise<void> {
+    try {
+      const url = `${this.baseUri}/json/1.3/deleteEmail`;
+      await this.networkService.post(url, {
+        request: {
+          application: this.applicationCode,
```
fmrsabino, line 72: The API key is not required for this call?

src/domain/interfaces/email-api.interface.ts
```diff
@@ -3,5 +3,21 @@ import { CreateEmailMessageDto } from '@/domain/email/entities/create-email-mess
...
+   * @param createEmailMessageDto.to - the email address of the recipient
+   * @param createEmailMessageDto.template - the email template reference
+   * @param createEmailMessageDto.subject - the email message subject
+   * @param createEmailMessageDto.substitutions - object containing a dictionary of substitutions
+   * to be applied to the email message template
+   */
   createMessage(createEmailMessageDto: CreateEmailMessageDto): Promise<void>;
+
+  /**
+   * Deletes an email address previously persisted into the underlying provider.
```
fmrsabino, line 18: nitpick: I would phrase this more in a way that we trigger the deletion since deletion does not happen on our side.

src/datasources/email-api/pushwoosh-api.service.ts
hectorgomezv: Apparently, it isn't: https://docs.pushwoosh.com/platform-docs/api-reference/email-api#deleteemail

I tried this without authentication and it seems to work (returns a 200 status code). Email addresses are registered automatically to the _PW application_ when creating an email message, but it seems (un)registration endpoints are permissionless. I guess it's intended to be used as a way to (un)subscribe users on their own.

src/domain/interfaces/email-api.interface.ts
hectorgomezv: Changed in [e175e47](https://github.com/safe-global/safe-client-gateway/pull/871/commits/e175e4711a2ae2bb39fc88a25ac2188350fa8fdf) 🙂

fmrsabino (dismissed):
🚀

fmrsabino (approved):

PR #872, opened 2023-11-21
Send unknown recovery transaction email

hectorgomezv:
This PR implements a notification when an unknown recovery transaction is received in an `AlertLog`. 

Domain changes:
- Injects `IEmailApi` into `AlertsRepository`, to create an email message as a user notification.
- Injects `IEmailRepository` into `AlertsRepository`, to retrieve the email addresses for the notification.

Configuration changes:
- Adds a new key `email.templates.unknownRecoveryTx` to hold the unknown recovery transaction email template reference.

Out of scope:
- At least one test should be added to `AlertsController` tests.

### DIFF
--- a/src/config/configuration.validator.spec.ts
+++ b/src/config/configuration.validator.spec.ts
@@ -13,7 +13,7 @@ describe('Configuration validator', () => {
   it('should detect missing mandatory configuration in production environment', () => {
     process.env.NODE_ENV = 'production';
     expect(() => validate(JSON.parse(fakeJson()))).toThrow(
-      /must have required property 'AUTH_TOKEN'.*must have required property 'EXCHANGE_API_KEY'.*must have required property 'ALERTS_PROVIDER_SIGNING_KEY'.*must have required property 'ALERTS_PROVIDER_API_KEY'.*must have required property 'ALERTS_PROVIDER_ACCOUNT'.*must have required property 'ALERTS_PROVIDER_PROJECT'.*must have required property 'EMAIL_API_APPLICATION_CODE'.*must have required property 'EMAIL_API_FROM_EMAIL'.*must have required property 'EMAIL_API_KEY'/,
+      /must have required property 'AUTH_TOKEN'.*must have required property 'EXCHANGE_API_KEY'.*must have required property 'ALERTS_PROVIDER_SIGNING_KEY'.*must have required property 'ALERTS_PROVIDER_API_KEY'.*must have required property 'ALERTS_PROVIDER_ACCOUNT'.*must have required property 'ALERTS_PROVIDER_PROJECT'.*must have required property 'EMAIL_API_APPLICATION_CODE'.*must have required property 'EMAIL_API_FROM_EMAIL'.*must have required property 'EMAIL_API_KEY'.*must have required property 'EMAIL_TEMPLATE_UNKNOWN_RECOVERY_TX'/,
     );
   });
 
@@ -31,6 +31,7 @@ describe('Configuration validator', () => {
         EMAIL_API_APPLICATION_CODE: faker.string.alphanumeric(),
         EMAIL_API_FROM_EMAIL: faker.internet.email(),
         EMAIL_API_KEY: faker.string.uuid(),
+        EMAIL_TEMPLATE_UNKNOWN_RECOVERY_TX: faker.string.alphanumeric(),
       }),
     ).toThrow(/LOG_LEVEL must be equal to one of the allowed values/);
   });
@@ -48,6 +49,7 @@ describe('Configuration validator', () => {
       EMAIL_API_APPLICATION_CODE: faker.string.alphanumeric(),
       EMAIL_API_FROM_EMAIL: faker.internet.email(),
       EMAIL_API_KEY: faker.string.uuid(),
+      EMAIL_TEMPLATE_UNKNOWN_RECOVERY_TX: faker.string.alphanumeric(),
     };
     const validated = validate(expected);
     expect(validated).toBe(expected);
--- a/src/domain/alerts/alerts.repository.interface.ts
+++ b/src/domain/alerts/alerts.repository.interface.ts
@@ -15,5 +15,5 @@ export interface IAlertsRepository {
    * @param chainId - chain where the alert log was generated
    * @param log - the {@link AlertLog} to decode
    */
-  handleAlertLog(chainId: string, log: AlertLog): void;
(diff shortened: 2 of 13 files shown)


src/domain/alerts/alerts.repository.ts
```diff
@@ -79,4 +90,28 @@ export class AlertsRepository implements IAlertsRepository {
...
+    });
+
+    if (emails.length) {
+      return this.emailApi.createMessage({
+        to: emails,
+        template: this.configurationService.getOrThrow<string>(
+          'email.templates.unknownRecoveryTx',
+        ),
+        subject: 'Unknown transaction attempt',
+        substitutions: {},
```
hectorgomezv, line 110: The contents for these fields are out of the scope of this PR, and will be reviewed later when real subjects and the needed substitutions in the final template are clarified.

src/domain/alerts/alerts.repository.ts
```diff
@@ -79,4 +90,28 @@ export class AlertsRepository implements IAlertsRepository {
...
+      });
+    }
+    this.loggingService.debug(
+      `An alert log for an invalid transaction with no verified emails associated was thrown for Safe ${log.address}`,
+    );
```
fmrsabino, line 115: nitpick: to switch the conditions around:

```typescript
if (!emails.length) {
      this.loggingService.debug(
        `An alert log for an invalid transaction with no verified emails associated was thrown for Safe ${log.address}`,
      );
      return;
    }

    return this.emailApi.createMessage({
      to: emails,
      template: this.configurationService.getOrThrow<string>(
        'email.templates.unknownRecoveryTx',
      ),
      subject: 'Unknown transaction attempt',
      substitutions: {},
    });
```

src/domain/alerts/alerts.repository.interface.ts
```diff
@@ -12,7 +12,8 @@ export interface IAlertsRepository {
 
   /**
    * Parses and notifies the user about the {@link log} from the Alerts provider
+   * @param chainId - chain where the alert log was generated
    * @param log - the {@link AlertLog} to decode
    */
-  handleAlertLog(log: AlertLog): void;
+  handleAlertLog(chainId: string, log: AlertLog): void;
```
fmrsabino, line 18: This should return `Promise<void>`.

src/config/configuration.validator.spec.ts
```diff
@@ -13,7 +13,7 @@ describe('Configuration validator', () => {
...
     process.env.NODE_ENV = 'production';
     expect(() => validate(JSON.parse(fakeJson()))).toThrow(
-      /must have required property 'AUTH_TOKEN'.*must have required property 'EXCHANGE_API_KEY'.*must have required property 'ALERTS_PROVIDER_SIGNING_KEY'.*must have required property 'ALERTS_PROVIDER_API_KEY'.*must have required property 'ALERTS_PROVIDER_ACCOUNT'.*must have required property 'ALERTS_PROVIDER_PROJECT'.*must have required property 'EMAIL_API_APPLICATION_CODE'.*must have required property 'EMAIL_API_FROM_EMAIL'.*must have required property 'EMAIL_API_KEY'/,
+      /must have required property 'AUTH_TOKEN'.*must have required property 'EXCHANGE_API_KEY'.*must have required property 'ALERTS_PROVIDER_SIGNING_KEY'.*must have required property 'ALERTS_PROVIDER_API_KEY'.*must have required property 'ALERTS_PROVIDER_ACCOUNT'.*must have required property 'ALERTS_PROVIDER_PROJECT'.*must have required property 'EMAIL_API_APPLICATION_CODE'.*must have required property 'EMAIL_API_FROM_EMAIL'.*must have required property 'EMAIL_API_KEY'.*must have required property 'EMAIL_TEMPLATE_UNKNOWN_RECOVERY_TX'/,
```
fmrsabino, line 16: We should evaluate the value that this test provides as this is bound to increase the more we add required keys.

It'd be better in my opinion to have a correct configuration as the baseline and generate tests where we remove required keys.

fmrsabino (changes requested):

src/domain/alerts/alerts.repository.ts
fmrsabino: I'd add a `TODO` here to keep track that this needs to be done.

src/domain/alerts/alerts.repository.ts
```diff
@@ -7,22 +7,34 @@ import { AlertLog } from '@/routes/alerts/entities/alert.dto.entity';
...
 
 @Injectable()
 export class AlertsRepository implements IAlertsRepository {
   constructor(
     @Inject(IAlertsApi)
     private readonly alertsApi: IAlertsApi,
+    @Inject(IEmailApi)
+    private readonly emailApi: IEmailApi,
+    @Inject(IEmailRepository)
+    private readonly emailRepository: IEmailRepository,
```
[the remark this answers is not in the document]
hectorgomezv: Yes, I get your point, but I think `IEmailRepository` shouldn't need to inject `IAlertsRepository`, and if it does, or we need to solve circular dependencies, I've used NestJS's `forwardRef` in the past and it does its job well for these cases, at least when no construction order needs to be followed during the app bootstrap (https://docs.nestjs.com/fundamentals/circular-dependency#forward-reference) 🙂

src/domain/alerts/alerts.repository.ts
```diff
@@ -33,8 +45,7 @@ export class AlertsRepository implements IAlertsRepository {
     );
 
     if (!decodedTransactions) {
-      // Transaction outside of specified ABI => notify user
-      return;
+      return this.notifyInvalidTransaction(chainId, log);
```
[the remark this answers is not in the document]
hectorgomezv: Makes sense, thanks! Changed in [65990f9](https://github.com/safe-global/safe-client-gateway/pull/872/commits/65990f9a4c21b6e17281eb61679349b56e5c9818)

src/domain/alerts/alerts.repository.interface.ts
hectorgomezv: Changed in [65990f9](https://github.com/safe-global/safe-client-gateway/pull/872/commits/65990f9a4c21b6e17281eb61679349b56e5c9818)

fmrsabino (approved):
🚀

src/config/configuration.validator.spec.ts
hectorgomezv: I've created an issue to track this, thanks!

PR #875, opened 2023-11-22
Set AppModule as a DynamicModule

fmrsabino:
The `AppModule` is now a `DynamicModule` – this allows configuring its configuration based on the `configuration` object.

This change is similar to https://github.com/safe-global/safe-client-gateway/commit/501bc74398faa11e3764ca6dd3979c9643c4052c which was later reverted in https://github.com/safe-global/safe-client-gateway/commit/84d48c0bf0cf012da7261669f9a35e290e21df1a.

The main reason for the revert was because the `configuration` object would not take the `.env` file into consideration since  this is a feature exclusive to the `ConfigurationModule` which is not initialised at this stage.

However, even if the `.env` is not loaded at this stage, we still have access to the `configuration` object which relies more on the environment variables set in the session, allowing entire features to be enabled/disabled with a single configuration entry.


src/routes/contracts/__tests__/get-contract.e2e-spec.ts
```diff
@@ -13,7 +14,7 @@ describe('Get contract e2e test', () => {
 
   beforeAll(async () => {
     const moduleRef = await Test.createTestingModule({
-      imports: [AppModule],
+      imports: [AppModule.register(configuration)],
```
hectorgomezv, line 17: I understand that for non-e2e tests the configuration object should be provided to the `register()` function, as this configuration object points to the _test configuration_. But in e2e tests, why is it needed to provide the `configuration` argument? The default value in `app.module.ts` would point to the same file (`@/config/entities/configuration`) 🤔

fmrsabino: It's not needed. I've removed the reference in those tests in [d074350](https://github.com/safe-global/safe-client-gateway/pull/875/commits/d074350c56820d558c42820b9bd0ca5067a4d9cc).

hectorgomezv (approved):
🚀🚀

PR #876, opened 2023-11-22
Add feature flag to the email feature

fmrsabino:
- Adds a feature flag to the email feature
- The feature can be enabled via the environment variable `FF_EMAIL`


fmrsabino:
Tests might fail on this one since every test is opening a connection to the database thus rejecting some connections for some tests (number of open connections reached the limit).

This is fixed by https://github.com/safe-global/safe-client-gateway/pull/869

hectorgomezv (approved):
🚀

hectorgomezv (approved):

PR #877, opened 2023-11-22
Retrieve verified signer email addresses for a Safe

hectorgomezv:
Datasource changes:
- Adds `getVerifiedSignerEmailsBySafeAddress` to `IEmailDataSource`. This allows the clients of the data source class to retrieve the email addresses associated with a given Safe address, which have been previously verified.

### DIFF
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -6,7 +6,6 @@ import { Email } from '@/datasources/email/entities/email.entity';
 import { EmailAddressDoesNotExistError } from '@/datasources/email/errors/email-address-does-not-exist.error';
 import * as shift from 'postgres-shift';
 import configuration from '@/config/entities/__tests__/configuration';
-import { orderBy } from 'lodash';
 
 const DB_CHAIN_ID_MAX_VALUE = 2147483647;
 
@@ -216,45 +215,39 @@ describe('Email Datasource Tests', () => {
         code: faker.number.int({ max: 999998 }).toString(),
       },
     ];
-    await Promise.all(
-      verifiedSigners.map(async ({ emailAddress, signer, code }) => {
-        await target.saveEmail({
-          chainId,
-          safeAddress,
-          emailAddress,
-          signer,
-          code,
-        });
-        await target.verifyEmail({
-          chainId: chainId,
-          safeAddress,
-          signer,
-        });
-      }),
-    );
-    await Promise.all(
-      nonVerifiedSigners.map(async ({ emailAddress, signer, code }) => {
-        await target.saveEmail({
-          chainId,
-          safeAddress,
-          emailAddress,
-          signer,
-          code,
-        });
-      }),
-    );
+    for (const { emailAddress, signer, code } of verifiedSigners) {
+      await target.saveEmail({
+        chainId,
+        safeAddress,
+        emailAddress,
+        signer,
+        code,
+      });
+      await target.verifyEmail({
+        chainId: chainId,
+        safeAddress,
+        signer,
+      });
+    }
+    for (const { emailAddress, signer, code } of nonVerifiedSigners) {
+      await target.saveEmail({
+        chainId,
+        safeAddress,
+        emailAddress,
+        signer,
+        code,
+      });
+    }
 
     const result = await target.getVerifiedSignerEmailsBySafeAddress({
       chainId,
       safeAddress,
     });
 
-    const expected = orderBy(
+    expect(result).toEqual(
       verifiedSigners.map((verifiedSigner) => ({
         email: verifiedSigner.emailAddress,
       })),
-      'email',
     );
-    expect(orderBy(result, 'email')).toEqual(expected);
   });
 });
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -187,4 +187,68 @@ describe('Email Datasource Tests', () => {
       }),
     ).rejects.toThrow(EmailAddressDoesNotExistError);
   });
+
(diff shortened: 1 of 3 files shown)


src/datasources/email/email.datasource.ts
```diff
@@ -8,6 +8,18 @@ import { Email } from '@/datasources/email/entities/email.entity';
...
+  async getVerifiedSignerEmailsBySafeAddress(args: {
+    chainId: string;
+    safeAddress: string;
+  }): Promise<{ email: string }[]> {
+    const emails = await this.sql<Email[]>`
+      SELECT email_address
+      FROM emails.signer_emails
+      WHERE chain_id = ${args.chainId} AND safe_address = ${args.safeAddress} AND verified is true`;
+
+    return emails.map((email) => ({ email: email.email_address }));
```
fmrsabino, line 20: Wouldn't the `SELECT` statement return just email addresses (so a `string`)?

src/domain/interfaces/email.datasource.interface.ts
```diff
@@ -1,6 +1,17 @@
 export const IEmailDataSource = Symbol('IEmailDataSource');
 
 export interface IEmailDataSource {
+  /**
+   * Gets the array of verified emails associated with a Safe address.
```
fmrsabino, line 5: nitpick: "Gets the verified emails associated with a Safe address"

src/datasources/email/email.datasource.spec.ts
```diff
@@ -187,4 +188,73 @@ describe('Email Datasource Tests', () => {
...
+        code: faker.number.int({ max: 999998 }).toString(),
+      },
+      {
+        emailAddress: faker.internet.email(),
+        signer: faker.finance.ethereumAddress(),
+        code: faker.number.int({ max: 999998 }).toString(),
+      },
+    ];
+    await Promise.all(
+      verifiedSigners.map(async ({ emailAddress, signer, code }) => {
```
fmrsabino, line 220: I would say a `for each` is better suited here because of the side effects. The result of the `map` is not being used.

src/datasources/email/email.datasource.spec.ts
```diff
@@ -187,4 +188,73 @@ describe('Email Datasource Tests', () => {
...
+        await target.saveEmail({
```
fmrsabino, line 221: Here we are waiting for the `saveEmail` before moving to the next iteration. Therefore `Promise.all` does not seem to "help" much.

I would however not use `Promise.all` here since it doesn't seem to provide a lot of benefits. Order of insertion is lost if the scheduler switches between the different `saveEmail` right?

src/datasources/email/email.datasource.spec.ts
```diff
@@ -187,4 +188,73 @@ describe('Email Datasource Tests', () => {
...
+      chainId,
+      safeAddress,
+    });
+
+    const expected = orderBy(
+      verifiedSigners.map((verifiedSigner) => ({
+        email: verifiedSigner.emailAddress,
+      })),
+      'email',
+    );
```
fmrsabino, line 257: This `orderBy` email address is not guaranteed by the database correct? If we want to be deterministic in the return then this should be covered either by the domain layer or the datasource itself.

fmrsabino (changes requested):

src/datasources/email/email.datasource.ts
fmrsabino: Ok nevermind. It returns the column name as a key 👍 

However beware that the type won't match `Email` completely. Instead it will be `{ email_address: string }`

fmrsabino: I think here it's probably better to use

```typescript
await this.sql<{ email_address: string }[]>``
```

or even

```typescript
await this.sql``
```

So that the rigth type expectations are set.

hectorgomezv: Yeah, good point, fixed in [f1ff610](https://github.com/safe-global/safe-client-gateway/pull/877/commits/f1ff61004d4b054ac07ac7b5750cec600b9d5d8a)

src/domain/interfaces/email.datasource.interface.ts
hectorgomezv: Changed in [f1ff610](https://github.com/safe-global/safe-client-gateway/pull/877/commits/f1ff61004d4b054ac07ac7b5750cec600b9d5d8a)

src/datasources/email/email.datasource.spec.ts
hectorgomezv: Changed in [f1ff610](https://github.com/safe-global/safe-client-gateway/pull/877/commits/f1ff61004d4b054ac07ac7b5750cec600b9d5d8a)

src/datasources/email/email.datasource.spec.ts
hectorgomezv: Yes, maybe it is a bit overkill to iterate using a `Promise.await` and a `.map` here. 

I've changed it in [f1ff610](https://github.com/safe-global/safe-client-gateway/pull/877/commits/f1ff61004d4b054ac07ac7b5750cec600b9d5d8a)

src/datasources/email/email.datasource.spec.ts
hectorgomezv: I've added an `ORDER BY` clause to the database query to be deterministic. I think it's consistent to order by the primary key (insertion order) but I'm open to other options.

Added in [f1ff610](https://github.com/safe-global/safe-client-gateway/pull/877/commits/f1ff61004d4b054ac07ac7b5750cec600b9d5d8a)

fmrsabino (approved):
👏 🚀

PR #878, opened 2023-11-23
Extend IEmailRepository

hectorgomezv:
This PR extends the `IEmailRepository` interface and its implementation. The new function added to the interface will be used by https://github.com/safe-global/safe-client-gateway/pull/872 

Domain changes:
- Adds `getVerifiedEmailsBySafeAddress` in order to retrieve the verified email addresses associated with a Safe address.

### DIFF
--- a/src/domain/email/email.repository.interface.ts
+++ b/src/domain/email/email.repository.interface.ts
@@ -1,6 +1,25 @@
 export const IEmailRepository = Symbol('IEmailRepository');
 
 export interface IEmailRepository {
+  /**
+   * Gets the verified emails associated with a Safe address.
+   *
+   * @param args.chainId - the chain id of where the Safe is deployed
+   * @param args.safeAddress - the Safe address to use as filter
+   */
+  getVerifiedEmailsBySafeAddress(args: {
+    chainId: string;
+    safeAddress: string;
+  }): Promise<string[]>;
+
+  /**
+   * Saves an email entry.
+   *
+   * @param args.chainId - the chain id of where the Safe is deployed
+   * @param args.safeAddress - the Safe address to which we should store the email address
+   * @param args.emailAddress - the email address to store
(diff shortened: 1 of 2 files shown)


src/domain/email/email.repository.ts
```diff
@@ -39,4 +39,13 @@ export class EmailRepository implements IEmailRepository {
       throw new EmailSaveError(args.chainId, args.safeAddress, args.account);
     }
   }
+
+  async getVerifiedEmailsBySafeAddress(args: {
+    chainId: string;
+    safeAddress: string;
+  }): Promise<string[]> {
```
hectorgomezv, line 46: `emailDataSource.getVerifiedSignerEmailsBySafeAddress` returns a `Promise<{ email: string }[]>. I tried to simplify this for the consumers of the repository by mapping it to `string[]`, but no strong opinion regarding this.

src/domain/email/email.repository.interface.ts
```diff
@@ -1,6 +1,25 @@
...
+  }): Promise<string[]>;
+
+  /**
+   * Saves an email entry.
+   *
+   * @param args.chainId - the chain id of where the Safe is deployed
+   * @param args.safeAddress - the Safe address to which we should store the email address
+   * @param args.emailAddress - the email address to store
+   * @param args.account - the owner address to which we should link the email address to
+   */
```
hectorgomezv, line 22: Not strictly related, but for consistency, I added documentation for both functions in the class.

src/domain/email/email.repository.ts
fmrsabino: This is subject to change in the future as the `Email` entities stabilise a bit. Probably this function would return an `Email` instance that would contain more data to be in line with other functions of the Repository.

i.e.: we would work more with an `Email` entity than a `string`.

fmrsabino (approved):

PR #879, opened 2023-11-23
Extract Alerts domain module

hectorgomezv:
**Problem**
To implement email sending from `AlertsRepository` (as https://github.com/safe-global/safe-client-gateway/pull/872 needs), injection of `EmailRepository` should be possible.

**Solution**
The solution is inspired by the existing `Email` modules separation between `EmailControllerModule` and `EmailDomainModule`. This PR does the same for the `Alert` feature:
- Extracts the `Alert` related domain classes to  `AlertsDomainModule`.
- Creates a new `AlertsControllerModule` that depends on `AlertsDomainModule`.

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -39,7 +39,7 @@ import { GlobalErrorFilter } from '@/routes/common/filters/global-error.filter';
 import { DataSourceErrorFilter } from '@/routes/common/filters/data-source-error.filter';
 import { ServeStaticModule } from '@nestjs/serve-static';
 import { RootModule } from '@/routes/root/root.module';
-import { AlertsModule } from '@/routes/alerts/alerts.module';
+import { AlertsControllerModule } from '@/routes/alerts/alerts.controller.module';
 import { ConfigFactory } from '@nestjs/config/dist/interfaces/config-factory.interface';
 import { EmailControllerModule } from '@/routes/email/email.controller.module';
 
@@ -49,6 +49,7 @@ export class AppModule implements NestModule {
   // into account. The .env file loading is done by the ConfigurationModule
   // which is not available at this stage.
   static register(configFactory: ConfigFactory = configuration): DynamicModule {
+    const isAlertsFeatureEnabled = configFactory()['features']['alerts'];
     const isEmailFeatureEnabled = configFactory()['features']['email'];
 
     return {
@@ -56,7 +57,9 @@ export class AppModule implements NestModule {
       imports: [
         // features
         AboutModule,
-        AlertsModule,
+        ...(isAlertsFeatureEnabled && isEmailFeatureEnabled
+          ? [AlertsControllerModule]
+          : []),
         BalancesModule,
         CacheHooksModule,
         ChainsModule,
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -39,9 +39,9 @@ import { GlobalErrorFilter } from '@/routes/common/filters/global-error.filter';
 import { DataSourceErrorFilter } from '@/routes/common/filters/data-source-error.filter';
 import { ServeStaticModule } from '@nestjs/serve-static';
 import { RootModule } from '@/routes/root/root.module';
-import { AlertsControllerModule } from '@/routes/alerts/alerts.controller.module';
 import { ConfigFactory } from '@nestjs/config/dist/interfaces/config-factory.interface';
 import { EmailControllerModule } from '@/routes/email/email.controller.module';
+import { AlertsControllerModule } from '@/routes/alerts/alerts.controller.module';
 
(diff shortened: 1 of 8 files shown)


src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -173,7 +173,7 @@ describe('Alerts (Unit)', () => {
       await app.close();
     });
 
-    it('returns 403 (Forbidden) for valid signature/invalid payload', async () => {
+    it('returns 404 (Not found) for valid signature/invalid payload', async () => {
```
hectorgomezv, line 176: I had to change this expectation because the route won't be mounted if the `alerts` feature is not active in the configuration, and therefore the service returns a 404 instead. 

If we agree on this approach (enabling/disabling the `alerts` modules via configuration), we can go further and remove `AlertsRouteGuard` as it won't be used anymore.

src/domain/alerts/alerts.repository.ts
```diff
@@ -7,12 +7,17 @@ import { AlertLog } from '@/routes/alerts/entities/alert.dto.entity';
...
 
 @Injectable()
 export class AlertsRepository implements IAlertsRepository {
   constructor(
     @Inject(IAlertsApi)
     private readonly alertsApi: IAlertsApi,
+    // TODO: IEmailRepository needs to be injected to be used in
+    // https://github.com/safe-global/safe-client-gateway/pull/872
+    @Inject(IEmailRepository)
+    private readonly emailRepository: IEmailRepository,
```
hectorgomezv, line 20: I've left this injection here just for visibility and for demonstration purposes, as the main goal of the PR is to make this injection possible, but in this commit the injected class is not used. It will be used by [the referenced PR](https://github.com/safe-global/safe-client-gateway/pull/872).

src/domain/email/email.domain.module.ts
```diff
@@ -1,8 +1,9 @@
-import { Module } from '@nestjs/common';
+import { Global, Module } from '@nestjs/common';
 import { EmailDataSourceModule } from '@/datasources/email/email.datasource.module';
 import { IEmailRepository } from '@/domain/email/email.repository.interface';
 import { EmailRepository } from '@/domain/email/email.repository';
 
+@Global()
```
fmrsabino, line 6: What is the reason for this to be a global module?

src/app.module.ts
```diff
@@ -49,14 +49,17 @@ export class AppModule implements NestModule {
...
+    const isAlertsFeatureEnabled = configFactory()['features']['alerts'];
     const isEmailFeatureEnabled = configFactory()['features']['email'];
 
     return {
       module: AppModule,
       imports: [
         // features
         AboutModule,
-        AlertsModule,
+        ...(isAlertsFeatureEnabled && isEmailFeatureEnabled
```
fmrsabino, line 60: This PR is doing more things than described I would say.

- Extracting a Alerts Domain Module shouldn't change how the feature is loaded on the application.
- The way this is bundled here raises the question of why we need two feature flags instead of one.

src/domain/alerts/alerts.domain.module.ts
```diff
@@ -0,0 +1,20 @@
+import { Global, Module } from '@nestjs/common';
+import { AlertsApiModule } from '@/datasources/alerts-api/alerts-api.module';
+import { AlertsRepository } from '@/domain/alerts/alerts.repository';
+import { IAlertsRepository } from '@/domain/alerts/alerts.repository.interface';
+import { DelayModifierDecoder } from '@/domain/alerts/contracts/delay-modifier-decoder.helper';
+import { SafeDecoder } from '@/domain/alerts/contracts/safe-decoder.helper';
+import { MultiSendDecoder } from '@/domain/alerts/contracts/multi-send-decoder.helper';
+
+@Global()
```
fmrsabino, line 9: Why is this module a global module?

hectorgomezv: No need, I've changed it in [f51fd4b](https://github.com/safe-global/safe-client-gateway/pull/879/commits/f51fd4b3fcf4eb16814563c84d2e497c239a76d0). Thanks!

src/domain/alerts/alerts.repository.ts
hectorgomezv: This change was removed. It will be done in a separate PR.

src/app.module.ts
hectorgomezv: I agree, I've changed both the PR description and its code. Now it only extracts the Alerts controller/domain modules. I'll be doing the other changes in separate PRs. Thanks.

fmrsabino (approved):
🚀

PR #882, opened 2023-11-27
Add functionality to retrieve an email

fmrsabino:
- Retrieving a single email from the datasource is now possible via `IEmailDataSource.getEmail`
- If the email does not exist, an `EmailAddressDoesNotExistError` is thrown.
- The previous `Email` entity was renamed to `EmailAddress` as it represented only the address component of an email.
- A new `Email` entity was added to the domain layer which represents an email registration.
- To increase encapsulation, the `Email` entity in the datasource was moved to the respective datasource file as it is only used within that file and should not be used by other components.


hectorgomezv (approved):
Awesome 🚀

PR #883, opened 2023-11-27
Add a configurable webhook execution delay

hectorgomezv:
Changes:
- Adds a configurable (via the environment variable `WEB_HOOK_EXECUTION_DELAY_MS`) delay before executing web hook. The default value for this is 5 seconds.
- The status code returned by the controller was changed to 202 as it is more accurate given the new behavior.

### DIFF
--- a/src/routes/cache-hooks/cache-hooks.controller.spec.ts
+++ b/src/routes/cache-hooks/cache-hooks.controller.spec.ts
@@ -25,9 +25,11 @@ describe('Post Hook Events (Unit)', () => {
   let safeConfigUrl;
   let fakeCacheService: FakeCacheService;
   let networkService;
+  let webHookExecutionDelayMs;
 
   beforeEach(async () => {
     jest.clearAllMocks();
+    jest.useFakeTimers();
 
     const moduleFixture: TestingModule = await Test.createTestingModule({
       imports: [AppModule.register(configuration)],
@@ -47,9 +49,16 @@ describe('Post Hook Events (Unit)', () => {
     const configurationService = moduleFixture.get(IConfigurationService);
     authToken = configurationService.get('auth.token');
     safeConfigUrl = configurationService.get('safeConfig.baseUri');
+    webHookExecutionDelayMs = configurationService.get(
+      'webHookExecutionDelayMs',
+    );
     networkService = moduleFixture.get(NetworkService);
 
     await app.init();
+  });
+
+  afterEach(() => {
+    jest.useRealTimers();
   });
 
   afterAll(async () => {
@@ -148,7 +157,7 @@ describe('Post Hook Events (Unit)', () => {
       .post(`/hooks/events`)
       .set('Authorization', `Basic ${authToken}`)
       .send(data)
-      .expect(200);
+      .expect(202);
   });
 
   it('returns 400 (Bad Request) on unknown payload', async () => {
@@ -223,9 +232,12 @@ describe('Post Hook Events (Unit)', () => {
       .post(`/hooks/events`)
       .set('Authorization', `Basic ${authToken}`)
       .send(data)
-      .expect(200);
+      .expect(202);
 
-    await expect(fakeCacheService.get(cacheDir)).resolves.toBeNull();
+    setTimeout(
+      () => expect(fakeCacheService.get(cacheDir)).resolves.toBeNull(),
+      webHookExecutionDelayMs,
+    );
   });
 
   it.each([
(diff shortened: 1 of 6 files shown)


src/routes/cache-hooks/cache-hooks.controller.ts
```diff
@@ -43,6 +43,6 @@ export class CacheHooksController {
       | PendingTransaction
       | SafeAppsUpdate,
   ): Promise<void> {
-    await this.service.onEvent(eventPayload);
+    this.service.onEvent(eventPayload);
```
hectorgomezv, line 46: This is changed to run the `onEvent` callback in a non-blocking fashion and avoid having the caller waiting for the event hook execution.

fmrsabino: I think it'd be simpler if the delay was tackled on a web-hook route only initially. So you'd have something like

```typescript
setTimeout(() => {
  this.service.onEvent(eventPayload);
}, 5000); // read timeout value from the configuration
```

I would also add a small test to this where the `service` is mocked and we use virtual timers. With the mock you can check that the service was only called after T seconds/milliseconds. Due to the scope of this issue it's also fine if the test is added as a separate PR 👍

hectorgomezv: I've changed it in [563547d](https://github.com/safe-global/safe-client-gateway/pull/883/commits/563547d36d6c47eb6cdbbab2242173d5723f4fbf)

I had to adjust the controller tests in this PR, they would fail otherwise, as `fakeCacheService.get` wouldn't return `null` immediately as the controller tests were expecting.

hectorgomezv: I also changed the status code returned by the controller to `202` as it is more accurate given the new behavior.

hectorgomezv:
The status code returned by the controller was changed to `202` as it is more accurate given the new behavior. 

The PR title and description were also changed.

src/routes/cache-hooks/cache-hooks.controller.ts
```diff
@@ -15,18 +22,29 @@ import { PendingTransaction } from '@/routes/cache-hooks/entities/pending-transa
...
+  ) {
+    this.webHookExecutionDelayMs = this.configurationService.getOrThrow<number>(
+      'webHookExecutionDelayMs',
+    );
+  }
 
   @UseGuards(BasicAuthGuard)
   @Post('/hooks/events')
-  @HttpCode(200)
+  @HttpCode(202)
```
fmrsabino, line 47: What's the reason for this change? Seems to be out of context or? 🤔

src/routes/cache-hooks/cache-hooks.controller.spec.ts
```diff
@@ -407,9 +428,12 @@ describe('Post Hook Events (Unit)', () => {
       .post(`/hooks/events`)
       .set('Authorization', `Basic ${authToken}`)
       .send(data)
-      .expect(200);
+      .expect(202);
 
-    await expect(fakeCacheService.get(cacheDir)).resolves.toBeNull();
+    setTimeout(
+      () => expect(fakeCacheService.get(cacheDir)).resolves.toBeNull(),
+      webHookExecutionDelayMs,
```
fmrsabino, line 435: This would add the assertion to the event loop which I don't think it's guaranteed to work after `webHookExecutionDelayMs`.

Did you try to use with `jest.advanceTimersByTime`?

hectorgomezv: > This would add the assertion to the event loop which I don't think it's guaranteed to work after `webHookExecutionDelayMs`.

Could you elaborate on why this is not guaranteed to work? 

> Did you try to use with `jest.advanceTimersByTime`?

I saw it in the documentation, but it only works in the context of jest mocks, right? `fakeCacheService` is not a jest mock in this class. I'm not following on how to use that here, could you elaborate on that?

fmrsabino: > Could you elaborate on why this is not guaranteed to work?

The assertion will be added to the event loop but that doesn't mean that it will be executed exactly at `webHookExecutionDelayMs`. It will be executed at the next event cycle. I believe there can be a use case where the deletion can happen after the assertion. 
 
> I saw it in the documentation, but it only works in the context of jest mocks, right? `fakeCacheService` is not a jest mock in this class. I'm not following on how to use that here, could you elaborate on that?

Not really. Jest replaces the timers with fake ones. It is not related with mocks. See https://jestjs.io/docs/timer-mocks

fmrsabino (approved):
We should revisit the way the timers work in these tests or have one where the the value retrieved from cache is still present and not null.

🚀

src/routes/cache-hooks/cache-hooks.controller.spec.ts
hectorgomezv: Synced internally, thanks for your comments @fmrsabino ! 🙂

PR #891, opened 2023-11-28
Clear safe info on MODULE_TRANSACTION events

fmrsabino:
An event of type `MODULE_TRANSACTION` can change the configuration of the safe. Therefore, the Safe configuration stored locally should be cleared.


hectorgomezv (approved):
🚀

PR #894, opened 2023-11-30
refactor: add `AlertsDecodersModule`

hectorgomezv:
This adds `AlertsDecodersModule` which encapsulates transaction decoders that are used in the context of alerts.


hectorgomezv:
> > Adds TestAlertsDecodersModule to have test doubles to be used in AlertsController tests.
> 
> Can you explain what you mean by "test doubles"? What are the reasons for not using the actual decoder implementations in the tests?

I intended to simplify things by mocking the decoders, so we can use them as in this PR: https://github.com/safe-global/safe-client-gateway/pull/895/files#diff-c7e0e49dea155fd49c0d76192b374f6b3731b09496e8e0a2dc32e7dad1b8b9b6R153-R161

If we agree we want to use actual alert payloads, then I think we need to work on the `alertBuilder` class, so it generates valid alerts (I'm getting `AbiEventSignatureNotFoundError` on `DelayModifierDecoder.decodeEventLog()` when I try to parse an alert payload generated with `alertBuilder`). I'm OK with doing it the other way around and start testing with the actual decoder implementations 🙂 We can close this PR if so.

iamacook:
As per discussion, I've removed the decoder test module in favour of decoding "real" transaction data in forthcoming tests (encoders for which are in https://github.com/safe-global/safe-client-gateway/pull/916).

PR #895, opened 2023-11-30
Add test for unknown recovery transaction attempt

hectorgomezv:
Changes:

- Adds `TestEmailApiModule`, which provides test doubles for testing `EmailApiModule` dependent modules.
- Adds a unit test to `AlertsController` tests to verify a notification is sent to the verified signer emails when an unknown recovery transaction is attempted.


iamacook:
I am going to merge this directly as it is purely test-focused and a prerequisite to test https://github.com/safe-global/safe-client-gateway/pull/922.

PR #897, opened 2023-11-30
Add functionality to resend email verification

fmrsabino:
- The `IEmailRepository` was extended to have a function for resending email verifications.
- The email verification can be resent if enough time has passed since this action was last triggered successfully. This timeframe is configurable with `EMAIL_VERIFICATION_CODE_RESEND_LOCK_WINDOW_MS` (default is 30 seconds).
- When resending the verification code, if the current code is still valid, then it is re-sent. If it expired, then a new one is generated, stored and re-sent to the same email address. The expiration time of the code can be set via `EMAIL_VERIFICATION_CODE_TTL_MS`.
- If this action is triggered for verified owners, an `EmailAlreadyVerifiedError` is thrown.
- If this action is triggered within `EMAIL_VERIFICATION_CODE_RESEND_LOCK_WINDOW_MS` from the last successful one, a `ResendVerificationTimespanError` is thrown.
- Else, it attempts to set a new code and send it. If the new code was not set, an `InvalidVerificationCodeError` is thrown.

### Out of scope

- Unit testing will be done from the route perspective and will cover the domain layer introduced here

### DIFF
--- a/src/domain/email/email.repository.ts
+++ b/src/domain/email/email.repository.ts
@@ -4,13 +4,38 @@ import codeGenerator from '@/domain/email/code-generator';
 import { EmailAddress } from '@/domain/email/entities/email.entity';
 import { IEmailRepository } from '@/domain/email/email.repository.interface';
 import { EmailSaveError } from '@/domain/email/errors/email-save.error';
+import { ResendVerificationTimespanError } from '@/domain/email/errors/verification-timeframe.error';
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { EmailAlreadyVerifiedError } from '@/domain/email/errors/email-already-verified.error';
+import { InvalidVerificationCodeError } from '@/domain/email/errors/invalid-verification-code.error';
 
 @Injectable()
 export class EmailRepository implements IEmailRepository {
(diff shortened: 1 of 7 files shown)


src/domain/email/email.repository.ts
```diff
@@ -19,24 +44,22 @@ export class EmailRepository implements IEmailRepository {
...
         emailAddress: email,
         safeAddress: args.safeAddress,
         signer: args.account,
         codeGenerationDate: new Date(),
       });
-
-      // TODO if successful, send the generated code (result.verification)
+      await this._sendEmailVerification({
+        ...args,
+        signer: args.account,
```
iamacook, line 60: Out of the scope of this PR but to take note: we should rename either `signer` to `account` or vice versa for consistency.

src/domain/email/errors/email-already-verified.error.ts
```diff
@@ -0,0 +1,7 @@
+export class EmailAlreadyVerifiedError extends Error {
+  constructor(args: { chainId: string; safeAddress: string; signer: string }) {
```
iamacook, line 2: Nit: as per my previous comment regarding `signer` naming.

PR #910, opened 2023-12-05
Add trusted query param to Transaction History

fmrsabino:
- The transaction history endpoint now has a `trusted` query parameter which can be used to filter transfer (incoming/outgoing) depending on the token's trusted status (this status is something that is set on the Safe Transaction Service)
- Adds a feature flag to enable the newly added filtering. The feature can be enabled by setting `FF_TRUSTED_TOKENS` to `true`
- If the filtering is enabled on the endpoint it is then applied to Incoming and Outgoing ERC20 and ERC721 transfers:
  * If `trusted` is present and `true` then the transfer is included.
  * Else (if it's either `false` or we couldn't get the token info) the transfer is excluded.

### DIFF
--- a/src/routes/transactions/entities/transfer-transaction-info.entity.ts
+++ b/src/routes/transactions/entities/transfer-transaction-info.entity.ts
@@ -35,3 +35,9 @@ export class TransferTransactionInfo extends TransactionInfo {
     this.transferInfo = transferInfo;
   }
 }
+
+export function isTransferTransactionInfo(
+  txInfo: TransactionInfo,
+): txInfo is TransferTransactionInfo {
+  return txInfo.type === 'Transfer';
+}
--- a/src/routes/transactions/entities/transfers/erc20-transfer.entity.ts
+++ b/src/routes/transactions/entities/transfers/erc20-transfer.entity.ts
@@ -14,6 +14,8 @@ export class Erc20Transfer extends Transfer {
   logoUri: string | null;
   @ApiPropertyOptional({ type: Number, nullable: true })
   decimals: number | null;
+  @ApiPropertyOptional({ type: Boolean, nullable: true })
+  trusted: boolean | null;
 
   constructor(
     tokenAddress: string,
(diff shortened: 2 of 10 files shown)


src/routes/transactions/entities/transfer-transaction-info.entity.ts
```diff
@@ -35,3 +35,9 @@ export class TransferTransactionInfo extends TransactionInfo {
     this.transferInfo = transferInfo;
   }
 }
+
+export function isTransferTransactionInfo(
+  txInfo: TransactionInfo,
+): txInfo is TransferTransactionInfo {
+  return txInfo.type === 'Transfer';
```
iamacook, line 42: Out of scope here, but maybe we should consider creating an enum for the `type` of `TransactionInfo`, should there be any changes to the `type` value. What do you think?

src/routes/transactions/entities/transfers/erc20-transfer.entity.ts
```diff
@@ -30,5 +33,10 @@ export class Erc20Transfer extends Transfer {
     this.tokenSymbol = tokenSymbol;
     this.logoUri = logoUri;
     this.decimals = decimals;
+    this.trusted = trusted;
   }
 }
+
+export function isErc20Transfer(transfer: Transfer): transfer is Erc20Transfer {
+  return transfer.type === 'ERC20';
```
iamacook, line 41: My previous comment regarding a `type` enum is valid here too.

src/routes/transactions/entities/transfers/erc721-transfer.entity.ts
```diff
@@ -12,19 +12,29 @@ export class Erc721Transfer extends Transfer {
...
     this.tokenSymbol = tokenSymbol;
     this.logoUri = logoUri;
+    this.trusted = trusted;
   }
 }
+
+export function isErc721Transfer(
+  transfer: Transfer,
+): transfer is Erc721Transfer {
+  return transfer.type === 'ERC721';
```
iamacook, line 39: My previous comment regarding a `type` enum is valid here too.

iamacook (approved):

PR #919, opened 2023-12-06
Add resend email verification route

fmrsabino:
- Adds a new route to resend email verifications – `PUT chains/:chainId/safes/:safeAddress/emails/verify-resend`.
- This route returns:
  * `202 Accepted` – if the service accepted the request to resend the verification email.
  * `409 Conflict` – if the account already has a verified email address.
  * `429 Too Many Requests` – if attempting to resend the verification email within the "lock window".
- Renames `email.controller.spec.ts` to `email.controller.save-email.spec.ts` to group the tests targeting different routes in different files.

The expected payload for this route is the `account` (owner) to which we should resend the email verification to:

```javascript
{
  "account": <string>
}
```

### DIFF
--- a/src/routes/email/email.controller.resend-verification.spec.ts
+++ b/src/routes/email/email.controller.resend-verification.spec.ts
@@ -1,0 +1,242 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { AppModule } from '@/app.module';
+import { CacheModule } from '@/datasources/cache/cache.module';
+import { TestCacheModule } from '@/datasources/cache/__tests__/test.cache.module';
+import configuration from '@/config/entities/__tests__/configuration';
+import { RequestScopedLoggingModule } from '@/logging/logging.module';
+import { TestLoggingModule } from '@/logging/__tests__/test.logging.module';
+import { NetworkModule } from '@/datasources/network/network.module';
+import { TestNetworkModule } from '@/datasources/network/__tests__/test.network.module';
+import { TestAppProvider } from '@/__tests__/test-app.provider';
+import { EmailDataSourceModule } from '@/datasources/email/email.datasource.module';
+import { TestEmailDatasourceModule } from '@/datasources/email/__tests__/test.email.datasource.module';
+import * as request from 'supertest';
+import { faker } from '@faker-js/faker';
+import { IEmailDataSource } from '@/domain/interfaces/email.datasource.interface';
+import { chainBuilder } from '@/domain/chains/entities/__tests__/chain.builder';
+import { EmailControllerModule } from '@/routes/email/email.controller.module';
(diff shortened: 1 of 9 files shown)


src/routes/email/email.controller.resend-verification.spec.ts
```diff
@@ -0,0 +1,242 @@
...
+    // Advance timer to a time within resendLockWindowMs
+    jest.advanceTimersByTime(resendLockWindowMs - 1);
+    await request(app.getHttpServer())
+      .put(
+        `/v1/chains/${chain.chainId}/safes/${safe.address}/emails/verify-resend`,
+      )
+      .send({
+        account: account,
+      })
+      .expect(429);
```
iamacook, line 130: We're not checking the error body here as we do in other the other tests.

src/routes/email/exception-filters/email-already-verified.exception-filter.ts
```diff
@@ -0,0 +1,21 @@
...
+  Catch,
+  ExceptionFilter,
+  HttpStatus,
+} from '@nestjs/common';
+import { EmailAlreadyVerifiedError } from '@/domain/email/errors/email-already-verified.error';
+import { Response } from 'express';
+
+@Catch(EmailAlreadyVerifiedError)
+export class EmailAlreadyVerifiedExceptionFilter implements ExceptionFilter {
+  catch(exception: EmailAlreadyVerifiedError, host: ArgumentsHost): any {
```
iamacook, line 12: Nit: can we change `any` to `void`? Type checking is otherwise disabled.

src/routes/email/exception-filters/resend-verification-timespan-error.exception-filter.ts
```diff
@@ -0,0 +1,23 @@
...
+  HttpStatus,
+} from '@nestjs/common';
+import { Response } from 'express';
+import { ResendVerificationTimespanError } from '@/domain/email/errors/verification-timeframe.error';
+
+@Catch(ResendVerificationTimespanError)
+export class ResendVerificationTimespanExceptionFilter
+  implements ExceptionFilter
+{
+  catch(exception: ResendVerificationTimespanError, host: ArgumentsHost): any {
```
iamacook, line 14: Nit: as above re. changing `any` to `void`

src/routes/email/exception-filters/email-already-verified.exception-filter.ts
fmrsabino: The return type comes from the interface `ExceptionFilter` (NestJS).

I've removed the return type hint here though. [f0f5b82](https://github.com/safe-global/safe-client-gateway/pull/919/commits/f0f5b8208024628b4bab921e6e5d36f40e9a95fb)

iamacook (approved):

PR #920, opened 2023-12-07
Add email verification route

fmrsabino:
Depends on https://github.com/safe-global/safe-client-gateway/pull/919

- Adds a new route to verify email addresses – `PUT /v1/chains/:chainId/safes/:safeAddress/emails/verify`.
- This route sets an email address as verified if successful.
- This route returns:
  * `204 No Content` – if the request was successful and that the client does not need to navigate away from the current page. This status code is also returned if the email was already verified.
  * `400 Bad Request` – if the code is not valid (wrong or expired code).

 This route accepts the following payload:

 ```javascript
 {
   "account": <string>,
   "code": <string>
 }
 ```

### DIFF
--- a/src/domain/email/email.repository.interface.ts
+++ b/src/domain/email/email.repository.interface.ts
@@ -44,4 +44,22 @@ export interface IEmailRepository {
     safeAddress: string;
     signer: string;
   }): Promise<void>;
+
+  /**
+   * Verifies an email address with the provided code.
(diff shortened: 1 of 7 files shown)


iamacook (approved):

fmrsabino:
@iamacook had to rebase due to conflicts with the `main` branch.

PR #921, opened 2023-12-08
refactor: create `IEncoder` for transaction/event mocking

iamacook:
This refactors the encoders added in #916 to extend upon `IBuilder`. A new `IEncoder` has been created, with a base `Encoder` to extend from.

Note: some small modifications were made to `IBuilder` and `Builder` to facilitate this.


src/__tests__/builder.ts
```diff
@@ -1,21 +1,21 @@
 export interface IBuilder<T> {
-  with<K extends keyof T>(key: K, value: T[K]): IBuilder<T>;
+  with<K extends keyof T>(key: K, value: T[K]): this;
```
fmrsabino, line 2: Due to the scope of the `IBuilder` and how it's used in many tests I would create a PR just with these changes.

src/__tests__/builder.ts
```diff
@@ -1,21 +1,21 @@
...
 
   build(): T;
 }
 
 export class Builder<T> implements IBuilder<T> {
-  private constructor(private target: Partial<T>) {}
+  constructor(private target: Partial<T>) {}
```
fmrsabino, line 8: We can start with `protected` as I understand this is only going to be used by subclasses.

fmrsabino (changes requested):

src/__tests__/encoder.ts
```diff
@@ -0,0 +1,17 @@
+import { Hex } from 'viem';
+
+import { Builder, IBuilder } from '@/__tests__/builder';
+
+export interface IEncoder<T, E = Hex> extends IBuilder<T> {
```
fmrsabino, line 5: Since this is an extension of the `IBuilder` class and it's going to be mostly used for building payloads for testing, I would rename this to `IEncoderBuilder`

src/__tests__/encoder.ts
```diff
@@ -0,0 +1,17 @@
...
+  encode(): E;
+}
+
+export class Encoder<T, E = Hex> extends Builder<T> implements IEncoder<T, E> {
+  encode(): E {
+    throw new Error('Method not implemented.');
+  }
```
fmrsabino, line 12: `encode` does seem to hint that this might be better suited for an abstract method.

Since the base `Builder` class already has the required properties minus the encoding functionality probably the `Encoder` class can be removed altogether and current implementations of this subclass would implement the `IEncoder` interface and extend the `Builder` class. What do you think?

fmrsabino (changes requested):

hectorgomezv:
Great job! 👏🏻 I agree with @fmrsabino comments. Otherwise, it looks good to me.

src/__tests__/builder.ts
iamacook: As per discussion, I've created a [separate PR](https://github.com/safe-global/safe-client-gateway/pull/947) to tackle this.

src/__tests__/builder.ts
iamacook: Adjusted in f7bfad5d54cfc2a28765809536b4989f838b2626.

src/__tests__/encoder.ts
iamacook: Renamed in f7bfad5d54cfc2a28765809536b4989f838b2626.

src/__tests__/encoder.ts
iamacook: As per discussion, we will iterate on this afterwards.

fmrsabino (approved):
👏

PR #922, opened 2023-12-08
feat: add notifications for recovery proposals

iamacook:
This adds base logic for sending emails of valid recovery proposals to verified email addresses with the new, predicted Safe Account setup structure.


src/domain/alerts/alerts.repository.ts
```diff
@@ -51,33 +52,69 @@ export class AlertsRepository implements IAlertsRepository {
...
+  private decodeTransactionAdded(data: Hex) {
+    try {
+      const decoded = this.safeDecoder.decodeFunctionData({ data });
+
+      if (decoded.functionName !== 'execTransaction') {
+        return [decoded];
+      }
+
+      return this.multiSendDecoder
+        .mapMultiSendTransactions(decoded.args[2])
```
fmrsabino, line 79: nitpick: I would extract `decoded.args[2]` so that it's more clear on what it represents.

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -109,78 +125,525 @@ describe('Alerts (Unit)', () => {
...
+              return Promise.reject(`No matching rule for url: ${url}`);
+          }
+        });
+
+        await request(app.getHttpServer())
+          .post('/v1/alerts')
+          .set('x-tenderly-signature', signature)
+          .set('date', timestamp)
+          .send(alert)
+          .expect(200)
```
fmrsabino, line 195: Out of scope for this PR but [HTTP 202](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/202) better represents the outcome of the alerts route.

src/domain/alerts/alerts.repository.ts
```diff
@@ -51,33 +52,69 @@ export class AlertsRepository implements IAlertsRepository {
...
+        decodedTransactions,
+      });
+
+      return this._notifyNewSafeSetup({ chainId, safeAddress, newSafe });
+    } catch {
+      return this._notifyUnknownTransaction(chainId, log);
+    }
+  }
+
+  private decodeTransactionAdded(data: Hex) {
```
hectorgomezv, line 70: Nitpick: could we prefix private functions with `_`? I know it's not a standard through all the codebase, but at least to stay consistent within this file.

src/domain/alerts/alerts.repository.ts
```diff
@@ -51,33 +52,69 @@ export class AlertsRepository implements IAlertsRepository {
...
-      chainId,
-      address: decodedEvent.args.to,
+    try {
+      const safeAddress = decodedEvent.args.to;
+
+      const newSafe = await this.predictNewSafeSetup({
+        chainId,
+        safeAddress,
+        decodedTransactions,
+      });
```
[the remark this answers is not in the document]
hectorgomezv: Nitpick: just adding more name options: `updatedSafeSetup`, `updatedSafeState`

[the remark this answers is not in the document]
iamacook: I opted for `newSafeState` in ab0c720fca0ecfeba81054f22fc9ad347c96765d.

src/domain/alerts/alerts.repository.ts
```diff
@@ -51,33 +52,69 @@ export class AlertsRepository implements IAlertsRepository {
...
+
+      return this._notifyNewSafeSetup({ chainId, safeAddress, newSafe });
+    } catch {
+      return this._notifyUnknownTransaction(chainId, log);
+    }
```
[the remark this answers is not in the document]
iamacook: Good point. `_notifyNewSafeSetup` (to be renamed) would throw if there is a non-owner management transaction but the decoding (currently) notifies of that further up.

What do you think about removing the `catch` from within `decodeTransactionAdded` and calling it within the `try` instead? I think that would be a bit clearer. cc @hectorgomezv

src/domain/alerts/alerts.repository.ts
iamacook: Adjusted in ab0c720fca0ecfeba81054f22fc9ad347c96765d.

src/domain/alerts/alerts.repository.ts
iamacook: Extracted in ab0c720fca0ecfeba81054f22fc9ad347c96765d.

src/domain/alerts/alerts.repository.ts
```diff
@@ -51,33 +52,69 @@ export class AlertsRepository implements IAlertsRepository {
...
+
+      return this.multiSendDecoder
+        .mapMultiSendTransactions(decoded.args[2])
+        .flatMap(({ data }) => this.safeDecoder.decodeFunctionData({ data }));
+    } catch {
+      return null;
+    }
+  }
+
+  private async predictNewSafeSetup(args: {
```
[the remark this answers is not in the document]
iamacook: Adjusted in ab0c720fca0ecfeba81054f22fc9ad347c96765d.

src/domain/alerts/alerts.repository.ts
```diff
@@ -142,4 +163,34 @@ export class AlertsRepository implements IAlertsRepository {
       });
     }
   }
+
+  private async _notifyNewSafeSetup(args: {
```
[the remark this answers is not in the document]
iamacook: Adjusted in ab0c720fca0ecfeba81054f22fc9ad347c96765d.

src/domain/alerts/alerts.repository.ts
```diff
@@ -142,4 +163,34 @@ export class AlertsRepository implements IAlertsRepository {
...
+    chainId: string;
+    safeAddress: string;
+    newSafe: Safe;
```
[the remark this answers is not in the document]
iamacook: Indeed. `safe.address` is instead referenced from ab0c720fca0ecfeba81054f22fc9ad347c96765d.

src/routes/alerts/alerts.controller.spec.ts
iamacook: I'll address this in a follow up PR.

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -109,78 +125,525 @@ describe('Alerts (Unit)', () => {
...
+          ),
+          to: expectedTargetEmailAddresses,
+        });
       });
+
+      it.todo('notifies about alerts with multiple logs');
+    });
+
+    describe('it notifies about an invalid transaction attempt', () => {
+      it('notifies about an invalid transaction attempt', async () => {
```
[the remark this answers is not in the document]
iamacook: I've added a comment in ab0c720fca0ecfeba81054f22fc9ad347c96765d.

src/routes/alerts/alerts.controller.spec.ts
iamacook: I've created [a separate PR](https://github.com/safe-global/safe-client-gateway/pull/948) that tackles this. I will update the tests accordingly when merging.

src/domain/alerts/alerts.repository.ts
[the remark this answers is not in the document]
hectorgomezv: Yes, I think it makes more sense and it will simplify the code 👍🏻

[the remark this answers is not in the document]
iamacook: I've included the suggested refactor in 110716087c1e6c60867165586081cdbf23c478dc.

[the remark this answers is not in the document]
fmrsabino: > Good point. `_notifyNewSafeSetup` (to be renamed) would throw if there is a non-owner management transaction but the decoding (currently) notifies of that further up.

You mean `_mapSafeSetup` right? 

> What do you think about removing the `catch` from within `decodeTransactionAdded` and calling it within the `try` instead? I think that would be a bit clearer. cc @hectorgomezv

There's multiple layers of error handling – the service itself (`AlertsService`) also has a try catch which logs a warning but then returns to the caller 200.

I think in the end it's important to notify the user even on payloads that we do not expect so we should probably not throw as I commented before.

src/domain/alerts/alerts.repository.ts
```diff
@@ -142,4 +165,33 @@ export class AlertsRepository implements IAlertsRepository {
...
+  private async _notifySafeSetup(args: {
+    chainId: string;
+    newSafeState: Safe;
+  }): Promise<void> {
+    const emails = await this.emailRepository.getVerifiedEmailsBySafeAddress({
+      chainId: args.chainId,
+      safeAddress: args.newSafeState.address,
+    });
+
+    if (!emails.length) {
```
fmrsabino, line 168: I think we should have this condition as soon as we know the safe address. If there are no verified email addresses we shouldn't need to do all the mapping.

src/routes/alerts/alerts.controller.spec.ts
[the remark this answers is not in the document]
hectorgomezv: Awesome! Thanks.

src/domain/alerts/alerts.repository.ts
hectorgomezv: I agree, good catch. It should be a common precondition before emitting any type of alert.

src/domain/alerts/alerts.repository.ts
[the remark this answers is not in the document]
iamacook: > You mean `_mapSafeSetup` right?

Yes. I've reverted my change for now as it has an influence on my [proposed retrieval of the Safe address for unknown transactions](https://github.com/safe-global/safe-client-gateway/pull/923) as well. Refactoring this would make sense there once we are in agreement.

src/domain/alerts/alerts.repository.ts
iamacook: I would suggest we keep this optimisation in mind and include it once we agree upon the [retrieval of the Safe address for unknown transactions](https://github.com/safe-global/safe-client-gateway/pull/923).

iamacook: I've included the suggestion in d8d3ab5081b8c96368aeebab93455741f2210c6e with a comment as this will need to be revisited when we are in agreement regarding the [retrieval of the Safe address for unknown transactions](https://github.com/safe-global/safe-client-gateway/pull/923).

src/domain/alerts/alerts.repository.ts
```diff
@@ -38,46 +39,97 @@ export class AlertsRepository implements IAlertsRepository {
...
   async handleAlertLog(chainId: string, log: AlertLog): Promise<void> {
+    const emails = await this.emailRepository.getVerifiedEmailsBySafeAddress({
+      chainId,
+      // TODO: This is the address of the module, _not_ the Safe
+      // Discussion in https://github.com/safe-global/safe-client-gateway/pull/923
+      safeAddress: log.address,
+    });
+
+    if (emails.length === 0) {
+      return this.loggingService.debug(
```
fmrsabino, line 50: nitpick: Logging for me it's more of a side-effect of the function. While it works because of the matching return types, if that changes in the future we might have issues with the return type on this function.

fmrsabino (approved):
🚀

src/domain/alerts/alerts.repository.ts
iamacook: Addressed in 47123ceec383947a3757e50dde2e110255be40a7.

fmrsabino (approved):

PR #923, opened 2023-12-08
feat: fetch `target` for unknown transactions

iamacook:
This adds a proposed method of fetching the `target` of the recovery (Delay Modifier) module when `TransactionAdded` events could not be decoded and thus no Safe address is known.

### DIFF
--- a/src/domain/alerts/alerts.repository.ts
+++ b/src/domain/alerts/alerts.repository.ts
@@ -1,5 +1,12 @@
 import { Inject, Injectable } from '@nestjs/common';
-import { Hex } from 'viem';
+import {
+  Hex,
+  createPublicClient,
+  getAddress,
+  getContract,
+  http,
+  parseAbi,
+} from 'viem';
 import { IAlertsRepository } from '@/domain/alerts/alerts.repository.interface';
 import { IAlertsApi } from '@/domain/interfaces/alerts-api.interface';
 import { AlertsRegistration } from '@/domain/alerts/entities/alerts.entity';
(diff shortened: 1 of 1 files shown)


src/domain/alerts/alerts.repository.ts
```diff
@@ -117,18 +132,53 @@ export class AlertsRepository implements IAlertsRepository {
...
   }
 
-  private async _notifyUnknownTransaction(
-    chainId: string,
-    log: AlertLog,
-  ): Promise<void> {
-    const emails = await this.emailRepository.getVerifiedEmailsBySafeAddress({
-      chainId,
-      safeAddress: log.address,
+  async getRecoveryTarget(args: {
```
hectorgomezv, line 135: Overall I think this is the way to go, but I'd like to have some separation of layers here. I'd consider the blockchain as another data source for us, and I'd extract this functionality fo the `datasources` layer, by creating something like a `BlockchainAPI/OnChainAPI` (the naming can be improved 🙂) with its interface on the domain side, instead of doing this data retrieval here right away. Wdyt? (cc. @fmrsabino)

We could also leave that for a followup PR I guess.

iamacook: I agree that a separate data source makes the most sense. As this [area of the code is evolving](https://github.com/safe-global/safe-client-gateway/pull/922), I will look into creating a new data source that we could then use here.

We can keep this open for discussion until then.

iamacook:
I'm going to close this as we will aim to use the [modules endpoint of the Transaction Service](https://github.com/safe-global/safe-transaction-service/blob/79b1b984c50d127cae8ebf1cb0a38f1484ec72a5/safe_transaction_service/history/views.py#L1202-L1204) instead.

PR #927, opened 2023-12-11
Add timezone_offset query param to transaction routes

fmrsabino:
Adds a `timezone_offset` query parameter to the following routes:
- `GET /v1/chains/:chainId/transactions/:id`
- `GET /v1/chains/:chainId/safes/:safeAddress/multisig-transactions`
- `GET /v1/chains/:chainId/safes/:safeAddress/module-transactions`
- `POST /v1/chains/:chainId/transactions/:safeTxHash/confirmations`
- `GET /v1/chains/:chainId/safes/:safeAddress/incoming-transfers`
- `POST /v1/chains/:chainId/transactions/:safeAddress/propose`

The `timezone_offset` functions in the same way as the Transaction History and Transaction Queue endpoints – given a `timezone_offset` in milliseconds, it will shift the results by that number of milliseconds.

### DIFF
--- a/src/routes/transactions/mappers/module-transactions/module-transaction-details.mapper.spec.ts
+++ b/src/routes/transactions/mappers/module-transactions/module-transaction-details.mapper.spec.ts
@@ -84,6 +84,42 @@ describe('ModuleTransactionDetails mapper (Unit)', () => {
     });
   });
 
+  it('should return an execution date with a timezone offset', async () => {
+    const chainId = faker.string.numeric();
+    const safe = safeBuilder().build();
+    const transaction = moduleTransactionBuilder()
+      .with('safe', safe.address)
+      .build();
+    const txStatus =
+      sample(Object.values(TransactionStatus)) ?? TransactionStatus.Success;
+    statusMapper.mapTransactionStatus.mockReturnValue(txStatus);
+    const txInfo = transferTransactionInfoBuilder().build();
+    transactionInfoMapper.mapTransactionInfo.mockResolvedValue(txInfo);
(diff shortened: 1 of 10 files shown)


src/routes/transactions/mappers/module-transactions/module-transaction-details.mapper.spec.ts
```diff
@@ -109,7 +109,7 @@ describe('ModuleTransactionDetails mapper (Unit)', () => {
       trustedDelegateCallTarget,
     );
 
-    const actual = await mapper.mapDetails(chainId, transaction);
+    const actual = await mapper.mapDetails(chainId, transaction, 0);
```
hectorgomezv, line 112: Is there any reason to pass `0` in these tests? I mean, could it be an opportunity to test that `timezoneOffsetMs` is actually added to the value in the response's `executedAt` property?

hectorgomezv (approved):
Great job 👏🏻 (I've left a minor comment)

src/routes/transactions/mappers/module-transactions/module-transaction-details.mapper.spec.ts
fmrsabino: I agree. Although `0` was used else the expectation of the test would have to be changed. I've added new tests instead. [8fad952](https://github.com/safe-global/safe-client-gateway/pull/927/commits/8fad9525e2dfa87f90facd261053503823e2983b)

hectorgomezv (approved):
🚀🚀🚀

PR #935, opened 2023-12-12
Remove humanDescription feature flag

fmrsabino:
- Removes `features.humanDescription` – this feature is considered stable and is now officially part of the provided API.
- Adds a new feature flag for "rich fragments" – rich fragments are currently under development and are considered to be a separate feature to human-readable descriptions.
- This feature flag can be enabled via `FF_RICH_FRAGMENTS`

### DIFF
--- a/src/routes/transactions/mappers/common/transaction-info.mapper.ts
+++ b/src/routes/transactions/mappers/common/transaction-info.mapper.ts
@@ -65,15 +65,20 @@ export class MultisigTransactionInfoMapper {
     const dataSize =
       dataByteLength >= 2 ? Math.floor((dataByteLength - 2) / 2) : 0;
 
-    const richDecodedInfo = this.isRichFragmentsEnabled
-      ? await this.humanDescriptionMapper.mapRichDecodedInfo(
-          transaction,
-          chainId,
-        )
-      : null;
+    const richDecodedInfo =
+      await this.humanDescriptionMapper.mapRichDecodedInfo(
+        transaction,
+        chainId,
+      );
 
     const humanDescription =
       this.humanDescriptionMapper.mapHumanDescription(richDecodedInfo);
+
+    // If the rich fragment feature is disabled, we set it as undefined.
+    // Undefined properties are not rendered on the response
+    const richDecodedInfoApiProperty = this.isRichFragmentsEnabled
+      ? richDecodedInfo
+      : undefined;
 
     if (this.isCustomTransaction(value, dataSize, transaction.operation)) {
       return await this.customTransactionMapper.mapCustomTransaction(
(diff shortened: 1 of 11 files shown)


src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -65,16 +65,15 @@ export class MultisigTransactionInfoMapper {
...
           transaction,
           chainId,
         )
       : null;
 
-    const humanDescription = this.isHumanDescriptionEnabled
-      ? this.humanDescriptionMapper.mapHumanDescription(richDecodedInfo)
-      : null;
+    const humanDescription =
+      this.humanDescriptionMapper.mapHumanDescription(richDecodedInfo);
```
usame-algan, line 76: This takes `richDecodedInfo` as a parameter which would be `null` if the feature flag is disabled. We could pass `null` to the mapper further down in that case instead.

fmrsabino: Good catch! However this can complicate things as if we set it to `null` afterwards we need to assume that the mappers will never use the `richDecodedInfo` for other parts of the response (excluding `richDecodedInfo`) as it's the case with `humanDescription`.

Setting it to `null` afterwards would be the easiest solution but not sure if it'd be robust compared to performing the check on the final payloads (more occurrences 🤔 )

hectorgomezv: I'm not sure I'm following the logic.

If `this.isRichFragmentsEnabled` is `false`, then `richDecodedInfo` is `null`, and per https://github.com/safe-global/safe-client-gateway/blob/55f36a0b0c54b9d2ee30892c5555c38ed7dd2eaf/src/routes/transactions/mappers/common/human-description.mapper.ts#L82 `humanDescription` will be also `null`. So both `richDecodedInfo` and `humanDescription` will be `null` if `this.isRichFragmentsEnabled` is `false`, right?

If I'm not wrong both variables still depend on the same FF, so no decoupling between features is being made, is that correct?

hectorgomezv (changes requested):

src/routes/transactions/mappers/common/transaction-info.mapper.ts
fmrsabino: We can additionally set the property to `undefined` afterwards as `undefined` properties would not be rendered on the API level.

@usame-algan on the client side do we expect the key `richDecodedInfo` to be present?

usame-algan: I just checked and we don't rely on the existence of the `richDecodedInfo` in safe-wallet-web so it can be safely set to be `undefined` 👍

hectorgomezv: Got it, thanks for explaining @fmrsabino 👍🏻

hectorgomezv (approved):

PR #936, opened 2023-12-12
Remove noncesRoute feature flag

fmrsabino:
Removes the `noncesRoute` feature flag. The route to retrieve the safe address nonce data is now available by default in `GET /v1/chains/:chainId/safes/:safeAddress/nonces`


hectorgomezv (approved):
🚀🚀

PR #944, opened 2023-12-15
Rename signer property to account

fmrsabino:
Renames the `signer` reference to `account` – `account` is more abstract and encapsulates better the meaning behind this property i.e.: the signer was the one who might have triggered an action but does not necessarily mean that the email is to be associated to that signer (e.g.: signer can have permissions to associate the email to another account).


src/routes/email/email.service.ts
```diff
@@ -23,7 +23,7 @@ export class EmailService {
   }): Promise<void> {
     return this.repository.resendEmailVerification({
       ...args,
-      signer: args.account,
+      account: args.account,
     });
```
hectorgomezv, line 27: I think this can be simplified to:
```javascript
return this.repository.resendEmailVerification(args);
```

src/routes/email/email.service.ts
```diff
@@ -35,7 +35,7 @@ export class EmailService {
   }): Promise<any> {
     return this.repository.verifyEmailAddress({
       ...args,
-      signer: args.account,
+      account: args.account,
     });
```
hectorgomezv, line 39: And this can be now:
```javascript
return this.repository.verifyEmailAddress(args);
```

hectorgomezv (changes requested):

src/routes/email/email.service.ts
fmrsabino: Good catch! Addressed in [6990105](https://github.com/safe-global/safe-client-gateway/pull/944/commits/6990105ff17819e91620f80a63d99667b1dca99f)

src/routes/email/email.service.ts
fmrsabino: [6990105](https://github.com/safe-global/safe-client-gateway/pull/944/commits/6990105ff17819e91620f80a63d99667b1dca99f)

fmrsabino:
@safe-global/safe-client-gateway-maintainers Forgot to mention: the migration was edited directly. This was done because it hasn't been integrated with a live environment yet so it would not be worth it to create another migration until that happens.

hectorgomezv (approved):
🚀🚀

PR #945, opened 2023-12-15
Add missing interface declaration

fmrsabino:
`BackboneRepository` should implement `IBackboneRepository`. The mapping is already being done in the `DomainModule`.


hectorgomezv (approved):
👏🏻 good catch!

PR #947, opened 2023-12-18
refactor: return `this` from `IBuilder['with']`

iamacook:
This adjusts the `ReturnType` of `IBuilder['with']` in preparation for [extending it](https://github.com/safe-global/safe-client-gateway/pull/921/files#r1422237932). Calling `with` currently returns a `Builder` instance, when we should otherwise be returning the potentially extended instance.


fmrsabino (approved):
🚀

PR #950, opened 2023-12-18
feat: add blockchain data source

iamacook:
In preparation for retrieving [Safe addresses via recovery modules](https://github.com/safe-global/safe-client-gateway/pull/923#discussion_r1430128476), this adds a `IBlockchainDataSource` for retrieval of the viem `PublicClient`. The following functionality is avaialable:

- `getPublicClient` - returns a viem `PublicClient` for interfacing with the given chain.

`createPublicClient` of `viem` is injected directly within the `BlockchainDataSourceModule` instead of a wrapped version as we already have "viem-opinionated" code in the codebase and would otherwise need create a vast interface.


iamacook:
I'm going to close this as we will aim to use the [modules endpoint of the Transaction Service](https://github.com/safe-global/safe-transaction-service/blob/79b1b984c50d127cae8ebf1cb0a38f1484ec72a5/safe_transaction_service/history/views.py#L1202-L1204) instead.

PR #953, opened 2023-12-18
feat: add `/api/v1/modules/{address}/safes/` to `TransactionApi`

iamacook:
That extends the `TransactionApi` to support the [`/v1/modules/{address}/safes/` endpoint of the Transaction Service](https://github.com/safe-global/safe-transaction-service/blob/79b1b984c50d127cae8ebf1cb0a38f1484ec72a5/safe_transaction_service/history/views.py#L1202-L1204), adding the following methods:

  - `getSafesByModule` - fetches and caches endpoint result
  - `clearSafesByModule` - clears existing cache of endpoint result

Note: `clearSafesByModule` will be integrated in a follow up PR.

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -14,6 +14,7 @@ export class CacheRouter {
   private static readonly MASTER_COPIES_KEY = 'master_copies';
   private static readonly MESSAGE_KEY = 'message';
   private static readonly MESSAGES_KEY = 'messages';
+  private static readonly MODULES_KEY = 'modules';
   private static readonly MODULE_TRANSACTION_KEY = 'module_transaction';
   private static readonly MODULE_TRANSACTIONS_KEY = 'module_transactions';
   private static readonly MULTISIG_TRANSACTION_KEY = 'multisig_transaction';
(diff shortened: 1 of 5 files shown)


src/datasources/cache/cache.router.ts
```diff
@@ -14,6 +14,7 @@ export class CacheRouter {
   private static readonly MASTER_COPIES_KEY = 'master_copies';
   private static readonly MESSAGE_KEY = 'message';
   private static readonly MESSAGES_KEY = 'messages';
+  private static readonly MODULES_KEY = 'modules';
```
hectorgomezv, line 17: Nitpick: if I'm not wrong we are storing Safe addresses behind this key, so I'd use another name, like `module_safes` (taken from `owner_safes` example).

hectorgomezv:
I've left a small comment only, it looks good otherwise, great job! 👏🏻

src/datasources/cache/cache.router.ts
iamacook: I changed it in 970f68198f654722658cdfce7929e210556aecb3 / bba9179e176321e4aaf6eacd0b7a4c475ded8417 to `module_safes`.

src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -173,4 +173,41 @@ describe('TransactionApi', () => {
       expect(httpErrorFactory.from).toHaveBeenCalledTimes(1);
     });
   });
+
+  describe('Modules', () => {
+    it('should return Safes with module enables', async () => {
```
fmrsabino, line 178: typo:

```suggestion
    it('should return Safes with module enabled', async () => {
```

fmrsabino (approved):
👏

src/datasources/transaction-api/transaction-api.service.spec.ts
iamacook: Fixed in 7655818de0c1138f1f72ba1f6b89921f926fe3b0.

fmrsabino (approved):

PR #954, opened 2023-12-18
feat: add `getSafesByModule` and `clearSafesByModule` to `SafeRepository`

iamacook:
This adds the following functionality to the `SafeRepository`, as well as integrating it with the `CacheHooksService`:

  - `getSafesByModule` - fetches and caches endpoint result
  - `clearSafesByModule` - clears existing cache of endpoint result

`clearSafesByModule` is called within the `CacheHooksService` whenever a `EventType.MODULE_TRANSACTION` event occurs.

Depends on https://github.com/safe-global/safe-client-gateway/pull/953.

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -150,10 +150,7 @@ export class CacheRouter {
     chainId: string;
     moduleAddress: string;
   }): CacheDir {
-    return new CacheDir(
-      this.getSafesByModuleCacheKey(args), // TODO: Refactor other `CacheDir` keys to use respective key methods
-      '',
-    );
+    return new CacheDir(CacheRouter.getSafesByModuleCacheKey(args), '');
   }
 
   static getSafesByModuleCacheKey(args: {
--- a/src/routes/cache-hooks/cache-hooks.controller.spec.ts
+++ b/src/routes/cache-hooks/cache-hooks.controller.spec.ts
@@ -768,4 +768,37 @@ describe('Post Hook Events (Unit)', () => {
       webHookExecutionDelayMs,
     );
   });
+
+  it.each([
+    {
+      type: 'MODULE_TRANSACTION',
+      module: faker.finance.ethereumAddress(),
+      txHash: faker.string.hexadecimal({ length: 32 }),
+    },
+  ])('$type clears Safes by module', async (payload) => {
+    const chainId = faker.string.numeric();
+    const safeAddress = faker.finance.ethereumAddress();
+    const moduleAddress = faker.finance.ethereumAddress();
+    const cacheDir = new CacheDir(
+      `${chainId}_module_safes_${moduleAddress}`,
+      '',
+    );
+    await fakeCacheService.set(cacheDir, faker.string.alpha());
+    const data = {
+      address: safeAddress,
+      chainId: chainId,
+      ...payload,
+    };
+
+    await request(app.getHttpServer())
+      .post(`/hooks/events`)
+      .set('Authorization', `Basic ${authToken}`)
+      .send(data)
+      .expect(202);
+
+    setTimeout(
+      () => expect(fakeCacheService.get(cacheDir)).resolves.toBeNull(),
+      webHookExecutionDelayMs,
+    );
+  });
 });
--- a/src/routes/cache-hooks/cache-hooks.service.ts
+++ b/src/routes/cache-hooks/cache-hooks.service.ts
@@ -108,6 +108,10 @@ export class CacheHooksService {
       // - the transaction executed – clear multisig transaction
       // - the safe configuration - clear safe info
       case EventType.EXECUTED_MULTISIG_TRANSACTION:
+        // Important: we should be clearing Safes by module here but we don't have module address
+        // As this is only being used for recovery, whereby our UI deploys new module per Safe
+        // we can assume that the Safes-module relationship does not change for said module
+
         promises.push(
           this.collectiblesRepository.clearCollectibles({
             chainId: event.chainId,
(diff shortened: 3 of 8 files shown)


src/routes/cache-hooks/cache-hooks.controller.spec.ts
```diff
@@ -768,4 +768,54 @@ describe('Post Hook Events (Unit)', () => {
...
+      `${chainId}_modules_${moduleAddress}`,
+      faker.string.alpha(),
+    );
+    await fakeCacheService.set(cacheDir, faker.string.alpha());
+    const data = {
+      address: safeAddress,
+      chainId: chainId,
+      ...payload,
+    };
+    networkService.get.mockImplementation((url) => {
```
hectorgomezv, line 799: Why is this network call mock needed? The cache key is already populated by `fakeCacheService.set` in the line 793, right?

src/domain/modules/modules.repository.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface IModulesRepository {
+  getSafesByModule(args: {
+    chainId: string;
+    moduleAddress: string;
+  }): Promise<SafesByModule>;
+
+  clearSafesByModule(args: {
+    chainId: string;
+    moduleAddress: string;
+  }): Promise<void>;
```
hectorgomezv, line 14: Nitpick: no strong opinion regarding this, but given these methods retrieve/clear Safe addresses and not modules, do you think it may have sense to have them on `SafeRepository` instead of creating this new `ModulesRepository`?

src/domain/modules/modules.repository.ts
```diff
@@ -0,0 +1,32 @@
...
+  ) {}
+
+  async getSafesByModule(args: {
+    chainId: string;
+    moduleAddress: string;
+  }): Promise<SafesByModule> {
+    const api = await this.transactionApiManager.getTransactionApi(
+      args.chainId,
+    );
+    return await api.getSafesByModule(args.moduleAddress);
```
hectorgomezv, line 20: I'd add validation to this function before returning the result to the caller.

I think that we could even use an existing validator, `SafeListValidator`, what do you think?

hectorgomezv (changes requested):
Great job!

src/routes/cache-hooks/cache-hooks.controller.spec.ts
iamacook: It's a leftover from using another test as a reference 🙈 I've removed it in cbc68a8b4ec8699a56ccdd1a95c5226325bec844.

src/domain/modules/modules.repository.interface.ts
iamacook: I've migrated it in cbc68a8b4ec8699a56ccdd1a95c5226325bec844.

src/domain/modules/modules.repository.ts
iamacook: Added in cbc68a8b4ec8699a56ccdd1a95c5226325bec844.

src/routes/cache-hooks/cache-hooks.service.ts
```diff
@@ -78,6 +78,7 @@ export class CacheHooksService {
       // - the list of all executed transactions for the safe
       // - the list of module transactions for the safe
       // - the safe configuration
+      // - the list of Safes for the module
```
fmrsabino, line 81: Adding/Removing a module is not necessarily a module transaction I believe (from the Safe Transaction Service).

hectorgomezv: That's a good point and something I've also thought about, it doesn't imply a module transaction, in fact, I think it should be a Multisig one, right? In that case, we would need to clear also on `EXECUTED_MULTISIG_TRANSACTION` event.

src/datasources/cache/cache.router.ts
```diff
@@ -153,10 +153,7 @@ export class CacheRouter {
     chainId: string;
     moduleAddress: string;
   }): CacheDir {
-    return new CacheDir(
-      this.getSafesByModuleCacheKey(args), // TODO: Refactor other `CacheDir` keys to use respective key methods
-      '',
-    );
```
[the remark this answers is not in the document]
hectorgomezv: I think it's due to this PR: https://github.com/safe-global/safe-client-gateway/pull/960 But yeah, being strict it can be removed in that PR instead 😄

src/routes/cache-hooks/cache-hooks.service.ts
fmrsabino: A Module should be able to technically change the state of the Safe so I believe that having this under a Module Event is correct but not complete. An `EXECUTED_MULTISIG_TRANSACTION` should also be able to change the state of the Safe.

fmrsabino: Actually it is likely that both of these events can have the same side-effects. We should consider merging them if that's the case.

src/datasources/cache/cache.router.ts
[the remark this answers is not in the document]
iamacook: Due to the order of PRs, #960 was merged before this so the note was no longer needed.

src/routes/cache-hooks/cache-hooks.service.ts
iamacook: After discussion and due to the required changes for clearing the cache on `EXECUTED_MULTISIG_TRANSACTION` events, I have added a `TODO` in the relevant place and will tackle this in a follow up PR.

iamacook: As our UI deploys a new recovery module per Safe we can also assume that the Safes-module relationship won't change so invalidating the cache shouldn't be an issue for now. (I've updated the aforementioned comment accordingly.)

hectorgomezv (approved):

PR #960, opened 2023-12-19
refactor: call key methods for `CacheDir` keys

iamacook:
There is some disparity of `get*CacheDir` functions (not) using their respective `get*CacheKey` getters in the `CacheRouter`. If they exist, all `CacheDir`s now use their key getters.


fmrsabino (approved):
🚀

hectorgomezv (approved):

PR #961, opened 2023-12-19
fix: retrieve Safe address via recovery module address

iamacook:
This adjusts recovery alerts to fetch the Safe address via the module address of the alert as well as adjusting test coverage accordingly.

Depends on https://github.com/safe-global/safe-client-gateway/pull/953 then https://github.com/safe-global/safe-client-gateway/pull/954.

### DIFF
--- a/src/domain/alerts/alerts.repository.ts
+++ b/src/domain/alerts/alerts.repository.ts
@@ -10,7 +10,6 @@ import { MultiSendDecoder } from '@/domain/alerts/contracts/multi-send-decoder.h
 import { IEmailApi } from '@/domain/interfaces/email-api.interface';
 import { IEmailRepository } from '@/domain/email/email.repository.interface';
 import { ILoggingService, LoggingService } from '@/logging/logging.interface';
-import { IModulesRepository } from '@/domain/modules/modules.repository.interface';
 import { IConfigurationService } from '@/config/configuration.service.interface';
 import { ISafeRepository } from '@/domain/safe/safe.repository.interface';
 import { Safe } from '@/domain/safe/entities/safe.entity';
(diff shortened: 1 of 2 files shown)


src/domain/alerts/alerts.repository.ts
```diff
@@ -32,77 +33,83 @@ export class AlertsRepository implements IAlertsRepository {
...
-    const decodedEvent = this.delayModifierDecoder.decodeEventLog({
-      data: log.data as Hex,
-      topics: log.topics as [Hex, Hex, Hex],
-    });
+    // Recovery module is deployed per Safe so we can assume that it is only enabled on one
+    const safeAddress = safes[0];
 
-    const decodedTransactions = this._decodeTransactionAdded(
-      decodedEvent.args.data,
-    );
```
iamacook, line 63: This was moved into the `try` to ensure users are notified.

src/domain/alerts/alerts.repository.ts
```diff
@@ -32,77 +33,83 @@ export class AlertsRepository implements IAlertsRepository {
...
-        return [decoded];
-      }
+    const decoded = this.safeDecoder.decodeFunctionData({ data });
 
-      const execTransactionData = decoded.args[2];
-      return this.multiSendDecoder
-        .mapMultiSendTransactions(execTransactionData)
-        .flatMap(({ data }) => this.safeDecoder.decodeFunctionData({ data }));
-    } catch {
-      return null;
```
iamacook, line 104: This will be caught within `handleAlertLog` and therefore notify users.

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -181,9 +181,9 @@ describe('Alerts (Unit)', () => {
           switch (url) {
             case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
               return Promise.resolve({ data: chain });
-            case `${chain.transactionService}/api/v1/safes/${getAddress(
```
iamacook, line 184: As we are now fetching the Safe address via the Transaction Service and _not_ decoding it with `viem`, we do not need to checksum this here.

hectorgomezv (approved):
🚀

PR #963, opened 2023-12-19
Sort supported currency codes 

hectorgomezv:
**Context**
This is a follow-up from this comment: https://github.com/safe-global/safe-client-gateway/pull/898#discussion_r1425269639

**Changes**
- Adds a log when the mandatory fiat code (`USD`) is not returned by the provider's supported currencies list.
- Sort the returned codes alphabetically.

### DIFF
--- a/src/domain/prices/entities/schemas/fiat-codes.schema.ts
+++ b/src/domain/prices/entities/schemas/fiat-codes.schema.ts
@@ -8,6 +8,10 @@ export const fiatCodesSchema: JSONSchemaType<FiatCodes> = {
   $id: FIAT_CODES_SCHEMA_ID,
   type: 'array',
   items: { type: 'string' },
+  allOf: [
+    { contains: { type: 'string', enum: ['usd'] } },
+    { contains: { type: 'string', enum: ['eur'] } },
+  ],
   minItems: 1,
   uniqueItems: true,
 };
--- a/src/domain/prices/entities/schemas/fiat-codes.schema.ts
+++ b/src/domain/prices/entities/schemas/fiat-codes.schema.ts
@@ -8,10 +8,6 @@ export const fiatCodesSchema: JSONSchemaType<FiatCodes> = {
   $id: FIAT_CODES_SCHEMA_ID,
   type: 'array',
   items: { type: 'string' },
-  allOf: [
-    { contains: { type: 'string', enum: ['usd'] } },
-    { contains: { type: 'string', enum: ['eur'] } },
-  ],
   minItems: 1,
   uniqueItems: true,
 };
--- a/src/domain/prices/prices.repository.interface.ts
+++ b/src/domain/prices/prices.repository.interface.ts
@@ -16,7 +16,7 @@ export interface IPricesRepository {
 
   /**
    * Gets the list of supported fiat codes.
-   * @returns alphabetically ordered list of uppercase strings representing the supported fiat codes.
+   * @returns an alphabetically ordered list of uppercase strings representing the supported fiat codes.
    */
   getFiatCodes(): Promise<string[]>;
 }
--- a/src/domain/prices/prices.repository.interface.ts
+++ b/src/domain/prices/prices.repository.interface.ts
@@ -16,8 +16,7 @@ export interface IPricesRepository {
 
   /**
    * Gets the list of supported fiat codes.
-   * USD and EUR fiat codes are fixed, and should be the first ones in the result list.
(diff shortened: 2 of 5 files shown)


src/domain/prices/entities/schemas/fiat-codes.schema.ts
```diff
@@ -8,6 +8,10 @@ export const fiatCodesSchema: JSONSchemaType<FiatCodes> = {
   $id: FIAT_CODES_SCHEMA_ID,
   type: 'array',
   items: { type: 'string' },
+  allOf: [
+    { contains: { type: 'string', enum: ['usd'] } },
+    { contains: { type: 'string', enum: ['eur'] } },
+  ],
```
fmrsabino, line 14: Why do we have this requirement?

hectorgomezv: This is inherited from the documentation written for the Rust CGW implementation. 

I took this as an informal requirement, but as an alternative way, if you think this is excessive, we could also sort the array and put those codes (`USD, EUR`) only if they are contained in the provider response. No strong opinion on this.

fmrsabino: According to the comment that you shared and the one available [here](https://github.com/5afe/safe-client-gateway/pull/315), it mentions that USD and EUR should come first but it doesn't state that they should be present.

That's an important difference as we are now setting both USD and EUR as a requirement to the provider. While this can be considered it was not a requirement so far. That is different than if `USD` and `EUR` are present in the response they should come first.

hectorgomezv: I agree, in fact, I think the existent implementation could be seen as a bug, as it adds artificially the `USD, EUR` fiat codes to the response array. This gives incorrect data to the clients, that might ask for a token price denominated in those currencies when they are not actually available.

So summing it up, what do you think about the proposal:

> we could also sort the array and put those codes (USD, EUR) only if they are contained in the provider response

fmrsabino: > So summing it up, what do you think about the proposal:
> 
> > we could also sort the array and put those codes (USD, EUR) only if they are contained in the provider response

Currently, I don't think there's any benefit in shifting their positions in the return so I would leave the order as is returned by the provider.

hectorgomezv: I've removed the hard validation part in [7873f7c](https://github.com/safe-global/safe-client-gateway/pull/963/commits/7873f7c30b368a27c62fc75c0ba29f2c214adc56)

src/domain/prices/prices.repository.ts
```diff
@@ -44,6 +46,18 @@ export class PricesRepository implements IPricesRepository {
 
   async getFiatCodes(): Promise<string[]> {
     const result = await this.coingeckoApi.getFiatCodes();
-    return this.fiatCodesValidator.validate(result);
+
+    const fiatCodes = this.fiatCodesValidator
+      .validate(result)
+      .map((item) => item.toUpperCase())
```
fmrsabino, line 52: Wonder if this should be seen as a route requirement and not a domain one (the `toUpperCase` one).

Given its usage, this is mostly a nitpick though.

fmrsabino (approved):
🚀

src/domain/prices/prices.repository.ts
hectorgomezv: Hmm, good point, I guess that depends on whatever we consider. No strong opinion on my side, if you consider this should be moved to the routes layer we can change it of course 🙂

src/domain/prices/prices.repository.interface.ts
```diff
@@ -14,5 +14,9 @@ export interface IPricesRepository {
     fiatCode: string;
   }): Promise<AssetPrice[]>;
 
+  /**
+   * Gets the list of supported fiat codes.
+   * @returns alphabetically ordered list of uppercase strings representing the supported fiat codes.
```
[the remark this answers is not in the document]
hectorgomezv: Thanks! (I had to re-request your review after merging the suggestion)

src/domain/prices/prices.repository.ts
fmrsabino: Let's leave it like this for now. It's a specific use-case as is stands.

fmrsabino (approved):

PR #964, opened 2023-12-19
fix: remove Safes by modules caching

iamacook:
Depends on #961

This removes caching of Safes by modules as it was deemed an unnecessary optimisation that we will revisit later if necessary.

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.spec.ts
+++ b/src/datasources/transaction-api/transaction-api.service.spec.ts
@@ -173,41 +173,4 @@ describe('TransactionApi', () => {
       expect(httpErrorFactory.from).toHaveBeenCalledTimes(1);
     });
   });
-
-  describe('Modules', () => {
-    it('should return Safes with module enabled', async () => {
-      const moduleAddress = faker.finance.ethereumAddress();
-      const safesByModule = {
-        safes: [
-          faker.finance.ethereumAddress(),
-          faker.finance.ethereumAddress(),
-        ],
-      };
-      mockDataSource.get.mockResolvedValueOnce(safesByModule);
-
-      const actual = await service.getSafesByModule(moduleAddress);
-
-      expect(actual).toBe(safesByModule);
-      expect(mockDataSource.get).toHaveBeenCalledWith({
-        cacheDir: new CacheDir(`${chainId}_module_safes_${moduleAddress}`, ''),
-        url: `${baseUrl}/api/v1/modules/${moduleAddress}/safes/`,
-        notFoundExpireTimeSeconds: notFoundExpireTimeSeconds,
-        expireTimeSeconds: defaultExpirationTimeInSeconds,
-      });
-      expect(httpErrorFactory.from).toHaveBeenCalledTimes(0);
-    });
-
-    it('should map error on error', async () => {
-      const moduleAddress = faker.finance.ethereumAddress();
-      const error = new Error('some error');
-      const expected = new DataSourceError('some data source error');
-      mockDataSource.get.mockRejectedValueOnce(error);
-      mockHttpErrorFactory.from.mockReturnValue(expected);
-
-      await expect(service.getSafesByModule(moduleAddress)).rejects.toThrow(
-        expected,
-      );
-      expect(httpErrorFactory.from).toHaveBeenCalledTimes(1);
-    });
-  });
 });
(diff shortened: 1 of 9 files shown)


src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -178,12 +178,13 @@ describe('Alerts (Unit)', () => {
         );
 
         networkService.get.mockImplementation((url) => {
+          console.log(url);
```
hectorgomezv, line 181: `console.log` left here 😄

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -178,12 +178,13 @@ describe('Alerts (Unit)', () => {
...
           switch (url) {
             case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
               return Promise.resolve({ data: chain });
-            case `${chain.transactionService}/api/v1/safes/${getAddress(
-              safe.address,
-            )}`:
+            case `${chain.transactionService}/api/v1/modules/${delayModifier}/safes/`:
```
hectorgomezv, line 185: These changes are included in another PR if I'm not wrong (probably related to the conflict regarding this file).

src/domain/alerts/alerts.repository.ts
```diff
@@ -39,70 +39,74 @@ export class AlertsRepository implements IAlertsRepository {
   }
 
   async handleAlertLog(chainId: string, log: AlertLog): Promise<void> {
-    const emails = await this.emailRepository.getVerifiedEmailsBySafeAddress({
+    const moduleAddress = log.address;
```
hectorgomezv, line 42: This file needs rebasing I think

src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -173,41 +173,4 @@ describe('TransactionApi', () => {
       expect(httpErrorFactory.from).toHaveBeenCalledTimes(1);
     });
   });
-
```
hectorgomezv, line 176: Why are these tests removed? They still relevant, as the endpoint still exist (but it uses NetworkService directly instead of CacheFirstDatasource), or am I missing something?

hectorgomezv (changes requested):
Awesome job, I've just left minor comments 😄

src/routes/alerts/alerts.controller.spec.ts
iamacook: 🙈 Removed in 8469ed0a094b62c49b0ca23355456cc35c32e3ad.

src/routes/alerts/alerts.controller.spec.ts
iamacook: Rebasing fixed this.

src/domain/alerts/alerts.repository.ts
iamacook: Indeed. Should be fixed now.

src/datasources/transaction-api/transaction-api.service.spec.ts
iamacook: You're correct. I've added them again in the context of not caching in 8469ed0a094b62c49b0ca23355456cc35c32e3ad.

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -178,11 +178,12 @@ describe('Alerts (Unit)', () => {
         );
 
         networkService.get.mockImplementation((url) => {
+          console.log(url);
           switch (url) {
             case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
               return Promise.resolve({ data: chain });
             case `${chain.transactionService}/api/v1/modules/${delayModifier}/safes/`:
-              return Promise.resolve({ data: { safes: [safe.address] } });
+              return Promise.resolve({ safes: [safe.address] });
```
fmrsabino, line 185: This change can be a bit confusing since the expected return type has been `NetworkResponse`.

I know that this is a default and not a requirement but I wonder if we should change that since this seems like an outlier 🤔

fmrsabino (approved):

src/routes/alerts/alerts.controller.spec.ts
iamacook: Good catch. I missed an inconsistency between the `CacheFirstDataSource` and using the `NetworkService` directly. I've adjusted it in 98d925cac1549b16372a4d297e4e96c60f14d219 accordingly.

fmrsabino (approved):
🚀

hectorgomezv (approved):

Issue #893, opened 2023-11-29
Check request lifespan instead of using a webhook execution delay


hectorgomezv:
**Context**
As a result of https://github.com/safe-global/safe-client-gateway/issues/881 a delay in the webhooks execution was introduced. This delay prevents race conditions between the services when the cache invalidation associated with a webhook happens while there's an ongoing request to the TXS, that stores outdated data in the cache.

**Proposed solution**
- Remove the delay, as it affects every webhook, not only the ones impacted by the race condition described above.
- Implement a mechanism to check the **last invalidation time for a given key is older than the start time for the ongoing request** that is trying to write the cache key. (TBD)

PR #966, opened 2023-12-20
Refactor CacheFirstDataSource

hectorgomezv:
**Context**
This is preliminary work before implementing https://github.com/safe-global/safe-client-gateway/issues/893, because that task would increase the cyclomatic complexity of the `CacheFirstDataSource.get` function. So the idea is to decrease its complexity beforehand.

**Changes:**
- Refactor `CacheFirstDataSource.get` by splitting its logic into two new private functions.

### DIFF
--- a/src/datasources/cache/cache.first.data.source.ts
+++ b/src/datasources/cache/cache.first.data.source.ts
@@ -69,7 +69,7 @@ export class CacheFirstDataSource {
   /**
    * Gets the data from the contents stored in the cache.
    */
-  private _getFromCachedData<T>({ key, field }, cached: string): Promise<T> {
+  private _getFromCachedData<T>({ key, field }: CacheDir, cached: string): Promise<T> {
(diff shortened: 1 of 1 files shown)


src/datasources/cache/cache.first.data.source.ts
```diff
@@ -91,6 +66,38 @@ export class CacheFirstDataSource {
     }
   }
 
+  /**
+   * Gets the data from the contents stored in the cache.
+   */
+  private _getFromCachedData<T>({ key, field }, cached: string): Promise<T> {
```
iamacook, line 72:
```suggestion
  private _getFromCachedData<T>({ key, field }: CacheDir, cached: string): Promise<T> {
```

iamacook (approved):

iamacook (approved):

hectorgomezv:
Refactor `CacheFirstDataSource.get` by splitting its logic into two new private functions.

PR #968, opened 2023-12-21
Add email update functionality

iamacook:
The adds the following methods to the `EmailDataSource`:

- `updateEmail` - updates an entry for the given Safe/account, setting it as unverified.

When calling the above method, the `EmailService` sets the provided email as unverified and dispatches a verification email.

This functionality is propagated to the `chains/:chainId/safes/:safeAddress/emails` endpoint via a [new `PUT` method](https://www.notion.so/safe-global/Technical-Specification-Email-Registration-1b897da501d541ce973cd846bbca4763?pvs=4#b266ad747d624e1da3b514556bffd575). It is guarded by timestamp, owner and signature (prefixed with `email-update-`) guards.

### DIFF
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -329,4 +329,67 @@ describe('Email Datasource Tests', () => {
       }),
     ).rejects.toThrow(EmailAddressDoesNotExistError);
   });
+
+  it('updates emails successfully', async () => {
+    const chainId = faker.number.int({ max: DB_CHAIN_ID_MAX_VALUE });
+    const safeAddress = faker.finance.ethereumAddress();
+    const emailAddress = new EmailAddress(faker.internet.email());
+    const account = faker.finance.ethereumAddress();
+    const code = faker.string.numeric();
+    const codeGenerationDate = faker.date.recent();
+
+    await target.saveEmail({
+      chainId: chainId.toString(),
+      safeAddress,
+      emailAddress: new EmailAddress(faker.internet.email()),
+      account,
+      code: faker.string.numeric(),
+      codeGenerationDate: faker.date.recent(),
+    });
+
+    await target.updateEmail({
+      chainId: chainId.toString(),
+      safeAddress,
+      emailAddress,
+      account,
+      code,
+      codeGenerationDate,
+    });
+
+    const email = await target.getEmail({
+      chainId: chainId.toString(),
+      safeAddress,
+      account,
+    });
+
+    expect(email).toMatchObject({
+      chainId: chainId.toString(),
+      emailAddress,
+      isVerified: false,
+      safeAddress,
+      account,
+      verificationCode: code,
+      verificationSentOn: null,
+    });
+  });
+
+  it('updating a non-existent email throws', async () => {
+    const chainId = faker.number.int({ max: DB_CHAIN_ID_MAX_VALUE });
+    const safeAddress = faker.finance.ethereumAddress();
+    const emailAddress = new EmailAddress(faker.internet.email());
+    const account = faker.finance.ethereumAddress();
+    const code = faker.string.numeric();
+    const codeGenerationDate = faker.date.recent();
+
+    await expect(
+      target.updateEmail({
+        chainId: chainId.toString(),
+        safeAddress,
+        emailAddress,
+        account,
+        code,
+        codeGenerationDate,
+      }),
+    ).rejects.toThrow(EmailAddressDoesNotExistError);
+  });
 });
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -331,24 +331,31 @@ describe('Email Datasource Tests', () => {
   });
 
   it('updates emails successfully', async () => {
-    const chainId = faker.number.int({ max: DB_CHAIN_ID_MAX_VALUE });
+    const chainId = faker.number.int({ max: DB_CHAIN_ID_MAX_VALUE }).toString();
     const safeAddress = faker.finance.ethereumAddress();
+    const prevEmailAddress = new EmailAddress(faker.internet.email());
     const emailAddress = new EmailAddress(faker.internet.email());
     const account = faker.finance.ethereumAddress();
     const code = faker.string.numeric();
     const codeGenerationDate = faker.date.recent();
 
     await target.saveEmail({
-      chainId: chainId.toString(),
+      chainId,
       safeAddress,
-      emailAddress: new EmailAddress(faker.internet.email()),
+      emailAddress: prevEmailAddress,
       account,
       code: faker.string.numeric(),
       codeGenerationDate: faker.date.recent(),
     });
 
+    await target.verifyEmail({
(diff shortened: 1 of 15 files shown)


src/routes/email/guards/email-update.guard.ts
```diff
@@ -0,0 +1,71 @@
...
+ * To use this guard, the route should have:
+ * - the 'chainId' declared as a parameter
+ * - the 'safeAddress' declared as a parameter
+ * - the 'emailAddress' as part of the JSON body (top level)
+ * - the 'account' as part of the JSON body (top level)
+ * - the 'signature' as part of the JSON body (top level) - see message format to be signed
+ * - the 'timestamp' as part of the JSON body (top level)
+ */
+@Injectable()
+export class EmailUpdateGuard implements CanActivate {
```
iamacook, line 30: I have created a new guard for now but would like to revisit this in a follow up PR to refactor the three into one.

src/datasources/email/email.datasource.ts
```diff
@@ -175,4 +175,30 @@ export class EmailDataSource implements IEmailDataSource {
...
+
+  async updateEmail(args: {
+    chainId: string;
+    safeAddress: string;
+    emailAddress: EmailAddress;
+    account: string;
+    code: string;
+    codeGenerationDate: Date;
+  }): Promise<void> {
+    const [email] = await this.sql<Email[]>`UPDATE emails.account_emails
```
fmrsabino, line 187: An update to the email address should set the `verified` field to `false`.

src/domain/email/email.repository.ts
```diff
@@ -149,6 +150,33 @@ export class EmailRepository implements IEmailRepository {
...
+    chainId: string;
+    safeAddress: string;
+    emailAddress: string;
+    account: string;
+  }): Promise<void> {
+    const email = new EmailAddress(args.emailAddress);
+    const verificationCode = this._generateCode();
+
+    try {
+      await this.emailDataSource.updateEmail({
```
fmrsabino, line 163: What if the email address that we are updating is the same.

E.g.: A verified `foo@example.com` is updated with `foo@example.com`. This action should probably result in a client error to signal that the same email was provided, and we don't update it on our side (as it stands, updating it to the same value would result in setting a verified email as unverified).

src/routes/email/email.controller.update-email.spec.ts
```diff
@@ -0,0 +1,212 @@
...
+import { generatePrivateKey, privateKeyToAccount } from 'viem/accounts';
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { NetworkService } from '@/datasources/network/network.service.interface';
+import { IEmailDataSource } from '@/domain/interfaces/email.datasource.interface';
+import { chainBuilder } from '@/domain/chains/entities/__tests__/chain.builder';
+import { safeBuilder } from '@/domain/safe/entities/__tests__/safe.builder';
+import { getAddress } from 'viem';
+import { EmailControllerModule } from '@/routes/email/email.controller.module';
+
+describe('Email controller update email tests', () => {
```
fmrsabino, line 24: It looks like we are missing a test which tests the expected response if an `EmailUpdateError` is thrown.

fmrsabino (changes requested):

src/datasources/email/email.datasource.ts
iamacook: Adjusted in 234b4ce38efb7be32c9940dff56dcc943e511a90.

src/routes/email/email.controller.update-email.spec.ts
iamacook: Added in 234b4ce38efb7be32c9940dff56dcc943e511a90.

src/routes/email/email.controller.update-email.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+        account: account.address,
+        timestamp,
+        signature,
+      })
+      .expect(409)
+      .expect({
+        statusCode: 409,
+        message: 'Email address matches that of the Safe owner.',
+      });
+  });
```
fmrsabino, line 149: nitpick: maybe it's worth to add an additional assertion here – checking if the update was not called at all even with 409 as a result.

```typescript
expect(emailDatasource.updateEmail).toBeCalledTimes(0)
```

Should apply to other tests also (which have the same expectation).

src/routes/email/email.controller.update-email.spec.ts
fmrsabino: It's still missing right? 🤔 `EmailUpdateError` in this case should return a `500` or similar.

src/datasources/email/email.datasource.spec.ts
```diff
@@ -329,4 +329,74 @@ describe('Email Datasource Tests', () => {
...
+      account,
+      code: faker.string.numeric(),
+      codeGenerationDate: faker.date.recent(),
+    });
+
+    await target.verifyEmail({
+      chainId,
+      safeAddress,
+      account,
+    });
```
fmrsabino, line 355: Verification is optional when updating an email right? As in we can update unverified emails.

fmrsabino (changes requested):

src/routes/email/email.controller.update-email.spec.ts
```diff
@@ -0,0 +1,305 @@
...
+          return Promise.resolve({ data: chain });
+        case `${chain.transactionService}/api/v1/safes/${safe.address}`:
+          return Promise.resolve({ data: safe });
+        default:
+          return Promise.reject(new Error(`Could not match ${url}`));
+      }
+    });
+    emailDatasource.getEmail.mockResolvedValue({
+      emailAddress: new EmailAddress(prevEmailAddress),
+    });
```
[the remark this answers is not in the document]
iamacook: Added in 5674ff88d2bdb104e784280af3968346b9446d87.

src/routes/email/email.controller.update-email.spec.ts
iamacook: Added to the relevant tests in 5674ff88d2bdb104e784280af3968346b9446d87.

src/datasources/email/email.datasource.spec.ts
iamacook: I've added a second test in 5674ff88d2bdb104e784280af3968346b9446d87. The two check that updates are "correct" (setting the verified flag to `false`, etc) when the current email was (un-)verified.

src/routes/email/email.controller.update-email.spec.ts
iamacook: I initially thought the previously added test covered this. I've removed the `EmailUpdateError` as it was wrapping not only updates but also email verification errors. Either [throw and cause a `500`](https://github.com/safe-global/safe-client-gateway/blob/5674ff88d2bdb104e784280af3968346b9446d87/src/routes/email/email.controller.update-email.spec.ts#L203) as of 5674ff88d2bdb104e784280af3968346b9446d87.

fmrsabino (approved):
🚀 👏

PR #970, opened 2023-12-21
Check datasources request lifespan before caching its response

### LINKED ISSUE #893: Check request lifespan instead of using a webhook execution delay

hectorgomezv:
Closes #893 

This PR adds an `invalidationTimeMs` when cache keys are deleted. That key will be checked when a response from the data sources (Config Service, Transaction Service) arrives at the CGW, so the response would be only cached if its request was initiated before the associated `invalidationTimeMs`.

Changes: 
- Adds a `_shouldBeCached` function to `CacheFirstDataSource` that determines whether a response data coming from a data source should be cached or not, depending on the timestamp check.
- Writes `invalidationTimeMs` when a `CacheService.deleteByKey` call is invoked on `TransactionApi` or `ConfigApi`. 
- Modifies the related test classes accordingly.

Out of scope:
- On `ConfigApi` there are some cache key deletions by pattern. These are not tackled on this PR (`TODO` annotations were left), as this needs some additional logic. Will be tackled in a separate PR. 
- Remove the `WEB_HOOK_EXECUTION_DELAY_MS` configuration and its related code (will be also done separately as a followup).

### DIFF
--- a/src/datasources/cache/__tests__/fake.cache.service.spec.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.spec.ts
@@ -8,6 +8,7 @@ describe('FakeCacheService', () => {
 
   beforeEach(async () => {
     target = new FakeCacheService();
+    jest.useRealTimers();
   });
 
   it('sets key', async () => {
@@ -23,7 +24,7 @@ describe('FakeCacheService', () => {
     expect(target.keyCount()).toBe(1);
   });
 
-  it('deletes key', async () => {
+  it('deletes key without setting invalidationTimeMs', async () => {
     const key = faker.string.alphanumeric();
     const field = faker.string.alphanumeric();
     const cacheDir = new CacheDir(key, field);
@@ -34,6 +35,24 @@ describe('FakeCacheService', () => {
 
     await expect(target.get(cacheDir)).resolves.toBe(undefined);
     expect(target.keyCount()).toBe(0);
+  });
+
+  it('deletes key and sets invalidationTimeMs', async () => {
+    jest.useFakeTimers();
+    const now = jest.now();
+    const key = faker.string.alphanumeric();
+    const field = faker.string.alphanumeric();
+    const cacheDir = new CacheDir(key, field);
+    const value = faker.string.alphanumeric();
+
+    await target.set(cacheDir, value, 0);
+    await target.deleteByKey(key, true);
+
+    await expect(target.get(cacheDir)).resolves.toBe(undefined);
+    await expect(
+      target.get(new CacheDir(`invalidationTimeMs:${cacheDir.key}`, '')),
+    ).resolves.toBe(now.toString());
+    expect(target.keyCount()).toBe(1);
   });
 
   it('clears keys', async () => {
--- a/src/datasources/cache/__tests__/fake.cache.service.spec.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.spec.ts
@@ -8,7 +8,6 @@ describe('FakeCacheService', () => {
 
   beforeEach(async () => {
     target = new FakeCacheService();
-    jest.useRealTimers();
   });
 
   it('sets key', async () => {
@@ -24,19 +23,6 @@ describe('FakeCacheService', () => {
     expect(target.keyCount()).toBe(1);
   });
 
-  it('deletes key without setting invalidationTimeMs', async () => {
-    const key = faker.string.alphanumeric();
-    const field = faker.string.alphanumeric();
-    const cacheDir = new CacheDir(key, field);
-    const value = faker.string.alphanumeric();
-
-    await target.set(cacheDir, value, 0);
-    await target.deleteByKey(key);
-
-    await expect(target.get(cacheDir)).resolves.toBe(undefined);
-    expect(target.keyCount()).toBe(0);
-  });
-
   it('deletes key and sets invalidationTimeMs', async () => {
     jest.useFakeTimers();
     const now = jest.now();
@@ -46,13 +32,14 @@ describe('FakeCacheService', () => {
     const value = faker.string.alphanumeric();
 
     await target.set(cacheDir, value, 0);
-    await target.deleteByKey(key, true);
+    await target.deleteByKey(key);
 
     await expect(target.get(cacheDir)).resolves.toBe(undefined);
     await expect(
       target.get(new CacheDir(`invalidationTimeMs:${cacheDir.key}`, '')),
     ).resolves.toBe(now.toString());
     expect(target.keyCount()).toBe(1);
+    jest.useRealTimers();
   });
 
   it('clears keys', async () => {
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -21,14 +21,12 @@ export class FakeCacheService implements ICacheService {
     this.cache = {};
   }
 
-  deleteByKey(key: string, setInvalidationTime?: boolean): Promise<number> {
+  deleteByKey(key: string): Promise<number> {
     delete this.cache[key];
-    if (setInvalidationTime) {
-      this.set(
-        new CacheDir(`invalidationTimeMs:${key}`, ''),
-        Date.now().toString(),
-      );
-    }
+    this.set(
+      new CacheDir(`invalidationTimeMs:${key}`, ''),
+      Date.now().toString(),
+    );
     return Promise.resolve(1);
   }
 
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -21,9 +21,9 @@ export class FakeCacheService implements ICacheService {
     this.cache = {};
   }
 
-  deleteByKey(key: string): Promise<number> {
+  async deleteByKey(key: string): Promise<number> {
     delete this.cache[key];
-    this.set(
+    await this.set(
       new CacheDir(`invalidationTimeMs:${key}`, ''),
       Date.now().toString(),
     );
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -21,8 +21,14 @@ export class FakeCacheService implements ICacheService {
     this.cache = {};
   }
 
-  deleteByKey(key: string): Promise<number> {
+  deleteByKey(key: string, setInvalidationTime?: boolean): Promise<number> {
     delete this.cache[key];
+    if (setInvalidationTime) {
+      this.set(
+        new CacheDir(`invalidationTimeMs:${key}`, ''),
+        Date.now().toString(),
+      );
+    }
     return Promise.resolve(1);
   }
 
--- a/src/datasources/cache/cache.first.data.source.spec.ts
+++ b/src/datasources/cache/cache.first.data.source.spec.ts
@@ -74,7 +74,7 @@ describe('CacheFirstDataSource', () => {
     const notFoundExpireTimeSeconds = faker.number.int();
     const data = JSON.parse(fakeJson());
     const invalidationTimeMs = jest.now(); // invalidation happens at this point in time
-    fakeCacheService.set(
+    await fakeCacheService.set(
       new CacheDir(`invalidationTimeMs:${cacheDir.key}`, ''),
       invalidationTimeMs.toString(),
     );
@@ -106,7 +106,7 @@ describe('CacheFirstDataSource', () => {
     const notFoundExpireTimeSeconds = faker.number.int();
     const data = JSON.parse(fakeJson());
     const invalidationTimeMs = jest.now() + 1; // invalidation happens 1 ms after this point in time
-    fakeCacheService.set(
+    await fakeCacheService.set(
       new CacheDir(`invalidationTimeMs:${cacheDir.key}`, ''),
       invalidationTimeMs.toString(),
     );
@@ -129,14 +129,14 @@ describe('CacheFirstDataSource', () => {
     expect(actual).toEqual(data);
     expect(mockNetworkService.get).toHaveBeenCalledTimes(1);
     expect(fakeCacheService.keyCount()).toBe(1); // only invalidation timestamp is cached
-    expect(await fakeCacheService.get(cacheDir)).toEqual(undefined); // item is not cached
+    await expect(fakeCacheService.get(cacheDir)).resolves.toEqual(undefined); // item is not cached
   });
 
   it('should return the cached data without calling the underlying network interface', async () => {
     const cacheDir = new CacheDir(faker.word.sample(), faker.word.sample());
     const notFoundExpireTimeSeconds = faker.number.int();
     const rawJson = fakeJson();
-    fakeCacheService.set(cacheDir, rawJson);
+    await fakeCacheService.set(cacheDir, rawJson);
     mockNetworkService.get.mockImplementation((url) =>
       Promise.reject(`Unexpected request to ${url}`),
     );
--- a/src/datasources/cache/cache.first.data.source.spec.ts
+++ b/src/datasources/cache/cache.first.data.source.spec.ts
@@ -62,8 +62,10 @@ describe('CacheFirstDataSource', () => {
 
     expect(actual).toEqual(data);
     expect(mockNetworkService.get).toHaveBeenCalledTimes(1);
-    expect(fakeCacheService.keyCount()).toBe(1); // only data is cached (as no invalidation happened yet a this point in time)
-    expect(await fakeCacheService.get(cacheDir)).toEqual(JSON.stringify(data));
+    expect(fakeCacheService.keyCount()).toBe(1); // only data is cached (as no invalidation happened yet at this point in time)
+    await expect(fakeCacheService.get(cacheDir)).resolves.toEqual(
+      JSON.stringify(data),
+    );
   });
 
   it('should return the network data and it should cache it if the last invalidation happened before the request was initiated', async () => {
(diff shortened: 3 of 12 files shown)


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -914,5 +925,16 @@ export class TransactionApi implements ITransactionApi {
       messageHash: args.messageHash,
     });
     await this.cacheService.deleteByKey(key);
+    await this._setInvalidationTimeForKey(key);
+  }
+
+  private async _setInvalidationTimeForKey(key: string): Promise<void> {
```
hectorgomezv, line 931: Note: instead of having this function here, we could opt for moving this logic to the `CacheService` itself. That would in fact simplify things, but would also write the `invalidationTimeMs` entry on each deletion, which could be a problem if we want to trigger a deletion and avoid doing that.

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -35,11 +36,78 @@ describe('CacheFirstDataSource', () => {
...
+
+    const actual = await cacheFirstDataSource.get({
+      cacheDir,
+      url: targetUrl,
+      notFoundExpireTimeSeconds,
+    });
+
+    expect(actual).toEqual(data);
+    expect(mockNetworkService.get).toHaveBeenCalledTimes(1);
+    expect(fakeCacheService.keyCount()).toBe(1);
```
iamacook, line 65: Nit: can we add a comment here clarifying that the invalidation key has yet to be added.

src/datasources/transaction-api/transaction-api.service.ts
iamacook: I am in favour of moving this as there is a lot of repetition and setting it is also easily forgotten. We could add a flag argument or a different method for setting the `invalidationTimeMs` alongside deletion. We can discuss this further and implement it later though.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -914,5 +925,16 @@ export class TransactionApi implements ITransactionApi {
...
+    await this.cacheService.set(
+      new CacheDir(`invalidationTimeMs:${key}`, ''),
```
iamacook, line 933: Nit: I'd suggest creating a helper function to get the key as it is used in multiple places.

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: I've added several comments explaining the test expectations in [f895765](https://github.com/safe-global/safe-client-gateway/pull/970/commits/f89576594ef369e007eac0e242330f957c05b3b7), let me know if it's not very understandable, or you want to write it differently 🙂

src/datasources/transaction-api/transaction-api.service.ts
hectorgomezv: I've moved this logic to the `CacheService` in [1d4da4c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/1d4da4c0eabcd9f200cb0e9e813d8535421aabea) so it's now being used in 3 places (besides test classes). 

I don't have a strong opinion about it, I'm fine with creating an `utils.ts` for the data sources layer, extracting the string part to a constant somewhere, or letting it as it is.

src/datasources/transaction-api/transaction-api.service.ts
hectorgomezv: I've moved this logic to the `CacheService` implementations on a single commit ([1d4da4c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/1d4da4c0eabcd9f200cb0e9e813d8535421aabea)). But I have to say I don't have a strong opinion regarding this, there are pros and cons of each option. @fmrsabino what's your take on this? 🙂

src/datasources/cache/redis.cache.service.ts
```diff
@@ -42,9 +51,20 @@ export class RedisCacheService
     return await this.client.hGet(cacheDir.key, cacheDir.field);
   }
 
-  async deleteByKey(key: string): Promise<number> {
+  async deleteByKey(
+    key: string,
+    setInvalidationTime?: boolean,
```
fmrsabino, line 56: In which situations do we call this with `false`?

src/datasources/cache/__tests__/fake.cache.service.spec.ts
```diff
@@ -8,6 +8,7 @@ describe('FakeCacheService', () => {
 
   beforeEach(async () => {
     target = new FakeCacheService();
+    jest.useRealTimers();
```
fmrsabino, line 11: Since this was done because of the change of a single test I would move this to that test wdyt?

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -35,11 +36,78 @@ describe('CacheFirstDataSource', () => {
...
+
+    const actual = await cacheFirstDataSource.get({
+      cacheDir,
+      url: targetUrl,
+      notFoundExpireTimeSeconds,
+    });
+
+    expect(actual).toEqual(data);
+    expect(mockNetworkService.get).toHaveBeenCalledTimes(1);
+    expect(fakeCacheService.keyCount()).toBe(1); // only data is cached (as no invalidation happened yet a this point in time)
```
fmrsabino, line 65:
```suggestion
    expect(fakeCacheService.keyCount()).toBe(1); // only data is cached (as no invalidation happened yet at this point in time)
```

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -35,11 +36,78 @@ describe('CacheFirstDataSource', () => {
...
+    expect(await fakeCacheService.get(cacheDir)).toEqual(JSON.stringify(data));
```
fmrsabino, line 66: nitpick: mostly syntactic sugar that we have been using in other tests.

```suggestion
    await expect(fakeCacheService.get(cacheDir)).resolves.toEqual(
      JSON.stringify(data),
    );
```

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -35,11 +36,78 @@ describe('CacheFirstDataSource', () => {
...
+      notFoundExpireTimeSeconds,
+    });
+
+    expect(actual).toEqual(data);
+    expect(mockNetworkService.get).toHaveBeenCalledTimes(1);
+    expect(fakeCacheService.keyCount()).toBe(2); // both data and invalidation timestamp are cached
+    expect(await fakeCacheService.get(cacheDir)).toEqual(JSON.stringify(data)); // item is cached
+  });
+
+  it("should return the network data but it shouldn't not cache it if the last invalidation happened after the request was initiated", async () => {
```
fmrsabino, line 101:
```suggestion
  it("should return the network data but it should not cache it if the last invalidation happened after the request was initiated", async () => {
```

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -35,11 +36,78 @@ describe('CacheFirstDataSource', () => {
...
+    const targetUrl = faker.internet.url({ appendSlash: false });
+    const cacheDir = new CacheDir(faker.word.sample(), faker.word.sample());
+    const notFoundExpireTimeSeconds = faker.number.int();
+    const data = JSON.parse(fakeJson());
+    const invalidationTimeMs = jest.now() + 1; // invalidation happens 1 ms after this point in time
+    fakeCacheService.set(
```
fmrsabino, line 107: We are missing an `await` here (similarly in other tests)

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -24,15 +25,31 @@ const mockLoggingService = {
...
+      }
+      throw Error(`Unexpected key: ${key}`);
+    });
+
     redisCacheService = new RedisCacheService(
       redisClientTypeMock,
       mockLoggingService,
+      mockConfigurationService,
     );
+    jest.useRealTimers();
```
fmrsabino, line 52: We set the fake timers only in one test so I'd move this to that test for now.

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -5,6 +5,7 @@ import { RedisCacheService } from '@/datasources/cache/redis.cache.service';
 import { RedisClientType } from 'redis';
 import clearAllMocks = jest.clearAllMocks;
 import { fakeJson } from '@/__tests__/faker';
+import { IConfigurationService } from '@/config/configuration.service.interface';
 
 const redisClientType = {
```
fmrsabino, line 10: Out of scope for this PR but we should probably consider converting this to an integration test with Redis itself as we have integration tests at this level with the Postgres database for example.

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -153,7 +170,7 @@ describe('RedisCacheService', () => {
     expect(redisClientTypeMock.quit).toHaveBeenCalledTimes(0);
   });
 
-  it('Deleting key calls delete', async () => {
+  it("Deleting key calls delete and don't set invalidationTime", async () => {
```
fmrsabino, line 173: I understand the need for this test but this situation only happens because the interface that is offered allows for disabling this behaviour (although actual usage seems to be limited to having the `invalidationTime` always registered).

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -165,6 +182,25 @@ describe('RedisCacheService', () => {
...
+    expect(redisClientTypeMock.hGet).toHaveBeenCalledTimes(0);
+    expect(redisClientTypeMock.hSet).toHaveBeenCalledTimes(1);
+    expect(redisClientTypeMock.hSet).toHaveBeenCalledWith(
+      `invalidationTimeMs:${key}`,
+      '',
+      now.toString(),
+    );
+    expect(redisClientTypeMock.hDel).toHaveBeenCalledTimes(0);
+    expect(redisClientTypeMock.quit).toHaveBeenCalledTimes(0);
+  });
```
fmrsabino, line 202: We can also add an assertion which verifies that the key was set with a given TTL.

fmrsabino (changes requested):

src/datasources/transaction-api/transaction-api.service.ts
fmrsabino: If the end result is that we create a new key with the invalidation time then I would say that this should be ok and have it created all the time to simplify things.

I'd argue that the value on having this key every time outweighs the branching logic that we would need to have on each feature that we want to cache (as it stands right now of course. This can change in the future).

This is also related with my comment [here](https://github.com/safe-global/safe-client-gateway/pull/970#discussion_r1435047173)

src/datasources/cache/redis.cache.service.ts
hectorgomezv: I was thinking of giving the client of this class the ability to indicate if the `invalidationTimeMs` should be written. But maybe this can be simplified as there's no situation where this will be called with `false` value. 

I'm going to remove the optional parameter if you agree, so the calls are also simplified.

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: `FakeCacheService` functions are synchronous (unlike async implementations like `RedisCacheService`).

`FakeCacheService` functions are called with no `await` in this file as you said, `health.controller.spec.ts` or `flush.controller.spec.ts`. I think it makes no difference, but we could agree on a common approach and unify all the calls in a separate PR, wdyt?

src/datasources/cache/redis.cache.service.ts
fmrsabino: I think we can assume that we will generate this entry every time. We can iterate on it if things change 👍

src/datasources/cache/redis.cache.service.spec.ts
hectorgomezv: I agree, it's a good idea for a follow-up task, happy to tackle it separately 👍

src/datasources/cache/redis.cache.service.ts
hectorgomezv: Perfect, thanks, changed in [b92584c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/b92584cf5c50470b34305c8c9afffcc884c9fb21)

src/datasources/cache/__tests__/fake.cache.service.spec.ts
hectorgomezv: Done in [b92584c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/b92584cf5c50470b34305c8c9afffcc884c9fb21)

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: Agree with your suggestion but when changing `shouldn't` to `should not` we can use single quotes. Change: [b92584c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/b92584cf5c50470b34305c8c9afffcc884c9fb21)

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: You are right, I'm going to add the `await` statement here, and  will create a separate PR to fix it in other classes 👍

src/datasources/cache/redis.cache.service.spec.ts
hectorgomezv: Agreed, changed in [b92584c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/b92584cf5c50470b34305c8c9afffcc884c9fb21)

src/datasources/cache/redis.cache.service.spec.ts
hectorgomezv: Removed in [b92584c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/b92584cf5c50470b34305c8c9afffcc884c9fb21)

src/datasources/cache/redis.cache.service.spec.ts
hectorgomezv: Checks added in [b92584c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/b92584cf5c50470b34305c8c9afffcc884c9fb21)

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: Done in [e4d030f](https://github.com/safe-global/safe-client-gateway/pull/970/commits/e4d030ffdc0a86bf23d2170be3f200856769c27e)

src/datasources/cache/cache.first.data.source.spec.ts
hectorgomezv: (And [01321e7](https://github.com/safe-global/safe-client-gateway/pull/970/commits/01321e7b46fdf3e8914f83348cdb94393babaff1) 😅)

src/datasources/config-api/config-api.service.ts
```diff
@@ -59,9 +59,11 @@ export class ConfigApi implements IConfigApi {
 
   async clearChains(): Promise<void> {
     const pattern = CacheRouter.getChainsCachePattern();
+    const key = CacheRouter.getChainsCacheKey();
     await Promise.all([
-      this.cacheService.deleteByKey(CacheRouter.getChainsCacheKey()),
+      this.cacheService.deleteByKey(key),
       this.cacheService.deleteByKeyPattern(pattern),
+      // TODO: call _setInvalidationTimeForKey for each item matching the pattern
```
fmrsabino, line 66: This function needs to be reviewed as the `flush` endpoint shouldn't be used anymore anyway.

src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -133,6 +134,14 @@ describe('TransactionApi', () => {
   });
 
   describe('Clear Local Balances', () => {
+    beforeEach(() => {
+      jest.useFakeTimers();
+    });
+
+    afterAll(() => {
+      jest.useRealTimers();
+    });
```
fmrsabino, line 143: This shouldn't be required.

fmrsabino (changes requested):

src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -23,6 +23,10 @@ export class FakeCacheService implements ICacheService {
 
   deleteByKey(key: string): Promise<number> {
     delete this.cache[key];
+    this.set(
```
fmrsabino, line 26: We are missing an `await` here.  `set` does mostly sync work but in the end returns an empty promise. However, this function shouldn't be required to know that most of the work is done synchronously.

fmrsabino (changes requested):

src/datasources/config-api/config-api.service.ts
hectorgomezv: Yes, I agree this function shouldn't be necessary once the `flush` endpoint is deprecated. We could review it in a separate issue if you agree 🙂

src/datasources/cache/__tests__/fake.cache.service.ts
hectorgomezv: Changed in [11b5c6c](https://github.com/safe-global/safe-client-gateway/pull/970/commits/11b5c6c32aecfed991a309d512ca0a9fce3b3346)

(To be able to use an `await` statement here, `deleteByKey` has to be marked as `async` also)

fmrsabino (approved):
Great work! 👏 🚀

PR #973, opened 2023-12-22
Set Jest timezone to UTC

iamacook:
This sets the timezone of Jest to UTC via a [`globalSetup`](https://jestjs.io/docs/configuration#globalsetup-string) file. It cannot be set via our `e2e-setup` file as it needs run once before everything.


fmrsabino (approved):
🚀

PR #980, opened 2023-12-26
Add DELETED_MULTISIG_TRANSACTION event

hectorgomezv:
Changes: 
- Adds `DELETED_MULTISIG_TRANSACTION` event to the events supported by the CGW. 

Out of scope:
- `DELETED_MULTISIG_TRANSACTION` event handling: it will be implemented in a separate PR.

PR #981, opened 2023-12-26
Handle DeletedMultisigTransaction event type

hectorgomezv:
Depends on #980 

Changes:
- Adds the handling logic for a `DeletedMultisigTransaction` event (i.e.: the service will call `clearMultisigTransactions` and `clearMultisigTransaction` when this type of event is received).


src/routes/cache-hooks/cache-hooks.service.ts
```diff
@@ -44,6 +45,7 @@ export class CacheHooksService {
   async onEvent(
```
fmrsabino, line 45: We are missing the type on `cache-hooks.controller.ts`. We should consider adding a union type given the amount of types involved here 😅

fmrsabino (approved):

src/routes/cache-hooks/cache-hooks.service.ts
hectorgomezv: Makes total sense to me, I'll add the union type in a separate PR if you agree!

(I've added the new type to `CacheHooksController` in [27f4dee](https://github.com/safe-global/safe-client-gateway/pull/981/commits/27f4dee496e4a740c986c8a6b0887501a324e6ff), thanks! 🙂)

PR #985, opened 2023-12-28
RedisCacheService tests refactor

hectorgomezv:
Context: https://github.com/safe-global/safe-client-gateway/pull/970#discussion_r1435060366

Changes:
- Modifies `redis.cache.service.spec.ts` to use a real Redis test instance instead of a mocked class.

### DIFF
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -135,8 +135,8 @@ export default (): ReturnType<typeof configuration> => ({
     },
   },
   redis: {
-    host: faker.internet.domainName(),
-    port: faker.internet.port().toString(),
+    host: process.env.REDIS_HOST || 'localhost',
+    port: process.env.REDIS_PORT || '6379',
   },
   relay: { limit: faker.number.int({ min: 1 }) },
   safeConfig: {
--- a/src/datasources/cache/redis.cache.service.spec.ts
+++ b/src/datasources/cache/redis.cache.service.spec.ts
@@ -30,6 +30,10 @@ describe('RedisCacheService', () => {
     redisClient = await redisClientFactory();
   });
 
+  afterAll(async () => {
+    await redisClient.quit();
+  });
+
   beforeEach(async () => {
     clearAllMocks();
     defaultExpirationTimeInSeconds = faker.number.int();
@@ -161,10 +165,5 @@ describe('RedisCacheService', () => {
     await expect(redisCacheService.ping()).rejects.toThrow();
     // Connection is reopened after this test execution
     redisClient = await redisClientFactory();
-  });
-
-  afterAll(async () => {
-    await redisClient.flushAll();
-    await redisClient.quit();
   });
 });
--- a/src/datasources/cache/redis.cache.service.spec.ts
+++ b/src/datasources/cache/redis.cache.service.spec.ts
@@ -6,6 +6,7 @@ import { RedisClientType } from 'redis';
 import { fakeJson } from '@/__tests__/faker';
 import { IConfigurationService } from '@/config/configuration.service.interface';
 import clearAllMocks = jest.clearAllMocks;
+import { redisClientFactory } from '@/__tests__/redis-client.factory';
 
 const redisClientType = {
   hGet: jest.fn(),
@@ -34,6 +35,11 @@ describe('RedisCacheService', () => {
   let redisCacheService: RedisCacheService;
   let defaultExpirationTimeInSeconds: number;
   const keyPrefix = '';
+  let redisClient: RedisClientType;
+
+  beforeAll(async () => {
+    redisClient = await redisClientFactory();
+  });
 
   beforeEach(async () => {
     clearAllMocks();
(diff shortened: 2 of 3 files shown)


src/config/entities/__tests__/configuration.ts
```diff
@@ -135,8 +135,8 @@ export default (): ReturnType<typeof configuration> => ({
     },
   },
   redis: {
-    host: faker.internet.domainName(),
-    port: faker.internet.port().toString(),
+    host: process.env.REDIS_HOST || 'localhost',
+    port: process.env.REDIS_PORT || '6379',
```
fmrsabino, line 139: The description of the PR mentions a test instance but if we follow the setup that we have with `docker-compose` then the development instance will be used.

While we use Redis as a key-value store to temporarily store data, if you run the tests after a debug session for example, you might get the wrong data (and assertions might fail).

fmrsabino (changes requested):

src/config/entities/__tests__/configuration.ts
fmrsabino: This is in part now mitigated with https://github.com/safe-global/safe-client-gateway/pull/998

iamacook (approved):

src/config/entities/__tests__/configuration.ts
hectorgomezv: Good point, I also thought about this, and I was thinking of using a separate Redis _database_ (logic data separation) between test data and _development_ data. I decided against doing that in this PR to avoid introducing too much noise, but we can easily set `database: 1` when running tests (if not specified, all the data is stored in database 0 in Redis). Context: https://redis.io/commands/select/

What do you think about this solution? It wouldn't overlap with https://github.com/safe-global/safe-client-gateway/pull/998, as running tests in a separate database doesn't help with parallel-running test collisions.

fmrsabino: I'm more in favour of a separate Redis instance for the tests. The database selection would require more changes on the production code to have a default one selected.

One advantage of having a separate instance is also performance – running the tests while running a development build would not affect the underlying performance (even with different databases).

Although this is not as critical as a separate database for testing due to the nature of the data that we store in Redis.

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -77,147 +69,100 @@ describe('RedisCacheService', () => {
...
-    expect(redisClientTypeMock.quit).toHaveBeenCalledTimes(1);
+    // Connection is closed, this is expected to throw an error
+    await expect(redisCacheService.ping()).rejects.toThrow();
+    // Connection is reopened after this test execution
+    redisClient = await redisClientFactory();
+  });
+
+  afterAll(async () => {
+    await redisClient.flushAll();
+    await redisClient.quit();
```
fmrsabino, line 166: I'd move this to before the tests so that it's easier to compare the setup/teardown.

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -77,147 +69,100 @@ describe('RedisCacheService', () => {
...
-    expect(redisClientTypeMock.hDel).toHaveBeenCalledTimes(0);
-    expect(redisClientTypeMock.quit).toHaveBeenCalledTimes(1);
+    // Connection is closed, this is expected to throw an error
+    await expect(redisCacheService.ping()).rejects.toThrow();
+    // Connection is reopened after this test execution
+    redisClient = await redisClientFactory();
+  });
+
+  afterAll(async () => {
+    await redisClient.flushAll();
```
fmrsabino, line 165: This can affect other tests. For now the prefix avoid conflicts in the key-space but if we trigger a `FLUSHALL` on the instance then we might affect the state of the tests that depend on the same instance.

fmrsabino (changes requested):

src/config/entities/__tests__/configuration.ts
hectorgomezv: > The database selection would require more changes on the production code to have a default one selected.
> 
> One advantage of having a separate instance is also performance – running the tests while running a development build would not affect the underlying performance (even with different databases).
> 

I totally agree with the performance part, even if it's not critical I'd say. Regarding the changes to production code, I think the tests involving Redis are using a test  `RedisClient`, provided by `redisClientFactory` (under the `__tests__` folder), so the database selection (via parameter, or even hardcoding the database number) would affect test classes only if I'm not wrong.

But yeah, overall I also think that having a separate instance makes more sense, and it's more consistent with the approach we took for the Postgres database also 👍🏻

src/datasources/cache/redis.cache.service.spec.ts
hectorgomezv: I agree, thanks! Changed in [e07b503](https://github.com/safe-global/safe-client-gateway/pull/985/commits/e07b50302262b71c265cb32598e51825d4809346)

fmrsabino (approved):
👏 🚀

fmrsabino (approved):

PR #994, opened 2024-01-04
Trigger email deletion from provider

fmrsabino:
Triggers the email address deletion on the provider side when the `DELETE chains/:chainId/safes/:safeAddress/emails` is called.

If the provider returns an error, the email is not deleted from our database.

### DIFF
--- a/src/domain/email/email.repository.ts
+++ b/src/domain/email/email.repository.ts
@@ -9,6 +9,7 @@ import { IConfigurationService } from '@/config/configuration.service.interface'
 import { EmailAlreadyVerifiedError } from '@/domain/email/errors/email-already-verified.error';
 import { InvalidVerificationCodeError } from '@/domain/email/errors/invalid-verification-code.error';
 import { EmailUpdateMatchesError } from '@/domain/email/errors/email-update-matches.error';
+import { IEmailApi } from '@/domain/interfaces/email-api.interface';
 
 @Injectable()
 export class EmailRepository implements IEmailRepository {
@@ -20,6 +21,7 @@ export class EmailRepository implements IEmailRepository {
     private readonly emailDataSource: IEmailDataSource,
     @Inject(IConfigurationService)
     private readonly configurationService: IConfigurationService,
+    @Inject(IEmailApi) private readonly emailApi: IEmailApi,
   ) {
     this.verificationCodeResendLockWindowMs =
       this.configurationService.getOrThrow(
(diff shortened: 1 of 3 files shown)


src/domain/email/email.repository.ts
```diff
@@ -147,7 +149,11 @@ export class EmailRepository implements IEmailRepository {
     safeAddress: string;
     account: string;
   }): Promise<void> {
-    return this.emailDataSource.deleteEmail(args);
+    const email = await this.emailDataSource.getEmail(args);
+    await this.emailApi.deleteEmailAddress({
+      emailAddress: email.emailAddress.value,
+    });
```
fmrsabino, line 155: The errors that can be returned when deleting the email are either non-existent or not-documented – https://docs.pushwoosh.com/platform-docs/api-reference/email-api#deleteemail

This is something that we need to further clarify as if we get an error that an email was already deleted on the provider side then we should not throw an error, and continue with the deletion on our side instead.

src/routes/email/email.controller.delete-email.spec.ts
```diff
@@ -97,6 +112,9 @@ describe('Email controller delete email tests', () => {
       })
       .expect(204)
       .expect({});
+
+    expect(emailApi.deleteEmailAddress).toBeCalledTimes(1);
```
hectorgomezv, line 116: Nitpick: according to Jest docs, `toBeCalledTimes` is deprecated in favor of `toHaveBeenCalledTimes` 🙂

src/routes/email/email.controller.delete-email.spec.ts
```diff
@@ -245,4 +266,51 @@ describe('Email controller delete email tests', () => {
...
+          return Promise.resolve({ data: chain });
+        case `${chain.transactionService}/api/v1/safes/${safe.address}`:
+          return Promise.resolve({ data: safe });
+        default:
+          return Promise.reject(new Error(`Could not match ${url}`));
+      }
+    });
+    emailDatasource.getEmail.mockResolvedValueOnce(email);
+    emailApi.deleteEmailAddress.mockRejectedValue(new Error('Some error'));
+    emailDatasource.deleteEmail.mockResolvedValueOnce();
```
hectorgomezv, line 301: Nitpick: I think this call to `mockResolvedValueOnce` is not needed as the function `emailDatasource.deleteEmail` is not expected to be called.

hectorgomezv (changes requested):
Great job 👏🏻👏🏻  I've only left two nitpick comments 🙂

src/routes/email/email.controller.delete-email.spec.ts
fmrsabino: Good catch! [034c228](https://github.com/safe-global/safe-client-gateway/pull/994/commits/034c2286ce2058625a214fc4b44e06d40f3be97b)

src/routes/email/email.controller.delete-email.spec.ts
fmrsabino: Agreed 👍 [034c228](https://github.com/safe-global/safe-client-gateway/pull/994/commits/034c2286ce2058625a214fc4b44e06d40f3be97b)

hectorgomezv (approved):
👏🏻🚀

PR #998, opened 2024-01-05
Use cache key prefix

fmrsabino:
- Provides a `CacheKeyPrefix` in `CacheModule` – this prefix will be prepended to the cache key being set/read.
- The prefix when running the app (outside a test context) is an empty string.
- Tests that require a Redis instance can now be run in parallel given that they can set a random prefix before running the test suite (e.g.: with `crypto.randomUUID`)

Local testing showed an improvement of around 50% when running the tests concurrently. Note however that E2E tests are bound to networking conditions.

### DIFF
--- a/src/datasources/cache/redis.cache.service.spec.ts
+++ b/src/datasources/cache/redis.cache.service.spec.ts
@@ -3,9 +3,9 @@ import { ILoggingService } from '@/logging/logging.interface';
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 import { RedisCacheService } from '@/datasources/cache/redis.cache.service';
 import { RedisClientType } from 'redis';
-import clearAllMocks = jest.clearAllMocks;
 import { fakeJson } from '@/__tests__/faker';
 import { IConfigurationService } from '@/config/configuration.service.interface';
+import clearAllMocks = jest.clearAllMocks;
 
 const redisClientType = {
   hGet: jest.fn(),
@@ -33,6 +33,7 @@ const mockConfigurationService = jest.mocked(configurationService);
 describe('RedisCacheService', () => {
   let redisCacheService: RedisCacheService;
   let defaultExpirationTimeInSeconds: number;
+  const keyPrefix = '';
 
   beforeEach(async () => {
     clearAllMocks();
@@ -48,6 +49,7 @@ describe('RedisCacheService', () => {
       redisClientTypeMock,
       mockLoggingService,
       mockConfigurationService,
+      keyPrefix,
     );
   });
 
--- a/src/datasources/cache/redis.cache.service.ts
+++ b/src/datasources/cache/redis.cache.service.ts
@@ -5,6 +5,7 @@ import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 import { ICacheReadiness } from '@/domain/interfaces/cache-readiness.interface';
 import { ILoggingService, LoggingService } from '@/logging/logging.interface';
 import { IConfigurationService } from '@/config/configuration.service.interface';
+import { CacheKeyPrefix } from '@/datasources/cache/constants';
 
 @Injectable()
 export class RedisCacheService
@@ -18,6 +19,7 @@ export class RedisCacheService
     @Inject(LoggingService) private readonly loggingService: ILoggingService,
     @Inject(IConfigurationService)
     private readonly configurationService: IConfigurationService,
+    @Inject(CacheKeyPrefix) private readonly keyPrefix: string,
   ) {
     this.defaultExpirationTimeInSeconds =
       this.configurationService.getOrThrow<number>(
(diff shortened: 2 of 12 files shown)


src/routes/about/__tests__/get-about.e2e-spec.ts
```diff
@@ -4,19 +4,28 @@ import * as request from 'supertest';
...
+      .useValue(cacheKeyPrefix)
+      .compile();
 
     app = moduleRef.createNestApplication();
     await app.init();
   });
 
+  afterAll(async () => {
+    await app.close();
+  });
```
fmrsabino, line 27: I've moved the `afterAll` to the section before each test. In my opinion this makes it easier to see what is done before and after each test (as we can have quite a few tests before we get to the `afterEach`, `afterAll`)

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -33,6 +33,7 @@ const mockConfigurationService = jest.mocked(configurationService);
 describe('RedisCacheService', () => {
   let redisCacheService: RedisCacheService;
   let defaultExpirationTimeInSeconds: number;
+  const keyPrefix = '';
```
iamacook, line 36: Is there a reason we don't want to test with the prefix here?

src/datasources/cache/redis.cache.service.ts
```diff
@@ -38,32 +40,39 @@ export class RedisCacheService
...
+      new CacheDir(`invalidationTimeMs:${keyWithPrefix}`, ''),
       Date.now().toString(),
       this.defaultExpirationTimeInSeconds,
     );
     return result;
   }
 
   async deleteByKeyPattern(pattern: string): Promise<void> {
-    for await (const key of this.client.scanIterator({ MATCH: pattern })) {
+    const patternWithPrefix = `${this.keyPrefix}${pattern}`;
```
iamacook, line 72: Nit: we could create a helper to prefix keys. It should work in all of these methods.

iamacook: I think it would also be beneficial to have a separator between the prefix and key for clarity. At the moment it's not clear where the prefix ends and the chain ID starts.

src/datasources/cache/redis.cache.service.ts
```diff
@@ -38,32 +40,39 @@ export class RedisCacheService
...
   }
 
   async deleteByKey(key: string): Promise<number> {
+    const keyWithPrefix = `${this.keyPrefix}${key}`;
     // see https://redis.io/commands/unlink/
-    const result = await this.client.unlink(key);
+    const result = await this.client.unlink(keyWithPrefix);
     await this.set(
-      new CacheDir(`invalidationTimeMs:${key}`, ''),
+      new CacheDir(`invalidationTimeMs:${keyWithPrefix}`, ''),
```
fmrsabino, line 64: There's a bug in this implementation where the prefix will be added again because we are calling `set` here which also adds the prefix.

src/datasources/cache/redis.cache.service.ts
```diff
@@ -18,6 +19,7 @@ export class RedisCacheService
     @Inject(LoggingService) private readonly loggingService: ILoggingService,
     @Inject(IConfigurationService)
     private readonly configurationService: IConfigurationService,
+    @Inject(CacheKeyPrefix) private readonly keyPrefix: string,
```
hectorgomezv, line 22: Nitpick: What do you think about adding some documentation/comments regarding the purpose of this prefix, its intended usage, or the reason why it is needed? I'm not sure this is the place for it, may be somewhere else. I'm considering that this `prefix` may be not self-explanatory when reading the code.

fmrsabino: I agree. Added in [20dc54f](https://github.com/safe-global/safe-client-gateway/pull/998/commits/20dc54f91051f3a9afceaaffdaa9cdf19f91a11a)

src/datasources/cache/redis.cache.service.ts
fmrsabino: Fixed in [20dc54f](https://github.com/safe-global/safe-client-gateway/pull/998/commits/20dc54f91051f3a9afceaaffdaa9cdf19f91a11a)

src/datasources/cache/redis.cache.service.spec.ts
fmrsabino: I've decided to create a new set of tests to add the prefix: [20dc54f](https://github.com/safe-global/safe-client-gateway/pull/998/commits/20dc54f91051f3a9afceaaffdaa9cdf19f91a11a)

The main reason is because I didn't want to change the validation tests that we have currently to handle the prefix since the application will run without one.

The tests introduced in `redis.cache.service.key-prefix.spec.ts` should cover this though!

hectorgomezv (approved):
Awesome job! 🚀

iamacook (approved):

PR #1006, opened 2024-01-09
Return UTC timestamps

fmrsabino:
- The transaction history endpoint now returns timestamps without shifting those timestamps with the `timezone_offset`
- This means that the `timezone_offset` will be only used for grouping.
- Removes the option to set a `timezone_offset` for the following endpoints:
  * `GET chains/:chainId/safes/:safeAddress/multisig-transactions`
  * `GET chains/:chainId/safes/:safeAddress/module-transactions`
  * `GET chains/:chainId/safes/:safeAddress/incoming-transfers`
  * `GET chains/:chainId/safes/:safeAddress/transactions/queued`
  * `POST chains/:chainId/transactions/:safeAddress/propose`
- Removes respective tests.


iamacook (approved):
LGTM!

~I have a suggestion for a later PR - we should consider centralised timestamp retrieval logic.

In transaction specific mappers we are directly accessing properties, e.g. `executionDate`, `submissionDate`, etc. whereas we have a `getTransactionTimestamp` helper in `transaction-history-mapper` that could do this for us.

It would be more consistent (and safer) for us to use the latter everywhere. What do you think?~

The above didn't take into account the differences between queue and history.

src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -156,18 +155,19 @@ export class TransactionsHistoryMapper {
   ): TransactionDomainGroup[] {
     return Object.entries(
       groupBy(transactions, (transaction) => {
-        return this.getDayFromTransactionDate(
-          transaction,
-          timezoneOffset,
-        ).getTime();
+        return [
```
hectorgomezv, line 158: Nit: I think returning an array here is not necessary for the grouping, right?
```javascript
groupBy(transactions, (transaction) =>
  this.getDayFromTransactionDate(transaction, timezoneOffset).getTime(),
)
```

hectorgomezv (approved):
Great job! 👏🏻👏🏻

src/routes/transactions/mappers/transactions-history.mapper.ts
fmrsabino: It is not. Good catch.

I was trying an alternative solution where the return would be a tuple that would include the UTC timestamp (to avoid indexing the first transaction in the followup `map`).

fmrsabino: Fixed in [918d1b3](https://github.com/safe-global/safe-client-gateway/pull/1006/commits/918d1b34c189ff1972da1d8a6589f97d08b79148)

hectorgomezv (approved):

PR #1007, opened 2024-01-10
Update (unknown) recovery email subject and substituions

iamacook:
This updates the `subject` and `substitutions` provided when sending (unknown) recovery emails.

The subject of unknown recovery transaction emails have been aligned with the client implementation:

- "Unknown transaction attempt" -> "Malicious transaction"

A new `IEmailTemplate` service has been created, namely a provider specific `PushwoodTemplate` for creating substitute-specific HTML elements.

(Unknown) recovery attempts now both receive the following:

- `chainId`
- `safeAddress` - pre-formatted as our provider does not allow JavaScript. We want to colour the centre of the address.

Recovery attempts also have the following for the new Safe setup:

- `owners` - pre-formatted as our provider does not allow mapping over arrays in templates, as well as formatting of the addresses explained above for `safeAddress`.
- `threshold`

Note: the above is subject to change after consultation with design and I am also in contact with out provider regarding the mapping of `owners`.

### DIFF
--- a/src/datasources/email-template/pushwood-template.service.spec.ts
+++ b/src/datasources/email-template/pushwood-template.service.spec.ts
@@ -1,0 +1,46 @@
+import { PushwooshTemplate } from '@/datasources/email-template/pushwoosh-template.service';
+import { faker } from '@faker-js/faker';
+import { getAddress } from 'viem';
+
+describe('PushwoodTemplate', () => {
+  const service = new PushwooshTemplate();
+
+  describe('addressToHtml', () => {
+    it('should return an element with a checksummed address', () => {
+      const address = faker.finance.ethereumAddress();
+      const checksummedAddress = getAddress(address);
+
+      const start = checksummedAddress.slice(0, 6);
+      const center = checksummedAddress.slice(6, -4);
+      const end = checksummedAddress.slice(-4);
+
+      const expected = `${start}<span id="address-center">${center}</span>${end}`;
+
+      expect(service.addressToHtml(address)).toEqual(expected);
+    });
+  });
+
+  describe('addressListToHtml', () => {
+    it('should return a list element with checksummed addresses', () => {
+      const addresses = [
+        faker.finance.ethereumAddress(),
+        faker.finance.ethereumAddress(),
+      ];
+      const checksummedAddresses = addresses.map((address) => {
+        return getAddress(address);
+      });
+
+      const expected = `<ul>${checksummedAddresses
+        .map((checmsummedAddress) => {
+          const start = checmsummedAddress.slice(0, 6);
+          const center = checmsummedAddress.slice(6, -4);
+          const end = checmsummedAddress.slice(-4);
+
+          return `<li>${start}<span id="address-center">${center}</span>${end}</li>`;
+        })
+        .join('')}</ul>`;
+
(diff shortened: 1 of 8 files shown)


src/domain/alerts/alerts.repository.ts
```diff
@@ -183,19 +192,43 @@ export class AlertsRepository implements IAlertsRepository {
...
+  }
+
+  /**
+   * Checksums and formats the address to be displayed in the email as
+   * PushWoosh doesn't support JavaScript
+   *
+   * 0x1234...7890 => 0x1234<span id="address-center">.../span>7890
+   * @param address - address to be formatted
+   */
+  private _formatAddress(address: string): string {
```
hectorgomezv, line 223: I'd say the `domain` layer is the place to include provider-specific (PushWoosh in this case) logic.

Maybe we could left it here in a first iteration, but in my opinion it would make more sense to have something like a provider-specific templates service (something like `pushwoosh-templates.service.ts` in this case) in the `datasources` layer, following a common service interface that could contain functions like `singleAddressToHtml` and `addressListToHtml` for example, so in the future another emailing provider would need to implement that interface, and the `domain` layer would stay provider-agnostic.

As I said, maybe could be seen as a follow-up instead of a matter for this PR. Thoughts? @fmrsabino @iamacook 🙂

hectorgomezv:
Great job! Besides the concern I wrote [here](https://github.com/safe-global/safe-client-gateway/pull/1007#discussion_r1447247163), it looks good to me!

src/domain/alerts/alerts.repository.ts
iamacook: Good point and I agree. I will create the appropriate service now.

iamacook: I adjusted the structure in 553dd9c8dc0351d651beeecc5de4cac0c096e549 and updated test coverage in 2df8bbc0d64e06431628a953c9a5f18d42bb19dd..

src/datasources/email-template/pushwood-template.service.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+      const addresses = [
+        faker.finance.ethereumAddress(),
+        faker.finance.ethereumAddress(),
+      ];
+      const checksummedAddresses = addresses.map((address) => {
+        return getAddress(address);
+      });
+
+      const expected = `<ul>${checksummedAddresses
+        .map((checmsummedAddress) => {
```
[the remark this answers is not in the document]
iamacook: Fixed in 014c3e3052ebdbd8501e875533347023dcab4d3a.

src/domain/interfaces/email-template.interface.ts
```diff
@@ -0,0 +1,19 @@
+export const IEmailTemplate = Symbol('IEmailTemplate');
+
+/**
+ * Templates are required (but used sparingly) as PushWoosh doesn't support
+ * JavaScript or arrays passed as substitution variables
```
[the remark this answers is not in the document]
iamacook: Changed to your suggestion in 014c3e3052ebdbd8501e875533347023dcab4d3a.

hectorgomezv (approved):

PR #1009, opened 2024-01-10
Add web app link substitution and add explorer URL to address HTML

iamacook:
This updates the email template `substitutions` for (unknown) recovery transactions:

- The `chainId` has been changed to a pre-generated `webAppUrl` as the provider has no access to the `shortName` of the chain.
- All addresses now link to the relevant block explorer with an icon.

The `owners` substitution has been made more declarative and renamed to `ownersListHtml`, and the `safeAddress` removed as [the design](https://www.figma.com/file/pO43pjVwR34Uy1ihMsY3OM/Recovery-Hub?type=design&node-id=1402-25942&mode=design&t=icygDNmHToLh0xlX-0) no longer requires it.


src/datasources/email-template/pushwoosh-template.service.ts
```diff
@@ -1,24 +1,75 @@
...
+import { Inject, Injectable } from '@nestjs/common';
 import { getAddress } from 'viem';
 import { IEmailTemplate } from '@/domain/interfaces/email-template.interface';
+import { IChainsRepository } from '@/domain/chains/chains.repository.interface';
 
 @Injectable()
 export class PushWooshTemplate implements IEmailTemplate {
-  addressToHtml(address: string): string {
+  constructor(
+    @Inject(IChainsRepository)
```
hectorgomezv, line 9: I think we don't have an hard rule for this, but I'd try to not inject a Repository at this level (datasource layer). 

What do you think about retrieving the `Chain` object and passing it to the methods in `IEmailTemplate` (i.e.: passing `chain: Chain` instead of `chainId:  string`)? That would also have other performance-wise beneficial side-effects, because `AlertsRepository._notifySafeSetup` for example calls the three functions of `IEmailTemplate`, and with the current implementation it would translate into three calls to the repository, which could be just one by retrieving the `Chain` beforehand.

iamacook: I've adjusted the code to pass the `Chain` as an argument in f08a727f880da61260c379da75737d62ecca97c0. As well as the performance improvement, it simplified some of the code as well.

hectorgomezv (approved):
🚀🚀

src/datasources/email-template/pushwoosh-template.service.ts
```diff
@@ -1,24 +1,71 @@
...
 
-    return `${start}<span id="${CSS_ID}">${center}</span>${end}`;
+    return `<span>${start}<span id="${CSS_ID}">${center}</span>${end}</span>`;
+  }
+
+  private addressToExplorerLink(args: {
+    chain: Chain;
+    address: string;
+  }): string {
+    const ADDRESS_TEMPLATE = '{{address}}';
```
hectorgomezv, line 49: Nit: what about removing this variable and using the value string inline? It's only used once in the file. Alternatively, we could make it static if you prefer to keep the variable reference.

src/datasources/email-template/pushwoosh-template.service.ts
```diff
@@ -1,24 +1,71 @@
...
+
+    const addressedExplorerUrl =
+      args.chain.blockExplorerUriTemplate.address.replace(
+        ADDRESS_TEMPLATE,
+        args.address,
+      );
+    const explorerName = new URL(addressedExplorerUrl).hostname;
+
+    const base64LinkIcon = this.imageToBase64(
```
hectorgomezv, line 58: With the current implementation `readFileSync` is called on each `addressToExplorerLink` execution, and I think that could be optimized easily by loading the image into memory on the class initialisation phase (class constructor or static reference).

src/datasources/email-template/pushwoosh-template.service.ts
```diff
@@ -1,24 +1,71 @@
...
 import { IEmailTemplate } from '@/domain/interfaces/email-template.interface';
+import { Chain } from '@/routes/chains/entities/chain.entity';
+import { readFileSync } from 'fs';
+import { join } from 'path';
 
 @Injectable()
 export class PushWooshTemplate implements IEmailTemplate {
-  addressToHtml(address: string): string {
+  addressToSafeWebAppUrl(args: { chain: Chain; safeAddress: string }): string {
+    return `https://app.safe.global/home?safe=${args.chain.shortName}:${args.safeAddress}`;
```
hectorgomezv, line 11: Does it make sense to set this base url via configuration? I mean, could it be a scenario where the Safe Web App lives on a different base url? (staging/development environments for example)

hectorgomezv (changes requested):

src/datasources/email-template/pushwoosh-template.service.ts
iamacook: This was moved in a40eb1ca59a440ffd68a3a4a1316dc59cc05fe0f.

src/datasources/email-template/pushwoosh-template.service.ts
iamacook: This no longer exists. It turns out that Gmail doesn't support base64 images so I opted for a unicode character instead in a40eb1ca59a440ffd68a3a4a1316dc59cc05fe0f.

src/datasources/email-template/pushwoosh-template.service.ts
iamacook: I've added a `safeWebApp.baseUri` configuration value in a40eb1ca59a440ffd68a3a4a1316dc59cc05fe0f.

src/datasources/email-template/pushwoosh-template.service.ts
hectorgomezv: Great, solved then 😄

hectorgomezv (approved):
Awesome job 👏🏻

PR #1010, opened 2024-01-11
Ignore E2E tests in coverage

iamacook:
This excludes our `*.e2e.spec.ts` files from our test coverage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,7 +90,8 @@
     "coverageDirectory": "../coverage",
     "coveragePathIgnorePatterns": [
       "index.ts",
-      ".+\\/__tests__\\/.+\\.factory.(t|j)s"
+      ".+\\/__tests__\\/.+\\.factory.(t|j)s",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -90,7 +90,8 @@
     "coverageDirectory": "../coverage",
     "coveragePathIgnorePatterns": [
       "index.ts",
-      ".+\\/__tests__\\/.+\\.factory.(t|j)s"
+      ".+\\/__tests__\\/.+\\.factory.(t|j)s",
+      ".e2e.spec.ts"
```
fmrsabino, line 94: This should be `.e2e-spec.ts` right?

iamacook: Very true. It seems like Jest is forgiving and accepted the above. I've changed it to `.e2e-spec.ts` in 5101f9ec6788ebf8bd628877f0edaf8fe500d5f7, however.

fmrsabino (approved):
👏 🚀

PR #1014, opened 2024-01-11
Use same global setup in all Jest configs

iamacook:
This updates our custom Jest configurations (`jest-all.json` and `jest-e2e.json`) to use the same global setup as our standard tests (setting the timezone of tests to UTC).


fmrsabino (approved):
🚀

PR #1016, opened 2024-01-12
Throttle email edits according to verification code TTL

iamacook:
This throttle email edit requests according to the TTL of verification codes. If an edit request occurs again within the TTL of the generated verification occurs, a `429` is returned.

- `EmailRepository['editEmail']` now includes a condition checking the presence of `verificationGeneratedOn` for the current email, then checking whether the current request is before or after the TTL and throwing accordingly.
- `EditTimespanError` error and associated `EmailEditTimespanExceptionFilter` exception filter have been included.
- Associated test coverage added.

### DIFF
--- a/src/domain/email/email.repository.ts
+++ b/src/domain/email/email.repository.ts
@@ -171,6 +171,7 @@ export class EmailRepository implements IEmailRepository {
     const newEmail = new EmailAddress(args.emailAddress);
     const currentEmail = await this.emailDataSource.getEmail(args);
 
+    // Prevent subsequent edit if verification code still valid
     if (
       currentEmail.verificationGeneratedOn &&
       this._isEmailVerificationCodeValid(currentEmail)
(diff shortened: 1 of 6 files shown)


src/domain/email/email.repository.ts
```diff
@@ -170,6 +171,19 @@ export class EmailRepository implements IEmailRepository {
     const newEmail = new EmailAddress(args.emailAddress);
     const currentEmail = await this.emailDataSource.getEmail(args);
 
+    if (
+      currentEmail.verificationGeneratedOn &&
+      this._isEmailVerificationCodeValid(currentEmail)
```
hectorgomezv, line 176: Not sure I get the purpose of this condition. The error seems to be thrown when the currentEmail has a previously generated verificationCode, and that verificationCode is valid, is that correct? Maybe I'm missing something.

iamacook: > Not sure I get the purpose of this condition. The error seems to be thrown when the currentEmail has a previously generated verificationCode, and that verificationCode is valid, is that correct?

Yes, this prevents editing again within the TTL window as the `verificationCode` is only valid within it.

hectorgomezv: Oh, you are right, I was misinterpreting the requirement 😅 Thanks!

hectorgomezv (approved):
Great! 🚀

hectorgomezv (approved):

PR #1023, opened 2024-01-15
Reduce reliance on `lodash`

iamacook:
This migrates trivial `lodash` functions to their vanilla counterparts, reducing the import cost and lowering the bundle size, e.g. `isArray` which is no longer used:

image

The following functions have been migrated:

- [`get`](https://lodash.com/docs/4.17.15#get) -> accessing property directly with optional chaining
- [`inRange`](https://lodash.com/docs/4.17.15#inRange) -> `>=` and `<`
- [`isArray`](https://lodash.com/docs/4.17.15#isArray) ->`Array.isArray`
- [`flatten`](https://lodash.com/docs/4.17.15#flatten) -> `flat`
- [`head(arr)`](https://lodash.com/docs/4.17.15#head) -> arr[0]
- [`last(arr)`](https://lodash.com/docs/4.17.15#last) -> arr[.at(-1)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at)

### DIFF
--- a/src/routes/transactions/transactions.service.ts
+++ b/src/routes/transactions/transactions.service.ts
@@ -489,7 +489,7 @@ export class TransactionsService {
   private getFirstTransactionNonce(
     page: Page<DomainMultisigTransaction>,
   ): number | null {
-    return page.results[0].nonce ?? null;
+    return page.results[0]?.nonce ?? null;
   }
 
   private getLastTransactionNonce(
--- a/src/routes/transactions/transactions.service.ts
+++ b/src/routes/transactions/transactions.service.ts
@@ -1,5 +1,4 @@
 import { Inject, Injectable } from '@nestjs/common';
-import { head, last } from 'lodash';
 import { MultisigTransaction as DomainMultisigTransaction } from '@/domain/safe/entities/multisig-transaction.entity';
 import { SafeRepository } from '@/domain/safe/safe.repository';
 import { ISafeRepository } from '@/domain/safe/safe.repository.interface';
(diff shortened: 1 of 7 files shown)


src/routes/transactions/transactions.service.ts
```diff
@@ -490,12 +489,12 @@ export class TransactionsService {
   private getFirstTransactionNonce(
     page: Page<DomainMultisigTransaction>,
   ): number | null {
-    return head(page.results)?.nonce ?? null;
+    return page.results[0].nonce ?? null;
```
hectorgomezv, line 492: Is it possible that `page.results` is `null` in this context? It seems the prior implementation was protected against that (`head(page.results)` would return `undefined`, but `page.results[0]` would throw an uncaught error) 🤔

iamacook: Good catch! I've added optional chaining in e592ccb48f96095fc675168ec1e04d189c445dbf.

src/routes/transactions/transactions.service.ts
```diff
@@ -490,12 +489,12 @@ export class TransactionsService {
   private getFirstTransactionNonce(
     page: Page<DomainMultisigTransaction>,
   ): number | null {
-    return head(page.results)?.nonce ?? null;
+    return page.results[0]?.nonce ?? null;
```
hectorgomezv, line 492:
```suggestion
    return page.results?.[0]?.nonce ?? null;
```

I think we need to cover the case where `page.results` is `null` with optional chaining before accessing the array.

src/routes/transactions/transactions.service.ts
```diff
@@ -490,12 +489,12 @@ export class TransactionsService {
...
   }
 
   private getLastTransactionNonce(
     page: Page<DomainMultisigTransaction>,
   ): number | null {
-    return last(page.results)?.nonce ?? null;
+    return page.results.at(-1)?.nonce ?? null;
```
hectorgomezv, line 498:
```suggestion
    return page.results?.at(-1)?.nonce ?? null;
```

hectorgomezv (approved):

fmrsabino (approved):
🚀

PR #1027, opened 2024-01-15
Decode `multiSend` alerts first and update tests to match module transactions

iamacook:
Recovery transactions are not wrapped in an `execTransaction` as they are executed by a module (see [example](https://sepolia.etherscan.io/tx/0x83eae8585eb068e7de8e641eb1a720c77e4e3d7fa876188d50aab13312a6fa7a#eventlog)). Attempting to decode the incoming alert as an `execTransaction` would fail if it is a `multiSend`-based recovery attempt.

This inverts the associated logic to first try decoding `multiSend` and then Safe transactions. The relevant tests have been updated accordingly.

Changes:

- `AlertsRepository['_decodeTransactionAdded']` tries to decode the `multiSend` call and then Safe methods either within the batch or if the initial decode fails.
- Each Safe method decode attempt is now handled by `AlertsRepository['_decodeRecoveryTransaction']` which checks that the function name matches that expected, otherwise throwing.
- `alerts.controller.spec` mock transactions have been updated to _not_ wrap `multiSend` recoveries in `execTransaction` calls.

### DIFF
--- a/src/domain/alerts/contracts/safe-decoder.helper.ts
+++ b/src/domain/alerts/contracts/safe-decoder.helper.ts
@@ -7,7 +7,6 @@ const OWNER_MANAGER_ABI = parseAbi([
   'function removeOwner(address prevOwner, address owner, uint256 _threshold)',
   'function swapOwner(address prevOwner, address oldOwner, address newOwner)',
   'function changeThreshold(uint256 _threshold)',
-  // Batched owner management transactions are executed via MultiSend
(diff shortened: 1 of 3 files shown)


src/domain/alerts/contracts/safe-decoder.helper.ts
```diff
@@ -7,7 +7,6 @@ const OWNER_MANAGER_ABI = parseAbi([
   'function removeOwner(address prevOwner, address owner, uint256 _threshold)',
   'function swapOwner(address prevOwner, address oldOwner, address newOwner)',
   'function changeThreshold(uint256 _threshold)',
-  // Batched owner management transactions are executed via MultiSend
```
iamacook, line 10: This was an incorrect assumption and has therefore been removed.

hectorgomezv (approved):

PR #1028, opened 2024-01-15
Add email subscriptions datasource

fmrsabino:
Adds two new tables: `subscriptions` and `account_subscriptions`.

The `subscriptions` table holds a collection of subscription types:

|         id         |         key          |     name      |
|:------------------:|:--------------------:|:-------------:|
| SERIAL PRIMARY KEY | TEXT NOT NULL UNIQUE | TEXT NOT NULL |

The `key` should be used to reference the category in a human-readable format. The `name` should be used for rendering purposes.

A default entry is added with `migrations/000001_initial`: `key='account_recovery' name='Account Recovery'`. (Note: migration was edited directly because there's no deployment yet).

The `account_subscriptions` table holds the data on the account subscriptions which are active (i.e., if there is a row, then there is a subscription).

| account_id | subscription_id |
|:----------:|:---------------:|
| FOREIGN KEY (emails.account_emails) | FOREIGN KEY (emails.subscriptions) |

---
The following functionality was added to `IEmailDataSource`:
- `getSubscriptions` – Gets all the subscriptions for the account on chain id, with the specified Safe address.
- `subscribe` – Subscribes the account on chain id, with the Safe address to a category.
- `unsubscribe` – Unsubscribes from the notification category with the provided category key.
- `unsubscribeAll` – Unsubscribes from all notification categories.

### DIFF
--- a/migrations/00001_initial/index.sql
+++ b/migrations/00001_initial/index.sql
@@ -13,5 +13,28 @@ CREATE table emails.account_emails
     verification_code              text,
     verification_code_generated_on timestamp with time zone,
     verification_sent_on           timestamp with time zone,
+    unsubscription_token           uuid                  NOT NULL,
     UNIQUE (chain_id, safe_address, account)
 );
+
+DROP TABLE IF EXISTS emails.subscriptions CASCADE;
+CREATE TABLE emails.subscriptions
+(
+    id   SERIAL PRIMARY KEY,
+    key  TEXT NOT NULL UNIQUE,
+    name TEXT NOT NULL
+);
+
+-- Add the default subscription type: account_recovery
+INSERT INTO emails.subscriptions (key, name)
+VALUES ('account_recovery', 'Account Recovery');
+
+DROP TABLE IF EXISTS emails.account_subscriptions CASCADE;
+CREATE TABLE emails.account_subscriptions
+(
+    account_id      INT,
+    subscription_id INT,
+    FOREIGN KEY (account_id) REFERENCES emails.account_emails (id) ON DELETE CASCADE,
+    FOREIGN KEY (subscription_id) REFERENCES emails.subscriptions (id) ON DELETE CASCADE,
+    UNIQUE (account_id, subscription_id)
+)
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -541,7 +541,7 @@ describe('Email Datasource Tests', () => {
     expect(subscriptions).toContainEqual(subscription);
   });
 
-  it('unsubscribes to a category successfully', async () => {
+  it('unsubscribes from a category successfully', async () => {
     const chainId = faker.number.int({ max: DB_CHAIN_ID_MAX_VALUE }).toString();
     const safeAddress = faker.finance.ethereumAddress();
     const emailAddress = new EmailAddress(faker.internet.email());
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -600,7 +600,7 @@ describe('Email Datasource Tests', () => {
     expect(currentSubscriptions).not.toContainEqual(subscriptions[0]);
   });
 
-  it('unsubscribes to non existent category', async () => {
+  it('unsubscribes from a non-existent category', async () => {
     const chainId = faker.number.int({ max: DB_CHAIN_ID_MAX_VALUE }).toString();
     const safeAddress = faker.finance.ethereumAddress();
     const emailAddress = new EmailAddress(faker.internet.email());
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -661,7 +661,7 @@ describe('Email Datasource Tests', () => {
     expect(result).toHaveLength(0);
   });
 
-  it('unsubscribes to all categories successfully', async () => {
+  it('unsubscribes from all categories successfully', async () => {
     const chainId = faker.number.int({ max: DB_CHAIN_ID_MAX_VALUE }).toString();
     const safeAddress = faker.finance.ethereumAddress();
     const emailAddress = new EmailAddress(faker.internet.email());
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -598,6 +598,7 @@ describe('Email Datasource Tests', () => {
     });
     expect(result).toContainEqual(subscriptions[0]);
     expect(currentSubscriptions).not.toContainEqual(subscriptions[0]);
+    expect(currentSubscriptions).toContainEqual(subscriptions[1]);
   });
 
   it('unsubscribes from a non-existent category', async () => {
@@ -658,7 +659,14 @@ describe('Email Datasource Tests', () => {
       token: faker.string.uuid(),
     });
 
+    const currentSubscriptions = await target.getSubscriptions({
+      chainId,
+      safeAddress,
+      account,
+    });
     expect(result).toHaveLength(0);
+    // Length of 1 since there is a default subscription
+    expect(currentSubscriptions).toHaveLength(1);
   });
 
   it('unsubscribes from all categories successfully', async () => {
@@ -705,8 +713,7 @@ describe('Email Datasource Tests', () => {
       categoryKey: subscriptions[1].key,
     });
 
-    // Unsubscribe from one category
-    await target.unsubscribeAll({
+    const result = await target.unsubscribeAll({
       token: unsubscriptionToken,
     });
 
@@ -716,5 +723,8 @@ describe('Email Datasource Tests', () => {
       account,
     });
     expect(currentSubscriptions).toHaveLength(0);
+    expect(result).toHaveLength(3);
+    expect(result).toContainEqual(subscriptions[0]);
+    expect(result).toContainEqual(subscriptions[1]);
   });
 });
--- a/src/datasources/email/email.datasource.spec.ts
+++ b/src/datasources/email/email.datasource.spec.ts
@@ -31,7 +31,7 @@ describe('Email Datasource Tests', () => {
   });
 
   afterEach(async () => {
-    await sql`TRUNCATE emails.account_emails CASCADE`;
+    await sql`TRUNCATE TABLE emails.account_emails, emails.account_subscriptions CASCADE`;
   });
 
   afterAll(async () => {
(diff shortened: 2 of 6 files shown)


src/datasources/email/email.datasource.spec.ts
```diff
@@ -440,7 +466,255 @@ describe('Email Datasource Tests', () => {
...
+
+    const subscriptions = await target.getSubscriptions({
+      chainId,
+      safeAddress,
+      account,
+    });
+    expect(subscriptions).toContainEqual(subscription);
+  });
+
+  it('unsubscribes to a category successfully', async () => {
```
iamacook, line 544:
```suggestion
  it('unsubscribes from a category successfully', async () => {
```

src/datasources/email/email.datasource.spec.ts
```diff
@@ -440,7 +466,255 @@ describe('Email Datasource Tests', () => {
...
+    const currentSubscriptions = await target.getSubscriptions({
+      chainId,
+      safeAddress,
+      account,
+    });
+    expect(result).toContainEqual(subscriptions[0]);
+    expect(currentSubscriptions).not.toContainEqual(subscriptions[0]);
+  });
+
+  it('unsubscribes to non existent category', async () => {
```
iamacook, line 603:
```suggestion
  it('unsubscribes from a non-existent category', async () => {
```

src/datasources/email/email.datasource.spec.ts
```diff
@@ -440,7 +466,255 @@ describe('Email Datasource Tests', () => {
...
+    // Unsubscribe from one category
+    const result = await target.unsubscribe({
+      categoryKey: subscriptions[0].key,
+      token: faker.string.uuid(),
+    });
+
+    expect(result).toHaveLength(0);
+  });
+
+  it('unsubscribes to all categories successfully', async () => {
```
iamacook, line 664:
```suggestion
  it('unsubscribes from all categories successfully', async () => {
```

migrations/00001_initial/index.sql
```diff
@@ -13,5 +13,28 @@ CREATE table emails.account_emails
...
+
+DROP TABLE IF EXISTS emails.account_subscriptions CASCADE;
+CREATE TABLE emails.account_subscriptions
+(
+    account_id      INT,
+    subscription_id INT,
+    FOREIGN KEY (account_id) REFERENCES emails.account_emails (id) ON DELETE CASCADE,
+    FOREIGN KEY (subscription_id) REFERENCES emails.subscriptions (id) ON DELETE CASCADE,
+    UNIQUE (account_id, subscription_id)
+)
```
iamacook, line 40: Nit: can we prettify this to match the `emails.account_emails` format?

src/datasources/email/email.datasource.spec.ts
```diff
@@ -440,7 +466,255 @@ describe('Email Datasource Tests', () => {
...
+      token: unsubscriptionToken,
+    });
+
+    const currentSubscriptions = await target.getSubscriptions({
+      chainId,
+      safeAddress,
+      account,
+    });
+    expect(result).toContainEqual(subscriptions[0]);
+    expect(currentSubscriptions).not.toContainEqual(subscriptions[0]);
```
iamacook, line 600: I'd suggest we also add a check for the subscription of `subscriptions[1]` here to be on the safe side.

src/datasources/email/email.datasource.spec.ts
```diff
@@ -440,7 +466,255 @@ describe('Email Datasource Tests', () => {
...
+      unsubscriptionToken,
+    });
+
+    // Unsubscribe from one category
+    const result = await target.unsubscribe({
+      categoryKey: subscriptions[0].key,
+      token: faker.string.uuid(),
+    });
+
+    expect(result).toHaveLength(0);
```
iamacook, line 661: I think we should also check that the subscriptions of `emailAddress` has not been touched after this.

src/datasources/email/email.datasource.ts
```diff
@@ -75,12 +83,24 @@ export class EmailDataSource implements IEmailDataSource {
...
         VALUES (${args.chainId}, ${args.emailAddress.value}, ${args.safeAddress}, ${args.account}, ${args.code},
-                ${args.codeGenerationDate})
+                ${args.codeGenerationDate}, ${args.unsubscriptionToken})
+        RETURNING *
+    `;
+
+    // When registering a new email address, the account_recovery subscription should be enabled by default
+    await this.sql`
+        INSERT INTO emails.account_subscriptions (account_id, subscription_id)
+        SELECT ae.id, s.id
```
iamacook, line 99: Nit: please resolve if I'm not understanding correctly but are `ae` and `s` self-define variables? If so, could we name them clearer?

src/datasources/email/email.datasource.ts
```diff
@@ -199,4 +221,87 @@ export class EmailDataSource implements IEmailDataSource {
...
+
+  async subscribe(args: {
+    chainId: string;
+    safeAddress: string;
+    account: string;
+    categoryKey: string;
+  }): Promise<DomainSubscription[]> {
+    const subscriptions = await this
+      .sql`INSERT INTO emails.account_subscriptions (account_id, subscription_id)
+           SELECT ae.id AS account_id,
```
iamacook, line 255: Nit: as before regarding variable names. Please resolve if I misunderstand.

hectorgomezv (changes requested):
Great job! 🚀🚀🚀

Just left some nit comments.

migrations/00001_initial/index.sql
fmrsabino: Hmm actually this was done with the same formatter as the one used for `emails.account_emails` (the one bundled with IntelliJ). We can think about adding a tool to format SQL files separately. Wdyt?

src/datasources/email/email.datasource.spec.ts
fmrsabino: This is kind of covered in `expect(result).toContainEqual(subscriptions[0])` since it should return the subscriptions removed but I will add the assertion which I think is also valid!

migrations/00001_initial/index.sql
iamacook: Sure

src/datasources/email/email.datasource.ts
fmrsabino: They are aliases to the table names (they come from the `FROM` clause). I will make the variables clearer!

src/datasources/email/email.datasource.ts
```diff
@@ -199,4 +221,87 @@ export class EmailDataSource implements IEmailDataSource {
...
+  }
+
+  async subscribe(args: {
+    chainId: string;
+    safeAddress: string;
+    account: string;
+    categoryKey: string;
+  }): Promise<DomainSubscription[]> {
+    const subscriptions = await this
+      .sql`INSERT INTO emails.account_subscriptions (account_id, subscription_id)
```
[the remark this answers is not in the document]
fmrsabino: Can you show an example on where you'd like to have the parentheses?

Regarding indentation it is left to prettier. I can change it locally but it is then formatted as you see here.

iamacook (approved):

src/datasources/email/email.datasource.ts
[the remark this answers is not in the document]
hectorgomezv: I was thinking in something like this to distinguish a bit more the main query (INSERT) from the subquery (nested SELECT):

```sql
INSERT INTO emails.account_subscriptions (account_id, subscription_id)
    (SELECT account_emails.id AS account_id,
            subscriptions.id  AS subscription_id
    FROM emails.account_emails
            CROSS JOIN emails.subscriptions
    WHERE account_emails.chain_id = ${args.chainId}
    AND account_emails.safe_address = ${args.safeAddress}
    AND account_emails.account = ${args.account}
    AND subscriptions.key = ${args.categoryKey})
RETURNING *`;
```

This would also apply here: https://github.com/safe-global/safe-client-gateway/blob/040d7d6378f9fe26daf364d342d3a9481c361b8a/src/datasources/email/email.datasource.ts#L98 (out of the scope of this PR)

But feel free to ignore this, as we commented in the past since we don't have strict rules regarding SQL formatting yet.

src/datasources/email/email.datasource.spec.ts
fmrsabino: Addressed in [040d7d6](https://github.com/safe-global/safe-client-gateway/pull/1028/commits/040d7d6378f9fe26daf364d342d3a9481c361b8a)

src/datasources/email/email.datasource.spec.ts
fmrsabino: Addressed in [040d7d6](https://github.com/safe-global/safe-client-gateway/pull/1028/commits/040d7d6378f9fe26daf364d342d3a9481c361b8a)

src/datasources/email/email.datasource.ts
fmrsabino: Addressed in [040d7d6](https://github.com/safe-global/safe-client-gateway/pull/1028/commits/040d7d6378f9fe26daf364d342d3a9481c361b8a)

src/datasources/email/email.datasource.ts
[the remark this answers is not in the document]
fmrsabino: With this example in particular I actually don't find that the parentheses help that much because it'd be for most of the query.

However if we talk about indentation then this is something that we should align. It'd be better to have a tool to format SQL queries instead.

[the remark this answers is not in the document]
hectorgomezv: While I agree it's not critical to include parenthesis in this particular case regarding readability, I do think it's a common/good practice to use them for subqueries (even if they enclose most of the query). I see this practice in all the references I'm checking:
- https://www.postgresql.org/docs/current/functions-subquery.html
- https://www.postgresql.org/docs/current/queries-with.html
- https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/SELECT.html#GUID-CFA006CA-6FF1-4972-821E-6996142A51C6__I2066912
- https://www.oreilly.com/library/view/practical-postgresql/9781449309770/ch04s07.html

But it's fine by me if you think we don't need it. I agree we can defer these discussions until we have rules/a tool to format SQL 👍

hectorgomezv (approved):

src/datasources/email/email.datasource.ts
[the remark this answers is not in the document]
fmrsabino: Sounds good. I don't have a strong opinion but since it improves readability I added the parentheses 👍 Added in [81f95ca](https://github.com/safe-global/safe-client-gateway/pull/1028/commits/81f95cac601715d232425e390d9090640f918d9e)

iamacook (approved):

PR #1037, opened 2024-01-16
Add email subscription cancellation routes

fmrsabino:
- Adds `DELETE /v1/subscriptions?category=<string>&token=<string>`
  * This endpoint deletes the subscription of the given `category`, if a valid token is provided.
  * The token must be a UUID. A `400` is returned otherwise.
- Adds `DELETE /v1/subscriptions/all?token=<string>`
  * This endpoint deletes all subscriptions of an account with the associated `token`.
  * The token must be a UUID. A `400` is returned otherwise.
- Saving a new email with `EmailRepository` now subscribes to the Account Recovery category – this is a default subscription for all new accounts registered.

An indication of the token's existence is not returned to the client i.e. if an invalid token is provided the response is still successful. This was done while considering that the endpoint is unauthenticated and the link should be _private_ to the email owner. Therefore, no feedback is given regarding the validity of said token.

### DIFF
--- a/src/routes/subscriptions/subscription.controller.spec.ts
+++ b/src/routes/subscriptions/subscription.controller.spec.ts
@@ -52,7 +52,7 @@ describe('Subscription Controller tests', () => {
     await app.close();
   });
 
-  it('deletes category successfully', async () => {
+  it('unsubscribes from a category successfully', async () => {
     const subscriptionKey = faker.word.sample();
     const subscriptionName = faker.word.sample(2);
     const token = faker.string.uuid();
--- a/src/routes/subscriptions/subscription.controller.spec.ts
+++ b/src/routes/subscriptions/subscription.controller.spec.ts
@@ -1,0 +1,168 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { AppModule } from '@/app.module';
+import configuration from '@/config/entities/__tests__/configuration';
+import { EmailControllerModule } from '@/routes/email/email.controller.module';
+import { EmailApiModule } from '@/datasources/email-api/email-api.module';
+import { TestEmailApiModule } from '@/datasources/email-api/__tests__/test.email-api.module';
+import { CacheModule } from '@/datasources/cache/cache.module';
+import { TestCacheModule } from '@/datasources/cache/__tests__/test.cache.module';
+import { RequestScopedLoggingModule } from '@/logging/logging.module';
+import { TestLoggingModule } from '@/logging/__tests__/test.logging.module';
+import { NetworkModule } from '@/datasources/network/network.module';
+import { TestNetworkModule } from '@/datasources/network/__tests__/test.network.module';
+import { TestAppProvider } from '@/__tests__/test-app.provider';
+import * as request from 'supertest';
+import { faker } from '@faker-js/faker';
+import { Subscription } from '@/domain/account/entities/subscription.entity';
+import { TestAccountDataSourceModule } from '@/datasources/account/__tests__/test.account.datasource.module';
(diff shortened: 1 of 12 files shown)


src/routes/subscriptions/subscription.controller.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    await request(app.getHttpServer())
+      .delete(`/v1/subscriptions/?category=${subscriptionKey}&token=${token}`)
+      .expect(200)
+      .expect({});
+
+    expect(emailDatasource.unsubscribe).toHaveBeenCalledWith({
+      categoryKey: subscriptionKey,
+      token: token,
+    });
+    expect(emailDatasource.unsubscribeAll).toHaveBeenCalledTimes(0);
```
iamacook, line 75: Nit: can we also check that it no longer exists in the database?

src/routes/subscriptions/subscription.controller.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    await request(app.getHttpServer())
+      .delete(`/v1/subscriptions/all?token=${token}`)
+      .expect(200)
+      .expect({});
+
+    expect(emailDatasource.unsubscribeAll).toHaveBeenCalledWith({
+      token: token,
+    });
+    expect(emailDatasource.unsubscribe).toHaveBeenCalledTimes(0);
```
iamacook, line 126: Nit: can we also check here that the database updated successfully?

src/routes/subscriptions/subscription.controller.spec.ts
fmrsabino: The test setup does not require a database (it's replaced with a fake). So that assertion would not work here because there's no database.

However that integration is tested in `src/datasources/email/email.datasource.spec.ts`.

The goal of this test is to cover everything up until the database call. Database calls are then covered in the test that I mentioned.

src/routes/subscriptions/subscription.controller.spec.ts
fmrsabino: See https://github.com/safe-global/safe-client-gateway/pull/1037#discussion_r1459039745

iamacook (dismissed):

src/routes/subscriptions/subscription.controller.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+
+    app = await new TestAppProvider().provide(moduleFixture);
+    await app.init();
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+
+  it('deletes category successfully', async () => {
```
hectorgomezv, line 54: Very nitpick, but: I'd say `unsubscribes from a caategory` or similiar, instead of `deletes category` (also in the other test cases). Otherwise I am under the impresion the test verifies the removal of a category itself.

Also, pretty unrelated to thi PR: the endpoing uses the keyword `category`, which I think it's meaningful, but at the database/datasource level we use `subscription` to refer to a category/topic. Does it make sense to unify this semantically?

hectorgomezv (dismissed):
👏🏻👏🏻👏🏻

fmrsabino:
I will wait for https://github.com/safe-global/safe-client-gateway/pull/1050 to be merged and rebase this one as it makes more sense given the changes that the other introduces.

src/routes/subscriptions/subscription.controller.spec.ts
fmrsabino: Done in [e075ea0](https://github.com/safe-global/safe-client-gateway/pull/1037/commits/e075ea00852417361e6628660ca91308a1d25852)

PR #1038, opened 2024-01-16
Remove flush endpoint

hectorgomezv:
Depends on https://github.com/safe-global/safe-config-service/pull/1027 **(this PR should be merged after that one)**

**Changes:**
- Removes the `FlushModule` routes module, including controllers, services, and entities.
- Removes flush-associated classes in the domain layer.

**Out of scope:**
It's out of the scope of this PR to determine if the cache management classes can be simplified (e.g. by removing pattern-based invalidations). This will be tackled separately.


iamacook (approved):

fmrsabino (approved):
🚀

PR #1041, opened 2024-01-17
Remove type assertions

iamacook:
This removes type assertions in favour of reliance on return types or variable type definitions. Type assertions "force" a value to be correct and can therefore mean that an incorrect value is valid.

Pleast note: there are two remaining assertions but their removal causes errors. They are fixed in #1040 and #1043.

### DIFF
--- a/src/domain/balances/entities/balance.entity.ts
+++ b/src/domain/balances/entities/balance.entity.ts
@@ -1,9 +1,18 @@
 import { BalanceToken } from '@/domain/balances/entities/balance.token.entity';
 
-export interface Balance {
-  tokenAddress: string | null;
-  token: BalanceToken | null;
+interface NativeBalance {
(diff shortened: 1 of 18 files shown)


iamacook:
I'm converting this to a draft whilst I investigate the remaining assertions.

src/domain/balances/entities/balance.entity.ts
```diff
@@ -1,9 +1,18 @@
...
+  tokenAddress: string;
+  token: BalanceToken;
+  balance: string;
+}
+
+export type Balance = (NativeBalance | Erc20Balance) & {
   fiatBalance: string | null;
   fiatConversion: string | null;
-}
+};
```
iamacook, line 18: The split is required to maintain type specificity in `BalancesService['_mapBalance']`.

PR #1043, opened 2024-01-17
Remove assertions from token type mapping

iamacook:
This removes the assertions from human readable token type types. As the types are no longer asserted, associated error handling for invalid types had to be added as well.


fmrsabino (approved):
🚀

PR #1044, opened 2024-01-18
Add linting to enforce return types and assertion consistency

iamacook:
With the removal of angle bracket (`<Type>value`) type assertion (#1041), favouring return types and `as` assertions (if absolutely necessary), this adds the following linting rules and fixes any issues associated with them.

### Errors
Require explicit return types on functions and class methods. ([`explicit-function-return-type`](https://typescript-eslint.io/rules/explicit-function-return-type/))

### Warnings
Enforce consistent usage of `as` type assertions. ([`consistent-type-assertions`](https://typescript-eslint.io/rules/consistent-type-assertions/))

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -15,9 +15,13 @@ module.exports = {
   ignorePatterns: ['.eslintrc.js'],
   rules: {
     '@typescript-eslint/interface-name-prefix': 'off',
-    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'error',
     '@typescript-eslint/explicit-module-boundary-types': 'off',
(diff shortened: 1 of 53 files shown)


.eslintrc.js
```diff
@@ -15,9 +15,13 @@ module.exports = {
   ignorePatterns: ['.eslintrc.js'],
   rules: {
     '@typescript-eslint/interface-name-prefix': 'off',
-    '@typescript-eslint/explicit-function-return-type': 'off',
+    '@typescript-eslint/explicit-function-return-type': 'error',
```
iamacook, line 18: I've set this to `error` as it is now our source of truth having removed assertions.

fmrsabino (approved):

PR #1047, opened 2024-01-18
Use feature flag instead of import in `recovery.controller` test

iamacook:
This uses the feature flag instead of importing the `Recovery` module directly in the `recovery.controller` test.


fmrsabino (approved):
🚀

hectorgomezv (approved):

PR #1048, opened 2024-01-19
Remove `axios` in favour of `fetch`

iamacook:
This removes `axios`, replacing it with `fetch`. A new `FetchClient` is now used in the `NetworkModule`.

A new `fetchClientFactory` replaces what was `axiosFactory`. It returns a fetch "client" that adheres to the previous requirements/interface, throwing the correct error object structures to match: `NetworkOtherError`, `NetworkRequestError` and `NetworkResponseError`

The `AxiosNetworkService` is now `FetchNetworkService` and all associated test coverage has been accordingly updated to work with the new factory.


src/datasources/network/entities/network.request.entity.ts
```diff
@@ -1,4 +1,4 @@
 export interface NetworkRequest {
-  headers?: Record<string, string | number | boolean>;
+  headers?: Record<string, string>;
```
iamacook, line 2: `fetch` only permits certain header types:

```ts
type HeadersInit = [string, string][] | Record<string, string> | Headers;
```

So as to not make many changes, I adjusted this to suit the object notation.

src/datasources/network/network.module.ts
```diff
@@ -1,24 +1,55 @@
...
-    httpsAgent: new https.Agent({ keepAlive: true }),
-  });
+
+  return async <T>(url: string, options: RequestInit): Promise<{ data: T }> => {
+    // If general error - NetworkOtherError
+    const response = await fetch(url, {
+      ...options,
+      signal: AbortSignal.timeout(requestTimeout),
+      keepalive: true,
+    });
```
fmrsabino, line 28: If an error is thrown in the `fetch` then this error will propagate upwards instead of being mapped to our `NetworkRequestError`

fmrsabino (changes requested):

src/datasources/network/network.module.ts
```diff
@@ -1,24 +1,55 @@
...
+
+    const data = await response.json().catch(() => null);
+
+    if (!response.ok) {
+      if (!data) {
+        // Response error (no response) - NetworkRequestError
+        throw {
+          request: new URL(url),
+        };
```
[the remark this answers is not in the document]
fmrsabino: On a second thought it should also be ok to leave the types out of the module and have those in the `FetchNetworkService` as you currently have.

src/datasources/errors/http-error-factory.spec.ts
```diff
@@ -30,15 +29,6 @@ describe('HttpErrorFactory', () => {
     expect(actual.message).toBe('Service unavailable');
   });
 
-  it('should create an HttpException with 503 status when there is an any other network error', async () => {
-    const httpError = new NetworkOtherError('some other error');
```
iamacook, line 34: `NetworkOtherError` is no longer thrown.

src/datasources/network/fetch.network.service.ts
```diff
@@ -0,0 +1,110 @@
...
+    if (error.response) {
+      this.logErrorResponse(error, responseTimeMs);
+      throw new NetworkResponseError(error.response.status, error.data);
+    } else {
+      throw new NetworkRequestError(error.request);
+    }
+  }
+
+  private logErrorResponse(error, responseTimeMs: number): void {
+    this.loggingService.debug({
```
iamacook, line 100: We should evaluate how much we are using this log as it is a `debug`.

src/datasources/network/network.service.interface.ts
```diff
@@ -4,19 +4,16 @@ import { NetworkResponse } from '@/datasources/network/entities/network.response
...
     data: object,
     networkRequest?: NetworkRequest,
-  ): Promise<R>;
+  ): Promise<NetworkResponse<T>>;
 
-  delete<T = any, R = NetworkResponse<T>>(
-    url: string,
-    data?: object,
-  ): Promise<R>;
+  delete<T>(url: string, data?: object): Promise<NetworkResponse<T>>;
```
iamacook, line 18: These types were not correct. They were previously returning `any`. Changing these caused the other type fixes in this PR.

src/domain/prices/entities/asset-price.entity.ts
```diff
@@ -1,3 +1,3 @@
 export interface AssetPrice {
-  [assetName: string]: Record<string, number>;
+  [assetName: string]: Record<string, number | null>;
```
iamacook, line 2: This (and the following) type changes are changed according to the logic of `coingecko-api.service.ts`.

src/datasources/network/network.module.ts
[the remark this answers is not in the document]
iamacook: After discussion and adjustments, I've added a note to tackle this in a follow-up PR. Throwing specific errors here causes a waterfall of type errors that I think would be better fixed separately.

src/datasources/network/network.module.ts
```diff
@@ -1,24 +1,55 @@
...
+            ...response,
+            data,
+          },
+        };
+      }
+    }
+
+    return {
+      data: data as T,
+    };
```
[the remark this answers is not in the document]
iamacook: Adjusted in 3c3a5f5053f1d8d0524d13cfbe9c2302153f6572.

src/datasources/network/network.module.ts
iamacook: This is now caught and thrown correctly as of 3c3a5f5053f1d8d0524d13cfbe9c2302153f6572.

src/datasources/network/network.module.spec.ts
```diff
@@ -19,14 +22,26 @@ describe('NetworkModule', () => {
...
     await app.init();
 
-    expect(axios.defaults.timeout).toBe(httpClientTimeout);
+    try {
+      await fetchClient('', {
+        method: 'GET',
+      });
+    } catch {
+      // fetch response is not mocked but we are only concerned with RequestInit options
+    }
```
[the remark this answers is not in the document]
iamacook: Updated in 3c3a5f5053f1d8d0524d13cfbe9c2302153f6572.

src/datasources/network/fetch.network.service.spec.ts
```diff
@@ -234,27 +252,15 @@ describe('AxiosNetworkService', () => {
...
-      const data = { some_data: 'some_data' };
-
-      await target.delete(url, data);
-
-      expect(axiosMock.delete).toHaveBeenCalledTimes(1);
-      expect(axiosMock.delete).toHaveBeenCalledWith(url, { data: data });
-    });
-
-    it(`delete forwards unknown error as NetworkOtherError`, async () => {
+    it(`get uses DELETE method`, async () => {
```
fmrsabino, line 255: "delete uses DELETE method"

src/datasources/network/network.service.interface.ts
fmrsabino: `any` was the default right? So a return type could hint on what value `T` should've taken here (talking about the changes before this diff).

fmrsabino (approved):

src/datasources/network/fetch.network.service.spec.ts
iamacook: Fixed in c9ba85577a8d93c61ebf5d8aa38d80fbcaf2aebe.

src/datasources/network/network.service.interface.ts
iamacook: Very true. I've reverted the default value removal in c9ba85577a8d93c61ebf5d8aa38d80fbcaf2aebe, as well as removed the associated generics that were required because of the change.

fmrsabino: We can discuss the value of this default too. I was not necessarily against the change proposed here but wondering that the type could come from somewhere else (even if not explicitly declared in the generic) 😄

fmrsabino (approved):

PR #1051, opened 2024-01-22
Enable strict mode

iamacook:
This enables the `strict` flag of TypeScript to "a wide range of type checking behavior that results in stronger guarantees of program correctness" and updates the associated issues found, notably:

- Error type checking ("standard" errrors that are thrown are `unknown`)
- Assertions added to DTOs ([as per NestJS' author suggestion](https://github.com/nestjs/nest/issues/4178#issuecomment-614006219))

### DIFF
--- a/src/datasources/errors/http-error-factory.ts
+++ b/src/datasources/errors/http-error-factory.ts
@@ -1,9 +1,7 @@
 import { HttpStatus, Injectable } from '@nestjs/common';
-import {
-  NetworkError,
-  NetworkResponseError,
-} from '@/datasources/network/entities/network.error.entity';
+import { NetworkResponseError } from '@/datasources/network/entities/network.error.entity';
 import { DataSourceError } from '@/domain/errors/data-source.error';
+import { isObject } from 'lodash';
 
 /**
  * Maps a {@link NetworkError} or {@link Error} into a {@link DataSourceError}
@@ -16,7 +14,7 @@ import { DataSourceError } from '@/domain/errors/data-source.error';
  */
 @Injectable()
 export class HttpErrorFactory {
-  from(source: NetworkError | Error): DataSourceError {
+  from(source: unknown): DataSourceError {
     if (isNetworkResponseError(source)) {
       const errorMessage: string = source.data?.message ?? 'An error occurred';
       return new DataSourceError(errorMessage, source.status);
@@ -29,8 +27,12 @@ export class HttpErrorFactory {
   }
 }
 
-function isNetworkResponseError(
-  error: NetworkError | Error,
-): error is NetworkResponseError {
-  return 'status' in error && error.status >= 400 && error.status < 600;
+function isNetworkResponseError(error: unknown): error is NetworkResponseError {
+  return (
+    isObject(error) &&
+    'status' in error &&
+    typeof error.status === 'number' &&
+    error.status >= 400 &&
+    error.status < 600
+  );
 }
--- a/src/domain/email/entities/create-email-message.dto.entity.ts
+++ b/src/domain/email/entities/create-email-message.dto.entity.ts
@@ -1,7 +1,7 @@
 export class CreateEmailMessageDto {
-  to: string[];
-  template: string;
-  subject: string;
-  substitutions: Record<string, unknown> | null;
+  to!: string[];
+  template!: string;
+  subject!: string;
+  substitutions!: Record<string, unknown> | null;
   emailMessageId?: string;
 }
--- a/src/logging/utils.spec.ts
+++ b/src/logging/utils.spec.ts
@@ -1,0 +1,54 @@
+import { asError } from '@/logging/utils';
+
+describe('asError', () => {
+  it('should return the same error if thrown is an instance of Error', () => {
+    const thrown = new Error('test error');
+
+    expect(asError(thrown)).toEqual(new Error('test error'));
+  });
+
+  it('should return the same error if thrown is a superset of Error', () => {
+    class CustomError extends Error {
+      constructor(message: string) {
+        super(message);
+      }
+    }
+    const thrown = new CustomError('test error');
+
+    expect(asError(thrown)).toEqual(new CustomError('test error'));
+  });
+
+  it('should return a new Error instance with the thrown value if thrown is a string', () => {
+    const thrown = 'test error';
+
+    const result = asError(thrown);
+    expect(result).toEqual(new Error('test error'));
+
+    // If stringified:
+    expect(result).not.toEqual(new Error('"test error'));
+  });
+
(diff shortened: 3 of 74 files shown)


src/datasources/errors/http-error-factory.ts
```diff
@@ -29,8 +27,12 @@ export class HttpErrorFactory {
...
-): error is NetworkResponseError {
-  return 'status' in error && error.status >= 400 && error.status < 600;
+function isNetworkResponseError(error: unknown): error is NetworkResponseError {
+  return (
+    isObject(error) &&
+    'status' in error &&
+    typeof error.status === 'number' &&
+    error.status >= 400 &&
+    error.status < 600
+  );
```
iamacook, line 37: This can be reduced when we [restructure the errors thrown in the network client](https://github.com/safe-global/safe-client-gateway/blob/dd79a201e2aed3031c2a862f4390cf9ffa1f0fd5/src/datasources/network/network.module.ts#L23).

src/logging/utils.ts
```diff
@@ -34,3 +36,23 @@ export function formatRouteLogMessage(
     detail: detail ?? null,
   };
 }
+
+export const asError = (thrown: unknown): Error => {
```
iamacook, line 40: This is a helper [copied over from the web](https://github.com/safe-global/safe-wallet-web/blob/946d56e1273317cdde33a324109aefd070b14827/src/services/exceptions/utils.ts#L1) for extracting the error `message` when in strict mode.

src/routes/data-decode/pipes/get-data-decoded.dto.validation.pipe.ts
```diff
@@ -27,7 +32,9 @@ export class GetDataDecodedDtoValidationPipe
     try {
       return this.genericValidator.validate(this.isValid, data);
     } catch (err) {
-      err.status = HttpStatus.BAD_REQUEST;
+      if (err instanceof HttpException) {
+        Object.assign(err, { status: HttpStatus.BAD_REQUEST });
+      }
```
iamacook, line 37: This is a temporary workaround as the way we are setting the `status` is incorrect (it is a readonly property). We should revisit the generic validator to throw `HttpStatus.BAD_REQUEST` directly.

fmrsabino (changes requested):

src/logging/utils.ts
iamacook: Added in fc58887c6782d4416d3ac7505c59a57ffa7e817c.

tsconfig.json
```diff
@@ -20,6 +20,7 @@
     "resolveJsonModule": true,
     "paths": {
       "@/*": ["src/*"]
-    }
+    },
+    "strict": true
```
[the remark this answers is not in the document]
hectorgomezv: The same applies to [noImplicitAny](https://www.typescriptlang.org/tsconfig#noImplicitAny) and [strictBindCallApply](https://www.typescriptlang.org/tsconfig#strictBindCallApply), right? The docs say they are 
transitively `true` if `strict` is true.

src/domain/email/entities/create-email-message.dto.entity.ts
```diff
@@ -1,7 +1,7 @@
 export class CreateEmailMessageDto {
-  to: string[];
-  template: string;
-  subject: string;
-  substitutions: Record<string, unknown> | null;
+  to!: string[];
```
hectorgomezv, line 2: I don't have a strong opinion regarding this, I've never faced this dilemma before, but adding the non-null assertion operator (`!`) to each property in each class seems a bit like a workaround 🤔 

I guess a more _permanent_ solution could add constructors to the classes and specify the access modifier on each property. That would also force us to review which classes can be converted to interfaces or types. 

Of course this wouldn't be covered in this PR, just food for thought. What do you think @iamacook @fmrsabino ?

hectorgomezv: Nevermind, I've just seen @fmrsabino comment here: https://github.com/safe-global/safe-client-gateway/pull/1051/files#r1463236193, I think it reflects the same concern 🙂

tsconfig.json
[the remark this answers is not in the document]
iamacook: I removed `strictNullChecks` in fc58887c6782d4416d3ac7505c59a57ffa7e817c and `strictBingCallApply` in 66be018be289c720ec6e3a03e59527d9cdc3e4ca. `noImplictAny` is a bit more complicated. I've created [a separate PR](https://github.com/safe-global/safe-client-gateway/pull/1057) for it as it requires other adjustments in the codebase.

src/domain/email/entities/create-email-message.dto.entity.ts
iamacook: I am impartial. I think we can add constructors later if we prefer?

For now, I think the assertion is the "quickest" solution and is [the recommendation of the NestJS author](https://github.com/nestjs/nest/issues/4178#issuecomment-614006219).

tsconfig.json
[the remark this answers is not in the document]
hectorgomezv: Thanks @iamacook, I've just seen your PR for removing `noImplicitAny`, will review that also! 👏🏻

src/domain/email/entities/create-email-message.dto.entity.ts
hectorgomezv: Sure, I agree the solution is sufficient for now. Later we can revisit this and think about the constructors or even migrating some classes to types/interfaces if it's feasible. All good from my side!

fmrsabino: > I am impartial. I think we can add constructors later if we prefer?
> 
> For now, I think the assertion is the "quickest" solution and is [the recommendation of the NestJS author](https://github.com/nestjs/nest/issues/4178#issuecomment-614006219).

Yes. I'd add these in a separate PR.

src/logging/utils.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+  });
+
+  it('should return a new Error instance with the message thrown value if thrown is object with message', () => {
+    const thrown = { message: 'test error' };
+
+    const result = asError(thrown);
+    expect(result).toEqual(new Error('test error'));
+
+    // If stringified:
+    expect(result).not.toEqual(new Error('{"message":"test error"}'));
```
fmrsabino, line 38: What is the reason to test the equality in one assertion and testing a different one (non-equality) in the following assertion?

fmrsabino (approved):
👏 🚀

src/logging/utils.spec.ts
iamacook: I wanted to make sure that the `message` was used and not the object.

PR #1057, opened 2024-01-23
Enable `noImplicitAny` in TypeScript

iamacook:
Relies on #1051

This enables the [`noImplicitAny` flag](https://www.typescriptlang.org/tsconfig#noImplicitAny) in TypeScript, preventing the compiler from falling back to a type of `any` for a variable when it cannot be inferred.

All non-defined types have, as such, been defined and associated errors fixed in accordance. (It predominantly found issues with our tests.)


src/routes/common/decorators/utils.spec.ts
```diff
@@ -1,13 +1,14 @@
...
+import { Request } from 'express';
 
 describe('utils tests', () => {
   describe('getRouteUrl tests', () => {
     const request = {
       get: jest.fn(),
       originalUrl: faker.system.filePath(),
       protocol: faker.internet.protocol(),
-    } as unknown as any;
+    } as unknown as jest.MockedObjectDeep<Request>;
```
hectorgomezv, line 11:
```suggestion
    } as jest.MockedObjectDeep<Request>;
```

In fact, I think we could revisit some `as unknown` casts we are doing through the code (I can tackle this in a separate PR) 🙂

hectorgomezv:
Great job! 👏🏻 LGTM, I'll approve it after the conflicts are resolved 🙂

src/routes/common/decorators/utils.spec.ts
iamacook: Iremoved this particular instance in 771fef093993851cd5fcdf1b3161d3ac4af4fb95 but we definitely should tackle the others.

hectorgomezv (approved):

PR #1060, opened 2024-01-23
Prevent explicit use of `any`

iamacook:
Whilst #1068 prevents an implied `any`, this enabled the [`no-explicit-any` lint rule](https://typescript-eslint.io/rules/no-explicit-any/) which prevents explicit usage of `any` to increase type safety. ("`any` is a dangerous "escape hatch" from the type system.")


src/app.provider.ts
```diff
@@ -1,8 +1,13 @@
...
+import {
+  DynamicModule,
+  INestApplication,
+  VersioningType,
+} from '@nestjs/common';
 import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
 import { NestFactory } from '@nestjs/core';
 import { IConfigurationService } from '@/config/configuration.service.interface';
 import { json } from 'express';
+import { TestingModule } from '@nestjs/testing';
```
fmrsabino, line 10: This dependency is not available in the final image. We should avoid having references to testing utilities in production code.

fmrsabino (changes requested):

src/app.provider.ts
```diff
@@ -66,13 +71,17 @@ export abstract class AppProvider {
     app: INestApplication,
   ) => void)[];
 
-  public async provide(module: any): Promise<INestApplication> {
+  public async provide(
+    module: DynamicModule | TestingModule,
```
fmrsabino, line 75: I would not restrict the module to be a `DynamicModule` here.

Given that NestJS expected `any` in `NestFactory.create` we can have a generic type instead.


```typescript
export abstract class AppProvider<T> {
  ...
  public async provide(module: T): Promise<INestApplication> {
    ...
  }

  protected abstract getApp(module: T): Promise<INestApplication>;
}
```

src/datasources/errors/http-error-factory.spec.ts
```diff
@@ -24,7 +24,9 @@ describe('HttpErrorFactory', () => {
 
     expect(actual instanceof DataSourceError).toBe(true);
```
fmrsabino, line 25: Here we know that `actual` is a `DataSourceError` so we can avoid the `instanceof` operator

src/routes/data-decode/entities/data-decoded-parameter.entity.ts
```diff
@@ -8,13 +8,19 @@ export class DataDecodedParameter {
   @ApiProperty()
   value: unknown;
   @ApiProperty()
-  valueDecoded?: Record<string, any> | Record<string, any>[] | undefined;
+  valueDecoded?:
+    | Record<string, unknown>
+    | Record<string, unknown>[]
```
fmrsabino, line 13: OOC how are these serialised in the Swagger interface?

fmrsabino (changes requested):

src/app.provider.ts
iamacook: Good idea! I've adjusted it to use generics in 646f900c7186f2e596033fe23ebb9e1a8202c7bf.

src/datasources/errors/http-error-factory.spec.ts
iamacook: I removed all `instanceof `checks in 646f900c7186f2e596033fe23ebb9e1a8202c7bf.

src/routes/data-decode/entities/data-decoded-parameter.entity.ts
iamacook: Both before and after the changes in this PR, it is shown as `{}`.

src/__tests__/test-app.provider.ts
```diff
@@ -27,11 +27,12 @@ export class TestAppProvider extends AppProvider {
...
 
-  protected getApp(module: any): Promise<INestApplication> {
-    if (!(module instanceof TestingModule))
-      return Promise.reject(
-        `${module.constructor.name} is not a TestingModule`,
-      );
+  protected getApp(module: T): Promise<INestApplication> {
+    if (!(module instanceof TestingModule)) {
+      const name =
+        typeof module === 'function' ? module.constructor.name : typeof module;
```
iamacook, line 33: This change was necessary due to type restrictions.

iamacook: Changed to remove second `typeof` in c6b50ff83e8d969c24f62d7a68addb42f3288aa3.

src/__tests__/test-app.provider.ts
```diff
@@ -27,13 +27,12 @@ export class TestAppProvider extends AppProvider {
...
-    module: DynamicModule | TestingModule,
-  ): Promise<INestApplication> {
-    if (!(module instanceof TestingModule))
-      return Promise.reject(
-        `${module.constructor.name} is not a TestingModule`,
-      );
+  protected getApp(module: T): Promise<INestApplication> {
+    if (!(module instanceof TestingModule)) {
+      const name =
+        typeof module === 'function' ? module.constructor.name : module;
```
fmrsabino, line 33: What is the goal of this change? module might be an object for example. This would print

```
[object Object] is not a TestingModule
```

iamacook: We have to satisfy TypeScript. As [we are now using generics](https://github.com/safe-global/safe-client-gateway/pull/1060#discussion_r1469380455), `module` is `unknown`.

It's theoretically a `DynamicModule` so we could do something like this but I'd welcome your suggestion:

```ts
if (!(module instanceof TestingModule)) {
  const message = module instanceof DynamicModule
    ? `${module.constructor.name} is not a TestingModule`
    : 'Not a TestingModule'
  return Promise.reject(message);
}
return Promise.resolve(module.createNestApplication());
```

fmrsabino: I think this introduces complexity for an unlikely use-case that should only happen in a test environment. I'd say to keep it simple for now and reject the promise with something like "Provided module is not a TestingModule". Wdyt?

iamacook: Agreed. I simplified it in 9e246808069e9a68cd3f8d1eaa2ab20cce6c3e64.

fmrsabino (approved):
👏 🚀

PR #1062, opened 2024-01-23
Add Valk Balances provider

hectorgomezv:
### Summary
This PR adds a new Ethereum address balance provider (Valk). This provider is disabled by default. To use this provider on a specific chain, the chain ID should be included in the feature flag environment variable `FF_VALK_BALANCES_CHAIN_IDS`. 

This PR also introduces the concept of `BalancesApiManager`. This class acts as a common router/façade for the available balance providers (starting with Valk, but potentially opening the door for more providers). 

```mermaid
graph TD;
    BalancesRepository-->BalancesApiManager;
    BalancesApiManager-->BalancesApi-1;
    BalancesApiManager-->BalancesApi-2;
    BalancesApiManager-->BalancesApi-N;
```

### Changes
* Three new environment variables are introduced:
  - `VALK_BASE_URI`: contains the base URI for the Valk API. Has a default value in the configuration.
  - `VALK_API_KEY`: contains the API key for the Valk API. Defaults to `undefined`. This is done on purpose as it shouldn't be needed to provide an API key for each potential provider the CGW uses (moreover considering the provider could be not managing any chain at a given point in time). 
  - `FF_VALK_BALANCES_CHAIN_IDS`: as stated above, this manages the chains that are associated with the Valk API for balance retrieval. The expected values are comma-separated numbers (e.g.: `1,100,137`)
* Valk-specific configuration is added to `configuration.ts` under the key `balances.providers.valk`.
* `BalancesApiManager` is added along with its tests.
* `BalancesApiModule` has been added with only one `BalancesApi` implementation so far.
* Valk-specific entities, and `ValkBalancesApi` (with tests) have been added.
* `BalancesController` tests have been extended to test the new integration.

### Out of scope
- `GET /v1/balances/supported-fiat-codes` endpoint adjustments will be done in a separate PR.
- The current balances provider (Transaction Service + CoinGecko) could be considered a `BalancesApi` implementation and could be included in the `BalancesApiManager` routing. This was left out of the scope intentionally to avoid doing too many (potentially transient) changes. It could be done in a separate PR if required.

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -24,7 +24,7 @@ export default () => ({
       valk: {
         baseUri:
           process.env.VALK_BASE_URI ||
-          'https://merlin-api-v1.cf/api/merlin/public/',
+          'https://merlin-api-v1.cf/api/merlin/public',
         apiKey: process.env.VALK_API_KEY,
         chains: {
           1: { chainName: 'eth' },
@@ -97,7 +97,8 @@ export default () => ({
   features: {
     richFragments: process.env.FF_RICH_FRAGMENTS?.toLowerCase() === 'true',
     email: process.env.FF_EMAIL?.toLowerCase() === 'true',
-    valkBalancesChainIds: process.env.VALK_BALANCES_CHAIN_IDS?.split(',') ?? [],
+    valkBalancesChainIds:
+      process.env.FF_VALK_BALANCES_CHAIN_IDS?.split(',') ?? [],
   },
   httpClient: {
     // Timeout in milliseconds to be used for the HTTP client.
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -18,6 +18,31 @@ export default () => ({
   auth: {
     token: process.env.AUTH_TOKEN,
   },
+  balances: {
+    balancesTtlSeconds: parseInt(process.env.BALANCES_TTL_SECONDS ?? `${300}`),
+    providers: {
+      valk: {
+        baseUri:
+          process.env.VALK_BASE_URI ||
+          'https://merlin-api-v1.cf/api/merlin/public/',
+        apiKey: process.env.VALK_API_KEY,
+        chains: {
+          1: { chainName: 'eth' },
+          10: { chainName: 'op' },
+          100: { chainName: 'xdai' },
+          1101: { chainName: 'pze' },
+          1313161554: { chainName: 'aurora' },
+          137: { chainName: 'matic' },
+          324: { chainName: 'era' },
+          42161: { chainName: 'arb' },
+          42220: { chainName: 'celo' },
+          43114: { chainName: 'avax' },
+          56: { chainName: 'bsc' },
+          8453: { chainName: 'base' },
+        },
+      },
+    },
+  },
   db: {
     postgres: {
       host: process.env.POSTGRES_HOST || 'localhost',
@@ -72,6 +97,7 @@ export default () => ({
   features: {
     richFragments: process.env.FF_RICH_FRAGMENTS?.toLowerCase() === 'true',
     email: process.env.FF_EMAIL?.toLowerCase() === 'true',
+    valkBalancesChainIds: process.env.VALK_BALANCES_CHAIN_IDS?.split(',') ?? [],
   },
   httpClient: {
     // Timeout in milliseconds to be used for the HTTP client.
--- a/src/datasources/balances-api/balances-api.manager.ts
+++ b/src/datasources/balances-api/balances-api.manager.ts
@@ -20,7 +20,7 @@ export class BalancesApiManager implements IBalancesApiManager {
     this.valkBalancesApi = valkBalancesApi;
   }
 
-  isExternalized(chainId: string): boolean {
+  useExternalApi(chainId: string): boolean {
     const allProvidersChainIds = this.valkBalancesChainIds;
     return allProvidersChainIds.includes(chainId);
   }
--- a/src/datasources/balances-api/balances-api.manager.ts
+++ b/src/datasources/balances-api/balances-api.manager.ts
@@ -25,9 +25,13 @@ export class BalancesApiManager implements IBalancesApiManager {
   }
 
   getBalancesApi(chainId: string): IBalancesApi {
-    if (this.valkBalancesChainIds.includes(chainId)) {
+    if (this._isSupportedByValk(chainId)) {
       return this.valkBalancesApi;
     }
     throw new Error(`Chain ID ${chainId} balances provider is not configured`);
+  }
+
+  private _isSupportedByValk(chainId: string): boolean {
+    return this.valkBalancesChainIds.includes(chainId);
   }
 }
--- a/src/datasources/balances-api/balances-api.manager.ts
+++ b/src/datasources/balances-api/balances-api.manager.ts
@@ -1,0 +1,34 @@
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { IValkBalancesApi } from '@/datasources/balances-api/valk-balances-api.service';
+import { IBalancesApi } from '@/domain/interfaces/balances-api.interface';
+import { IBalancesApiManager } from '@/domain/interfaces/balances-api.manager.interface';
+import { Inject, Injectable } from '@nestjs/common';
+
+@Injectable()
+export class BalancesApiManager implements IBalancesApiManager {
+  private readonly valkBalancesChainIds: string[];
+  private readonly valkBalancesApi: IBalancesApi;
+
+  constructor(
+    @Inject(IConfigurationService)
+    private readonly configurationService: IConfigurationService,
+    @Inject(IValkBalancesApi) valkBalancesApi: IBalancesApi,
+  ) {
+    this.valkBalancesChainIds = this.configurationService.getOrThrow<string[]>(
+      'features.valkBalancesChainIds',
+    );
+    this.valkBalancesApi = valkBalancesApi;
+  }
+
+  isExternalized(chainId: string): boolean {
+    const allProvidersChainIds = this.valkBalancesChainIds;
+    return allProvidersChainIds.includes(chainId);
+  }
+
+  getBalancesApi(chainId: string): IBalancesApi {
+    if (this.valkBalancesChainIds.includes(chainId)) {
+      return this.valkBalancesApi;
+    }
+    throw new Error(`Chain ID ${chainId} balances provider is not configured`);
+  }
+}
--- a/src/datasources/balances-api/balances-api.manager.ts
+++ b/src/datasources/balances-api/balances-api.manager.ts
@@ -21,8 +21,7 @@ export class BalancesApiManager implements IBalancesApiManager {
   }
 
   useExternalApi(chainId: string): boolean {
-    const allProvidersChainIds = this.valkBalancesChainIds;
-    return allProvidersChainIds.includes(chainId);
+    return this.valkBalancesChainIds.includes(chainId);
   }
 
   getBalancesApi(chainId: string): IBalancesApi {
--- a/src/datasources/balances-api/valk-balances-api.service.ts
+++ b/src/datasources/balances-api/valk-balances-api.service.ts
@@ -109,7 +109,7 @@ export class ValkBalancesApi implements IBalancesApi {
           : this._mapNativeBalance(valkBalance)),
         fiatBalance,
         fiatConversion,
-      } as Balance;
+      };
     });
   }
 
--- a/src/datasources/balances-api/valk-balances-api.service.ts
+++ b/src/datasources/balances-api/valk-balances-api.service.ts
@@ -1,0 +1,15 @@
+import { Balance } from '@/domain/balances/entities/balance.entity';
+import { IBalancesApi } from '@/domain/interfaces/balances-api.interface';
+import { Injectable } from '@nestjs/common';
+
+export const IValkBalancesApi = Symbol('IValkBalancesApi');
+
+@Injectable()
+export class ValkBalancesApi implements IBalancesApi {
+  getBalances(): Promise<Balance[]> {
+    throw new Error('Method not implemented.');
+  }
+  clearBalances(): Promise<void> {
+    throw new Error('Method not implemented.');
+  }
+}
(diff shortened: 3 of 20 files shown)


src/domain/balances/balances.repository.ts
```diff
@@ -66,16 +110,6 @@ export class BalancesRepository implements IBalancesRepository {
     );
   }
 
-  async clearLocalBalances(args: {
```
hectorgomezv, line 69: This function was moved upward to keep class public functions/interface members first, and private/utility functions later.

src/domain/balances/entities/schemas/balance.schema.ts
```diff
@@ -26,6 +26,8 @@ const balanceSchema: Schema = {
     tokenAddress: { type: 'string', nullable: true, default: null },
     token: { anyOf: [{ type: 'null' }, { $ref: 'balance-token.json' }] },
     balance: { type: 'string' },
+    fiatBalance: { type: 'string', nullable: true, default: null },
+    fiatConversion: { type: 'string', nullable: true, default: null },
```
hectorgomezv, line 30: These properties were present in the domain entity but weren't included in the schema. This is probably a leftover of the previous implementation (pre-Coingecko) where these were calculated at the route level.

src/routes/balances/balances.service.ts
```diff
@@ -81,6 +81,7 @@ export class BalancesService {
   }
 
   async getSupportedFiatCodes(): Promise<string[]> {
+    // TODO: take into consideration BalancesRepository available fiat codes.
```
hectorgomezv, line 84: I left a TODO here as I think this can be done in a separate PR to avoid creating too much noise.

src/domain/balances/entities/balance.entity.ts
```diff
@@ -1,12 +1,12 @@
 import { BalanceToken } from '@/domain/balances/entities/balance.token.entity';
 
-interface NativeBalance {
+export interface NativeBalance {
```
hectorgomezv, line 3: I had to export these two interfaces to avoid transpiler errors by explicitly state the return types in `valk-balances-api.service.ts`.

src/datasources/cache/cache.router.ts
```diff
@@ -48,6 +49,20 @@ export class CacheRouter {
...
+  }): string {
+    return `${args.chainId}_${CacheRouter.VALK_BALANCES_KEY}_${args.safeAddress}`;
+  }
+
+  static getValkBalancesCacheDir(args: {
+    chainId: string;
+    safeAddress: string;
+  }): CacheDir {
+    return new CacheDir(CacheRouter.getValkBalancesCacheKey(args), '');
+  }
```
hectorgomezv, line 64: I think having separated cache prefixes by provider is the way to go. Otherwise, the balances would be cached regardless of the provider, so changing the configuration of a chain could make an integration process the response of a different provider and therefore erroring for some time (until it expired). 

Once the migration is complete and only one provider is used, the provider prefix can be removed.

src/datasources/balances-api/valk-balances-api.service.ts
```diff
@@ -0,0 +1,143 @@
...
+          tokenAddress: null,
+          token: null,
+          balance,
+          fiatBalance,
+          fiatConversion,
+        } as NativeBalance & {
+          fiatBalance: string | null;
+          fiatConversion: string | null;
+        };
+      }
```
iamacook, line 123: Nit: I'd separate the `Erc20Balance`/`NativeBalance` objects into their own variables then cast to `Balance` with the required values for better type safety.

```ts
const erc20Balance: Erc20Balance = {
  ...
}
return {
  ...erc20Balance,
  ...
} as Balance
```

src/datasources/cache/cache.router.ts
iamacook: Let's add a TODO to remove the prefix.

src/datasources/balances-api/balances-api.manager.ts
```diff
@@ -0,0 +1,34 @@
...
+    private readonly configurationService: IConfigurationService,
+    @Inject(IValkBalancesApi) valkBalancesApi: IBalancesApi,
+  ) {
+    this.valkBalancesChainIds = this.configurationService.getOrThrow<string[]>(
+      'features.valkBalancesChainIds',
+    );
+    this.valkBalancesApi = valkBalancesApi;
+  }
+
+  isExternalized(chainId: string): boolean {
```
iamacook, line 23: Nit: I think we could improve the name a bit, e.g. something like `useProvider`, `useExternalApi`.

src/datasources/balances-api/valk-balances-api.service.ts
```diff
@@ -0,0 +1,143 @@
...
+  getChainName(chainId: string): string {
+    const chainName = this.chainsConfiguration[Number(chainId)]?.chainName;
+    if (!chainName)
+      throw Error(
+        `Chain ${chainId} balances retrieval via Valk is not configured`,
+      );
+    return chainName;
+  }
+
+  async clearBalances(args: {
```
iamacook, line 136: Nit: I'd try to be declarative here with the namingg as well, e.g. `clearExternalBalancesApi`.

src/domain/interfaces/balances-api.manager.interface.ts
```diff
@@ -0,0 +1,24 @@
+import { IBalancesApi } from '@/domain/interfaces/balances-api.interface';
+
+export const IBalancesApiManager = Symbol('IBalancesApiManager');
+
+export interface IBalancesApiManager {
+  /**
+   * Determines whether an external balances provider is being used
```
iamacook, line 7:
```suggestion
   * Determines whether an external balances provider is used
```

src/domain/interfaces/balances-api.manager.interface.ts
```diff
@@ -0,0 +1,24 @@
...
+   * for a given chain ID.
+   *
+   * @param chainId - the chain identifier to check.
+   * @returns true if the chain balances are externalized.
```
iamacook, line 11:
```suggestion
   * @returns true if an external API is used for the chain balances.
```

src/domain/interfaces/balances-api.manager.interface.ts
```diff
@@ -0,0 +1,24 @@
...
+   */
+  isExternalized(chainId: string): boolean;
+
+  /**
+   * Gets an {@link IBalancesApi} implementation.
+   * Each chain is associated to an implementation (i.e.: to a balances
```
iamacook, line 17:
```suggestion
   * Each chain is associated with an implementation (i.e.: to a balances
```

src/config/entities/configuration.ts
```diff
@@ -18,6 +18,31 @@ export default () => ({
...
+          100: { chainName: 'xdai' },
+          1101: { chainName: 'pze' },
+          1313161554: { chainName: 'aurora' },
+          137: { chainName: 'matic' },
+          324: { chainName: 'era' },
+          42161: { chainName: 'arb' },
+          42220: { chainName: 'celo' },
+          43114: { chainName: 'avax' },
+          56: { chainName: 'bsc' },
+          8453: { chainName: 'base' },
```
iamacook, line 41: Nit: are these defined by Valk as `chainName`s? They look like `shortName`s to me.

hectorgomezv: I don't have an opinion here, `chainName` was chosen in order to align it with the existing `prices.chains` configuration. Valk refers to these values as `id` in their documentation. The values can be seen as a somewhat counterpart of our ERC-3770 `shortName` that we have in the Config Service, but there are differences in the values, I don't think they are following that convention. 

So as I said, no opinion about the property name, we could rename it to `shortName` both here and into `prices.chains` in a separate PR, but that wouldn't be very strict because as I commented none of them is following ERC-3770's `shortName` concept.

src/datasources/cache/cache.router.ts
hectorgomezv: Added in [4f5723f](https://github.com/safe-global/safe-client-gateway/pull/1062/commits/4f5723f06f49a21b090547d02f90155e3a297cd2)

src/datasources/balances-api/balances-api.manager.ts
```diff
@@ -0,0 +1,34 @@
...
+    const allProvidersChainIds = this.valkBalancesChainIds;
+    return allProvidersChainIds.includes(chainId);
+  }
+
+  getBalancesApi(chainId: string): IBalancesApi {
+    if (this.valkBalancesChainIds.includes(chainId)) {
```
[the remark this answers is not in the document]
hectorgomezv: My intention here was to implement something like a _router_ so the execution would go through each provider and check if the `chainId` is included in the chain IDs the provider manages. At this point, as only one provider is available I agree we could reuse the other function in the interface. Still, it may be clearer to check the provider IDs explicitly to understand the purpose of the function (and also keeping in mind that the cost is the same), wdyt?

src/datasources/balances-api/valk-balances-api.service.ts
hectorgomezv: Hmm, I guess that would also imply to rename `getBalances` to `getExternalBalances`. No strong opinion, but from a domain-side point of view (as this interface serves as API for the domain), I'd say we should abstract the precise origin of the balances (external provider vs. Transaction Service). And also, eventually, all the balances could become "external" so no distinction should be necessary. Thoughts @iamacook @fmrsabino? 🙂

src/datasources/balances-api/balances-api.manager.ts
hectorgomezv: I agree, changed to `useExternalApi` in [4f5723f](https://github.com/safe-global/safe-client-gateway/pull/1062/commits/4f5723f06f49a21b090547d02f90155e3a297cd2)

src/datasources/balances-api/valk-balances-api.service.ts
hectorgomezv: I think I found an implementation that is simpler than the one I initially did (just one return point) and also rolls back the need to export `Erc20Balance` and `NativeBalance` (increasing encapsulation). You can check it out in 

If you find this implementation a bit convoluted or you think it provides less type safety, I'd be happy to implement it as you proposed! 🙂

hectorgomezv: Sorry for introducing more confusion, but I also moved up the public function (`getBalances`, `clearBalances`) and I made explicitly private the private ones (including the `_` prefix) :) [eff163c](https://github.com/safe-global/safe-client-gateway/pull/1062/commits/eff163cb9d44f9d482eb5300b4d0024009c0cf57)

src/datasources/balances-api/balances-api.manager.ts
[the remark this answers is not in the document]
iamacook: Makes sense to me - you need not include my suggestion. We should instead add a comment or include provider-specific helpers, e.g. `isSupportByValk` to be more declarative.

src/datasources/balances-api/valk-balances-api.service.ts
iamacook: Fair point. As this is in a provider-specific service then let's it as is.

src/datasources/balances-api/valk-balances-api.service.ts
```diff
@@ -0,0 +1,134 @@
...
+      const balanceAmount = getNumberString(valkBalance.balance);
+      const fiatBalance = getNumberString(
+        (valkBalance.balance / Math.pow(10, valkBalance.decimals)) * price,
+      );
+      const fiatConversion = getNumberString(price);
+
+      // Valk returns a string representing the native coin (e.g.: 'eth') as
+      // token_address for native coins balances. An hex is returned for ERC20 tokens.
+      return {
+        ...(isHex(valkBalance.token_address)
```
iamacook, line 104: Let's use `isAddress` here to make sure it's not `0x123`, for example.

src/datasources/balances-api/valk-balances-api.service.ts
```diff
@@ -0,0 +1,134 @@
...
+                logoUri: valkBalance.logo ?? '',
+              },
+              balance: balanceAmount,
+            }
+          : {
+              tokenAddress: null,
+              token: null,
+              balance: balanceAmount,
+            }),
+        fiatBalance,
```
iamacook, line 120: You can move `balance` out of the ternary.

```suggestion
            }
          : {
              tokenAddress: null,
              token: null,
            }),
        balance: balanceAmount,
        fiatBalance,
```

---

The below is a nit and not necessary so I'll leave the decision up to you.

If you extracted the mapping, you could be more explicit with the return types, e.g. something like:

```ts
mapErc20Balance(valkBalance: ValkBalance): Erc20Balance
mapNativeBalance(valkBalance: ValkBalance): NativeBalance
mapValkBalance(valkBalance: ValkBalance): Balance {
  return {
    ...isAddress(valkBalance.token_address) ? mapErc20Balance(valkBalance) : mapNativeBalance(valkBalance)
    fiatBalance,
    fiatConversion
  }
}
```

src/datasources/balances-api/valk-balances-api.service.ts
hectorgomezv: Changed in [9ec50af](https://github.com/safe-global/safe-client-gateway/pull/1062/commits/9ec50afe2fbb0b467535f4121b8191e53db9e10e), thanks!

src/datasources/balances-api/valk-balances-api.service.ts
hectorgomezv: Good suggestion! I've adopted this mapping functions approach 🙂 in [9ec50af](https://github.com/safe-global/safe-client-gateway/pull/1062/commits/9ec50afe2fbb0b467535f4121b8191e53db9e10e)

src/datasources/balances-api/valk-balances-api.service.ts
```diff
@@ -0,0 +1,145 @@
...
+
+      // Valk returns a string representing the native coin (e.g.: 'eth') as token_address
+      // for native coins balances. An Ethereum address is returned for ERC20 tokens.
+      return {
+        ...(isAddress(valkBalance.token_address)
+          ? this._mapErc20Balance(valkBalance)
+          : this._mapNativeBalance(valkBalance)),
+        fiatBalance,
+        fiatConversion,
+      } as Balance;
```
iamacook, line 112: With the mappers we no longer need cast:

```suggestion
      };
```

iamacook (approved):

PR #1064, opened 2024-01-24
Refactor trusted filtering

iamacook:
As per [suggestion](https://github.com/safe-global/safe-client-gateway/pull/1061#discussion_r1464091362), this refactors the filtering of trusted transactions to be simpler:

#### Changes
- `for` has been changed to an initial fetching of nested transactions, then `filter`ing
- Mappers that are only used as flags have been adjusted to return `boolean`s
  - `mapZeroValueTransfer` -> `transferHasValue`
  - `mapTrustedTransfer` -> `isTrustedTransfer`

### DIFF
--- a/src/routes/transactions/mappers/transactions-history.mapper.ts
+++ b/src/routes/transactions/mappers/transactions-history.mapper.ts
@@ -188,27 +188,21 @@ export class TransactionsHistoryMapper {
     onlyTrusted: boolean,
   ): Promise<TransactionItem[]> {
     const limitedTransfers = transfers.slice(0, this.maxNestedTransfers);
-    const result: TransactionItem[] = [];
 
-    for (const transfer of limitedTransfers) {
-      const nestedTransaction = await this.transferMapper.mapTransfer(
-        chainId,
-        transfer,
-        safe,
-      );
+    const nestedTransactions = await Promise.all(
+      limitedTransfers.map((transfer) =>
+        this.transferMapper.mapTransfer(chainId, transfer, safe),
+      ),
+    );
 
-      const transferWithValue = this.mapZeroValueTransfer(nestedTransaction);
-      // If we do not have a transfer with value, we do not add it to the result
-      if (!transferWithValue) continue;
+    return nestedTransactions
+      .filter((nestedTransaction) => {
+        // If we do not have a transfer with value, we do not add it to the result
+        if (!this.transferHasValue(nestedTransaction)) return false;
 
-      const trustedTransfer = onlyTrusted
-        ? this.mapTrustedTransfer(transferWithValue)
-        : transferWithValue;
-
-      if (!trustedTransfer) continue;
-      result.push(new TransactionItem(nestedTransaction));
-    }
-    return result;
+        return !onlyTrusted || this.isTrustedTransfer(nestedTransaction);
+      })
+      .map((nestedTransaction) => new TransactionItem(nestedTransaction));
   }
 
   /**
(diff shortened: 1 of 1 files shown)


src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -188,27 +188,21 @@ export class TransactionsHistoryMapper {
...
+    return nestedTransactions
+      .filter((nestedTransaction) => {
+        // If we do not have a transfer with value, we do not add it to the result
+        if (!this.mapZeroValueTransfer(nestedTransaction)) return false;
 
-      if (!trustedTransfer) continue;
-      result.push(new TransactionItem(nestedTransaction));
-    }
-    return result;
+        return !onlyTrusted || this.mapTrustedTransfer(nestedTransaction);
```
fmrsabino, line 203: I find this line here a but confusing: the `filter` expects a predicate to filter the elements present in the collection.

However we are using the _truthy_ value with `this.mapTrustedTransfer(nestedTransaction)` so we might end with the following filter results:

- `true` – if we want to include all transactions.
- `false || Transaction` – if we want to include only trusted transactions – this results in `Transaction` being returned in the predicate.
- `false || null` – if we want to include only trusted transactions – this results in `null` being returned from the predicate.

I'd therefore suggest `isX` where `isX` returns `boolean` instead of `mapX` since using a `map` inside of a `filter` hints to some side-effects.

fmrsabino (changes requested):

src/routes/transactions/mappers/transactions-history.mapper.ts
iamacook: > I'd therefore suggest `isX` where `isX` returns `boolean` instead of `mapX` since using a `map` inside of a `filter` hints to some side-effects.

Changed accordingly in d6e0feb49a00dafc31d09bb040f2f1bc883fe066.

src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -188,54 +188,41 @@ export class TransactionsHistoryMapper {
...
-   * Returns Null otherwise.
+   * Returns true if it is an ERC20 transfer with value.
+   * Returns false otherwise.
    *
    * @private
    */
-  private mapZeroValueTransfer(transaction: Transaction): Transaction | null {
-    if (!isTransferTransactionInfo(transaction.txInfo)) return transaction;
-    if (!isErc20Transfer(transaction.txInfo.transferInfo)) return transaction;
+  private isTransferWithVaklue(transaction: Transaction): boolean {
```
fmrsabino, line 214: typo: `isTransferWithValue`

src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -188,54 +188,41 @@ export class TransactionsHistoryMapper {
...
+      .filter((nestedTransaction): boolean => {
+        // If we do not have a transfer with value, we do not add it to the result
+        if (!this.isTransferWithVaklue(nestedTransaction)) return false;
 
-      if (!trustedTransfer) continue;
-      result.push(new TransactionItem(nestedTransaction));
-    }
-    return result;
+        return !onlyTrusted || this.isTrustedTransfer(nestedTransaction);
+      })
```
fmrsabino, line 204: nitpick:

```typescript
.filter((nestedTransaction): boolean => {
  // We are interested in transfers that:
  // - Have value and:
  // - If onlyTrusted is true then it should be a trusted transfer
  // - If onlyTrusted is false then any transfer is valid
  return (
    this.isTransferWithValue(nestedTransaction) &&
    (!onlyTrusted || this.isTrustedTransfer(nestedTransaction))
  );
})
```

src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -188,54 +188,41 @@ export class TransactionsHistoryMapper {
...
-  private mapZeroValueTransfer(transaction: Transaction): Transaction | null {
-    if (!isTransferTransactionInfo(transaction.txInfo)) return transaction;
-    if (!isErc20Transfer(transaction.txInfo.transferInfo)) return transaction;
+  private isTransferWithVaklue(transaction: Transaction): boolean {
+    if (!isTransferTransactionInfo(transaction.txInfo)) return true;
+    if (!isErc20Transfer(transaction.txInfo.transferInfo)) return true;
 
-    if (transaction.txInfo.transferInfo.value === '0') return null;
-    return transaction;
+    return transaction.txInfo.transferInfo.value !== '0';
```
fmrsabino, line 218: This might be a bug (even in the previous version) as a value of `'00'` should probably be considered as zero too.

fmrsabino (changes requested):

src/routes/transactions/mappers/transactions-history.mapper.ts
iamacook: Fixed in 15489f36e5464501c0356a4652eae757dd04c396.

src/routes/transactions/mappers/transactions-history.mapper.ts
iamacook: Included in 15489f36e5464501c0356a4652eae757dd04c396.

src/routes/transactions/mappers/transactions-history.mapper.ts
iamacook: I've changed it to compare the numerical value in 15489f36e5464501c0356a4652eae757dd04c396.

fmrsabino (approved):
🚀

PR #1065, opened 2024-01-24
Remove unnecessary `expect` in test

iamacook:
This removes an unnecessary `expect` in the `asError` test as it is checked prior as per https://github.com/safe-global/safe-client-gateway/pull/1051#discussion_r1464590239.

### DIFF
--- a/src/logging/utils.spec.ts
+++ b/src/logging/utils.spec.ts
@@ -33,9 +33,6 @@ describe('asError', () => {
 
     const result = asError(thrown);
     expect(result).toEqual(new Error('test error'));
-
-    // If stringified:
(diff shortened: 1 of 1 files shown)


src/logging/utils.spec.ts
```diff
@@ -33,9 +33,6 @@ describe('asError', () => {
 
     const result = asError(thrown);
     expect(result).toEqual(new Error('test error'));
-
-    // If stringified:
-    expect(result).not.toEqual(new Error('{"message":"test error"}'));
```
fmrsabino, line 38: What about the test `should return a new Error instance with the thrown value if thrown is a string`? Shouldn't that additional assertion be removed also?

fmrsabino (approved):
🚀

PR #1075, opened 2024-01-26
Migrate from `master-copies` to `singletons` on `TransactionApi`

### LINKED ISSUE #1071: Migrate from `master-copies` to `singletons` endpoint
The `/v1/about/master-copies/` endpoint is marked as deprecated on the Transaction Service, noting that `/v1/about/singletons/` should be used instead. We should migrate in accordance with this.

The return types of both endpoints are the same so we can switch the URL without causing disruptions to clients.

iamacook:
Closes #1071

The `master-copies` endpoint of the Transaction Service has been deprecated, replaced by `singletons`. Both of which have the same return type. This migrates to the `singletons` endpoint.

### Changes

- `TransactionApi['getMasterCopies']` -> `getSingletons` (using new endpoint)
- `MasterCopy` domain entity -> `Singleton`
- `masterCopySchema` -> `singletonSchema`
- `MasterCopyValidator` -> `SingletonValidator`
- `CacheRouter['getMasterCopiesCacheDir]'` -> `getSingletonsCacheDir` (using ne key `master_copies` -> `singletons`)
- `ChainsRepository['getMasterCopies']` -> `getSingletons`
- `masterCopyBuilder` -> `singletonBuilder`

Associated usage and test coverage of the above has also been renamed accordingly.


fmrsabino (approved):
👏 🚀

PR #1076, opened 2024-01-29
Set invalidation time of keys deleted via pattern

iamacook:
This adjusts `RedisCacheService['deleteByKeyPattern']` to set the `invalidationTimeMs:*` of each key deleted as they were not being set:

### Changes

- New `setInvalidationTime` method extracted from and called in `deleteByKey`.
- Said method called in `deleteByKeyPattern` with non-prefixed keys (from new `_removePrefixFromKey` method).

Test coverage has been added for the above in the same manner as that testing the invalidation times set in `deleteByKey`.

### DIFF
--- a/src/datasources/cache/redis.cache.service.spec.ts
+++ b/src/datasources/cache/redis.cache.service.spec.ts
@@ -136,7 +136,9 @@ describe('RedisCacheService', () => {
     );
   });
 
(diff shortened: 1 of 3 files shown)


iamacook:
Closing in favour of [removing `deleteByKeyPattern`](https://github.com/safe-global/safe-client-gateway/pull/1077) as it is no longer used.

PR #1092, opened 2024-01-31
Clear transaction from cache immediately upon deletion

iamacook:
This immediately clears a deleted transaction from the cache in case an invalidation event is not received. If an event is not received, the transaction could be signed again and therefore "added" again.


hectorgomezv (approved):
👏🏻👏🏻👏🏻

PR #1093, opened 2024-01-31
Add `isTestnet` flag from config service

iamacook:
Depends on release of https://github.com/safe-global/safe-config-service/pull/1036

This add the new `isTestnet` flag on the config service, implemented as optional as to not break the gateway whilst not released on the config service.


fmrsabino (approved):
🚀

PR #1096, opened 2024-01-31
Add Zerion Balances provider

hectorgomezv:


src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -0,0 +1,155 @@
...
+      .filter((zb) => zb.attributes.flags.displayable === true)
+      .map((zb) => {
+        const implementation = zb.attributes.fungible_info.implementations.find(
+          (implementation) => implementation.chain_id === chainName,
+        );
+        if (!implementation)
+          throw Error(
+            `Zerion error: ${chainName} implementation not found for balance ${zb.id}`,
+          );
+        const fiatBalance = getNumberString(zb.attributes.value ?? 0);
```
fmrsabino, line 99: Having `0` as a fallback can be dangerous as `0` is a valid value. What about skipping items with 0 value?

src/routes/balances/__tests__/controllers/get-externalized-balances.balances.controller.spec.ts
```diff
@@ -22,12 +22,24 @@ import { TestLoggingModule } from '@/logging/__tests__/test.logging.module';
...
+  zerionAttributesBuilder,
+  zerionFungibleInfoBuilder,
+  zerionImplementationBuilder,
+  zerionQuantityBuilder,
+  zerionBalanceBuilder,
+  zerionFlagsBuilder,
+  zerionBalancesBuilder,
+} from '@/datasources/balances-api/entities/__tests__/zerion-balance.entity.builder';
 
 describe('Balances Controller (Unit)', () => {
```
fmrsabino, line 35: For readability, I'd say to split the controller tests between the two providers. What do you think?

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -0,0 +1,155 @@
...
+  private _mapErc20Balance(
+    zerionBalanceAttributes: ZerionAttributes,
+    tokenAddress: string,
+  ): Erc20Balance {
+    const { fungible_info, quantity } = zerionBalanceAttributes;
+    return {
+      tokenAddress,
+      token: {
+        name: fungible_info.name!,
+        symbol: fungible_info.symbol!,
```
iamacook, line 121: Why do we need to assert these?

src/config/entities/configuration.ts
```diff
@@ -41,6 +43,27 @@ export default () => ({
           8453: { chainName: 'base' },
         },
       },
+      zerion: {
+        baseUri: process.env.ZERION_BASE_URI || 'https://api.zerion.io',
+        apiKey: process.env.ZERION_API_KEY,
+        chains: {
```
fmrsabino, line 49: Are these names standardised? We should consider if the Safe Config Service should play a role with storing these names.

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -0,0 +1,155 @@
...
+
+  async clearBalances(args: {
+    chainId: string;
+    safeAddress: string;
+  }): Promise<void> {
+    const key = CacheRouter.getZerionBalancesCacheKey(args);
+    await this.cacheService.deleteByKey(key);
+  }
+
+  private _getChainName(chainId: string): string {
```
iamacook, line 147: Nit: this confused me again as in the Valk implementation (I was expecting "Polygon", for example). I think we should change this to be `shortName`, etc. in a follow up PR.

src/datasources/balances-api/balances-api.manager.spec.ts
```diff
@@ -16,48 +16,71 @@ const valkBalancesApi = {
...
+  clearBalances: jest.fn(),
+} as IBalancesApi;
+
+const zerionBalancesApiMock = jest.mocked(zerionBalancesApi);
+
 describe('Balances API Manager Tests', () => {
   describe('useExternalApi checks', () => {
     configurationServiceMock.getOrThrow.mockImplementation((key) => {
       if (key === 'features.valkBalancesChainIds') return ['1', '2', '3'];
+      if (key === 'features.zerionBalancesChainIds') return ['4', '5', '6'];
```
fmrsabino, line 30: If a chain is present in both collections then it'd return the Valk provider right? This would be a configuration issue so not that relevant given that this is a testing phase.

fmrsabino (changes requested):

src/datasources/balances-api/zerion-balances-api.service.ts
hectorgomezv: Good catch. Their values are not required (can be null) according to Zerion docs, b it's probably better to use nullish coalescing as for `logoUri`. Let me know if you prefer an alternative solution.

src/datasources/balances-api/zerion-balances-api.service.ts
hectorgomezv: Ok, makes sense to me, will do it in a separate PR then 👍🏻 thanks

src/datasources/balances-api/zerion-balances-api.service.ts
hectorgomezv: Done in [d37d2b9](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/d37d2b9f58b8ed28950623131da0e8f76ce86470)

src/datasources/balances-api/balances-api.module.ts
```diff
@@ -7,6 +7,10 @@ import {
   IValkBalancesApi,
   ValkBalancesApi,
 } from '@/datasources/balances-api/valk-balances-api.service';
+import {
+  IZerionBalancesApi,
+  ZerionBalancesApi,
+} from '@/datasources/balances-api/zerion-balances-api.service';
 
 @Global()
```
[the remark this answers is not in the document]
hectorgomezv: It doesn't need to. It was a bad assumption, thanks. Changed in [2b91828](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/2b91828fa16d3a9a478461eb961910310bc11a5f)

src/datasources/balances-api/balances-api.module.ts
```diff
@@ -15,6 +19,7 @@ import {
     HttpErrorFactory,
```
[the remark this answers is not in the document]
hectorgomezv: Fixed in [2b91828](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/2b91828fa16d3a9a478461eb961910310bc11a5f)

src/config/entities/configuration.ts
```diff
@@ -41,6 +43,27 @@ export default () => ({
...
+          // 11155111 (Sepolia) is not available on Zerion
+          // 11155111: { chainName: '' },
+          1313161554: { chainName: 'aurora' },
+          137: { chainName: 'polygon' },
+          324: { chainName: 'zksync-era' },
+          42161: { chainName: 'arbitrum' },
+          42220: { chainName: 'celo' },
+          43114: { chainName: 'avalanche' },
+          56: { chainName: 'binance-smart-chain' },
+          8453: { chainName: 'base' },
```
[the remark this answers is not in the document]
hectorgomezv: They are already sorted in lexicographical order, but I agree it makes more sense to order them in numerical order. Changed in [2b91828](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/2b91828fa16d3a9a478461eb961910310bc11a5f)

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -0,0 +1,155 @@
...
+      );
+    }
+  }
+
+  private _mapBalances(
+    chainName: string,
+    zerionBalances: ZerionBalance[],
+  ): Balance[] {
+    return zerionBalances
+      .filter((zb) => zb.attributes.flags.displayable === true)
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [2b91828](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/2b91828fa16d3a9a478461eb961910310bc11a5f)

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -0,0 +1,155 @@
...
+  async getBalances(args: {
+    chainId: string;
+    safeAddress: string;
+    fiatCode: string;
+  }): Promise<Balance[]> {
+    try {
+      const cacheDir = CacheRouter.getZerionBalancesCacheDir(args);
+      const chainName = this._getChainName(args.chainId);
+      const currency = args.fiatCode.toLowerCase();
+      const url = `${this.baseUri}/v1/wallets/${args.safeAddress}/positions?filter[chain_ids]=${chainName}&currency=${currency}&sort=value`;
```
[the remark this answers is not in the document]
hectorgomezv: Yes, thanks! Changed in [2b91828](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/2b91828fa16d3a9a478461eb961910310bc11a5f)

src/datasources/balances-api/zerion-balances-api.service.ts
hectorgomezv: I think skipping 0-value items is not an option as we would skip SAFE atm for example. But instead of returning 0 from the data-source layer, we can propagate `null` from here and let the route class map it to 0 (it is doing it in https://github.com/safe-global/safe-client-gateway/blob/d37d2b9f58b8ed28950623131da0e8f76ce86470/src/routes/balances/balances.service.ts#L78)

I've changed the fallback in [2b91828](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/2b91828fa16d3a9a478461eb961910310bc11a5f)

Hopefully, we would be able to simplify the code a little if a unique provider is used in the future.

src/config/entities/configuration.ts
hectorgomezv: They are not. These are the names the providers use on their APIs, they don't map 1:1 to the Safe Config Service ones, or ERC-3770. 

As a follow-up, we could consider if it makes sense to push this configuration to the Config Service 👍🏻

src/routes/balances/__tests__/controllers/get-externalized-balances.balances.controller.spec.ts
hectorgomezv: Yes, I think it makes sense, thanks! 

Done in [4d522b5](https://github.com/safe-global/safe-client-gateway/pull/1096/commits/4d522b531cbc41a12aab02d271404db7e1cb5a96)

src/datasources/balances-api/balances-api.manager.spec.ts
hectorgomezv: Yes, I also thought about that. Probably the most correct solution would be to check at bootstrap time that the provider's configured chains don't overlap, but given this is a testing phase, I thought it would add too much effort/complexity for a low benefit.

In my opinion that would be the way to solve the issue if several providers would be a permanent scenario.

src/config/entities/configuration.ts
iamacook: Reading that there are not standardised, nor match ERC-3770 short names, ignore suggestion to [rename them](https://github.com/safe-global/safe-client-gateway/pull/1096#discussion_r1475934308).

src/datasources/balances-api/entities/__tests__/zerion-balance.entity.builder.ts
```diff
@@ -0,0 +1,71 @@
...
+    .with('icon', { url: faker.internet.url() })
+    .with('implementations', [
+      zerionImplementationBuilder().build(),
+      zerionImplementationBuilder().build(),
+    ]);
+}
+
+export function zerionQuantityBuilder(): IBuilder<ZerionQuantity> {
+  return new Builder<ZerionQuantity>()
+    .with('int', faker.number.int().toString())
```
iamacook, line 34: Nit:
```suggestion
    .with('int', faker.string.numeric())
```

iamacook (approved):

src/config/entities/configuration.ts
fmrsabino: > As a follow-up, we could consider if it makes sense to push this configuration to the Config Service 👍🏻

I'd say that your solution is totally fine. We can re-evaluate adding this to the Safe Config Service but given that they are not standardised, this also seems to be a good place to have the configuration.

src/datasources/balances-api/balances-api.manager.spec.ts
fmrsabino: > but given this is a testing phase, I thought it would add too much effort/complexity for a low benefit.

Yes exactly 👍 No action needed here it was mostly to clarify the setup.

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -66,12 +66,19 @@ export class ZerionBalancesApi implements IBalancesApi {
...
       const { data } = await this.dataSource.get<ZerionBalances>({
         cacheDir,
         url,
         notFoundExpireTimeSeconds: this.defaultNotFoundExpirationTimeSeconds,
-        networkRequest: { headers: { Authorization: `Basic ${this.apiKey}` } },
+        networkRequest: {
+          headers: { Authorization: `Basic ${this.apiKey}` },
+          params: {
+            'filter[chain_ids]': chainName,
+            currency,
```
fmrsabino, line 78: nitpick: a refactor on the name of this property would result in the wrong request being done. While the shorthand version is convenient this is something that we should pay attention to in case we change it in the future.

fmrsabino (approved):
👏

src/datasources/balances-api/zerion-balances-api.service.ts
hectorgomezv: Yeah, totally agree. I have another PR in the making to use `NetworkRequest.params` on `valk-balances-api.service.ts` as well, so I'll make the change there, thanks!

iamacook (approved):

PR #1099, opened 2024-02-01
Adjust supported currencies

hectorgomezv:
### Summary
With the addition of `BalancesManagerAPI`, several balance providers can be added to the configuration of the CGW. The available currencies returned through the endpoint `GET /v1/balances/supported-fiat-codes` should take into consideration the available currencies for these providers.

### Changes
- Adds the available currencies configuration for both Zerion and Valk balances providers.
- Adds `getFiatCodes` function to `BalancesManagerApi`, which should intersect the available currencies for all the configured providers.
- Adds `getFiatCodes` function to `BalancesApi` interface, and its concrete implementation on `ValkBalancesApi` and `ZerionBalancesApi`.
- Adds the associated tests.


src/config/entities/__tests__/configuration.ts
```diff
@@ -38,6 +38,14 @@ export default (): ReturnType<typeof configuration> => ({
...
           8453: { chainName: faker.string.sample() },
         },
+        currencies: [
+          ...Array.from({ length: faker.number.int({ min: 2, max: 5 }) }, () =>
+            faker.finance.currencyCode(),
+          ),
+          'ETH',
+          'EUR',
+          'USD',
+        ],
```
iamacook, line 48: Do we need the harcoded values here? There could potentially be duplicates.

src/config/entities/__tests__/configuration.ts
```diff
@@ -55,6 +63,15 @@ export default (): ReturnType<typeof configuration> => ({
...
         },
+        currencies: [
+          ...Array.from({ length: faker.number.int({ min: 2, max: 5 }) }, () =>
+            faker.finance.currencyCode().toLowerCase(),
+          ),
+          'btc',
+          'eth',
+          'eur',
+          'usd',
+        ],
```
iamacook, line 74: As above regarding duplicates.

iamacook (approved):

src/config/entities/__tests__/configuration.ts
hectorgomezv: > Do we need the harcoded values here?

I hardcoded some currency codes in the test configuration to be able to make some assertions at `balances.controlller.spec.ts`. Otherwise, we can't figure out the result of the intersection between providers' configured codes because we are not mocking the configuration at the Controller tests. I think doing so would add noise to the PR. But ofc, let me know if you think there's a better way to test the intersection without hardcoding some codes in the test configuration 🙂

> There could potentially be duplicates

I agree! It may be more correct to avoid duplicates (even if the intersection would hide the duplication). I've fixed it in [d6f35a0](https://github.com/safe-global/safe-client-gateway/pull/1099/commits/d6f35a006240b85a4e1f2a9743a33342c0b1e1c4)

src/config/entities/__tests__/configuration.ts
hectorgomezv: Also done in [d6f35a0](https://github.com/safe-global/safe-client-gateway/pull/1099/commits/d6f35a006240b85a4e1f2a9743a33342c0b1e1c4)

iamacook (approved):

PR #1101, opened 2024-02-02
Extract Verification Code table

fmrsabino:
The verification code related data was extracted from the Accounts' table. Verification Codes are temporary data that should only be present while an account is being verified.

Previously, after an account was verified, most of the data tied to the verification process was set to `null` as it was still part of the `accounts` table. By extracting the verification code related data, we can now delete the respective rows instead of holding null values that were mostly used for temporary purposes.

### DIFF
--- a/src/datasources/account/account.datasource.ts
+++ b/src/datasources/account/account.datasource.ts
@@ -5,8 +5,10 @@ import { AccountDoesNotExistError } from '@/datasources/account/errors/account-d
 import {
   Account as DomainAccount,
   EmailAddress,
+  VerificationCode as DomainVerificationCode,
 } from '@/domain/account/entities/account.entity';
 import { Subscription as DomainSubscription } from '@/domain/account/entities/subscription.entity';
+import { VerificationCodeDoesNotExistError } from '@/datasources/account/errors/verification-code-does-not-exist.error';
 
 interface Account {
   id: number;
@@ -15,10 +17,14 @@ interface Account {
   safe_address: string;
   signer: string;
   verified: boolean;
-  verification_code: string | null;
-  verification_code_generated_on: Date | null;
-  verification_sent_on: Date | null;
   unsubscription_token: string;
+}
+
+interface VerificationCode {
+  account_id: number;
+  code: string;
+  generated_on: Date;
+  sent_on: Date | null;
 }
 
 interface Subscription {
@@ -70,9 +76,35 @@ export class AccountDataSource implements IAccountDataSource {
       isVerified: account.verified,
       safeAddress: account.safe_address,
       signer: account.signer,
-      verificationCode: account.verification_code,
-      verificationGeneratedOn: account.verification_code_generated_on,
-      verificationSentOn: account.verification_sent_on,
+    };
+  }
+
+  async getAccountVerificationCode(args: {
+    chainId: string;
+    safeAddress: string;
+    signer: string;
+  }): Promise<DomainVerificationCode> {
+    const [verificationCode] = await this.sql<VerificationCode[]>`SELECT *
+                                                                  FROM verification_codes
+                                                                           inner join accounts on accounts.id = verification_codes.account_id
+                                                                  WHERE account_id = accounts.id
+                                                                    AND chain_id = ${args.chainId}
+                                                                    AND safe_address = ${args.safeAddress}
+                                                                    AND signer = ${args.signer}
+    `;
+
+    if (!verificationCode) {
+      throw new VerificationCodeDoesNotExistError(
+        args.chainId,
+        args.safeAddress,
+        args.signer,
+      );
+    }
+
+    return {
+      code: verificationCode.code,
+      generatedOn: verificationCode.generated_on,
+      sentOn: verificationCode.sent_on,
     };
   }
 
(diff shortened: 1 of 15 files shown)


src/datasources/account/account.datasource.ts
```diff
@@ -100,116 +157,153 @@ export class AccountDataSource implements IAccountDataSource {
...
+        FROM accounts
+        WHERE chain_id = ${args.chainId}
+          AND safe_address = ${args.safeAddress}
+          AND signer = ${args.signer}
+          AND verified = false
+        ON CONFLICT (account_id)
+            DO UPDATE SET code         = ${args.code},
+                          generated_on = ${args.codeGenerationDate}
+        RETURNING *
+    `;
```
fmrsabino, line 173: Since the Verification Code is now separate from the Accounts, setting a verification code now does an "upsert" if the account is not verified.

src/domain/interfaces/account.datasource.interface.ts
```diff
@@ -96,7 +104,7 @@ export interface IAccountDataSource {
     chainId: string;
     safeAddress: string;
     signer: string;
-  }): Promise<void>;
+  }): Promise<Account>;
```
hectorgomezv, line 107: Are these return types changes related to this PR? 

The return type for this one in particular is not used anywhere AFAIK.

src/domain/account/account-repository.service.ts
```diff
@@ -199,33 +221,36 @@ export class AccountRepository implements IAccountRepository {
...
       throw new EmailEditMatchesError(args);
     }
 
-    const verificationCode = this._generateCode();
+    const newVerificationCode = this._generateCode();
 
     await this.accountDataSource.updateAccountEmail({
       chainId: args.chainId,
-      code: verificationCode,
+      code: newVerificationCode,
```
hectorgomezv, line 235: This `newVerificationCode` is not being updated in https://github.com/safe-global/safe-client-gateway/blob/5e7817f74560efba5ad02c5a2b2f33d017e6baba/src/datasources/account/account.datasource.ts#L277 or am I missing something?

Related to this, just for consideration: should we have an specific function to update the verification code instead of updating it through `updateAccountEmail`?

src/domain/account/entities/__tests__/verification-code.builder.ts
```diff
@@ -0,0 +1,10 @@
+import { Builder, IBuilder } from '@/__tests__/builder';
+import { VerificationCode } from '@/domain/account/entities/account.entity';
+import { faker } from '@faker-js/faker';
+
+export function verificationCodeBuilder(): IBuilder<VerificationCode> {
+  return new Builder<VerificationCode>()
+    .with('code', faker.string.numeric({ length: 6 }))
+    .with('generatedOn', new Date())
```
hectorgomezv, line 8: Should we randomize this date value at builder level?

hectorgomezv (changes requested):

src/domain/interfaces/account.datasource.interface.ts
fmrsabino: Not exactly related no. It came from an iteration of returning always the updated account. I will remove it as we don't use the resulting type currently.

src/domain/account/account-repository.service.ts
fmrsabino: > This `newVerificationCode` is not being updated in
> 
> https://github.com/safe-global/safe-client-gateway/blob/5e7817f74560efba5ad02c5a2b2f33d017e6baba/src/datasources/account/account.datasource.ts#L277
> 
> or am I missing something?

Good catch! Will fix it.


> Related to this, just for consideration: should we have an specific function to update the verification code instead of updating it through `updateAccountEmail`?

That functionality is already provided with `AccountDataSource.setEmailVerificationCode`.

src/domain/account/entities/__tests__/verification-code.builder.ts
fmrsabino: For this one I think it's better to not randomize it as most code generations should take the current `Date` value as a default. Testing a value that does not represent the "now" seems to be less likely than one with the current value. Additionally the goal of the builders is to reduce verbosity in the tests with "sensible" defaults.

src/domain/account/account-repository.service.ts
fmrsabino: Fixed in [1029720](https://github.com/safe-global/safe-client-gateway/pull/1101/commits/1029720c35fe0377c02abd0e9d0abe197d2ae609).

Added relevant assertions. Also added a `TODO` with a proposed change to tackle in a separate PR as it was not really within the context of this change.

src/domain/interfaces/account.datasource.interface.ts
fmrsabino: Removed in [1029720](https://github.com/safe-global/safe-client-gateway/pull/1101/commits/1029720c35fe0377c02abd0e9d0abe197d2ae609)

src/datasources/account/account.datasource.ts
```diff
@@ -100,116 +157,153 @@ export class AccountDataSource implements IAccountDataSource {
...
-                                                RETURNING *`;
+  }): Promise<DomainVerificationCode> {
+    const [verificationCode] = await this.sql<VerificationCode[]>`
+        INSERT INTO verification_codes (account_id, code, generated_on)
+        SELECT id, ${args.code}, ${args.codeGenerationDate}
+        FROM accounts
+        WHERE chain_id = ${args.chainId}
+          AND safe_address = ${args.safeAddress}
+          AND signer = ${args.signer}
+          AND verified = false
```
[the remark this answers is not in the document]
fmrsabino: Yes 👍 [1029720](https://github.com/safe-global/safe-client-gateway/pull/1101/commits/1029720c35fe0377c02abd0e9d0abe197d2ae609)

src/domain/interfaces/account.datasource.interface.ts
```diff
@@ -1,6 +1,8 @@
 import {
   Account,
   EmailAddress,
+  VerificationCode,
+  VerificationCode as DomainVerificationCode,
```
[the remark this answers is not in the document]
fmrsabino: Removed in [1029720](https://github.com/safe-global/safe-client-gateway/pull/1101/commits/1029720c35fe0377c02abd0e9d0abe197d2ae609)

hectorgomezv (approved):
👏🏻👏🏻👏🏻

PR #1102, opened 2024-02-05
Add `chainLogoUri` field from config service

### LINKED ISSUE #463: Add Chain logo to
In the mobile app have to new design for network selection screen, to comply with the design we need to get chain logo as part of  chain info returned from the CGW chain endpoint. 

We have a [ticket](https://github.com/safe-global/safe-config-service/issues/883) already to add this to the config service so it will be the source of this info

iamacook:
Closes #463

Depends on release of https://github.com/safe-global/safe-config-service/pull/1039

This add the new `chainLogoUri` field on the config service, implemented as optional as to not break the gateway whilst not released on the config service.


fmrsabino (approved):
🚀

PR #1104, opened 2024-02-05
Clear notification_types contents on tests execution

hectorgomezv:
### Summary
The `notification_types` table is not being cleared after the tests' execution. This causes the table rows to incrementally grow on each execution, which fails when a randomly generated `notification_type` row collides with an existing one.

### Changes
- Truncates `notification_types` table after each test execution.


fmrsabino (approved):
🚀

PR #1116, opened 2024-02-07
Add route to retrieve email address

fmrsabino:
- Adds a new route to retrieve an email address: `GET /v1/chains/:chainId/safes/:safeAddress/emails/:signerAddress`
- This route returns the `emailAddress` and respective verification status for authenticated requests. If there are no emails associated with the signer a `404 Not Found` is returned. 

```javascript
GET /v1/chains/:chainId/safes/:safeAddress/emails/:signerAddress

{
  email: <string>, // the signer's email address
  verified: <boolean>, // the current verification status of the email
}
```

- This new route requires two headers to be set:
  * `Safe-Wallet-Signature` – which includes the signature of the expected message
  * `Safe-Wallet-Signature-Timestamp` – which includes the timestamp that is signed with the message
- The expected message to be signed is: `email-retrieval-${chainId}-${safe}-${signer}-${timestamp}`
- A signature is only valid for 5 minutes. If a signature is re-used after these 5 minutes, a `403 Forbidden` is returned.

### Future work

We can move the authentication components to the headers instead for the other endpoints that require some sort of authentication. This decouples the authentication part with the payload part of each request which was an issue when building this `GET` endpoint.

### DIFF
--- a/src/routes/email/guards/timestamp.guard.ts
+++ b/src/routes/email/guards/timestamp.guard.ts
@@ -1,4 +1,4 @@
-import { CanActivate, ExecutionContext, Type, mixin } from '@nestjs/common';
+import { CanActivate, ExecutionContext, mixin, Type } from '@nestjs/common';
 
 /**
  * Returns a guard mixin that can be used to check if a 'timestamp'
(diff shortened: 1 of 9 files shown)


src/routes/email/guards/timestamp.guard.ts
```diff
@@ -13,7 +13,9 @@ export const TimestampGuard = (maxElapsedTimeMs: number): Type<CanActivate> => {
     canActivate(context: ExecutionContext): boolean {
       const request = context.switchToHttp().getRequest();
 
-      const timestampRaw = request.body['timestamp'];
+      const timestampRaw =
+        request.body['timestamp'] ??
+        request.headers['safe-wallet-signature-timestamp'];
```
fmrsabino, line 18: Had to add this fallback since this guard was expecting the timestamp to be given in the body of the request.

hectorgomezv (approved):
👏🏻👏🏻👏🏻

PR #1120, opened 2024-02-08
Add PreExecutionLogGuard to POST /v1/hooks/events route

### LINKED ISSUE #1118: Add additional logging to `POST /v1/hooks/events` route
### Summary
We log the request data when the request cycle is finalized (either in complete or error states).

For this particular `POST /v1/hooks/events` route, we want to log the request data as soon as it reaches the service.

hectorgomezv:
Closes #1118 

### Changes:
- Adds `PreExecutionLogGuard` to `CacheHooksController.postEvent`, which prints a log line holding the request data each time an event reaches the service.

### DIFF
--- a/src/routes/cache-hooks/guards/pre-execution.guard.ts
+++ b/src/routes/cache-hooks/guards/pre-execution.guard.ts
@@ -1,0 +1,38 @@
+import {
+  CanActivate,
+  ExecutionContext,
+  Inject,
+  Injectable,
+} from '@nestjs/common';
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
+import { formatRouteLogMessage } from '@/logging/utils';
+
+/**
+ * The PreExecutionLogGuard guard outputs a log line containing the request data.
+ */
+@Injectable()
+export class PreExecutionLogGuard implements CanActivate {
+  private readonly PRE_EXECUTION_LOGGING_DETAIL = 'pre-execution-log';
+  private readonly PRE_EXECUTION_LOGGING_STATUS = 202;
+
+  constructor(
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
+  ) {}
+
+  canActivate(context: ExecutionContext): boolean {
+    const startTimeMs: number = performance.now();
+    const httpContext = context.switchToHttp();
+    const request = httpContext.getRequest();
+
+    this.loggingService.info(
+      formatRouteLogMessage(
+        this.PRE_EXECUTION_LOGGING_STATUS,
+        request,
+        startTimeMs,
+        this.PRE_EXECUTION_LOGGING_DETAIL,
+      ),
+    );
+
+    return true;
+  }
+}
--- a/src/routes/cache-hooks/guards/pre-execution.guard.ts
+++ b/src/routes/cache-hooks/guards/pre-execution.guard.ts
@@ -1,38 +1,30 @@
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
 import {
   CanActivate,
   ExecutionContext,
   Inject,
   Injectable,
 } from '@nestjs/common';
-import { ILoggingService, LoggingService } from '@/logging/logging.interface';
-import { formatRouteLogMessage } from '@/logging/utils';
 
 /**
- * The PreExecutionLogGuard guard outputs a log line containing the request data.
+ * The PreExecutionLogGuard guard outputs a log line containing parts of the request data.
+ * Currently only the request path is being logged.
  */
 @Injectable()
 export class PreExecutionLogGuard implements CanActivate {
-  private readonly PRE_EXECUTION_LOGGING_DETAIL = 'pre-execution-log';
-  private readonly PRE_EXECUTION_LOGGING_STATUS = 202;
+  private static readonly PRE_EXECUTION_LOGGING_DETAIL = 'pre-execution-log';
 
   constructor(
     @Inject(LoggingService) private readonly loggingService: ILoggingService,
   ) {}
 
   canActivate(context: ExecutionContext): boolean {
-    const startTimeMs: number = performance.now();
     const httpContext = context.switchToHttp();
     const request = httpContext.getRequest();
-
-    this.loggingService.info(
-      formatRouteLogMessage(
-        this.PRE_EXECUTION_LOGGING_STATUS,
-        request,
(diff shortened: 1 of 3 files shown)


src/routes/common/interceptors/route-logger.interceptor.ts
```diff
@@ -27,6 +27,10 @@ import { Request, Response } from 'express';
  */
 @Injectable()
 export class RouteLoggerInterceptor implements NestInterceptor {
+  private readonly PRE_EXECUTION_LOGGING_ROUTE = '/v1/hooks/events';
+  private readonly PRE_EXECUTION_LOGGING_DETAIL = 'pre-execution-log';
+  private readonly PRE_EXECUTION_LOGGING_STATUS = 202;
```
hectorgomezv, line 32: These values are hardcoded, as this can be seen more like a transient feature. 

`202 - Accepted` is used as `formatRouteLogMessage` has a mandatory `statusCode` argument, and I'm trying to not do many changes across classes for this.

hectorgomezv: > Have you considered tackling this in CacheHooksController?

Yes, but I added the logging to this interceptor to be as close as possible to the request cycle start (this interceptor is executed before the code in the controller function body if I'm not wrong).

> Also you don't need to call formatRouteLogMessage unless you want to log all the parameters tied to the route. Is that the case? 🤔

I agree, but I thought it would be nice to reuse the mappings provided by that utility and have the full information about the request in the logs.

Anyway, I don't have a strong opinion regarding these topics, we can move the logging to the controller if you prefer, but as I said if a request execution gets stuck in the interceptor for some reason, the log won't be generated 🙂

fmrsabino: So we have logs for when the request finishes. That's the expectation and that's why we have the status code available.

With this change we are mixing pre-call logs with post-call logs. This can twist the metrics that we have for `/v1/hooks/events`.

If there's any reason to believe that this interceptor is somehow blocking calls then I can understand the change. Else I do think that `CacheHooksController` or another interceptor specialised in incoming requests is a better fit.

fmrsabino (changes requested):

src/routes/cache-hooks/guards/pre-execution.guard.ts
```diff
@@ -0,0 +1,38 @@
...
+    const request = httpContext.getRequest();
+
+    this.loggingService.info(
+      formatRouteLogMessage(
+        this.PRE_EXECUTION_LOGGING_STATUS,
+        request,
+        startTimeMs,
+        this.PRE_EXECUTION_LOGGING_DETAIL,
+      ),
+    );
```
fmrsabino, line 34: I'd choose a different format for the route. This log ends up being printed anyways by the `RouteLoggerInterceptor`. This log can represent something specific to this guard. The `request` part only has the path of the route which we know for this one (and it's already being distinguished with `PRE_EXECUTION_LOGGING_DETAIL`).

What about the following:

```typescript
canActivate(context: ExecutionContext): boolean {
  this.loggingService.info({
    type: PreExecutionLogGuard.PRE_EXECUTION_LOGGING_DETAIL,
  });
  return true;
}
```

src/routes/cache-hooks/guards/pre-execution.guard.ts
```diff
@@ -0,0 +1,38 @@
...
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
+import { formatRouteLogMessage } from '@/logging/utils';
+
+/**
+ * The PreExecutionLogGuard guard outputs a log line containing the request data.
+ */
+@Injectable()
+export class PreExecutionLogGuard implements CanActivate {
+  private readonly PRE_EXECUTION_LOGGING_DETAIL = 'pre-execution-log';
+  private readonly PRE_EXECUTION_LOGGING_STATUS = 202;
```
fmrsabino, line 16: These can be `static`.

src/routes/cache-hooks/guards/pre-execution.guard.ts
```diff
@@ -0,0 +1,38 @@
...
+  CanActivate,
+  ExecutionContext,
+  Inject,
+  Injectable,
+} from '@nestjs/common';
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
+import { formatRouteLogMessage } from '@/logging/utils';
+
+/**
+ * The PreExecutionLogGuard guard outputs a log line containing the request data.
```
fmrsabino, line 11: nitpick: the only request data that is being logged is the path. It does not include the body of the response for example.

fmrsabino (changes requested):

src/routes/cache-hooks/guards/pre-execution.guard.ts
hectorgomezv: So by doing that, we would have a log line emitted by this guard and another log line emitted by the `RouteLoggerInterceptor`. The first would not have a route param, but the second would. 

That would let us check each line having `PreExecutionLogGuard.PRE_EXECUTION_LOGGING_DETAIL` is paired to another log line with `route: '/v1/hooks/events'`. That may work when the guard is only associated with a single route.

hectorgomezv: I think it's a bit more consistent to log the route in the guard as well, but let me know if you think otherwise.

I've changed the format to log the path and the discrimination string in [5c9cfb7](https://github.com/safe-global/safe-client-gateway/pull/1120/commits/5c9cfb7870538e0177c956a10ef91b3e7b9b620f)

src/routes/cache-hooks/guards/pre-execution.guard.ts
hectorgomezv: I agree, thanks! [5c9cfb7](https://github.com/safe-global/safe-client-gateway/pull/1120/commits/5c9cfb7870538e0177c956a10ef91b3e7b9b620f)

src/routes/cache-hooks/guards/pre-execution.guard.ts
hectorgomezv: Changed [5c9cfb7](https://github.com/safe-global/safe-client-gateway/pull/1120/commits/5c9cfb7870538e0177c956a10ef91b3e7b9b620f)

PR #1129, opened 2024-02-12
Add `GET` `/v1/safes` overview endpoint

iamacook:
This adds a new `GET` `/v1/safes` endpoint for "overviews" of specified Safes, returning the following:

```ts
type AddressInfo = {
  value: string,
  name: string | null,
  logoUri: string | null
}

type Response = Array<{
  chainId: string;
  address: AddressInfo;
  owners: Array<AddressInfo>;
  threshold: number;
  fiatTotal: string; // e.g. "69.42"
  queued: number;
  awaitingConfirmation: number;
}>
```

It requires the following search parameters: `currency` and `safes`. It also accepts `walletAddress`, `exclude_spam` and `trusted` search parameters:

#### `currency`

The desired fiat to return the balance of the Safe in, e.g. `USD`

#### `safes`

A CAIP-10 comma separated list of Safe addresses, without the EIP155 namespace, e.g. `1:0x123...,1:0x456...,137:0x789`.

⚠️ Note: this is limited to a maximum of 7 but can be modified by setting the `MAX_SAFE_OVERVIEWS` env. var.

#### `walletAddress`

The owner which is checked for awaiting confirmations as a "standard" address, e.g. `0x123...`

#### `trusted` (default: `false`)

Whether to return "trusted" transactions or not.

#### `exclude_spam` (default: `true`)

Whether to exclude "spam" tokens or not.

### DIFF
--- a/src/domain/balances/balances.repository.interface.ts
+++ b/src/domain/balances/balances.repository.interface.ts
@@ -10,6 +10,7 @@ export interface IBalancesRepository {
   getBalances(args: {
     chainId: string;
     safeAddress: string;
+    fiatCode: string;
     trusted?: boolean;
     excludeSpam?: boolean;
   }): Promise<Balance[]>;
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -130,6 +130,9 @@ export default (): ReturnType<typeof configuration> => ({
     history: {
       maxNestedTransfers: faker.number.int({ min: 1, max: 5 }),
     },
+    safe: {
+      maxOverviews: faker.number.int({ min: 1, max: 5 }),
+    },
   },
   prices: {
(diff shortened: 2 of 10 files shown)


iamacook:
Marking as a draft whilst I fix the tests.

src/domain/balances/balances.repository.interface.ts
```diff
@@ -10,6 +10,7 @@ export interface IBalancesRepository {
   getBalances(args: {
     chainId: string;
     safeAddress: string;
+    fiatCode: string;
```
hectorgomezv, line 13: Good catch! 🙂

hectorgomezv (commented):
Besides the incompatibilities with newer code changes in `safes.controller.overview.spec.ts`, it looks good to me!

hectorgomezv (approved):
🚀

PR #1130, opened 2024-02-12
Add getAccounts to AccountDataSource

fmrsabino:
Adds a function to retrieve all the accounts tied to a Safe Address. This is particularly useful when only the Safe Address and chain id are available (e.g.: alert triggering).

The functionality was previously covered by `getVerifiedAccountEmailsBySafeAddress`. However, it was only returning the string values of the respective email addresses. While this covers some use-cases (to whom to send an email address), some other properties of the `Account` are still useful when sending email address such as the ability to check and send the unsubscription token.

### DIFF
--- a/src/datasources/account/account.datasource.spec.ts
+++ b/src/datasources/account/account.datasource.spec.ts
@@ -5,7 +5,13 @@ import { faker } from '@faker-js/faker';
 import { AccountDoesNotExistError } from '@/domain/account/errors/account-does-not-exist.error';
 import * as shift from 'postgres-shift';
 import configuration from '@/config/entities/__tests__/configuration';
-import { EmailAddress } from '@/domain/account/entities/account.entity';
+import {
+  Account,
+  EmailAddress,
+  VerificationCode,
+} from '@/domain/account/entities/account.entity';
+import { accountBuilder } from '@/domain/account/entities/__tests__/account.builder';
+import { verificationCodeBuilder } from '@/domain/account/entities/__tests__/verification-code.builder';
 
 const DB_CHAIN_ID_MAX_VALUE = 2147483647;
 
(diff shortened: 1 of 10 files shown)


src/domain/alerts/alerts.repository.ts
```diff
@@ -100,7 +101,14 @@ export class AlertsRepository implements IAlertsRepository {
...
     } catch {
-      await this._notifyUnknownTransaction({ chainId, safeAddress, emails });
+      const emails = verifiedAccounts.map(
+        (account) => account.emailAddress.value,
+      );
+      await this._notifyUnknownTransaction({
+        chainId,
+        safeAddress,
+        emails,
+      });
```
iamacook, line 111: It's out of the scop eof this PR, but we will need to send the unsubscribe token here eventually.

src/domain/alerts/alerts.repository.ts
```diff
@@ -242,6 +251,9 @@ export class AlertsRepository implements IAlertsRepository {
       };
     });
 
+    const emails = verifiedAccounts.map(
+      (account) => account.emailAddress.value,
+    );
     return this.emailApi.createMessage({
```
iamacook, line 257: As above regarding the token.

src/datasources/account/account.datasource.spec.ts
```diff
@@ -238,88 +244,157 @@ describe('Account DataSource Tests', () => {
...
-        unsubscriptionToken: faker.string.uuid(),
-      },
-      {
-        emailAddress: new EmailAddress(faker.internet.email()),
-        signer: faker.finance.ethereumAddress(),
-        code: faker.number.int({ max: 999998 }).toString(),
-        codeGenerationDate: faker.date.recent(),
-        unsubscriptionToken: faker.string.uuid(),
-      },
+    const verifiedAccounts: [Account, VerificationCode][] = [
```
iamacook, line 247: Nice one for adding the types 💪

iamacook (approved):

PR #1141, opened 2024-02-13
Add `GelatoApi` service

iamacook:
This adds a Gelato-based `IRelayApi` service for sponsoring transactions which caches relay attempt counts:

- A new `relay` cache entry
- `GelatoApi` service with relay/count retrieval methods

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "test:all:cov": "jest --coverage --config ./test/jest-all.json"
   },
   "dependencies": {
+    "@gelatonetwork/relay-sdk": "^5.5.5",
     "@nestjs/cli": "^10.3.1",
     "@nestjs/common": "^10.3.3",
     "@nestjs/config": "^3.1.1",
--- a/src/datasources/relay-api/gelato-api.service.spec.ts
+++ b/src/datasources/relay-api/gelato-api.service.spec.ts
@@ -12,7 +12,7 @@ const mockGelatoClient = jest.mocked({
 } as jest.MockedObjectDeep<InstanceType<typeof GelatoRelay>>);
 
 const mockLoggingService = {
-  info: jest.fn(),
+  warn: jest.fn(),
 } as jest.MockedObjectDeep<ILoggingService>;
 
 describe('GelatoApi', () => {
@@ -39,7 +39,6 @@ describe('GelatoApi', () => {
       const chainId = faker.string.numeric();
       const address = faker.finance.ethereumAddress();
       const cacheDir = new CacheDir(`${chainId}_relay_${address}`, '');
-
       const count = faker.number.int();
 
       await fakeCacheService.set(cacheDir, count.toString());
@@ -72,9 +71,7 @@ describe('GelatoApi', () => {
       const data = faker.string.hexadecimal() as Hex;
       const apiKey = faker.string.sample();
       const taskId = faker.string.alphanumeric();
-
       fakeConfigurationService.set(`gelato.apiKey.${chainId}`, apiKey);
-
       mockGelatoClient.sponsoredCall.mockResolvedValue({ taskId });
 
       const result = await target.relay({
@@ -103,7 +100,6 @@ describe('GelatoApi', () => {
       const data = faker.string.hexadecimal() as Hex;
       const gasLimit = faker.number.bigInt();
       const apiKey = faker.string.sample();
-
       fakeConfigurationService.set(`gelato.apiKey.${chainId}`, apiKey);
 
       await target.relay({
@@ -141,13 +137,10 @@ describe('GelatoApi', () => {
     });
 
     it('should increment the count after relaying', async () => {
-      const setSpy = jest.spyOn(fakeCacheService, 'set');
-
       const chainId = faker.string.numeric();
       const address = faker.finance.ethereumAddress() as Hex;
       const data = faker.string.hexadecimal() as Hex;
       const apiKey = faker.string.sample();
-
       fakeConfigurationService.set(`gelato.apiKey.${chainId}`, apiKey);
 
       await target.relay({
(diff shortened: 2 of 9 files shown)


src/datasources/relay-api/gelato-api.service.ts
```diff
@@ -0,0 +1,86 @@
...
+        gasLimit,
+      },
+    );
+
+    await this.incrementRelayCount({
+      chainId: args.chainId,
+      address: args.to,
+    }).catch((error) => {
+      // If we fail to increment count, we should not fail the relay
+      this.loggingService.info(error.message);
```
fmrsabino, line 67: A `warn` level would be a better fit here as we don't expect this to happen but it's also non-blocking.

package.json
```diff
@@ -23,6 +23,7 @@
     "test:all:cov": "jest --coverage --config ./test/jest-all.json"
   },
   "dependencies": {
+    "@gelatonetwork/relay-sdk": "^5.5.5",
```
fmrsabino, line 26: We should consider not using the `@gelatonetwork/relay-sdk`. As it stands it does not have tests – https://github.com/gelatodigital/relay-sdk/tree/v5.5.5 and also pulls yet another network library (`axios`)

Can be done in a follow up PR.

src/datasources/relay-api/gelato-api.service.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  private getRelayGasLimit(gasLimit: bigint): bigint {
+    return gasLimit + GelatoApi.GAS_LIMIT_BUFFER;
+  }
+
+  private async incrementRelayCount(args: {
+    chainId: string;
+    address: string;
+  }): Promise<void> {
+    const cacheDir = CacheRouter.getRelayCacheDir(args);
```
fmrsabino, line 81: nitpick: would move this closer to where it's actually being used (so closer to the `set`).

src/datasources/relay-api/gelato-api.service.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      mockLoggingService,
+    );
+  });
+
+  describe('getRelayCount', () => {
+    it('should return the current count from the cache', async () => {
+      const chainId = faker.string.numeric();
+      const address = faker.finance.ethereumAddress();
+      const cacheDir = new CacheDir(`${chainId}_relay_${address}`, '');
+
```
fmrsabino, line 42:
```suggestion
```

src/datasources/relay-api/gelato-api.service.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+  });
+
+  describe('relay', () => {
+    it('should relay the payload', async () => {
+      const chainId = faker.string.numeric();
+      const address = faker.finance.ethereumAddress() as Hex;
+      const data = faker.string.hexadecimal() as Hex;
+      const apiKey = faker.string.sample();
+      const taskId = faker.string.alphanumeric();
+
```
fmrsabino, line 75: nitpick:
```suggestion
```

src/datasources/relay-api/gelato-api.service.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      fakeConfigurationService.set(`gelato.apiKey.${chainId}`, apiKey);
+
```
fmrsabino, line 77: nitpick:
```suggestion
```

src/datasources/relay-api/gelato-api.service.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      );
+    });
+
+    it('should add a gas buffer if a gas limit is provided', async () => {
+      const chainId = faker.string.numeric();
+      const address = faker.finance.ethereumAddress() as Hex;
+      const data = faker.string.hexadecimal() as Hex;
+      const gasLimit = faker.number.bigInt();
+      const apiKey = faker.string.sample();
+
```
fmrsabino, line 106: nitpick:
```suggestion
```

src/datasources/relay-api/gelato-api.service.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+        target.relay({
+          chainId,
+          to: address,
+          data,
+        }),
+      ).rejects.toThrow();
+    });
+
+    it('should increment the count after relaying', async () => {
+      const setSpy = jest.spyOn(fakeCacheService, 'set');
```
fmrsabino, line 144: I don't think we should mix the use of a `fake` and `spy` here. We can get the count set by the fake cache and assert the returned value.

src/datasources/relay-api/gelato-api.service.spec.ts
```diff
@@ -0,0 +1,187 @@
...
+      });
+
+      expect(setSpy).toHaveBeenCalledWith(
+        new CacheDir(`${chainId}_relay_${address}`, ''),
+        '1',
+      );
+    });
+
+    it('should not fail the relay if incrementing the count fails', async () => {
+      const setSpy = jest.spyOn(fakeCacheService, 'set');
```
fmrsabino, line 166: Same here. If the goal is to measure the number of calls probably a fake implementation is the wrong tool for this test.

fmrsabino (changes requested):

src/datasources/relay-api/gelato-api.service.ts
iamacook: Changed in 641026000723aeba2b9c4abce2179b1b7d54ecd1.

package.json
iamacook: I'll investigate this in a follow up.

src/datasources/relay-api/gelato-api.service.ts
iamacook: Moved in 641026000723aeba2b9c4abce2179b1b7d54ecd1.

src/datasources/relay-api/gelato-api.service.spec.ts
iamacook: Removed in 641026000723aeba2b9c4abce2179b1b7d54ecd1.

src/datasources/relay-api/gelato-api.service.spec.ts
iamacook: Removed in 641026000723aeba2b9c4abce2179b1b7d54ecd1.

src/datasources/relay-api/gelato-api.service.spec.ts
iamacook: Removed in 641026000723aeba2b9c4abce2179b1b7d54ecd1.

src/datasources/relay-api/gelato-api.service.spec.ts
iamacook: Removed in 641026000723aeba2b9c4abce2179b1b7d54ecd1.

src/datasources/relay-api/gelato-api.service.spec.ts
iamacook: Changed to check the return value in 641026000723aeba2b9c4abce2179b1b7d54ecd1.

src/datasources/relay-api/gelato-api.service.spec.ts
iamacook: This is needed to throw setting the cache. I've moved it to where it's used in 641026000723aeba2b9c4abce2179b1b7d54ecd1 and added a comment. Let me know your thoughts.

fmrsabino: While I understand the use-case I think it can be confusing to mock/change the behaviour of the fake object via Jest.

We should either extend the fake object to provide what we need for the tests or use a mocked instance. This works and is mostly a minor comment so no need to address it now.

iamacook: I overlooked this comment before merging. I will pass a mocked instance instead in a separate PR.

PR #1144, opened 2024-02-14
Add officiality check for Safe and MultiSend contracts in `LimitAddressesMapper`

iamacook:
This adds officiality checks for the Safes to-be-limited, as well as the MultiSend contract being interacted with. (The ProxyFactory needs not be checked as the `singleton` of the Safe-to-be-deployed is.)

- `LimitAddressesMapper['isOfficialMastercopy']` method added
- `LimitAddressesMapper['isOfficialMultiSendDeployment']` method added
- Officiality checks added in place

Note: small tweaks and test coverage is also included

PR #1147, opened 2024-02-15
Generate and infer from ABIs of `@safe-global/safe-deployments`

iamacook:
⚠️ Depends on #1144

This adds a script to extract `const` asserted ABIs from the `@safe-global/safe-deployments` package and uses them in their respective `AbiDecoder`

## Changes

- Add `scripts/generate-abis.ts` script, running on `postinstall`, to generate ABIs in `/dist/abis/safe/*`
- Use the ABIs in `MultiSendDecoder`, `SafeDecoder` and `ProxyFactoryDecoder`


scripts/generate-abis.ts
```diff
@@ -0,0 +1,68 @@
...
+    const versionDir = path.join(assetsDir, version);
+
+    // ...parse the ABI for each asset
+    for (const assetFile of fs.readdirSync(versionDir)) {
+      // Read the asset JSON
+      const assetPath = path.join(assetsDir, version, assetFile);
+      const assetJson = fs.readFileSync(assetPath, 'utf8');
+
+      // Parse the asset JSON
+      const { contractName, abi } = JSON.parse(assetJson);
```
fmrsabino, line 53: This can be a bottleneck if the JSON file is large. We should look into working with the Stream API if possible – https://nodejs.org/api/stream.html#stream

However reading and validating the JSON file might require the complete context and might be tricky to process it in chunks without relying on an external tool. Currently we would be holding the entire JSON file in memory (the `assetJson`).

We can potentially use `createWriteStream` with this solution:

```typescript
const stream = fs.createWriteStream(filePath);
stream.write(
  '// This file is auto-generated by scripts/generate-abis.ts\nexport default',
);
stream.write(data);
stream.write('as const;');
stream.end();
```

But we wouldn't get a lot of benefits out of it given that the stream would be used to write the entire "chunk" (i.e.: the entire file) that is in memory (instead of having a `pipeline` to process chunks of it).

Given the complexity of validating the JSON file in chunks I would say that this solution is totally fine as it's also a step of the build process and not a service runtime operation.

.gitignore
```diff
@@ -48,3 +48,6 @@ lerna-debug.log*
 
 # Database mounted volume
 data
+
+# ABIs
+/src/abis/safe
```
fmrsabino, line 53: nitpick: newline a the end of file

.prettierignore
```diff
@@ -1 +1,2 @@
 /.yarn
+/src/abis
```
fmrsabino, line 2: nitpick: newline a the end of file

scripts/generate-abis.ts
```diff
@@ -0,0 +1,68 @@
...
+  process.cwd(),
+  'node_modules',
+  '@safe-global',
+  'safe-deployments',
+  'dist',
+  'assets',
+);
+
+// Path to directory where ABIs will be written
+const outputDir = path.join(process.cwd(), 'src', 'abis', 'safe');
```
fmrsabino, line 28: I don't think we should generate source files to the `src` folder as these are usually manually written and we should avoid mixing generated sources with the sources that we track on the version control.

This also creates the wrong expectation that if someone creates source code on `/src/abis` that source code would be deleted by this script but also ignored by Git (given the `.gitignore` rules).

Have you considered the `dist` folder of the project instead?

fmrsabino (changes requested):

scripts/generate-abis.ts
iamacook: As you said, I don't think there will be much benefit from using the Steam API as this doesn't run at runtime. However, I altered to using it in 9dff79a72bbe821c2a4bb3861949c01770d340c4 as it gives a clear separation of: comment, data and const assertion. I've added a comment summarising your points for future reference as well.

.gitignore
iamacook: Inconsequential as of 3ee2bbeb0c6ad411221151d30b9ead8cc3331cef.

.prettierignore
iamacook: Inconsequential as of 3ee2bbeb0c6ad411221151d30b9ead8cc3331cef.

scripts/generate-abis.ts
iamacook: It outputs to `/dist/abis/safe` as of 9dff79a72bbe821c2a4bb3861949c01770d340c4. I had to make some config adjustments to allow for `@/dist` imports but it should be self-explanatory.

scripts/generate-abis.ts
```diff
@@ -56,9 +56,22 @@ function main() {
       const fileName = `${contractName}.abi.ts`;
       const filePath = path.join(versionOutputDir, fileName);
 
-      const content = `// This file is auto-generated by scripts/generate-abis.ts
-export default ${JSON.stringify(abi, null, 2)} as const;`;
-      fs.writeFileSync(filePath, content);
+      // It is generally better to use the Steam API for larger files
```
fmrsabino, line 59: nitpick: Stream API

scripts/generate-abis.ts
```diff
@@ -25,14 +25,14 @@ const assetsDir = path.join(
 );
 
 // Path to directory where ABIs will be written
-const outputDir = path.join(process.cwd(), 'src', 'abis', 'safe');
+const outputDir = path.join(process.cwd(), 'dist', 'abis', 'safe');
```
fmrsabino, line 28: You'd need to bundle this in the docker image too 😬

scripts/generate-abis.ts
iamacook: Fixed in 0e7c2b494c9f3084fcf805103fc9f43a8920658a.

scripts/generate-abis.ts
iamacook: I _think_ I did so in 0e7c2b494c9f3084fcf805103fc9f43a8920658a. Please let me know if it's correct.

Dockerfile
```diff
@@ -11,6 +11,7 @@ COPY --chown=node:node package.json yarn.lock .yarnrc.yml tsconfig*.json ./
 RUN --mount=type=cache,target=/root/.yarn yarn
 COPY --chown=node:node assets ./assets
 COPY --chown=node:node migrations ./migrations
+COPY --chown=node:node scripts/generate-abis.ts ./scripts/generate-abis.ts
```
fmrsabino, line 14: Should be:

```
COPY --chown=node:node scripts/generate-abis.js ./scripts/generate-abis.js
```

This step also needs to happen before `RUN --mount=type=cache,target=/root/.yarn yarn` as the `postinstall` script runs then.

fmrsabino (changes requested):

Dockerfile
iamacook: I corrected the name in bb1e9cb458ccabc93c988e0e1cfa537a0a08c0a7 and moved the step in f3d7016936fe0dc19b6d44754109af3581ed74a8.

fmrsabino (approved):
Awesome! 👏 🚀

PR #1148, opened 2024-02-15
Replace Safe/ERC-20 `ContractHelper`s with `AbiDecoder`s

iamacook:
This removes the Safe/ERC-20 `ContractHelper` classes, in favour of `AbiDecoder`s.

## Changes

- Create an ERC-20 `AbiDecoder`.
- Remove `ContractHelper` and their Safe/ERC-20 variants.
- Adjust code to use Safe/ERC-20 `AbiDecoder` variants.
- Move provided `AbiDecoder`s from `relay.module` to `relay-decoders.module` and provide that instead.


src/domain/relay/contracts/__tests__/erc20-encoder.builder.ts
```diff
@@ -14,16 +14,11 @@ class Erc20TransferEncoder<T extends Erc20TransferArgs>
...
-    'function transfer(address to, uint256 value)' as const;
-
   encode(): Hex {
-    const abi = parseAbi([Erc20TransferEncoder.FUNCTION_SIGNATURE]);
-
     const args = this.build();
 
     return encodeFunctionData({
-      abi,
+      abi: erc20Abi,
```
fmrsabino, line 21: Nice. I didn't know that `viem` bundled these ABIs.

fmrsabino (approved):
👏 

Unrelated but caught my eye with the path of the files present in this diff: we should probably move the files under `src/domain/contracts/contracts` to something like `src/domain/contracts/decoders`.

iamacook:
> we should probably move the files under `src/domain/contracts/contracts` to something like `src/domain/contracts/decoders`.

Agreed. I will create a follow up PR moving them.

PR #1121, opened 2024-02-09
Add Zerion NFTs retrieval

hectorgomezv:
### Summary
This PR implements NFTs (ERC-721 tokens) retrieval using the Zerion API.

### Changes
- Adds `ZerionCollectible` entity and its related interfaces/builders.
- Adds `getCollectibles/clearCollectibles` methods to the `BalancesApi` interface.
- Implement `getCollectibles/clearCollectibles` on `ZerionBalancesApi`.

### Out of scope
- `PaginationData` entity needs to be refactored to accommodate Zerion's pagination format (`string` offset). This PR implements a mapping between the service pagination format and the Zerion API one.

PR #1149, opened 2024-02-15
Remove Valk BalancesAPI

hectorgomezv:
Depends on #1121 

### Changes
- Removes Valk-related code (BalancesApi implementation, configuration, entities, etc.)


src/config/entities/__tests__/configuration.ts
```diff
@@ -21,35 +21,6 @@ export default (): ReturnType<typeof configuration> => ({
...
-              { length: faker.number.int({ min: 2, max: 5 }) },
-              () => faker.finance.currencyCode(),
-            ),
-            'ETH',
-            'EUR',
-            'USD',
-          ]),
-        ),
-      },
       zerion: {
```
iamacook, line 24: Need will still keep this level of nesting? Although, with the manager, I suppose we do.

iamacook (dismissed):

src/config/entities/__tests__/configuration.ts
hectorgomezv: Good point, after adding the TransactionApi to the manager I will review this configuration format 👍 Thanks!

iamacook (approved):

PR #1151, opened 2024-02-15
Add SafeBalancesApi

hectorgomezv:
### Summary
This PR moves balances/collectibles retrieval functions from `TransactionApi` to a new `SafeBalancesApi` class, which implements `BalancesApi`. The goal is to have the Safe Transaction Service as a data provider for both `TransactionApi` and `SafeBalancesApi`.

### Changes
- Moves `TransactionApi` balances-related functions (`getBalances`, `clearBalances`, `getCollectibles`, `clearCollectibles`) to `SafeBalancesApi`.
- Adds `SafeBalancesApi` routing (between different chain-specific Transaction Services) to `BalancesApiManager` (heavily inspired by `TransactionApiManager`).
- Unifies naming to stick to the interface (between `clearLocalBalances` and `clearBalances`).
- Simplifies both `BalancesRepository` and `CollectiblesRepository`, as some complexity is pushed to `BalancesApiManager`.

### Out of scope
- Further simplification can be addressed as `BalancesRepository` uses `PricesRepository`. A follow-up PR will also push this prices-related complexity to the data source layer, so `BalancesRepository` will get simplified by abstracting CoinGecko-specific logic from the domain layer.

### Other options considered
- Instead of moving functions from `TransactionApi`, it was considered to convert `TransactionApi` to an implementation of `BalancesApi`. This would avoid some duplicated code, but given that several `TransactionApi` exist in the service (one per chain), this would lead to having a `TransactionApiManager` injected into `BalancesApiManager`, and therefore _two-level routing logic_ (to decide whether to use a `TransactionApi` or not at first, and to decide which `TransactionApi` to use after). This seems to make the overall architecture more complex and harder to follow.

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -51,7 +51,6 @@ export class CacheRouter {
     );
   }
 
-  // TODO: remove this prefixed key if eventually only one balances provider is used
   static getZerionBalancesCacheKey(args: {
     chainId: string;
     safeAddress: string;
--- a/src/datasources/balances-api/balances-api.manager.spec.ts
+++ b/src/datasources/balances-api/balances-api.manager.spec.ts
@@ -1,13 +1,33 @@
 import { IConfigurationService } from '@/config/configuration.service.interface';
 import { BalancesApiManager } from '@/datasources/balances-api/balances-api.manager';
+import { CacheFirstDataSource } from '@/datasources/cache/cache.first.data.source';
+import { ICacheService } from '@/datasources/cache/cache.service.interface';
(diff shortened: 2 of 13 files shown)


src/datasources/cache/cache.router.ts
```diff
@@ -51,7 +51,6 @@ export class CacheRouter {
     );
   }
 
-  // TODO: remove this prefixed key if eventually only one balances provider is used
```
hectorgomezv, line 54: This is not applicable anymore, since it will be necessary to have prefixes as Transaction Service will be used as a fallback, and a distinction between providers still will be needed.

iamacook (approved):

PR #1153, opened 2024-02-15
Add relay controller

iamacook:
This adds a feature-flag controlled relay controller a route for relaying and another for getting the number of remaining relays possible.

## `POST` `/v1/chains/:chainId/relay`

This endpoint relays the specified transaction providing it is valid:

- `execTransaction` (of an official mastercopy):
  - Sending funds (native/ERC-20) to another party
  - Transaction cancellation
  - Safe interaction (owner management, etc.)
- `multiSend` (from an official deployment) consisting of `execTransaction`s of the same Safe
- `createProxyWithNonce` (deploying an official mastercopy)

It accepts the following payload:

```ts
type Request = {
  to: string; // Transaction recipient
  data: string; // Transaction calldata
  gasLimit?: string;
}
```

Note: if the `gasLimit` is provided, a 150k buffer will be added as per [Gelato's recommendation](https://docs.gelato.network/developer-services/relay/quick-start/optional-parameters).

## `GET` `/v1/chains/:chainId/relay/:safeAddress`

This endpoint returns the `limit` and `remaining` number of relays possible for the specified Safe:

```ts
type Response = {
  remaining: number;
  limit: number;
}
```

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -49,7 +49,7 @@ export class AppModule implements NestModule {
   // into account. The .env file loading is done by the ConfigurationModule
   // which is not available at this stage.
   static register(configFactory = configuration): DynamicModule {
-    const { email: isEmailFeatureEnabled, relay: isRelayFeatureEnables } =
+    const { email: isEmailFeatureEnabled, relay: isRelayFeatureEnabled } =
       configFactory()['features'];
 
     return {
@@ -77,7 +77,7 @@ export class AppModule implements NestModule {
         MessagesModule,
         NotificationsModule,
         OwnersModule,
-        ...(isRelayFeatureEnables ? [RelayControllerModule] : []),
+        ...(isRelayFeatureEnabled ? [RelayControllerModule] : []),
         RootModule,
         SafeAppsModule,
         SafesModule,
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -40,6 +40,7 @@ import { RootModule } from '@/routes/root/root.module';
 import { EmailControllerModule } from '@/routes/email/email.controller.module';
 import { AlertsControllerModule } from '@/routes/alerts/alerts.controller.module';
 import { RecoveryModule } from '@/routes/recovery/recovery.module';
+import { RelayControllerModule } from '@/routes/relay/relay.controller.module';
 import { SubscriptionControllerModule } from '@/routes/subscriptions/subscription.module';
 
 @Module({})
@@ -48,7 +49,8 @@ export class AppModule implements NestModule {
   // into account. The .env file loading is done by the ConfigurationModule
   // which is not available at this stage.
   static register(configFactory = configuration): DynamicModule {
-    const isEmailFeatureEnabled = configFactory()['features']['email'];
+    const { email: isEmailFeatureEnabled, relay: isRelayFeatureEnables } =
+      configFactory()['features'];
 
     return {
       module: AppModule,
@@ -75,6 +77,7 @@ export class AppModule implements NestModule {
         MessagesModule,
         NotificationsModule,
         OwnersModule,
+        ...(isRelayFeatureEnables ? [RelayControllerModule] : []),
         RootModule,
         SafeAppsModule,
         SafesModule,
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -127,6 +127,7 @@ export default () => ({
     email: process.env.FF_EMAIL?.toLowerCase() === 'true',
     zerionBalancesChainIds:
       process.env.FF_ZERION_BALANCES_CHAIN_IDS?.split(',') ?? [],
+    relay: process.env.FF_RELAY?.toLowerCase() === 'true',
   },
   httpClient: {
     // Timeout in milliseconds to be used for the HTTP client.
@@ -182,6 +183,9 @@ export default () => ({
     baseUri:
       process.env.RELAY_PROVIDER_API_BASE_URI || 'https://api.gelato.digital',
     limit: parseInt(process.env.RELAY_THROTTLE_LIMIT ?? `${5}`),
+    apiKey: {
+      100: process.env.GELATO_API_KEY_GNOSIS_CHAIN,
+    },
   },
   safeConfig: {
     baseUri:
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -1,5 +1,4 @@
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
-import { getAddress } from 'viem';
 
 export class CacheRouter {
   private static readonly ALL_TRANSACTIONS_KEY = 'all_transactions';
@@ -409,8 +408,7 @@ export class CacheRouter {
   }
 
   static getRelayKey(args: { chainId: string; address: string }): string {
-    // Ensure address is checksummed to always have a consistent cache key
-    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${getAddress(args.address)}`;
+    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${args.address}`;
   }
 
   static getRelayCacheDir(args: {
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -1,4 +1,5 @@
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
+import { getAddress } from 'viem';
 
 export class CacheRouter {
   private static readonly ALL_TRANSACTIONS_KEY = 'all_transactions';
@@ -408,7 +409,8 @@ export class CacheRouter {
   }
 
   static getRelayKey(args: { chainId: string; address: string }): string {
-    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${args.address}`;
+    // Ensure address is checksummed to always have a consistent cache key
+    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${getAddress(args.address)}`;
   }
 
   static getRelayCacheDir(args: {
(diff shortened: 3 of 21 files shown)


src/datasources/cache/cache.router.ts
```diff
@@ -408,7 +409,8 @@ export class CacheRouter {
   }
 
   static getRelayKey(args: { chainId: string; address: string }): string {
-    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${args.address}`;
+    // Ensure address is checksummed to always have a consistent cache key
+    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${getAddress(args.address)}`;
```
iamacook, line 413: I'm not happy with this being located here but as the cache could theorertically be accessed from any module, I think it makes the most sense. Please let me know your thoughts.

src/datasources/relay-api/gelato-api.service.spec.ts
```diff
@@ -170,59 +126,6 @@ describe('GelatoApi', () => {
       ).rejects.toThrow();
     });
 
-    it('should increment the count after relaying', async () => {
```
iamacook, line 173: This is now done inside the service.

src/domain/relay/limit-addresses.mapper.spec.ts
```diff
@@ -66,7 +66,7 @@ describe('LimitAddressesMapper', () => {
       const safe = safeBuilder().build();
       const safeAddress = getAddress(safe.address);
       const data = execTransactionEncoder()
-        .with('to', getAddress(faker.finance.ethereumAddress()))
+        .with('value', faker.number.bigInt())
```
iamacook, line 69: This mock was incorrect. A `value` should be being sent but the builder defaults to `0`.

src/domain/relay/limit-addresses.mapper.spec.ts
```diff
@@ -85,7 +85,6 @@ describe('LimitAddressesMapper', () => {
       const safe = safeBuilder().build();
       const safeAddress = getAddress(safe.address);
       const data = execTransactionEncoder()
-        .with('to', getAddress(faker.finance.ethereumAddress()))
```
iamacook, line 88: This is redundant.

src/domain/relay/limit-addresses.mapper.spec.ts
```diff
@@ -306,6 +304,7 @@ describe('LimitAddressesMapper', () => {
       const safeAddress = getAddress(safe.address);
       const data = execTransactionEncoder()
         .with('to', safeAddress)
+        .with('value', faker.number.bigInt())
```
iamacook, line 307: As above regarding a `value` being needed.

src/domain/relay/limit-addresses.mapper.spec.ts
```diff
@@ -325,7 +324,6 @@ describe('LimitAddressesMapper', () => {
       const safe = safeBuilder().build();
       const safeAddress = getAddress(safe.address);
       const data = execTransactionEncoder()
-        .with('to', getAddress(faker.finance.ethereumAddress()))
```
iamacook, line 328: This is redundant.

src/routes/relay/relay.controller.spec.ts
```diff
@@ -0,0 +1,1646 @@
...
+            .expect(201)
+            .expect({
+              taskId,
+            });
+        });
+      });
+    });
+
+    describe('Transaction validation', () => {
+      // TODO: Return a 422 when the transactions are invalid
```
iamacook, line 887: I'm aware of this. I want to revisit it in a separate PR to not increase the complexity of this one.

iamacook: I've opened a [separate PR](https://github.com/safe-global/safe-client-gateway/pull/1154) implementing this.

src/datasources/cache/cache.router.ts
hectorgomezv: I'm more in favor of calling viem's `getAddress` on the domain, not here. I'd keep this class as clean as possible regarding dependencies. Otherwise we could also consider this check can applied to other functions in the this class handling eth addresses. I think we should assume the data coming from the domain is already validated/sanitized.

But besides having said that, it is not a strong opinion, I'm happy to see what you guys think @iamacook @fmrsabino

src/routes/relay/entities/relay.dto.entity.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export class RelayDto {
+  @ApiProperty()
+  to!: string;
+
+  @ApiProperty()
+  data!: string;
+
+  @ApiPropertyOptional({ type: BigInt, nullable: true })
+  gasLimit: string | undefined;
```
hectorgomezv, line 11: For entities nullable properties we are using `${type} | null`. I'm not sure this could have a lot of implications, but maybe it's more consistent to stick to that instead of `${type} | undefined`, what do you think? Also a default value in the schema could be considered 🤔

src/routes/relay/relay.controller.spec.ts
```diff
@@ -0,0 +1,1646 @@
...
+              to: safeAddress,
+              data,
+              gasLimit: gasLimit.toString(),
+            })
+            .expect(201)
+            .expect({
+              taskId,
+            });
+
+          // The gasLimit should have a buffer added
```
hectorgomezv, line 184: Not strictly related to this PR, but if I understand correctly, when the user specifies a custom `gasLimit`, 150k wei is added to the value the user specified. Do you think we should document this publicly in our API (Swagger maybe)?

src/routes/relay/relay.controller.spec.ts
```diff
@@ -0,0 +1,1646 @@
...
+            });
+        });
+
+        // execTransaction (execTransaction)
+        it('should return 201 calling execTransaction on a nested Safe', async () => {
+          const chainId = faker.helpers.arrayElement(supportedChainIds);
+          const chain = chainBuilder().with('chainId', chainId).build();
+          const safe = safeBuilder().build();
+          const safeAddress = getAddress(safe.address);
+          const data = execTransactionEncoder()
```
hectorgomezv, line 630: This is not intended to be tackled in this PR, it's more for a follow-up I guess, but it seems all the above test cases share the same setup except for the `data` property (depending on the transaction type). Not sure it is totally possible, but this seems a good candidate for parametrising the tests with `it.each`, by iterating on the `data` parameter. What do you think?

src/domain/relay/relay.repository.ts
```diff
@@ -56,7 +51,28 @@ export class RelayRepository {
...
+
+    // If we fail to increment count, we should not fail the relay
+    await Promise.allSettled(
+      relayAddresses.map((address) => {
+        return this.incrementRelayCount({
+          chainId: relayPayload.chainId,
+          address,
+        });
+      }),
+    );
```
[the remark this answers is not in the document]
hectorgomezv: This `Promise.allSettled` will call multiple times `incrementRelayCount` in parallel, right?  

As `incrementRelayCount` increments a counter, theoretically this could lead to a race condition. Maybe I'm missing something, but I'd change this implementation to:

```javascript
    for (const address of relayAddresses) {
      await this.incrementRelayCount({
        chainId: relayPayload.chainId,
        address,
      });
    }
```

hectorgomezv (changes requested):
Awesome work! Just small comments 👏🏻

src/datasources/cache/cache.router.ts
iamacook: I have moved the checksumming to the domain in a wrapper function in dc73902b4838bcb57f7a97a74f997cc572677a30.

src/domain/relay/relay.repository.ts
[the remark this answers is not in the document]
iamacook: Good point. I've implemented your suggestion in dc73902b4838bcb57f7a97a74f997cc572677a30.

src/config/entities/configuration.ts
```diff
@@ -127,6 +127,7 @@ export default () => ({
     email: process.env.FF_EMAIL?.toLowerCase() === 'true',
     zerionBalancesChainIds:
       process.env.FF_ZERION_BALANCES_CHAIN_IDS?.split(',') ?? [],
+    relay: process.env.FF_RELAY?.toLowerCase() === 'true',
```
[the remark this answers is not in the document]
iamacook: Added in dc73902b4838bcb57f7a97a74f997cc572677a30.

src/app.module.ts
```diff
@@ -48,7 +49,8 @@ export class AppModule implements NestModule {
   // into account. The .env file loading is done by the ConfigurationModule
   // which is not available at this stage.
   static register(configFactory = configuration): DynamicModule {
-    const isEmailFeatureEnabled = configFactory()['features']['email'];
+    const { email: isEmailFeatureEnabled, relay: isRelayFeatureEnables } =
```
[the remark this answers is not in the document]
iamacook: Fixed in dc73902b4838bcb57f7a97a74f997cc572677a30.

src/routes/relay/entities/relay.dto.entity.ts
iamacook: Changed in dc73902b4838bcb57f7a97a74f997cc572677a30.

src/routes/relay/relay.controller.spec.ts
iamacook: I'm not sure about the importance of this as the Gateway is inherently for internal use. Nonetheless, I've added a TODO to add it in a followup PR. There is a comment regarding the buffer in the `GelatoApi` as well.

src/routes/relay/relay.controller.spec.ts
iamacook: Refactored in dc73902b4838bcb57f7a97a74f997cc572677a30.

hectorgomezv: Thanks! 🙏🏻

hectorgomezv (approved):
👏🏻

PR #1154, opened 2024-02-16
Throw custom errors from relayer

iamacook:
This adds custom `Error`s with respective `ExceptionFilter`s for the following to the relayer:

- Invalid `multiSend`
- Invalid transfers
- Unofficial mastercopies
- Unofficial MultiSend contracts


src/domain/relay/errors/invalid-multisend.error.ts
```diff
@@ -0,0 +1,7 @@
+export class InvalidMultiSendError extends Error {
+  constructor() {
+    super(
+      'Invalid `multiSend` call. The batch is not all `execTransaction` calls to same address.',
```
fmrsabino, line 4: nitpick: I'd remove the ` as I assume this is mostly for rendering in markdown.

src/domain/relay/exception-filters/invalid-transfer.exception-filter.ts
```diff
@@ -0,0 +1,21 @@
...
+import { InvalidTransferError } from '@/domain/relay/errors/invalid-transfer.error';
+
+@Catch(InvalidTransferError)
+export class InvalidTransferExceptionFilter implements ExceptionFilter {
+  catch(_: InvalidTransferError, host: ArgumentsHost): void {
+    const ctx = host.switchToHttp();
+    const response = ctx.getResponse<Response>();
+
+    response.status(HttpStatus.UNPROCESSABLE_ENTITY).json({
+      message: 'Invalid transfer.',
```
fmrsabino, line 17: We can use the error message here as it seems to provide even more context. Same as with the others.

src/domain/relay/errors/unofficial-mastercopy.error.ts
```diff
@@ -0,0 +1,7 @@
+export class UnofficialMastercopyError extends Error {
```
fmrsabino, line 1: nitpick: `UnofficialMasterCopyError`

src/domain/relay/errors/invalid-multisend.error.ts
iamacook: Removed in 4a9d3787f8f6f122c9eb31a6e5e4cb00acc1c6e4.

src/domain/relay/exception-filters/invalid-transfer.exception-filter.ts
iamacook: Changed in 4a9d3787f8f6f122c9eb31a6e5e4cb00acc1c6e4.

src/domain/relay/errors/unofficial-mastercopy.error.ts
iamacook: Changed in 4a9d3787f8f6f122c9eb31a6e5e4cb00acc1c6e4.

src/datasources/cache/cache.router.ts
```diff
@@ -407,7 +408,8 @@ export class CacheRouter {
   }
 
   static getRelayKey(args: { chainId: string; address: string }): string {
-    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${args.address}`;
+    // Ensure address is checksummed to always have a consistent cache key
+    return `${args.chainId}_${CacheRouter.RELAY_KEY}_${getAddress(args.address)}`;
```
fmrsabino, line 412: Missed this one: I think this should be outside of the scope of this PR. Also, at this level we should be as close as possible to Redis primitives. Setting the key as a checksummed address should happen in outer layers (that handle more ethereum related logic).

fmrsabino (changes requested):

src/datasources/cache/cache.router.ts
iamacook: Agreed. This is a remnant from an older change that shouldn't be there. I manually removed it in 7c9718a097a267c29733988e72fb26f7f5971361.

fmrsabino (approved):
🚀

PR #1155, opened 2024-02-16
Add legacy relay service route

iamacook:
This adds the relay service routes to the relay controller for redirection purposes before clients migrate (as well as associated test coverage):

- `POST` `/v1/relay/:chainId`
- `GET` `/v1/relay/:chainId/:safeAddress`

### DIFF
--- a/src/routes/relay/relay.controller.ts
+++ b/src/routes/relay/relay.controller.ts
@@ -8,10 +8,7 @@ import { RelayDtoValidationPipe } from '@/routes/relay/pipes/relay.validation.pi
 @ApiTags('relay')
 @Controller({
   version: '1',
-  path: [
-    'relay/:chainId', // TODO: Remove after client has migrated
-    'chains/:chainId/relay',
-  ],
+  path: 'chains/:chainId/relay',
 })
 export class RelayController {
   constructor(private readonly relayService: RelayService) {}
--- a/src/routes/relay/relay.controller.ts
+++ b/src/routes/relay/relay.controller.ts
@@ -8,7 +8,10 @@ import { RelayDtoValidationPipe } from '@/routes/relay/pipes/relay.validation.pi
 @ApiTags('relay')
 @Controller({
   version: '1',
-  path: 'chains/:chainId/relay',
+  path: [
+    'relay/:chainId', // TODO: Remove after client has migrated
+    'chains/:chainId/relay',
+  ],
 })
 export class RelayController {
   constructor(private readonly relayService: RelayService) {}
--- a/src/routes/relay/relay.legacy.controller.spec.ts
+++ b/src/routes/relay/relay.legacy.controller.spec.ts
@@ -63,7 +63,7 @@ describe('Relay controller', () => {
           to: safeAddress,
           data,
         })
-        .expect(302)
+        .expect(308)
         .expect((res) => {
           expect(res.get('location')).toBe('/v1/chains/:chainId/relay');
         });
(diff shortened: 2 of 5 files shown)


src/routes/relay/relay.controller.ts
```diff
@@ -8,7 +8,10 @@ import { RelayDtoValidationPipe } from '@/routes/relay/pipes/relay.validation.pi
 @ApiTags('relay')
 @Controller({
   version: '1',
-  path: 'chains/:chainId/relay',
+  path: [
+    'relay/:chainId', // TODO: Remove after client has migrated
+    'chains/:chainId/relay',
+  ],
```
fmrsabino, line 14: Having a different controller for the legacy route can be more beneficial to separate concerns (legacy vs non-legacy routes). We can accomplish this with a `@Redirect` which would return a 302 by default to the clients signalling that the resource was moved which is also what we want to signal.


```typescript
@Controller({
  version: '1',
  path: 'relay/:chainId',
})
export class RelayLegacyController {
  @Post()
  @Redirect('/v1/chains/:chainId/relay')
  relay(): void {}

  @Get(':safeAddress')
  @Redirect('/v1/chains/:chainId/relay/:safeAddress')
  getRelaysRemaining(): void {
  }
}
```

fmrsabino (changes requested):

src/routes/relay/relay.controller.ts
iamacook: Changed in 4a94684980e5829ef6d81cf00a298c4d22ee7027.

src/routes/relay/relay.legacy.controller.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+      const safeAddress = faker.finance.ethereumAddress();
+      const data = faker.string.hexadecimal();
+
+      await request(app.getHttpServer())
+        .post(`/v1/relay/${chainId}`)
+        .send({
+          to: safeAddress,
+          data,
+        })
+        .expect(302)
```
fmrsabino, line 66: We should probably switch this to [308 Permanent Redirect](https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#permanent_redirections). Right now it signals a temporary one which is not the case for these routes.

Also 302 creates the expectation that the clients should follow-up on the redirect with a `GET` which is not the case for this route.

src/routes/relay/relay.legacy.controller.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+    });
+  });
+  describe('GET /v1/relay/:chainId/:safeAddress', () => {
+    it('should return 302 and redirect to the new endpoint', async () => {
+      const chainId = faker.string.numeric();
+      const safeAddress = faker.finance.ethereumAddress();
+
+      await request(app.getHttpServer())
+        .get(`/v1/relay/${chainId}/${safeAddress}`)
+        .expect(302)
```
fmrsabino, line 79: Given my comment above we can use [301 Moved Permanently](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/301) given that we are dealing with a `GET` request.

src/routes/relay/relay.legacy.controller.spec.ts
iamacook: Changed in 3f2db35fc53e88dc0419f8b40fba9a327e61c49b.

src/routes/relay/relay.legacy.controller.spec.ts
iamacook: Changed in 3f2db35fc53e88dc0419f8b40fba9a327e61c49b.

fmrsabino (approved):
👏

PR #1163, opened 2024-02-19
Mark `isTestnet` as required

iamacook:
This marks `Chain['isTestnet']` as required now that the [clients are using it successfully](https://github.com/safe-global/safe-wallet-web/blob/1be52b50d8f79eef46e2e5da621442a87f0b0234/src/components/common/NetworkSelector/index.tsx#L40).


fmrsabino (approved):
🚀

PR #1164, opened 2024-02-19
Send unsubscription token when sending email

fmrsabino:
Sends a `unsubscriptionUrl` substitution to the email provider. This URL includes the token which can be used to unsubscribe from further email notifications.

Since this token is unique, and a token cannot be mapped to a target email address if multiple email addresses are involved, the `AlertsRepository` now triggers multiple calls to `IEmailApi.createMessage` (one per each recipient).

### DIFF
--- a/src/domain/alerts/alerts.repository.ts
+++ b/src/domain/alerts/alerts.repository.ts
@@ -238,24 +238,37 @@ export class AlertsRepository implements IAlertsRepository {
     accountsToNotify: Account[];
   }): Promise<void> {
     const chain = await this.chainRepository.getChain(args.chainId);
-
     const webAppUrl = this.urlGenerator.addressToSafeWebAppUrl({
       chain,
       safeAddress: args.safeAddress,
     });
 
-    const emails = args.accountsToNotify.map(
-      (account) => account.emailAddress.value,
-    );
-    return this.emailApi.createMessage({
-      to: emails,
-      template: this.configurationService.getOrThrow<string>(
-        'email.templates.unknownRecoveryTx',
-      ),
-      subject: AlertsRepository.UNKNOWN_TX_EMAIL_SUBJECT,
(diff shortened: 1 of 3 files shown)


src/domain/alerts/alerts.repository.ts
```diff
@@ -238,24 +238,37 @@ export class AlertsRepository implements IAlertsRepository {
...
+        },
+      });
+    });
+
+    await Promise.allSettled(emailActions).then((results) => {
+      results.forEach((result, index) => {
+        if (result.status === 'rejected') {
+          const signer = args.accountsToNotify.at(index)?.signer;
+          this.loggingService.warn(
+            `Error sending email to user with account ${signer}, for safe ${args.safeAddress} on chain ${args.chainId}`,
```
iamacook, line 268: Nit:

```suggestion
            `Error sending email to user with account ${signer}, for Safe ${args.safeAddress} on chain ${args.chainId}`,
```

src/domain/alerts/alerts.repository.ts
```diff
@@ -280,20 +293,34 @@ export class AlertsRepository implements IAlertsRepository {
...
+        },
+      });
+    });
+
+    await Promise.allSettled(emailActions).then((results) => {
+      results.forEach((result, index) => {
+        if (result.status === 'rejected') {
+          const signer = args.accountsToNotify.at(index)?.signer;
+          this.loggingService.warn(
+            `Error sending email to user with account ${signer}, for safe ${args.newSafeState.address} on chain ${args.chainId}`,
```
iamacook, line 320:
```suggestion
            `Error sending email to user with account ${signer}, for Safe ${args.newSafeState.address} on chain ${args.chainId}`,
```

iamacook (approved):

iamacook (approved):

PR #1175, opened 2024-02-20
Push all SafeBalancesApi related logic to the datasources layer

hectorgomezv:
### Summary
This is a follow-up of https://github.com/safe-global/safe-client-gateway/pull/1151

With the introduction of `SafeBalancesApi` the balances retrieval architecture was simplified by routing all the balances and collectibles providers/fallbacks through a single point of responsibility (`BalancesApiManager`). Nevertheless, some pieces were left out of the scope of the original PR, as `PricesRepository` was left as an independent piece in the domain layer, and some CoinGecko-specific logging was still present in the domain as well. This refactor tries to address those problems.

### Changes
- Simplify `CollectiblesRepository` and `BalancesRepository` by pushing provider-specific logic to the data sources layer.
- Simplify `BalancesApiManager`: `useExternalApi` is no longer exposed outside the class. All the routing logic is encapsulated.
- `PricesApi` was renamed as `CoingeckoApi`, as it’s the unique provider for token prices. It was also moved to the data sources layer as it’s not exposed to the domain. It’s only used by `SafeBalancesApi`.
- Safe balances cache keys were renamed by adding a `safe` prefix.

PR #1178, opened 2024-02-20
Move prices provider configuration schema

hectorgomezv:
Depends on #1175 

### Summary
This is a follow up of #1175 

This comment (https://github.com/safe-global/safe-client-gateway/pull/1149#discussion_r1490874638) served as an inspiration for this change. Of course, any feedback/change is very welcome.

### Changes
Moves `prices.*` configuration key to the `balances.` key tree. Since the prices retrieval is now isolated into `SafeBalancesApi`, it makes no sense to have the configuration of the Coingecko prices provider as an independent key.


src/config/entities/__tests__/configuration.ts
```diff
@@ -21,6 +21,80 @@ export default (): ReturnType<typeof configuration> => ({
   balances: {
     balancesTtlSeconds: faker.number.int(),
     providers: {
+      safe: {
```
iamacook, line 24: Is there a particular reason to name this `safe`? We don't nest providers in other configurations, e.g. `relay`.

hectorgomezv: It was an intent to align with the `BalancesApiManager` strategy and have the providers (`SafeBalancesApi` and `ZerionBalancesApi` under the `balances` tree. 

Regarding the nesting, we could also omit the `providers` key and have just `balances.zerion` and `balances.safe` (or another key instead of `safe` if you find it confusing. What do you think?

iamacook: I'm happy to keep it as. Thank you for the clarification!

iamacook (dismissed):

iamacook (approved):

PR #1184, opened 2024-02-21
Fix `RelayDto['gasLimit']` Swagger type

iamacook:
This changes the Swagger type for `RelayDto['gasLimit']` from `BigInt` to `String`.

### DIFF
--- a/src/routes/relay/entities/relay.dto.entity.ts
+++ b/src/routes/relay/entities/relay.dto.entity.ts
@@ -11,7 +11,7 @@ export class RelayDto {
   data!: string;
 
   @ApiPropertyOptional({
-    type: BigInt,
+    type: String,
     nullable: true,
     description: `If specified, a gas buffer of 150k will be added on top of the expected gas usage for the transaction.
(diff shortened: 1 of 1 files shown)


src/routes/relay/entities/relay.dto.entity.ts
```diff
@@ -11,7 +11,7 @@ export class RelayDto {
   data!: string;
 
   @ApiPropertyOptional({
-    type: BigInt,
+    type: String,
```
fmrsabino, line 14: Is `BigInt` not supported?

iamacook: It accepts it type-wise but the request is receiving a `String`, not a `BigInt`.

fmrsabino: We should consider maybe mapping this to a a `BigInt` before it reaches our domain. Non-numerical values should not work with the relay.

fmrsabino (approved):

src/routes/relay/entities/relay.dto.entity.ts
iamacook: Good point. I'll investigate it in a separate PR.

PR #1185, opened 2024-02-21
Add bug and feature issue templates

iamacook:
This adds two issue templates:

- Bug report
- Feature request

Adding these templates prevents "blank" issues from being created but this can be adjusted by adding a config file should we deem it necessary.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug-report.yml
+++ b/.github/ISSUE_TEMPLATE/bug-report.yml
@@ -1,0 +1,32 @@
+name: Bug report
+description: Create a new issue reporting a bug
+body:
+  - type: textarea
+    id: description
+    attributes:
+      label: Description
+      description: A clear and concise description of what the bug is.
+    validations:
+      required: true
+  - type: textarea
+    id: reproduction-steps
+    attributes:
+      label: Reproduction steps
+      description: Steps to reproduce the behavior.
+      placeholder: 1. ...
+        2. ...
+        3. ...
+    validations:
+      required: true
+  - type: textarea
+    id: expected-behavior
+    attributes:
+      label: Expected behavior
+      description: What did you expect to happen?
+    validations:
+      required: true
+  - type: textarea
+    id: additional-info
+    attributes:
+      label: Additional information
+      description: Add any other context about the problem here e.g. screenshots, logs, etc.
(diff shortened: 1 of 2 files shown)


.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,32 @@
+name: Bug report
+description: Create a new issue reporting a bug
```
fmrsabino, line 2: What do you think about adding the bug label here?

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,32 @@
...
+    attributes:
+      label: Expected behavior
+      description: What did you expect to happen?
+    validations:
+      required: true
+  - type: textarea
+    id: additional-info
+    attributes:
+      label: Additional information
+      description: Add any other context about the problem here e.g. screenshots, logs, etc.
```
fmrsabino, line 32: We can consider adding the Chain ID and Safe Address as optional fields. I'd say for now it's ok to have it as free-form text though and we can check the patterns present in future reports.

.github/ISSUE_TEMPLATE/bug-report.yml
```diff
@@ -0,0 +1,32 @@
...
+    validations:
+      required: true
+  - type: textarea
+    id: reproduction-steps
+    attributes:
+      label: Reproduction steps
+      description: Steps to reproduce the behavior.
+      placeholder: 1. ...
+        2. ...
+        3. ...
```
fmrsabino, line 18: I think you might need to use `|` for the newlines here. See https://yaml.org/spec/1.2.2/#812-literal-style

.github/ISSUE_TEMPLATE/bug-report.yml
iamacook: It's a good idea! Added in fbaf003ea96a97d534ed5084e197825d5539df21.

.github/ISSUE_TEMPLATE/bug-report.yml
iamacook: Agreed. I think it's best to keep it as simple as possible for the time being as well.

.github/ISSUE_TEMPLATE/bug-report.yml
iamacook: Added in fbaf003ea96a97d534ed5084e197825d5539df21 according to [GitHub's example](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms#yaml-issue-form-template).

fmrsabino (approved):
🚀

PR #1189, opened 2024-02-21
Test each chain depending on supported contract versions

iamacook:
This adds a chain-contract version map for each relevant contract and loops over the relay tests accordingly.

### DIFF
--- a/src/routes/relay/relay.controller.spec.ts
+++ b/src/routes/relay/relay.controller.spec.ts
@@ -46,11 +46,27 @@ import {
 } from '@safe-global/safe-deployments';
 import { createProxyWithNonceEncoder } from '@/domain/relay/contracts/__tests__/proxy-factory-encoder.builder';
 
-const SAFE_VERSIONS = ['1.0.0', '1.1.1', '1.2.0', '1.3.0', '1.4.1'];
-const SAFE_L2_VERSIONS = ['1.3.0', '1.4.1'];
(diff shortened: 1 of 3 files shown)


src/routes/relay/relay.controller.spec.ts
```diff
@@ -58,7 +58,14 @@ describe('Relay controller', () => {
...
   let relayUrl: string;
-  const supportedChainIds = Object.keys(configuration().relay.apiKey);
+
+  // Remove Sepolia from tests as only >=1.3.0 are deployed there
+  const supportedChainIds = Object.keys(configuration().relay.apiKey).filter(
+    (chainId) => {
+      const sepoliaChainId = '11155111';
+      return chainId !== sepoliaChainId;
+    },
+  );
```
fmrsabino, line 68: I think it's valuable that we test every chain and this points out to a different issue where the version mapping is dependent of the chain (and we don't have that right now).

Right now it's Sepolia that is missing one version but we can potentially have a broad range of supported chains which by themselves support different contract versions.

To unblock the pipeline I'd say this fix is fine but we should consider adjusting the way we test each chain/safe version.

fmrsabino (approved):

src/routes/relay/relay.controller.spec.ts
iamacook: As I had capacity, I adjusted the test to loop over the entire suite for each chain in 9e70f16d39f7bef304d5c94e128d5c1a6bf316a2.

hectorgomezv (approved):

PR #1192, opened 2024-02-22
Make Page count property nullable

hectorgomezv:
### Summary
Some data providers (e.g.: Safe Transaction Service) provide a `count` attribute within the response payload when requesting a `Page` of data. This attribute is used in the CGW to build the `next` cursor attribute.

But some other providers don't provide this `count` (e.g.: Zerion). In order to bypass this limitation, the CGW was returning the amount of items on the page, but this is not correct (see https://github.com/safe-global/safe-client-gateway/blob/18f400dee910866e3be482595159a23a226b1c0d/src/datasources/balances-api/zerion-balances-api.service.ts#L142).

### Changes
- Allows `null` values on `Page.count` attribute.


iamacook (approved):
We will need to propagate this change to the SDK. cc @safe-global/safe-web

PR #1194, opened 2024-02-22
Coerce `RelayDto['gasLimit']` to `bigint`

iamacook:
This coerces `RelayDto['gasLimit'` to a `bigint` is specified and initially validated within the service so that the domain needs not adjust the type.

### DIFF
--- a/src/routes/relay/relay.controller.module.ts
+++ b/src/routes/relay/relay.controller.module.ts
@@ -3,10 +3,11 @@ import { RelayDomainModule } from '@/domain/relay/relay.domain.module';
 import { RelayService } from '@/routes/relay/relay.service';
 import { RelayController } from '@/routes/relay/relay.controller';
 import { RelayLegacyController } from '@/routes/relay/relay.legacy.controller';
+import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
 
 @Module({
   imports: [RelayDomainModule],
-  providers: [RelayService],
+  providers: [HttpErrorFactory, RelayService],
   controllers: [RelayController, RelayLegacyController],
 })
 export class RelayControllerModule {}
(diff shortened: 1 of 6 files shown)


src/routes/relay/relay.service.ts
```diff
@@ -11,6 +16,7 @@ export class RelayService {
   constructor(
     @Inject(IConfigurationService) configurationService: IConfigurationService,
     private readonly relayRepository: RelayRepository,
+    private readonly httpErrorFactory: HttpErrorFactory,
```
fmrsabino, line 19: We don't need this injected.

src/routes/relay/relay.controller.module.ts
```diff
@@ -3,10 +3,11 @@ import { RelayDomainModule } from '@/domain/relay/relay.domain.module';
 import { RelayService } from '@/routes/relay/relay.service';
 import { RelayController } from '@/routes/relay/relay.controller';
 import { RelayLegacyController } from '@/routes/relay/relay.legacy.controller';
+import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
 
 @Module({
   imports: [RelayDomainModule],
-  providers: [RelayService],
+  providers: [HttpErrorFactory, RelayService],
```
fmrsabino, line 10: If the `httpErrorFactory` is removed from the `RelayService` we can also remove this provider.

fmrsabino (changes requested):

src/routes/relay/relay.service.ts
iamacook: Removed in 7bd3e3d48d8d989ace3f9a8163817e9662ebb736.

src/routes/relay/relay.controller.module.ts
iamacook: Removed in 7bd3e3d48d8d989ace3f9a8163817e9662ebb736.

fmrsabino (approved):

PR #1195, opened 2024-02-23
Generate chain-specific deployment versions for tests

iamacook:
This adds a function to generate type safe mappings of chain IDs to deployment versions from the `@safe-global/safe-deployments` package e.g.

```ts
const versions = getDeploymentVersionsByChainIds(
  'CompatibilityFallbackHandler',
  ['100', '11155111'],
);

expect(versions).toStrictEqual({
  '100': ['1.3.0', '1.4.1'],
  '11155111': ['1.3.0', '1.4.1'],
});
```

### DIFF
--- a/src/__tests__/deployments.helper.spec.ts
+++ b/src/__tests__/deployments.helper.spec.ts
@@ -1,0 +1,125 @@
+import { getDeploymentVersionsByChainIds } from '@/__tests__/deployments.helper';
+
+describe('Deployments helper', () => {
+  describe('getDeploymentVersionsByChainIds', () => {
+    it('should return all CompatibilityFallbackHandler versions for the specified chain ID', () => {
+      const versions = getDeploymentVersionsByChainIds(
+        'CompatibilityFallbackHandler',
+        ['100', '11155111'],
+      );
+
+      expect(versions).toStrictEqual({
+        '100': ['1.3.0', '1.4.1'],
+        '11155111': ['1.3.0', '1.4.1'],
+      });
+    });
+
+    it('should return all CreateCall versions for the specified chain ID', () => {
+      const versions = getDeploymentVersionsByChainIds('CreateCall', [
+        '100',
+        '11155111',
+      ]);
+
+      expect(versions).toStrictEqual({
+        '100': ['1.3.0', '1.4.1'],
+        '11155111': ['1.3.0', '1.4.1'],
+      });
+    });
+
+    it('should return all DefaultCallbackHandler versions for the specified chain ID', () => {
+      const versions = getDeploymentVersionsByChainIds(
+        'DefaultCallbackHandler',
+        ['100', '11155111'],
+      );
+
+      expect(versions).toStrictEqual({
+        '100': ['1.1.1'],
+        '11155111': [],
+      });
+    });
+
+    it('should return all MultiSend for the specified chain ID', () => {
+      const versions = getDeploymentVersionsByChainIds('MultiSend', [
+        '100',
+        '11155111',
+      ]);
+
+      expect(versions).toStrictEqual({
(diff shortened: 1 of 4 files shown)


src/__tests__/deployments.helper.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  return versionsByChain;
+}
+
+export function getDeploymentVersionsByChainIds(
+  contractAlias: keyof typeof deploymentAliases,
+  chainIds: Array<string>,
+): Record<string, Array<string>> {
+  return chainIds.reduce<Record<string, Array<string>>>((acc, chainId) => {
+    acc[chainId] = getDeploymentVersionsByChainId(contractAlias, chainId);
```
fmrsabino, line 105: For each chain we are parsing all the asset files even though they should not change which can make this call potentially more expensive that it should be.

src/__tests__/deployments.helper.spec.ts
```diff
@@ -0,0 +1,125 @@
+import { getDeploymentVersionsByChainIds } from '@/__tests__/deployments.helper';
+
+describe('Deployments helper', () => {
+  describe('getDeploymentVersionsByChainIds', () => {
+    it('should return all CompatibilityFallbackHandler versions for the specified chain ID', () => {
+      const versions = getDeploymentVersionsByChainIds(
```
fmrsabino, line 6: For each test, we parse a JSON file `N` times (where `N` is the number of asset files which does not change). My suggestion is to do an initial test setup where we get the deployments object and then each test asserts the respective contents.

fmrsabino (changes requested):

src/__tests__/deployments.helper.ts
iamacook: I changed this in f30d5ba9b9b4bfcd7ee92e7e7cfa9930c275ef79 to generate a singleton map of contracts-chain-versions:

```ts
type VersionsByChainIdByDeploymentMap = {
  [contractAlias: string]: {
    [chainId: string]: string[] | undefined
  };
};
```

Please let me know what you think.

src/__tests__/deployments.helper.spec.ts
iamacook: I've kept this function as is but it now references a singleton map of contracts-chain-versions. See my [previous comment](https://github.com/safe-global/safe-client-gateway/pull/1195#discussion_r1502480176) for more information.

src/__tests__/deployments.helper.ts
```diff
@@ -0,0 +1,126 @@
...
+  }
+
+  return versionsByDeploymentByChainIdSingleton;
+}
+
+export function getDeploymentVersionsByChainIds(
+  contractAlias: keyof typeof deploymentAliases,
+  chainIds: Array<string>,
+): Record<string, Array<string>> {
+  const map = getVersionsByChainIdByDeploymentMap();
```
fmrsabino, line 121: I think we should store this `map` on a global scope. This would only be loaded for testing purposes and avoid parsing the file multiple times on different `contractAlias`.

fmrsabino (changes requested):

src/__tests__/deployments.helper.ts
iamacook: Each call [was returning the singleton](https://github.com/safe-global/safe-client-gateway/pull/1195/commits/f30d5ba9b9b4bfcd7ee92e7e7cfa9930c275ef79#diff-8f7da0203c50595cd5516d468648a8a1d77aa9cb26e979339ad55c0429c01c1eR66-R72) so parsing only happened once. Nonetheless, it is instead called globally as of 83a4669872b91a720529fb3afb8608a35835f4f3.

fmrsabino (approved):

PR #1203, opened 2024-02-26
Validate `RelayDto` using `zod`

iamacook:
## Summary

This is a proof of concept using `zod` over `ajv` for validation. The main reasons being:

- type refinement/coercion
- automatic type inferral

In order to use `zod` with NestJS, a custom `ZodValidationPipe` is necessary. Whilst there are third party packages that achieve this, this custom pipe is a "bare bones" implementation.

The demonstrative focus of this demo is the `RelayDtoSchema`:

- `to`, the string is checked for validity (is an Ethereum address) and coerced to be checksummed.
- `data`, the string is checked for validity (is hexadecimal).
- `version`, the string is checked for validity (is SemVer).
- `gasLimit`, the string is checked for validity (either `undefined` or a numerical string) and coerced to `null` or `BigInt`.

Using the `RelayDtoSchema`, the `RelayDto` types are automatically inferred with the aforementioned coerctions by using `zod`'s `z.infer` utility type.

```ts
  export class RelayDto implements z.infer<typeof RelayDtoSchema> {
    ...
  }
```

It is important to note that we already have `zod` installed as a peer dependency so this will not increase our bundle size. Should we want to opt for `zod` over `ajv`, we could sequentially migrate to it.

## Changes

- `zod` installed and TypeScript adjusted to enable `strictNullChecks` (a pre-requisite)
- New schemas with tests added
  - `AddressSchem` to validate and checmsum addresses
  - `HexSchema` to validate hexadecimal strings
  - `RelayDtoSchema` to validate `RelayDto` and coerce accordingly
- `ZodValidationPipe` created and added to route
- `ZodErrorFilter` to map validation errors
- Associated relay code/tests updated


tsconfig.json
```diff
@@ -21,5 +21,6 @@
       "@/*": ["src/*"],
     },
     "strict": true,
+    "strictNullChecks": true,
```
iamacook, line 24: This is a pre-requisite for `zod`.

src/validation/entities/schemas/__tests__/hex.schema.spec.ts
```diff
@@ -0,0 +1,24 @@
...
+describe('HexSchema', () => {
+  it('should validate a valid hex string', () => {
+    const value = faker.string.hexadecimal();
+
+    const result = HexSchema.safeParse(value);
+
+    expect(result.success).toBe(true);
+    if (!result.success) {
+      throw new Error('HexSchema failed to validate a valid hex string.');
+    }
```
hectorgomezv, line 13: Are this kind of assertions (here and also in `address.schema.spec.ts` and `relay.dto.schema.spec.ts`) needed? I mean, if `result.success` is `false` Jest will inform of the failure.

src/routes/relay/entities/schemas/__tests__/relay.dto.schem.spec.ts
```diff
@@ -0,0 +1,133 @@
+import { RelayDtoSchema } from '@/routes/relay/entities/schemas/relay.dto.schema';
```
hectorgomezv, line 1: Nit: there is a typo in the filename (`schem.spec.ts`) 😄

src/routes/common/filters/zod-error.filter.ts
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * This {@link ExceptionFilter} catches any {@link ZodError} thrown
+ * at the route level.
+ *
+ * It builds a JSON payload which contains a code and a message.
+ * The code and message used are read from {@link ZodError} issues.
+ */
+@Catch(ZodError)
+export class ZodErrorFilter implements ExceptionFilter {
```
hectorgomezv, line 18: Nit: what about renaming this class and `ZodValidationPipe` to remove the `Zod` reference from the name? Ideally the caller would use a "validation" filter/pipe without knowing the underlying library implementing the validation checks. (No strong opinion)

src/validation/entities/schemas/__tests__/hex.schema.spec.ts
iamacook: Yes. The types of `zod` are very strict, e.g.

```ts
type Result = { success: true, data: ... } | { success: false, error: ... }
```

They're needed to appease TypeScript:

image

src/routes/relay/entities/schemas/__tests__/relay.dto.schem.spec.ts
iamacook: 🦅 Fixed in 65586e04007c76e036ebaa09a03c5e124d417e82.

src/routes/common/filters/zod-error.filter.ts
iamacook: I've renamed the pipe to `ValidationPipe` but think we should keep the filter as is because of the error type. What do you think?

src/validation/entities/schemas/__tests__/hex.schema.spec.ts
hectorgomezv: I see, thanks for the explanation. This is not a strong opinion at all, but wdyt about:
```javascript
expect(result.success && result.data).toBe(value);
```
Instead of:
```javascript
expect(result.success).toBe(true);
if (!result.success) {
  throw new Error('HexSchema failed to validate a valid hex string.');
}
expect(result.data).toBe(value);
```

I'm just saving some lines which it seems would get repeated in many tests, but this is absilutely nitpick, I'm fine with your code 👏🏻

src/routes/common/filters/zod-error.filter.ts
hectorgomezv: Hmm, yes, makes sense as it is more specific and tied to the library 👍 Thanks!

hectorgomezv (approved):

PR #1215, opened 2024-02-27
Refactor cache-write functions

hectorgomezv:
## Summary
Before this PR, `ICacheService.set` had two main implementations in the service: `FakeCacheService` and `RedisCacheService`.

The implementation of these classes was different. For `FakeCacheService`:
- If the optional `expireTimeSeconds` parameter was **not set**, the value provided was set with no TTL.
- If the optional `expireTimeSeconds` parameter was **set**, the value provided was set with no TTL.
So, in `FakeCacheService`, the optional parameter `expireTimeSeconds` was meaningless, the implementation was not taking it into consideration.

### DIFF
--- a/src/datasources/cache/__tests__/fake.cache.service.spec.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.spec.ts
@@ -15,8 +15,9 @@ describe('FakeCacheService', () => {
       faker.string.alphanumeric(),
     );
     const value = faker.string.alphanumeric();
+    const expireTimeSeconds = faker.number.int({ min: 1 });
 
-    await target.set(cacheDir, value, 0);
+    await target.setWithExpiration(cacheDir, value, expireTimeSeconds);
 
     await expect(target.get(cacheDir)).resolves.toBe(value);
     expect(target.keyCount()).toBe(1);
@@ -29,8 +30,9 @@ describe('FakeCacheService', () => {
     const field = faker.string.alphanumeric();
     const cacheDir = new CacheDir(key, field);
     const value = faker.string.alphanumeric();
+    const expireTimeSeconds = faker.number.int({ min: 1 });
 
-    await target.set(cacheDir, value, 0);
+    await target.setWithExpiration(cacheDir, value, expireTimeSeconds);
     await target.deleteByKey(key);
 
     await expect(target.get(cacheDir)).resolves.toBe(undefined);
@@ -39,19 +41,5 @@ describe('FakeCacheService', () => {
     ).resolves.toBe(now.toString());
     expect(target.keyCount()).toBe(1);
     jest.useRealTimers();
-  });
-
-  it('clears keys', async () => {
-    const actions: Promise<void>[] = [];
-    for (let i = 0; i < 5; i++) {
-      actions.push(
-        target.set(new CacheDir(`key${i}`, `field${i}`), `value${i}`, 0),
-      );
-    }
-
-    await Promise.all(actions);
-    target.clear();
-
-    expect(target.keyCount()).toBe(0);
   });
 });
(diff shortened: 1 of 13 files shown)


src/datasources/cache/__tests__/fake.cache.service.spec.ts
```diff
@@ -15,8 +15,9 @@ describe('FakeCacheService', () => {
       faker.string.alphanumeric(),
     );
     const value = faker.string.alphanumeric();
+    const expireTimeSeconds = faker.number.int({ min: 1 });
 
-    await target.set(cacheDir, value, 0);
+    await target.setWithExpiration(cacheDir, value, expireTimeSeconds);
```
iamacook, line 20: Nit: let's update the description of this test to reflect that it's with expiration. We should also consider adding a test for `set`. What do you think?

hectorgomezv: Good point! Done in [ec1c982](https://github.com/safe-global/safe-client-gateway/pull/1215/commits/ec1c982243db33ed394ab687d06063e6e0cfbcc3)

src/datasources/cache/redis.cache.service.ts
```diff
@@ -31,7 +31,7 @@ export class RedisCacheService
     return this.client.ping();
   }
 
-  async set(
+  async setWithExpiration(
```
fmrsabino, line 34: By splitting the parameters of the `SET` operation into multiple functions, we increase the number of functions that we need to implement in case we want to extend its functionality.

If we need to check if a specific component was called with a certain number of parameters, a mocked object might be a better fit for this (while we can still have an integration test with the actual implementation to check if the TTL was set).

See SET available options – https://redis.io/commands/set/

hectorgomezv: The gotcha here is we are not using `SET` but `HSET`, which does not provide these options to manage the TTL. But this is not relevant anymore as I think the PR you merged (https://github.com/safe-global/safe-client-gateway/pull/1219) solves this problem accurately by passing the `NX` option into the `expire` call 🙂

I will reset this PR and use it exclusively to align our `ICacheService` implementations.

src/datasources/cache/__tests__/fake.cache.service.spec.ts
```diff
@@ -16,7 +16,7 @@ describe('FakeCacheService', () => {
     );
     const value = faker.string.alphanumeric();
 
-    await target.set(cacheDir, value, 0);
+    await target.set(cacheDir, value, faker.number.int());
```
iamacook, line 19: If `0` is given as the `expireTimeSeconds`, it [shouldn't set the key](https://github.com/safe-global/safe-client-gateway/blob/80134e6ff4bf3dc190988b3485db3803a76d5bd3/src/datasources/cache/__tests__/fake.cache.service.ts#L45-L47) and `faker.number.int` [can return `0`](https://fakerjs.dev/api/number.html#int). We should use a `min` of `1` for these examples.

src/datasources/cache/cache.first.data.source.spec.ts
```diff
@@ -58,6 +58,7 @@ describe('CacheFirstDataSource', () => {
       cacheDir,
       url: targetUrl,
       notFoundExpireTimeSeconds,
+      expireTimeSeconds: faker.number.int(),
```
iamacook, line 61: As before regarding a `min` option.

hectorgomezv: Fixed in [1676ade](https://github.com/safe-global/safe-client-gateway/pull/1215/commits/1676ade45f02da84872f938601a84b03163a93b0)

PR #1216, opened 2024-02-27
Improve `NetworkResponseError` error message propagation

iamacook:
## Summary

This improves the propagation of `NetworkResponseError` messages in the following order:

1. Django `nonFieldErrors[0]`
2. Standard error `message`
3. Response `statusText`
4. Default of `"An error occurred"

## Changes

- Adds new `HttpErrorFactory['getNetworkResponseErrorMessage']` method returning the above "ranked" error messages.

## Notes

The above can be tested by trying to delete transactions that were proposed _before_ deleteion was possible.


iamacook:
I'm going to convert this to a draft as there is ongoing discussion regarding the error standards.

iamacook:
I'm going to close this as it is superseeded by https://github.com/safe-global/safe-client-gateway/pull/1226 (where we map the error inside the relevant service instead).

PR #1218, opened 2024-02-27
Normalize Ethereum addresses

fmrsabino:
- Normalizes Ethereum addresses in the `AccountRepository` – this avoids that the same address is represented in different ways (thus resulting in duplicate entries).
- The normalization step for Ethereum addresses results in a checksummed address.
- The `InvalidAddressError` error mapping was added to `EmailService` – an exception filter was considered but this results in the wrong status code being logged in `RouteLoggerInterceptor` since it's neither an instance of `HttpException` nor  `DataSourceError`.

### DIFF
--- a/src/datasources/account/account.datasource.ts
+++ b/src/datasources/account/account.datasource.ts
@@ -9,6 +9,7 @@ import {
 } from '@/domain/account/entities/account.entity';
 import { Subscription as DomainSubscription } from '@/domain/account/entities/subscription.entity';
 import { VerificationCodeDoesNotExistError } from '@/datasources/account/errors/verification-code-does-not-exist.error';
+import { getAddress } from 'viem';
 
 interface Account {
   id: number;
@@ -39,8 +40,8 @@ export class AccountDataSource implements IAccountDataSource {
 
   async getAccount(args: {
     chainId: string;
-    safeAddress: string;
-    signer: string;
+    safeAddress: `0x${string}`;
+    signer: `0x${string}`;
   }): Promise<DomainAccount> {
     const [account] = await this.sql<Account[]>`SELECT *
                                                 FROM accounts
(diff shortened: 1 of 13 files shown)


src/datasources/account/account.datasource.ts
```diff
@@ -59,40 +60,39 @@ export class AccountDataSource implements IAccountDataSource {
       chainId: account.chain_id.toString(),
       emailAddress: new EmailAddress(account.email_address),
       isVerified: account.verified,
-      safeAddress: account.safe_address,
-      signer: account.signer,
+      safeAddress: getAddress(account.safe_address),
+      signer: getAddress(account.signer),
```
iamacook, line 64: Theoretically, all addresses returned in this data source should be checksummed. Is there a reason we checksum them again?

src/routes/email/email.controller.delete-email.spec.ts
```diff
@@ -74,16 +74,20 @@ describe('Email controller delete email tests', () => {
     await app.close();
   });
 
-  it('deletes email successfully', async () => {
+  it.each([
+    // non-checksummed address
+    { safeAddress: faker.finance.ethereumAddress() },
```
iamacook, line 79: Nit: what do you think about calling `toLowerCase` to be explicit? (And in case it ever changes to return a checksummed address.)

src/datasources/account/account.datasource.ts
fmrsabino: That's correct. However in order to comply with [the Account  interface](https://github.com/safe-global/safe-client-gateway/blob/775159f7348c78adeb24744dbff8118a93ce8ea2/src/domain/account/entities/account.entity.ts#L11-L12), it does need to be converted to `0x{string}`.

Checksumming the address might prove to be more costly than required indeed however we would still need to validate that it's indeed a `0x{string}` 🤔

src/routes/email/email.controller.delete-email.spec.ts
fmrsabino: Makes total sense. Added [1ad54f7](https://github.com/safe-global/safe-client-gateway/pull/1218/commits/1ad54f722841490e3d93ca861b0e730021783d19).

iamacook (approved):

PR #1222, opened 2024-02-28
Add endpoint to delete contracts from Tenderly

iamacook:
⚠️ Depends on #1221

## Summary

This adds a new `DELETE` `/v1/chains/:chainId/safes/:safeAddress/recovery/:moduleAddress` endpoint to delete contracts from Tenderly according to [their API](https://docs.tenderly.co/reference/api#tag/Contracts/operation/renameContract).

If we do not delete contract (here, recovery modules), users will continue to receive Alerts if an event is emitted. This is unlikely if the module is not activated on the Safe, but still plausible.

## Changes

- Add `TenderlyApid['deleteContract']`
- Add `AlertsRepository['deleteContract']`
- Add `DELETE` `/v1/chains/:chainId/safes/:safeAddress/recovery/:moduleAddress` and associated validation
- Update/extend associated tests


src/routes/recovery/recovery.controller.ts
```diff
@@ -26,4 +35,17 @@ export class RecoveryController {
       addRecoveryModuleDto,
     });
   }
+
+  @HttpCode(200)
+  @Delete()
+  async deleteRecoveryModule(
+    @Param('chainId') chainId: string,
+    @Body(DeleteRecoveryModuleDtoValidationPipe)
+    deleteRecoveryModuleDto: DeleteRecoveryModuleDto,
```
fmrsabino, line 44: Since we are deleting a resource which is identified by the module address, I would say it's more semantically clear that it is part of the path: `DELETE /v1/chains/:chainId/safes/:safeAddress/recovery/:moduleAddress`

src/routes/recovery/recovery.controller.ts
```diff
@@ -26,4 +35,17 @@ export class RecoveryController {
       addRecoveryModuleDto,
     });
   }
+
+  @HttpCode(200)
```
fmrsabino, line 39: If we don't return a body in the response, then `204 No Content` might be a better fit.

fmrsabino (changes requested):

src/routes/recovery/recovery.controller.ts
iamacook: Moved to the route in 82952c4763b1f9a0705743980b8febc963eafa67.

src/routes/recovery/recovery.controller.ts
iamacook: Changed in 82952c4763b1f9a0705743980b8febc963eafa67.

src/routes/recovery/recovery.controller.ts
```diff
@@ -36,16 +34,15 @@ export class RecoveryController {
     });
   }
 
-  @HttpCode(200)
-  @Delete()
+  @HttpCode(204)
+  @Delete('/:moduleAddress')
```
fmrsabino, line 38: OOC: I don't think you need the slash or do you? 🤔

fmrsabino (approved):
👏 🚀

src/routes/recovery/recovery.controller.ts
iamacook: It does not seem to have an influence when there or not. I'll merge this as is.

fmrsabino (approved):
🚀 👏

PR #1225, opened 2024-02-28
Extend ICacheService interface by storing incrementable numeric values

hectorgomezv:
## Summary
This adds a new functionality to `ICacheService` interface: setting keys on the underlying cache that hold `number` values and can be incremented sequentially.

## Changes
- Adds `ICacheService.incrementAndGet` function. It also adds the function code on each implementation class.

### DIFF
--- a/src/datasources/cache/__tests__/fake.cache.service.spec.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.spec.ts
@@ -59,7 +59,7 @@ describe('FakeCacheService', () => {
     expect(target.keyCount()).toBe(0);
   });
 
-  it('increments the value of a key', async () => {
+  it('creates a missing key and increments its value', async () => {
     const key = faker.string.alphanumeric();
     const firstResult = await target.incrementAndGet(key, undefined);
     expect(firstResult).toEqual(1);
@@ -70,5 +70,16 @@ describe('FakeCacheService', () => {
     }
 
     expect(results).toEqual([2, 3, 4, 5, 6]);
+  });
+
+  it('increments the value of an existing key', async () => {
+    const key = faker.string.alphanumeric();
+    const initialValue = faker.number.int({ min: 100 });
+    target.initNumericalValue(key, initialValue);
+
+    for (let i = 1; i <= 5; i++) {
+      const result = await target.incrementAndGet(key, undefined);
+      expect(result).toEqual(initialValue + i);
+    }
   });
 });
--- a/src/datasources/cache/__tests__/fake.cache.service.spec.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.spec.ts
@@ -75,7 +75,11 @@ describe('FakeCacheService', () => {
   it('increments the value of an existing key', async () => {
     const key = faker.string.alphanumeric();
     const initialValue = faker.number.int({ min: 100 });
-    target.initNumericalValue(key, initialValue);
+    await target.set(
+      new CacheDir(key, ''),
+      initialValue,
+      faker.number.int({ min: 1 }),
+    );
 
     for (let i = 1; i <= 5; i++) {
       const result = await target.increment(key, undefined);
--- a/src/datasources/cache/__tests__/fake.cache.service.spec.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.spec.ts
@@ -61,12 +61,12 @@ describe('FakeCacheService', () => {
 
   it('creates a missing key and increments its value', async () => {
     const key = faker.string.alphanumeric();
-    const firstResult = await target.incrementAndGet(key, undefined);
+    const firstResult = await target.increment(key, undefined);
     expect(firstResult).toEqual(1);
 
     const results: number[] = [];
     for (let i = 0; i < 5; i++) {
-      results.push(await target.incrementAndGet(key, undefined));
+      results.push(await target.increment(key, undefined));
     }
 
     expect(results).toEqual([2, 3, 4, 5, 6]);
@@ -78,7 +78,7 @@ describe('FakeCacheService', () => {
     target.initNumericalValue(key, initialValue);
 
     for (let i = 1; i <= 5; i++) {
-      const result = await target.incrementAndGet(key, undefined);
+      const result = await target.increment(key, undefined);
       expect(result).toEqual(initialValue + i);
     }
   });
--- a/src/datasources/cache/__tests__/fake.cache.service.spec.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.spec.ts
@@ -58,4 +58,17 @@ describe('FakeCacheService', () => {
 
     expect(target.keyCount()).toBe(0);
   });
+
+  it('increments the value of a key', async () => {
+    const key = faker.string.alphanumeric();
+    const firstResult = await target.incrementAndGet(key, undefined);
+    expect(firstResult).toEqual(1);
+
+    const results: number[] = [];
+    for (let i = 0; i < 5; i++) {
+      results.push(await target.incrementAndGet(key, undefined));
+    }
+
+    expect(results).toEqual([2, 3, 4, 5, 6]);
+  });
 });
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -41,7 +41,7 @@ export class FakeCacheService implements ICacheService {
 
   set(
     cacheDir: CacheDir,
-    value: string | number, // this changed
+    value: string | number,
     expireTimeSeconds: number | undefined,
   ): Promise<void> {
     if (!expireTimeSeconds || expireTimeSeconds <= 0) {
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -56,7 +56,7 @@ export class FakeCacheService implements ICacheService {
     return Promise.resolve();
   }
 
-  incrementAndGet(
+  increment(
     cacheKey: string,
     // eslint-disable-next-line @typescript-eslint/no-unused-vars
     expireTimeSeconds: number | undefined,
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -2,7 +2,7 @@ import { ICacheService } from '@/datasources/cache/cache.service.interface';
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 
 export class FakeCacheService implements ICacheService {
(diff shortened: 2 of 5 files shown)


src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -2,7 +2,7 @@ import { ICacheService } from '@/datasources/cache/cache.service.interface';
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 
 export class FakeCacheService implements ICacheService {
-  private cache: Record<string, Record<string, string>> = {};
+  private cache: Record<string, Record<string, string> | number> = {};
```
hectorgomezv, line 5: I had to extend this fake `cache` definition to hold `number` values behind `string` keys.

src/datasources/cache/cache.service.interface.ts
```diff
@@ -12,4 +12,9 @@ export interface ICacheService {
   get(cacheDir: CacheDir): Promise<string | undefined>;
 
   deleteByKey(key: string): Promise<number>;
+
+  incrementAndGet(
```
hectorgomezv, line 16: I think `incrementAndGet` is more descriptive for the clients of this interface API than simply `incr`, but of course suggestions on the naming are welcome.

src/datasources/cache/__tests__/fake.cache.service.spec.ts
```diff
@@ -58,4 +58,17 @@ describe('FakeCacheService', () => {
 
     expect(target.keyCount()).toBe(0);
   });
+
+  it('increments the value of a key', async () => {
+    const key = faker.string.alphanumeric();
+    const firstResult = await target.incrementAndGet(key, undefined);
```
iamacook, line 64: This initially confused me - it creates missing keys, then increments them. It matches the functionality of Redis though so I think it's fine. I would, however, adjust the description to reflect this and add a second test that has a pre-existing value in the cache, starting from a higher number.

src/datasources/cache/__tests__/fake.cache.service.ts
iamacook: I find it strange that Redis [returns a number after incrementing vs. "getting" a value](https://redis.io/commands/incr/). What do you think about converting the `INCR` return value to a string to keep the same type? We could then remove all casting as well.

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -144,4 +145,24 @@ describe('RedisCacheService', () => {
     // Connection is reopened after this test execution
     redisClient = await redisClientFactory();
   });
+
+  it('Increments and gets the value of a key', async () => {
```
iamacook, line 149: I would suggest the same as my previous comment regarding a second test.

hectorgomezv: Added in [5f7073a](https://github.com/safe-global/safe-client-gateway/pull/1225/commits/5f7073afaea1b5398057abebd6bdfbba2bc9141b)

src/datasources/cache/__tests__/fake.cache.service.ts
iamacook: > Actually, to remove the casting we would need to convert the number to `Record<string, string>`, right?

Yes, or do a type check before, e.g. `typeof === 'string'`.

> The Redis cache is accessed with `HGET/HSET`, which works with Redis _hashes_, and with `INCR`, which works with plain key-values. The `cache` in-memory variable in this test class tries to reflect this duality.

I see. It's odd that it can now return either a string or number. We will have to do the aforementioned checks within the codebase.

src/datasources/cache/redis.cache.service.ts
```diff
@@ -70,6 +70,17 @@ export class RedisCacheService
...
 
+  async incrementAndGet(
+    cacheKey: string,
+    expireTimeSeconds: number | undefined,
+  ): Promise<number> {
+    const result = await this.client.incr(cacheKey);
+    if (expireTimeSeconds !== undefined && expireTimeSeconds > 0) {
+      await this.client.expire(cacheKey, expireTimeSeconds, 'NX');
+    }
+    return result;
```
fmrsabino, line 81: We should perform these two steps on a transaction. In the event that `INCR` is successful and `EXPIRE` is not, we are setting a key which doesn't expire ever. Suggestion:

```typescript
const transaction = this.client.multi().incr(cacheKey);
if (expireTimeSeconds !== undefined && expireTimeSeconds > 0) {
  transaction.expire(cacheKey, expireTimeSeconds, 'NX');
}
const [incrRes] = await transaction.get(cacheKey).exec();
return Number(incrRes);
```

src/datasources/cache/cache.service.interface.ts
fmrsabino: I would rename this to just `increment`. The result of the increment can be documented instead. Else it seems that we are composing different actions which should be the responsibility of the caller.

Example: the caller might choose to ignore the contents of `increment` which makes `increment` self-explanatory. But the result would still be there if needed.

fmrsabino (changes requested):

src/datasources/cache/redis.cache.service.ts
hectorgomezv: I do agree this would be more strict if wrapped into a transaction, but at the same time, I think it would have an associated performance cost. 

> In the event that INCR is successful and EXPIRE is not, we are setting a key which doesn't expire ever

Regarding this case, the subsequent call to the `increment` function would execute the `expire` command, and the key would have a TTL then. But it is true that if there is no subsequent call the key would live forever in the cache.

So I've added your suggestion in [66f1152](https://github.com/safe-global/safe-client-gateway/pull/1225/commits/66f1152d921315ee265df6e48f67afcb63edead5)

Thanks!

src/datasources/cache/cache.service.interface.ts
hectorgomezv: Done in [66f1152](https://github.com/safe-global/safe-client-gateway/pull/1225/commits/66f1152d921315ee265df6e48f67afcb63edead5)

iamacook (approved):

src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -49,7 +51,25 @@ export class FakeCacheService implements ICacheService {
...
+      this.cache[cacheKey] = 1;
+    } else {
+      this.cache[cacheKey] = ++(this.cache[cacheKey] as number);
+    }
+    return Promise.resolve(this.cache[cacheKey] as number);
+  }
+
+  initNumericalValue(cacheKey: string, value: number): void {
+    this.cache[cacheKey] = value;
+  }
```
fmrsabino, line 74: I find this function a little bit out of place as this is mostly setting a value. By extending the Fake with specialised functions we make it harder for the caller to know how they should use it.

For example if you replace the real cache with a Fake one, the callers will only call `increment` because `initNumericalValue` is not part of interface being offered. 

Shouldn't this behaviour be covered already by the `increment` implementation?

hectorgomezv: I agree with you, but this is needed in order to test the case where there is a pre-existent value behind a key, and it gets incremented, as @iamacook pointed out in [this comment](https://github.com/safe-global/safe-client-gateway/pull/1225#discussion_r1507292230): 

> add a second test that has a pre-existing value in the cache, starting from a higher number.

With the current set of functions exposed by `FakeCacheService`, we don't have a way to set a numeric value behind a key. The interface only provides a way to `set` by passing a `CacheDir`, i.e.: setting a Redis hash, not a simple key-value. 

This additional function is not needed in the `RedisCacheService` as we have the actual Redis client available in the tests, so we can _bypass_ the service.

I can see that this approach is not super consistent, as it introduces differences between the implementations (even considering `FakeCacheService` is test-only). I'm open to alternatives to solve this, some that comes to my mind are:
- Expose the `cache` class variable from `FakeCacheService` to the outside world, to be able to modify its state from outside the class.
- Add a `set` function to both implementations (and remove this `initNumericValue` then) that maps Redis `SET` command. The already existent `set` functions are mapping `HSET` right now, so this would be a wider refactor.
- Do not test the case that @iamacook requested (see above).

Do you have ny preference/additional solution @fmrsabino ?

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -144,4 +145,37 @@ describe('RedisCacheService', () => {
...
+
+    for (let i = 1; i <= 5; i++) {
+      const result = await redisCacheService.increment(key, undefined);
+      expect(result).toEqual(initialValue + i);
+    }
+
+    const ttl = await redisClient.ttl(key);
+    expect(ttl).toBeGreaterThan(0);
+    expect(ttl).toBeLessThanOrEqual(expireTime);
+  });
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [166f8e1](https://github.com/safe-global/safe-client-gateway/pull/1225/commits/166f8e1555769c3671daaaee8c5e649ce259f227)

src/datasources/cache/__tests__/fake.cache.service.ts
hectorgomezv: OK, I've added your suggestion. I agree it's not perfect because based on the type of the value provided, the cache will be written differently, so, for example, writing a `string` in the cache's _first level_ is not possible, and also to write a `number` into the _second level_ (inside the _Redis hash_ this class mimics). But I agree it's sufficient and maybe a more consistent one.

I guess the ideal solution if we want to write both plain key-values and hashes (something that we didn't need before now), would be to map Redis `set` and `hset` 🤔

fmrsabino: Yes, I believe that for now this solution works but we need to come back to it to provide a more robust implementation of the fake. No need to tackle in this PR 👍

hectorgomezv: Changed in [797c9f2](https://github.com/safe-global/safe-client-gateway/pull/1225/commits/797c9f211d41722511d547963477ff597e208f8e)

Thanks!

src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -34,22 +34,42 @@ export class FakeCacheService implements ICacheService {
...
-    return Promise.resolve(this.cache[cacheDir.key][cacheDir.field]);
+    return Promise.resolve(
+      (this.cache[cacheDir.key] as Record<string, string>)[cacheDir.field],
+    );
   }
 
   set(
     cacheDir: CacheDir,
-    value: string,
+    value: string | number, // this changed
```
fmrsabino, line 44:
```suggestion
    value: string | number,
```

fmrsabino (approved):

PR #1226, opened 2024-02-29
Propagate Django error messages from Transaction Service

iamacook:
## Summary

This adds an error mapper to the `TransactionApi` service that maps Django error messages (`nonFieldErrors[0]`) if present to be the "standard" error `message`.

## Changes

- Add `TransactionApi['mapError']` and use it within all calls
- Add test coverage for `TransactionApi`

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.spec.ts
+++ b/src/datasources/transaction-api/transaction-api.service.spec.ts
@@ -116,7 +116,7 @@ describe('TransactionApi', () => {
 
     const errorMessage = faker.word.words();
     it.each([
-      ['Django', { nonFieldErrors: [errorMessage] }],
+      ['Transaction Service', { nonFieldErrors: [errorMessage] }],
       ['standard', new Error(errorMessage)],
     ])(`should forward a %s error`, async (_, error) => {
       const data = faker.string.hexadecimal();
@@ -169,7 +169,7 @@ describe('TransactionApi', () => {
 
     const errorMessage = faker.word.words();
     it.each([
-      ['Django', { nonFieldErrors: [errorMessage] }],
+      ['Transaction Service', { nonFieldErrors: [errorMessage] }],
       ['standard', new Error(errorMessage)],
     ])(`should forward a %s error`, async (_, error) => {
       const getBackboneUrl = `${baseUrl}/api/v1/about`;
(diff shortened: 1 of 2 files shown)


src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -26,20 +40,17 @@ const configurationService = {
 } as jest.MockedObjectDeep<IConfigurationService>;
 const mockConfigurationService = jest.mocked(configurationService);
 
-const httpErrorFactory = {
-  from: jest.fn(),
-} as jest.MockedObjectDeep<HttpErrorFactory>;
-const mockHttpErrorFactory = jest.mocked(httpErrorFactory);
```
iamacook, line 32: I migrated to using the real HttpErrorFactory as we were mocking and testing the mocked functionality which didn't make much sense. Would welcome your thoughts regarding this.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -860,4 +862,17 @@ export class TransactionApi implements ITransactionApi {
...
   }
+
+  private mapError(error: unknown): unknown {
+    if (error instanceof NetworkResponseError) {
+      // Map Django error as "standard" error message if present
+      const djangoError = get(error.data, 'nonFieldErrors[0]');
+      if (djangoError) {
+        return new NetworkResponseError(error.url, error.response, {
+          message: djangoError,
+        });
```
hectorgomezv, line 873: Nit: should we abstract from the technology used by the data source? I mean, what about changing the `django` references here and in the tests for another generic concept like `transactionService`?

hectorgomezv: And also other nits: 
- `nonFieldErrors[0]` might be extracted to a class constant, wdyt? 
- This assumes we are only interested in the first error in the payload, aren't we? I'm ok with this as long as this is ok for the clients.

hectorgomezv (approved):
I've only left small nits, but it LGTM 👏🏻

Glad to see a lot of tests were added 💪🏻

src/datasources/transaction-api/transaction-api.service.ts
iamacook: I generalised all references to "Django" (preferring "Transaction Service") in 931505d2b0fdb5eae083f4fd891acea1b14fa97b, as well as moving the path to a constant. I also added a comment explaining the reasoning for returning the first error element - a string is preferred.

PR #1228, opened 2024-02-29
Add Zerion API rate limit

hectorgomezv:
## Summary
This PR adds a cache-based rate limit to the `ZerionBalancesApi` outgoing requests.

## Changes
- Adds `ZERION_RATE_LIMIT_PERIOD_SECONDS` environment variable, holding the number of seconds to clear the calls counter on each _cycle_.
- Adds `ZERION_RATE_LIMIT_CALLS_BY_PERIOD` environment variable, holding the number of calls allowed on each _cycle_.
- Adds a `ZerionBalancesApi._checkRateLimit`, which checks the ability to make a call to Zerion before doing it.

### DIFF
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -96,8 +96,8 @@ export default (): ReturnType<typeof configuration> => ({
         },
       },
       zerion: {
-        baseUri: faker.internet.url({ appendSlash: false }),
         apiKey: faker.string.hexadecimal({ length: 32 }),
+        baseUri: faker.internet.url({ appendSlash: false }),
         chains: {
           1: { chainName: faker.string.sample() },
           10: { chainName: faker.string.sample() },
@@ -123,6 +123,8 @@ export default (): ReturnType<typeof configuration> => ({
             'usd',
           ]),
         ),
+        limitPeriodSeconds: faker.number.int({ min: 1, max: 10 }),
+        limitCalls: faker.number.int({ min: 1, max: 10 }),
       },
     },
   },
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -54,8 +54,8 @@ export default () => ({
         },
       },
       zerion: {
-        baseUri: process.env.ZERION_BASE_URI || 'https://api.zerion.io',
         apiKey: process.env.ZERION_API_KEY,
+        baseUri: process.env.ZERION_BASE_URI || 'https://api.zerion.io',
         chains: {
           1: { chainName: 'ethereum' },
           10: { chainName: 'optimism' },
@@ -91,6 +91,12 @@ export default () => ({
           'try',
           'zar',
         ],
+        limitPeriodSeconds: parseInt(
+          process.env.ZERION_RATE_LIMIT_PERIOD_SECONDS ?? `${10}`,
+        ),
+        limitCalls: parseInt(
+          process.env.ZERION_RATE_LIMIT_CALLS_BY_PERIOD ?? `${2}`,
+        ),
       },
     },
   },
--- a/src/datasources/balances-api/zerion-balances-api.service.ts
+++ b/src/datasources/balances-api/zerion-balances-api.service.ts
@@ -9,6 +9,7 @@ import {
   ZerionCollectible,
   ZerionCollectibles,
 } from '@/datasources/balances-api/entities/zerion-collectible.entity';
+import { LimitReachedError } from '@/datasources/network/entities/errors/limit-reached.error';
 import { CacheFirstDataSource } from '@/datasources/cache/cache.first.data.source';
 import { CacheRouter } from '@/datasources/cache/cache.router';
 import {
@@ -24,7 +25,9 @@ import {
 import { Collectible } from '@/domain/collectibles/entities/collectible.entity';
 import { getNumberString } from '@/domain/common/utils/utils';
 import { Page } from '@/domain/entities/page.entity';
+import { DataSourceError } from '@/domain/errors/data-source.error';
 import { IBalancesApi } from '@/domain/interfaces/balances-api.interface';
+import { LoggingService, ILoggingService } from '@/logging/logging.interface';
 import { Inject, Injectable } from '@nestjs/common';
 
 export const IZerionBalancesApi = Symbol('IZerionBalancesApi');
@@ -39,8 +42,14 @@ export class ZerionBalancesApi implements IBalancesApi {
   private readonly defaultNotFoundExpirationTimeSeconds: number;
   private readonly fiatCodes: string[];
 
+  // Number of seconds for each rate-limit cycle
+  private readonly limitPeriodSeconds: number;
+  // Number of allowed calls on each rate-limit cycle
+  private readonly limitCalls: number;
+
   constructor(
     @Inject(CacheService) private readonly cacheService: ICacheService,
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
     @Inject(IConfigurationService)
     private readonly configurationService: IConfigurationService,
     private readonly dataSource: CacheFirstDataSource,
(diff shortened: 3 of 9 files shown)


src/config/entities/__tests__/configuration.ts
```diff
@@ -96,8 +96,8 @@ export default (): ReturnType<typeof configuration> => ({
         },
       },
       zerion: {
-        baseUri: faker.internet.url({ appendSlash: false }),
         apiKey: faker.string.hexadecimal({ length: 32 }),
+        baseUri: faker.internet.url({ appendSlash: false }),
```
hectorgomezv, line 100: This is only to sort these keys alphabetically.

src/config/entities/configuration.ts
```diff
@@ -54,8 +54,8 @@ export default () => ({
         },
       },
       zerion: {
-        baseUri: process.env.ZERION_BASE_URI || 'https://api.zerion.io',
         apiKey: process.env.ZERION_API_KEY,
+        baseUri: process.env.ZERION_BASE_URI || 'https://api.zerion.io',
```
hectorgomezv, line 58: Sorting keys alphabetically.

src/datasources/cache/cache.first.data.source.ts
```diff
@@ -40,15 +41,22 @@ export class CacheFirstDataSource {
...
    * @param args.notFoundExpireTimeSeconds - the time to live in seconds for the error when the item is not found
+   * @param args.isRateLimited - indicates if the call is rate-limited. If so, and the data is not already cached, an error will be thrown
    */
   async get<T>(args: {
     cacheDir: CacheDir;
     url: string;
     notFoundExpireTimeSeconds: number;
     networkRequest?: NetworkRequest;
     expireTimeSeconds?: number;
+    isRateLimited?: boolean;
```
hectorgomezv, line 52: This argument is added to inform the function the data should be retrieved exclusively from the cache. 

Another option, besides adding this parameter, would be to inject both `ICacheService` and `INetworkService` into the caller (`ZerionBalancesApi`) and perform the logic there, but I thought having this parameter could be valuable for other classes in the future. 

Having that said, I haven't a strong opinion about the alternatives. Feedback is welcomed :)

src/datasources/cache/cache.router.ts
```diff
@@ -31,6 +31,7 @@ export class CacheRouter {
   private static readonly TRANSFERS_KEY = 'transfers';
   private static readonly ZERION_BALANCES_KEY = 'zerion_balances';
   private static readonly ZERION_COLLECTIBLES_KEY = 'zerion_collectibles';
+  private static readonly ZERION_RATE_LIMIT_KEY = 'zerion_rate_limit';
```
iamacook, line 34: I can forsee us using this for relaying. Maybe we should rename the key to `rate_limit` and then the method accept a prefix, e.g. `"zerion"` in this case. What do you think?

src/routes/balances/__tests__/controllers/zerion-balances.controller.spec.ts
```diff
@@ -425,5 +442,153 @@ describe('Balances Controller (Unit)', () => {
...
+            .expect(200);
+        }
+
+        await request(app.getHttpServer())
+          .get(
+            `/v1/chains/${chain.chainId}/safes/${safeAddress}/balances/${crypto.randomUUID()}`,
+          )
+          .expect(429);
+
+        expect(networkService.get.mock.calls.length).toBe(limitCalls + 1);
```
iamacook, line 590: Nit: can we add a comment here regarding the `+ 1` relating to the Config Service call? It initially confused me.

src/datasources/cache/cache.first.data.source.ts
hectorgomezv:
```
We need not inject anything. We could throw if the rate limit was hit
```

That was something I also thought at first glance, but it wasn't so simple. We need to avoid throwing if the data is available in the cache. Otherwise, we would be blocking all the `getBalances` calls, not only the ones that need a call to Zerion.

src/datasources/cache/cache.router.ts
hectorgomezv: Hmm, that would imply we are using the same rate limit parameters for all the APIs we want to rate limit, wouldn't it? I think it is valuable to decide the throughput for each case.

src/routes/balances/__tests__/controllers/zerion-balances.controller.spec.ts
hectorgomezv: Absolutely! Good point.

Added in [395394b](https://github.com/safe-global/safe-client-gateway/pull/1228/commits/395394bf4b4f8fbc44293488ddf74f17c798ce36)

src/datasources/cache/cache.router.ts
hectorgomezv: Yeah, good suggestion, added in [7f4a85b](https://github.com/safe-global/safe-client-gateway/pull/1228/commits/7f4a85b6d803176df8a0d3894921bdc636508988)

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -24,23 +25,32 @@ import {
...
 import { IBalancesApi } from '@/domain/interfaces/balances-api.interface';
+import { LoggingService, ILoggingService } from '@/logging/logging.interface';
 import { Inject, Injectable } from '@nestjs/common';
 
 export const IZerionBalancesApi = Symbol('IZerionBalancesApi');
 
 @Injectable()
 export class ZerionBalancesApi implements IBalancesApi {
   private static readonly collectiblesSorting = '-floor_price';
+  private static readonly RATE_LIMIT_CACHE_KEY_PREFIX = 'zerion';
```
iamacook, line 38: Nit: I'd either match the casing of `collectiblesSorting` or change that to screaming snake case. I'm in favour of the latter.

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -24,23 +25,32 @@ import {
...
   private readonly apiKey: string | undefined;
   private readonly baseUri: string;
   private readonly chainsConfiguration: Record<number, ChainAttributes>;
   private readonly defaultExpirationTimeInSeconds: number;
   private readonly defaultNotFoundExpirationTimeSeconds: number;
   private readonly fiatCodes: string[];
```
iamacook, line 45: Nit: unnecessary newline.

src/datasources/cache/cache.first.data.source.ts
iamacook: I see. I would still lean towards moving this out of the data source. If that's difficult we can keep it as is though, but I'd tweak the name, e.g. `shouldImmediatelyReturn`.

src/datasources/cache/redis.cache.service.spec.ts
```diff
@@ -173,4 +173,20 @@ describe('RedisCacheService', () => {
     expect(ttl).toBeGreaterThan(0);
     expect(ttl).toBeLessThanOrEqual(expireTime);
   });
+
+  it('increments the value of an existing key', async () => {
```
iamacook, line 177: This is a duplicate of the test above.

hectorgomezv: Oops, it got duplicated in a rebase. Thanks!

Removed in [23d9d5e](https://github.com/safe-global/safe-client-gateway/pull/1228/commits/23d9d5e6730da71544fe9b883c2cc35d8fe55fa1)

src/datasources/balances-api/zerion-balances-api.service.ts
hectorgomezv: Removed in [23d9d5e](https://github.com/safe-global/safe-client-gateway/pull/1228/commits/23d9d5e6730da71544fe9b883c2cc35d8fe55fa1)

src/datasources/cache/cache.first.data.source.ts
hectorgomezv: I've removed the parameter as you suggested, and now all the logic is performed within `ZerionBalancesApi`. If more data source APIs need to be rate-limited we can think in including back the parameter :) 

[23d9d5e](https://github.com/safe-global/safe-client-gateway/pull/1228/commits/23d9d5e6730da71544fe9b883c2cc35d8fe55fa1)

iamacook (approved):

PR #1229, opened 2024-02-29
Scaffold locking repository

iamacook:
## Summary

This adds a scaffold for the locking repository. It builds upon the proposed implementation of #1227.

## Changes

- Add validation for locking events
- Add validation for ranks
- Create `ILockingRepository`:
  - `getRank`
  - `getLeaderboard`
  - `getLockingHistory`


src/datasources/locking-api/locking-api.module.ts
```diff
@@ -0,0 +1,10 @@
+import { Global, Module } from '@nestjs/common';
+import { ILockingApi } from '@/domain/interfaces/locking-api.interface';
+import { LockingApi } from '@/datasources/locking-api/locking-api.service';
+
+@Global()
```
fmrsabino, line 5: The module doesn't need to be a global one.

src/domain.module.ts
```diff
@@ -52,13 +52,19 @@ import { IHumanDescriptionRepository } from '@/domain/human-description/human-de
...
+import { LockingEventValidator } from '@/domain/locking/locking-event.validator';
+import { LockingApiModule } from '@/datasources/locking-api/locking-api.module';
 
 @Global()
 @Module({
   imports: [
     BalancesApiModule,
     ConfigApiModule,
     HumanDescriptionApiModule,
+    LockingApiModule,
```
fmrsabino, line 67: I would leave this module out of the `DomainModule` for now: the most recent approach to have smaller modules is preferable from a maintenance perspective (see `AccountDataSourceModule`).

This is also to say that we should break down this module into smaller ones in the future.

src/domain/interfaces/validator.interface.ts
```diff
@@ -1,3 +1,9 @@
+import { Page } from '@/domain/entities/page.entity';
+
 export interface IValidator<T> {
   validate(data: unknown): T;
 }
+
+export interface IPageValidator<T> {
+  validatePage(data: unknown): Page<T>;
+}
```
fmrsabino, line 9: We already have this interface; see `src/domain/interfaces/page-validator.interface.ts`

src/domain/locking/entities/schemas/locking-event.schema.ts
```diff
@@ -0,0 +1,89 @@
...
+  $id: LOCKING_EVENT_SCHEMA_ID,
+  type: 'object',
+  discriminator: { propertyName: 'type' },
+  oneOf: [lockEventSchema, unlockEventSchema, withdrawEventSchema],
+};
+
+export const LOCKING_EVENT_PAGE_SCHEMA_ID =
+  'https://safe-client.safe.global/schemas/locking/locking-event-page.json';
+
+// TODO: Use buildPageSchema when types are fixed
```
fmrsabino, line 69: One of the main differences that I see between `buildPageSchema` and `lockingEventPageSchema` is that `count` is nullable in `lockingEventPageSchema`. But besides that, `buildPageSchema` can be used or?

fmrsabino (changes requested):

src/datasources/locking-api/locking-api.module.ts
iamacook: Removed in 53082b79be2c075827de2f2c059f984030fbe1b8.

src/domain.module.ts
iamacook: Removed in 53082b79be2c075827de2f2c059f984030fbe1b8.

src/domain/interfaces/validator.interface.ts
iamacook: Adjusted in 53082b79be2c075827de2f2c059f984030fbe1b8.

src/domain/locking/entities/schemas/locking-event.schema.ts
iamacook: `count` [is nullable everywhere](https://github.com/safe-global/safe-client-gateway/pull/1192), afaiu. Whilst we can use the `buildPageSchema` function, we'd need to coerce the type to `Schema`, which is inherently wrong.

I did manage to get the types working but it requires changing a lot of schemas across the project. I would sooner leave this comment and create a function with `zod` when we've migrated. Would you prefer I do so in this PR though?

src/domain.module.ts
iamacook: A domain-specific module was added in the follow-up PR in 43c1d35a03d799bffdeb4b50eaf2d0342e9b006c, connecting everything.

src/domain.module.ts
```diff
@@ -89,9 +91,11 @@ import { BalancesApiModule } from '@/datasources/balances-api/balances-api.modul
     DelegateValidator,
     EstimationsValidator,
     SingletonValidator,
+    LockingEventValidator,
     MessageValidator,
     ModuleTransactionValidator,
     MultisigTransactionValidator,
+    RankValidator,
```
fmrsabino, line 98: Related with https://github.com/safe-global/safe-client-gateway/pull/1229#discussion_r1510912118: I would also include the validators in the domain-specific module (i.e.: import them only if needed by the module).

src/domain/locking/entities/schemas/locking-event.schema.ts
fmrsabino: > Would you prefer I do so in this PR though?

No let's tackle these issues in the Zod migration. This was more so that the schemas would be aligned across the project (as we already have one schema for the `Page`).

fmrsabino (changes requested):

fmrsabino (approved):
🚀

PR #1217, opened 2024-02-27
Add `proposer` to `MultisigExecutionInfo`

iamacook:
## Summary

This adds the `MultisigTransaction['proposer']` from the Transaction Service as `MultisigExecutionInfo['proposer']` (`AddressInfo`) for the client to reference for deletion. (Only the proposer can delete a transaction.)

## Changes

- Add `proposer` to `MultisigTransaction` entity (`null` pre-deletion integration)
- Add validation for `proposer` to `multisigTransactionSchema`
- Forward `proposer` in `MultisigTransactionExecutionInfoMapper['mapExecutionInfo']`
- Add `proposer` to `multisigTransactionBuilder`
- Add/update associated tests

PR #1234, opened 2024-03-01
Move `proposer` from `executionInfo` to `detailedExecutionInfo`

iamacook:
## Summary

This moves the [recently added `proposer`
](https://github.com/safe-global/safe-client-gateway/pull/1217), from the `executionInfo` to the `detailedExecutionInfo` of a multisig transaction

## Motivation

1. The `proposer` is not required in the transactions list (where only `executionInfo` is returned).
2. Semantically, `proposer` matches the other properties of `detailedExecutionInfo`.

## Changes

- `proposer` removed from `executionInfo`
- `proposer` added to `detailedExecutionInfo`
- Associated tests updated

### DIFF
--- a/src/routes/transactions/mappers/multisig-transactions/multisig-transaction-execution-details.mapper.ts
+++ b/src/routes/transactions/mappers/multisig-transactions/multisig-transaction-execution-details.mapper.ts
@@ -42,6 +42,9 @@ export class MultisigTransactionExecutionDetailsMapper {
               confirmation.submissionDate.getTime(),
             ),
         );
+    const proposer = transaction.proposer
+      ? new AddressInfo(transaction.proposer)
+      : null;
 
     const [gasTokenInfo, executor, refundReceiver, rejectors] =
       await Promise.all([
@@ -77,6 +80,7 @@ export class MultisigTransactionExecutionDetailsMapper {
       rejectors,
       gasTokenInfo,
       transaction.trusted,
+      proposer,
     );
   }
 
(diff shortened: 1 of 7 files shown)


src/routes/transactions/mappers/multisig-transactions/multisig-transaction-execution-details.mapper.ts
```diff
@@ -42,6 +42,9 @@ export class MultisigTransactionExecutionDetailsMapper {
               confirmation.submissionDate.getTime(),
             ),
         );
+    const proposer = transaction.proposer
+      ? new AddressInfo(transaction.proposer)
+      : null;
```
fmrsabino, line 47: Do we need to return the full Address Info if we have it? (implies calling the `AddressInfoHelper` instead).

iamacook: Adding that would result in an another request. As the client is (currently) only using this as a flag, I don't think it's needed. We can always add it later if necessary. What do you think?

fmrsabino: > Adding that would result in an another request. As the client is (currently) only using this as a flag, I don't think it's needed. We can always add it later if necessary. What do you think?

Sounds good 👍 We can it later if the requirements change.

fmrsabino (approved):
🚀

PR #1235, opened 2024-03-01
Add support for Database certificate pinning

fmrsabino:
- Adds support for pinning a CA certificate.
- If SSL is enabled:
  * `POSTGRES_SSL_REQUEST_CERT` can be used to make Postgres request a certificate from the clients. (default `true`)
  * `POSTGRES_SSL_CA_PATH` can be used to specify the absolute path to the certificate file (no default).
  * More details can be found in https://nodejs.org/api/tls.html#tlscreateserveroptions-secureconnectionlistener
- The self-signed certificated is a X509 certificate with validity of 365 days. The Common Name (CN) is `localhost`.

### Docker

- The local development instance `db`, does not have SSL configured (i.e.: no changes).
- The local test instance `db-test`, has SSL configured and will use the self-signed certificate available under `<projectRoot>/db_config/test`.
  * With the described setup, we are testing if the connection options are valid.


src/datasources/account/account.datasource.spec.ts
```diff
@@ -13,19 +13,37 @@ import {
 import { accountBuilder } from '@/domain/account/entities/__tests__/account.builder';
 import { verificationCodeBuilder } from '@/domain/account/entities/__tests__/verification-code.builder';
 import { getAddress } from 'viem';
+import * as fs from 'fs';
+import * as path from 'path';
```
hectorgomezv, line 17: Nitpick: it might be better to select the imported functions instead of the whole modules:
```javascript
import { readFileSync } from 'fs';
import { join} from 'path';
```

src/datasources/account/account.datasource.spec.ts
```diff
@@ -13,19 +13,37 @@ import {
...
+import * as process from 'process';
```
hectorgomezv, line 18: I think this is not needed, `process` is a global object: https://nodejs.org/docs/latest-v20.x/api/globals.html#process

docker-compose.yml
```diff
@@ -24,6 +24,15 @@ services:
       POSTGRES_PASSWORD: ${POSTGRES_TEST_PASSWORD:-postgres}
     ports:
       - '${POSTGRES_TEST_PORT:-5433}:5432'
+    volumes:
+      - ./db_config/test/server.crt:/var/lib/postgresql/server.crt:ro
+      - ./db_config/test/server.key:/var/lib/postgresql/server.key:ro
```
hectorgomezv, line 29: I think these permission descriptors (`:ro`) are not enough.
```
FATAL:  private key file "/var/lib/postgresql/server.key" has group or world access
```

I had to manually set the needed chmod permissions (`0640`) in my machine to make this work. It seems it is consistent with the way the volume permissions work: https://stackoverflow.com/a/50325808

If we can't find a better workaround for this, I think we would need to document this required step in the project `README.md` file.

hectorgomezv (changes requested):

docker-compose.yml
fmrsabino: Good catch. I think it should be on whoever is running the repo to manage the correct permissions of the files (while we document on how to do so). R/W permissions for the only only are enough according to https://www.postgresql.org/docs/current/ssl-tcp.html#SSL-SETUP. Documented the steps in. [765de50](https://github.com/safe-global/safe-client-gateway/pull/1235/commits/765de5047d9dced2c9b007dc2c015a71ba9a4423)

src/datasources/account/account.datasource.spec.ts
fmrsabino: Fixed in [765de50](https://github.com/safe-global/safe-client-gateway/pull/1235/commits/765de5047d9dced2c9b007dc2c015a71ba9a4423)

src/datasources/account/account.datasource.spec.ts
fmrsabino: Fixed in [765de50](https://github.com/safe-global/safe-client-gateway/pull/1235/commits/765de5047d9dced2c9b007dc2c015a71ba9a4423)

hectorgomezv (approved):

PR #1233, opened 2024-03-01
Add CowSwapApi

fmrsabino:
Closes #1197 

- Adds an initial version of the CowSwapApi API.
- A CowSwapApi instance is tied to a specific chain (and base url). Support was added for Mainnet, Gnosis Chain and Sepolia.
- Adds support for getting an order via `GET /api/v1/orders/<uid>`.

Issue #1198, opened 2024-02-23
Enrich CoW Swap transactions (orders) with the respective order information


fmrsabino:
### Description

The transactions that are pending execution (queued transactions) or the transactions already executed (transaction history) should show the respective order information.

### Requirements

Order details should include:
- Buy / Sell amount and respective token information
- Price ratio
- Order validity
- Order status
- CoW Swap Explorer URL
- Filled status

PR #1239, opened 2024-03-04
Add SetPreSignatureDecoder

fmrsabino:
Adds the `SetPreSignatureDecoder`, a decoder helper which can be used to retrieve a CoW Swap Order UID from any `setPreSignature` transaction.

This represent preliminary work for https://github.com/safe-global/safe-client-gateway/issues/1198. With the ability to retrieve the Order UID (this PR) and the integration of the CoW Swap API done in https://github.com/safe-global/safe-client-gateway/pull/1233, the CGW components can now retrieve the data of CoW Swap orders and map the respective data to be rendered to the clients.

### DIFF
--- a/src/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper.ts
+++ b/src/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper.ts
@@ -1,6 +1,7 @@
-import { Injectable } from '@nestjs/common';
+import { Inject, Injectable } from '@nestjs/common';
 import { AbiDecoder } from '@/domain/contracts/decoders/abi-decoder.helper';
-import { getFunctionSelector, isHex, parseAbi } from 'viem';
+import { getFunctionSelector, parseAbi } from 'viem';
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
 
 export const abi = parseAbi([
   'function setPreSignature(bytes calldata orderUid, bool signed)',
@@ -10,7 +11,9 @@ export const abi = parseAbi([
 export class SetPreSignatureDecoder extends AbiDecoder<typeof abi> {
   private readonly setPreSignatureFunctionSelector: `0x${string}`;
 
-  constructor() {
+  constructor(
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
+  ) {
     super(abi);
     this.setPreSignatureFunctionSelector = getFunctionSelector(abi[0]);
   }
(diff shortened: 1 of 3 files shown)


src/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper.ts
```diff
@@ -0,0 +1,32 @@
...
+  /**
+   * Gets the Order UID associated with the provided transaction data.
+   *
+   * @param data - the transaction data for the setPreSignature call
+   * @returns {`0x${string}`} the order UID or null if the data does not represent a setPreSignature transaction
+   */
+  getOrderUid(data: `0x${string}`): `0x${string}` | null {
+    if (!data.startsWith(this.setPreSignatureFunctionSelector)) return null;
+    const { functionName, args } = this.decodeFunctionData({ data });
+    if (functionName !== 'setPreSignature') return null;
```
iamacook, line 27: Line 25 and 27 check the same thing here. If the former is removed, `decodeFunctionData` can throw though.

src/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (!args || !args[0] || typeof args[0] !== 'string' || !isHex(args[0]))
```
iamacook, line 28: These checks are theoretically not needed on a type level.

Are you checking for "fake" signatures? If so, you don't need to check `args[0]` or the `typeof` as the arguments are always an array (if present) and `isHex` accepts an `unknown` type.

src/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper.ts
fmrsabino: Good call about the `throw` (honestly I think this should be documented). I will wrap it in a `try/catch` block and return `null`. The goal of this function is not to provide the caller the opportunity to deal with the error.

src/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper.ts
fmrsabino: Right! At some point I had a state where the type checks were not working but `decodeFunctionData` does return the correct types. Fixed in [9dca8df](https://github.com/safe-global/safe-client-gateway/pull/1239/commits/9dca8df8a3cde5031ea9fd75fc93e15fb4a30639)

src/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper.ts
fmrsabino: Fixed in [9dca8df](https://github.com/safe-global/safe-client-gateway/pull/1239/commits/9dca8df8a3cde5031ea9fd75fc93e15fb4a30639)

iamacook (approved):

PR #1247, opened 2024-03-05
Guard adding/removing recovery contracts

iamacook:
## Summary

This adds guards to the following routes for addding/removing recovery contracts from Tenderly:

### `POST` `/v1/chains/:chainId/safes/:safeAddress/recovery`

A signature is required in order to successfully add a recovery contract. The following message need be signed (where ${} represents placeholder values for the respective data):

`enable-recovery-alerts-${chainId}-${safeAddress}-${moduleAddress}-${signer}-${timestamp}`

 To use this guard, the route should have:

- the `chainId` as part of the path parameters
- the `safeAddress` as part of the path parameters
- the `moduleAddress` as part of the JSON body (top level - unchanged)
- the `signer` as part of the JSON body (top level - new)
- the `Safe-Wallet-Signature` header set to the signature
- the `Safe-Wallet-Signature-Timestamp` header set to the signature timestamp

On top of this, the 'Safe-Wallet-Signature-Timestamp' need be within the past 5 minutes and the 'signer' an owner of the Safe.

### `DELETE` `chains/:chainId/safes/:safeAddress/recovery/:moduleAddress`

A signature is required in order to successfully remove a recovery contract. The following message need be signed (where ${} represents placeholder values for the respective data):

`enable-recovery-alerts-${chainId}-${safeAddress}-${moduleAddress}-${signer}-${timestamp}`

 To use this guard, the route should have:

- the `chainId` as part of the path parameters
- the `safeAddress` as part of the path parameters
- the `moduleAddress` as part of the JSON body (top level - unchanged)
- the `signer` as part of the JSON body (top level - new)
- the `Safe-Wallet-Signature` header set to the signature
- the `Safe-Wallet-Signature-Timestamp` header set to the signature timestamp

On top of this, the 'Safe-Wallet-Signature-Timestamp' need be within the past 5 minutes and the 'signer' an owner of the Safe.

## Changes

- Create `EnableRecoveryAlertsGuard`/`DisableRecoveryAlertsGuard`
- Add guards to respective routes


src/routes/recovery/guards/disable-recovery-alerts.guard.ts
```diff
@@ -0,0 +1,71 @@
...
+  ExecutionContext,
+  Inject,
+  Injectable,
+} from '@nestjs/common';
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
+import { verifyMessage } from 'viem';
+
+/**
+ * The DisableRecoveryAlertsGuard guard should be used on routes that require
+ * authenticated actions for deleting recovery contracts.
```
fmrsabino, line 12: We are not deleting the recovery contracts with this signature. We are disabling the alerts that can be triggered by a recovery contract.

src/routes/recovery/recovery.controller.spec.ts
```diff
@@ -82,27 +102,137 @@ describe('Recovery (Unit)', () => {
...
     });
 
-    it('should get a validation error', async () => {
+    it('should prevent requests older than 5 minutes', async () => {
+      const addRecoveryModuleDto = addRecoveryModuleDtoBuilder().build();
+      const privateKey = generatePrivateKey();
+      const signer = privateKeyToAccount(privateKey);
+      const chain = chainBuilder().build();
+      const safe = safeBuilder().with('owners', [signer.address]).build();
+      const timestamp = jest.now() - (5 * 60 * 1000 + 1); // 5m + 1ms
```
fmrsabino, line 122: Not really something that needs to be changed but other tests advance the timers instead of signing an older timestamp. https://github.com/safe-global/safe-client-gateway/blob/97c3188a1d30a313227b9d5c664ff92bdc61b59d/src/routes/email/email.controller.save-email.spec.ts#L184

Both should be valid.

fmrsabino (changes requested):

src/routes/recovery/guards/disable-recovery-alerts.guard.ts
iamacook: Technically we are deleting the contracts from Tenderly. It should match its name though so I adjusted the comment in cab30bf4823a04b278ad156652bfe0b319df3044.

src/routes/recovery/recovery.controller.spec.ts
iamacook: Changed in cab30bf4823a04b278ad156652bfe0b319df3044.

src/routes/recovery/guards/disable-recovery-alerts.guard.ts
fmrsabino: Yes it's more like deleting from a registry that is used for Alerts. I think "that require authenticated actions for deleting recovery contracts" can be misleading in the context of smart contract.

fmrsabino (approved):

PR #1248, opened 2024-03-05
Align and validate `ILockingApi` return types

iamacook:
## Summary

This updates the locking service types from data source to controller in accordance with the forthcoming service. Associated validation was added, as well as tests, accordingly.

## Changes

- `LockingEventSchema` and `RankSchema` created and types inferred from
- `LockingApi` updated to reference forthcoming locking history endpoint
- Associated tests/builder added accordingly


src/domain/locking/entities/schemas/rank.schema.ts
```diff
@@ -1,25 +1,7 @@
...
-};
-
-export const RANK_PAGE_SCHEMA_ID =
-  'https://safe-client.safe.global/schemas/locking/rank-page.json';
-
-export const rankPageSchema: Schema = buildPageSchema(
-  RANK_PAGE_SCHEMA_ID,
-  rankSchema,
-);
+export const RankPageSchema = buildZodPageSchema(RankSchema);
```
fmrsabino, line 7: I wonder if we should declare the different `TypePageSchema` as it is basically just setting an actual type for a `Page`. We can see how this evolves but one alternative would be that we would not have a const/type to represent each paginated response 🤔

fmrsabino (approved):
🚀

src/domain/interfaces/locking-api.interface.ts
```diff
@@ -12,7 +12,7 @@ export interface ILockingApi {
   }): Promise<Page<Rank>>;
 
   getLockingHistory(args: {
-    safeAddress?: string;
+    safeAddress: string;
```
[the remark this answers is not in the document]
iamacook: I made the change in 22cc90703bd0b748cc97308644b1ff11d36890a6 but in order for this to be "true" I also added validation for the param as it comes from the URL. This is the first time validating a param in the project (to my knowledge) so please lmk what you think.

src/domain/locking/entities/schemas/rank.schema.ts
iamacook: I am impartial. Whilst migrating other schemas, I noticed that there are often no "page schemas" and `results` are mapped and validated directly.

I am in favour of seeing how this develops and aligning them all once we've fully migrated to `zod`. What do you think?

src/domain/interfaces/locking-api.interface.ts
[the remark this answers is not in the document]
fmrsabino: I would say this is ok. If the address does not start with `0x` we should exit early. If the address needs to be checksummed or not is a different concern though.

src/domain/locking/entities/schemas/rank.schema.ts
fmrsabino: > I am in favour of seeing how this develops and aligning them all once we've fully migrated to `zod`. What do you think?

Let's see how it develops then 👍

fmrsabino (approved):

PR #1260, opened 2024-03-06
Migrate `Collectible` to `zod`

iamacook:
## Summary

This migrates the validation of `Collectible` to `zod`.

## Changes

- Removes `CollectiblesValidator` and associated schema
- Adds `CollectibleSchema` and infers type(s) from it
- Propagate required changes in codebase
- Update associated tests


src/domain/collectibles/entities/schemas/collectible.schema.ts
```diff
@@ -1,40 +1,17 @@
...
-  },
-  required: ['address', 'tokenName', 'tokenSymbol', 'logoUri', 'id'],
-};
+export const CollectibleSchema = z.object({
+  address: AddressSchema,
+  tokenName: z.string(),
+  tokenSymbol: z.string(),
+  logoUri: z.string(),
+  id: z.string(),
+  // AJV remnant - do not use format: 'uri' as it fails on some payloads that should be considered valid
```
hectorgomezv, line 10: Nit: do you think the AJV reference is needed? We could just comment out that we don't enforce URL validation here, wdyt?

hectorgomezv (approved):

src/domain/collectibles/entities/schemas/collectible.schema.ts
iamacook: I changed the comment in 01237f555881832ebe717e380095241ffb44c2a1 but I think we should revisit this in the future. If we can make the schemas stricter, it would be beneficial.

hectorgomezv (approved):

PR #1261, opened 2024-03-06
Migrate `Contract` to `zod`

iamacook:
## Summary

This migrates the validation of `Contract` to `zod`.

## Changes

- Removes `ContractsValidator` and associated schema
- Adds `ContractSchema` and infers type from it
- Propagate required changes in codebase
- Update associated tests


src/domain/contracts/entities/contract.entity.ts
```diff
@@ -1,8 +1,4 @@
-export interface Contract {
-  address: string;
-  name: string;
-  displayName: string;
-  logoUri: string;
```
iamacook, line 5: I checked the Transaction Service and this can return `null` which has been reflected in the new `ContractSchema`.

src/routes/common/address-info/address-info.helper.ts
```diff
@@ -100,7 +100,10 @@ export class AddressInfoHelper {
       case 'CONTRACT':
         return this.contractsRepository
           .getContract({ chainId, contractAddress: address })
-          .then((c) => new AddressInfo(c.address, c.displayName, c.logoUri));
+          .then(
+            (c) =>
+              new AddressInfo(c.address, c.displayName, c.logoUri ?? undefined),
```
iamacook, line 105: In accordance with the potential `null`, this needs to fallback to `undefined` as `AddessInfo` only accepts `string | undefined` as a `logoUri`.

src/domain/contracts/entities/contract.entity.ts
hectorgomezv: Good catch! In fact, in the schema it was considered nullable already.

src/routes/common/address-info/address-info.helper.ts
hectorgomezv: Wdyt about leaving a TODO to change the `AddressInfo` constructor to accept `string | null` as a `logoUri` in a different PR?

hectorgomezv (approved):

src/routes/common/address-info/address-info.helper.ts
iamacook: Good point. I've added a comment in ea4c64b57678a9f128a3457f3f6a2ac9ae57e4a6.

hectorgomezv (approved):

PR #1267, opened 2024-03-06
Only allow enabling/disabling recovery alerts for modules of given Safe

iamacook:
## Summary

We currently prevent the user from enabling/disabling recovery alerts unless they submit a valid signature. This does, not, however, mean that the module is tied to the specified Safe.

This includes a check that the specified `moduleAddress` is enabled on that `safeAddress`.

## Changes

- Add check that `moduleAddress` is enabled on `safeAddress` in `EnableRecoveryAlertsGuard`/`DisableRecoveryAlertsGuard`
- Add test coverage


src/routes/recovery/guards/disable-recovery-alerts.guard.ts
```diff
@@ -67,5 +69,17 @@ export class DisableRecoveryAlertsGuard implements CanActivate {
...
+      const { safes } = await this.safeRepository.getSafesByModule({
+        chainId,
+        moduleAddress,
+      });
+
+      return safes.some((safe) => getAddress(safe) === getAddress(safeAddress));
+    } catch (e) {
+      this.loggingService.debug(e);
+      return false;
+    }
```
fmrsabino, line 83: We can merge the try/catch block with the one above. Alternatively you can extract this part to a new `Guard` in favour of composability (e.g.: `SafeHasModuleGuard`).

However for the context of this PR and since we only perform this check here I would say that we can merge the try/catch block first.

fmrsabino (changes requested):

src/routes/recovery/guards/disable-recovery-alerts.guard.ts
iamacook: Merged in c3d6cff8a5ccbc4f53a9e597513c0d73cc46b7f8.

src/routes/recovery/guards/enable-recovery-alerts.guard.ts
```diff
@@ -67,5 +69,17 @@ export class EnableRecoveryAlertsGuard implements CanActivate {
...
+    try {
+      const { safes } = await this.safeRepository.getSafesByModule({
+        chainId,
+        moduleAddress,
+      });
+
+      return safes.some((safe) => getAddress(safe) === getAddress(safeAddress));
+    } catch (e) {
+      this.loggingService.debug(e);
+      return false;
```
fmrsabino, line 82: Same here 🙏 

https://github.com/safe-global/safe-client-gateway/pull/1267#discussion_r1519432903

fmrsabino (changes requested):

src/routes/recovery/guards/enable-recovery-alerts.guard.ts
iamacook: Merged in db117f82d687f40dcc2a002ba2e3b4c8c7faf449.

PR #1268, opened 2024-03-06
Migrate `DataDecoded` to `zod`

iamacook:
## Summary

This migrates the validation of `DataDecoded` to `zod`.

## Changes

- Create `DataDecodedSchema`
- Migrated `DataDecodedRepository['getDataDecoded']` to use new schema
- Add notes to remove to-be-deprecated validator


iamacook:
Marking as a draft until #1276 is merged to reduce the scope.

src/domain/data-decoder/entities/schemas/data-decoded.schema.ts
```diff
@@ -1,5 +1,22 @@
...
+import { z } from 'zod';
 
+export const DataDecodedParameterSchema = z.object({
+  name: z.string(),
+  type: z.string(),
+  // z.unknown() makes the property optional but it should be defined
+  value: z.custom<Required<unknown>>(),
+  valueDecoded: z
+    .union([z.record(z.unknown()), z.array(z.record(z.unknown()))])
+    .optional(),
```
hectorgomezv, line 11: Is there a reason for just keeping this as `optional()` instead of defining it as `nullish().default(null)` as we do in other places?

Also, the second member of the union (`z.array(z.record(z.unknown()))`) implies the array should contain objects, is this correct?

hectorgomezv (approved):
I've left small doubts only, otherwise LGTM.

hectorgomezv (approved):

PR #1272, opened 2024-03-07
Migrate `Event` to `zod`

iamacook:
## Summary

This migrates the validation of `Event` to `zod`:

- `ChainUpdate`
- `DeletedMultisigTransaction`
- `ExecutedTransaction`
- `IncomingEther`
- `IncomingToken`
- `MessageCreated`
- `ModuleTransaction`
- `NewConfirmation`
- `NewMessageConfirmation`
- `OutgoingEther`
- `OutgoingToken`
- `PeningTransaction`
- `SafeAppsUpdate`

## Changes

- Remove `EventValidationPipe` and associated schema.
- Add `WebHookSchema`, constituted of schemas for all above listed events and infer types from them.
- Update associated tests.


src/routes/cache-hooks/cache-hooks.controller.ts
```diff
@@ -17,7 +18,9 @@ export class CacheHooksController {
   @UseGuards(PreExecutionLogGuard, BasicAuthGuard)
   @Post('/hooks/events')
   @HttpCode(202)
-  async postEvent(@Body(EventValidationPipe) event: Event): Promise<void> {
+  async postEvent(
+    @Body(new ValidationPipe(WebHookSchema)) event: Event,
```
fmrsabino, line 22: What is the response code if the we receive an event that we don't know how to handle?

fmrsabino: Nevermind, forgot that we had a global exception filter:

https://github.com/safe-global/safe-client-gateway/blob/ba418d0cd54e5074384a8710ff9176ea11f9ca86/src/routes/common/filters/zod-error.filter.ts#L25-L39

So the expected status code would be 422 which I would say is correct for this route 👍

src/routes/cache-hooks/entities/schemas/__tests__/chain-update.schema.spec.ts
```diff
@@ -0,0 +1,52 @@
...
+          code: 'invalid_literal',
+          expected: 'CHAIN_UPDATE',
+          path: ['type'],
+          message: 'Invalid literal value, expected "CHAIN_UPDATE"',
+        },
+      ]),
+    );
+  });
+
+  it.each([['type' as const], ['chainId' as const]])(
```
fmrsabino, line 36: Do we need to use tuples in this test (over the actual string values)?

fmrsabino (approved):

src/routes/cache-hooks/entities/schemas/__tests__/chain-update.schema.spec.ts
iamacook: Removed in 90ae30c716907fc58d03eb6231abe3ac7eb6e304.

fmrsabino (approved):
🚀 👏

PR #1288, opened 2024-03-12
Allow relaying of all ERC-20 methods

iamacook:
## Summary

This facilitates the relaying of _all_ ERC-20 methods, with validation of `transfer` and `transferFrom` methods:

- `transfer` transactions that are being executed to 'self' (the Safe) _cannot_ be relayed.
- `transferFrom` transactions that are being executed to 'self' or transferring from the `sender` to `receiver` _cannot_ be relayed.

## Motivation

We were previous _only_ allowing "valid" `transfer` relays. It was otherwise [not possible to relay other methods](https://github.com/safe-global/safe-client-gateway/blob/main/src/domain/relay/limit-addresses.mapper.ts#L144).

This is an inherent bug that was ported over from the relay service - it was [_only_ validating `transfer` calldata](https://github.com/safe-global/safe-gelato-relay-service/blob/main/src/routes/relay/entities/schema/transactions/exec-transaction.ts#L41). Other methods were able to relay as they passed the ['self'-validation](https://github.com/safe-global/safe-gelato-relay-service/blob/main/src/routes/relay/entities/schema/transactions/exec-transaction.ts#L47-L48).

Note: although there is no validation of `transferFrom` on the relay service, adding it aligns with that of `transfer`.

## Changes

- Add validation of `transferFrom` relays
- Allow relay of other methods of ERC-20
- Add respective `IEncoder`s for mocking calldata
- Add test coverage

### DIFF
--- a/src/domain/relay/limit-addresses.mapper.ts
+++ b/src/domain/relay/limit-addresses.mapper.ts
@@ -141,8 +141,6 @@ export class LimitAddressesMapper {
       });
     }
 
-    // Other ERC-20 methods are valid if they are...
-
     // Only transaction to other party is valid
     const toSelf = execTransactionArgs.to === args.to;
     if (!toSelf) {
@@ -185,6 +183,7 @@ export class LimitAddressesMapper {
   }
 
   private isValidErc20Transfer(args: { to: string; data: Hex }): boolean {
+    // Can throw but called after this.erc20Decoder.helpers.isTransfer
     const erc20DecodedData = this.erc20Decoder.decodeFunctionData({
       data: args.data,
     });
@@ -200,6 +199,7 @@ export class LimitAddressesMapper {
   }
 
   private isValidErc20TransferFrom(args: { to: string; data: Hex }): boolean {
+    // Can throw but called after this.erc20Decoder.helpers.isTransferFrom
     const erc20DecodedData = this.erc20Decoder.decodeFunctionData({
       data: args.data,
     });
--- a/src/domain/relay/limit-addresses.mapper.ts
+++ b/src/domain/relay/limit-addresses.mapper.ts
@@ -119,63 +119,102 @@ export class LimitAddressesMapper {
   }
 
   private isValidExecTransactionCall(args: { to: string; data: Hex }): boolean {
-    let execTransaction: { data: Hex; to: Hex; value: bigint };
-    // If transaction is an execTransaction
+    const execTransactionArgs = this.getExecTransactionArgs(args.data);
+    // Not a valid execTransaction call
+    if (!execTransactionArgs) {
+      return false;
+    }
+
+    // Only ERC-20 transfer to other party is valid
+    if (this.erc20Decoder.helpers.isTransfer(execTransactionArgs.data)) {
+      return this.isValidErc20Transfer({
+        to: args.to,
+        data: execTransactionArgs.data,
+      });
+    }
+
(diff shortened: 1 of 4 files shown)


src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -134,16 +134,31 @@ export class LimitAddressesMapper {
...
+        const [sender, recipient] = erc20DecodedData.args;
+        return (
+          sender !== recipient &&
+          // TODO: Propagate checksummed address types from RelayDto from controller
+          recipient !== getAddress(args.to)
+        );
+      }
+
+      // Otherwise allow other methods
+      return true;
```
fmrsabino, line 161: We are returning `true` while only checking the ERC20 related data. We have some further checks that avoid calls targeting "self". Imagine that you have an ERC20 method which is not `transfer` and is no` transferFrom`: shouldn't we validate that the `execTransaction` is not targeting the Safe itself also (the transaction would no succeed if the Safe could not handle that method but we would still relay right?).

iamacook: I simplified the code in dbb5ca98c7f795a1bd0dfb466138f35f85e45f35 to explicitly handle `transfer`/`transferFrom` validation in their own conditions, thus allowing all other ERC-20 methods, providing they are not trying to be called on the Safe. The changes simplify the code, as well as include test cases.

src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -119,48 +119,102 @@ export class LimitAddressesMapper {
...
+
+    // Only ERC-20 transferFrom to other party is valid
+    if (this.erc20Decoder.helpers.isTransferFrom(execTransactionArgs.data)) {
+      return this.isValidErc20TransferFrom({
+        to: args.to,
+        data: execTransactionArgs.data,
+      });
+    }
+
+    // Other ERC-20 methods are valid if they are...
```
fmrsabino, line 144: nitpick: from this point it's not really specific to ERC-20 functions but any function.

src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -119,48 +119,102 @@ export class LimitAddressesMapper {
...
     }
+  }
 
-    // If a transaction does not target 'self' consider it valid
-    if (args.to !== execTransaction.to) {
-      return true;
+  private isValidErc20Transfer(args: { to: string; data: Hex }): boolean {
+    const erc20DecodedData = this.erc20Decoder.decodeFunctionData({
+      data: args.data,
+    });
```
fmrsabino, line 190: `decodeFunctionData` can throw. but I think it should be ok to throw here. This function should be called given that we know that we are dealing with an ERC20 Transfer already. Throwing also creates visibility in case we have application bugs.

fmrsabino (approved):
👏 🚀

src/domain/relay/limit-addresses.mapper.ts
iamacook: I removed the comment in edb9749fb6c722676f61148ad36909dfd53d9f79.

src/domain/relay/limit-addresses.mapper.ts
iamacook: I added a comment to keep note in edb9749fb6c722676f61148ad36909dfd53d9f79.

fmrsabino (approved):

PR #1296, opened 2024-03-13
Migrate `SafeApp` to `zod`

hectorgomezv:
## Summary

This migrates the validation of `SafeApp` to `zod`.

## Changes

- Removes `SafeAppValidator` and associated schema.
- Adds `SafeAppSchema` and infers type(s) from it.
- Propagate required changes in codebase.
- Update associated tests.
- Changes `iconUrl` from `string` to `string | null`.


src/domain/safe-apps/entities/schemas/safe-app.schema.ts
```diff
@@ -1,105 +1,31 @@
...
-    'features',
-    'socialProfiles',
-  ],
-};
+import { z } from 'zod';
+
+export const SafeAppAccessControlSchema = z.object({
+  type: z.string(),
+  value: z.array(z.string().url()).nullish().default(null),
+});
```
iamacook, line 6:
```suggestion
export const SafeAppAccessControlSchema = z.discriminatedUnion('type', [
  z.object({
    type: z.literal(SafeAppAccessControlPolicies.DomainAllowlist),
    value: z.array(z.string().url()).nullish().default(null),
  }),
  z.object({
    type: z.literal(SafeAppAccessControlPolicies.NoRestrictions),
  }),
  z.object({
    type: z.literal(SafeAppAccessControlPolicies.Unknown),
  }),
]);
```

src/domain/safe-apps/entities/schemas/safe-app.schema.ts
```diff
@@ -1,105 +1,31 @@
...
+  id: z.number(),
+  url: z.string().url(),
+  name: z.string(),
+  description: z.string(),
+  chainIds: z.array(z.number()),
+  accessControl: SafeAppAccessControlSchema,
+  tags: z.array(z.string()),
+  features: z.array(z.string()),
+  socialProfiles: z.array(SafeAppSocialProfileSchema),
+  iconUrl: z.string().url().nullish().default(null),
```
iamacook, line 28: Outside of the scope of this PR but we need add this to the SDK.

hectorgomezv:
Thanks for the suggestion @iamacook! I've committed it, in fact, I had the exact same code and relaxed the validation later (see [this difff](https://github.com/safe-global/safe-client-gateway/pull/1296/commits/32c103fecb3ea31cce7b1bd5eab5d7aecf730ca7#diff-16274fb3e876760c022692a1e18bab1117c0d8837aac80ea95d1cf3edc6a6eb4)) because I'm not sure it makes a lot of sense to validate the `accessControl` property in the CGW, as its logic is more on the Config Service. Adding a new `AccessControl` type there will break the validation in the CGW, and I think it shouldn't as the CGW doesn't own the logic for this property. 

But I also think your suggestion is right since it should be tackled in a separate PR, and we are purely migrating the validation mechanism here 👍🏻 

Thanks again!

PR #1299, opened 2024-03-13
Add support for decoding CoW Swap orders

fmrsabino:
Part of #1198

- Adds `SwapOrderMapper` – this mapper will attempt to decode transactions that represent a `preSignature` CoW Swap order.
- Successful decoding depends on multiple conditions. If any of the following conditions occurs, the transaction is mapped to a custom transaction instead (i.e.: without any decoding in place):
  * Transaction data is null
  * Order UID cannot be retrieved
  * Buy or Sell token info cannot be retrieved
  * Unsupported order statuses (statuses which are not `fulfilled`, `open`, `cancelled` or `expired`).
- Adds an `orderBuilder` to support creating multiple instances of `Order` with different properties.
- The decoding of CoW Swap orders is only available if `FF_SWAPS_DECODING` is enabled.

### DIFF
--- a/src/domain/swaps/entities/__tests__/order.builder.ts
+++ b/src/domain/swaps/entities/__tests__/order.builder.ts
@@ -16,7 +16,13 @@ export function orderBuilder(): IBuilder<Order> {
     .with('sellAmount', faker.number.bigInt({ min: 1 }))
     .with('buyAmount', faker.number.bigInt({ min: 1 }))
     .with('validTo', faker.date.future().getTime())
-    .with('appData', faker.string.hexadecimal())
+    .with(
+      'appData',
+      JSON.stringify({
+        version: faker.system.semver(),
+        metadata: {},
+      }),
+    )
     .with('feeAmount', faker.number.bigInt({ min: 1 }))
     .with('kind', faker.helpers.arrayElement(['buy', 'sell']))
     .with('partiallyFillable', faker.datatype.boolean())
@@ -40,7 +46,7 @@ export function orderBuilder(): IBuilder<Order> {
     .with('creationDate', faker.date.recent())
     .with('class', faker.helpers.arrayElement(['market', 'limit', 'liquidity']))
     .with('owner', getAddress(faker.finance.ethereumAddress()))
-    .with('uid', faker.string.uuid())
+    .with('uid', faker.string.hexadecimal({ length: 112 }))
     .with(
       'availableBalance',
       faker.datatype.boolean() ? faker.number.bigInt({ min: 1 }) : null,
@@ -66,7 +72,9 @@ export function orderBuilder(): IBuilder<Order> {
       'ethflowData',
       faker.datatype.boolean()
         ? {
-            refundTxHash: faker.string.hexadecimal(),
+            refundTxHash: faker.datatype.boolean()
+              ? faker.string.hexadecimal()
+              : null,
             userValidTo: faker.date.future().getTime(),
           }
         : null,
--- a/src/domain/swaps/entities/__tests__/order.builder.ts
+++ b/src/domain/swaps/entities/__tests__/order.builder.ts
@@ -1,0 +1,101 @@
+import { Order } from '@/domain/swaps/entities/order.entity';
+import { Builder, IBuilder } from '@/__tests__/builder';
+import { faker } from '@faker-js/faker';
+import { getAddress } from 'viem';
+
+export function orderBuilder(): IBuilder<Order> {
+  return new Builder<Order>()
+    .with('sellToken', getAddress(faker.finance.ethereumAddress()))
+    .with('buyToken', getAddress(faker.finance.ethereumAddress()))
+    .with(
+      'receiver',
+      faker.datatype.boolean()
+        ? getAddress(faker.finance.ethereumAddress())
+        : null,
+    )
+    .with('sellAmount', faker.number.bigInt({ min: 1 }))
+    .with('buyAmount', faker.number.bigInt({ min: 1 }))
+    .with('validTo', faker.date.future().getTime())
+    .with('appData', faker.string.hexadecimal())
+    .with('feeAmount', faker.number.bigInt({ min: 1 }))
+    .with('kind', faker.helpers.arrayElement(['buy', 'sell']))
+    .with('partiallyFillable', faker.datatype.boolean())
+    .with(
+      'sellTokenBalance',
+      faker.helpers.arrayElement(['erc20', 'internal', 'external']),
+    )
+    .with('buyTokenBalance', faker.helpers.arrayElement(['erc20', 'internal']))
+    .with(
+      'signingScheme',
+      faker.helpers.arrayElement(['eip712', 'ethsign', 'presign', 'eip1271']),
+    )
+    .with('signature', faker.string.hexadecimal())
+    .with(
+      'from',
+      faker.datatype.boolean()
+        ? getAddress(faker.finance.ethereumAddress())
+        : null,
+    )
+    .with('quoteId', faker.datatype.boolean() ? faker.number.int() : null)
+    .with('creationDate', faker.date.recent())
+    .with('class', faker.helpers.arrayElement(['market', 'limit', 'liquidity']))
+    .with('owner', getAddress(faker.finance.ethereumAddress()))
+    .with('uid', faker.string.uuid())
+    .with(
+      'availableBalance',
+      faker.datatype.boolean() ? faker.number.bigInt({ min: 1 }) : null,
+    )
+    .with('executedSellAmount', faker.number.bigInt({ min: 1 }))
+    .with('executedSellAmountBeforeFees', faker.number.bigInt({ min: 1 }))
+    .with('executedBuyAmount', faker.number.bigInt({ min: 1 }))
+    .with('executedFeeAmount', faker.number.bigInt({ min: 1 }))
+    .with('invalidated', faker.datatype.boolean())
+    .with(
+      'status',
+      faker.helpers.arrayElement([
+        'presignaturePending',
+        'open',
+        'fulfilled',
+        'cancelled',
+        'expired',
+      ]),
+    )
+    .with('fullFeeAmount', faker.number.bigInt({ min: 1 }))
+    .with('isLiquidityOrder', faker.datatype.boolean())
+    .with(
+      'ethflowData',
+      faker.datatype.boolean()
+        ? {
+            refundTxHash: faker.string.hexadecimal(),
+            userValidTo: faker.date.future().getTime(),
+          }
+        : null,
+    )
+    .with(
+      'onchainUser',
+      faker.datatype.boolean()
+        ? getAddress(faker.finance.ethereumAddress())
+        : null,
+    )
+    .with(
+      'onchainOrderData',
+      faker.datatype.boolean()
+        ? {
+            sender: getAddress(faker.finance.ethereumAddress()),
+            placementError: faker.helpers.arrayElement([
+              'QuoteNotFound',
+              'ValidToTooFarInFuture',
+              'PreValidationError',
+            ]),
+          }
+        : null,
+    )
+    .with(
+      'executedSurplusFee',
+      faker.datatype.boolean() ? faker.number.bigInt({ min: 1 }) : null,
+    )
+    .with(
+      'fullAppData',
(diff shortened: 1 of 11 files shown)


src/domain/swaps/entities/__tests__/order.builder.ts
```diff
@@ -0,0 +1,101 @@
...
+        'expired',
+      ]),
+    )
+    .with('fullFeeAmount', faker.number.bigInt({ min: 1 }))
+    .with('isLiquidityOrder', faker.datatype.boolean())
+    .with(
+      'ethflowData',
+      faker.datatype.boolean()
+        ? {
+            refundTxHash: faker.string.hexadecimal(),
```
hectorgomezv, line 69: As this is nullable in the schema, what about `refundTxHash: faker.datatype.boolean() ? faker.string.hexadecimal() : null`

src/routes/transactions/entities/swap-order-info.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+  TransactionInfoType,
+} from '@/routes/transactions/entities/transaction-info.entity';
+import {
+  ApiExtraModels,
+  ApiProperty,
+  ApiPropertyOptional,
+} from '@nestjs/swagger';
+
+export class TokenInfo {
+  @ApiPropertyOptional()
```
hectorgomezv, line 12:
```suggestion
  @ApiPropertyOptional({ type: Number, nullable: true })
```

src/routes/transactions/entities/swap-order-info.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.logo = args.logo;
+    this.symbol = args.symbol;
+    this.amount = args.amount;
+  }
+}
+
+@ApiExtraModels(TokenInfo)
+export abstract class SwapOrderTransactionInfo extends TransactionInfo {
+  @ApiProperty()
+  status: 'open' | 'fulfilled' | 'cancelled' | 'expired';
```
hectorgomezv, line 31: What about exporting these unions into a type? I mean, something like:
```javascript
export type SwapOrderTransactionInfoStatus =
  | 'open'
  | 'fulfilled'
  | 'cancelled'
  | 'expired';

export type DefaultSwapOrderTransactionInfoStatus =
  | 'open'
  | 'cancelled'
  | 'expired';

export type SwapOrderTransactionInfoOrderKind = 'buy' | 'sell';
```

It could be applied to test classes also:
```javascript
.with('status', status as SwapOrderTransactionInfoStatus)
```

hectorgomezv: No strong opinion though, but we could extract the string literals ('open', 'fulfilled', etc.) as well to constants, as they are used in several files (`SwapOrderMapper` for example).

src/domain/swaps/entities/__tests__/order.builder.ts
```diff
@@ -0,0 +1,101 @@
...
+      faker.datatype.boolean()
+        ? getAddress(faker.finance.ethereumAddress())
+        : null,
+    )
+    .with('sellAmount', faker.number.bigInt({ min: 1 }))
+    .with('buyAmount', faker.number.bigInt({ min: 1 }))
+    .with('validTo', faker.date.future().getTime())
+    .with('appData', faker.string.hexadecimal())
+    .with('feeAmount', faker.number.bigInt({ min: 1 }))
+    .with('kind', faker.helpers.arrayElement(['buy', 'sell']))
```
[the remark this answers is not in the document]
fmrsabino: The main reason why I didn't add `unknown` here it's because the default builder should return a "sensible" default i.e.: in a testing scenario, you should most of the times expect a "normal" Order.

However this only applies to the default values. Setting `unknown` is of course still allowed in a testing scenario. We would otherwise need to write the expected values most of the time to avoid having `unknown`.

src/domain/swaps/entities/__tests__/order.builder.ts
```diff
@@ -0,0 +1,101 @@
...
+      'from',
+      faker.datatype.boolean()
+        ? getAddress(faker.finance.ethereumAddress())
+        : null,
+    )
+    .with('quoteId', faker.datatype.boolean() ? faker.number.int() : null)
+    .with('creationDate', faker.date.recent())
+    .with('class', faker.helpers.arrayElement(['market', 'limit', 'liquidity']))
+    .with('owner', getAddress(faker.finance.ethereumAddress()))
+    .with('uid', faker.string.uuid())
```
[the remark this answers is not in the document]
fmrsabino: The `uid` from CoW Swap is actually an hex string. Looks like 112 characters long without the `0x`. I will change the faker value to return something similar instead.

src/domain/swaps/entities/__tests__/order.builder.ts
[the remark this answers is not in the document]
hectorgomezv: Ok, it makes sense, thanks for the explanation 🙂

src/domain/swaps/entities/__tests__/order.builder.ts
```diff
@@ -0,0 +1,103 @@
...
+    .with(
+      'receiver',
+      faker.datatype.boolean()
+        ? getAddress(faker.finance.ethereumAddress())
+        : null,
+    )
+    .with('sellAmount', faker.number.bigInt({ min: 1 }))
+    .with('buyAmount', faker.number.bigInt({ min: 1 }))
+    .with('validTo', faker.date.future().getTime())
+    .with('appData', faker.string.hexadecimal())
```
iamacook, line 19: Out of scope of this PR, but is there a reason we aren't using the `HexSchema` for the `faker.string.hexadecimal` values here?

src/routes/transactions/entities/swap-order-info.entity.ts
```diff
@@ -0,0 +1,99 @@
...
+    this.orderKind = args.orderKind;
+    this.sellToken = args.sellToken;
+    this.buyToken = args.buyToken;
+    this.expiresTimestamp = args.expiresTimestamp;
+  }
+}
+
+@ApiExtraModels(TokenInfo)
+export class FulfilledSwapOrderTransactionInfo extends SwapOrderTransactionInfo {
+  @ApiPropertyOptional()
```
iamacook, line 63:
```suggestion
  @ApiPropertyOptional({ type: String, nullable: true })
```

src/routes/transactions/mappers/common/swap-order.mapper.ts
```diff
@@ -0,0 +1,239 @@
...
+  async mapSwapOrder(
+    chainId: string,
+    transaction: MultisigTransaction | ModuleTransaction,
+    dataSize: number,
+  ): Promise<SwapOrderTransactionInfo | CustomTransactionInfo> {
+    if (!isHex(transaction.data)) {
+      return this._mapUnknownOrderStatus(chainId, transaction, dataSize);
+    }
+
+    const orderUid: `0x${string}` | null =
```
iamacook, line 82:
```suggestion
    const orderUid =
```

src/routes/transactions/mappers/common/swap-order.mapper.ts
```diff
@@ -0,0 +1,239 @@
...
+
+  private _mapFulfilledOrderStatus(args: {
+    buyToken: TokenAmount;
+    sellToken: TokenAmount;
+    order: Order;
+  }): SwapOrderTransactionInfo {
+    if (args.order.kind === 'unknown') {
+      throw new Error('Unknown order kind');
+    }
+    const surplusFee: string | null = args.order.executedSurplusFee
```
iamacook, line 180:
```suggestion
    const surplusFee = args.order.executedSurplusFee
```

src/routes/transactions/mappers/common/swap-order.mapper.ts
```diff
@@ -0,0 +1,239 @@
...
+      expiresTimestamp: args.order.validTo,
+      surplus: surplusFee,
+      executionPrice: this._getExecutionPrice(args.sellToken, args.buyToken),
+    });
+  }
+
+  private _getExecutedSurplusFee(
+    executedSurplusFee: bigint,
+    token: Token & { decimals: number },
+  ): string | null {
```
iamacook, line 200:
```suggestion
  ): string {
```

src/domain/swaps/entities/__tests__/order.builder.ts
fmrsabino: `appData` shouldn't be considered an hexadecimal but a stringified JSON. I will update it.

src/routes/transactions/mappers/common/swap-order.mapper.ts
fmrsabino: Do you prefer to have the type removed?

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: For this use-case I actually think that having the union declared in-place is simpler (one less redirection for the reader of what the value might be).

If we declared the union type, we would still need to specify the respective values in `SwapOrderMapper` (maybe with the exception of the `switch/case` where we could use the declared union type).

src/routes/transactions/mappers/common/swap-order.mapper.ts
fmrsabino: Changed in [e3a4a9c](https://github.com/safe-global/safe-client-gateway/pull/1299/commits/e3a4a9c33aac2d31572dd3cb51e6cacec9ed3515)

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: Changed in [e3a4a9c](https://github.com/safe-global/safe-client-gateway/pull/1299/commits/e3a4a9c33aac2d31572dd3cb51e6cacec9ed3515)

src/domain/swaps/entities/__tests__/order.builder.ts
fmrsabino: Changed in [e3a4a9c](https://github.com/safe-global/safe-client-gateway/pull/1299/commits/e3a4a9c33aac2d31572dd3cb51e6cacec9ed3515)

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: Changed in [e3a4a9c](https://github.com/safe-global/safe-client-gateway/pull/1299/commits/e3a4a9c33aac2d31572dd3cb51e6cacec9ed3515) (should be `String` though).

src/domain/swaps/entities/__tests__/order.builder.ts
fmrsabino: Changed in [e3a4a9c](https://github.com/safe-global/safe-client-gateway/pull/1299/commits/e3a4a9c33aac2d31572dd3cb51e6cacec9ed3515)

src/domain/swaps/entities/__tests__/order.builder.ts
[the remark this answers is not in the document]
fmrsabino: Changed in [e3a4a9c](https://github.com/safe-global/safe-client-gateway/pull/1299/commits/e3a4a9c33aac2d31572dd3cb51e6cacec9ed3515)

src/routes/transactions/mappers/common/swap-order.mapper.ts
fmrsabino: I've updated the names to be more explicit on what we expect of the value (since it's a label) – [b56ec3b](https://github.com/safe-global/safe-client-gateway/pull/1299/commits/b56ec3bbc1530ceb762c0a77d495bb99368bff3c)

src/routes/transactions/entities/swap-order-info.entity.ts
hectorgomezv: Oops, `String`, yes 😄

Thanks!

src/routes/transactions/entities/swap-order-info.entity.ts
hectorgomezv: Ok, it makes sense. For `SwapOrderMapper` I was thinking about using not the union but static constants, but I agree with you, maybe it adds more complexity overall. I'm fine with your version, thanks for looking into it!

hectorgomezv (approved):

PR #1302, opened 2024-03-14
Add OrderSchema tests

fmrsabino:
## Summary

- Adds unit tests to `OrderSchema`.

## Changes

- Changes `.default('unknown')` to `.catch('unknown')` in the `OrderSchema` as the default value would only be set if no value was provided. If an unknown value is provided we should still set the respective value as `unknown`.

### DIFF
--- a/src/domain/swaps/entities/schemas/order.schema.spec.ts
+++ b/src/domain/swaps/entities/schemas/order.schema.spec.ts
@@ -14,10 +14,9 @@ describe('OrderSchema', () => {
   });
 
   it('hexadecimal signature should be valid', () => {
-    const order = {
-      ...orderBuilder().build(),
-      signature: faker.string.hexadecimal() as `0x${string}`,
-    };
+    const order = orderBuilder()
+      .with('signature', faker.string.hexadecimal() as `0x${string}`)
+      .build();
 
     const result = OrderSchema.safeParse(order);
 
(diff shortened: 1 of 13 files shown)


src/domain/swaps/entities/schemas/order.schema.ts
```diff
@@ -62,7 +62,7 @@ export const OrderSchema = z.object({
           'PreValidationError',
           'unknown',
         ])
-        .default('unknown')
+        .catch('unknown')
         .nullish()
         .default(null),
```
iamacook, line 67:
```suggestion
        .catch('unknown'),
```

iamacook (changes requested):
Can we also add test conditions for:

- Address checksumming of `sellToken`, `buyToken`, `receiver`, `from`, `owner`, `onChainUser` and `onchainUser.sender`
- Defaulting to `null` for `receiver`, `from`, `quoteId`, `availableBalance`, `ethflowData`, `onchainUser`, `executedSurplusFee` and `fullAppData`.

I've also noticed that there are some hexadecimal values that could use the `HexSchema` in the `OrderSchema` that could then be tested as well: `signature` and `ethflowData.refundTxHash`.

src/domain/swaps/entities/schemas/order.schema.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+
+    const result = OrderSchema.safeParse(order);
+
+    expect(result.success).toBe(true);
+  });
+
+  it('hexadecimal signature should be valid', () => {
+    const order = {
+      ...orderBuilder().build(),
+      signature: faker.string.hexadecimal() as `0x${string}`,
```
iamacook, line 19: Is there a reason to now use `with` in these tests?

src/domain/swaps/entities/schemas/order.schema.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  it('should fallback to unknown kind on order with an invalid kind', () => {
+    const order = { ...orderBuilder().build(), kind: faker.string.sample() };
+
+    const result = OrderSchema.safeParse(order);
+
+    expect(result.success && result.data.kind).toBe('unknown');
+  });
+
+  it.each<keyof Order>([
```
iamacook, line 114: Nice type usage!

iamacook (approved):

src/domain/swaps/entities/schemas/order.schema.spec.ts
fmrsabino: No reason! Will fix it! 👍

fmrsabino: I've fixed it in [107cf08](https://github.com/safe-global/safe-client-gateway/pull/1302/commits/107cf0878fbb2dead19a4885feadb58ec472fd1c).

Additionally I merged the "unknown" tests into one (parameterised test).

iamacook (approved):

PR #1315, opened 2024-03-18
Migrate `Message`/`MessageConfirmation` to `zod`

iamacook:
## Summary

This migrates the validation of `Message` to `zod`.

## Changes

- Remove `MessageValidator` and associated schemas.
- Add test-covered `MessageSchema`/`MessageConfirmationSchema` and infer types from them.
- Propagate type requirements.
- Update tests accordingly.


src/domain/messages/entities/schemas/message.schema.ts
```diff
@@ -1,57 +1,27 @@
...
+import { z } from 'zod';
 
-export const MESSAGE_CONFIRMATION_SCHEMA_ID =
-  'https://safe-client.safe.global/schemas/messages/message-confirmation.json';
+export const MessageConfirmationSchema = z.object({
+  created: z.coerce.date(),
+  modified: z.coerce.date(),
+  owner: AddressSchema,
+  signature: HexSchema,
+  signatureType: z.nativeEnum(SignatureType),
```
iamacook, line 12: I didn't add a fallback here as we control this type and would otherwise need to extend `SignatureType` with `UNKNOWN`, for example. Would appreciate your thoughts.

hectorgomezv: I think it's OK to not have a fallback value, as you said we own this type internally and it haven't changed for years. And if it does and we don't notice, we should be able to catch the related errors in development environments.

src/domain/messages/entities/schemas/__tests__/message.schema.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+
+        expect(
+          !result.success &&
+            result.error.issues.length === 1 &&
+            result.error.issues[0].path.length === 1 &&
+            result.error.issues[0].path[0] === key,
+        ).toBe(true);
+      },
+    );
+  });
```
hectorgomezv, line 246: Nit: these two tests are independent of the type of the items in the Page, what about extracting them to a `page.schema.spec.ts`? Otherwise we would be repeating them for all the `Page<T>` types.

This is not an strong opinion btw, and it would also require to export `PageSchema` somewhere as well, so I think this can be tackled in a separate PR if you agree with the suggestion.

hectorgomezv (approved):

src/domain/messages/entities/schemas/__tests__/message.schema.spec.ts
iamacook: It's a valid point, which I agree with. I removed them in 1a939c76af3ca4fbea86456f98c2d9428aa2758f and will create a follow up PR for the `PageSchema`.

hectorgomezv: Great, thanks!

hectorgomezv (approved):

hectorgomezv (approved):

PR #1317, opened 2024-03-18
Migrate Token to zod

hectorgomezv:
## Summary

This migrates the validation of `Token` to `zod`.

## Changes

- Remove `TokenValidator` and associated schemas.
- Add test-covered `TokenSchema` and infer type from it.
- Propagate type requirements.
- Update tests accordingly.


src/domain/tokens/entities/schemas/token.schema.ts
```diff
@@ -1,37 +1,16 @@
...
+import { TokenType } from '@/domain/tokens/entities/token.entity';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { z } from 'zod';
 
-export const TOKEN_SCHEMA_ID =
-  'https://safe-client.safe.global/schemas/tokens/token.json';
+export const TokenSchema = z.object({
+  address: AddressSchema,
+  decimals: z.number().nullish().default(null),
+  logoUri: z.string(),
```
iamacook, line 9: I would've said that this could be a `z.string().url()` but I see it's just a `string` in AJV. Perhaps we should revisit this later.

src/domain/tokens/entities/schemas/token.schema.ts
```diff
@@ -1,37 +1,16 @@
...
-};
-
-export const TOKEN_PAGE_SCHEMA_ID =
-  'https://safe-client.safe.global/schemas/tokens/token-page.json';
-
-export const tokenPageSchema: Schema = buildPageSchema(
-  TOKEN_PAGE_SCHEMA_ID,
-  tokenSchema,
-);
+export const TokenPageSchema = buildZodPageSchema(TokenSchema);
```
iamacook, line 16: Can we add tests for these?

hectorgomezv: Added in [316e401](https://github.com/safe-global/safe-client-gateway/pull/1317/commits/316e401981f992491f825cf2420f3cece771a46f)

And adapted to the new constraints in [cccaf2e](https://github.com/safe-global/safe-client-gateway/pull/1317/commits/cccaf2ede0ebd3f067c10490ef7b1270cf250163)

src/domain/tokens/entities/schemas/token.schema.ts
hectorgomezv: I agree, this maybe this would be nullable as well according to the Transaction Service docs, but we can revisit it later as it might need adjustments in other parts of the code as well 🙂

iamacook (approved):

PR #1321, opened 2024-03-18
PoC: SiWe with JWTs

iamacook:
This demonstrates an implementation of [SiWe](https://docs.login.xyz/general-information/siwe-overview/eip-4361) with JWTs. A user can authenticate themselves and remain authenticated by access tokens:

1. User requests `nonce` from `GET` `/v1/auth/nonce`.
2. User signs the `nonce` in a SiWe message and send the signature and message (as an object) to `POST` `/v1/auth/verify`.
3. If valid, an `accessToken` and `tokenType` is returned for `Authorization` header (“Bearer xyz”).

Included is an `AuthGuard` that validates the above `accessToken` when attached to a request, allowing the request if valid or throwing if not.

Note: the official SiWe package allows for string or object messages but uses `ethers`. As we are using `viem`, a standard-matching schema is used in place, allowing us to:

1. Reduce dependency reliance
2. Validate the input
3. Infer the type


src/datasources/jwt/jwt.module.ts
```diff
@@ -0,0 +1,43 @@
...
+      options?: jwt.SignOptions,
+    ): string => {
+      return jwt.sign(payload, secret, options);
+    },
+    verify: <T>(
+      token: string,
+      options?: Omit<jwt.VerifyOptions, 'complete'>,
+    ): T => {
+      // As we are not `complete` decoding, it will be T if verified
+      return jwt.verify(token, secret, options) as T;
```
iamacook, line 24: Should we opt for JWTs, we shoul also set [the issuer, etc.](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1) to confirm the expected entity signed it.

iamacook:
I am going to close this as it has been broken down for individual review.

PR #1338, opened 2024-03-20
Add filledPercentage to SwapOrderTransactionInfo

fmrsabino:
Adds `filledPercentage` to the `SwapOrderTransactionInfo` – it represents how much the order was filled in percentual terms (returning the first two decimal digits).

### DIFF
--- a/src/routes/transactions/entities/swap-order-info.entity.ts
+++ b/src/routes/transactions/entities/swap-order-info.entity.ts
@@ -51,7 +51,10 @@ export abstract class SwapOrderTransactionInfo extends TransactionInfo {
   @ApiProperty({ description: 'The timestamp when the order expires' })
   expiresTimestamp: number;
 
-  @ApiProperty({ description: 'The filled percentage of the order' })
+  @ApiProperty({
+    description: 'The filled percentage of the order',
+    examples: ['0.00', '50.75', '100.00'],
+  })
   filledPercentage: string;
 
   protected constructor(args: {
--- a/src/routes/transactions/entities/swap-order-info.entity.ts
+++ b/src/routes/transactions/entities/swap-order-info.entity.ts
@@ -51,12 +51,16 @@ export abstract class SwapOrderTransactionInfo extends TransactionInfo {
   @ApiProperty({ description: 'The timestamp when the order expires' })
   expiresTimestamp: number;
 
+  @ApiProperty({ description: 'The filled percentage of the order' })
+  filledPercentage: string;
+
   protected constructor(args: {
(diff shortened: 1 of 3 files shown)


src/routes/transactions/entities/swap-order-info.entity.ts
```diff
@@ -51,12 +51,16 @@ export abstract class SwapOrderTransactionInfo extends TransactionInfo {
   @ApiProperty({ description: 'The timestamp when the order expires' })
   expiresTimestamp: number;
 
+  @ApiProperty({ description: 'The filled percentage of the order' })
```
iamacook, line 54: Out of the scope of this PR but what do you think about adding `example` to these? I understand the value from the PR description, but maybe not in a few months 🙈

src/routes/transactions/mappers/common/swap-order.mapper.spec.ts
```diff
@@ -380,4 +382,61 @@ describe('Swap Order Mapper tests', () => {
...
+        order = {
+          ...order,
+          executedBuyAmount: BigInt(executedAmount),
+          buyAmount: BigInt(amount),
+        };
+      } else if (order.kind === 'sell') {
+        order = {
+          ...order,
+          executedSellAmount: BigInt(executedAmount),
+          sellAmount: BigInt(amount),
```
iamacook, line 420: Nit question: is there a reason we spread the object and not just assign the values directly, e.g.
```suggestion
        order.executedBuyAmount = BigInt(executedAmount)
        order.buyAmount = BigInt(amount)
      } else if (order.kind === 'sell') {
        order.executedSellAmount = BigInt(executedAmount)
        order.sellAmount = BigInt(amount)
```

iamacook (approved):

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: Good point! Added examples in [7c939f5](https://github.com/safe-global/safe-client-gateway/pull/1338/commits/7c939f5a5ebab7af689bd6e372923f188ce5a215)

src/routes/transactions/mappers/common/swap-order.mapper.spec.ts
fmrsabino: No reason. I'm used to field immutability by default so I created an object with a "copy" of the other one while overriding some values. Fixed here [7c939f5](https://github.com/safe-global/safe-client-gateway/pull/1338/commits/7c939f5a5ebab7af689bd6e372923f188ce5a215)

iamacook (approved):

PR #1348, opened 2024-03-27
Add support for decoding Swap Orders in Multisends

fmrsabino:
- Swap Orders inside a Multisend should be decoded in the same way as regular ones.
- Detection of a Multisend order happens on the first level of transactions (i.e. it does not detect Swap transactions if a Multisend contains another Multisend).
- The `mapSwapOrder` function also changed:
  * It is no longer restricted to a `MultisigTransaction` or `ModuleTransaction` – this was done to support nested transactions which do not have all the properties of a Multisig or Module transactions.
  * It no longer maps to a `CustomTransactionInfo` – this is now a decision left for the callers. Instead, the promise is rejected if the swap transaction couldn't be decoded successfully.

### DIFF
--- a/src/routes/transactions/mappers/common/transaction-info.mapper.ts
+++ b/src/routes/transactions/mappers/common/transaction-info.mapper.ts
@@ -19,6 +19,10 @@ import { NativeCoinTransferMapper } from '@/routes/transactions/mappers/common/n
 import { SettingsChangeMapper } from '@/routes/transactions/mappers/common/settings-change.mapper';
 import { SetPreSignatureDecoder } from '@/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper';
 import { SwapOrderMapper } from '@/routes/transactions/mappers/common/swap-order.mapper';
+import { MultiSendDecoder } from '@/domain/contracts/decoders/multi-send-decoder.helper';
+import { isHex } from 'viem';
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
+import { SwapOrderTransactionInfo } from '@/routes/transactions/entities/swap-order-info.entity';
 
 @Injectable()
 export class MultisigTransactionInfoMapper {
@@ -43,6 +47,7 @@ export class MultisigTransactionInfoMapper {
     @Inject(ITokenRepository) private readonly tokenRepository: TokenRepository,
     @Inject(IConfigurationService)
     private readonly configurationService: IConfigurationService,
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
     private readonly dataDecodedParamHelper: DataDecodedParamHelper,
     private readonly customTransactionMapper: CustomTransactionMapper,
     private readonly settingsChangeMapper: SettingsChangeMapper,
@@ -52,6 +57,7 @@ export class MultisigTransactionInfoMapper {
     private readonly humanDescriptionMapper: HumanDescriptionMapper,
     private readonly setPreSignatureDecoder: SetPreSignatureDecoder,
     private readonly swapOrderMapper: SwapOrderMapper,
+    private readonly multisendDecoder: MultiSendDecoder,
   ) {
     this.isRichFragmentsEnabled = this.configurationService.getOrThrow(
       'features.richFragments',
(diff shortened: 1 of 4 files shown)


src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -52,6 +57,7 @@ export class MultisigTransactionInfoMapper {
     private readonly humanDescriptionMapper: HumanDescriptionMapper,
     private readonly setPreSignatureDecoder: SetPreSignatureDecoder,
     private readonly swapOrderMapper: SwapOrderMapper,
+    private readonly multisendDecoder: MultiSendDecoder,
```
hectorgomezv, line 60: nitpick: `multiSendDecoder`

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -180,13 +185,73 @@ export class MultisigTransactionInfoMapper {
...
+  private findSwapOrder(data: `0x${string}`): `0x${string}` | null {
+    // The swap order can be in the transaction data directly
+    if (this.isSwapOrder({ data })) {
+      return data;
+    }
+    // or in the data of a multisend transaction
+    if (this.multisendDecoder.helpers.isMultiSend(data)) {
+      const transactions = this.multisendDecoder.mapMultiSendTransactions(data);
+      // TODO If we can build a sorted hash map of the transactions, we can avoid iterating all of them
+      //  as we know the pattern of a Swap Order.
```
hectorgomezv, line 245: Sounds like a good solution, but building the hash map would also imply to iterate the full array, wouldn't? The worst case scenario is the `transactions` array does not contain a swap order so the iteration reaches the last element. 

Btw, related to it: is it possible to have several swap orders in the same multisend transaction? My understanding here is only one can be present, as the code is returning the first occurrence, is this correct?

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -180,13 +185,73 @@ export class MultisigTransactionInfoMapper {
     );
   }
 
-  private isCoWSwapOrder(
-    transaction: MultisigTransaction | ModuleTransaction,
-  ): boolean {
+  private isSwapOrder(transaction: { data?: `0x${string}` }): boolean {
```
hectorgomezv, line 188: Very nitpicky: I'm under the impression that `mapSwapOrder` determines whether a transaction is a swap order or not. So this function name is a bit confusing at first glance, I'd say this one checks the data within the transaction, what about renaming to `isSwapOrderData`? Do you think it would make the class API clearer? 

Not strong opinion, I left this up to you.

hectorgomezv (commented):
Great job! Only small comments 🙂

src/routes/transactions/mappers/common/transaction-info.mapper.ts
fmrsabino: > Sounds like a good solution, but building the hash map would also imply iterating the full array, wouldn't it? The worst-case scenario is the `transactions` array does not contain a swap order so the iteration reaches the last element.

Correct. The Hash Map solution would be mostly to optimise accessing the collection while keeping the iterative properties.

> Btw, related to it: is it possible to have several swap orders in the same multisend transaction? My understanding here is only one can be present, as the code is returning the first occurrence, is this correct?

It is. However the interface (and the CGW) will only support decoding of a single transaction in that collection. The expected use case is to have a MultiSend with a Token Approval + Swapping (although the MultiSend can contain anything).

src/routes/transactions/mappers/common/transaction-info.mapper.ts
fmrsabino: > Very nitpicky: I'm under the impression that `mapSwapOrder` determines whether a transaction is a swap order or not. So this function name is a bit confusing at first glance, I'd say this one checks the data within the transaction, what about renaming it to `isSwapOrderData`? Do you think it would make the class API clearer?
> 
> Not a strong opinion btw, I left this up to you 🙂

The goal of `mapSwapOrder` is to map a transaction to a Swap Order, it's not really to check if the transaction is a Swap Order although it can be used for that (nullability test).

The `isSwapOrder` is a utility function which works on the "transaction" level as in, it should work with any object that is a transaction-like (that's why the requirement of the argument is to have an optional `data` argument).

So in summary:
- `isSwapOrder` – returns `true` if the transaction is a Swap Order.
- `mapSwapOrder` – returns the `SwapOrder` if the transaction was successfully mapped to a Swap Order, `null` otherwise.

src/routes/transactions/mappers/common/transaction-info.mapper.ts
fmrsabino: [6c234fe](https://github.com/safe-global/safe-client-gateway/pull/1348/commits/6c234fe792f2987ed822ccbdf2374ea860a67b70)

hectorgomezv (approved):
🚀🚀🚀

PR #1356, opened 2024-04-03
Set Cache-Control to no-cache

fmrsabino:
- Adds a Nest interceptor which, when applied, sets the `Cache-Control` directive to `no-cache`. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#directives
- Adjusts the redirects on the `RelayLegacyController`:
  * With `res.redirect`, a response was being sent to the clients, and the interceptor would fail to set the header (resulting in a 500 – "Cannot set headers after they are sent to the client").
  * By returning an object with `{ url }` with the intended redirect in addition to the decorator `@Redirect`, the header is set correctly before the body is sent out to the clients. See https://docs.nestjs.com/controllers#redirection


src/cache-control.interceptor.ts
```diff
@@ -0,0 +1,25 @@
+import {
```
iamacook, line 1: Nit: Should we move this to `src/routes/common/inteceptors`?

fmrsabino: Agree 100%. Changed in [73f6ad7](https://github.com/safe-global/safe-client-gateway/pull/1356/commits/73f6ad71c0fc47c1625ce77b689eda56ab7e7cc9).

iamacook (approved):

PR #1357, opened 2024-04-03
Add `AuthApiModule` datasource

iamacook:
## Summary

This adds the `AuthApiModule` datasource as part of a test implementation for SiWe authentication, breaking down https://github.com/safe-global/safe-client-gateway/pull/1321 for review.

It assumes the [EIP-4361 (SiWe) standard](https://eips.ethereum.org/EIPS/eip-4361) for authentication. Relative type names are therefore explicit.

Note: a proprietary implementation was chosen for SiWe message parsing as it reduces the necessity for the `siwe-parser` package. Said package is also reliant on `ethers`, which we do not use in the project.

## Changes

- Add `AuthApiModule`/`SiweApiService`, adhering to `IAuthApi`:
  - `generateNonce` returns an alphanumic nonce of at least 8 characters.
  -  `verifyMessage` accepts and validates a signed SiWe-standard message
- `toSignableSiwMessage` converts an object-based SiWe message to a standard-adherent string that would've been signed.
- Appropriate test coverage for the above.

### DIFF
--- a/src/datasources/auth-api/utils/__tests__/to-signable-siwe-message.spec.ts
+++ b/src/datasources/auth-api/utils/__tests__/to-signable-siwe-message.spec.ts
@@ -1,0 +1,328 @@
+import { toSignableSiweMessage } from '@/datasources/auth-api/utils/to-signable-siwe-message';
+import { siweMessageBuilder } from '@/domain/auth/entities/__tests__/siwe-message.builder';
+import { faker } from '@faker-js/faker';
+
+describe('toSignableSiweMessage', () => {
+  it('should return a signable message with all fields', () => {
+    const message = siweMessageBuilder()
+      .with('resources', [
+        faker.internet.url(),
+        faker.internet.url(),
+        faker.internet.url(),
+      ])
+      .build();
+
+    const result = toSignableSiweMessage(message);
+
+    // Origin is built correctly from scheme and domain
+    expect(result)
+      .toBe(`${message.scheme}://${message.domain} wants you to sign in with your Ethereum account:
+${message.address}
+
+${message.statement}
+
+URI: ${message.uri}
+Version: ${message.version}
+Chain ID: ${message.chainId}
+Nonce: ${message.nonce}
+Issued At: ${message.issuedAt}
+Expiration Time: ${message.expirationTime}
+Not Before: ${message.notBefore}
+Request ID: ${message.requestId}
+Resources:
+- ${message.resources![0]}
+- ${message.resources![1]}
+- ${message.resources![2]}`);
+  });
+
+  describe('statement', () => {
+    it('should add a new line before and after the statement', () => {
+      const message = siweMessageBuilder()
+        .with('scheme', undefined)
+        .with('statement', faker.lorem.sentence())
+        .with('expirationTime', undefined)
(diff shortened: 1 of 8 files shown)


src/datasources/auth-api/utils/to-signable-siwe-message.ts
```diff
@@ -0,0 +1,53 @@
+import { SiweMessage } from '@/domain/auth/entities/siwe-message.entity';
+
+/**
+ * The following adheres to the EIP-4361 (SiWe) standard
+ * {@link https://eips.ethereum.org/EIPS/eip-4361}
+ */
+export function toSignableSiweMessage(message: SiweMessage): string {
```
hectorgomezv, line 7: Nitpick: Is this intended to be used outside the scope of `SiweApi`? I mean, if it's only used by that class, have you considered to include it there to increase encapsulation, instead of exporting the function globally?

src/datasources/auth-api/utils/__tests__/to-signable-siwe-message.spec.ts
```diff
@@ -0,0 +1,328 @@
...
+URI: ${message.uri}
+Version: ${message.version}
+Chain ID: ${message.chainId}
+Nonce: ${message.nonce}
+Issued At: ${message.issuedAt}`);
+    });
+  });
+
+  describe('requestId', () => {
+    it('should add the requestId if present', () => {
```
hectorgomezv, line 182: Nitpick: I see the value on having tests verifying the absence of these optional fields, but the case where they are present should be convered by the first one (`should return a signable message with all fields`). Maybe we could get rid of some of them, wdyt? No strong opinion though, I left this totally up to you, if you see value on having them I'm totally OK with it.

hectorgomezv (approved):

src/datasources/auth-api/utils/to-signable-siwe-message.ts
iamacook: I would like to encapsulate but it [will be needed for testing](https://github.com/safe-global/safe-client-gateway/blob/dbb327c929120737eb5b21905400c405c56ab28b/src/routes/auth/auth.controller.spec.ts#L112) the controller.

It is only "used" here (outside of tests) so I think it belongs under the `auth-api`. What do you think?

src/datasources/auth-api/utils/__tests__/to-signable-siwe-message.spec.ts
iamacook: I'd be in favour of keeping them. I wanted to remain as explicit as possible because it adheres to the [message field standards](https://eips.ethereum.org/EIPS/eip-4361#message-fields) (we should ideally be using the [official package](https://www.npmjs.com/package/siwe) - reasons for not are explained in the description).

src/datasources/auth-api/utils/to-signable-siwe-message.ts
hectorgomezv: Hmm, then I'd say it makes sense to keep it as it is (globally exported function). To access it in that testing class we would need to export the function in `IAuthApi` and it makes no sense IMO, as it is purely SiWe-related.

Thanks for explaining!

PR #1358, opened 2024-04-03
Allow decoding of `multiSend` `transactions` with empty `data`

iamacook:
## Summary

This fixes errors thrown when decoding "empty" `data` of a batched `multiSend` transaction.

```
Slice starting at offset "170" is out-of-bounds (size: 170).

Version: viem@2.9.6
```

## Changes

If attempting to `slice` the `data` of a transaction at the same start/end byte offset, empty date (`0x`) is returned instead.


fmrsabino (approved):

src/domain/contracts/decoders/multi-send-decoder.helper.ts
```diff
@@ -60,11 +60,11 @@ export class MultiSendDecoder extends AbiDecoder<typeof MultiSendCallOnly130> {
...
       );
 
-      const data = slice(
-        transactions,
-        cursor,
-        (cursor += hexToNumber(dataLength)),
-      );
+      const data =
+        // slice throws if start/end byte offsets are the same
+        hexToNumber(dataLength) === 0
```
hectorgomezv, line 65: I was wondering if a condition more inline with the comment would be to check `cursor === nextCursor` but it might make the code a bit more complex, I think it's good as it is.

But as a nitpick, maybe we can extract `hexToNumber(dataLength)` to a variable in order to avoid calculating it twice.

hectorgomezv (approved):

src/domain/contracts/decoders/multi-send-decoder.helper.ts
iamacook: > I was wondering if a condition more inline with the comment would be to check `cursor === nextCursor` but it might make the code a bit more complex, I think it's good as it is.
> 
> But as a nitpick, maybe we can extract `hexToNumber(dataLength)` to a variable in order to avoid calculating it twice.

I played around a bit and I think it is better as is. After extracting the `hexToNumber` to a variable (3b2ee6196ad89df711c1e70240e6a74870890232), I think it is more self-documenting.

hectorgomezv (approved):
🚀

PR #1359, opened 2024-04-03
Add default URI of locking service

iamacook:
## Summary

This adds the production locking service as the `locking.baseUri` default value.


fmrsabino (approved):
🚀

PR #1360, opened 2024-04-03
Add `SiweMessageSchema`

iamacook:
## Summary

This adds the `SiweMessageSchema` as part of a test implementation for SiWe authentication, breaking down https://github.com/safe-global/safe-client-gateway/pull/1321 for review.

It assumes the [EIP-4361 (SiWe) standard](https://eips.ethereum.org/EIPS/eip-4361) for authentication. Relative type names are therefore explicit.

This will be used to validate incoming signature requests and that they adhere to the standard, as well as when parsing verified tokens from clients.

Note: a proprietary implementation was chosen for SiWe message parsing as it reduces the necessity for the `siwe-parser` package. Said package is also reliant on `ethers`, which we do not use in the project.

## Changes

- Add standard-aherent `SiweMessageSchema` and associated tests.


hectorgomezv (approved):
👏🏻👏🏻

PR #1363, opened 2024-04-04
Remove `JwtModule` from global scope and validation of its env. vars.

iamacook:
## Summary

This removes the `JwtModule` from the global scope, as well as validation of its env. vars.

## Changes

- Removes the `Global` decorator from the `JwtModule`
- Remove validation of `JWT_ISSUER` and `JWT_SECRET` env. vars.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -102,6 +102,11 @@
 # The AUTH_TOKEN should always be set
 #AUTH_TOKEN=
 
+# The issuer and secret used to sign and verify JWTs.
+# The JWT_ISSUER and JWT_SECRET should always be set
+#JWT_ISSUER=
+#JWT_SECRET=
+
 # Log level for the service.
 #LOG_LEVEL=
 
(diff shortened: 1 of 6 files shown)


fmrsabino (commented):
I think the main issue is because the module is `@Global`. The Module is not being used anywhere so it shouldn't load. But the fact that it is `@Global` would make NestJS build it and make it available in the dependency graph.

https://github.com/safe-global/safe-client-gateway/blob/1c172c5a00aac6cb2e95358501cc74ba389ee470/src/datasources/jwt/jwt.module.ts#L17

iamacook:
> I think the main issue is because the module is `@Global`. The Module is not being used anywhere so it shouldn't load. But the fact that it is `@Global` would make NestJS build it and make it available in the dependency graph.

Good point. This is a remnant from the PoC. I removed the decorator in 1c6cd00e29c5a759ee4e8cce8a573c1bd980ebab.

fmrsabino (approved):
🚀

PR #1367, opened 2024-04-04
Remove DomainModule

fmrsabino:
- The `DomainModule` was removed in favor of smaller domain modules.
- Each feature now should explicitly declare which section of the domain is required in their respective modules.
- This increases encapsulation of the provided features. For Dynamic Modules, it allows loading only the required components instead of the entire domain.


hectorgomezv (approved):
👏🏻👏🏻

iamacook (dismissed):
As a follow up, we should align the style of the existing domain-specific modules. There are some that are in separate files, e.g. `*.domain.module.ts`.

fmrsabino:
> As a follow up, we should align the style of the existing domain-specific modules. There are some that are in separate files, e.g. `*.domain.module.ts`.

I agree. I've left them in the same file as most of the new ones are only about the component that is present on the same file.

iamacook (approved):

PR #1368, opened 2024-04-05
Add `AuthController`

iamacook:
## Summary

This adds the `AuthController` as part of a test implementation for SiWe authentication, breaking down https://github.com/safe-global/safe-client-gateway/pull/1321 for review.

## Changes

- Add `AuthModule`
- Add `AuthController`, enabled behind a feature flag
  - `GET` `/v1/auth/nonce` - returns nonce to be signed in SiWe message
  - `POST` `/v1/auth/verify` - verify signature and return access token
    - Add `VerifyAuthMessageDtoSchema`
- Add `AuthService`
  - `generateNonce` - get nonce to include in SiWe message
  - `verify` - verify signature according to expiration, nonce and validity of signature
- Appropriate test coverage for the above.


src/routes/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+      .verifyMessage(args)
+      // Don't prevent nonce from being deleted
+      .catch(() => false);
+
+    const cacheDir = CacheRouter.getAuthNonceCacheDir(args.message.nonce);
+    const cachedNonce = await this.cacheService.get(cacheDir);
+    const isValidNonce = cachedNonce === args.message.nonce;
+
+    // Delete nonce from cache to prevent replay attacks
+    await this.cacheService.deleteByKey(cacheDir.key);
```
fmrsabino, line 148: This breaks the abstraction that we have in the routes when it comes to accessing the Datasources. This should go through the domain layer instead.

src/routes/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+   * 4. Return an access token if all checks pass.
+   *
+   * @param args.request - Express request object.
+   * @param args.verifyAuthMessageDto - DTO containing the message and signature to verify.
+   *
+   * The following adhere to JWT standard {@link https://datatracker.ietf.org/doc/html/rfc7519}
+   *
+   * @returns accessToken - JWT access token
+   * @returns tokenType - token type ('Bearer') to be used in the `Authorization` header
+   * @returns notBefore - expoch from when token is valid (if applicable, otherwise null)
```
fmrsabino, line 68: `epoch`

src/routes/auth/auth.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ValidationPipe } from '@/validation/pipes/validation.pipe';
+import { AuthService } from '@/routes/auth/auth.service';
+import { VerifyAuthMessageDtoSchema } from '@/routes/auth/entities/schemas/verify-auth-message.dto.schema';
+import { VerifyAuthMessageDto } from '@/routes/auth/entities/schemas/verify-auth-message.dto';
+
+/**
+ * The AuthController is responsible for handling authentication:
+ *
+ * 1. Calling `/v1/auth/nonce` returns a unique nonce to be signed.
+ * 2. The client signs this nonce in a SiWe message, sending it and
```
fmrsabino, line 13: The spec mentions it as `SIWE`. We should keep it that way in my opinion.

src/routes/auth/auth.controller.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(private readonly authService: AuthService) {}
+
+  @Get('nonce')
+  async getNonce(): Promise<{
+    nonce: string;
+  }> {
+    return this.authService.getNonce();
+  }
+
+  @HttpCode(202)
```
fmrsabino, line 31: What is the reason behind returning `202` (and not `200` for example).

src/routes/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+   * @returns expiresIn - time in seconds until the token expires (if applicable, otherwise null)
+   */
+  async verify(args: {
+    request: Request;
```
fmrsabino, line 72: We don't need the `request` here.

src/routes/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    const { message } = args.verifyAuthMessageDto;
+
+    const dateWhenTokenIsValid = message.notBefore
+      ? new Date(message.notBefore)
+      : null;
+    const dateWhenTokenExpires = message.expirationTime
+      ? new Date(message.expirationTime)
+      : null;
```
fmrsabino, line 95: Can't we get the `message` and respective field coercions with Zod? The `message` is an `hexSchema` so how would it have the properties `notBefore` and `expirationTime`?

src/routes/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  CacheService,
+  ICacheService,
+} from '@/datasources/cache/cache.service.interface';
+import { CacheRouter } from '@/datasources/cache/cache.router';
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { SiweMessage } from '@/domain/auth/entities/siwe-message.entity';
+
+@Injectable()
+export class AuthService {
+  static readonly AUTH_TOKEN_TOKEN_TYPE = 'Bearer';
```
fmrsabino, line 16: We can increase encapsulation here and leave this field `private`.

src/config/entities/configuration.ts
```diff
@@ -17,6 +17,9 @@ export default () => ({
   applicationPort: process.env.APPLICATION_PORT || '3000',
   auth: {
     token: process.env.AUTH_TOKEN,
+    nonceTtlSeconds: parseInt(
+      process.env.AUTH_NONCE_TTL_SECONDS ?? `${60 * 60}`,
+    ),
```
fmrsabino, line 22: Is there a reference or recommended minimum value from the spec?

fmrsabino (changes requested):

src/routes/auth/auth.service.ts
iamacook: Moved to `AuthRepository` in 52e1c66649c3a40858d05b1258748cfe621b3b26.

src/routes/auth/auth.service.ts
iamacook: Changed in 52e1c66649c3a40858d05b1258748cfe621b3b26.

src/routes/auth/auth.controller.ts
iamacook: Changed in 52e1c66649c3a40858d05b1258748cfe621b3b26.

src/routes/auth/auth.controller.ts
iamacook: This is a leftover from my testing. Changed to `200` in 52e1c66649c3a40858d05b1258748cfe621b3b26.

src/routes/auth/auth.service.ts
iamacook: Removed in 52e1c66649c3a40858d05b1258748cfe621b3b26.

src/routes/auth/auth.service.ts
iamacook: > Can't we get the `message` and respective field coercions with Zod? The `message` is an `hexSchema` so how would it have the properties `notBefore` and `expirationTime`?

`message` is [an object](https://github.com/safe-global/safe-client-gateway/blob/main/src/domain/auth/entities/siwe-message.entity.ts#L12). It does not coerce `notBefore` or `expirationTime` as the field values would have been signed _as is_ by the user.

We receive the values and then convert them to [the would-have-been-signed message](https://github.com/safe-global/safe-client-gateway/blob/main/src/datasources/auth-api/siwe-api.service.ts#L29) for comparison. If we coerced the values, we would have to change them back to a string when creating said message. Would it be preferable to do so?

Note: I understand the logic flow is quite complex and I intend to documen this.

src/routes/auth/auth.service.ts
iamacook: This value [is needed](https://github.com/safe-global/safe-client-gateway/blob/df2bc24a2eae01e4ac7007af40d2e1ac9d81c027/src/routes/auth/guards/auth.guard.ts#L37) by the `AuthGuard` in the follow up (#1369). Would you prefer I set it as `private` here then back?

src/config/entities/configuration.ts
iamacook: No, this is an arbitrary value. It is effectively the period when a user should sign the message. Do you think we should decrease it?

I considered setting it to the `expirationTime` but that value is not always present and the nonce is cleared if the message is expired.

src/routes/auth/auth.service.ts
```diff
@@ -0,0 +1,89 @@
+import { Inject, Injectable, UnauthorizedException } from '@nestjs/common';
+import { VerifyAuthMessageDto } from '@/routes/auth/entities/schemas/verify-auth-message.dto';
+import { IAuthRepository } from '@/domain/auth/auth.repository.interface';
+import { IJwtService } from '@/datasources/jwt/jwt.service.interface';
```
fmrsabino, line 4: We should also move the access to this interface to be done via the `IAuthRepository` (as it's a Data Source).

fmrsabino (changes requested):

src/routes/auth/auth.service.ts
fmrsabino: I see. I might have actually mixed it up with another schema because, `message` is not an `HexSchema`. Sorry for the confusion!

src/domain/auth/auth.repository.ts
```diff
@@ -1,26 +1,148 @@
...
   }): Promise<boolean> {
-    return this.authApi.verifyMessage(args);
+    const isExpired =
+      !!args.message.expirationTime &&
+      new Date(args.message.expirationTime) < new Date();
+
+    const isValidSignature = await this.authApi
+      .verifyMessage(args)
+      // Don't prevent nonce from being deleted
+      .catch(() => false);
```
hectorgomezv, line 132: Nitpick: wdyt about using a `try-catch-finally` block here? The `nonce` deletion seems to fit in a `finally` clause at it needs to be always executed, and we can also have specific errors for each condition (expired, invalid nonce, invalid signature) that could be managed in a `catch` block.

hectorgomezv (approved):

src/domain/auth/auth.repository.ts
iamacook: Refactored in 0ce020796f2e6da46916af649e7160d62810bf8e.

hectorgomezv (approved):

src/domain/auth/auth.repository.ts
```diff
@@ -1,26 +1,155 @@
...
+
+    try {
+      // Verification is not necessary, message has expired
+      if (isExpired) {
+        return false;
+      }
+
+      const [isValidSignature, cachedNonce] = await Promise.all([
+        this.authApi.verifyMessage(args),
+        await this.cacheService.get(cacheDir),
```
fmrsabino, line 139: We don't need the `await` here right?

src/domain/auth/auth.repository.ts
```diff
@@ -1,26 +1,155 @@
...
+  /**
+   * Verifies that a message is valid according to its expiration date,
+   * signature and nonce.
+   *
+   * @param args.message - SiWe message in object form
+   * @param args.signature - signature from signing the message
+   *
+   * @returns boolean - whether the message is valid
+   */
+  private async isAuthorized(args: {
```
fmrsabino, line 121: nitpick: maybe `isValid` is a better description of what we do here 🤔

Authorization seems more related with resource access. A request can be valid (expiration time, nonce) but not authorised to access a resource.

src/routes/auth/entities/schemas/verify-auth-message.dto.ts
```diff
@@ -0,0 +1,4 @@
+import { VerifyAuthMessageDtoSchema } from '@/routes/auth/entities/schemas/verify-auth-message.dto.schema';
+import { z } from 'zod';
+
+export type VerifyAuthMessageDto = z.infer<typeof VerifyAuthMessageDtoSchema>;
```
fmrsabino, line 4: nitpick: to merge this file with the entity one (less reference jumping around the project).

src/config/entities/configuration.ts
iamacook: As per discussion, this was reduced to five minutes in 8c6934c772570d3a66c079a042e3cf28af7a1f6e.

src/domain/auth/auth.repository.ts
iamacook: We don't. Removed in 719b0f301fa3901b5ec468b7c90c04936b592f8c.

src/domain/auth/auth.repository.ts
iamacook: Changed in 719b0f301fa3901b5ec468b7c90c04936b592f8c.

src/routes/auth/entities/schemas/verify-auth-message.dto.ts
iamacook: Merged in 719b0f301fa3901b5ec468b7c90c04936b592f8c.

fmrsabino (approved):
🚀 👏

PR #1369, opened 2024-04-05
Add `AuthGuard`

iamacook:
## Summary

This adds the`AuthGuard` as part of a test implementation for SiWe authentication, breaking down https://github.com/safe-global/safe-client-gateway/pull/1321 for review. It is used to protect a route by a signed SiWe message.

## Changes

- Add `AuthGuard` with tests.


src/routes/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,74 @@
...
+
+/**
+ * The AuthGuard should be used to protect routes that require authentication.
+ *
+ * It checks for the presence of a valid JWT access token in the request and
+ * verifies its validity before allowing access to the route.
+ *
+ * 1. Check for the presence of an access token in the request.
+ * 2. Verify the token's validity and decode it.
+ * 3. Ensure the decoded token is that of a valie SiweMessage.
```
fmrsabino, line 20: "valid"

src/routes/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,74 @@
+import {
+  CanActivate,
+  ExecutionContext,
+  Inject,
+  Injectable,
+} from '@nestjs/common';
+import { IAuthRepository } from '@/domain/auth/auth.repository.interface';
+import { IJwtService } from '@/datasources/jwt/jwt.service.interface';
```
fmrsabino, line 8: To keep the separation of responsibilities between the layers, this should call a domain handler.

src/routes/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,74 @@
...
+        return null;
+      }
+    })();
+
+    // Token is either not yet valid, expired or malformed
+    if (!payload) {
+      return false;
+    }
+
+    const result = SiweMessageSchema.safeParse(payload);
```
fmrsabino, line 58: Should we validate the schema first? If the schema is not valid from the get go no need to validate the containing properties 🤔

fmrsabino (changes requested):

src/routes/auth/guards/auth.guard.ts
fmrsabino: Disregard the above as the `payload` is only retrieved via the `jwtService` so this is correct 👍

src/domain/auth/auth.repository.interface.ts
```diff
@@ -6,12 +6,17 @@ export const IAuthRepository = Symbol('IAuthRepository');
 export interface IAuthRepository {
   generateNonce(): Promise<{ nonce: string }>;
 
-  verify(args: { message: SiweMessage; signature: `0x${string}` }): Promise<{
+  verifyMessage(args: {
```
iamacook, line 9: Renamed for clarity as `verifyAccessToken` was added here.

src/routes/auth/guards/auth.guard.ts
iamacook: Fixed in b5aed5145112dbbc6648486f44d67c343fb43d7c.

src/routes/auth/guards/auth.guard.ts
iamacook: I added a new `AuthRepository['verifyAccessToken']` method in b5aed5145112dbbc6648486f44d67c343fb43d7c which is now called here instead.

src/routes/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * The AuthGuard should be used to protect routes that require authentication.
+ *
+ * It checks for the presence of a valid JWT access token in the request and
+ * verifies its validity before allowing access to the route.
+ *
+ * 1. Check for the presence of an access token in the request.
+ * 2. Verify the token's validity and decode it.
+ * 3. Ensure the decoded token is that of a valid SiweMessage.
+ * 4. Ensure the token is not expired or not yet valid.
```
hectorgomezv, line 20: `4. Ensure the token is not expired and is already valid.`

(Maybe I'm mixing up things)

iamacook: You're spot on. Updated in 0d1fae29290f4892c659b2006c70c41e727c5623.

hectorgomezv (approved):
🚀

src/routes/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,66 @@
...
+      request,
+      AuthService.AUTH_TOKEN_TOKEN_TYPE,
+    );
+
+    // No token in the request
+    if (!accessToken) {
+      return false;
+    }
+
+    const payload = this.authRepository.verifyAccessToken(accessToken);
```
fmrsabino, line 43: The schema of a JWT payload might not match the one required by SIWE right? In JWT the payload would contain a set of claims which is specific to JWT – https://www.iana.org/assignments/jwt/jwt.xhtml

src/routes/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!result.success) {
+      return false;
+    }
+
+    const { expirationTime, notBefore } = result.data;
+
+    const now = new Date();
+    // Should be covered by JWT verification but we check in case
+    const isExpired = !!expirationTime && new Date(expirationTime) < now;
+    const isValid = notBefore ? new Date(notBefore) < now : true;
```
fmrsabino, line 62: I don't think we should have custom logic to verify a JWT token as it might differ in some aspects. We should delegate this to the JWT library only.

fmrsabino (changes requested):

src/routes/auth/guards/auth.guard.ts
iamacook: As of `0d0c355763a73da4bb910f1fd82f33b90bcef21e`, we are now only storing `signer_address` in the payload.

src/routes/auth/guards/auth.guard.ts
```diff
@@ -0,0 +1,47 @@
...
+  constructor(
+    @Inject(IAuthRepository) private readonly authRepository: IAuthRepository,
+  ) {}
+
+  canActivate(context: ExecutionContext): boolean {
+    const request = context.switchToHttp().getRequest();
+
+    const accessToken = this.authRepository.getAccessToken(
+      request,
+      AuthService.AUTH_TOKEN_TOKEN_TYPE,
```
fmrsabino, line 32: nitpick: out of context for this PR – we should probably move this to the `AuthRepository` as it's used by both the domain and routes.

fmrsabino (approved):
👏 🚀

src/routes/auth/guards/auth.guard.ts
iamacook: Agreed. As we will migrate to using cookies, this will ultimately be removed so I will merge this as is for now.

PR #1372, opened 2024-04-05
Extract Alerts configuration

fmrsabino:
- The Alerts related configuration and environment keys were extracted to the `AlertsApiModule` and `AlertsControllerModule`.
- By separating the properties of the Alerts feature to its own configuration object which is loaded dynamically, we can make some properties set as required in the application **only if the respective feature is loaded**.
- This effectively reduces the number of configuration properties that is needed to set for the application to launch, when features are under development or a set of features is not needed by the hosting party.
- In order to avoid key collision, the feature level configuration now uses namespaces (with `registerAs`).
  * `alerts-api` for keys used in `/datasources/alerts-api`
  * `alerts-route` for keys used in `/routes/alerts` 

```mermaid
classDiagram
    ConfigurationModule ..> rootConfiguration
    AlertsApiConfigurationModule ..> alertsApiConfiguration
    AlertsConfigurationModule ..> alertsConfiguration
    NestConfigService o-- ConfigurationModule
    NestConfigService o-- AlertsApiConfigurationModule
    NestConfigService o-- AlertsConfigurationModule
    IConfigurationService -- NestConfigService
    FeatureA ..|> IConfigurationService
    FeatureB ..|> IConfigurationService
    FeatureC ..|> IConfigurationService

    <<Interface>> IConfigurationService
    <<Service>> NestConfigService
    <<Module>> ConfigurationModule
    <<Module>> AlertsApiConfigurationModule
    <<Module>> AlertsConfigurationModule

    note for AlertsApiConfigurationModule "Only loaded if feature flag is set to true"
    note for AlertsConfigurationModule "Only loaded if feature flag is set to true"

    class IConfigurationService {
    }
    class NestConfigService {
    }
    class ConfigurationModule {
    }
    class rootConfiguration {
        about
        applicationPort
        ...
    }
    class AlertsApiConfigurationModule {
    }
    class alertsApiConfiguration {
        apiKey
        baseUri
        account
        project
    }
    class AlertsConfigurationModule {
    }
    class alertsConfiguration {
        signingKey
    }

```

### DIFF
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -7,13 +7,6 @@ export default (): ReturnType<typeof configuration> => ({
     version: faker.system.semver(),
     buildNumber: faker.string.numeric(),
   },
-  alerts: {
-    baseUri: faker.internet.url({ appendSlash: false }),
-    signingKey: faker.string.nanoid(32),
-    apiKey: faker.string.hexadecimal({ length: 32 }),
-    account: faker.string.sample(),
-    project: faker.string.sample(),
-  },
   applicationPort: faker.internet.port().toString(),
   auth: {
     token: faker.string.hexadecimal({ length: 32 }),
@@ -171,7 +164,7 @@ export default (): ReturnType<typeof configuration> => ({
   express: { jsonLimit: '1mb' },
   features: {
     richFragments: true,
-    email: true,
+    email: false,
     zerionBalancesChainIds: ['137'],
     locking: true,
     relay: true,
--- a/src/datasources/alerts-api/configuration/alerts-api.configuration.module.ts
+++ b/src/datasources/alerts-api/configuration/alerts-api.configuration.module.ts
@@ -19,5 +19,5 @@ export class AlertsApiConfigurationModule {
   }
 }
 
-export const alertsApiConfigurationModule =
+export const ALERTS_API_CONFIGURATION_MODULE =
   AlertsApiConfigurationModule.register(alertsApiConfiguration);
--- a/src/datasources/alerts-api/configuration/alerts-api.configuration.module.ts
+++ b/src/datasources/alerts-api/configuration/alerts-api.configuration.module.ts
@@ -1,0 +1,23 @@
+import { DynamicModule, Module } from '@nestjs/common';
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { ConfigFactory } from '@nestjs/config/dist/interfaces/config-factory.interface';
+import { ConfigModule } from '@nestjs/config';
+import { NestConfigurationService } from '@/config/nest.configuration.service';
+import alertsApiConfiguration from '@/datasources/alerts-api/configuration/alerts-api.configuration';
+
+@Module({})
+export class AlertsApiConfigurationModule {
+  static register(configFactory: ConfigFactory): DynamicModule {
+    return {
+      module: AlertsApiConfigurationModule,
+      imports: [ConfigModule.forFeature(configFactory)],
+      providers: [
+        { provide: IConfigurationService, useClass: NestConfigurationService },
+      ],
+      exports: [IConfigurationService],
+    };
+  }
+}
+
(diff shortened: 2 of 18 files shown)


src/config/entities/__tests__/configuration.ts
```diff
@@ -171,7 +164,7 @@ export default (): ReturnType<typeof configuration> => ({
   express: { jsonLimit: '1mb' },
   features: {
     richFragments: true,
-    email: true,
+    email: false,
```
fmrsabino, line 167: This was set to `false` because otherwise, every test would have to include the now optional Alerts configuration (resulting in more changes).

Since this feature is opt-in, each test should decide if it should include the feature or not.

src/datasources/alerts-api/configuration/alerts-api.configuration.ts
```diff
@@ -0,0 +1,21 @@
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
iamacook, line 1: Nit: can we add a comment why this is necessary?

src/routes/alerts/configuration/alerts.configuration.ts
```diff
@@ -0,0 +1,13 @@
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
```
iamacook, line 1: Nit: as before regarding a comment.

src/datasources/alerts-api/configuration/alerts-api.configuration.module.ts
```diff
@@ -0,0 +1,23 @@
...
+      imports: [ConfigModule.forFeature(configFactory)],
+      providers: [
+        { provide: IConfigurationService, useClass: NestConfigurationService },
+      ],
+      exports: [IConfigurationService],
+    };
+  }
+}
+
+export const alertsApiConfigurationModule =
```
iamacook, line 22: Nit: I was initially confused regarding the naming of this (starting with a lowercase character). What do you think about adjusting to to something like `RegisteredAlertsApiConfigurationModule`?

src/datasources/alerts-api/configuration/alerts-api.configuration.ts
fmrsabino: Not needed. Leftover from local testing.

[3be66bc](https://github.com/safe-global/safe-client-gateway/pull/1372/commits/3be66bcdefeb534bec04fbec96acf7397dca4a5d)

src/datasources/alerts-api/configuration/alerts-api.configuration.module.ts
hectorgomezv: I'd be in favor of using `SCREAMING_SNAKE_CASE` here 👍

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -1618,7 +1630,7 @@ describe('Alerts (Unit)', () => {
         const alert = alertBuilder().build();
         const timestamp = Date.now().toString();
         const signature = fakeTenderlySignature({
-          signingKey,
+          signingKey: faker.string.nanoid(32),
```
hectorgomezv, line 1633: Why is this needed? `signingKey` is already being taken from `src/routes/alerts/configuration/__tests__/alerts.configuration.ts` right?

fmrsabino: It is not because the feature flag for this test is supposed to be false – https://github.com/safe-global/safe-client-gateway/blob/3be66bcdefeb534bec04fbec96acf7397dca4a5d/src/routes/alerts/alerts.controller.spec.ts#L1604

So this test won't have the alerts configuration available.

src/datasources/alerts-api/configuration/alerts-api.configuration.module.ts
iamacook: I also agree.

hectorgomezv (approved):
👏🏻

PR #1373, opened 2024-04-08
Improve SiWe nonce generation

iamacook:
## Summary

This improves the nonce generation for SiWe messages, favouring generation instead of modifying a UUID.

## Changes

- Leverage `crypto.getRandomValues` over `crypto.randomUUID`.

### DIFF
--- a/src/datasources/auth-api/siwe-api.service.ts
+++ b/src/datasources/auth-api/siwe-api.service.ts
@@ -7,6 +7,8 @@ import { verifyMessage } from 'viem';
 
 @Injectable()
 export class SiweApi implements IAuthApi {
+  private readonly NONCE_LENGTH = 8;
+
   constructor(
     @Inject(LoggingService)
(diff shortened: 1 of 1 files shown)


src/datasources/auth-api/siwe-api.service.ts
```diff
@@ -7,6 +7,8 @@ import { verifyMessage } from 'viem';
 
 @Injectable()
 export class SiweApi implements IAuthApi {
+  private readonly NONCE_LENGTH = 8;
```
fmrsabino, line 10: I'd set it as `static`.

fmrsabino (changes requested):

src/datasources/auth-api/siwe-api.service.ts
iamacook: Changed in e607443959055f523163be550fa5670763fc96d7.

fmrsabino (approved):

PR #1382, opened 2024-04-09
[WIP] PoC: consume events from TXS queue

hectorgomezv:
## Summary
This PR adds a `QueuesApiModule`, which includes a `QueuesApiService`. This service takes responsibility for consuming AMQP message queues and exposes a `subscribe` function, which allows the clients of its API to set up a callback function that will be executed on every message reception in a given queue.

```mermaid
sequenceDiagram
    CacheHooksService->>+QueuesRepository: subscribe(topicName, callback)
    QueuesRepository->>+QueuesApiService: subscribe(topicName, callback)
    QueueConsumer->>QueuesApiService: message
    QueuesApiService->>QueuesRepository: message
    QueuesRepository->>CacheHooksService: message
    CacheHooksService->>CacheHooksService: onEvent(message)
```

## Changes
- Adds a development RabbitMQ server instance to the `docker-compose.yml` file.
- Adds a development RabbitMQ server instance to the `ci.yml` file.
- Adds `QueuesApiModule` along with `QueuesApiService`, which allows subscribing functions to new messages received in a given AMQP queue.
- Adds `QueuesRepositoryModule` along with `QueuesRepositiory` in the domain layer.
- `HealthController` has been modified to take into consideration the state of the queue consumer.


src/datasources/queues/queue-consumer.service.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+export const IQueueConsumerService = Symbol('IQueueConsumerService');
+
+export interface IQueueConsumerService {
+  /**
+   * Checks if the configured queue consumer is connected.
+   */
+  isReady(): boolean;
+
+  subscribe(
+    queueName: string,
```
hectorgomezv, line 12: Initially the `queueName` will be the name of the queue where the TXS is publishing the event messages, but this parameter opens the door to subscribing to different queues.

src/datasources/queues/queue-consumer.service.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+    fn: (msg: ConsumeMessage) => Promise<void>,
```
hectorgomezv, line 13: The function `fn` would be called every time a message arrives to the queue named `queueName`.

src/domain/health/health.repository.ts
```diff
@@ -1,26 +1,33 @@
...
       await this.cacheService.ping();
-      return HealthEntity.READY;
     } catch (error) {
       this.loggingService.warn('Cache service connection is not established');
       return HealthEntity.NOT_READY;
     }
+
+    return this.queueConsumerService.isReady()
+      ? HealthEntity.READY
+      : HealthEntity.NOT_READY;
```
hectorgomezv, line 31: By having this, `QueueConsumerService` is now a dependency while checking the service health. i.e.: `/health/ready` endpoint now depends on the status of the queue consumer. I thought the events consumption could be seen as a hard dependency for the CGW (when the FF is active), so I opted for including it here.

src/routes/cache-hooks/cache-hooks.controller.ts
```diff
@@ -14,26 +14,35 @@ import { Event } from '@/routes/cache-hooks/entities/event.entity';
...
     @Body(new ValidationPipe(WebHookSchema)) event: Event,
   ): Promise<void> {
-    this.service.onEvent(event).catch((error) => {
-      this.loggingService.error(error);
-    });
+    if (!this.isEventHooksEndpointEnabled) {
+      this.service.onEvent(event).catch((error) => {
+        this.loggingService.error(error);
+      });
+    }
```
hectorgomezv, line 46: Here I opted for just disabling the reception of events via HTTP _silently_, instead of unmounting the controller, to avoid disrupting the Events Service, which theoretically would still send HTTP requests when the queue consumption is activated.

src/routes/cache-hooks/cache-hooks.service.ts
```diff
@@ -28,7 +33,30 @@ export class CacheHooksService {
...
+          const event: Event = WebHookSchema.parse(
+            JSON.parse(Buffer.from(content.data).toString()),
+          );
+          await this.onEvent(event);
+        } catch (err) {
+          this.loggingService.error(err);
+        }
+      },
+    );
+  }
```
hectorgomezv, line 59: Here I made `CacheHooksService` implement `OnModuleInit` and take responsibility for subscribing to the queue. I'd prefer to have this kind of functionality on the `domain` layer, but it needs the `onEvent` function. So I think there are two options for a future improvement:
- Moving the `onEvent` function (along with all events entities) to the `domain` (by creating an `EventsRepository`).
- Subscribe from the `domain` (by creating an `EventsRepository`) and call this `onEvent` function in the `routes` layer.

I'm leaning towards the first option. Thoughts?

src/__tests__/retry.ts
```diff
@@ -0,0 +1,33 @@
...
+        throw error;
+      }
+    }
+
+    await milliseconds(delayMs);
+    attempt++;
+    return execute();
+  };
+  return execute();
+}
```
hectorgomezv: Tests-only utility which takes a function as a parameter and executes it until the execution succeeds. The main point of having something like this is checking state modified by asynchronous processes that are not mocked and are out of Jest scope, like a RabbitMQ queue messages consumption. 

It would be simpler to use a `setTimeout` and check the state was modified after a given amount of time, but setting that time lapse could be problematic and could lead to flaky tests IMO. If the lapse is too short, it might fail sometimes, and if it's too long, it would impact the tests execution time, as we probably need to use this multiple times in the tests.

src/__tests__/retry.ts
```diff
@@ -0,0 +1,33 @@
...
```
hectorgomezv, line 33: Test-related utility which takes a function as a parameter, and executes this function repeatedly until success or until the attempts threshold is not reached. This can help to verify some piece of state that is not being mocked and is outside Jest's scope, like the asynchronous consumption of a RabbitMQ queue.

A simpler solution could be just using `setTimeout` but setting up the time lapse could be problematic, as a short one could lead to flaky tests under some conditions, and a longer one would impact the tests execution time.

src/datasources/queues/queue-consumer.service.ts
```diff
@@ -0,0 +1,41 @@
...
+    fn: (msg: ConsumeMessage) => Promise<void>,
+  ): Promise<void> {
+    if (this.isEventsQueueEnabled) {
+      await this.consumer.channel.consume(queueName, async (msg) => {
+        await fn(msg);
+        await this.consumer.channel.ack(msg);
+      });
+      this.loggingService.info(`Subscribed to queue: ${queueName}`);
+    }
+  }
```
hectorgomezv, line 40: Another option besides checking `this.isEventsQueueEnabled` is each function would be injecting a _mock_ `QueueConsumerService` in `QueueConsumerModule`, depending on whether the FF is active or not. I opted for not taking that approach as it would need to make `QueueConsumerModule` dynamic, and given that other parts of the application also depend on it (`HealthRepositoryModule`, `CacheHooksModule`), it would make the overall setup more convoluted in my opinion, also considering this FF would be transitory.

src/domain/health/health.repository.ts
```diff
@@ -6,21 +6,27 @@ import {
...
       await this.cacheService.ping();
-      return HealthEntity.READY;
     } catch (error) {
       this.loggingService.warn('Cache service connection is not established');
       return HealthEntity.NOT_READY;
     }
+
+    return this.queuesRepository.isReady()
+      ? HealthEntity.READY
+      : HealthEntity.NOT_READY;
```
hectorgomezv, line 30: By having this, `QueueConsumerService` is now a dependency while checking the service health. i.e.: `/health/ready` endpoint now depends on the status of the queue consumer. I thought the events consumption could be seen as a hard dependency for the CGW (when the FF is active), so I opted for including it here.

hectorgomezv:
Closing this PR as it's a PoC that will be implemented through several smaller and production-ready PRs.

PR #1384, opened 2024-04-09
Add error messages for custom validation schemas

iamacook:
## Summary

This adds error messages for custom validation schemas.

## Changes

- Reused schemas now have the following error messages:
  - `NumericStringSchema` - 'Invalid base-10 numeric string'
  - `HexSchema` - 'Invalid "0x" notated hex string'
  - `AddressSchema` - 'Invalid address'
- A few isolated fields that have custom validation have had error messages added.
- The above changes have been propegated across test coverage.

### DIFF
--- a/src/validation/entities/schemas/hex.schema.ts
+++ b/src/validation/entities/schemas/hex.schema.ts
@@ -1,4 +1,6 @@
 import { z } from 'zod';
 import { isHex } from 'viem';
 
-export const HexSchema = z.string().refine(isHex);
+export const HexSchema = z.string().refine(isHex, {
+  message: 'Invalid hex string',
+});
--- a/src/validation/entities/schemas/hex.schema.ts
+++ b/src/validation/entities/schemas/hex.schema.ts
@@ -2,5 +2,5 @@ import { z } from 'zod';
 import { isHex } from 'viem';
 
 export const HexSchema = z.string().refine(isHex, {
-  message: 'Invalid hex string',
+  message: 'Invalid "0x" notated hex string',
 });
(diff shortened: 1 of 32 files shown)


hectorgomezv (approved):
👏🏻

src/validation/entities/schemas/hex.schema.ts
```diff
@@ -1,4 +1,6 @@
 import { z } from 'zod';
 import { isHex } from 'viem';
 
-export const HexSchema = z.string().refine(isHex);
+export const HexSchema = z.string().refine(isHex, {
+  message: 'Invalid hex string',
+});
```
fmrsabino, line 6: The `isHex` requires a string to start with `0x` right? If so, I would add this to the error message as a string like `fff` can be considered an hex string for the clients.

src/validation/entities/schemas/numeric-string.schema.ts
```diff
@@ -4,4 +4,6 @@ function isNumeric(value: unknown): boolean {
   return typeof value === 'string' && !isNaN(Number(value));
 }
 
-export const NumericStringSchema = z.string().refine(isNumeric);
+export const NumericStringSchema = z.string().refine(isNumeric, {
+  message: 'Invalid numeric string',
+});
```
fmrsabino, line 9: Do we consider only 10-base numbers? Or in any base?

src/validation/entities/schemas/hex.schema.ts
iamacook: Good point. I updated the message to be `'Invalid "0x" notated hex string"` in 30596f4249a57df3c13f8aedd61c025d817dd81b.

fmrsabino (approved):
🚀

PR #1388, opened 2024-04-10
Split auth domain

iamacook:
## Summary

This splits the auth domain into SiWe and JWT relative repositories.

## Changes

- Remove `AuthDomainModule`
- Split `AuthRepository` into `SiweRepository` and `JwtRepository`, moving associated entities, etc. into each domain.
- Rename `AuthApi` to `SiweApi`


src/routes/auth/guards/auth.guard.ts
```diff
@@ -38,10 +40,26 @@ export class AuthGuard implements CanActivate {
...
     try {
-      this.authRepository.verifyAccessToken(accessToken);
+      this.jwtRepository.verifyToken(accessToken, JwtAccessTokenPayloadSchema);
       return true;
     } catch {
       return false;
     }
   }
+
+  private getAccessToken(request: Request, tokenType: string): string | null {
```
iamacook, line 50: I moved this here as it is isolated to this guard. As we will likely pivot to cookies, this is temporary.

src/domain/jwt/jwt.repository.ts
```diff
@@ -0,0 +1,30 @@
...
+      notBefore?: number;
+    },
+  ): string {
+    return this.jwtService.sign(payload, options);
+  }
+
+  verifyToken<T extends z.ZodTypeAny>(
+    accessToken: string,
+    schema: T,
+  ): z.ZodType<z.infer<T>> {
```
fmrsabino, line 26: We are leaking Zod types in a domain interface. Can we use more concrete types here? (as in types that we know that we will be able to handle)

src/domain/siwe/siwe.repository.ts
```diff
@@ -0,0 +1,82 @@
...
+   * Generates a unique nonce and stores it in cache for later verification.
+   *
+   * @returns nonce - unique string to be signed
+   */
+  async generateNonce(): Promise<{ nonce: string }> {
+    const nonce = this.authApi.generateNonce();
+
+    // Store nonce for reference to verify/prevent replay attacks
+    const cacheDir = CacheRouter.getAuthNonceCacheDir(nonce);
+    await this.cacheService.set(cacheDir, nonce, this.nonceTtlInSeconds);
```
fmrsabino, line 38: This should be part of the functionality of the `IAuthApi`. The `SiweRepository` shouldn't need to know that this is stored in a cache.

fmrsabino (changes requested):

src/domain/jwt/jwt.repository.ts
iamacook: I changed this in e0ee482bfd60773bb9423aac0cb5252b157da667 to explicitly use the `JwtAccessTokenSchema` so we can expect the concrete type.

src/domain/siwe/siwe.repository.ts
iamacook: I moved all nonce interfacing to the `IAuthApi` in e0ee482bfd60773bb9423aac0cb5252b157da667.

src/domain/interfaces/auth-api.interface.ts
```diff
@@ -9,4 +9,10 @@ export interface IAuthApi {
     message: SiweMessage;
     signature: `0x${string}`;
   }): Promise<boolean>;
+
+  cacheNonce(nonce: string): Promise<void>;
```
fmrsabino, line 13: We should avoid mentioning the cache backend (if possible). We can then change the backend without changing the interface. 

```suggestion
  storeNonce(nonce: string): Promise<void>;
```

src/domain/interfaces/auth-api.interface.ts
```diff
@@ -9,4 +9,10 @@ export interface IAuthApi {
...
+
+  getCachedNonce(nonce: string): Promise<string | undefined>;
```
fmrsabino, line 15: Similar to the above ☝️ 

```suggestion
  getNonce(nonce: string): Promise<string | undefined>;
```

src/domain/interfaces/auth-api.interface.ts
```diff
@@ -9,4 +9,10 @@ export interface IAuthApi {
...
+
+  clearCachedNonce(nonce: string): Promise<void>;
```
fmrsabino, line 17: Similar to the above ☝️ 

```suggestion
  clearNonce(nonce: string): Promise<void>;
```

src/domain/siwe/siwe.repository.interface.ts
```diff
@@ -0,0 +1,24 @@
...
+export const ISiweRepository = Symbol('ISiweRepository');
+
+export interface ISiweRepository {
+  generateNonce(): Promise<{ nonce: string }>;
+
+  isValidMessage(args: VerifyAuthMessageDto): Promise<boolean>;
+}
+
+@Module({
+  imports: [AuthApiModule],
```
fmrsabino, line 15: Maybe we should rename the `AuthApiModule` as it's now SIWE related only 🤔

fmrsabino (changes requested):

src/domain/interfaces/auth-api.interface.ts
iamacook: Changed in b15fa98e39c07d2987f9f21eafce4ea6e318ec21.

src/domain/interfaces/auth-api.interface.ts
iamacook: Changed in b15fa98e39c07d2987f9f21eafce4ea6e318ec21.

src/domain/interfaces/auth-api.interface.ts
iamacook: Changed in b15fa98e39c07d2987f9f21eafce4ea6e318ec21.

src/domain/siwe/siwe.repository.interface.ts
iamacook: I renamed it (and all other SiWe-associated datasource elements) to use `Siwe*` in place of `Auth*` in b15fa98e39c07d2987f9f21eafce4ea6e318ec21.

fmrsabino (approved):
👏

PR #1391, opened 2024-04-10
Fix token used for fee label

fmrsabino:
The sell token should be used for the fee label (instead of the buy token).


fmrsabino:
Had to fix some conflicts @iamacook 🙇

iamacook (approved):

PR #1395, opened 2024-04-11
Add surplusLabel to Order

fmrsabino:
- Adds the surplus property to a Fulfilled order.
- The surplus is computed with the absolute difference between the initial token amount and the executed amount of the order.

### DIFF
--- a/src/routes/transactions/mappers/common/swap-order.mapper.ts
+++ b/src/routes/transactions/mappers/common/swap-order.mapper.ts
@@ -208,7 +208,7 @@ export class SwapOrderMapper {
         args.sellToken,
         args.buyToken,
       ),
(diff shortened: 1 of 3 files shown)


src/routes/transactions/mappers/common/swap-order.mapper.ts
```diff
@@ -197,6 +208,7 @@ export class SwapOrderMapper {
         args.sellToken,
         args.buyToken,
       ),
+      surplusLabel: surplusLabel,
```
iamacook, line 211: Nit:

```suggestion
      surplusLabel,
```

iamacook (approved):

iamacook (approved):

PR #1401, opened 2024-04-12
Move relay count CRUD to datasource

iamacook:
## Summary

This moves the relay count get/set methods to the `IRelayApi` datasource in accordance with the separation of layers throughout the project.

## Changes

- Add `IRelayApi['getRelayCount' | 'setRelayCount']`
- Use above methods in `RelayRepository
- Update tests accordingly

### DIFF
--- a/src/datasources/relay-api/gelato-api.service.ts
+++ b/src/datasources/relay-api/gelato-api.service.ts
@@ -11,6 +11,8 @@ import {
   CacheService,
   ICacheService,
 } from '@/datasources/cache/cache.service.interface';
+import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
+import { getAddress } from 'viem';
 
 @Injectable()
 export class GelatoApi implements IRelayApi {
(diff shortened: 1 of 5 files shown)


src/datasources/relay-api/gelato-api.service.ts
```diff
@@ -63,4 +73,37 @@ export class GelatoApi implements IRelayApi {
...
+    address: string;
+    count: number;
+  }): Promise<void> {
+    const cacheDir = this.getRelayCacheKey(args);
+    await this.cacheService.set(
+      cacheDir,
+      args.count.toString(),
+      this.ttlSeconds,
+    );
+  }
```
hectorgomezv, line 97: This is out of the scope of this PR, but maybe at some point it would be nice to refactor `RelayRepository.relay` to use an atomic operation for getting and incrementing the counter value, instead of separate operations. (We can leverage the already existing `ICacheService.increment` function, which server this purpose).

Although I think it's not critical, as the case where the counter can be retrieved from the cache but at the same time the increased value cannot be persisted shouldn't happen, it would be the best practice IMO.

hectorgomezv (approved):

src/datasources/relay-api/gelato-api.service.ts
iamacook: I agree, and that it should be done separately. I will make a note to open a follow up PR with the refactor.

PR #1404, opened 2024-04-15
Add support for `presignaturePending` orders

fmrsabino:
- `presignaturePending` orders are now decoded.
- This is a new requirement as orders in this state were not previously supported.


usame-algan (approved):
Looks good 👍

iamacook (approved):

PR #1405, opened 2024-04-15
Extract JWT configuration

fmrsabino:
- Extracts the JWT configuration out of the root configuration object into its own configuration object.
- This means that if validation is added to the respective properties, the variables won't be required to be set if the module is not enabled.

### DIFF
--- a/src/datasources/jwt/configuration/__tests__/jwt.configuration.ts
+++ b/src/datasources/jwt/configuration/__tests__/jwt.configuration.ts
@@ -1,7 +1,11 @@
 import { registerAs } from '@nestjs/config';
 import { faker } from '@faker-js/faker';
+import { JwtConfiguration } from '@/datasources/jwt/configuration/jwt.configuration';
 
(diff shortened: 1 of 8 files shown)


src/datasources/jwt/configuration/__tests__/jwt.configuration.ts
```diff
@@ -0,0 +1,7 @@
+import { registerAs } from '@nestjs/config';
+import { faker } from '@faker-js/faker';
+
+export default registerAs('jwt', () => ({
```
iamacook, line 4: Nit: we could create an `interface` from the "actual" `registerAs` to ensure that the signature or the `ReturnType` is the same.

fmrsabino: [fdb2d39](https://github.com/safe-global/safe-client-gateway/pull/1405/commits/fdb2d39d8e46451b3ef48f8c3c712e05fc1b6db2)

PR #1413, opened 2024-04-16
Add expiration time limit

iamacook:
## Summary

This adds a 15 minute limit for the validity period of JWT tokens. (It is currently hardcoded, but will eventually be injected.)

## Changes

- Change the `SiweMessageSchema` to a function that accepts a validity period.
- Pass a value to `getSiweMessageSchema` in `VerifyAuthMessageDto`.
- Add test case, as well as update existing tests.


src/routes/auth/entities/schemas/__tests__/verify-auth-message.dto.schema.spec.ts
```diff
@@ -2,10 +2,18 @@ import { siweMessageBuilder } from '@/domain/siwe/entities/__tests__/siwe-messag
 import { VerifyAuthMessageDtoSchema } from '@/routes/auth/entities/verify-auth-message.dto.entity';
 import { faker } from '@faker-js/faker';
 
+const MAX_VALIDITY_PERIOD_IN_MS = 15 * 60 * 1_000; // 15 minutes
+
 describe('VerifyAuthMessageDto', () => {
```
hectorgomezv, line 7: Maybe I'm not getting it 100% right here, but I'm thinking whether we should call `jest.useFakeTimers` also here (as in `AuthController` tests) or not, to make sure the edge cases (`faker` returning threshold dates) doesn't make the tests flaky.

iamacook: That's true. I've adjusted it to include fake timers in 8fdb401f73bbe4551722d0f5df324013723bf632.

hectorgomezv (approved):

PR #1415, opened 2024-04-16
Update `AuthGuard` to only grant access to Safe owners

iamacook:
⚠️ Depends on #1405

## Summary

This updates the `AuthGuard` to only authorise access owners of the requested Safe.

Providing a signer has authenticated themselves with SiWe, the session JWT token is now referenced against the owners of desired Safe and (if they are an owner) granted access.

Note: this does not remove the existing `OnlySafeOwnerGuard`. It will eventually replace all authentication guards used for email routes in a follow up PR.

## Changes

- Add validated `chain_id` claim to authentication JWT token payload.
- Modify `AuthGuard` to verify authenticated session is of a Safe owner.
- Add/update associated test coverage.

### DIFF
--- a/src/routes/auth/auth.service.ts
+++ b/src/routes/auth/auth.service.ts
@@ -30,9 +30,10 @@ export class AuthService {
       throw new UnauthorizedException();
     }
 
-    const { address, notBefore, expirationTime } = args.message;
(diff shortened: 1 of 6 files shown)


iamacook:
I'm converting this to a draft as we agreed the owner check should be isolated in the `OnlySafeOwnerGuard`.

PR #1426, opened 2024-04-18
Return non formatted Order data

fmrsabino:
## Summary

The service was previously following a render-ready approach when it came to Swap Orders (i.e.: pre-formatted labels that should be rendered as is).

While this approach is a valid one, the current setup that we have on the Wallet doesn't allow for this type of localisation to be driven by this web service.

Some examples include:
- Position of the currency symbol
- Use of comma vs period for thousands or decimals
- Difference in representation of negative amounts
- LTR vs RTL

Given that in the current state, the frontend clients are expected to have more information regarding the user localisation, this responsibility is then shifted from the CGW to the clients.

This has a couple implications on the API:

## Changes

- Removes the following labels (they now need to be computed on the client side):
  * feeLabel
  * executionPriceLabel
  * surplusLabel
  * filledPercentage
- Extends the `TokenInfo` with:
  * Token name
  * trusted status
- Token amounts are now returned without decimals (correct decimal formatting should be set by the clients).

```javascript
{
  "type": "SwapOrder",
  "uid": "string",
  "status": "presignaturePending | open | fulfilled | cancelled | expired | unknown",
  "kind": "buy | sell",
  "class": "market | limit | liquidity | unknown"
  "validUntil": "number",
  "sellAmount": "string",
  "buyAmount": "string",
  "executedSellAmount": "string",
  "executedBuyAmount": "string",
  "sellToken": "TokenInfo", // see below
  "buyToken": "TokenInfo",  // see below
  "explorerUrl": "URL",
  "executedSurplusFee": "string | null"
}

"TokenInfo": {
  "address": "0x{string}",
  "decimals": "number",
  "logoUri": "string | null",
  "name": "string",
  "symbol": "string",
  "trusted": "boolean"
}
```

### DIFF
--- a/src/routes/transactions/entities/swap-order-info.entity.ts
+++ b/src/routes/transactions/entities/swap-order-info.entity.ts
@@ -66,7 +66,7 @@ export class SwapOrderTransactionInfo extends TransactionInfo {
   kind: 'buy' | 'sell';
 
   @ApiProperty({ description: 'The timestamp when the order expires' })
-  validTo: number;
+  validUntil: number;
 
   @ApiProperty({
     description: 'The sell token raw amount (no decimals)',
@@ -111,7 +111,7 @@ export class SwapOrderTransactionInfo extends TransactionInfo {
     uid: string;
     status: OrderStatus;
     kind: 'buy' | 'sell';
-    validTo: number;
+    validUntil: number;
     sellAmount: string;
     buyAmount: string;
     executedSellAmount: string;
@@ -125,7 +125,7 @@ export class SwapOrderTransactionInfo extends TransactionInfo {
     this.uid = args.uid;
     this.status = args.status;
     this.kind = args.kind;
-    this.validTo = args.validTo;
+    this.validUntil = args.validUntil;
     this.sellAmount = args.sellAmount;
     this.buyAmount = args.buyAmount;
     this.executedSellAmount = args.executedSellAmount;
--- a/src/routes/transactions/entities/swap-order-info.entity.ts
+++ b/src/routes/transactions/entities/swap-order-info.entity.ts
@@ -7,7 +7,7 @@ import {
   ApiProperty,
   ApiPropertyOptional,
 } from '@nestjs/swagger';
-import { OrderStatus } from '@/domain/swaps/entities/order.entity';
+import { OrderClass, OrderStatus } from '@/domain/swaps/entities/order.entity';
 
 export class TokenInfo {
   @ApiProperty({ description: 'The token address' })
@@ -65,6 +65,11 @@ export class SwapOrderTransactionInfo extends TransactionInfo {
   @ApiProperty({ enum: ['buy', 'sell'] })
   kind: 'buy' | 'sell';
 
+  @ApiProperty({
+    enum: OrderClass,
+  })
+  class: OrderClass;
+
   @ApiProperty({ description: 'The timestamp when the order expires' })
   validUntil: number;
 
@@ -111,6 +116,7 @@ export class SwapOrderTransactionInfo extends TransactionInfo {
     uid: string;
     status: OrderStatus;
     kind: 'buy' | 'sell';
+    class: OrderClass;
     validUntil: number;
     sellAmount: string;
     buyAmount: string;
@@ -125,6 +131,7 @@ export class SwapOrderTransactionInfo extends TransactionInfo {
     this.uid = args.uid;
     this.status = args.status;
     this.kind = args.kind;
+    this.class = args.class;
     this.validUntil = args.validUntil;
     this.sellAmount = args.sellAmount;
     this.buyAmount = args.buyAmount;
--- a/src/routes/transactions/entities/swap-order-info.entity.ts
+++ b/src/routes/transactions/entities/swap-order-info.entity.ts
@@ -7,50 +7,86 @@ import {
   ApiProperty,
   ApiPropertyOptional,
 } from '@nestjs/swagger';
+import { OrderStatus } from '@/domain/swaps/entities/order.entity';
 
 export class TokenInfo {
+  @ApiProperty({ description: 'The token address' })
+  address: `0x${string}`;
+
+  @ApiProperty({ description: 'The token decimals' })
+  decimals: number;
+
   @ApiPropertyOptional({
     type: String,
     nullable: true,
-    description: 'The token logo',
+    description: 'The logo URI for the token',
   })
-  logo: string | null;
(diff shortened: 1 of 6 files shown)


compojoom (approved):
i like the new approach! Thanks for implementing it!

src/routes/transactions/entities/swap-order-info.entity.ts
```diff
@@ -7,168 +7,132 @@ import {
...
 
-  constructor(args: { logo: string | null; symbol: string; amount: string }) {
-    this.logo = args.logo;
+  constructor(args: {
+    address: `0x${string}`;
+    decimals: number;
+    logoUri: string | null;
+    name: string;
+    symbol: string;
+    trusted: boolean;
```
iamacook, line 41: What do you think about aligning these to be more in line with the `Erc20Transfer` entity, e.g.

```suggestion
    tokenAddress: `0x${string}`;
    decimals: number;
    logoUri: string | null;
    tokenName: string;
    tokenSymbol: string;
    trusted: boolean;
```

src/routes/transactions/entities/swap-order-info.entity.ts
```diff
@@ -7,168 +7,132 @@ import {
...
-  @ApiProperty({ description: 'The sell token of the order' })
-  sellToken: TokenInfo;
-
-  @ApiProperty({ description: 'The buy token of the order' })
-  buyToken: TokenInfo;
+  kind: 'buy' | 'sell';
 
   @ApiProperty({ description: 'The timestamp when the order expires' })
-  expiresTimestamp: number;
+  validTo: number;
```
iamacook, line 69: Nit: I'd tweak this to `validUntil` or match that of the comment, e.g. `expiresAt`.

src/routes/transactions/entities/swap-order-info.entity.ts
```diff
@@ -7,168 +7,132 @@ import {
...
-    this.feeLabel = args.feeLabel;
-    this.executionPriceLabel = args.executionPriceLabel;
-    this.surplusLabel = args.surplusLabel;
-  }
-}
+  @ApiProperty({ description: 'The sell token of the order' })
+  sellToken: TokenInfo;
+
+  @ApiProperty({ description: 'The buy token of the order' })
+  buyToken: TokenInfo;
```
iamacook, line 95: What do you think about putting these in nested structures? We could reuse the types, e.g.

```ts
type OrderInfo = {
  amount: string;
  executedAmount: string;
  token: TokenInfo
}

type SwapOrderTransactionInfo = {
  buy: OrderInfo;
  sell: OrderInfo;
  ...
}
```

iamacook (commented):
Everything looks good go to me, I just have a few of structural suggestions. Let me know what you think, otherwise we can approve it!

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: @compojoom @usame-algan any opinion on this? I've tried to match it closer to what the CoW Swap API returns. Do you prefer this alternative (involves more nesting: Root -> OrderInfo -> TokenInfo)

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: I've renamed it to be closer to the `Order` object. I do prefer `validUntil` though.

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: Since the object is of a `Token` I think it unnecessary to prepend the `token` to its properties. If the frontend is already using an entity to represent a token I think we can evaluate reusing it though. @compojoom @usame-algan what's your take on this one?

src/routes/transactions/entities/swap-order-info.entity.ts
fmrsabino: Adjust in [af164ec](https://github.com/safe-global/safe-client-gateway/pull/1426/commits/af164ec085dc60df6de5c9620f3024f04b4a9adb)

src/routes/transactions/entities/swap-order-info.entity.ts
usame-algan: I think both would work. Since there are always exactly two tokens in a swap, not nesting it is fine as it won't need to be extended in the future. I would be in favour of leaving as is since its less work.

src/routes/transactions/entities/swap-order-info.entity.ts
compojoom: I think the same as @fmrsabino - since the object is already called Token, it is a bit weird if props have a Token prefix.

compojoom:
@fmrsabino I just realised that we still don't have the class of the order, so we won't know if we are dealing with limit order. Could you please add this.

compojoom (changes requested):
@fmrsabino I just realised that we still don't have the class of the order, so we won't know if we are dealing with limit order. Could you please add this.

fmrsabino:
> @fmrsabino I just realised that we still don't have the class of the order, so we won't know if we are dealing with limit order. Could you please add this.

Was adding it here https://github.com/safe-global/safe-client-gateway/pull/1427 to reduce the scope of the PR. But since it's just a field and the entire API is being reworked I will merge it to this branch 👍

fmrsabino:
@compojoom merged in [7c2b510](https://github.com/safe-global/safe-client-gateway/pull/1426/commits/7c2b510f47861eab2c67334b86acc8a0c6719032)

iamacook (approved):

compojoom (approved):

compojoom:
I was wondering about "unknown" 
export enum OrderClass {
  Market = 'market',
  Limit = 'limit',
  Liquidity = 'liquidity',
  Unknown = 'unknown',
}

I saw "unknown" somewhere else as well. What is it? Why do we have it? As far I understand the orderBook API - there can't be an "unknown" class? Why do we have this?

fmrsabino:
> I was wondering about "unknown" export enum OrderClass { Market = 'market', Limit = 'limit', Liquidity = 'liquidity', Unknown = 'unknown', }
> 
> I saw "unknown" somewhere else as well. What is it? Why do we have it? As far I understand the orderBook API - there can't be an "unknown" class? Why do we have this?

It is used as a fallback value by the CGW in case there are new types introduced on the Order Book API. It allows us to use type unions / enums on our side even if 3rd party services introduce new types.

PR #1428, opened 2024-04-19
Event poisoning filter demo

iamacook:
⚠️ This is demonstrative implementation that serves as a reference for discussion/further development.

## Motivation

We are seeing an increase in address poisoning, whereby a fake transfer event is emitted **after a legitimate one**. These "poisoned events" aim to trick users into copying poisoned addresses and transfering funds to them. The attack tends to occur **immediately after** legitimate, **outgoing transfers**.

## Summary

This alters the logic of the `TransactionHistoryMapper` to facilitate and implement filtering of addres poisoning.

### Alteration of logic flow

The last transaction of the previous page (“previous transaction”) was (and still is) fetched into order to calculate the date label for the final transaction of the history page.

Before this PR, the date extracted from this domain transaction using a helper for each type of transaction, e.g. multisig, creation, module, etc. as the timestamp differs by each.

Now, the “previous transaction” is immediately mapped in order to use it as filter reference for the final transaction of the page, on top of getting the date label from it. This removes the necessity for a date label-relevant helper function that essentially duplicated the timestamp mapping of each transaction type, making the source of truth the mapped transactions themselves. It also simplifies a lot of associated logic regarding the history mapping.

Note: if this implementation is deemed viable, the logic refactor and filtering will be separated into different PRs.

### Filtering

This implements an **opinionated** filter that removes suspected poison events from the transaction history according to the following criteria:

- They occur relative to outgoing transfers. (Incoming transfers _do_ occur but they require legitimate funds.)
- They happen immediately after legitimate transfers. (Not always the case.)
- They are of the same value.
- They are to >=4 character matching "vanity address" recipients.

The following should be considered:

- Comparison against transfers within `multiSend` transactions. (If comparing neighbours, this won't work due to "distance".)
- Parameterisation of the above filter (although a feature flag may be sufficient.)
- Flagging the transactions as `trusted` vs. hiding them (and whether we rely on the `trusted` path param.)

## Changes

- Map all domain transactions, including the "previous" one, before filtering and adding date labels.
- Add/adjust tests accordingly.


src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -21,18 +22,10 @@ import { MultisigTransactionMapper } from '@/routes/transactions/mappers/multisi
...
-    | MultisigTransaction
-    | ModuleTransaction
-    | EthereumTransaction
-    | CreationTransaction
-  )[];
-}
-
 @Injectable()
 export class TransactionsHistoryMapper {
+  private readonly VANITY_ADDRESS_CHARS = 4;
```
hectorgomezv, line 27: Nitpick: for the final implementation maybe we can extract the number of chars considered as prefix/suffix to the configuration?

hectorgomezv (approved):
Looks really good! 🚀

iamacook:
For future reference: I'll split this into two PRs instead of merging it - one with the refactor and a second with the address poisoning logic.

iamacook:
I'm going to close this as it has been split into #1505 and #1524.

PR #1434, opened 2024-04-22
Add SwapOrderHelper

fmrsabino:
## Summary

The functionality that was present in the `SwapOrderMapper` can be shared with other components (outside of the mapping module for a collection of transactions).

This functionality was then extracted to a `SwapOrderHelper` which contains the following functionality:
- `findSwapOrder` – no changes. Given a transaction, extract the Swap Order from it if available (supports MultiSend).
- `getOrder` – given an Order ID, attempts to retrieve its data alongside the buy and sell token information.
- `getOrderExplorerUrl` – no changes. Given an `Order` object, builds the Swap Order explorer URL.

This helper is available via the `SwapOrderHelperModule`.

## Changes

- Adds a `SwapOrderHelper`.
- Moves functionality from `SwapOrderMapper` and `MultisigTransactionInfoMapper` to the `SwapOrderHelper`.
- Extracts the `TokenInfo` schema in order to be used by other components.

### DIFF
--- a/src/routes/transactions/helpers/swap-order.helper.ts
+++ b/src/routes/transactions/helpers/swap-order.helper.ts
@@ -1,0 +1,132 @@
+import { Inject, Injectable, Module } from '@nestjs/common';
+import { MultiSendDecoder } from '@/domain/contracts/decoders/multi-send-decoder.helper';
+import { SetPreSignatureDecoder } from '@/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper';
+import {
+  ITokenRepository,
+  TokenRepositoryModule,
+} from '@/domain/tokens/token.repository.interface';
+import {
+  ISwapsRepository,
+  SwapsRepository,
+} from '@/domain/swaps/swaps.repository';
+import { Token } from '@/domain/tokens/entities/token.entity';
+import { Order } from '@/domain/swaps/entities/order.entity';
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { SwapsRepositoryModule } from '@/domain/swaps/swaps-repository.module';
+
+@Injectable()
+export class SwapOrderHelper {
+  private readonly swapsExplorerBaseUri: string =
+    this.configurationService.getOrThrow('swaps.explorerBaseUri');
+
+  constructor(
+    private readonly multiSendDecoder: MultiSendDecoder,
+    private readonly setPreSignatureDecoder: SetPreSignatureDecoder,
+    @Inject(ITokenRepository)
+    private readonly tokenRepository: ITokenRepository,
+    @Inject(ISwapsRepository) private readonly swapsRepository: SwapsRepository,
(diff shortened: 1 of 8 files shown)


src/routes/transactions/helpers/swap-order.helper.ts
```diff
@@ -0,0 +1,132 @@
...
+   * The returned object includes:
+   * - `order`: An object representing the order.
+   * - `sellToken`: The Token object with a mandatory `decimals` property
+   * - `buyToken`: Similar to `sellToken`, for the token being purchased in the order.
+   *
+   * @throws {Error} Throws an error if the order `kind` is 'unknown'.
+   * @throws {Error} Throws an error if either the sellToken or buyToken object has null decimals.
+   */
+  async getOrder(args: { chainId: string; orderUid: `0x${string}` }): Promise<{
+    order: Omit<Order, 'kind'> & { kind: Exclude<Order['kind'], 'unknown'> };
```
iamacook, line 76:
```suggestion
    order: Order & { kind: Exclude<Order['kind'], 'unknown'> };
```

src/routes/transactions/helpers/swap-order.helper.ts
```diff
@@ -0,0 +1,132 @@
...
+    ]);
+
+    if (buyToken.decimals === null || sellToken.decimals === null) {
+      throw new Error('Invalid token decimals');
+    }
+
+    return {
+      order: { ...order, kind: order.kind },
+      buyToken: { ...buyToken, decimals: buyToken.decimals },
+      sellToken: { ...sellToken, decimals: sellToken.decimals },
```
iamacook, line 105: It's a shame that these need be remapped like this. (I also played around and it seems like the only way to appease TypeScript though.)

iamacook (approved):

src/routes/transactions/helpers/swap-order.helper.ts
fmrsabino: Yeah I've tried declaring it in multiple ways but it seems that we need to redeclare those properties again in order to have the type correctly narrowed 🤔

src/routes/transactions/helpers/swap-order.helper.ts
fmrsabino: Interesting, at some point this was not working as I expected but maybe that was when I didn't have `{ ...order, kind: order.kind }`.

The original idea was to remove the `kind` completely from the order and add a new one without `unknown` (thus the two separate steps here) – i.e. I didn't want all the `Order.kind` +  `Order.kind except 'unknown'`.

iamacook: Both approaches are correct but I noticed that the tokens use the "other" method so I suggested this for consistency. Afaiu, it overwrites the former. Prior to my suggestion, it is more explicit. I would suggest we use one or the other.

iamacook (approved):

PR #1439, opened 2024-04-22
Rename Order.class to Order.orderClass

fmrsabino:
## Summary

Renames `Order.class` to `Order.orderClass` to avoid using JS reserved keywords as property names – https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words.

This was supposed to be included in becda8f8b489a7b0058b64c5178d2ea396b581e8 but instead, the `status` property was the one renamed (it should remain as `status`).

## Changes

- Renames `class` to `orderClass`
- Renames `orderStatus` to `status`


usame-algan (approved):
Thanks 🙌

iamacook (approved):

PR #1440, opened 2024-04-22
Change AMQP prefetch configuration

hectorgomezv:
## Summary
The AMQP prefetch value was configured to `0` (therefore unlimited) assuming the explicit `ack` emitted by the consumers would automatically round-robin the delivery of the messages between them. 

But this does not consider the scenario where the service pods _cold-start_ and multiple messages are in the queue. In this scenario, we want to still round-robin between the pods, instead of letting one pod fill up with all the messages.

(Optional) Reference documentation: https://www.rabbitmq.com/docs/consumer-prefetch

Note: the value `100` was not chosen following any special logic, any non-zero value would serve the purpose of the PR. There are no strict rules regarding this value, it can be adjusted via service configuration if needed.

## Changes
- Changes the default value for `amqp.prefetch` configuration value from `0` to `100`.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -118,8 +118,7 @@
 # (default=fanout)
 #AMQP_EXCHANGE_MODE=
 # The message prefetch amount (max non-acknowledged messages being processed)
-# The default is 0, so no throttling.
-# (default=0)
+# (default=100)
 #AMQP_PREFETCH=
 
 # Authorization token to use privileged endpoints.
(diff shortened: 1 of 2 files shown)


iamacook (changes requested):
We need to also update the `.env.sample` file to mention `100` as well.

hectorgomezv:
> We need to also update the .env.sample file to mention 100 as well.

Thanks! Changes in [7604394](https://github.com/safe-global/safe-client-gateway/pull/1440/commits/760439478c5c35f4e03a010e08fcaca3df31aaee)

iamacook (approved):

PR #1450, opened 2024-04-24
Include not listed Safe Apps on transaction mappings

hectorgomezv:
## Summary
Depends on https://github.com/safe-global/safe-config-service/pull/1111

This PR integrates the `onlyListed` Config Service query param when calling `GET /safe-apps/` endpoint. 

## Changes
- Adds `onlyListed` query parameter.
- Sets it to `false` on transaction mapping.
- Sets it to `true` when retrieving Safe Apps directly from the CGW API.


fmrsabino (approved):
🚀

PR #1452, opened 2024-04-24
Forward the Order receiver

fmrsabino:
## Summary

A CoW Swap Order contains an optional `receiver` which represents the target address that should receive the proceeds from the trade. This field is required by the clients of this service API so it should be forwarded.

## Changes

- Forwards the Order `receiver` in the Confirmation View and Transaction decoding

### DIFF
--- a/src/routes/transactions/entities/swaps/swap-order-info.entity.ts
+++ b/src/routes/transactions/entities/swaps/swap-order-info.entity.ts
@@ -22,6 +22,7 @@ export interface OrderInfo {
   executedBuyAmount: string;
   explorerUrl: URL;
   executedSurplusFee: string | null;
+  receiver: string | null;
   fullAppData: Record<string, unknown> | null;
 }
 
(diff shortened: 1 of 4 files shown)


src/routes/transactions/entities/swaps/swap-order-info.entity.ts
```diff
@@ -91,6 +92,13 @@ export class SwapOrderTransactionInfo
   })
   executedSurplusFee: string | null;
 
+  @ApiPropertyOptional({
+    type: String,
+    nullable: true,
+    description: 'The (optional) address to receive the proceeds of the trade',
+  })
```
fmrsabino, line 99: The Open API docs regarding the Order should be mostly the same which points to sharing these decorators. However that would require creating an abstract class with this information (which we can consider).

So far, we strived for composing the different components over inheriting them (in the Swaps feature).

iamacook (approved):

PR #1456, opened 2024-04-25
Rename `JwtRepository` to `AuthRepository`

iamacook:
## Summary

This renames the domain `JwtRepository` (now `AuthRepository`) as the logic is customised to our auth implementation.

## Changes

- Rename `JwtRepository` and interface to `AuthRepository`
- Rename `JwtAccessTokenPayload` to `AuthPayload`, as well as associated builder
- Rename `JwtAccessTokenPayloadSchema` to `AuthPayload`
- Amend associated tests with appropriate naming


iamacook:
For future reference: this was originally moved to the routes layer after discussion with @hectorgomezv. However, after review we decided it would be better to remain on the domain layer, just renamed.

There is still a question of whether the JWT datasource is inherently a datasource and (if not) all should be located on the route layer. However, we agreed that this can be iterated on as it is currently easier to follow spread across layers.

hectorgomezv (approved):

PR #1457, opened 2024-04-25
Add settlementContract to Order

fmrsabino:


fmrsabino:
Closing this one since it's part of an undocumented API from CoW which might be removed (so we can't rely on this field).

PR #1458, opened 2024-04-25
Allow Swap App restriction

fmrsabino:
## Summary

Adds the ability to restrict the decoding functionality to specific Swap applications. Swap applications can be identified with Metadata that is added to each order by the respective apps. The property which contains this metadata is `fullAppData` – https://explorer.cow.fi/appdata?tab=encode

## Changes

- Two configuration properties were added:
  * `SWAPS_RESTRICT_APPS` – this configuration controls if the App restriction should be enabled or disabled.
  * `SWAPS_ALLOWED_APPS` – this configuration contains the comma separated collection of Swap apps which are allowed to be decoded. This configuration only has an effect if `SWAPS_RESTRICT_APPS` is set to `true`.

### DIFF
--- a/src/routes/transactions/mappers/common/swap-order.mapper.ts
+++ b/src/routes/transactions/mappers/common/swap-order.mapper.ts
@@ -1,4 +1,4 @@
-import { Injectable, Module } from '@nestjs/common';
+import { Inject, Injectable, Module } from '@nestjs/common';
 import { SetPreSignatureDecoder } from '@/domain/swaps/contracts/decoders/set-pre-signature-decoder.helper';
 import { SwapOrderTransactionInfo } from '@/routes/transactions/entities/swaps/swap-order-info.entity';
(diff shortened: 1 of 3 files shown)


iamacook (approved):

src/routes/transactions/mappers/common/swap-order.mapper.ts
```diff
@@ -29,6 +36,13 @@ export class SwapOrderMapper {
       orderUid,
     });
 
+    if (
+      this.restrictApps &&
+      !this.allowedApps.has(order.fullAppData['appCode'])
```
iamacook, line 41: I think it would be good to add a comment with a link to the CoW docs here. What do you think?

fmrsabino: Maybe in the Config itself what do you think?

iamacook (approved):

PR #1459, opened 2024-04-25
Populate `Request['auth']` in AuthGuard

iamacook:
## Summary

In order to keep authentication and authorisation separate, this implements the basis for validation the currently authenticated user.

We are verifying the JWT token of the user and hence their authentication the `AuthGuard`. After authentication has succeeded, we can assume that the parsed payload is inherently valid. It is therefore now assigned to `Request['accessToken']` which can be accessed for comparison, e.g. if the `chain_id` matches that which we expect.

For easier access, a new `AuthPayloadDecorator` returns the `Request['accessToken']` values.

Note: test coverage will be included when migrating currently authenticated routes to use the new JWT-based system.

## Changes

- Assign the verified `AuthPayload` to `Request['accessToken']` in `AuthGuard`
- Add `AuthPayloadDecorator`


src/routes/auth/decorators/auth-payload.decorator.ts
```diff
@@ -0,0 +1,13 @@
+import { createParamDecorator, ExecutionContext } from '@nestjs/common';
+import { Request } from 'express';
+import { AuthPayload } from '@/domain/auth/entities/auth-payload.entity';
+
+/**
+ * Route decorator that extracts the {@link AuthPayload}
+ */
+export const AuthPayloadDecorator = createParamDecorator(
```
hectorgomezv, line 8: Nit: is it maybe a bit redundant to use `Decorator` in the decorator name? What about simply `@Auth`? 

(On [the NestJS documentation](https://docs.nestjs.com/custom-decorators) the examples have simpler names also, e.g.: `export const User = createParamDecorator(...)`)

hectorgomezv (approved):

src/routes/auth/decorators/auth-payload.decorator.ts
iamacook: I used `RouteUrlDecorator` as a reference but I agree in hindsight. I've renamed it in 3d66d93afa832080bb9869adca7f1da9442676d5.

hectorgomezv (approved):

src/routes/auth/decorators/auth.decorator.ts
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Route decorator that extracts the {@link AuthPayload}
+ */
+export const Auth = createParamDecorator(
+  (_: unknown, ctx: ExecutionContext): AuthPayload | undefined => {
+    const request: Request = ctx.switchToHttp().getRequest();
+    return request.auth;
+  },
+);
```
fmrsabino, line 13: I'd create a test for this. This can be done via a controller made specifically for testing.

src/routes/auth/guards/auth.guard.ts
```diff
@@ -6,12 +6,22 @@ import {
...
 import { AuthController } from '@/routes/auth/auth.controller';
+import { AuthPayload } from '@/domain/auth/entities/auth-payload.entity';
+import { Request } from 'express';
+
+declare module 'express' {
+  // Inject AuthPayload to express.Request
+  interface Request {
+    auth?: AuthPayload | undefined;
+  }
+}
```
fmrsabino, line 17: What is the advantage of doing it this way? Wouldn't it be simpler to access it via `request['auth']`?

fmrsabino (changes requested):

src/routes/auth/guards/auth.guard.ts
```diff
@@ -36,7 +46,7 @@ export class AuthGuard implements CanActivate {
     }
 
     try {
-      this.authRepository.verifyToken(accessToken);
+      request.auth = this.authRepository.verifyToken(accessToken);
```
fmrsabino, line 49: `auth` seems a bit generic. I think `accessToken` is more descriptive and more representative of what the property holds.

fmrsabino (changes requested):

src/routes/auth/decorators/auth.decorator.ts
iamacook: Added a test in f047089a1153d5234508a9c1b4d45e85255cd483.

src/routes/auth/guards/auth.guard.ts
iamacook: This adds the property to the global types so it can be accessed in other files.

src/routes/auth/guards/auth.guard.ts
iamacook: Renamed in f047089a1153d5234508a9c1b4d45e85255cd483.

src/routes/auth/guards/auth.guard.ts
fmrsabino: But wouldn't that be done via the decorator?

iamacook: Would you prefer the type declaration be moved to the decorator? It is not possible to access the property unless it exists on the type:

> Element implicitly has an 'any' type because expression of type '"accessToken"' can't be used to index type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'.
  Property 'accessToken' does not exist on type 'Request<ParamsDictionary, any, any, ParsedQs, Record<string, any>>'.ts(7053)

image

image

fmrsabino: I thought you could access the property without redeclaring the module. In that case I think it should be ok 👍

fmrsabino (approved):
🚀

PR #1460, opened 2024-04-25
Authorise email retrieval relative to JWT authentication

iamacook:
## Summary

Prior to this, retrieving accounts required a user to sign a message with every request. This migrates the suboptimal authentication/authorisation to the new JWT/SIWE-based approach.

## Changes

- Replace `EmailRetrievalGuard` and `TimestampGuard` guards with `AuthGuard`
- Delete `EmailRetrievalGuard`
- Checksum `signer` of request
- Ensure authorisation payload exists and the `chain_id` and `signer_address` matches that of the request
- Add/update test coverage accordingly.


src/routes/email/email.controller.get-email.spec.ts
```diff
@@ -101,115 +105,219 @@ describe('Email controller get email tests', () => {
...
-        message: 'Forbidden resource',
-        error: 'Forbidden',
-        statusCode: 403,
-      });
+      .set('Cookie', [`access_token=${accessToken}`])
+      .expect(403);
+  });
 
-    expect(accountDataSource.getAccount).toHaveBeenCalledTimes(0);
+  // Note: this could be removed as we checksum the :signer but for robustness we should keep it
```
iamacook, line 215: This is relative to the type changes.

src/routes/email/email.controller.ts
```diff
@@ -37,20 +41,19 @@ export class EmailController {
...
-    EmailRetrievalGuard,
-    TimestampGuard(5 * 60 * 1000), // 5 minutes
-  )
+  @UseGuards(AuthGuard)
   @UseFilters(AccountDoesNotExistExceptionFilter)
   async getEmail(
     @Param('chainId') chainId: string,
     @Param('safeAddress') safeAddress: string,
-    @Param('signer') signer: string,
+    @Param('signer', new ValidationPipe(AddressSchema)) signer: `0x${string}`,
```
iamacook, line 49: This is where the address is checksummed.

src/routes/email/email.service.ts
```diff
@@ -76,8 +78,16 @@ export class EmailService {
   async getEmail(args: {
     chainId: string;
     safeAddress: string;
-    signer: string;
+    signer: `0x${string}`;
+    authPayload: AuthPayload | undefined;
   }): Promise<Email> {
+    if (
+      !args.authPayload ||
+      args.authPayload.signer_address !== args.signer ||
```
iamacook, line 86: This is the comparison responsible for the aforementioned checksumming requirements.

src/domain/account/account.repository.ts
```diff
@@ -58,10 +59,18 @@ export class AccountRepository implements IAccountRepository {
...
   }): Promise<Account> {
     const safeAddress = getAddress(args.safeAddress);
     const signer = getAddress(args.signer);
+    if (
+      !args.authPayload ||
+      args.authPayload.signer_address !== signer ||
+      args.authPayload.chain_id !== args.chainId
+    ) {
+      throw new UnauthorizedException();
+    }
```
hectorgomezv, line 73: Nit: just as a side note, not needed in this PR at all: I suspect we will need to repeat this logic in other auth-protected functions. So it might make sense to extract this kind of checks to a common place (utils file, domain auth service, etc.). We could create functions there like `checkChainAndSigner(authPayload: AuthPayload): void` and call them from several places.

But as I said, it's just an idea for future features.

hectorgomezv: Also, by having this logic isolated in a new class, we could move all the 401/403 status code tests to that class' tests (as it's done in this PR `email.controller.get-email.spec.ts` changes). 

I mean, if not, we would need to do multiple repeating tests for each function that we protect via authentication. I know the code currently has many similar tests per controller, but adding 5-6 test cases per each function we want to protect via authentication (which likely will increase in the future) would scale significantly the number of overall tests.

hectorgomezv (approved):
👏🏻

src/domain/account/account.repository.ts
iamacook: I created a new `AuthorizationRepository` in a3dde8a179328b6eb275b6e2e37e182568f2e006. It has an `assertChainAndSigner` method according to your suggestion. However, I think we should keep the `403`/`401` tests as they are somewhat of an E2E implementation and (as it's authorisation-based) I think it's more robust. What do you think?

hectorgomezv: Yeah, I'm fine with keeping the tests for both `AuthorizationRepository` and the controller.

Thanks for integrating the suggestion in the PR! 😄

hectorgomezv (approved):

src/config/entities/__tests__/configuration.ts
```diff
@@ -183,7 +183,7 @@ export default (): ReturnType<typeof configuration> => ({
     relay: true,
     swapsDecoding: true,
     historyDebugLogs: false,
-    auth: false,
+    auth: true,
```
fmrsabino, line 186: We aligned that this was supposed to be opt in order to prevent changes across all the tests. Also changes to those tests seem out of scope for the Email Retrieval.

fmrsabino (changes requested):

src/config/entities/__tests__/configuration.ts
iamacook: This was a leftover from issues I encountered with tests. I've changed it back to `false` and only opted into `auth` where necessary. It's reduced the footprint of configuration overrides as well.

src/domain/account/account.domain.module.ts
```diff
@@ -14,6 +16,10 @@ import { SubscriptionRepository } from '@/domain/subscriptions/subscription.repo
       provide: ISubscriptionRepository,
       useClass: SubscriptionRepository,
     },
+    {
+      provide: IAuthorizationRepository,
+      useClass: AuthorizationRepository,
```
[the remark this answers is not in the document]
iamacook: Changed in b47fd9ca1af7cc1ade1778ae324a39dd58fc9fb3.

src/domain/auth/authorization.repository.ts
```diff
@@ -0,0 +1,22 @@
...
+export class AuthorizationRepository implements IAuthorizationRepository {
+  assertChainAndSigner(args: {
+    chainId: string;
+    signerAddress: `0x${string}`;
+    authPayload: AuthPayload | undefined;
+  }): void {
+    if (
+      !args.authPayload ||
+      args.authPayload.chain_id !== args.chainId ||
+      // Ennsure a mixture of (non-)checksummed addresses match
```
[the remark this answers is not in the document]
iamacook: Changed in b47fd9ca1af7cc1ade1778ae324a39dd58fc9fb3.

src/domain/account/account.repository.ts
```diff
@@ -58,10 +62,18 @@ export class AccountRepository implements IAccountRepository {
   async getAccount(args: {
     chainId: string;
     safeAddress: string;
-    signer: string;
+    signer: `0x${string}`;
+    authPayload: AuthPayload | undefined;
```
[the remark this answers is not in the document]
iamacook: I missed this one! Added in b47fd9ca1af7cc1ade1778ae324a39dd58fc9fb3.

src/domain/auth/auth.repository.interface.ts
```diff
@@ -18,6 +18,16 @@ export interface IAuthRepository {
...
+
+  isChain(args: {
+    chainId: string;
+    authPayload: AuthPayload | undefined;
+  }): boolean;
+
+  isSigner(args: {
+    signerAddress: `0x${string}`;
+    authPayload: AuthPayload | undefined;
+  }): boolean;
```
fmrsabino, line 30: Reviewing it once more and comparing it to the functionality provided by `IAuthRepository`, maybe this makes sense to be part of a class which implements `AuthPayload` wdyt? As in, we are directly comparing external data _against_ the state of `authPayload`.

```typescript
export class AuthPayload implements z.infer<typeof AuthPayloadSchema> {
  readonly chain_id: string;
  readonly signer_address: `0x${string}`;

  constructor(chain_id: string, signer_address: `0x${string}`) {
    this.chain_id = chain_id;
    this.signer_address = signer_address;
  }

  isForChain(chainId: string): boolean {
    return this.chain_id === chainId;
  }

  isForSigner(signerAddress: `0x${string}`): boolean {
    return this.signer_address.toLowerCase() === signerAddress.toLowerCase();
  }
}
```

iamacook: Great suggestion. I made the relevant adjustments in f3a0bee5201c5af0b21818233a35984c753fea62. Lmk what you think.

iamacook:
I'm going to mark this as a draft and abstract the `AuthPayload` entity into a separate PR to reduce the scope of this.

src/routes/recovery/recovery.controller.spec.ts
```diff
@@ -58,6 +63,8 @@ describe('Recovery (Unit)', () => {
     const moduleFixture: TestingModule = await Test.createTestingModule({
       imports: [AppModule.register(testConfiguration)],
     })
+      .overrideModule(JWT_CONFIGURATION_MODULE)
```
iamacook, line 66: As this opts in to `features.email`, the override is needed as`AuthGuard` is now on the `EmailController`.

src/routes/subscriptions/subscription.controller.spec.ts
```diff
@@ -49,6 +54,8 @@ describe('Subscription Controller tests', () => {
     const moduleFixture: TestingModule = await Test.createTestingModule({
       imports: [AppModule.register(testConfiguration), EmailControllerModule],
     })
+      .overrideModule(JWT_CONFIGURATION_MODULE)
```
iamacook, line 57: As this opts in to `features.email`, the override is needed as`AuthGuard` is now on the `EmailController`.

src/routes/alerts/alerts.controller.spec.ts
```diff
@@ -112,6 +117,8 @@ describe('Alerts (Unit)', () => {
       const moduleFixture: TestingModule = await Test.createTestingModule({
         imports: [AppModule.register(testConfiguration)],
       })
+        .overrideModule(JWT_CONFIGURATION_MODULE)
```
iamacook, line 120: As this opts in to `features.email`, the override is needed as`AuthGuard` is now on the `EmailController`.

fmrsabino (approved):
👏 🚀

fmrsabino (approved):

PR #1477, opened 2024-04-29
Map `TransactionInfo` according to `Transfer['type']`

iamacook:
## Summary

We were mapping the types of `Transfer`s based on properties in the entity. However, a `type` decriminator exists. Migrating to this was marked as a TODO and this adjusts the code to instead use that.

## Changes

- Remove and migrate from using `isERC20Transfer` to instead check for `type === ERC20_TRANSFER`
- Remove and migrate from using `isERC721Transfer` to instead check for `type === ERC721_TRANSFER`
- Remove and migrate from using `isNativeTokenTransfer` to instead check for `type === ETHER_TRANSFER`

### DIFF
--- a/src/domain/safe/entities/transfer.entity.ts
+++ b/src/domain/safe/entities/transfer.entity.ts
@@ -14,51 +14,8 @@ export const TransferSchema = z.discriminatedUnion('type', [
 
 export const TransferPageSchema = buildPageSchema(TransferSchema);
 
-const hasTokenAddress = (
-  transfer: Transfer,
-): transfer is Transfer & {
-  tokenAddress: (ERC20Transfer | ERC721Transfer)['tokenAddress'];
-} => {
-  return 'tokenAddress' in transfer && transfer.tokenAddress != null;
-};
-
-const hasValue = (
-  transfer: Transfer,
-): transfer is Transfer & {
-  value: (ERC20Transfer | NativeTokenTransfer)['value'];
-} => {
-  return 'value' in transfer && transfer.value != null;
-};
-
-const hasTokenId = (
-  transfer: Transfer,
-): transfer is Transfer & {
-  tokenId: ERC721Transfer['tokenId'];
-} => {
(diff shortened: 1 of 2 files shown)


src/routes/transactions/mappers/transfers/transfer-info.mapper.ts
```diff
@@ -56,7 +51,7 @@ export class TransferInfoMapper {
     chainId: string,
     domainTransfer: DomainTransfer,
   ): Promise<Transfer> {
-    if (isERC20Transfer(domainTransfer)) {
+    if (domainTransfer.type === 'ERC20_TRANSFER') {
```
iamacook, line 54: This passes our tests and I also tested it against live data.

src/routes/transactions/mappers/transfers/transfer-info.mapper.ts
```diff
@@ -84,7 +79,7 @@ export class TransferInfoMapper {
         token?.logoUri,
         token?.trusted,
       );
-    } else if (isNativeTokenTransfer(domainTransfer)) {
+    } else if (domainTransfer.type === 'ETHER_TRANSFER') {
```
iamacook, line 82: These changes passe our tests and I also tested them against live data.

src/domain/safe/entities/transfer.entity.ts
```diff
@@ -14,51 +14,8 @@ export const TransferSchema = z.discriminatedUnion('type', [
...
-} => {
-  return 'tokenId' in transfer && transfer.tokenId != null;
-};
-
 export type ERC20Transfer = z.infer<typeof Erc20TransferSchema>;
 
-// TODO: Just check `type` is `"ERC20_TRANSFER"`
-export function isERC20Transfer(transfer: Transfer): transfer is ERC20Transfer {
-  return hasTokenAddress(transfer) && hasValue(transfer);
-}
```
iamacook, line 46: I removed the helpers as they are only used in one file.

fmrsabino (approved):

PR #1478, opened 2024-04-29
Instantiate an `AuthPayload` from request

iamacook:
## Summary

Authorisation will require comparison of the chain or signer address to the session present in the request. This creates an `AuthGuard` entity that has helpers for achieving this. It is instantiated by the `Auth` decorator according to the session present in the request

## Changes

- Rename `AuthPayload` to `AuthPayloadDto`
- Create `AuthPayload` entity
- Instantiate `AuthPayload` in `Auth` decorator
- Add/adjust associated test coverage


fmrsabino (approved):
👏 🚀

PR #1491, opened 2024-04-30
Add native currency support to SwapOrderHelper

fmrsabino:
- `SwapOrderHelper` now supports swaps with the respective chain native tokens.
- The Native Token is represented with the address `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE` (https://docs.cow.fi/cow-protocol/reference/sdks/cow-sdk/modules#buy_eth_address)
- If we fail to retrieve the native currency data (from the Safe Config Service), the decoding fails, and it attempts to decode it as a Custom transaction. The main reason for this is that Native Currencies might have different decimals which are important for showing the correct number of tokens being transferred.

### DIFF
--- a/src/routes/transactions/helpers/swap-order.helper.ts
+++ b/src/routes/transactions/helpers/swap-order.helper.ts
@@ -9,13 +9,22 @@ import {
   ISwapsRepository,
   SwapsRepository,
 } from '@/domain/swaps/swaps.repository';
-import { Token } from '@/domain/tokens/entities/token.entity';
+import { Token, TokenType } from '@/domain/tokens/entities/token.entity';
 import { Order } from '@/domain/swaps/entities/order.entity';
 import { IConfigurationService } from '@/config/configuration.service.interface';
 import { SwapsRepositoryModule } from '@/domain/swaps/swaps-repository.module';
+import {
+  ChainsRepositoryModule,
+  IChainsRepository,
+} from '@/domain/chains/chains.repository.interface';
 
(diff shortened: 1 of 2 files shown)


src/routes/transactions/helpers/swap-order.helper.ts
```diff
@@ -138,6 +149,46 @@ export class SwapOrderHelper {
...
+   * of either the native currency or the specified token.
+   * @throws {Error} Throws an error if the token data cannot be retrieved.
+   * @private
+   * @async
+   */
+  private async getToken(args: {
+    chainId: string;
+    address: `0x${string}`;
+  }): Promise<Token> {
+    if (args.address === SwapOrderHelper.NATIVE_CURRENCY_ADDRESS) {
```
iamacook, line 172: Are we sure that `args.address` with _always_ be checksummed? It would be a bit safer to either checksum `args.address` or compare them in lowercase. What do you think?

hectorgomezv (approved):

src/routes/transactions/helpers/swap-order.helper.ts
fmrsabino: Auto-merge merged the commit before I was able to tackle this. I will open a follow up PR.

fmrsabino: Tackled it here: https://github.com/safe-global/safe-client-gateway/pull/1493

PR #1492, opened 2024-04-30
Use LTS Node.js version

### LINKED ISSUE #1490: Use `lts` Node.js version
### Description

The Safe Client Gateway _de facto_ uses Node.js LTS version (its semver reference is being updated over time). 

It was decided by the maintainers to explicitly set the Node.js version to `alpine-lts` as the Docker base image, and also make the CI use `lts` explicitly when running service tests.

(Reference discussion: https://github.com/safe-global/safe-client-gateway/pull/1486#pullrequestreview-2030624928)

### Requirements

- Update Docker base images.
- Update CI runner setup.

hectorgomezv:
## Summary
Closes #1490 

This PR aims to change the Node.js version to the LTS one (long-term support).

This is a no-change as this `lts` tag references the same version that the service was using before this PR (`v20.12.2`)

## Changes
- Changes the Node.js version reference to `LTS`.


iamacook (approved):

Dockerfile
```diff
@@ -1,7 +1,7 @@
 #
 # BUILD CONTAINER
 #
-FROM node:20.12.2 as base
+FROM node:lts as base
```
fmrsabino, line 4: I think we should still need to pin it to a specific LTS version. So here you would pin it to `node:lts-iron`. 

`iron` is the codename for the current `v20.12` version.

Same would apply elsewhere.

hectorgomezv: Good point, maybe it's better to pin the LTS reference to a specific version. With this, we would still manually update the version every time a new LTS version is released, but I agree it's a bit more safer (even considering we have e2e tests).

I've changed it in [213ca9e](https://github.com/safe-global/safe-client-gateway/pull/1492/commits/213ca9e2db4cce0567866a3d0987d103e4d19d23)

fmrsabino: The issue with not pinning the version/release is that the pipeline might start failing on unrelated commits thus blocking merges and releases.

We shouldn't block the release cycle of the service because of a new LTS release as they might be incompatible in some ways. By explicitly changing the LTS version, we can tackle the issues that might arise with that change on the respective PR/commit.

hectorgomezv: Makes perfect sense 👍

fmrsabino (approved):
👏 🚀

PR #1495, opened 2024-04-30
Use TestQueuesApiModule in controller tests

hectorgomezv:
## Summary
This PR changes the `AppModule` created on controller tests: the module used for running these tests should be `TestQueuesApiModule` instead of `QueuesApiModule`. Controller tests shouldn’t need any running RabbitMQ instance and shouldn’t use the runtime RabbitMQ connection logic.

## Changes
- Override `QueuesApiModule` with `TestQueuesApiModule` on controller tests.


fmrsabino (approved):
🚀

PR #1500, opened 2024-05-02
Test PostgresDatabaseModule

fmrsabino:
- Add tests to validate that the connection provided by the `PostgresDatabaseModule` is valid (requires `db-test` to be available).


hectorgomezv (approved):
👏🏻

PR #1501, opened 2024-05-02
Implement GET /v2/chain/:chainId/delegates 

hectorgomezv:
## Summary
This PR adds the endpoint: 

```
/v2/chains/:chainId/delegates
```

This endpoint expects the same payload and returns the same data structure as the `v1` one, but calls the `v2` version of the delegates endpoint of the Safe Transaction Service. A `v2` version of the `delegates` endpoints is needed due to incompatibilities between the signatures. 

## Changes
- Adds `/v2/chains/:chainId/delegates` endpoint.


hectorgomezv:
I opted for creating new classes (controller, service, repository, etc.) for `v2`. Even if the `GET` endpoint does not contain breaking changes and the `v1` logic was reused, `DELETE` endpoints will require new logic and it seems cleaner to have that separated and just rename things after the migration is done.

src/app.module.ts
```diff
@@ -71,6 +72,7 @@ export class AppModule implements NestModule {
         ContractsModule,
         DataDecodedModule,
         DelegatesModule,
```
iamacook, line 74: Nit: we should add a TODO comment to delete the legacy route, and remove `V2` accordingly.

iamacook (approved):

src/app.module.ts
hectorgomezv: I completely agree. Added in [5e84006](https://github.com/safe-global/safe-client-gateway/pull/1501/commits/5e840060faeb683548aa6d7fb706ea76a048844e), thanks!

iamacook (approved):

PR #1504, opened 2024-05-03
Implement DELETE /v2/chain/:chainId/delegates

hectorgomezv:
Depends on #1503


src/routes/delegates/v2/delegates.v2.service.ts
```diff
@@ -59,4 +63,56 @@ export class DelegatesV2Service {
...
+
+    if (!response.results.length) {
+      this.loggingService.warn(
+        `Delegator for delegate ${delegate} and Safe ${safeAddress} not found`,
+      );
+      throw new NotFoundException();
+    }
+
+    return response.results[0].delegator;
+  }
```
iamacook, line 117: Are you trying to create one function for both the deletion of delegates and Safe delegates?

I think it would be simpler to have a function for each. Individual schemas/types should also result in no need for unnecessary `throw` conditions. What do you think?

hectorgomezv: Yes, I'm trying to unify both endpoints, as the deletion of "delegates" and "safe delegates" is essentially the same action in the end, the only difference is the client sending different parameters. Having one endpoint would be more in line with the Transaction Service API.

Do you think maintaining the two endpoints is more desirable? The clients will need to make changes either way, as the signatures needs to be calculated in a different manner.

iamacook: I assumed that we would continue to have both, apologies. I think aligning more with the Transaction Service makes sense, considering the clients will need to make respective adjustments as well.

iamacook (approved):

PR #1505, opened 2024-05-06
Map historical transactions before grouping them

iamacook:
## Summary

This simplifies the mapping of historical transactions in preparation for tackling address poisoning, breaking down #1428.

We currently group domain transactions "as is". This means that we have [duplicate logic](https://github.com/safe-global/safe-client-gateway/commit/991c552823c8f2fe3f460064ec97787266900e59#diff-bfd101886211fde88eb856ba9dbd3e719b1b6a389cf1c603840e6b3e6cf84be0L108-L126) for extracting the timestamp of domain transactions (of varying types) that is otherwise done in their respective mappers, e.g. [for multisig transactions](https://github.com/safe-global/safe-client-gateway/blob/address-poisoning-demo/src/routes/transactions/mappers/multisig-transactions/multisig-transaction.mapper.ts#L45).

This maps domain transactions before grouping them, as well as simplifying the grouping logic. (This will allow for simpler filtering of address poisoning attempts as all transactions will be the same type.)

## Changes

- Change `mapGroupTransactions` to only map singular transaction (`mapTransaction`)
- Map previous transaction (for date label calculation) if it exists
- Map domain transactions before grouping them
- Simplify grouping logic
- Update tests accordingly (the mock data has always been wrong)

### DIFF
--- a/src/routes/transactions/mappers/transactions-history.mapper.ts
+++ b/src/routes/transactions/mappers/transactions-history.mapper.ts
@@ -44,42 +44,23 @@ export class TransactionsHistoryMapper {
     if (transactionsDomain.length == 0) {
       return [];
     }
-    const previousTransaction = await (async (): Promise<
-      TransactionItem | undefined
-    > => {
-      const prevDomainTransaction = this.getPreviousItem(
-        offset,
-        transactionsDomain,
-      );
-      if (!prevDomainTransaction) {
-        return;
-      }
-      // We map in order to filter last list item against it
-      const mappedPreviousTransaction = await this.mapTransaction(
-        prevDomainTransaction,
-        chainId,
-        safe,
-        onlyTrusted,
-      );
-      // Remove first transaction that was requested to get previous day timestamp
-      transactionsDomain = transactionsDomain.slice(1);
+    // Must be retrieved before mapping others as we remove it from transactionsDomain
+    const previousTransaction = await this.getPreviousTransaction({
+      offset,
+      transactionsDomain,
+      chainId,
+      safe,
+      onlyTrusted,
+    });
 
-      return Array.isArray(mappedPreviousTransaction)
-        ? // All transfers should have same execution date but the last is "true" previous
-          mappedPreviousTransaction.at(-1)
-        : mappedPreviousTransaction;
-    })();
-
-    const transactions = await (async (): Promise<Array<TransactionItem>> => {
-      const mappedTransactions = await Promise.all(
-        transactionsDomain.map((transaction) => {
-          return this.mapTransaction(transaction, chainId, safe, onlyTrusted);
-        }),
-      );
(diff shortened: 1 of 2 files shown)


src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -134,34 +111,16 @@ export class TransactionsHistoryMapper {
...
-      // The actual value of the group should be in the UTC timezone instead
-      // A group should always have at least one transaction.
-      return {
-        timestamp: this.getTransactionTimestamp(transactions[0]).getTime(),
-        transactions: transactions,
-      };
+  ): TransactionItem[][] {
+    const grouped = groupBy(transactions, ({ transaction }) => {
+      // timestamp will always be defined for historical transactions
+      const date = new Date(transaction.timestamp ?? 0);
```
hectorgomezv, line 120: Have you considered here `CreationTransaction`? Those won't include a `timestamp` but a `created` field holding its timestamp I think.

src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -58,71 +44,62 @@ export class TransactionsHistoryMapper {
...
+    const transactions = await (async (): Promise<Array<TransactionItem>> => {
+      const mappedTransactions = await Promise.all(
+        transactionsDomain.map((transaction) => {
+          return this.mapTransaction(transaction, chainId, safe, onlyTrusted);
+        }),
+      );
+      return mappedTransactions
+        .filter(<T>(x: T): x is NonNullable<T> => x != null)
+        .flat();
+    })();
```
hectorgomezv, line 82: I'm not against this implementation, but have you considering extracting these to private functions instead of using IIFEs? Do you see an advantage on using the latter?

src/routes/transactions/mappers/transactions-history.mapper.ts
iamacook: As the transactions are first mapped, [`created` is mapped as `timestamp`](https://github.com/safe-global/safe-client-gateway/blob/refactor-history-mapping/src/routes/transactions/mappers/creation-transaction/creation-transaction.mapper.ts#L46) here.

src/routes/transactions/mappers/transactions-history.mapper.ts
iamacook: I originally only had an IIFE for retrieving the "previous" transaction to keep the `slice` in the main function as it can be somewhat confusing otherwise. I've moved it to it's own function and added a relevant comment in eada02d1a1f2eb2ea6b9dc2bf4ed5af165149105.

As for the `mappedTransaction` IIFE, it's unnecessary and I removed it.

hectorgomezv (approved):

PR #1507, opened 2024-05-06
Only return awaiting confirmations count if it is required

iamacook:
## Summary

`awaitingConfirmation` was previously being calculated according to whether `walletAddress` had submitted a confirmation, regardless of whether a sufficient confirmations exist. This meant that a fully confirmed transaction, when not signed by the given `walletAddress`, would still increase the awaiting confirmation count.

This now takes `confirmationsRequired` into account, only returning an awaiting count if it is `> 0`.

## Changes

- Take the `confirmationsRequired` into account when increasing the awaiting confirmation count
- Add test coverage

### DIFF
--- a/src/routes/safes/safes.service.ts
+++ b/src/routes/safes/safes.service.ts
@@ -206,16 +206,19 @@ export class SafesService {
     transactions: Array<MultisigTransaction>;
     walletAddress: string;
   }): number {
-    return args.transactions.reduce((acc, { confirmationsRequired, confirmations }) => {
-      const isConfirmed = confirmationsRequired === 0
-      const isSignedByWallet = confirmations?.some((confirmation) => {
-        return confirmation.owner === args.walletAddress;
-      })
-      if (!isConfirmed && !isSignedByWallet) {
-        acc++
-      }
-      return acc;
-    }, 0);
+    return args.transactions.reduce(
+      (acc, { confirmationsRequired, confirmations }) => {
+        const isConfirmed = confirmationsRequired === 0;
+        const isSignedByWallet = confirmations?.some((confirmation) => {
+          return confirmation.owner === args.walletAddress;
+        });
+        if (!isConfirmed && !isSignedByWallet) {
+          acc++;
+        }
+        return acc;
+      },
+      0,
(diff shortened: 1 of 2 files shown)


hectorgomezv (approved):

src/routes/safes/safes.service.ts
```diff
@@ -206,12 +206,19 @@ export class SafesService {
...
-      const isConfirmed = !!confirmations?.some((confirmation) => {
-        return confirmation.owner === args.walletAddress;
-      });
-      return isConfirmed ? acc - 1 : acc;
-    }, args.transactions.length);
+    return args.transactions.reduce(
+      (acc, { confirmationsRequired, confirmations }) => {
+        const isConfirmed = confirmationsRequired === 0;
+        const isSignedByWallet = confirmations?.some((confirmation) => {
+          return confirmation.owner === args.walletAddress;
```
katspaugh, line 213: Are these addresses always in the same format wrt checksum?

iamacook: Yes

src/routes/safes/safes.service.ts
```diff
@@ -206,12 +206,19 @@ export class SafesService {
...
-    return args.transactions.reduce((acc, { confirmations }) => {
-      const isConfirmed = !!confirmations?.some((confirmation) => {
-        return confirmation.owner === args.walletAddress;
-      });
-      return isConfirmed ? acc - 1 : acc;
-    }, args.transactions.length);
+    return args.transactions.reduce(
+      (acc, { confirmationsRequired, confirmations }) => {
+        const isConfirmed = confirmationsRequired === 0;
+        const isSignedByWallet = confirmations?.some((confirmation) => {
```
katspaugh, line 212: A slight optimization

```suggestion
        const isSignedByWallet = isConfirmed && confirmations?.some((confirmation) => {
```

iamacook: Added in 1dcf67e02f2fa9d046f63154ba84c64688319f71.

hectorgomezv (approved):

PR #1524, opened 2024-05-08
Filter out imitation transfers

iamacook:
## Summary

This adds `trusted` param-specific filtering of imitation transfers behind a feature flag. It removes outgoing, ERC-20 transfers that imitate the value and recipient in vanity of their neighbour, breaking down #1428.

It maps over the transaction history, keeping the transaction if the following conditions are met:

1. Was transaction executed by Safe?
2. Is there no previous transactions?
3. Is (previous transaction...
   - ..._not_ an ERC-20 transfer?
   - ...incoming?
   - ...of different value?
   - ...to same recipient?
   - ..._not_ a vanity address?

## Changes

- Add env. vars. for feature flag and number of chars. deemed vanity
- Add `ImitationTransactionsHelper`
   - `filterOutgoingErc20ImitationTransfers`
- Conditionally filter transfers if `trusted` param is `true` and feature enabled
- Add associated test coverage

### DIFF
--- a/src/routes/transactions/helpers/imitation-transactions.helper.ts
+++ b/src/routes/transactions/helpers/imitation-transactions.helper.ts
@@ -1,0 +1,104 @@
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { TransactionItem } from '@/routes/transactions/entities/transaction-item.entity';
+import {
+  isTransferTransactionInfo,
+  TransferDirection,
+} from '@/routes/transactions/entities/transfer-transaction-info.entity';
+import { isErc20Transfer } from '@/routes/transactions/entities/transfers/erc20-transfer.entity';
+import { Inject } from '@nestjs/common';
+
+export class ImitationTransactionsHelper {
+  private readonly vanityAddressChars: number;
+
+  constructor(
+    @Inject(IConfigurationService) configurationService: IConfigurationService,
+  ) {
+    this.vanityAddressChars = configurationService.getOrThrow(
+      'mappings.imitationTransactions.vanityAddressChars',
+    );
+  }
+
+  /**
+   * Filters out outgoing ERC20 transfers that imitate their direct predecessor in
+   * value and vanity (but not the exact) address of the recipient.
+   *
+   * @param transactions - list of transactions to filter
+   * @param previousTransaction - transaction to compare last {@link transactions} against
+   *
+   * Note: this only handles singular imitation transfers. It does not handle multiple
+   * imitation transfers in a row, nor does it compare batched multiSend transactions
+   * as the "distance" between those batched and their imitation may not be immediate.
+   */
+  filterOutgoingErc20ImitationTransfers(
+    transactions: Array<TransactionItem>,
+    previousTransaction: TransactionItem | undefined,
+  ): Array<TransactionItem> {
+    // TODO: Instead of filtering, mark transactions so client can display them differently
+    return transactions.filter((item, i, arr) => {
+      // Executed by Safe - cannot be imitation
+      if (item.transaction.executionInfo) {
+        return item;
+      }
+
+      const prevItem = i === arr.length - 1 ? previousTransaction : arr[i + 1];
+
+      // No reference transaction to filter against
+      if (!prevItem) {
+        return item;
+      }
+
+      const txInfo = item.transaction.txInfo;
+      const prevTxInfo = prevItem.transaction.txInfo;
+
+      if (
+        // Only consider transfers...
+        !isTransferTransactionInfo(txInfo) ||
+        !isTransferTransactionInfo(prevTxInfo) ||
+        // ...of ERC20s...
+        !isErc20Transfer(txInfo.transferInfo) ||
+        !isErc20Transfer(prevTxInfo.transferInfo)
+      ) {
+        return item;
+      }
+
+      // ...that are outgoing
+      const isOutgoing = txInfo.direction === TransferDirection.Outgoing;
+      const isPrevOutgoing =
+        prevTxInfo.direction === TransferDirection.Outgoing;
+      if (!isOutgoing || !isPrevOutgoing) {
+        return item;
+      }
+
+      // Imitation transfers are of the same value...
+      const isSameValue =
+        txInfo.transferInfo.value === prevTxInfo.transferInfo.value;
+      if (!isSameValue) {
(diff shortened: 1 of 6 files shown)


src/routes/transactions/helpers/imitation-transactions.helper.ts
```diff
@@ -0,0 +1,104 @@
...
+        !isTransferTransactionInfo(txInfo) ||
+        !isTransferTransactionInfo(prevTxInfo) ||
+        // ...of ERC20s...
+        !isErc20Transfer(txInfo.transferInfo) ||
+        !isErc20Transfer(prevTxInfo.transferInfo)
+      ) {
+        return item;
+      }
+
+      // ...that are outgoing
```
katspaugh, line 64: The pattern I've seen was 1) a legit outgoing transfer 2) incoming scam transfers.

Screenshot 2024-05-08 at 10 43 17

src/routes/transactions/helpers/imitation-transactions.helper.ts
```diff
@@ -0,0 +1,104 @@
...
+        return item;
+      }
+      return !this.isVanityAddress(
+        txInfo.recipient.value,
+        prevTxInfo.recipient.value,
+      );
+    });
+  }
+
+  private isVanityAddress(address1: string, address2: string): boolean {
```
katspaugh, line 92: Not sure "vanity" here represents the nature of this scam addresses. A better name would be an "imitator address" perhaps?

src/routes/transactions/helpers/imitation-transactions.helper.ts
iamacook: As incoming transfers are legitimate, [we proposed](https://www.notion.so/safe-global/Address-poisoning-heuristics-01153b329f544e9cb92384ebf87f70d5?pvs=4#e3171dc299bd4cc39d57a283ae104052) only filtering/marking outgoing ones.

src/routes/transactions/helpers/imitation-transactions.helper.ts
iamacook: Updated in 84a9cbb13848efab9c180485e2383b95fe9e0e99.

iamacook:
I'm marking this as a draft whilst addressing the tests.

src/routes/transactions/helpers/imitation-transactions.helper.ts
katspaugh: Not sure I understand why all the complicated logic is needed for just outgoing txs then. Aren't they trivially detectable by a missing/bad nonce?

iamacook: It doesn't matter about the type. Any transaction that transfers an ERC-20 [will emit a `Transfer` event](https://eips.ethereum.org/EIPS/eip-20#transfer-1), which is the "reference marker" used by imitators.

src/routes/transactions/helpers/imitation-transactions.helper.ts
```diff
@@ -0,0 +1,104 @@
...
+    previousTransaction: TransactionItem | undefined,
+  ): Array<TransactionItem> {
+    // TODO: Instead of filtering, mark transactions so client can display them differently
+    return transactions.filter((item, i, arr) => {
+      // Executed by Safe - cannot be imitation
+      if (item.transaction.executionInfo) {
+        return item;
+      }
+
+      const prevItem = i === arr.length - 1 ? previousTransaction : arr[i + 1];
```
hectorgomezv, line 43: Nit: could we include a brief comment explaining the logic followed here? Something like: we compare each item agains the next item, unless the current item is the last one, in which case we compare it against `previousTransaction`.

src/routes/transactions/helpers/imitation-transactions.helper.ts
```diff
@@ -0,0 +1,104 @@
...
+  filterOutgoingErc20ImitationTransfers(
+    transactions: Array<TransactionItem>,
+    previousTransaction: TransactionItem | undefined,
+  ): Array<TransactionItem> {
+    // TODO: Instead of filtering, mark transactions so client can display them differently
+    return transactions.filter((item, i, arr) => {
+      // Executed by Safe - cannot be imitation
+      if (item.transaction.executionInfo) {
+        return item;
+      }
```
hectorgomezv, line 41: Nit: `filter` expects a boolean return value to decide whether to include the item in the resulting array. I think it's OK to return the item itself, but being stricter maybe return `true` is a bit more appropriate?

src/routes/transactions/helpers/imitation-transactions.helper.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    // Same address is not an imitation
+    if (a1 === a2) {
+      return false;
+    }
+
+    const isSamePrefix =
+      // Ignore `0x` prefix
+      a1.slice(2, this.vanityAddressChars) ===
+      a2.slice(2, this.vanityAddressChars);
```
hectorgomezv, line 99: Just confirming: as this is slicing the `0x` part, only the two following characters are being compared. I mean for example `0x67b450ac7bec2aae5c13ebe4dbaf4f2fa5da0c8d` and `0x67a06f0fecfa567b0c52dc31db17a0b7ddf10252` are considered to be sharing the same prefix (`0x67`). Ommiting the `0x` part when comparing is something intended, right?

src/routes/transactions/helpers/imitation-transactions.helper.ts
iamacook: Added in 56f27fb54391f86585816b7df18a2d85265fc5ba.

src/routes/transactions/helpers/imitation-transactions.helper.ts
iamacook: Totally agree. Changed in 56f27fb54391f86585816b7df18a2d85265fc5ba.

src/routes/transactions/helpers/imitation-transactions.helper.ts
iamacook: Good catch, I've added `2` onto `this.vanityAddressChars` in 56f27fb54391f86585816b7df18a2d85265fc5ba.

hectorgomezv (approved):

PR #1527, opened 2024-05-08
Authorise recovery module adding/deleting with JWT authentication

iamacook:
## Summary

Prior to this, adding/deleting a recovery module address required signing a message with each request. This migrates the suboptiomal authentication/authorisation to the new JWT/SIWE-based approach.

The user needs to have authenticated via SIWE in order to add/remove a recovery module address to their Safe.

## Changes

- Remove `OnlySafeOwner`, `TimestampGuard`, `DisableRecoveryAlertsGuard` and `EnableRecoveryAlertsGuard` and associated tests
- Replace above guard usage `AuthGuard`
- Checksum incoming require Safe/module addresses
- Pass `AuthPayload` to domain and assert chain, signer and ownership before adding/deleting module
- Add/update associated test coverage


src/routes/recovery/recovery.controller.spec.ts
```diff
@@ -122,15 +129,6 @@ describe('Recovery (Unit)', () => {
...
         }
-        if (
-          url ===
-          `${chain.transactionService}/api/v1/modules/${addRecoveryModuleDto.moduleAddress}/safes/`
-        ) {
-          return Promise.resolve({
-            status: 200,
-            data: { safes: [safe.address] },
-          });
-        }
```
hectorgomezv, line 133: Why is this not needed anymore? I'm checking the deleted file `enable-recovery-alerts.guard.ts` and it seems a precondition for enabling and disabling the alerts was to check the module was already activated for the Safe (maybe I'm misinterpreting things), is this no longer true?

hectorgomezv (commented):
Great job! I've left a minor doubt only.

src/routes/recovery/recovery.controller.spec.ts
iamacook: Good point, I overlooked adding the check. I've included it (and added test coverage) as of 5b05a42855a3d6885d62caf5bd8f9020d76735b2.

hectorgomezv (approved):

PR #1529, opened 2024-05-09
Get prices provider configuration from Safe Config Service

hectorgomezv:
## Summary
This PR changes the source for the Prices Provider (currently CoinGecko) configuration from the CGW hardcoded (`configuration.ts`) one, by the `Chain` entity coming from the Safe Config Service.

In this first iteration, the hardcoded configuration (`configuration.ts`) is kept as a fallback. If the Prices Provider configuration data is not present in the Config Service response, then the fallback data is used.

For more context see the Config Service PR: https://github.com/safe-global/safe-config-service/pull/1124

## Changes
- Changes the primary source for both `nativeCoinId` and `chainName` properties in `CoingeckoApi` to the `Chain` entity passed.
- Changes several interfaces accordingly to pass the whole `Chain` entity (holding the needed configuration) instead of passing the `chainId` property only.
- Adds the prices provider configuration fields to the `Chain` entity/schema.
- Modifies the affected test classes.

### DIFF
--- a/src/datasources/balances-api/coingecko-api.service.spec.ts
+++ b/src/datasources/balances-api/coingecko-api.service.spec.ts
@@ -250,6 +250,67 @@ describe('CoingeckoAPI', () => {
     );
   });
 
+  // TODO: remove this after the prices provider data is migrated to the Config Service
+  it('should return and cache one token price (using the fallback configuration)', async () => {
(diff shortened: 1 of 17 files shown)


hectorgomezv:
I've added some changes to adapt the PR to the modified `pricesProvider` field structure (see https://github.com/safe-global/safe-config-service/pull/1124)

iamacook (approved):

PR #1542, opened 2024-05-14
Add flagging (alongside filtering) of imitation transfers

iamacook:
## Summary

The demonstative address poisoning feature currently filters imitation transfers. This means that there is no possibility of retrieving them by the client.

This adjusts the aforementioned to either flag _or_ filter dependent on an `imitation` query parameters. When requesting the transaction history, setting the query parameter to `false` filters out dusting attacks and to `true` (the default) returns with.

The new flag is added _only_ to outgoing ERC-20 transfers that imitate their direct predecessor in terms of value and recipient vanity.

Note: the mock data within tests is currently realworld data. On top of this, we should add mock data, as well as a mixture. A TODO has been added to the code signifying this.

## Changes

- Adjust configuration naming accordingly:
  - `features.imitationFiltering` -> `features.imitationMapping`
    - (`process.env.FF_IMITATION_FILTERING` -> `FF_IMITATION_MAPPING`)
  - `mapping.imitationTransactions` -> `mappings.imitation`
- Add `imitation` flag to `Erc20TransferEntity` (`transaction.txInfo.transferInfo.imitation`)
- Rename `ImitationTransactionsHelper` to `TransferImitationMapper`
- Add flagging logic to `TransferImitationMapper`, separate from filtering
- Add `imitation` query param. and pass it to mapper accordingly

### DIFF
--- a/src/routes/transactions/transactions.controller.ts
+++ b/src/routes/transactions/transactions.controller.ts
@@ -233,7 +233,7 @@ export class TransactionsController {
     timezoneOffsetMs: number,
     @Query('trusted', new DefaultValuePipe(true), ParseBoolPipe)
     trusted: boolean,
-    @Query('imitation', new DefaultValuePipe(false), ParseBoolPipe)
+    @Query('imitation', new DefaultValuePipe(true), ParseBoolPipe)
     imitation: boolean,
   ): Promise<Partial<TransactionItemPage>> {
     return this.transactionsService.getTransactionHistory({
--- a/src/routes/transactions/transactions.controller.ts
+++ b/src/routes/transactions/transactions.controller.ts
@@ -233,6 +233,8 @@ export class TransactionsController {
     timezoneOffsetMs: number,
     @Query('trusted', new DefaultValuePipe(true), ParseBoolPipe)
     trusted: boolean,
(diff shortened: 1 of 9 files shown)


src/routes/transactions/transactions.controller.ts
```diff
@@ -233,6 +233,8 @@ export class TransactionsController {
     timezoneOffsetMs: number,
     @Query('trusted', new DefaultValuePipe(true), ParseBoolPipe)
     trusted: boolean,
+    @Query('imitation', new DefaultValuePipe(false), ParseBoolPipe)
```
hectorgomezv, line 236: I'm not against this, just curious: we are filtering out imitation transfers by default, right? Is that the clients expected behaviour?

hectorgomezv (approved):

src/routes/transactions/transactions.controller.ts
iamacook: This was based on the demo implementation. As it directly affects the clients, setting it to `true` by default makes the most sense. I changed it in c2473aa1c93d665a2ed412471be74bd38e08c9ed.

hectorgomezv (approved):

PR #1543, opened 2024-05-14
Add Campaigns retrieval to LockingApi datasource

hectorgomezv:
## Summary

This PR adds the datasource layer part for getting a `Page` of `Campaigns` from the Safe Locking Service.

## Changes
- Adds a `Campaign` entity and schema.
- Adds a new `getCampaigns` function to the `LockingApi` datasource.


src/datasources/locking-api/locking-api.service.ts
```diff
@@ -24,6 +25,27 @@ export class LockingApi implements ILockingApi {
       this.configurationService.getOrThrow<string>('locking.baseUri');
   }
 
+  async getCampaigns(args: {
+    limit?: number;
+    offset?: number;
+  }): Promise<Page<Campaign>> {
+    try {
+      const url = `${this.baseUri}/api/v1/campaigns`;
+      const { data } = await this.networkService.get<Page<Campaign>>({
```
iamacook, line 34: Do we know how often the campaign data will update/whether it's cached by the Locking Service? It might be beneficial for us to cache it. What do you think?

src/domain/locking/entities/campaign.entity.ts
```diff
@@ -0,0 +1,4 @@
+import { CampaignSchema } from '@/domain/locking/entities/schemas/campaign.schema';
+import { z } from 'zod';
+
+export type Campaign = z.infer<typeof CampaignSchema>;
```
iamacook, line 4: In some places we locate the schema with the entity as it's a one liners. Should we do the same here?

Unrelated to this PR, we should also do the same everywhere else.

src/domain/locking/entities/schemas/__tests__/campaign.schema.spec.ts
```diff
@@ -0,0 +1,94 @@
...
+  it('should coerce lastUpdated to a date', () => {
+    const lastUpdated = faker.date.recent();
+    const campaign = campaignBuilder()
+      .with('lastUpdated', lastUpdated.toISOString() as unknown as Date)
+      .build();
+
+    const result = CampaignSchema.safeParse(campaign);
+
+    expect(result.success && result.data.lastUpdated).toBeInstanceOf(Date);
+  });
```
iamacook, line 46: Nit: we can refactor these to use `it.each`

src/datasources/locking-api/locking-api.service.ts
iamacook: After discussion, it has been confirmed that the Locking Service will cache. We will leave this as is for now.

src/domain/locking/entities/campaign.entity.ts
hectorgomezv: Agreed, changed in [9b82116](https://github.com/safe-global/safe-client-gateway/pull/1543/commits/9b821161c6a8779bba527895f7d11893fc500ad5)

PR #1544, opened 2024-05-15
Make `Chain['pricesProvider']` optional

iamacook:
## Summary

We currently haven't fully migrated to fetching prices provider configuration from the Config Service and we have a local configuration fallback. This does not, however, take into account validation of the fetched data.

When running the project locally against the production Config Service, the schema was throwing and preventing retrieval of chain data because of this.

This makes the `Chain['pricesProvider']` of the `ChainSchema` optional, as well as adds a temporary test and relative TODO comments.

## Changes

- Make `ChainSchema['pricesProvider']` `optional` and add test coverage
- Add relevant optional chaining to `CoingeckoApi`

### DIFF
--- a/src/datasources/balances-api/coingecko-api.service.spec.ts
+++ b/src/datasources/balances-api/coingecko-api.service.spec.ts
@@ -168,6 +168,7 @@ describe('CoingeckoAPI', () => {
     });
 
     const expectedCacheDir = new CacheDir(
+      // @ts-expect-error - TODO: remove after migration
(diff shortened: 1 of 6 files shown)


src/datasources/balances-api/coingecko-api.service.spec.ts
```diff
@@ -168,13 +168,15 @@ describe('CoingeckoAPI', () => {
     });
 
     const expectedCacheDir = new CacheDir(
+      // @ts-expect-error - TODO: remove after migration
```
iamacook, line 171: I chose to expect and error over assertion or optional chaining as this will error when we remove the `optional` added here. The others wouldn't.

PR #1545, opened 2024-05-15
Add LockingApi.getCampaignById

hectorgomezv:
Depends on https://github.com/safe-global/safe-client-gateway/pull/1543

## Changes
- Adds `ILockingApi.getCampaignById` along with its implementation.

### DIFF
--- a/src/datasources/locking-api/locking-api.service.ts
+++ b/src/datasources/locking-api/locking-api.service.ts
@@ -25,6 +25,16 @@ export class LockingApi implements ILockingApi {
       this.configurationService.getOrThrow<string>('locking.baseUri');
   }
 
+  async getCampaignById(campaignId: string): Promise<Campaign> {
+    try {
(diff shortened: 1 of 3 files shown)


src/datasources/locking-api/locking-api.service.ts
```diff
@@ -25,6 +25,16 @@ export class LockingApi implements ILockingApi {
       this.configurationService.getOrThrow<string>('locking.baseUri');
   }
 
+  async getCampaignById(campaignId: string): Promise<Campaign> {
+    try {
+      const url = `${this.baseUri}/api/v1/campaigns/${campaignId}`;
+      const { data } = await this.networkService.get<Campaign>({ url });
```
iamacook, line 31: Do we need to take caching into consideration? Or does the Locking Service?

iamacook: After discussion, it has been confirmed that the Locking Service will cache. We will leave this as is for now.

iamacook (approved):

PR #1553, opened 2024-05-16
Ensure `CreateMessageDto['safeAppId']` is an integer

iamacook:
## Summary

When creating an off-chain message, a `safeAppId` can be providing to map the details of a Safe App to it. The Transaction Service validates this to ensure that it is a an integer (including 0) or `null`. However, we currently allow any number/float. If a float is provided, for example, it will be allowed by the Gateway but not the Transaction Service.

This increases the strictness of message creation validation to only allow `null`, `0` or positive integers.

## Changes

- Increase strictness of `CreateMessageDtoSchema['safeAppId']` and add appropriate test coverage
- Mirror strictness in `createMessageDtoBuilder`

### DIFF
--- a/src/routes/messages/entities/schemas/__tests__/create-message.dto.schema.spec.ts
+++ b/src/routes/messages/entities/schemas/__tests__/create-message.dto.schema.spec.ts
@@ -136,7 +136,7 @@ describe('CreateMessageDtoSchema', () => {
       );
     });
 
-    it('should not validate optional safeAppId, defaulting to null', () => {
+    it('should validate without safeAppId, defaulting to null', () => {
       const createMessageDto = createMessageDtoBuilder().build();
       // @ts-expect-error - inferred type doesn't allow optional properties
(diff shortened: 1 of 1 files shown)


src/routes/messages/entities/schemas/__tests__/create-message.dto.schema.spec.ts
```diff
@@ -5,98 +5,167 @@ import { faker } from '@faker-js/faker';
...
+            expected: 'integer',
+            received: 'float',
+            message: 'Expected integer, received float',
+            path: ['safeAppId'],
+          },
+        ]),
+      );
+    });
+
+    it('should not validate optional safeAppId, defaulting to null', () => {
```
hectorgomezv, line 139:
```suggestion
    it('should validate without safeAppId, defaulting to null', () => {
```

hectorgomezv (approved):

hectorgomezv (approved):

PR #1564, opened 2024-05-17
Improve imitation detection to cover multiple and incoming transfers

iamacook:
⚠️ Although final, this should ideally be discussed before merging.

## Summary

Current address poisoning detection only compares the direct predecessor, flagging/filtering only outgoing transfers that imitate. This improves detection in the following manner:

- Comparison now occurs relative to a "lookup distance". This means that multiple imitation transafers within that distance, relative to the legitimate transaction are detected.
- Incoming transfers are also now detected.
- Imitation tokens that leverage varying decimals are also now detected.

## Changes

- Add lookup distance
- Compare sender/recipient according to direction of transaction
- Compare _formatted_ values


src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
```diff
@@ -22,99 +31,139 @@ export class TransferImitationMapper {
...
+        if (formattedValue !== prevFormattedValue) {
+          return false;
+        }
+
+        // Imitation transfers match in vanity of address
+        const prevRefAddress = this.getReferenceAddress(prevTxInfo);
+        return this.isImitatorAddress(refAddress, prevRefAddress);
+      });
+
+      txInfo.transferInfo.imitation = isImitation;
```
hectorgomezv, line 128: Nitpick: if the `for` loop `continue` before reaching this point, the value of `txInfo.transferInfo.imitation` is `null`. Do you see value in setting it to `false` in those situations? Maybe the API of this mapper is simpler if we enforce a boolean type in the field?

hectorgomezv (approved):

src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
iamacook: In 5a945cd1c5b703430c253115bc220e94438dfc82 I changed the default value of `Erc20Transfer['imitation']` to `false` in the and explicitly set it here as requested.

hectorgomezv (approved):
🚀

PR #1593, opened 2024-05-28
Remove chain-specific prices provider configuration

iamacook:
## Summary

The chain-specific configuration of the prices provider was [migrated to the Config Service](https://github.com/safe-global/safe-client-gateway/pull/1529), which is now live and configured on production. As such, we no longer need the local configuration or the temporary fallbacks that were in place.

This changes the `Chain['pricesProvider']` to be a required property, propegating the relevant adjustments as well as removing the local configuration.

## Changes

- Make `ChainSchema['pricesProvider']` (and all nested properties) required
- Remove `balances.providers.safe.prices.chains` from (test) configuration
- Remove fallbacks to configuration
- Remove all TODO comments
- Update tests accordingly


hectorgomezv (approved):
👏🏻

PR #1610, opened 2024-06-04
Add and validate against schema for `AssetPrice`

iamacook:
## Summary

When fetching `AssetPrice`s from the CoinGecko cache, we are blindly trusting `JSON.parse`. As it returns an `unknown` type, it is currently being asserted to `AssetPrice`. Whilst we can be relatively confident that the cache only has these, it is better to test these values.

This adds an `AssetPriceSchema` for validating against (with test coverage) and usage of it in the `CoingeckoApi`.

## Changes

- Create `AssetPriceSchema` (with test)
- Validate cache against it

### DIFF
--- a/src/datasources/balances-api/entities/__tests__/asset-price.schema.spec.ts
+++ b/src/datasources/balances-api/entities/__tests__/asset-price.schema.spec.ts
@@ -16,4 +16,45 @@ describe('AssetPriceSchema', () => {
 
     expect(result.success).toBe(true);
   });
+
+  it('should not validate an object with non-object values', () => {
+    const address = faker.finance.ethereumAddress();
+    const assetPrice = {
+      [address]: faker.number.int(),
+    };
+
+    const result = AssetPriceSchema.safeParse(assetPrice);
+
+    expect(!result.success && result.error.issues).toStrictEqual([
+      {
+        code: 'invalid_type',
+        expected: 'object',
+        message: 'Expected object, received number',
+        path: [address],
+        received: 'number',
+      },
+    ]);
+  });
+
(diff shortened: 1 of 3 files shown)


src/datasources/balances-api/entities/asset-price.entity.ts
```diff
@@ -1,3 +1,6 @@
-export interface AssetPrice {
-  [assetName: string]: Record<string, number | null>;
-}
+import { z } from 'zod';
+
+export type AssetPrice = z.infer<typeof AssetPriceSchema>;
+
+// TODO: Enforce Ethereum address keys (and maybe checksum them)
```
iamacook, line 5: I've left this note as it would tie in with refactoring the `CoingeckoApi`. As their API accepts checksummed addresses, we should only work with them - checksumming the response too.

src/datasources/balances-api/entities/__tests__/asset-price.schema.spec.ts
```diff
@@ -0,0 +1,19 @@
...
+          faker.number.float(),
+          null,
+        ]),
+      },
+    };
+
+    const result = AssetPriceSchema.safeParse(assetPrice);
+
+    expect(result.success).toBe(true);
+  });
```
hectorgomezv, line 18: I think we should include at least one test in which the validation not passing, wdyt?

iamacook: Added in 369edc9f04839313571bd5f137e44da0e49d1abd.

hectorgomezv (approved):

PR #1611, opened 2024-06-04
Add schemas for Zerion entities

iamacook:
## Summary

Requests from Zerion for balances/collectibles should return the expected data structure but it is never validated. This adds the relative validation schemas for both and parses against them to ensure validity.

## Changes

- Create `ZerionCollectiblesSchema` (and sub-schemas)
- Create `ZerionBalancesSchema` (and sub-schemas)
- Validate relevant data against the above in `ZerionBalancesApi`
- Update tests accordingly


src/datasources/balances-api/entities/zerion-collectible.entity.ts
```diff
@@ -1,40 +1,77 @@
...
+    .nullable(),
+  interface: z.string().nullable(),
+  name: z.string().nullable(),
+  token_id: z.string(),
+});
+
+const ZerionCollectibleAttributesSchema = z.object({
+  amount: z.string(),
+  changed_at: z.string(), // Coerce?
+  collection_info: ZerionCollectionInfoSchema.nullable(),
```
hectorgomezv, line 60: I'd say we can coerce this, but it's not used anywhere yet, so feel free to leave it as it is if you prefer.

hectorgomezv (approved):

src/datasources/balances-api/entities/zerion-collectible.entity.ts
iamacook: I forgot to remove this working note 🙈 Considering it's not used, I coerced it in 78a9197b4cd143e8e73bc060c66ed657275d8a04 for future use.

hectorgomezv (approved):

PR #1614, opened 2024-06-05
Add transaction hash to `Transaction` entity

iamacook:
## Summary

Each `Transaction` of the history list is returned with an `id` that is specific to that transaction. However, some of which could have been executed in bulk and therefore share the same transaction hash. As we are not returning the transaction hash, it is not possible to distinguish said executions without fetching the individual transaction via `id`.

As the client wants to start [grouping transactions executed in bulk](https://github.com/safe-global/safe-wallet-web/pull/3772), the transaction hash is required. Initially it was fetched per transaction, meaning _all_ transactions of the history was being fetched **on top of** what we currently return. As we have the transaction hash when mapping the list, however, this adds a new `txHash` field to the root of `Transaction` entities.

Note: the type is `string | null` as the `Transaction` entity is the same for the queue and history. It can be safely assumed (in the history) that the `txHash` is a `string` as it has been executed.

## Changes

- Add `txHash` to `Transaction` entity
- Populate `txHash` in creation, module, multisig and transfer mappers
- Update tests accordingly


hectorgomezv:
Looks good to me! What do you think about adding this field to the tests in the other `*.transactions.controller.spec.ts` files? Of course, it could be done separately.

```
transactions-history.controller.spec.ts
get-transaction-by-id.transactions.controller.spec.ts
list-incoming-transfers-by-safe.transactions.controller.spec.ts
list-module-transactions-by-safe.transactions.controller.spec.ts
list-multisig-transactions-by-safe.transactions.controller.spec.ts
preview-transaction.transactions.controller.spec.ts
```

hectorgomezv (approved):

iamacook:
> Looks good to me! What do you think about adding this field to the tests in the other `*.transactions.controller.spec.ts` files? Of course, it could be done separately.
> 
> ```
> transactions-history.controller.spec.ts
> get-transaction-by-id.transactions.controller.spec.ts
> list-incoming-transfers-by-safe.transactions.controller.spec.ts
> list-module-transactions-by-safe.transactions.controller.spec.ts
> list-multisig-transactions-by-safe.transactions.controller.spec.ts
> preview-transaction.transactions.controller.spec.ts
> ```

I'm quite surprised that these tests didn't break. I've updated the relevant tests (apart from preview as it's not present there) accordingly.

hectorgomezv:
> I'm quite surprised that these tests didn't break. I've updated the relevant tests (apart from preview as it's not present there) accordingly.

Yes, I think they didn't break because they are using Jest's `.toMatchObject` to verify the payloads, and that function accepts an object matching a subset of the properties.

The PR looks good to me now btw!

hectorgomezv (approved):

PR #1615, opened 2024-06-05
Add Counterfactual Safes ERC-20 balances

hectorgomezv:
## Summary
Currently, counterfactual Safes are not being indexed by Safe{Core}, and therefore their asset balances can only be obtained by calling `eth_getBalance` directly on a RPC. This approach faces a limitation: only the chain native balances are returned by the RPC call.

This PR leverages the Zerion API to get the balances (both native and ERC-20) of any undeployed (counterfactual) Safe.

## Changes
- `BalancesApiManager.getBalancesApi`now returns the injected `ZerionBalancesApi` when the Safe data is unavailable through the Safe Transaction Service. To accomplish that, the Safe address is now also passed to the function.
- A test to verify this behavior was added.
- Existing tests were modified accordingly to reflect this extra call to the Transaction Service API.

### DIFF
--- a/src/datasources/balances-api/balances-api.manager.ts
+++ b/src/datasources/balances-api/balances-api.manager.ts
@@ -51,6 +51,8 @@ export class BalancesApiManager implements IBalancesApiManager {
       return this.zerionBalancesApi;
     }
 
+    // SafeBalancesApi will be returned only if TransactionApi returns the Safe data.
+    // Otherwise ZerionBalancesApi will be returned as the Safe is considered counterfactual/not deployed.
     try {
       const transactionApi =
         await this.transactionApiManager.getTransactionApi(chainId);
--- a/src/datasources/balances-api/balances-api.manager.ts
+++ b/src/datasources/balances-api/balances-api.manager.ts
@@ -13,6 +13,7 @@ import { IConfigApi } from '@/domain/interfaces/config-api.interface';
 import { IPricesApi } from '@/datasources/balances-api/prices-api.interface';
 import { Inject, Injectable } from '@nestjs/common';
 import { intersection } from 'lodash';
+import { ITransactionApiManager } from '@/domain/interfaces/transaction-api.manager.interface';
 
 @Injectable()
 export class BalancesApiManager implements IBalancesApiManager {
@@ -30,6 +31,8 @@ export class BalancesApiManager implements IBalancesApiManager {
     private readonly httpErrorFactory: HttpErrorFactory,
     @Inject(IZerionBalancesApi) zerionBalancesApi: IBalancesApi,
     @Inject(IPricesApi) private readonly coingeckoApi: IPricesApi,
+    @Inject(ITransactionApiManager)
+    private readonly transactionApiManager: ITransactionApiManager,
   ) {
     this.zerionChainIds = this.configurationService.getOrThrow<string[]>(
       'features.zerionBalancesChainIds',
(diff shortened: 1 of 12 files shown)


src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -92,6 +92,7 @@ export class ZerionBalancesApi implements IBalancesApi {
     safeAddress: `0x${string}`;
     fiatCode: string;
   }): Promise<Balance[]> {
+    // TODO: check the fiatCode is supported.
```
hectorgomezv, line 95: This is not strictly related, but I've noted we make the request to Zerion when the `fiatCode` is invalid, so we should check it beforehand. I will do it in a separate PR if we are in an agreement.

src/datasources/balances-api/balances-api.manager.ts
```diff
@@ -40,11 +43,31 @@ export class BalancesApiManager implements IBalancesApiManager {
...
+    safeAddress: `0x${string}`,
+  ): Promise<IBalancesApi> {
     if (this.zerionChainIds.includes(chainId)) {
       return this.zerionBalancesApi;
     }
 
+    try {
+      const transactionApi =
+        await this.transactionApiManager.getTransactionApi(chainId);
+      await transactionApi.getSafe(safeAddress);
```
iamacook, line 57: Nit: I'd add a comment _why_ we're doing this.

src/datasources/balances-api/balances-api.manager.ts
```diff
@@ -40,11 +43,31 @@ export class BalancesApiManager implements IBalancesApiManager {
...
+      return this._getSafeBalancesApi(chainId);
+    } catch (err) {
```
iamacook, line 59: Nit: I'd we're not logging it, I'd remove the unused var. (I'm surprised the linter didn't pick this up.)

```suggestion
    } catch {
```

src/datasources/balances-api/balances-api.manager.ts
```diff
@@ -40,11 +43,31 @@ export class BalancesApiManager implements IBalancesApiManager {
...
+      return this.zerionBalancesApi;
+    }
+  }
+
+  async getFiatCodes(): Promise<string[]> {
+    const zerionFiatCodes = await this.zerionBalancesApi.getFiatCodes();
+    const safeFiatCodes = await this.coingeckoApi.getFiatCodes();
```
iamacook, line 66: We could make these concurrent. What do you think?
```suggestion
    const [zerionFiatCodes, safeFiatCodes] = await Promise.all([
      this.zerionBalancesApi.getFiatCodes(),
      this.coingeckoApi.getFiatCodes(),
    ]);
```

src/domain/interfaces/balances-api.manager.interface.ts
```diff
@@ -8,10 +8,17 @@ export interface IBalancesApiManager {
    * Each chain is associated with an implementation (i.e.: to a balances
    * provider) via configuration.
    *
+   * Note: if the Safe entity associated with the passed safeAddress cannot be retrieved
+   * by the TransactionApi for the input chain ID, then the Zerion API will be used.
+   *
    * @param chainId - the chain identifier to check.
+   * @param safeAddress - the safe address to check.
```
iamacook, line 15:
```suggestion
   * Note: if the Safe entity associated with th safeAddress cannot be retrieved
   * from the TransactionApi for the chainId, then the ZerionApi will be used.
   *
   * @param chainId - the chain identifier to check.
   * @param safeAddress - the Safe address to check.
```

src/datasources/balances-api/balances-api.manager.ts
hectorgomezv: Absolutely!

src/domain/interfaces/balances-api.manager.interface.ts
hectorgomezv: I've taken your suggestion in a different commit, as I think there's a typo on it:
```
associated with th safeAddress
```

I've changed it in [32c52a2](https://github.com/safe-global/safe-client-gateway/pull/1615/commits/32c52a21ac656ee3375d237d82204c5a0fca7c61)

src/datasources/balances-api/balances-api.manager.ts
hectorgomezv: Added in [32c52a2](https://github.com/safe-global/safe-client-gateway/pull/1615/commits/32c52a21ac656ee3375d237d82204c5a0fca7c61), please let me know if you find it not clear enough 🙂

iamacook (approved):

PR #1617, opened 2024-06-06
Check fiatCode before getting Zerion balances

hectorgomezv:
## Changes
- Adds a `fiatCode` check before calling the Zerion's balances endpoint.

### DIFF
--- a/src/datasources/balances-api/zerion-balances-api.service.ts
+++ b/src/datasources/balances-api/zerion-balances-api.service.ts
@@ -92,7 +92,7 @@ export class ZerionBalancesApi implements IBalancesApi {
     safeAddress: `0x${string}`;
     fiatCode: string;
   }): Promise<Balance[]> {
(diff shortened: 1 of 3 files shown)


src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -354,4 +354,9 @@ export class ZerionBalancesApi implements IBalancesApi {
     );
     if (current > this.limitCalls) throw new LimitReachedError();
   }
+
+  private _checkFiatCode(fiatCode: string): void {
+    if (!this.fiatCodes.includes(fiatCode.toUpperCase()))
+      throw new DataSourceError(`Unsupported currency code: ${fiatCode}`, 400);
+  }
```
iamacook, line 361: Nit: I wouldn't abstract this as the call is somewhat hard to follow. What do you think?

iamacook (approved):

PR #1618, opened 2024-06-06
Refactor SiWe-related logic to use viem

iamacook:
## Summary

We use a proprietary implementation of SiWe parsing due to dependency size/incompatibility. We don't accept a string `message`. We expect and object and create the `message` from it (with a proprietary function). The [official library](https://github.com/spruceid/siwe) parses the aforementioned into an object which follows the standard.

Now that viem has SiWe support and we have the library installed, this migrates to using it instead of our priorietary solution. It also adapts our implementation to accept a string `message` and parses it instead of relying on an object.

## Changes

- Removes verification logic from `SiweApi` datasource and
- Modify `ISiweRepository` it to accept a string `message` and` signature` (matching the SiWe standard)
- Update tests accordingly


src/domain/siwe/siwe.repository.ts
```diff
@@ -1,25 +1,49 @@
...
+
   constructor(
-    @Inject(ISiweApi)
-    private readonly siweApi: ISiweApi,
-  ) {}
+    @Inject(IConfigurationService)
+    private readonly configurationService: IConfigurationService,
+    @Inject(LoggingService)
+    private readonly loggingService: ILoggingService,
+    @Inject(CacheService) private readonly cacheService: ICacheService,
```
hectorgomezv, line 27: While I agree with accessing `view/siwe` functions directly from the domain, I also think it's a good idea to maintain the `CacheService/CacheRouter` usage within the datasources layer. Wdyt?

If you agree with that, maybe we could keep `SiweApiService` as a tiny/dummy class containing only the methods to encapsulate the cache calls, by keeping `storeNonce`, `getNonce` and `clearNonce` functions only in its interface.

src/domain/siwe/siwe.repository.interface.ts
```diff
@@ -1,18 +1,18 @@
...
-import { VerifyAuthMessageDto } from '@/routes/auth/entities/verify-auth-message.dto.entity';
 import { Module } from '@nestjs/common';
 
 export const ISiweRepository = Symbol('ISiweRepository');
 
 export interface ISiweRepository {
   generateNonce(): Promise<{ nonce: string }>;
 
-  isValidMessage(args: VerifyAuthMessageDto): Promise<boolean>;
+  verifyMessage(args: {
```
hectorgomezv, line 9: Nit: I don't have a strong opinion about this, but maybe the `is...` prefix on the function name is a bit more declarative as it always returns a boolean?

src/routes/auth/auth.controller.spec.ts
```diff
@@ -176,37 +185,34 @@ describe('AuthController', () => {
...
-            statusCode: 422,
+            message: 'Unauthorized',
+            statusCode: 401,
           });
         });
-      // Nonce not deleted
-      await expect(cacheService.get(cacheDir)).resolves.toBe(
-        nonceResponse.body.nonce,
-      );
+      // Nonce deleted
```
hectorgomezv, line 197: The nonce wasn't deleted before this change because this validation failed when the `ValidationPipe(VerifyAuthMessageDtoSchema)` was invoked at the route level, without cache access, right? (I think this _side-effect_ of the change makes total sense, I'm just trying to confirm my reasoning 🙂)

hectorgomezv (approved):

hectorgomezv (commented):
Looks good to me overall 👏 , just left minor comments/thoughts.

src/domain/siwe/siwe.repository.ts
iamacook: Agreed. I integrated your suggestion in c21988dcce5337f145a7ba04744c2aac15e88d52.

src/domain/siwe/siwe.repository.interface.ts
iamacook: I took the name from the official SiWe message but `is...` does make more sense. Reverted in c21988dcce5337f145a7ba04744c2aac15e88d52.

src/routes/auth/auth.controller.spec.ts
iamacook: Correct. I think it makes more sense having moved it.

hectorgomezv: Absolutely! Thanks!

hectorgomezv (approved):

PR #1619, opened 2024-06-07
Enable `@typescript-eslint/no-unsafe-call` lint rule

iamacook:
## Summary

With ESLint 9 the recommended rules were updated. As such, we disabled a few to prevent the upgrade spanning the entire codebase.

This reenables the `@typescript-eslint/no-unsafe-call` rule that was disabled.

Note: the `esModuleInterop` compiler option of TypeScript was enabled after defining the types for `postgres-shift`. Enabling it required all `import * as example from ...` to be changed to `import example from ...`.

## Changes

- Enable `@typescript-eslint/no-unsafe-call` lint rule
- Update respective errors


hectorgomezv (approved):
👏🏻

PR #1620, opened 2024-06-07
Change E2E testing Safe addresses

hectorgomezv:
## Changes
- Change the testing Safe address used in E2E tests to `0x8675B754342754A30A2AeF474D114d8460bca19b`

### DIFF
--- a/src/routes/common/__tests__/constants.ts
+++ b/src/routes/common/__tests__/constants.ts
@@ -1,11 +1,9 @@
-import { getAddress } from 'viem';
-
 interface TestSafe {
   address: `0x${string}`;
   owners: `0x${string}`[];
 }
 
 export const TEST_SAFE: TestSafe = {
-  address: getAddress('0x8675B754342754A30A2AeF474D114d8460bca19b'),
-  owners: [getAddress('0x6c15f69EE76DA763e5b5DB6f7f0C29eb625bc9B7')],
+  address: '0x8675B754342754A30A2AeF474D114d8460bca19b' as const,
+  owners: ['0x6c15f69EE76DA763e5b5DB6f7f0C29eb625bc9B7' as const],
 };
--- a/src/routes/common/__tests__/constants.ts
+++ b/src/routes/common/__tests__/constants.ts
@@ -1,0 +1,11 @@
+import { getAddress } from 'viem';
+
+interface TestSafe {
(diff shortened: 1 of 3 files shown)


iamacook (commented):
Nit: I think we should move these to a constants file. It could even be a mapping of Safe address to owners. What do you think?

hectorgomezv:
> Nit: I think we should move these to a constants file. It could even be a mapping of Safe address to owners. What do you think?

I totally agree. I've added the data to a `constants.ts` file under `src/routes/common/__tests__`. Please let me know if you think it should be stored in another place. Also, I've included the mappings in an object as you suggested. If you agree with the approach, what do you think about including the `chainId` also in that object?

src/routes/common/__tests__/constants.ts
```diff
@@ -0,0 +1,11 @@
+import { getAddress } from 'viem';
+
+interface TestSafe {
+  address: `0x${string}`;
+  owners: `0x${string}`[];
+}
+
+export const TEST_SAFE: TestSafe = {
+  address: getAddress('0x8675B754342754A30A2AeF474D114d8460bca19b'),
+  owners: [getAddress('0x6c15f69EE76DA763e5b5DB6f7f0C29eb625bc9B7')],
```
iamacook, line 10: Setting these to `as const` should negate the need to checksum them.

```suggestion
interface TestSafe {
  address: `0x${string}`;
  owners: `0x${string}`[];
}

export const TEST_SAFE: TestSafe = {
  address: '0x8675B754342754A30A2AeF474D114d8460bca19b' as const,
  owners: ['0x6c15f69EE76DA763e5b5DB6f7f0C29eb625bc9B7' as const],
```

iamacook (commented):
> I totally agree. I've added the data to a `constants.ts` file under `src/routes/common/__tests__`. Please let me know if you think it should be stored in another place. Also, I've included the mappings in an object as you suggested. If you agree with the approach, what do you think about including the `chainId` also in that object?

I think the `chainId` would be a good addition.

hectorgomezv:
> I think the `chainId` would be a good addition.

Added in [1d74f14](https://github.com/safe-global/safe-client-gateway/pull/1620/commits/1d74f14c31a279faf27dc654b26cc29bdab2c713)

iamacook (approved):

PR #1623, opened 2024-06-10
Add `IBlockchainApiManager` for on-chain interactions

iamacook:
## Summary

Smart contract wallets use EIP-1271 to sign messages. As we progress with SiWe-based authenticated, we will also need to add support for verification of messages signed with EIP-1271. Currently, our SiWe implementation assumes that signed messages come only from EOAs. In order to support verification for the above, we will need to make on-chain requests.

This adds a new `IBlockchainManager`, returning a client per chain. It can be used as a basis for future on-chain interactions, e.g. verification of EIP-1271 signatures.

It is important to note that w need a specific API key for the project.

## Changes

- Add relative configuration for Infura API key
- Add `IBlockchainApiManager` and implementation with test coverage
- Clear client on `CHAIN_UPDATE`
- Add relevant test coverage


src/datasources/blockchain/blockchain-api.manager.ts
```diff
@@ -0,0 +1,38 @@
...
+      chain,
+      this.destroyBlockchainApi,
+    );
+    return this.blockchainApiMap[chainId];
+  }
+
+  destroyBlockchainApi(chainId: string): void {
+    if (this.blockchainApiMap?.[chainId]) {
+      delete this.blockchainApiMap[chainId];
+    }
```
hectorgomezv, line 36: This is great. In fact, being strict we should implement the same destroy mechanism in `transaction-api.manager.ts`, as the Transaction Service URLs can also change (even is it's pretty unusual). Now it requires a CGW restart to get a new configuration for a given `TransactionApi` datasource. What do you think?

src/datasources/blockchain/blockchain-api.service.ts
```diff
@@ -0,0 +1,47 @@
...
+      'blockchain.infura.apiKey',
+    );
+  }
+
+  getClient(): PublicClient {
+    return createPublicClient({
+      chain: this.getChain(),
+      transport: http(),
+    });
+  }
```
hectorgomezv, line 27: I'm not totally against this, but wanted to known the approach better: why do we expose the `PublicClient` outside this class? Wouldn't it be more inline with the other datasource  classes (`*-api.service.ts`) to just expose the functions itself (even we are mostly wrapping the client in this case)?

I mean, instead of the having the clients of this class doing:
```
const blockchainClient = blockChainApi.getClient();
const blockNumber = await blockchainClient.getBlockNumber();
```
They would do something like:
```
const blockNumber = await blockchainApi.getBlockNumber();
```

Exposing the `PublicClient` seems a bit like defeating the purpose of having a datasource, which in my opinion has as one of its main benefits the encapsulation of the underlying implementation. What do you think?

src/datasources/blockchain/blockchain-api.service.ts
```diff
@@ -0,0 +1,47 @@
...
+import { Chain, PublicClient, createPublicClient, http } from 'viem';
+
+@Injectable()
+export class BlockchainApi implements IBlockchainApi {
+  private readonly infuraApiKey: string;
+
+  constructor(
+    private readonly configurationService: IConfigurationService,
+    private readonly chain: DomainChain,
+    public readonly destroyClient: (chainId: string) => void,
```
hectorgomezv, line 15: What's the purpose of this function? I understand `BlockchainApiManager.destroyBlockchainApi` is intended to delete the `BlockChainApi` entry in the map, but in this class this function doesn't do anything, right? It's intended to release some resources?

src/datasources/blockchain/blockchain-api.manager.ts
iamacook: Agree that it's unusual but I think it would be a good pattern to copy. We could even consider creating a specific interface for these managers that we implement in both.

src/datasources/blockchain/blockchain-api.service.ts
iamacook: Good point. `IBlockchainApi` is essentially redundant (it was a remanant from trialling the implementation). I removed it in e38f10916a1a2c4fba22cc228849c2777a96d1c6 as per your suggestion.

src/datasources/blockchain/blockchain-api.service.ts
iamacook: It is a [remnant from trialling the implementation](https://github.com/safe-global/safe-client-gateway/pull/1623#discussion_r1635019968). I removed it in e38f10916a1a2c4fba22cc228849c2777a96d1c6.

src/datasources/blockchain/blockchain-api.manager.ts
iamacook: I refactored the `TransactionApiManager` and `BalancesApiManager` to follow the same pattern with a common interface in https://github.com/safe-global/safe-client-gateway/pull/1640. When that is merged, I will propagate the changes into the `BlockchainApiManager` as well.

PR #1626, opened 2024-06-10
Cache Safe existence check for counterfactual balances

iamacook:
## Summary

In order to determine which datasource we use for balances, we need to know whether the Safe is yet deployed. As such, we currently called the Transaction Service, returning the Zerion API if the call fails (the Safe does not exist). This is detrimental as we _always_ make this request.

This introduces a new `isSafe` method on the `TransactionApi` that returns an _indefinitely_ cached `boolean` regarding deployment state. The cache is invalidated on the `SAFE_CREATED` event.

## Changes

- Check `isSafe` within `BalancesApiManager` accordingly and propagate
- Add new `${chainId}_safe_exists_${address}` cache routing

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -22,6 +22,7 @@ export class CacheRouter {
   private static readonly SAFE_APPS_KEY = 'safe_apps';
   private static readonly SAFE_BALANCES_KEY = 'safe_balances';
   private static readonly SAFE_COLLECTIBLES_KEY = 'safe_collectibles';
+  private static readonly SAFE_EXISTS_KEY = 'safe_exists';
   private static readonly SAFE_FIAT_CODES_KEY = 'safe_fiat_codes';
   private static readonly SAFE_KEY = 'safe';
   private static readonly SINGLETONS_KEY = 'singletons';
@@ -114,6 +115,23 @@ export class CacheRouter {
     safeAddress: `0x${string}`;
   }): string {
     return `${args.chainId}_${CacheRouter.SAFE_KEY}_${args.safeAddress}`;
+  }
+
+  static getIsSafeCacheDir(args: {
+    chainId: string;
+    safeAddress: `0x${string}`;
+  }): CacheDir {
+    return new CacheDir(
+      `${args.chainId}_${CacheRouter.SAFE_EXISTS_KEY}_${args.safeAddress}`,
+      '',
+    );
+  }
+
+  static getIsSafeCacheKey(args: {
+    chainId: string;
+    safeAddress: `0x${string}`;
+  }): string {
+    return `${args.chainId}_${CacheRouter.SAFE_EXISTS_KEY}_${args.safeAddress}`;
   }
 
   static getContractCacheDir(args: {
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -121,10 +121,7 @@ export class CacheRouter {
     chainId: string;
     safeAddress: `0x${string}`;
   }): CacheDir {
-    return new CacheDir(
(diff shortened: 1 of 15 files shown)


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -147,6 +150,69 @@ export class TransactionApi implements ITransactionApi {
...
+        throw this.httpErrorFactory.from(this.mapError(error));
+      }
+    })();
+
+    await this.cacheService.set(
+      cacheDir,
+      JSON.stringify(isSafe),
+      // We can indefinitely cache this as an address cannot "un-Safe" itself
+      // and invalidation (clearIsSafe) is called on SAFE_CREATED event
+      TransactionApi.INDEFINITE_EXPIRATION_TIME,
```
hectorgomezv, line 202: I'd store this for `INDEFINITE_EXPIRATION_TIME` if `isSafe === true` only. 

Imagine that for some reason (events source/queue malfunction, downtime, etc.) the service does not receive the `SAFE_CREATED` event. There is no problem if the Safe exists, because as you said it cannot be undeployed from the blockchain, but a false negative will be stored forever, as the `SAFE_CREATED` event won't be emitted again. What do you think?

src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -325,6 +333,128 @@ describe('TransactionApi', () => {
...
+      );
+
+      await expect(service.isSafe(safe.address)).rejects.toThrow(expected);
+
+      expect(cacheService.get).toHaveBeenCalledTimes(1);
+      expect(cacheService.get).toHaveBeenCalledWith(cacheDir);
+      expect(networkService.get).toHaveBeenCalledTimes(1);
+      expect(networkService.get).toHaveBeenCalledWith({
+        url: `${baseUrl}/api/v1/safes/${safe.address}`,
+      });
```
hectorgomezv, line 441: Nit: I'd also test here that `cacheService.set` is not called when an error in the upstream service occurs, wdyt?

hectorgomezv (changes requested):
Great job! 🚀

src/routes/cache-hooks/cache-hooks.controller.spec.ts
```diff
@@ -180,6 +181,9 @@ describe('Post Hook Events (Unit)', () => {
     },
     {
       type: 'SAFE_CREATED',
+      chainId: faker.string.numeric(),
```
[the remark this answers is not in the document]
iamacook: Removed in 43fb39a6923480a82d15146cb4317eb806821f1f.

src/datasources/cache/cache.router.ts
```diff
@@ -116,6 +117,23 @@ export class CacheRouter {
     return `${args.chainId}_${CacheRouter.SAFE_KEY}_${args.safeAddress}`;
   }
 
+  static getIsSafeCacheDir(args: {
+    chainId: string;
+    safeAddress: `0x${string}`;
+  }): CacheDir {
+    return new CacheDir(
+      `${args.chainId}_${CacheRouter.SAFE_EXISTS_KEY}_${args.safeAddress}`,
```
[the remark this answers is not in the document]
iamacook: Changed in 43fb39a6923480a82d15146cb4317eb806821f1f.

src/datasources/transaction-api/transaction-api.service.ts
iamacook: Good point. I changed it in 43fb39a6923480a82d15146cb4317eb806821f1f to set it as described, otherwise using the default expiration time.

src/datasources/transaction-api/transaction-api.service.spec.ts
iamacook: Added in 43fb39a6923480a82d15146cb4317eb806821f1f.

hectorgomezv (approved):

PR #1627, opened 2024-06-10
Support authentication with ERC-6492

iamacook:
## Summary

We currently only support EOAs for authentication. With smart contract wallets, the signature needs to be checked on-chain.

As viem supplies a method for the above, this integrates the `IBlockchainApiManager` with our SiWe implementation to support smart contract wallets.

## Changes

- Import `BlockchainApiManagerModule` in `SiweRepositoryModule`
- Modify signature validation in `SiweRepository` to verifies signatures of EOAs then on-chain
- Create `FakeBlockchainApiManager`/`TestBlockchainApiManager` for mocking `eth_call` responses

### DIFF
--- a/src/domain/siwe/siwe.repository.ts
+++ b/src/domain/siwe/siwe.repository.ts
@@ -8,7 +8,9 @@ import {
   generateSiweNonce,
   parseSiweMessage,
   validateSiweMessage,
+  verifySiweMessage,
 } from 'viem/siwe';
+import { IBlockchainApiManager } from '@/domain/interfaces/blockchain-api.manager.interface';
 
 @Injectable()
 export class SiweRepository implements ISiweRepository {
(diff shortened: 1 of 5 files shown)


src/domain/siwe/siwe.repository.ts
```diff
@@ -106,4 +110,37 @@ export class SiweRepository implements ISiweRepository {
...
+
+    // Else, verify hash on-chain using ERC-6492 for smart contract accounts
+    const blockchainApi = await this.blockchainApiManager.getBlockchainApi(
+      args.chainId,
+    );
+    const client = blockchainApi.getClient();
+    return client.verifySiweMessage({
+      message: args.message,
+      signature: args.signature,
+    });
```
hectorgomezv, line 144: Related with this other comment from the previous PR: https://github.com/safe-global/safe-client-gateway/pull/1623#discussion_r1634962430

Maybe it's simpler if we create and expose `blockchainApi.verifySiweMessage`. For the testing/mocking part it would simplify things also IMO, what do you think?

hectorgomezv (dismissed):

hectorgomezv (approved):

PR #1637, opened 2024-06-11
Add ZerionBalancesSchema tests

hectorgomezv:
## Changes
- Adds `ZerionBalances` entity schema unit tests.
- Changes `nullable()` by `nullish().default(null)` for several fields in the `ZerionBalance` schema, following the [Zerion API documentation](https://developers.zerion.io/reference/listwalletpositions).

### DIFF
--- a/src/datasources/balances-api/entities/zerion-balance.entity.spec.ts
+++ b/src/datasources/balances-api/entities/zerion-balance.entity.spec.ts
@@ -17,6 +17,7 @@ import {
   ZerionQuantitySchema,
 } from '@/datasources/balances-api/entities/zerion-balance.entity';
 import { faker } from '@faker-js/faker';
+import { getAddress } from 'viem';
 import { ZodError } from 'zod';
 
 describe('Zerion Balance Entity schemas', () => {
@@ -65,17 +66,17 @@ describe('Zerion Balance Entity schemas', () => {
 
       const result = ZerionBalanceSchema.safeParse(zerionBalance);
 
-      expect(!result.success && result.error).toStrictEqual(
-        new ZodError([
-          {
-            received: 'invalid',
-            code: 'invalid_literal',
-            expected: 'positions',
-            path: ['type'],
-            message: 'Invalid literal value, expected "positions"',
-          },
-        ]),
-      );
+      expect(result.success && result.data.type).toBe('unknown');
+    });
+
+    it('should fallback to "unknown" if type is not defined', () => {
+      const zerionBalance = zerionBalanceBuilder().build();
+      // @ts-expect-error - type is expected to be a 'positions' literal
+      delete zerionBalance['type'];
+
+      const result = ZerionBalanceSchema.safeParse(zerionBalance);
+
+      expect(result.success && result.data.type).toBe('unknown');
     });
 
     it('should not allow an invalid id value', () => {
@@ -98,7 +99,7 @@ describe('Zerion Balance Entity schemas', () => {
       );
     });
 
-    it.each(['type' as const, 'id' as const, 'attributes' as const])(
+    it.each(['id' as const, 'attributes' as const])(
       'should not allow %s to be undefined',
       (key) => {
         const zerionBalance = zerionBalanceBuilder().build();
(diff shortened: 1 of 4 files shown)


src/datasources/balances-api/entities/zerion-balance.entity.ts
```diff
@@ -19,41 +19,41 @@ export type ZerionBalance = z.infer<typeof ZerionBalanceSchema>;
 
 export type ZerionBalances = z.infer<typeof ZerionBalancesSchema>;
 
-const ZerionImplementationSchema = z.object({
+export const ZerionImplementationSchema = z.object({
```
hectorgomezv, line 22: The schemas in this file are exported to test them on the ad hoc tests.

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -222,9 +222,9 @@ export class ZerionBalancesApi implements IBalancesApi {
           throw Error(
             `Zerion error: ${chainName} implementation not found for balance ${zb.id}`,
           );
-        const { value } = zb.attributes;
+        const { value, price } = zb.attributes;
         const fiatBalance = value ? getNumberString(value) : null;
-        const fiatConversion = getNumberString(zb.attributes.price);
+        const fiatConversion = price ? getNumberString(price) : null;
```
hectorgomezv, line 227: `attributes.price` is not marked as required in the [Zerion API docs](https://developers.zerion.io/reference/listwalletpositions), so I've changed this to reflect the possibility of the field being `undefined`.

src/datasources/balances-api/entities/zerion-balance.entity.spec.ts
```diff
@@ -0,0 +1,557 @@
...
+
+      const result = ZerionBalanceSchema.safeParse(zerionBalance);
+
+      expect(result.success).toBe(true);
+    });
+
+    it('should not allow an invalid type value', () => {
+      const zerionBalance = zerionBalanceBuilder().build();
+      // @ts-expect-error - type is expected to be a 'positions' literal
+      zerionBalance.type = 'invalid';
```
iamacook, line 64: What do you think about adding an `unknown` literal here if it throws?

src/datasources/balances-api/entities/zerion-balance.entity.ts
```diff
@@ -19,41 +19,41 @@ export type ZerionBalance = z.infer<typeof ZerionBalanceSchema>;
...
   chain_id: z.string(),
-  address: z.string().nullable(),
+  address: z.string().nullish().default(null),
```
iamacook, line 24: We could use `AddressSchema` here (and add a test for checksumming).

src/datasources/balances-api/entities/zerion-balance.entity.spec.ts
```diff
@@ -0,0 +1,557 @@
...
+            message: 'Invalid literal value, expected "positions"',
+          },
+        ]),
+      );
+    });
+
+    it('should not allow an invalid id value', () => {
+      const zerionBalance = zerionBalanceBuilder().build();
+      // @ts-expect-error - id is expected to be a string
+      zerionBalance.id = faker.number.int();
```
iamacook, line 84: I was initially going to suggest we move this into the builder but this is a much more future-proof method of overwriting the values that we should consider doing elsewhere in schema tests - if the type changes to match that expected, TypeScript will complain. (If we casted it inside the `with` of a builder, it wouldn't.)

hectorgomezv: It makes perfect sense to me! 👍🏻

hectorgomezv: It makes perfect sense to me! 👍🏻

src/datasources/balances-api/entities/zerion-balance.entity.spec.ts
hectorgomezv: Added in [efadc24](https://github.com/safe-global/safe-client-gateway/pull/1637/commits/efadc242b984be3bbd2955dff58249aaf050779c)

iamacook (approved):

PR #1642, opened 2024-06-12
Check INFURA_API_KEY value at startup

hectorgomezv:
## Changes
- Adds `INFURA_API_KEY` to the sample `.env` file.
- Adds `INFURA_API_KEY` to `docker-compose.yml` file.
- Adds `INFURA_API_KEY` to the startup environment validator.

### DIFF
--- a/src/datasources/blockchain/blockchain-api.manager.ts
+++ b/src/datasources/blockchain/blockchain-api.manager.ts
@@ -59,6 +59,7 @@ export class BlockchainApiManager implements IBlockchainApiManager {
     };
   }
 
+  // Note: this assumes Infura as provider when using an API key as authentication method.
   private formatRpcUri(rpcUri: DomainChain['rpcUri']): string {
     return rpcUri.authentication === RpcUriAuthentication.ApiKeyPath
       ? rpcUri.value + this.infuraApiKey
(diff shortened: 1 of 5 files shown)


iamacook (approved):

src/datasources/blockchain/blockchain-api.manager.ts
```diff
@@ -59,6 +59,7 @@ export class BlockchainApiManager implements IBlockchainApiManager {
     };
   }
 
+  // Note: this assumes Infura as provider when using an API key as authentication method.
```
hectorgomezv, line 62: Let me know if you think this note is redundant @iamacook, but I think we have to be aware that if a `Chain` is configured with a different RPC provider, this function will try to use `INFURA_API_KEY` with the other provider URL, right?

iamacook: It will but I think it's fine. It's the same implementation as the Web.

We could alternatively check that the value contains `"infura"` in it. What do you think?

hectorgomezv: I'm totally OK with checking the URL, will do it separately 👍🏻

PR #1643, opened 2024-06-12
Improve `postgres-shift` types

iamacook:
## Summary

The package we use for migrations (`postgres-shift`) has no types. As such, we have a custom defintion file for it that is not "complete". This improves the definition to match [the integration](https://github.com/porsager/postgres-shift/blob/e7a4cb16d66c33c73f4f730b7d1de33b7757fea9/index.js).

## Changes

- Add documentation to definition
- Add types for `before`/`after` callbacks


iamacook:
I'm going to close these as we are likely migrating the code to our codebase for testing reasons.

PR #1652, opened 2024-06-14
Remove `Account` implementation and integration

iamacook:
## Summary

The current account implementation was built according to email notifications. As we are now putting more focus on accounts as a fundamental feature, we are reassessing the implementation.

This removes the current account/subscription integration, as well as their integration with alerts and emails. 

## Changes

- Remove account datasource
  - Remove test datasource from tests
- Remove account domain
- Remove subscription domain
- Decouple account/subscription integration from alerts:
  - Remove business logic
  - Remove redundant code
  - Remove relative test logic and skip those eventually required


src/config/entities/configuration.ts
```diff
@@ -135,20 +135,21 @@ export default () => ({
...
-    verificationCode: {
-      resendLockWindowMs: parseInt(
-        process.env.EMAIL_VERIFICATION_CODE_RESEND_LOCK_WINDOW_MS ??
-          `${30 * 1000}`,
-      ),
-      ttlMs: parseInt(
-        process.env.EMAIL_VERIFICATION_CODE_TTL_MS ?? `${5 * 60 * 1000}`,
-      ),
-    },
+    // TODO: Decide whether we reuse these when implementing the email service
```
hectorgomezv, line 138: Nitpick: not an strong opinion, but I'd more in favour of deleting this code instead of commenting it. We can back to the commits history if we need to reuse this. Wdyt?

iamacook: Removed in 7e7818d428d0dd2c3ca865b73be068cdb2d517e5.

hectorgomezv (approved):

hectorgomezv (approved):

PR #1654, opened 2024-06-14
Add `AccountsDatasource`

iamacook:
## Summary

This adds a new datasource for accounts, implementing a basic database with tables for `accounts` and `groups:

- `accounts` are address specific and containt an `id`, `group_id` and `address` column
- `groups` are will be for defining "shared" accounts, currently only having an `id`

Leveraging the database is the `IAccountsDatasource`, containing two methods: `createAccount` and `getAccount`.

## Changes

- Create `00001_accounts` migration to create database
- Create `Account` entity with test coverage
- Create `Group` entity with test coverage
- Create `IAccountsDatasource` with implementation and test override

### DIFF
--- a/migrations/00001_accounts/index.sql
+++ b/migrations/00001_accounts/index.sql
@@ -1,0 +1,13 @@
+DROP TABLE IF EXISTS groups,
+accounts CASCADE;
+
+CREATE TABLE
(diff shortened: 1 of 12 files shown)


iamacook:
We're going to merge this into #1652 before `main` in order to not create the previous tables.

PR #1655, opened 2024-06-14
Add `PostgresDatabaseMigrator` for (testing) migration

iamacook:
## Summary

We are relying of `postgres-shift` for migrations which we have to patch and also has limitations when trying to test using it (all migrations occur in a single transaction).

This adds a new `PostgresDatabaseMigrator` that has the core logic of `postgres-shift`, as well as a testing method. The `migrate` method mirrors `postgres-shift` and the `test` method reuses part of it, executing each migration in a separate transaction. It allows us to interact with the database _before_ and _after_ a migration has executed, stopping after the desired migration.

## Changes

- Create `PostgresDatabaseMigrator` and inject it
- Remove `postgres-shift` and associated definition/patch, replacing usage with the above


src/__tests__/migration-tester.ts
```diff
@@ -0,0 +1,108 @@
+// Note: the is a helper function to test migrations by postgres-shift.
+// @see https://github.com/porsager/postgres-shift/blob/e7a4cb16d66c33c73f4f730b7d1de33b7757fea9/index.js
```
iamacook, line 2: I did consider moving the code of the package into our codebase but this can be done in a separate PR. I have a local implementation stashed that migrates it and this test function into a singular class. What's our opinion? Should we opt for that instead?

package.json
```diff
@@ -82,7 +81,6 @@
       "json",
       "ts"
     ],
-    "rootDir": "src",
```
iamacook, line 85: This is necessary for jest to find the migration tests.

src/datasources/db/postgres-database.migrator.spec.ts
```diff
@@ -0,0 +1,318 @@
...
+  let target: PostgresDatabaseMigrator;
+
+  beforeEach(() => {
+    sql = dbFactory();
+    target = new PostgresDatabaseMigrator(sql);
+  });
+
+  afterEach(async () => {
+    // Drop example table after each test
+    await sql`drop table if exists tests`;
```
hectorgomezv, line 60:
```suggestion
    await sql`drop table if exists test`;
```

src/datasources/db/postgres-database.migrator.ts
```diff
@@ -0,0 +1,233 @@
...
+
+      await this.sql.begin((transaction) => {
+        return this.run({ transaction, migration });
+      });
+
+      if (isMigrationBeingTested) {
+        // Exit loop after testing migration
+        break shift;
+      }
+    }
```
hectorgomezv, line 105: Nitpick: the `shift` label is not needed, right? There's just one loop, so the `break;` statement would do the job.

Anyway, if labelling makes it more readable for you I'm fine with it.

hectorgomezv (commented):
Great job! Just left tiny nits 👏🏻

src/datasources/db/postgres-database.migrator.spec.ts
iamacook: Fixed in f994465bd3345e2870f0ce090c90c6b516bbcead.

src/datasources/db/postgres-database.migrator.ts
iamacook: In f994465bd3345e2870f0ce090c90c6b516bbcead I changed it to only test up to the specified migration.

hectorgomezv (approved):

Issue #1649, opened 2024-06-13
Add decoding for TWAP orders


iamacook:
### Description

To facilitate the TWAP-relevant designs, we need to add decoding of TWAP orders. These can either be a `multiSend` call [enabling the Fallback Handler and creating an order](https://app.safe.global/transactions/tx?safe=sep:0x31eaC7F0141837B266De30f4dc9aF15629Bd5381&id=multisig_0x31eaC7F0141837B266De30f4dc9aF15629Bd5381_0x012e431bdfdb03e4eaf996061a6a7bed1f0bd8853aa8c809eceadb6dff18e856) with the `ComposableCoW` contract (`createWithContext`) or just [creating an order](https://app.safe.global/transactions/tx?safe=sep:0x31eaC7F0141837B266De30f4dc9aF15629Bd5381&id=multisig_0x31eaC7F0141837B266De30f4dc9aF15629Bd5381_0x37edda94f70f30ffd106f7014374eeb4b441ee57d65ed30901de20ea887c7780) `createWithContext`.

Note: we believe we can [decode the `createWithContext` parameters](https://github.com/cowprotocol/cow-sdk/blob/5aa61a03d2ed9921c5f95522866b2af0ceb1c24d/src/composable/orderTypes/Twap.ts#L197) to get the required information (it calls ethers' `decodeFunctionData`). (We may need to make an API call to determine the fulfilment percentage.)

### Requirements

- Ascertain the required fields for the design
- Determine the sources of the aforementioned
- Map the fields into (batched) transactions to the relevant contract, calling `createWithContext`

### Additional information

- [Designs](https://www.notion.so/safe-global/Swap-order-queue-history-decoding-0f273a0f06f2465589628c7299f44e05)
- [Notion task](https://www.notion.so/safe-global/Decode-TWAPs-on-backend-c1f1590ae747487ea43895a97eedbdb4)

PR #1660, opened 2024-06-17
Add base decoder for ComposableCoW contract

iamacook:
## Summary

As part of #1649, we need to return extra swap information for TWAP decoding, e.g. the start time, number of parts, time between parts and duration of part.

This adds a new `ComposableCowDecoder`, based on the [required elements of the CoW SDK](https://github.com/cowprotocol/cow-sdk/blob/5aa61a03d2ed9921c5f95522866b2af0ceb1c24d/src/composable/orderTypes/Twap.ts) to detect and decode `createWithContext` calls. The reason for creating our own decoder is to reduce dependency reliance: the SDK has a peer dependency of ethers for example (which we don't use).

## Changes

- Add ComposableCoW ABI
- Create `ComposableCowDecoder`
- Add test coverage


src/domain/swaps/contracts/decoders/composable-cow-decoder.helper.ts
```diff
@@ -0,0 +1,269 @@
...
+}
+
+type StartTime =
+  | { startType: StartTimeValue.AT_MINING_TIME }
+  | { startType: StartTimeValue.AT_EPOCH; epoch: bigint };
+
+enum StartTimeValue {
+  AT_MINING_TIME = 'AT_MINING_TIME',
+  AT_EPOCH = 'AT_EPOCH',
+}
```
iamacook, line 269: I intend to eventually move these entities but will first wait for a pattern to develop as the relative mapper may inherit some of this.

src/domain/swaps/contracts/decoders/composable-cow-decoder.helper.ts
```diff
@@ -0,0 +1,269 @@
...
+      sellToken,
+      buyToken,
+      receiver,
+      sellAmount: partSellAmount * numberOfParts,
+      buyAmount: minPartLimit * numberOfParts,
+      startTime,
+      numberOfParts,
+      timeBetweenParts,
+      durationOfPart,
+      appData, // TODO: Decode?
```
iamacook, line 193: I'm not sure if this is yet feasible.

src/domain/swaps/contracts/decoders/composable-cow-decoder.helper.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+          partSellAmount: BigInt('213586875483862141750'),
+          receiver: '0x31eaC7F0141837B266De30f4dc9aF15629Bd5381',
+          sellToken: '0xbe72E441BF55620febc26715db68d3494213D8Cb',
+          span: BigInt('0'),
+          t: BigInt('1800'),
+          t0: BigInt('0'),
+        });
+      });
+
+      it.todo('should throw if TWAP handler is invalid');
```
iamacook, line 34: I will come back and add this coverage as it involves understanding of complex encoding.

hectorgomezv (approved):

Issue #1650, opened 2024-06-13
Add decoding for swap fulfilments


iamacook:
### Description

To facilitate the swap-relevant designs, we need to add decoding of fulfilments orders (including support for TWAPs, though that should be covered as standard". These can be determined by transfers interacting with the settlement contract.

Note: in order to retrieve the relevant information, we will need to make [API call(s)](https://docs.cow.fi/cow-protocol/reference/apis/orderbook) to get the order details (perhaps by first getting the order ID from the transaction hash, then requesting it via that).

### Requirements

- Ascertain the required fields for the design
- Determine the sources of the aforementioned
- Return the fields in transfers to the relevant contract

### Additional information

- [Designs](https://www.figma.com/design/VyA38zUPbJ2zflzCIYR6Nu/Swap?node-id=6980-41655&t=U6oIpF8zFU0IGYpy-0)
- [Notion task](https://www.notion.so/safe-global/Display-Order-Information-on-Sent-Receive-TXs-f89c62df3fee4731954b33a666410290)

PR #1675, opened 2024-06-19
Map TWAP swap (`settle`) orders

iamacook:
## Summary

Fulfilment orders of TWAP-based swaps are not currently decoded. This adds decoding of (`settle`) transactions, mapping the order to that of "standard" swaps, returning them as `SwapOrderTransactionInfo` transactions, resolving #1650.

## Changes

- Add `SwapOrderHelper['findTwapSwapOrder']` method for isolating transaction data to decode
- Add `SwapOrderMapper['mapTwapSwapOrder']` method for mapping TWAP-based swaps
- Integrate the aforementioned in `TransactionInfoMapper`


src/routes/transactions/entities/confirmation-view/confirmation-view.entity.ts
```diff
@@ -92,7 +92,7 @@ export class CowSwapConfirmationView implements Baseline, OrderInfo {
     type: String,
     description: 'The URL to the explorer page of the order',
   })
-  explorerUrl: URL;
+  explorerUrl: string;
```
iamacook, line 95: This change was necessary for tests to pass. This is already returned as a string by the controller but within the mapper it was a `URL` instance.

hectorgomezv (approved):

PR #1678, opened 2024-06-20
change orderStatus to status and export type to openAPI

compojoom:
## Summary
I'm changing the orderStatus prop on the TwapOrder to just status. This way it is the same as with swapOrder.

Also I'm exporting the TwapOrderTransactionInfo to openAPI.


src/routes/transactions/entities/transaction.entity.ts
```diff
@@ -39,6 +40,7 @@ export class Transaction {
       { $ref: getSchemaPath(CustomTransactionInfo) },
       { $ref: getSchemaPath(SettingsChangeTransaction) },
       { $ref: getSchemaPath(SwapOrderTransactionInfo) },
+      { $ref: getSchemaPath(TwapOrderTransactionInfo) },
```
hectorgomezv, line 43: `TwapOrderTransactionInfo` needs to be included in the `@ApiExtraModels` as well in order to be available within the class.

```
@ApiExtraModels(
  CreationTransactionInfo,
  CustomTransactionInfo,
  SettingsChangeTransaction,
  TransferTransactionInfo,
  ModuleExecutionInfo,
  MultisigExecutionInfo,
  SwapOrderTransactionInfo,
  TwapOrderTransactionInfo,
)
```

hectorgomezv (changes requested):
Great job! I've left just one small amendment.

src/routes/transactions/entities/transaction.entity.ts
compojoom: oh, thanks! Missed that.

hectorgomezv (approved):

Issue #1651, opened 2024-06-13
Extend confirmation view to support TWAPs


iamacook:
### Description

To facilitate TWAP-relevant designs, we need to add further information to the confirmation view. This can be determined by interaction with `createWithContext` on the `ComposableCoW` contract.

Note: we can likely use a similar approach to #1649 for retrieving the require data.

### Requirements

- Ascertain the required fields for the design
- Determine the sources of the aforementioned
- Return the fields in transfers to the relevant contract

### Additional information

- [Designs](https://www.figma.com/design/VyA38zUPbJ2zflzCIYR6Nu/Swap?node-id=6655-24108&t=ugU6GKAR0YVLmbBQ-0)
- [Notion task](https://www.notion.so/safe-global/Order-Confirm-Transaction-decoding-2d167e2db91a4ebaacd6bd26bf5a579e?pvs=25)

PR #1681, opened 2024-06-24
Add TWAP decoding to confirmation view

iamacook:
## Summary

Our confirmation view currently decodes "standard" swaps. This adds the relevant decoded to return the `TwapOrderInfo` from the confirmation view endpoint, resolving #1651.

## Changes

- Add `CowSwapTwapConfirmationView` with `COW_SWAP_TWAP_ORDER` `type` 
- Change types of `TwapOrderInfo` to ensure no issue with max. integers/time consistency
  - `numberOfParts`: number -> string
  - `timeBetweenParts`: string -> number
- Add `TwapOrderHelper['twapStructToPartialOrderInfo']` to reduce duplicate code
- Add TWAP decoding to `TransactionsViewService`


src/routes/transactions/mappers/common/swap-order.mapper.ts
```diff
@@ -14,6 +14,8 @@ export class SwapOrderMapper {
     private readonly swapOrderHelper: SwapOrderHelper,
   ) {}
 
+  // TODO: Handling of restricted Apps of TWAP mapping
```
iamacook, line 17: These will be handled in a separate PR as it is out of scope here.

PR #1691, opened 2024-06-25
Map swap-related transfers

iamacook:
## Summary

Incoming/outgoing transfers also cover `transferFrom` transactions. Swaps execute such transfers to/from the GPv2Settlement contract. Currently the decoding of these transactions are "standard".

This maps the relative order information to swap-related transfers if the sender/recipient address is the GPv2Settlement contract.

## Changes

The following are predominantly code migrations of existing code as transfer mapping was assumed to happen with the `TransactionInfoMapper`, whereas it should in the `TransferInfoMapper`:

- Create new `SwapTransferTransactionInfo`
- Create new `SwapTransferHelper`
- Create new `SwapTransferInfoMapper` and include it in `TransferInfoMapper`
- Adjust the relative entities


src/routes/transactions/mappers/transfers/transfer-info.mapper.ts
```diff
@@ -11,37 +11,67 @@ import { Erc721Transfer } from '@/routes/transactions/entities/transfers/erc721-
...
+          chainId,
+          safeAddress: safe.address,
+          domainTransfer,
+        });
+
+      if (swapTransfer) {
+        return swapTransfer;
+      }
+    }
+
```
hectorgomezv, line 69: This way (if the FFs are enabled), there is always a try to map the transfer as a `SwapTransferTransactionInfo`. 

Alternatively, we could expose an `isSwapTransferInfo` method so we can integrate the matching/mapping logic into `getTransferByType`. That way we could check if the transfer is actually a `SwapTransferTransactionInfo` and then call `mapSwapTransferInfo` inside `getTransferByType` as we do to map the rest of the transfer types. Wdyt?

src/routes/transactions/mappers/transfers/swap-transfer-info.mapper.spec.ts
```diff
@@ -0,0 +1,399 @@
...
+      token.symbol,
+      token.logoUri,
+      token.decimals,
+      token.trusted,
+    );
+    const order = orderBuilder()
+      .with('from', getAddress(sender.value))
+      .with('owner', safeAddress)
+      .with('buyToken', token.address)
+      .with('executedBuyAmount', BigInt(domainTransfer.value))
```
hectorgomezv, line 110: In these tests, I've just adjusted the `Order` payloads to match the logic implemented in `SwapTransferInfoMapper.findOrderByTransfer` not to filter out the order, i.e.: `executedBuyAmount` needs to match the transfer value, and `buyToken` address needs to match the token address. I'm assuming this logic is correct. cc. @iamacook @compojoom

hectorgomezv (approved):

PR #1694, opened 2024-06-26
Implement DELETE /v1/accounts/:address

hectorgomezv:
## Summary
This PR adds the endpoint:
```
DELETE /v1/accounts/:address
```

- The endpoint expects an `address` in the request path, targeting the Ethereum address associated with the account.
- The endpoint is authenticated using SIWE, so an `accessToken` associated with the address needs to be provided. Otherwise, a `403` status will be returned.

## Changes
- The endpoint `DELETE /v1/accounts/:address` was added to `AccountsController`, protected by `AuthGuard`.
- Associated tests and domain classes were added.

### DIFF
--- a/src/routes/accounts/accounts.controller.ts
+++ b/src/routes/accounts/accounts.controller.ts
@@ -3,12 +3,15 @@ import { Account } from '@/routes/accounts/entities/account.entity';
 import { CreateAccountDto } from '@/routes/accounts/entities/create-account.dto.entity';
 import { CreateAccountDtoSchema } from '@/routes/accounts/entities/schemas/create-account.dto.schema';
 import { AuthGuard } from '@/routes/auth/guards/auth.guard';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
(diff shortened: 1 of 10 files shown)


src/routes/accounts/accounts.controller.ts
```diff
@@ -33,4 +36,15 @@ export class AccountsController {
...
   }
+
+  @Delete(':address')
+  @UseGuards(AuthGuard)
+  @HttpCode(HttpStatus.NO_CONTENT)
+  async deleteAccount(
+    @Param('address', new ValidationPipe(AddressSchema)) address: `0x${string}`,
+    @Req() request: Request,
+  ): Promise<void> {
+    const auth = request.accessToken;
```
iamacook, line 47: With the `AuthGuard`, we will already have access to the `AuthPayload` here.

iamacook: I opened [a PR](https://github.com/safe-global/safe-client-gateway/pull/1722) changing this.

PR #1698, opened 2024-06-27
Filter out TWAP orders from not allowed apps

hectorgomezv:
## Summary
This PR implements `TwapOrderHelper.isAppAllowed`, which prevents the decoding of TWAP orders emitted by non-allowed apps.

## Changes
- Add `TwapOrderHelper.isAppAllowed`.
- Add test cases verifying the non-decoding of these orders.


src/routes/transactions/transactions-view.service.ts
```diff
@@ -172,9 +172,17 @@ export class TransactionsViewService {
...
     });
 
-    const [{ fullAppData }, buyToken, sellToken] = await Promise.all([
-      // Decode hash of `appData`
-      this.swapsRepository.getFullAppData(args.chainId, twapStruct.appData),
+    // Decode hash of `appData`
+    const fullAppData = await this.swapsRepository.getFullAppData(
+      args.chainId,
+      twapStruct.appData,
+    );
```
hectorgomezv, line 179: I've decupled this block from the tokens retrieval, so we can fail-fast if the app is not allowed.

src/routes/transactions/helpers/twap-order.helper.ts
```diff
@@ -167,6 +177,20 @@ export class TwapOrderHelper {
     });
   }
 
+  /**
+   * Checks if the app associated contained in fullAppData is allowed.
+   *
+   * @param fullAppData - object to which we should verify the app data with
+   * @returns true if the app is allowed, false otherwise.
+   */
+  public isAppAllowed(fullAppData: FullAppData): boolean {
```
iamacook, line 186: This duplicates that in the `SwapOrderHelper`. Maybe we should extract it somehow - what do you think? (I created a `disallow-twap-mapping` branch before going away that might be possible to reference.)

iamacook (commented):
We also need to add this to `SwapTransferInfoMapper['mapSwapTransferInfo']` after finding the order by transfer.

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -111,7 +111,10 @@ export class TwapOrderMapper {
       }
     }
 
-    // TODO: Handling of restricted Apps, calling `getToken` directly instead of multiple times in `getOrder` for sellToken and buyToken
+    if (!this.twapOrderHelper.isAppAllowed(fullAppData)) {
+      throw new Error(`Unsupported App: ${fullAppData.fullAppData?.appCode}`);
+    }
```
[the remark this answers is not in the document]
hectorgomezv: Added in [6c540a0](https://github.com/safe-global/safe-client-gateway/pull/1698/commits/6c540a0e504a2a4784c74187530848723f076e26)

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -101,6 +101,16 @@ export class TwapOrderMapper {
...
       try {
         const order = await this.swapsRepository.getOrder(chainId, orderUid);
+        const partFullAppData = await this.swapsRepository.getFullAppData(
+          chainId,
+          part.appData,
+        );
+        if (!this.twapOrderHelper.isAppAllowed(order)) {
+          throw new Error(
+            `Unsupported App: ${partFullAppData.fullAppData?.appCode}`,
+          );
```
iamacook, line 111:
```suggestion
        if (!this.twapOrderHelper.isAppAllowed(order)) {
          throw new Error(
            `Unsupported App: ${order.fullAppData?.appCode}`,
          );
```

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -85,7 +85,7 @@ export class TwapOrderMapper {
         : twapParts
       : [];
 
-    const { fullAppData } = await this.swapsRepository.getFullAppData(
+    const fullAppData = await this.swapsRepository.getFullAppData(
```
iamacook, line 88: We can move this below the checking of each part to reude request.

iamacook (commented):
I think we also need check the `appData` of [the "found" order](https://github.com/safe-global/safe-client-gateway/blob/restrict-twap-apps/src/routes/transactions/mappers/transfers/swap-transfer-info.mapper.ts#L56) when mapping swap transfers as well. We will otherwise map all swap transfers.

iamacook (approved):

PR #1703, opened 2024-06-28
Fix getNumberString utils function

### LINKED ISSUE #1701: Flaky test while testing balances fiatTotal values
### Description

The following difference happens sometimes when testing `/balances` endpoint responses.
```
 -     "fiatTotal": "4910.25",
 +     "fiatTotal": "NaN",
```

It is pending to determine/research if this is linked to a specific NodeJS version or other environment attribute.

`getNumberString` function could be involved in the issue. It may also be related to numeric string formatting (`.` vs `,`).

Screenshot 2024-06-28 at 12 35 03

hectorgomezv:
Closes #1701 

## Changes
- Changes `getNumberString` implementation to use a fixed locale, standard notation, and no grouping.

### DIFF
--- a/src/domain/common/utils/utils.ts
+++ b/src/domain/common/utils/utils.ts
@@ -1,6 +1,7 @@
 export function getNumberString(value: number): string {
   // Prevent scientific notation
-  return value.toLocaleString('fullwide', {
+  return value.toLocaleString('en-US', {
+    notation: 'standard',
     useGrouping: false,
   });
 }


src/domain/common/utils/utils.ts
```diff
@@ -1,6 +1,7 @@
 export function getNumberString(value: number): string {
   // Prevent scientific notation
-  return value.toLocaleString('fullwide', {
```
hectorgomezv, line 3: `'fullwide'` is not a standard locale identifier. This seems to be a bug in the original implementation, which JS is probably ignoring and therefore defaulting to the host locale.

PR #1705, opened 2024-06-28
Adjust access token cookie configuration

hectorgomezv:
## Summary
This change changes the configuration of the `accessToken` cookie, by setting `SameSite=Lax` when `CGW_ENV=production`, and `SameSite=None` otherwise.

## Changes
- Changes `AuthController.verify` function to set an authentication cookie with `SameSite=Lax` or `SameSite=None` depending on `CGW_ENV`.

### DIFF
--- a/src/routes/auth/auth.controller.ts
+++ b/src/routes/auth/auth.controller.ts
@@ -1,10 +1,19 @@
-import { Body, Controller, Get, Post, HttpCode, Res } from '@nestjs/common';
-import { ApiExcludeController } from '@nestjs/swagger';
-import { ValidationPipe } from '@/validation/pipes/validation.pipe';
-import { AuthService } from '@/routes/auth/auth.service';
-import { SiweDtoSchema, SiweDto } from '@/routes/auth/entities/siwe.dto.entity';
-import { Response } from 'express';
+import { IConfigurationService } from '@/config/configuration.service.interface';
 import { getMillisecondsUntil } from '@/domain/common/utils/time';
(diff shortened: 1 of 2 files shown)


src/routes/auth/auth.controller.ts
```diff
@@ -19,8 +28,19 @@ import { getMillisecondsUntil } from '@/domain/common/utils/time';
 @ApiExcludeController()
 export class AuthController {
   static readonly ACCESS_TOKEN_COOKIE_NAME = 'access_token';
+  static readonly ACCESS_TOKEN_COOKIE_SAME_SITE_LAX = 'lax';
+  static readonly ACCESS_TOKEN_COOKIE_SAME_SITE_NONE = 'none';
+  static readonly CGW_ENV_PRODUCTION = 'production';
+  private readonly cgwEnv: string;
```
iamacook, line 34: What do you think about adding an `isProduction` env. var. as well?

hectorgomezv: Do you mean instead of this one, or in addition to it? (I'm not against it, just aligning regarding the use cases it would have)

iamacook: I think the ternary comparison could be done against this value _in_ the configuration, e.g. there is an `application.isProduction` value.

hectorgomezv: It makes perfect sense, thanks! I've created a PR: https://github.com/safe-global/safe-client-gateway/pull/1724

PR #1715, opened 2024-07-02
Implement PUT /v1/accounts/:address/data-settings

hectorgomezv:
Depends on #1735 

## Summary
This PR adds the endpoint:
```
PUT /v1/accounts/:address/data-settings
```

- The endpoint expects a `UpsertAccountDataSettingsDto` request body with the following payload:
```
{
  "accountDataSettings": [
    {
      "id": [string],
      "enabled": [boolean]
    },
    {
      "id": [string],
      "enabled": [boolean]
    },
    ...
}
``` 
- The endpoint is authenticated using SIWE, so an `accessToken` associated with the address included in the URL needs to be provided. Otherwise, a `403` status will be returned.

## Changes
- The endpoint `PUT /v1/accounts/:address/data-settings` was added to `AccountsController`, protected by `AuthGuard`.
- Associated tests and domain classes were added.


migrations/__tests__/00003_account-data-settings.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+          { column_name: 'account_id' },
+          { column_name: 'enabled' },
+        ]),
+        rows: [],
+      },
+    });
+  });
+
+  it('should add one AccountDataSettings and update its row timestamps', async () => {
+    const accountAddress = faker.finance.ethereumAddress();
```
iamacook, line 74:
```suggestion
    const accountAddress = getAddress(faker.finance.ethereumAddress());
```

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -149,4 +150,150 @@ describe('AccountsDatasource tests', () => {
...
+            )?.id,
+            enabled: ads.enabled,
+            created_at: expect.any(Date),
+            updated_at: expect.any(Date),
+          })),
+        ),
+      );
+    });
+
+    it('updates existent account data settings successfully', async () => {
```
iamacook, line 187:
```suggestion
    it('updates existing account data settings successfully', async () => {
```

migrations/__tests__/00003_account-data-settings.spec.ts
```diff
@@ -0,0 +1,117 @@
...
+  address: `0x${string}`;
+}
+
+interface AccountDataTypeRow {
+  id: number;
+  created_at: Date;
+  updated_at: Date;
+  name: string;
+  description: string;
+  is_active: boolean;
```
iamacook, line 20: Is this a defined requirement? It's somewhat of a duplication of feature flags and the client equally defines what features they enable via the Config. Service.

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -149,4 +150,150 @@ describe('AccountsDatasource tests', () => {
...
+        dataTypeName: faker.lorem.slug(),
+        enabled: faker.datatype.boolean(),
+      });
+
+      await expect(
+        target.upsertAccountDataSettings(address, upsertAccountDataSettingsDto),
+      ).rejects.toThrow('Invalid data type.');
+    });
+
+    it('throws an error if an inactive data type is provided', async () => {
```
iamacook, line 282: This extends upon my other comment regarding this flag. Do we want to prevent the client from "registering" for settings outright? We need to consider _how_ this is going to be toggled. It's somewhat of an admin-related flag.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -64,4 +66,50 @@ export class AccountsDatasource implements IAccountsDatasource {
...
+    for (const ads of accountDataSettings) {
+      const dataType = dataTypes.find((dt) => dt.name === ads.dataTypeName);
+      if (!dataType) {
+        throw new UnprocessableEntityException('Invalid data type.');
+      }
+      if (!dataType.is_active) {
+        throw new UnprocessableEntityException(
+          `Data type ${dataType.name} is inactive.`,
+        );
+      }
```
iamacook, line 99: Relative to my aforementioned questions regarding this property.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -64,4 +66,50 @@ export class AccountsDatasource implements IAccountsDatasource {
...
+      const [inserted] = await this.sql<[AccountDataSetting]>`
+        INSERT INTO account_data_settings (account_id, account_data_type_id, enabled)
+        VALUES (${account.id}, ${dataType.id}, ${ads.enabled})
+        ON CONFLICT (account_id, account_data_type_id) DO UPDATE SET enabled = EXCLUDED.enabled
+        RETURNING *
+      `.catch((e) => {
+        throw new UnprocessableEntityException(
+          `Error updating data settings: ${asError(e).message}`,
+        );
+      });
```
iamacook, line 109: Would there be any benefit in doing these in one transaction?

src/domain/accounts/entities/upsert-account-data-settings.dto.entity.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from 'zod';
+
+export class UpsertAccountDataSettingsDto
+  implements z.infer<typeof UpsertAccountDataSettingsDtoSchema>
+{
+  accountDataSettings: {
+    dataTypeName: string;
```
iamacook, line 7: Will the client retrieve these "data types" from us? If so, it might be better to upsert them via ID. What do you think?

A small nit: I don't think we need prefix this with "dataType-" either.

src/routes/accounts/accounts.controller.spec.ts
```diff
@@ -584,4 +586,183 @@ describe('AccountsController', () => {
...
+      const accessToken = jwtService.sign(authPayloadDto);
+
+      await request(app.getHttpServer())
+        .put(`/v1/accounts/${address}/data-settings`)
+        .set('Cookie', [`access_token=${accessToken}`])
+        .send({ address })
+        .expect(403);
+    });
+
+    it('should propagate errors', async () => {
```
iamacook, line 745: As this is an auth-based endpoint, should we propagate errors?

src/routes/accounts/accounts.service.ts
```diff
@@ -69,4 +94,23 @@ export class AccountsService {
...
+      (dt) => dt.id === domainAccountDataSetting.account_data_type_id,
+    );
+
+    if (!dataType) {
+      throw new Error('Data type not found');
+    }
+
+    return {
+      dataTypeName: dataType.name,
+      dataTypeDescription: dataType.description,
```
iamacook, line 112: As before regarding the inclusion of an `id` and not prefixing with "dataType-".

migrations/00003_account-data-settings/index.sql
```diff
@@ -0,0 +1,15 @@
+DROP TABLE IF EXISTS account_data_settings CASCADE;
+
+CREATE TABLE account_data_settings (
+    account_id INTEGER NOT NULL,
+    account_data_type_id INTEGER NOT NULL,
```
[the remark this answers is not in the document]
hectorgomezv: If I understand correctly, setting them as primary key makes each combination both unique and not-null automatically (https://www.postgresql.org/docs/current/ddl-constraints.html#DDL-CONSTRAINTS-PRIMARY-KEYS)

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: (Related to https://github.com/safe-global/safe-client-gateway/pull/1715#discussion_r1668434862)

In my opinion and considering the changes in the data types will be likely uncommon, I'd opt for doing this via database migrations. And I'd include a new FF in the service `FF_RUN_MIGRATIONS`, so it would give us some extra control regarding that.

src/domain/accounts/entities/upsert-account-data-settings.dto.entity.ts
hectorgomezv: Initially, I thought about using ids instead of names for this payload. I opted for the names to make it more meaningful for the clients, but I agree it's simpler to implement this endpoint if we change back to ids. Wdyt?

src/routes/accounts/accounts.controller.spec.ts
hectorgomezv: Good point! The test description was wrong, the endpoint does not actually _propagate_ the error. It just sends `Internal server error` message on uncaught errors. I'm checking this in [6df2749](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/6df2749e90de12dcfb479e5f54d9520e5847c2ad)

src/routes/accounts/accounts.service.ts
hectorgomezv: Thanks! Will refactor this based on the outcome of [the other comment](https://github.com/safe-global/safe-client-gateway/pull/1715#discussion_r1668444149) 🙂

src/datasources/accounts/accounts.datasource.ts
hectorgomezv: The transaction was included in [76e14ea](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/76e14ea2747ff81e473d2b4e1916e85619a06c04)

migrations/__tests__/00003_account-data-settings.spec.ts
iamacook: I am happy with the justification but my main concern was the manual step required to (de-)activate a data type but [you gave clarity to that](https://github.com/safe-global/safe-client-gateway/pull/1715/files#r1668439084). I'm happy to stay with this implementation and we can iterate on it later if necessary.

src/routes/accounts/accounts.controller.ts
```diff
@@ -47,6 +51,22 @@ export class AccountsController {
     return this.accountsService.getDataTypes();
   }
 
+  @ApiOkResponse({ type: AccountDataSetting, isArray: true })
+  @Put(':address/data-settings')
+  @UseGuards(AuthGuard)
+  async upsertAccountDataSettings(
+    @Auth() authPayload: AuthPayload,
+    @Param('address', new ValidationPipe(AddressSchema)) address: `0x${string}`,
```
iamacook, line 59: We should rely on the authenticated `AuthPayload['address']` here. Otherwise, (as long as I'm authenticated) I could change the settings of any `address` here.

src/domain/accounts/entities/__tests__/upsert-account-data-settings.dto.entity.builder.ts
```diff
@@ -0,0 +1,13 @@
+import { IBuilder, Builder } from '@/__tests__/builder';
+import { UpsertAccountDataSettingsDto } from '@/domain/accounts/entities/upsert-account-data-settings.dto.entity';
+import { faker } from '@faker-js/faker';
+
+export function upsertAccountDataSettingsDtoBuilder(): IBuilder<UpsertAccountDataSettingsDto> {
+  return new Builder<UpsertAccountDataSettingsDto>().with(
+    'accountDataSettings',
+    Array.from({ length: faker.number.int({ min: 1, max: 10 }) }, () => ({
+      dataTypeName: faker.lorem.word(),
```
iamacook, line 9: Do you think we also need prefix this with "dataType-"?

src/domain/accounts/entities/upsert-account-data-settings.dto.entity.ts
iamacook: I think IDs would make sense as they remain static whereas names could be subject to change. I'm happy to merge this PR and change it in a follow-up though. (Not sure if it'd cause issues due to migrations though.)

src/domain/accounts/entities/upsert-account-data-settings.dto.entity.ts
```diff
@@ -0,0 +1,23 @@
...
+    enabled: boolean;
+  }[];
+
+  constructor(props: UpsertAccountDataSettingsDto) {
+    this.accountDataSettings = props.accountDataSettings;
+  }
+}
+
+export const UpsertAccountDataSettingDtoSchema = z.object({
+  dataTypeName: z.string(),
```
iamacook, line 17: As before regarding prefix.

src/routes/accounts/accounts.service.ts
```diff
@@ -53,6 +56,28 @@ export class AccountsService {
...
+      await this.accountsRepository.upsertAccountDataSettings({
+        authPayload: args.authPayload,
+        address: args.address,
+        upsertAccountDataSettings: {
+          accountDataSettings:
+            args.upsertAccountDataSettingsDto.accountDataSettings,
+        },
+      });
+
+    const dataTypes = await this.accountsRepository.getDataTypes();
```
iamacook, line 74: We could wrap these in a `Promise.all`.

src/domain/accounts/entities/upsert-account-data-settings.dto.entity.ts
hectorgomezv: IDs make total sense to me also, they were my first option as well, but later I took this approach to make it easier for the clients. But giving it a second thought, I agree it makes sense to have IDs instead of names to stay static. I will change it in this PR! Thanks! 😄

src/routes/accounts/accounts.controller.ts
hectorgomezv: As we agreed offline, we'll keep it as it is, as this address/auth alignment is been checked [here](https://github.com/safe-global/safe-client-gateway/blob/72449d149fa12382cd4c8c19416afbaaf519ddcc/src/domain/accounts/accounts.repository.ts).

src/routes/accounts/accounts.service.ts
hectorgomezv: Done in [2a6e988](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/2a6e988cbd0a6970b9aff2d1efd9b28d43342955)

src/domain/accounts/entities/upsert-account-data-settings.dto.entity.ts
hectorgomezv: Changed in [2a6e988](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/2a6e988cbd0a6970b9aff2d1efd9b28d43342955)

src/domain/accounts/entities/__tests__/upsert-account-data-settings.dto.entity.builder.ts
hectorgomezv: Done in dataTypeName

src/routes/accounts/entities/account-data-setting.entity.ts
```diff
@@ -0,0 +1,20 @@
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+
+export class AccountDataSetting {
+  @ApiProperty()
+  dataTypeName: string;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  dataTypeDescription: string | null;
```
iamacook, line 7: As above regarding the "dataType-" prefix.

src/routes/accounts/entities/upsert-account-data-settings.dto.entity.ts
```diff
@@ -0,0 +1,16 @@
+import { UpsertAccountDataSettingsDto as DomainUpsertAccountDataSettingsDto } from '@/domain/accounts/entities/upsert-account-data-settings.dto.entity';
+import { ApiProperty } from '@nestjs/swagger';
+
+class UpsertAccountDataSettingDto {
+  @ApiProperty()
+  id!: string;
```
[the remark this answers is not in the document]
iamacook: It'd be good to add a brief comment regarding this in the relevant places for later as well but it otherwise looks good to me!

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -62,6 +64,76 @@ export class AccountsDatasource implements IAccountsDatasource {
...
+            VALUES (${account.id}, ${accountDataSetting.id}, ${accountDataSetting.enabled})
+            ON CONFLICT (account_id, account_data_type_id) DO UPDATE SET enabled = EXCLUDED.enabled
+          `.catch((e) => {
+            throw new UnprocessableEntityException(
+              `Error updating data settings: ${asError(e).message}`,
+            );
+          });
+        }),
+      );
+    });
```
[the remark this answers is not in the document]
hectorgomezv: Good point! I've included the final selection in the transaction in [7840a53](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/7840a53a7fbd14e94d4b9428a7eaaec90bc7c712). I needed to replace the call to the class function `getAccountDataSettings` with the SQL query:

```
return sql<[AccountDataSetting]>`
        SELECT * FROM account_data_settings WHERE account_id = ${account.id}`;
```

This is due to the fact the function `getAccountDataSettings` uses `this.sql` instead of the `sql` instance provided inside the transaction. Otherwise, the result would not include the rows inserted inside the transaction.

This also has a positive side effect, as before by calling `getAccountDataSettings` it was executing an unnecessary `getAccount(address)` call.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -62,6 +64,76 @@ export class AccountsDatasource implements IAccountsDatasource {
...
+  }
+
+  private async checkDataTypes(
+    accountDataSettings: UpsertAccountDataSettingsDto['accountDataSettings'],
+  ): Promise<void> {
+    const activeDataTypes = await this.getActiveDataTypes();
+    const activeDataTypeIds = activeDataTypes.map((ads) => ads.id);
+    if (
+      !accountDataSettings
+        .map((ads) => ads.id)
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [7840a53](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/7840a53a7fbd14e94d4b9428a7eaaec90bc7c712)

src/routes/accounts/entities/account-data-setting.entity.ts
hectorgomezv: Changed in [7840a53](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/7840a53a7fbd14e94d4b9428a7eaaec90bc7c712)

src/routes/accounts/entities/upsert-account-data-settings.dto.entity.ts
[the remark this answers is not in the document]
hectorgomezv: Added in [7840a53](https://github.com/safe-global/safe-client-gateway/pull/1715/commits/7840a53a7fbd14e94d4b9428a7eaaec90bc7c712)

iamacook (approved):
Amazing work! 🚀🚀🚀

hectorgomezv:
- The endpoint PUT /v1/accounts/:address/data-settings was added to AccountsController, protected by AuthGuard.
- Associated tests and domain classes were added.

PR #1722, opened 2024-07-04
Use `AuthDecorator` for instantiation of account-related `AuthPayload`s

iamacook:
## Summary

When using the `AuthGuard`, we can be sure that the `accessToken` is present in the `request`. Currently, our account-related methods instantiate the relative `AuthPayload` individually. There is, however, a specific (`Auth`) decorator that handles instantiation.

This replaces the aforementioned for the creation, retrieval and deletion of accounts.

## Changes

- Expect an `AuthPayload` instead of `AuthPayloadDto` in `AccountsController`, `AccountsService` and `AccountsRepository`.
- Use `AuthPayload` directly for authorisation in `AccountsRepository`,


hectorgomezv (approved):
👏🏻

PR #1729, opened 2024-07-04
Remove redundant TODO comment

iamacook:
## Summary

Calling `ISwapsRepository['getOrder']` used to return the order as well as buy/sell tokens. Since this has been decoupled into respective methods, token details can be requested directly and are not requested when getting an order.

This removes a TODO that is/was related to the above. We are getting the token data directly after getting parts of a TWAP already.

## Changes

- Remove TODO comment


hectorgomezv (approved):

iamacook:
I'm going to close this as it was since fixed in `main`.

PR #1732, opened 2024-07-05
Only log when trying to map restricted swap transfers

iamacook:
## Summary

When mapping a swap transfer, we are currently throwing if the transfer was created by a disallowed app. This means that the mapping breaks when we should be falling back to "standard" transfer mapping.

This switches from throwing to instead logging and returning `null` instead of a mapped swap transfer. This means that the "standard" mapping will be done instead.

## Changes

- Log/return `null` instead of throwing when mapping a disallowed swap transfer
- Update test accordingly

### DIFF
--- a/src/routes/transactions/mappers/transfers/swap-transfer-info.mapper.ts
+++ b/src/routes/transactions/mappers/transfers/swap-transfer-info.mapper.ts
@@ -9,6 +9,7 @@ import { SwapTransferTransactionInfo } from '@/routes/transactions/swap-transfer
 import { getAddress, isAddressEqual } from 'viem';
 import { ISwapsRepository } from '@/domain/swaps/swaps.repository';
 import { Order } from '@/domain/swaps/entities/order.entity';
+import { LoggingService, ILoggingService } from '@/logging/logging.interface';
 
 @Injectable()
 export class SwapTransferInfoMapper {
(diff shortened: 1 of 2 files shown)


src/routes/transactions/mappers/transfers/swap-transfer-info.mapper.ts
```diff
@@ -61,7 +63,11 @@ export class SwapTransferInfoMapper {
 
     // TODO: Refactor with confirmation view, swaps and TWAPs
     if (!this.swapOrderHelper.isAppAllowed(order)) {
-      throw new Error(`Unsupported App: ${order.fullAppData?.appCode}`);
+      this.loggingService.warn(
+        `Unsupported App: ${order.fullAppData?.appCode}`,
+      );
+      // Don't throw in order to fallback to "standard" transfer mapping
+      return null;
```
hectorgomezv, line 70: Nit: while I see `null` returns in many other places related to transactions/swap/twaps mappings to indicate a failure, I also think we should refactor the `null` returns and change them for actual errors and do proper error handling in the upper layers/caller functions, wdyt? (This is just a general comment, I'm totally fine with this PR 🙂)

hectorgomezv (approved):

src/routes/transactions/mappers/transfers/swap-transfer-info.mapper.ts
iamacook: I agree. There is a common pattern with the finding then mapping of swap-related transactions that needs to be refactored in general. I think it would also solve this.

iamacook: "Proper error handling" has been included in https://github.com/safe-global/safe-client-gateway/pull/1733, but it should definitely be refactored as mentioned.

PR #1734, opened 2024-07-05
Generate specific function/event decoders in `AbiDecoder`

iamacook:
## Summary

We have a common pattern across the project when decoding function data/event logs, whereby we 1) decode 2) ensure it's the correct function/event 3) return the arguments. We also handle any thrown errors.

This refactors the `AbiDecoder` decoding logic for function data/event logs to automatically generate type safe, specific methods for each function/event, as we do with helpers.

The following example uses the ERC-20 ABI, highlighting the `transferFrom` function and `Transfer` event:

```ts
{
  helpers: {
    isTransferFrom: (data: `0x${string}`) => boolean
  },
  decodeFunctionData: {
    transferFrom: (data: `0x${string}`) => readonly [`0x${string}`, `0x${string}`, bigint] | null
  },
  decodeEventLog: {
    Transfer: (args: {
      data: `0x${string}`;
      topics: [] | [signature: `0x${string}`, ...args: `0x${string}`[]];
    }}> => {
      from: `0x${string}`;
      to: `0x${string}`;
      value: bigint;
    } | null
  }
}
```

## Changes

- Implement generators for function data/event log decoding
- Use the generated methods in place


src/domain/contracts/decoders/abi-decoder.helper.ts
```diff
@@ -1,78 +1,165 @@
...
+type FunctionDataDecoders<TAbi extends Abi> = {
+  [key in FunctionDataDecoder<TAbi>]: (
+    data: Hex,
+  ) => ContractFunctionArgs<TAbi, AbiStateMutability, key> | null;
+};
+
+export function _generateFunctionDataDecoders<TAbi extends Abi>(
+  abi: TAbi,
+): FunctionDataDecoders<TAbi> {
+  const createFunctionDataDecoder = (name: FunctionDataDecoder<TAbi>) => {
```
hectorgomezv, line 74: Nit: as on `_generateEventLogDecoders`, what do you think about making `createFunctionDataDecoder` an independent private function within the class?

src/domain/contracts/decoders/abi-decoder.helper.ts
```diff
@@ -1,78 +1,165 @@
...
+type Helper<TAbi extends Abi> = `is${Capitalize<ContractFunctionName<TAbi>>}`;
+
+type Helpers<TAbi extends Abi> = {
+  [key in Helper<TAbi>]: (data: Hex) => boolean;
+};
+
 export function _generateHelpers<TAbi extends Abi>(
   abi: Readonly<TAbi>,
 ): Helpers<TAbi> {
+  const createHelper = (name: AbiFunction) => {
```
hectorgomezv, line 42: Nit: as on `_generateEventLogDecoders`, what do you think about making `createHelper` an independent private function within the class?

src/domain/swaps/contracts/decoders/gp-v2-decoder.helper.ts
```diff
@@ -660,22 +657,13 @@ export class GPv2Decoder extends AbiDecoder<typeof GPv2Abi> {
...
-      if (decoded.functionName !== 'setPreSignature') {
-        throw new Error('Data is not of setPreSignature');
-      }
+    const decoded = this.decodeFunctionData.setPreSignature(data);
 
-      return decoded.args[0];
-    } catch (e) {
-      this.loggingService.debug(e);
+    if (!decoded) {
       return null;
```
hectorgomezv, line 663: Nit: Wdyt about logging a warning in this situation?

src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -156,39 +156,33 @@ export class LimitAddressesMapper {
...
-      }
-
-      return {
-        to: safeDecodedData.args[0],
-        value: safeDecodedData.args[1],
-        data: safeDecodedData.args[2],
-      };
-    } catch {
+    if (!safeDecodedData) {
       return null;
```
hectorgomezv, line 163: Nit: as before, should we log a warning here?

src/domain/contracts/decoders/abi-decoder.helper.ts
```diff
@@ -1,78 +1,165 @@
...
-  decodeFunctionData<TAbi extends Abi>(
-    args: Omit<DecodeFunctionDataParameters<TAbi>, 'abi'>,
-  ) {
-    return _decodeFunctionData({ ...args, abi: this.abi });
+        return decoded.args as ContractEventArgsFromTopics<
+          TAbi,
+          EventLogDecoder<TAbi>
+        >;
+      } catch {
+        return null;
```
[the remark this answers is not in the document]
iamacook: Changed in 36b45c17e2e254e6365f838b1b14e566de930810.

src/domain/contracts/decoders/abi-decoder.helper.ts
```diff
@@ -1,78 +1,165 @@
...
+    }): ContractEventArgsFromTopics<TAbi, ContractEventName<TAbi>> | null => {
+      try {
+        const decoded = decodeEventLog({
+          abi,
+          data: args.data,
+          topics: args.topics,
+        });
+
+        if (decoded.eventName !== name) {
+          return null;
```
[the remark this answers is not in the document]
iamacook: It can happen when if the function data is of a function in the ABI, but not that one expected. Changed it to log/throw in 36b45c17e2e254e6365f838b1b14e566de930810.

src/domain/contracts/decoders/abi-decoder.helper.ts
```diff
@@ -1,78 +1,165 @@
...
+  [key in EventLogDecoder<TAbi>]: (args: {
+    data: `0x${string}`;
+    topics: [signature: `0x${string}`, ...args: Array<`0x${string}`>] | [];
+  }) => ContractEventArgsFromTopics<TAbi, key> | null;
+};
+
+export function _generateEventLogDecoders<TAbi extends Abi>(
+  abi: TAbi,
+): EventLogDecoders<TAbi> {
+  const createEventLogDecoder = (name: EventLogDecoder<TAbi>) => {
```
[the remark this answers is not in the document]
iamacook: I adjusted it to use private functions instead in 36b45c17e2e254e6365f838b1b14e566de930810.

src/domain/contracts/decoders/abi-decoder.helper.ts
iamacook: I adjusted it to use private functions instead in 36b45c17e2e254e6365f838b1b14e566de930810.

src/domain/contracts/decoders/abi-decoder.helper.ts
iamacook: I adjusted it to use private functions instead in 36b45c17e2e254e6365f838b1b14e566de930810.

src/domain/swaps/contracts/decoders/gp-v2-decoder.helper.ts
iamacook: This shouldn't throw because of the initial `isSetPreSignature` so logging isn't needed.

src/domain/relay/limit-addresses.mapper.ts
iamacook: Logging now occurs if decoding fails in 36b45c17e2e254e6365f838b1b14e566de930810.

iamacook:
After discussion, we decided that handling errors relative to their respective calls is better than null-returns. Whilst this attempts to abstract away a common pattern, it goes against the aforementioned decision. As such, I'm going to close this but we can use it as reference should we need to decode more in the future.

PR #1736, opened 2024-07-08
Adjust `SwapsApiFactory` to extend `IApiManager`

iamacook:
## Summary

The `SwapsApiFactory` caches instances of `ISwapsApi` after instantiation. Whilst it currently relies on no cached values, it does not adhere to our `IApiManager` pattern.

This adjusts `SwapsApiFactory` to extend `IApiManager`.

## Changes

- Convert `SwapsApiFactory` to extend `IApiManager` and propagate changes

### DIFF
--- a/src/datasources/swaps-api/swaps-api.factory.ts
+++ b/src/datasources/swaps-api/swaps-api.factory.ts
@@ -20,9 +20,9 @@ export class SwapsApiFactory implements ISwapsApiFactory {
     private readonly configurationService: IConfigurationService,
   ) {}
 
-  get(chainId: string): ISwapsApi {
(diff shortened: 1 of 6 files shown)


src/datasources/swaps-api/swaps-api.factory.ts
```diff
@@ -20,9 +20,9 @@ export class SwapsApiFactory implements ISwapsApiFactory {
     private readonly configurationService: IConfigurationService,
   ) {}
 
-  get(chainId: string): ISwapsApi {
+  getApi(chainId: string): Promise<ISwapsApi> {
```
iamacook, line 23: This needs return a promise to conform with the interface.

PR #1738, opened 2024-07-08
Extract swap app allowance into helper

iamacook:
## Summary

Swap decoded is restricted to certain allowed apps. Now that we are decoding swap orders/transfers and TWAP orders, this check occurs in several places. There are dedicated `isAppAllowed` methods for both swap and TWAPs.

This extracts the multiple methods in a new `SwapApps` helper, calling it in every places instead.

## Changes

- Create and import `SwapAppsHelper`
- Replace all calls to `isAppAllowed` with that from the above helper


hectorgomezv (commented):
LGTM 👏🏻 But I think a rebase is needed to solve the conflicts.

hectorgomezv (approved):

PR #1753, opened 2024-07-10
Add FF_RUN_MIGRATIONS 

hectorgomezv:
## Summary
This PR adds a feature flag to (de-)activate the database migrations execution at the service startup.

## Changes
- Adds `FF_RUN_MIGRATIONS` to control the execution of database migrations at the application bootstrap.
- Adds a test suite for `PostgresDatabaseMigrationHook`.

### DIFF
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -107,6 +107,7 @@ export default (): ReturnType<typeof configuration> => ({
   },
   express: { jsonLimit: '1mb' },
   features: {
+    runMigrations: true,
     richFragments: true,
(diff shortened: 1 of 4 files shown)


iamacook (approved):

hectorgomezv:
As discussed offline, we see this as a temporal security measure to provide a switch that allows the team to block the database migrations from running by setting an environment variable. For the longer term, we aim to build a more elaborate mechanism where specific migrations can be linked to feature flags, so these migrations would only apply to the database if the feature flag is active. (cc. @iamacook)

iamacook (approved):

PR #1754, opened 2024-07-11
Add a retry mechanism to data-decoder e2e test

hectorgomezv:
## Summary
The E2E test file `data-decode.e2e-spec.ts` contains a flaky test failing on some CI executions. The first call to the Transaction Service for a non-cached data input takes some seconds to respond. Subsequent calls are much faster. This causes the first call to time out sometimes.

## Changes
- Adds a retry mechanism to the affected test.


hectorgomezv:
Closed in favor of #1755

PR #1757, opened 2024-07-11
Set MAX_TTL as TTL when caching SAFE_EXISTS_KEY

hectorgomezv:
## Summary
This PR fixes the TTL value when storing the `SAFE_EXISTS_KEY`. It was previously set to `-1`, but `RedisCacheService` explicitly avoids storing values when the TTL is less than zero. (See [this reference](https://github.com/safe-global/safe-client-gateway/blob/5a66c242d50fc32868b0bfb5c8a4467163053801/src/datasources/cache/redis.cache.service.ts#L39)) 

## Changes
- Changes `INDEFINITE_EXPIRATION_TIME` for `MAX_TTL`, with a value of `2^31 - 1` (maximum value allowed by Redis, maximum positive value of a 32-bit signed integer)

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.ts
+++ b/src/datasources/transaction-api/transaction-api.service.ts
@@ -31,7 +31,7 @@ import { get } from 'lodash';
 
 export class TransactionApi implements ITransactionApi {
   private static readonly ERROR_ARRAY_PATH = 'nonFieldErrors';
-  private static readonly INDEFINITE_EXPIRATION_TIME = -1;
+  private static readonly MAX_TTL = 2147483647;
 
   private readonly defaultExpirationTimeInSeconds: number;
   private readonly defaultNotFoundExpirationTimeSeconds: number;
@@ -199,7 +199,7 @@ export class TransactionApi implements ITransactionApi {
       JSON.stringify(isSafe),
       isSafe
         ? // We can indefinitely cache this as an address cannot "un-Safe" itself
-          TransactionApi.INDEFINITE_EXPIRATION_TIME
+          TransactionApi.MAX_TTL
         : this.defaultExpirationTimeInSeconds,
     );
 
--- a/src/datasources/transaction-api/transaction-api.service.ts
+++ b/src/datasources/transaction-api/transaction-api.service.ts
@@ -2,6 +2,7 @@ import { IConfigurationService } from '@/config/configuration.service.interface'
 import { CacheFirstDataSource } from '@/datasources/cache/cache.first.data.source';
 import { CacheRouter } from '@/datasources/cache/cache.router';
 import { ICacheService } from '@/datasources/cache/cache.service.interface';
+import { MAX_TTL } from '@/datasources/cache/constants';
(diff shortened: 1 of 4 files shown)


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -31,7 +31,7 @@ import { get } from 'lodash';
 
 export class TransactionApi implements ITransactionApi {
   private static readonly ERROR_ARRAY_PATH = 'nonFieldErrors';
-  private static readonly INDEFINITE_EXPIRATION_TIME = -1;
+  private static readonly MAX_TTL = 2147483647;
```
iamacook, line 34: Nit: I'd add a comment explaining this value and add some separators `2_147_483_647` or change it to `2^31 - 1`.

src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -199,7 +199,7 @@ export class TransactionApi implements ITransactionApi {
       JSON.stringify(isSafe),
       isSafe
         ? // We can indefinitely cache this as an address cannot "un-Safe" itself
-          TransactionApi.INDEFINITE_EXPIRATION_TIME
+          TransactionApi.MAX_TTL
```
iamacook, line 202: Considering `-1` is a "valid" TTL value, do you think we should intead modify [the client](https://github.com/safe-global/safe-client-gateway/blob/5a66c242d50fc32868b0bfb5c8a4467163053801/src/datasources/cache/redis.cache.service.ts#L39) to allow it instead? We otherwise lose the ability to cache indefinitely.

Should we want to prevent negative numbers entirely, we should remove the above comment.

iamacook (commented):
This is very sneaky - good catch!

src/datasources/transaction-api/transaction-api.service.ts
hectorgomezv: I agree, I thought about the same thing while fixing this, but I'm pretty confident that modifying the client would have an impact in other places of the code because if I recall correctly `CacheFirstDataSource.get` is called without an `expireTimeSeconds` value on purpose to prevent `CacheService` from caching the result.

So while I agree with your suggestion, I'd like to do it in a separate PR if you agree.

hectorgomezv: I found an example: https://github.com/safe-global/safe-client-gateway/blob/5a66c242d50fc32868b0bfb5c8a4467163053801/src/datasources/transaction-api/transaction-api.service.ts#L251

iamacook: Let's keep it as is then - the TTL equates to just over 68 years!

hectorgomezv: I've set it to `Number.MAX_SAFE_INTEGER - 1` in [bd424ec](https://github.com/safe-global/safe-client-gateway/pull/1757/commits/bd424ec02276ec00e9092172e52773772ca832d6) and I've added a note elaborating this decision. Thanks for the input!

iamacook (approved):

Issue #1750, opened 2024-07-10
Implement endpoint to return blocked addresses


iamacook:
### Description

Clients require a list of blocked addresses for swaps which is [currently hardcoded](https://github.com/safe-global/safe-wallet-web/blob/84050cc88bf353d06f5b794bf49ffa81e85fed9a/src/services/ofac/blockedAddressList.json).

Whilst this could be included in the Config Service, this list is likely to be updated often and manual management is difficult/error prone. Ideally we could implement a scheduled job to get the addresses from a dedicated source, e.g. [Chainalysis](https://go.chainalysis.com/chainalysis-oracle-docs.html) or Blockaid.

### Requirements

1. Implement an automatic job to fetch blocked addresses from a dedicated source.
2. Return aforementioned list from a dedicated endpoint, per chain.

iamacook:
After discussion within the team, we will instead opt for an RPC call to a sanctions oracle. cc @compojoom

PR #1758, opened 2024-07-12
Santioned addresses indexer

iamacook:
⚠️ This is demonstative, showcasing a basic indexer (with caveats) for https://github.com/safe-global/safe-client-gateway/issues/1750

## Summary

As clients restrict swap features based on address sactioning, a [hardcoded list of addresses](https://github.com/safe-global/safe-wallet-web/blob/84050cc88bf353d06f5b794bf49ffa81e85fed9a/src/services/ofac/blockedAddressList.json) is currently referenced. Whilst this works, it requires manual updating and release.

This lays the foundations for indexing of the [Chainalysis oracle for sanctioned addresses](https://go.chainalysis.com/chainalysis-oracle-docs.html). Indexing has its caveats as chains often restrict `eth_getLogs` calls meaning we would need to index in chucks (or pages). These limits vary by chain and, depending on the limit, could mean indexing takes a long time.

Note: whether we opt for this and cache/store the addresses, or instead rely on an API is to be decided.

## Changes

- Add `SanctionedAddressesIndexer`
- Add appropriate test coverage with respective builders

### DIFF
--- a/src/domain/blockchain/__tests__/log.builder.ts
+++ b/src/domain/blockchain/__tests__/log.builder.ts
@@ -1,0 +1,16 @@
+import { Builder, IBuilder } from '@/__tests__/builder';
+import { faker } from '@faker-js/faker';
+import { getAddress, Log } from 'viem';
+
+export function logBuilder(): IBuilder<Log> {
+  return new Builder<Log>()
+    .with('address', getAddress(faker.finance.ethereumAddress()))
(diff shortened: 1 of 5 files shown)


iamacook:
After discussion with the team/taking the aforementioned caveats into account, I will close this. We may, however, use it as reference as one of these will be the proposed solutions:

- Caching an external API to avoid indexing
- Request indexing by Core (if possible)
- Create a dedicated indexing service (based on this)

PR #1760, opened 2024-07-15
Add Accounts Datasource cache layer

hectorgomezv:
## Summary
This PR adds a cache layer to `AccountsDatasource`. 

The cache management for the datasource behavior regarding `Account` is:
- The `Account` is stored in the cache with the default TTL after creation.
- When reading an `Account` from the database, it gets stored in the cache with the default TTL.
- On `Account` deletion, the associated cache is cleared.

The cache management for the datasource behavior regarding `AccountDataType` is:
- The array of all `AccountDataTypes` is stored after a first read from the database with indefinite TTL.

The cache management for the datasource behavior regarding `AccountDataSettings` is:
- The array of all `AccountDataSettings` for a given `address`/`Account` is stored after a first read from the database with the default TTL.
- The array of all `AccountDataSettings` for a given `address`/`Account` is stored after a data update with the default TTL.

## Changes
- Adds cache management to `AccountDatasource`, along with its associated tests.
- The private function `getFromCacheOrExecuteAndCache` was added to `AccountDatasource`. This function abstracts the common logic for all its client functions: firstly it tries to get the data from the cache, and in its absence, tries to get the data from the database and stores the result (if any) in the cache.


src/datasources/db/entities/row.entity.ts
```diff
@@ -9,6 +9,6 @@ export type Row = z.infer<typeof RowSchema>;
  */
 export const RowSchema = z.object({
   id: z.number().int(),
-  created_at: z.date(),
-  updated_at: z.date(),
+  created_at: z.coerce.date(),
+  updated_at: z.coerce.date(),
```
hectorgomezv, line 13: This was changed because we store plain JSON in the cache, so the dates are stored as date-formatted strings, which need to be coerced.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -58,25 +94,47 @@ export class AccountsDatasource implements IAccountsDatasource {
     const { count } = await this
       .sql`DELETE FROM accounts WHERE address = ${address}`;
 
+    const { key } = CacheRouter.getAccountCacheDir(address);
+    await this.cacheService.deleteByKey(key);
```
iamacook, line 98: If this throws, the cache will remain whereas it will be removed from the database.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -138,4 +198,39 @@ export class AccountsDatasource implements IAccountsDatasource {
...
+   * @param ttl time to live for the cache
+   * @returns content from cache or query result
+   */
+  private async getFromCacheOrExecuteAndCache<T extends postgres.MaybeRow[]>(
+    cacheDir: CacheDir,
+    query: postgres.PendingQuery<T>,
+    ttl: number,
+  ): Promise<T> {
+    const { key, field } = cacheDir;
+    const cached = await this.cacheService.get(cacheDir);
```
iamacook, line 218: If this throws, should we instead continue with the database interaction?

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -58,25 +94,47 @@ export class AccountsDatasource implements IAccountsDatasource {
...
 
+  getActiveDataTypes(): Promise<AccountDataType[]> {
+    const cacheDir = CacheRouter.getActiveAccountDataTypesCacheDir();
+    return this.getFromCacheOrExecuteAndCache<AccountDataType[]>(
+      cacheDir,
+      this.sql<
+        AccountDataType[]
+      >`SELECT * FROM account_data_types WHERE is_active IS TRUE`,
+      MAX_TTL,
+    );
```
iamacook, line 113: We have all data types in the cache so we could derive this from `getDataTypes`. Is there a benefit in caching the both?

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -68,21 +102,59 @@ describe('AccountsDatasource tests', () => {
...
+      await target.createAccount(address);
+
+      // cache the account
+      await target.getAccount(address);
+
+      // delete the account from the database
+      await sql`DELETE FROM accounts WHERE address = ${address}`;
+
+      // check the account is still in the cache
+      const result = await target.getAccount(address);
```
iamacook, line 125: If the account has been deleted, we shouldn't keep it in cache.

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -103,53 +175,180 @@ describe('AccountsDatasource tests', () => {
...
+      const rows = await sql`
+        INSERT INTO account_data_types ${sql(dataTypes, 'name')} RETURNING (id)`;
+
+      // cache the data types
+      await target.getDataTypes();
+
+      // delete the data types from the database
+      await sql`
+        DELETE FROM account_data_types WHERE id 
+          IN ${sql(rows.map((id) => id.id))}`;
```
iamacook, line 256: As before regarding database and cache out of sync.

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -184,6 +383,44 @@ describe('AccountsDatasource tests', () => {
...
+        ${sql(accountDataSettingRows, 'account_id', 'account_data_type_id', 'enabled')} returning *`;
+
+      // cache the account data settings
+      await target.getAccountDataSettings(address);
+
+      // delete the account data settings
+      await sql`DELETE FROM account_data_settings`;
+
+      // check the account data settings are still in the cache
+      const actual = await target.getAccountDataSettings(address);
```
iamacook, line 411: As before regarding database and cache out of sync.

iamacook (commented):
I am a little concerned regarding the database and cache being out of sync. Perhaps we could [subscribe to database interactions](https://www.npmjs.com/package/postgres#realtime-subscribe) and update the cache accordingly?

src/datasources/accounts/accounts.datasource.ts
hectorgomezv: Yet, good point, maybe it's better to cache them all and just filter to get the active ones. Will do so 👌🏻

src/datasources/accounts/accounts.datasource.ts
hectorgomezv: I'd say if the `cacheService.get` call throws it's OK for the whole function to throw and do not continue with the database interaction. It's the same with the rest of the datasources (if the cache retrieval fails, they won't continue with the REST requests. Wdyt? From my perspective, the cache is a hard requirement for the service so it's ok to fail.

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: It's deleted from the database using a raw SQL query, so it is still present in the cache. Maybe this is a bit weird way to test the cache contents, in other tests the cache contents are read and checked directly using `fakeCacheService.get`, I combined both approaches in the tests, and this is a more _complete_ one because it gets the content using `getAccount(address)`, but maybe reading the cache directly is enough. Will change this approach 👍🏻

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: Answered in https://github.com/safe-global/safe-client-gateway/pull/1760#discussion_r1679010289

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: Answered in https://github.com/safe-global/safe-client-gateway/pull/1760#discussion_r1679010289

src/datasources/accounts/accounts.datasource.ts
hectorgomezv: Good catch, theoretically if the `DELETE` throws it won't be deleted, but maybe there are some scenarios where the deletion is performed and the cache invalidation is not, so I'm wrapping this on a `try-finally` statement in [cd76e37](https://github.com/safe-global/safe-client-gateway/pull/1760/commits/cd76e374714c6582c39d3842a0837cda1f4a54fa)

src/datasources/accounts/accounts.datasource.ts
hectorgomezv: I've removed the `AccountsDatasource.getActiveDataTypes`. In the end, it was just used internally, and as you suggested it's probably simpler to just store the data types and filter by activeness when needed. ([02d403d](https://github.com/safe-global/safe-client-gateway/pull/1760/commits/02d403db0a2a7c0ac65eacb6360270cbc6514a77))

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: The testing approach is changed in order to check the cache contents directly in [c611031](https://github.com/safe-global/safe-client-gateway/pull/1760/commits/c6110318a438e3d278ca59a05fc49f9b20ace5d6)

hectorgomezv:
> I am a little concerned regarding the database and cache being out of sync. Perhaps we could [subscribe to database interactions](https://www.npmjs.com/package/postgres#realtime-subscribe) and update the cache accordingly?

For Accounts and Counterfactual Safes, I’m considering just managing the cache directly when CRUDing items, it should be simple enough and I think we should be able to avoid out-of-syncs, but I’m open to changing it of course, if we see any problem regarding synchronization.

PR #1766, opened 2024-07-16
Add `GET` `/v1/sanction-addresses`, returning addresses from SDN list

iamacook:
## Summary

Our clients restrict access to the swap feature is an associated address is present on the SDN sanctions list. Currently, this list is [hardcoded](https://github.com/safe-global/safe-wallet-web/blob/84050cc88bf353d06f5b794bf49ffa81e85fed9a/src/services/ofac/blockedAddressList.json) and therefore needs manually updating.

This adds a new `GET` `/v1/sanctioned-addresses` endpoint the returns a list of addresses that are parsed/cached from the aforementioned SDN list.

Note: a follow up with include a relative CRON job to update the above accordingly.

## Changes

- Modify `FetchClient` to accordingly return text/json relative to the `content-type` of a response
- Add `SanctionedAddressesRepository` with method to get/update (and parse SDN list) sanctioned addresses
- Add new controller, returning the sanctioned addresses under `GET` `/v1/sanctioned-addresses`

### DIFF
--- a/src/routes/sanctioned-addresses/sanctioned-addresses.controller.ts
+++ b/src/routes/sanctioned-addresses/sanctioned-addresses.controller.ts
@@ -1,0 +1,18 @@
+import { SanctionedAddressesService } from '@/routes/sanctioned-addresses/sanctioned-addresses.service';
+import { Controller, Get } from '@nestjs/common';
+import { ApiTags } from '@nestjs/swagger';
+
+@ApiTags('sanctioned-addresses')
+@Controller({
+  version: '1',
+})
+export class SanctionedAddressesController {
+  constructor(
+    private readonly sanctionedAddressesService: SanctionedAddressesService,
+  ) {}
+
+  @Get('sanctioned-addresses')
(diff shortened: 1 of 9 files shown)


src/routes/sanctioned-addresses/sanctioned-addresses.controller.ts
```diff
@@ -0,0 +1,18 @@
...
+@ApiTags('sanctioned-addresses')
+@Controller({
+  version: '1',
+})
+export class SanctionedAddressesController {
+  constructor(
+    private readonly sanctionedAddressesService: SanctionedAddressesService,
+  ) {}
+
+  @Get('sanctioned-addresses')
```
iamacook, line 14: I'm not sure about the naming here.

iamacook (commented):
I've not included the CRON job in this PR but sooner in a follow up. It's relatively simple to implement as NestJS has a [native package](https://docs.nestjs.com/techniques/task-scheduling) to do so.

iamacook:
After discussion within the team, we will instead opt for an RPC call to a sanctions oracle. cc @compojoom

PR #1768, opened 2024-07-17
Refactor `JwtService` to expose issuer/secret

iamacook:
## Summary

The `JwtService` currently signs, verifies and decodes tokens according to our environment variables and there is no way of overriding these values.

As the Firebase datasource requires a JWT token signed with a custom issuer/secret, this refactors the aforementioned service to facilitate it.

## Changes

- Expose the option to specify issuer/secret to `JwtClient`, otherwise defaulting to env. vars. in `JwtService`
- Remove `options` that override claims on token, instead opting for their spec, e.g. `iat`, `nbf`, `exp` (instead of `issuedAt`, `notBefore`, `expiresIn`)
- Accept/return dates for the all spec. specific dateclaims
- Use `JwtService` in `FirebaseCloudMessagingApiService` instead of `jsonwebtoken`
- Update tests accordingly

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -35,6 +35,16 @@
 # The API Key to be used. If none is set, balances cannot be retrieved using this provider.
 #ZERION_API_KEY=
 
+# Push Notifications Provider - Firebase Cloud Messaging
+# Firebase API URL
+# (default=https://fcm.googleapis.com/v1/projects)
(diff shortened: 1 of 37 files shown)


hectorgomezv (dismissed):
👏🏻👏🏻👏🏻

hectorgomezv (approved):

PR #1770, opened 2024-07-17
Generify hook-related cache invalidation

iamacook:
## Summary

As the forthcoming notifications will be dispatched according to hook events, there should ideally be a distinction between cache- and notification-related hook logic, e.g. the notification logic not be implemented within the "cache-hooks".

This refactors the route "cache-hooks" to be a more generic, also adding distinction between Transaction Service and Config. Service events.

## Changes

- Rename `cache-hooks` to `hook` and the associated controller, service(s) and module accordingly
- Separate `EventType` into `TransactionEventType` and `ConfigEventType` and propagate changes


src/routes/hooks/entities/schemas/web-hook.schema.ts
```diff
@@ -0,0 +1,32 @@
...
+import { ModuleTransactionEventSchema } from '@/routes/hooks/entities/schemas/module-transaction.schema';
+import { NewConfirmationEventSchema } from '@/routes/hooks/entities/schemas/new-confirmation.schema';
+import { NewMessageConfirmationEventSchema } from '@/routes/hooks/entities/schemas/new-message-confirmation.schema';
+import { OutgoingEtherEventSchema } from '@/routes/hooks/entities/schemas/outgoing-ether.schema';
+import { OutgoingTokenEventSchema } from '@/routes/hooks/entities/schemas/outgoing-token.schema';
+import { PendingTransactionEventSchema } from '@/routes/hooks/entities/schemas/pending-transaction.schema';
+import { SafeAppsUpdateEventSchema } from '@/routes/hooks/entities/schemas/safe-apps-update.schema';
+import { SafeCreatedEventSchema } from '@/routes/hooks/entities/schemas/safe-created.schema';
+
+export const WebHookSchema = z.discriminatedUnion('type', [
```
hectorgomezv, line 17: Should we also rename this to `EventSchema` to stay consistent?

src/routes/hooks/hooks.service.ts
```diff
@@ -0,0 +1,36 @@
+import { IConfigurationService } from '@/config/configuration.service.interface';
+import { IQueuesRepository } from '@/domain/queues/queues-repository.interface';
+import { WebHookSchema } from '@/routes/hooks/entities/schemas/web-hook.schema';
+import { OnModuleInit } from '@nestjs/common';
+import { ConsumeMessage } from 'amqplib';
+import { Event } from '@/routes/hooks/entities/event.entity';
+import { ILoggingService } from '@/logging/logging.interface';
+
+export abstract class HooksService implements OnModuleInit {
```
hectorgomezv, line 9: While I'm totally aligned with the purpose of this refactor, I think I don't fully understand this abstract class part: are we planning to extend `HooksService` with another derived class for notifications (e.g.: `HooksNotificationsService`)? 

If that is the case, wouldn't that make `onModuleInit` execute twice, and therefore create two queue subscriptions? Would the `HooksController` dispatch the events to the proper implementation of `HooksService`, or all events would be consumed by all the implementation classes?

hectorgomezv (commented):
Superb job! 👏🏻 

I've just left minor comments/thoughts.

src/routes/hooks/hooks.service.ts
iamacook: As per our discussion, I've [opened a different PR](https://github.com/safe-global/safe-client-gateway/pull/1774) that moves the queue subscription into the domain, but accepts multiple listeners for said subscription. This should prevent multiple subscriptions. I will rebase this PR after merging that.

src/routes/hooks/entities/schemas/web-hook.schema.ts
iamacook: Renamed when rebasing according to https://github.com/safe-global/safe-client-gateway/pull/1770#discussion_r1682730520.

hectorgomezv (approved):
👏🏻

hectorgomezv (approved):

hectorgomezv (approved):

PR #1773, opened 2024-07-18
Add Counterfactual Safes Datasource

hectorgomezv:
## Summary
This PR adds a datasource & cache layer to store the Counterfactual Safes data. 

This datasource includes the following API:
- `createCounterfactualSafe`: stores a new CF Safe.
- `getCounterfactualSafe`: gets an existing CF Safe by chain ID and predicted address.
- `getCounterfactualSafesForAccount`: gets all the existing CF Safes for a given `Account`.
- `deleteCounterfactualSafe`: deletes an existing CF Safe by chain ID and predicted address.
- `deleteCounterfactualSafesForAccount`: deletes all the existing CF Safes for a given `Account`.

The cache management for the datasource behavior regarding `CounterfactualSafes` is:
- The `CounterfactualSafe` is stored in the cache with the default TTL after the first retrieval by ID.
- The array of `CounterfactualSafes` associated with an `Account` is stored in the cache with the default TTL after the first retrieval by `Account`.
- On `CounterfactualSafe` deletion, the associated cache is cleared.
- The cache for `CounterfactualSafes` is also cleared when the associated `Account` is deleted.

## Changes
- Adds `CounterfactualSafesDatasource` and associated tests.
- Adds the database migration `00004_counterfactual-safes`, which defines the `counterfactual_safes` table.
- The private function `getFromCacheOrExecuteAndCache` was moved to `@/datasources/db/utils.ts`, as it's used by several datasources now (`Accounts` and `CounterfactualSafes`).


src/datasources/db/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+ * If the specified {@link CacheDir} is not empty, the pointed content is returned.
+ *
+ * @param loggingService {@link ILoggingService} to use for logging
+ * @param cacheService {@link ICacheService} to use for caching
+ * @param cacheDir {@link CacheDir} to use for caching
+ * @param query query to execute
+ * @param ttl time to live for the cache
+ * @returns content from cache or query result
+ */
+// TODO: add tests
```
hectorgomezv, line 20: I'll add tests in a separate PR to avoid distracting from this PR's main purposes.

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.ts
```diff
@@ -0,0 +1,145 @@
...
+  ) {
+    this.defaultExpirationTimeInSeconds =
+      this.configurationService.getOrThrow<number>(
+        'expirationTimeInSeconds.default',
+      );
+  }
+
+  // TODO: the repository calling this function should:
+  // - check the AccountDataSettings to see if counterfactual-safes is enabled.
+  // - check the AccountDataType to see if it's active.
```
hectorgomezv, line 37: I've just left this TODO here for reference when implementing the Repository that will use this datasource.

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.spec.ts
```diff
@@ -0,0 +1,407 @@
...
+      // the cache is cleared after creating a new CF Safe for the same account
+      await target.createCounterfactualSafe(
+        account,
+        createCounterfactualSafeDtoBuilder().build(),
+      );
+      const afterCreation = await fakeCacheService.get(cacheDir);
+      expect(afterCreation).toBeUndefined();
+    });
+
+    // TODO: move this check to the repository.
```
hectorgomezv, line 115: I've also left this here for reference but this test should be moved to the Repository when it's implemented.

migrations/__tests__/00004_counterfactual-safes.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+  created_at: Date;
+  updated_at: Date;
+  id: number;
+  chain_id: string;
+  creator: string;
+  fallback_handler: string;
+  owners: string[];
+  predicted_address: string;
+  salt_nonce: string;
+  singleton_address: string;
```
iamacook, line 25:
```suggestion
  creator:`0x${string}`;
  fallback_handler: `0x${string}`;
  owners: `0x${string}`[];
  predicted_address: `0x${string}`;
  salt_nonce: string;
  singleton_address: `0x${string}`;
```

src/datasources/db/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+export async function getFromCacheOrExecuteAndCache<
+  T extends postgres.MaybeRow[],
+>(
+  loggingService: ILoggingService,
+  cacheService: ICacheService,
```
iamacook, line 25: To save having to pass this every time, perhaps we should consider changing this to a domain module of some kind?

src/domain/accounts/counterfactual-safes/entities/__tests__/create-counterfactual-safe.dto.entity.builder.ts
```diff
@@ -0,0 +1,18 @@
...
+export function createCounterfactualSafeDtoBuilder(): IBuilder<CreateCounterfactualSafeDto> {
+  return new Builder<CreateCounterfactualSafeDto>()
+    .with('chain_id', faker.string.numeric())
+    .with('fallback_handler', getAddress(faker.finance.ethereumAddress()))
+    .with('owners', [
+      getAddress(faker.finance.ethereumAddress()),
+      getAddress(faker.finance.ethereumAddress()),
+    ])
+    .with('predicted_address', getAddress(faker.finance.ethereumAddress()))
+    .with('salt_nonce', faker.string.uuid())
```
iamacook, line 15:
```suggestion
    .with('salt_nonce', faker.string.hexadecimal())
```

src/domain/accounts/counterfactual-safes/entities/counterfactual-safe.entity.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export type CounterfactualSafe = z.infer<typeof CounterfactualSafeSchema>;
+
+export const CounterfactualSafeSchema = RowSchema.extend({
+  chain_id: z.string(),
+  creator: AddressSchema,
+  fallback_handler: AddressSchema,
+  owners: z.array(AddressSchema),
+  predicted_address: AddressSchema,
+  salt_nonce: z.string(),
```
iamacook, line 14:
```suggestion
  salt_nonce: HexSchema,
```

src/domain/accounts/counterfactual-safes/entities/counterfactual-safe.entity.ts
```diff
@@ -0,0 +1,18 @@
...
+  singleton_address: AddressSchema,
+  threshold: z.number().int().gte(0),
```
iamacook, line 16:
```suggestion
  threshold: z.number().int().gte(1),
```

iamacook: Consider the minimum value, maybe we should also enforce a minimum number of `owner` elements as well. What do you think?

src/domain/accounts/counterfactual-safes/entities/create-counterfactual-safe.dto.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+    this.threshold = props.threshold;
+  }
+}
+
+export const CreateCounterfactualSafeDtoSchema = z.object({
+  chain_id: z.string(),
+  fallback_handler: AddressSchema,
+  owners: z.array(AddressSchema),
+  predicted_address: AddressSchema,
+  salt_nonce: z.string(),
```
iamacook, line 31:
```suggestion
  salt_nonce: HexSchema,
```

src/domain/accounts/counterfactual-safes/entities/create-counterfactual-safe.dto.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+  singleton_address: AddressSchema,
+  threshold: z.number().int().gte(0),
```
iamacook, line 33:
```suggestion
  threshold: z.number().int().gte(1),
```

(and as before regarding minimum length of `owners`.)

src/domain/accounts/counterfactual-safes/entities/counterfactual-safe.entity.ts
```diff
@@ -0,0 +1,18 @@
...
+  account_id: AccountSchema.shape.id,
+});
```
iamacook, line 18: Maybe we should extend the `CreateCounterfactualSafeDtoSchema` or inidividual field schemas if the property names are changed.

src/domain/accounts/counterfactual-safes/entities/create-counterfactual-safe.dto.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+export class CreateCounterfactualSafeDto
+  implements z.infer<typeof CreateCounterfactualSafeDtoSchema>
+{
+  chain_id: string;
+  fallback_handler: `0x${string}`;
+  owners: `0x${string}`[];
+  predicted_address: `0x${string}`;
+  salt_nonce: string;
+  singleton_address: `0x${string}`;
+  threshold: number;
```
iamacook, line 13: Nit: I think we should remain consistent and expect camelcase.

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.ts
```diff
@@ -0,0 +1,145 @@
...
+    const [counterfactualSafe] = await this.sql<CounterfactualSafe[]>`
+      INSERT INTO counterfactual_safes 
+      ${this.sql([this.mapCreationDtoToRow(account, createCounterfactualSafeDto)])}
+      RETURNING *`;
+    const { key } = CacheRouter.getCounterfactualSafesCacheDir(account.address);
+    await this.cacheService.deleteByKey(key);
+    return counterfactualSafe;
+  }
+
+  async getCounterfactualSafe(id: string): Promise<CounterfactualSafe> {
```
iamacook, line 51: I would expect us to retrieve it by the predicted address, otherwise we are exposing the internal ID.

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.ts
```diff
@@ -0,0 +1,145 @@
...
+      this.loggingService,
+      this.cacheService,
+      cacheDir,
+      this.sql<CounterfactualSafe[]>`
+        SELECT * FROM counterfactual_safes WHERE account_id = ${account.id}`,
+      this.defaultExpirationTimeInSeconds,
+    );
+  }
+
+  async deleteCounterfactualSafe(account: Account, id: string): Promise<void> {
```
iamacook, line 87: As above regarding predicted address vs. ID.

src/datasources/db/utils.ts
hectorgomezv: I agree, I was considering the same thing, but as I plan to add the tests separately, are you OK if I do both things together in a separate PR?

src/domain/accounts/counterfactual-safes/entities/counterfactual-safe.entity.ts
hectorgomezv: Do you mean extending DTOs from entity schemas? I haven't considered it, while I think in this case might make sense, I don't see it as a common pattern as DTOs shape does not match its _target entity_ schema necessarily, wdyt? (I can do it for this one, for sure)

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.ts
hectorgomezv: Answered in https://github.com/safe-global/safe-client-gateway/pull/1773#discussion_r1684141115

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.ts
hectorgomezv: We could also generate a unique uuid per CF Safe. 

And another thought: we need to keep in mind the endpoint to get this data would be authenticated.

src/datasources/db/utils.ts
iamacook: Of course!

src/domain/accounts/counterfactual-safes/entities/counterfactual-safe.entity.ts
iamacook: Let's keep it as tis then.

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.ts
iamacook: I would be more in favour of fetching by chain ID and address. It's easier for the clients imo.

src/domain/accounts/counterfactual-safes/entities/counterfactual-safe.entity.ts
hectorgomezv: Good point, done in [f2ca4c0](https://github.com/safe-global/safe-client-gateway/pull/1773/commits/f2ca4c0847951435d54b09bde99a34f1bf63c0ea)

src/domain/accounts/counterfactual-safes/entities/create-counterfactual-safe.dto.entity.ts
hectorgomezv: I agree, fixed in [485ece5](https://github.com/safe-global/safe-client-gateway/pull/1773/commits/485ece55c7f0ab4f2c494a1ea66156f46db630e3)

My initial thought was to use snake_case in the domain DTO and camelCase in the route DTO. This would align better with the fact we are returning snake_case entities in the domain (so the DTO would be snake_case also), but it makes no sense with the current naming standard, which seems to be:

**Input (DTO)**
Route: camelCase
Domain: camelCase
Datasource: here the mapping camelCase-> snake_case is done

**Output (result entity)**
Datasource: snake_case
Domain: snake_case
Route: here the mapping snake_case -> camelCase is done

What do you think about mapping both DTO and entities in the same layer? I'd say the datasource seems the proper place. That would restrict the usage of snake_case to the datasource only, and the domain would only see camelCase for both DTOs and entities.

So the proposal would be:

Input (DTO)
Route: camelCase
Domain: camelCase
Datasource: here the mapping camelCase-> snake_case is done

Output (result entity)
Datasource: here the mapping snake_case -> camelCase is done
Domain: camelCase
Route: camelCase

If you are happy with the proposal, I'll take care of it in a separate PR.

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.ts
hectorgomezv: Done in [30f64bf](https://github.com/safe-global/safe-client-gateway/pull/1773/commits/30f64bf2cb6dd431bc2e863c75232d9fa5bb1467)

hectorgomezv: I've added [a second PR](https://github.com/safe-global/safe-client-gateway/pull/1776) to type the datasource arguments, as it's otherwise easy to make an error while passing parameters between the layers. This aligns the implementation more with other classes in the codebase.

src/domain/accounts/entities/__tests__/account.builder.ts
```diff
@@ -5,8 +5,8 @@ import { getAddress } from 'viem';
 
 export function accountBuilder(): IBuilder<Account> {
   return new Builder<Account>()
-    .with('id', faker.number.int({ max: Number.MAX_SAFE_INTEGER }))
-    .with('group_id', faker.number.int({ max: Number.MAX_SAFE_INTEGER }))
+    .with('id', faker.number.int({ max: 1_000_000 }))
+    .with('group_id', faker.number.int({ max: 1_000_000 }))
```
hectorgomezv, line 9: This is an arbitrary value, I think this can be extracted to a constants file, under a descriptive name like `DB_MAX_SAFE_INTEGER` or something similar, wdyt?

hectorgomezv:
After accepting them, I've reverted the commits where `saltNonce` was defined as hexadecimal in the `CreateCounterfactualSafeDto` schema. According to [this](https://www.notion.so/safe-global/Counterfactual-multichain-2ae1c8b081dc47a3ad9e22c8239db9bc?pvs=26&qid=#f8a60c218ce84e09917825a7efe002af), it's expected to be a numeric string, right?

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.spec.ts
```diff
@@ -0,0 +1,430 @@
...
+      // the cache is cleared after creating a new CF Safe for the same account
+      await target.createCounterfactualSafe(
+        account,
+        createCounterfactualSafeDtoBuilder().build(),
+      );
+      expect(await fakeCacheService.get(cacheDir)).toBeUndefined();
+    });
+
+    // TODO: move this check to the repository.
+    it.skip('should throw if the counterfactual-safes storage setting is disabled', async () => {
```
iamacook, line 116: Is this intentionally skipped?

src/domain/accounts/entities/__tests__/account.builder.ts
iamacook: Sounds good to me. (I checked the default `max` to make sure we're not increasing it and it's `Number.MAX_SAFE_INTEGER`.)

migrations/00004_counterfactual-safes/index.sql
```diff
@@ -0,0 +1,23 @@
...
+    owners VARCHAR(42)[] NOT NULL,
+    predicted_address VARCHAR(42) NOT NULL,
+    salt_nonce VARCHAR(255) NOT NULL,
+    singleton_address VARCHAR(42) NOT NULL,
+    threshold INTEGER NOT NULL,
+    created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    account_id INTEGER NOT NULL,
+    FOREIGN KEY (account_id) REFERENCES accounts(id) ON DELETE CASCADE,
+    CONSTRAINT unique_chain_address UNIQUE (chain_id, predicted_address)
```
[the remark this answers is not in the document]
hectorgomezv: Good point. Theoretically, predicted address collisions shouldn't happen between different accounts, but we can add the account_id to the unique restriction for consistency and also for performance improvement while checking the db index.

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.spec.ts
hectorgomezv: Yes, this was intended to be moved to the repository, which will be added in a forthcoming PR. But giving this a second thought, I've just removed it from this PR as it's out of context. I will include it in the repository PR.

migrations/__tests__/00004_counterfactual-safes.spec.ts
```diff
@@ -0,0 +1,236 @@
...
+        return sql<
+          CounterfactualSafesRow[]
+        >`SELECT * FROM counterfactual_safes WHERE account_id = ${accountRows[0].id}`;
+      },
+    });
+
+    expect(counterfactualSafesRows).toHaveLength(0);
+  });
+
+  it('should throw an error if the unique(chain_id, predicted_address) constraint is violated', async () => {
```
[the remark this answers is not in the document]
hectorgomezv: I've added it to the test description here [b6ae6a7](https://github.com/safe-global/safe-client-gateway/pull/1773/commits/b6ae6a7ef4f73295047cce15a4a36397944419ea), thanks!

src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.spec.ts
hectorgomezv: Removed in [b6ae6a7](https://github.com/safe-global/safe-client-gateway/pull/1773/commits/b6ae6a7ef4f73295047cce15a4a36397944419ea)

src/domain/accounts/counterfactual-safes/entities/create-counterfactual-safe.dto.entity.ts
iamacook: I agree with the proposal!

iamacook (approved):

iamacook (approved):

PR #1774, opened 2024-07-18
Move hook domain logic from routes to domain

iamacook:
## Summary

With the forthcoming implementation of notification dispatch according to events, hooks will also trigger enqueuing notifications. This moves the domain-considered logic from `CacheHooksService` to a new `HooksRepository` in the domain. There the aforementioned notification-relevant logic will be placed.

## Changes

- Move domain logic from `CacheHooksService` to `HooksRepository`
- Propagate changes accordingly


src/domain/queues/queues-repository.ts
```diff
@@ -1,18 +1,45 @@
...
     @Inject(IQueuesApiService) private readonly queuesApi: IQueuesApiService,
-  ) {}
+    @Inject(IConfigurationService)
+    private readonly configurationService: IConfigurationService,
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
+  ) {
+    this.queueName = this.configurationService.getOrThrow<string>('amqp.queue');
+  }
+
+  onModuleInit(): Promise<void> {
```
hectorgomezv, line 25: I think we might face a race condition here: if this `onModuleInit` function is executed before `CacheHooksService.onModuleinit`, then `this.listeners` would be empty and therefore the first events wouldn't be processed by any listener, right? I'm not sure the would be lost as they wouldn't be ACKed but maybe the events processing order is modified. I'd avoid relying on the classes instantation order, but maybe I'm missing something.

What do you think about moving the complete `CacheHooksService` implementation to the domain layer instead (including its `onEvent` function). To me, it was always a bit weird that this core logic live in the routes layer so this might be a good opportunity to move it to its own class in the domain.

iamacook: Good points, with which I agree with. As of 6271732a3782d10e70f6f461531e2b22ffdfccd6, I've moved the relevant `CacheHooksService` logic to the domain. Please let me know what you think.

hectorgomezv (approved):
🚀

PR #1776, opened 2024-07-19
Type datasources arguments

hectorgomezv:
## Summary
This PR changes the Accounts and CounterfactualSafes datasource function from using an enumeration of arguments to a fixed typed object as an argument, in alignment with the rest of the classes in the service.

## Changes
- Adds `args: {...}` as an argument to all non-singular argument functions on both `AccountsDatasource` and `CounterfactualSafesDatasource`.

### DIFF
--- a/src/domain/interfaces/counterfactual-safes.datasource.interface.ts
+++ b/src/domain/interfaces/counterfactual-safes.datasource.interface.ts
@@ -7,25 +7,27 @@ export const ICounterfactualSafesDatasource = Symbol(
 );
 
 export interface ICounterfactualSafesDatasource {
-  createCounterfactualSafe(
-    account: Account,
-    createCounterfactualSafeDto: CreateCounterfactualSafeDto,
-  ): Promise<CounterfactualSafe>;
+  createCounterfactualSafe(args: {
+    account: Account;
+    createCounterfactualSafeDto: CreateCounterfactualSafeDto;
+  }): Promise<CounterfactualSafe>;
 
-  getCounterfactualSafe(
-    chainId: string,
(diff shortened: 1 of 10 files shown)


src/domain/interfaces/counterfactual-safes.datasource.interface.ts
```diff
@@ -7,25 +7,27 @@ export const ICounterfactualSafesDatasource = Symbol(
...
+  deleteCounterfactualSafe(args: {
+    account: Account;
+    chainId: string;
+    predictedAddress: `0x${string}`;
+  }): Promise<void>;
 
-  deleteCounterfactualSafesForAccount(account: Account): Promise<void>;
+  deleteCounterfactualSafesForAccount(args: {
+    account: Account;
+  }): Promise<void>;
```
iamacook, line 32: Nit: we don't use objects for singular arguments elsewhere (`getCounterfactualSafesForAccount` and `deleteCounterfactualSafesForAccount`). Do you intend to extend them?

hectorgomezv: I don't have an strong opinion, maybe it's better to stick to the standards we have been following and use objects only for non-singular arguments. Will change that, thanks!

hectorgomezv: I've removed the types for single arguments in [1d0832e](https://github.com/safe-global/safe-client-gateway/pull/1776/commits/1d0832eba1b1e72bdf25fb974ce02f7cc71da0a3)

iamacook (approved):

PR #1785, opened 2024-07-24
Improve TWAP order mapping with `activeOrderUid`

iamacook:
## Summary

In order to cancel a TWAP, the current order UID needs to be referenced. As we calculate each of these, it makes sense to return it as part of the transaction info (as well as at the request of the client).

This adds a new `activeOrderUid` to the `TwapOrderTransactionInfo`. On top of this, as we were previously calculating the status based on the initial 10 parts of a TWAP, or the last for those with >10, we now fetch the status from the active part. This fixes the issue whereby large part TWAPs were returning the incorrect status - final parts are not created meaning they [would always throw](https://github.com/safe-global/safe-client-gateway/blob/b4f144746fb5b9b78cca7fa38046dc06cc7ea3eb/src/routes/transactions/mappers/common/twap-order.mapper.ts#L129).

## Changes

- Add `activeOrderUid` to `TwapOrderTransactionInfo`.
- Fetch status relative to `activeOrderUid`.


src/routes/transactions/mappers/common/twap-order.mapper.spec.ts
```diff
@@ -680,6 +786,7 @@ describe('TwapOrderMapper', () => {
   });
 
   it('should map the TWAP order if source apps are restricted and a part order fullAppData matches any of the allowed apps', async () => {
+    configurationService.set('swaps.maxNumberOfParts', 2);
```
iamacook, line 789: This test began failing without this.

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -195,37 +193,110 @@ export class TwapOrderMapper {
...
+        owner: args.safeAddress,
+        // If there is no "active" part, we use the last one to get the status
+        order: args.twapParts.slice(-1)[0],
+      });
+
+    // If we already have the order, there is no need to fetch it again
+    if (args.partOrders) {
+      const order = args.partOrders.find((order) => order.uid === orderUid);
+      if (order) {
+        return order.status;
```
compojoom, line 284: Maybe I'm misunderstanding this, but if you just return the order status, wouldn't that be wrong for "fulfilled" orders? If the status is fulfilled for one part, it doesn't mean that this applies to the whole TWAP.

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -195,37 +193,110 @@ export class TwapOrderMapper {
...
+      }
     }
 
-    return OrderStatus.Unknown;
+    try {
+      const order = await this.swapsRepository.getOrder(args.chainId, orderUid);
+      // If the active order is fulfilled, the next one may not yet exist
+      // and fetching it would fail so we assume it's open
+      if (args.activeOrderUid && order.status === OrderStatus.Fulfilled) {
```
schmanu, line 292: Whenever this order exists we have to return the total twap order status as Open.
Even if the activePart is still open or even if it got cancelled (as it is possible to cancel a single part without cancelling the whole order)

src/routes/transactions/mappers/common/twap-order.mapper.ts
iamacook: The logic regarding status generation is different as of https://github.com/safe-global/safe-client-gateway/pull/1785#discussion_r1691083471 which should now solve this.

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -195,37 +193,109 @@ export class TwapOrderMapper {
...
+      const orderAlreadyFetched = !!args.partOrders?.some(
+        (order) => order.uid === orderUidToCheck,
+      );
+      if (!orderAlreadyFetched) {
+        await this.swapsRepository.getOrder(args.chainId, orderUidToCheck);
+      }
+
+      // If the order was created, the TWAP is either open or fulfilled depending
+      // on whether the order part is active
+      return args.activeOrderUid ? OrderStatus.Open : OrderStatus.Fulfilled;
```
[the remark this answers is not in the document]
iamacook: Added in 686505d64d83d75e5f8bfbe39e2571905b479ed6.

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -195,37 +193,114 @@ export class TwapOrderMapper {
...
+    // this does not affect TWAP status
+
+    const finalOrderUid = this.gpv2OrderHelper.computeOrderUid({
+      chainId: args.chainId,
+      owner: args.safeAddress,
+      order: args.twapParts.slice(-1)[0], // Last TWAP part
+    });
+
+    // If active order is final one, it's fulfilled as it exists
+    if (args.activeOrderUid === finalOrderUid) {
```
schmanu, line 281: I think we still have to fetch in order to know that it exists.
`twapParts` and `activeOrderUid` are both just computed here. So if the last twap would be active this will always be true.

src/routes/transactions/mappers/common/twap-order.mapper.ts
```diff
@@ -195,37 +193,114 @@ export class TwapOrderMapper {
...
+      const orderAlreadyFetched = !!args.partOrders?.some(
+        (order) => order.uid === orderUidToCheck,
+      );
+      if (!orderAlreadyFetched) {
+        await this.swapsRepository.getOrder(args.chainId, orderUidToCheck);
+      }
+
+      // If the order exists, the TWAP is either open or fulfilled depending
+      // on whether the order is active
+      return args.activeOrderUid ? OrderStatus.Open : OrderStatus.Fulfilled;
```
schmanu, line 299: I think we could remove the above check (where I commented) and instead do something like this here:
```suggestion
      return args.activeOrderUid ? args.activeOrderUid === finalOrderUid ? OrderStatus.Fulfilled : OrderStatus.Open : OrderStatus.Fulfilled;
```

src/routes/transactions/mappers/common/twap-order.mapper.spec.ts
```diff
@@ -841,6 +948,7 @@ describe('TwapOrderMapper', () => {
     });
 
     expect(result).toEqual({
+      activeOrderUid: null,
       buyAmount: '51576509680023161648',
       buyToken: {
         address: buyToken.address,
```
schmanu, line 954: Should we add these (or similar) testcases for the Order Status mapping?

- Cancelled Order with 3 parts where only the first part exists and the second part is currently active.
- Fulfilled Order with 3 parts where the last part is currently active and the repository returns fulfilled for it.

src/routes/transactions/mappers/common/twap-order.mapper.spec.ts
```diff
@@ -427,7 +430,14 @@ describe('TwapOrderMapper', () => {
...
 
-    mockSwapsRepository.getOrder.mockResolvedValueOnce(part2);
+    mockSwapsRepository.getOrder.mockImplementation(
+      async (_chainId: string, orderUid: string) => {
+        if (orderUid === part2.uid) {
+          return Promise.resolve(part2);
+        }
+        return Promise.reject(new NotFoundException());
+      },
+    );
```
iamacook, line 440: I encountered issues in the new test cases that were due to incorrectly calculated order UIDs. As such, I adjusted these mocks to be more specific.

src/routes/transactions/mappers/common/twap-order.mapper.ts
iamacook: Adjusted in f20a158137e033f1620ec10484a6501a82057e7f.

src/routes/transactions/mappers/common/twap-order.mapper.ts
iamacook: Adjusted in f20a158137e033f1620ec10484a6501a82057e7f.

src/routes/transactions/mappers/common/twap-order.mapper.spec.ts
iamacook: Added in f20a158137e033f1620ec10484a6501a82057e7f.

src/routes/transactions/mappers/common/twap-order.mapper.spec.ts
hectorgomezv: It's weird, I can't replicate the failure, maybe it's a flaky test for some reason. Or maybe the value is set on another test as the configuration is not cleaned between the tests if I'm not wrong. Anyway, I'm good with setting this explictly here.

hectorgomezv (approved):

PR #1795, opened 2024-07-31
Counterfactual Safes deletion endpoints

hectorgomezv:
## Summary
This PR adds a subset of the Counterfactual Safes storage controller endpoints. Specifically:
```
DELETE /v1/accounts/:address/storage/counterfactual-safes/:chainId/:predictedAddress
DELETE /v1/accounts/:address/storage/counterfactual-safes
```

The first one deletes an existing Counterfactual Safe by the creator's account address, the chain ID, and the predicted address of the Safe to be deployed.

The second deletes all the Counterfactual Safes associated with the address, i.e.: the address being the creator of the Counterfactual Safes.

## Changes
- Add `DELETE /v1/accounts/:address/storage/counterfactual-safes/:chainId/:predictedAddress`
- Add `DELETE /v1/accounts/:address/storage/counterfactual-safes`


src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
```diff
@@ -584,4 +584,337 @@ describe('CounterfactualSafesController', () => {
...
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(403);
+
+      expect(accountsRepository.getAccount).not.toHaveBeenCalled();
+      expect(
+        counterfactualSafesDataSource.deleteCounterfactualSafe,
+      ).not.toHaveBeenCalled();
+    });
+
+    it('returns 403 if chain_id is not a valid chain ID', async () => {
```
iamacook, line 742: What do you think about adding another test case that ensure the token is valid if signed on a different chain?

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
```diff
@@ -584,4 +584,337 @@ describe('CounterfactualSafesController', () => {
...
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(403);
+
+      expect(accountsRepository.getAccount).not.toHaveBeenCalled();
+      expect(
+        counterfactualSafesDataSource.deleteCounterfactualSafesForAccount,
+      ).not.toHaveBeenCalled();
+    });
+
+    it('returns 403 if chain_id is not a valid chain ID', async () => {
```
iamacook, line 900: What do you think about adding another test case that ensure the token is valid if signed on a different chain?

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
```diff
@@ -584,4 +584,337 @@ describe('CounterfactualSafesController', () => {
       ).not.toHaveBeenCalled();
     });
   });
+
+  describe('Delete Counterfactual Safe', () => {
+    it('should delete a Counterfactual Safe', async () => {
```
iamacook, line 589: We should add test coverage regarding a datasource error. These assume that deletion was successful.

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
hectorgomezv: I think that is not applicable in this case as the route URL does not contain a Chain ID, right?

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
hectorgomezv: Added in [650d3f6](https://github.com/safe-global/safe-client-gateway/pull/1795/commits/650d3f6adc010330d6a12092afbd3a09e3df7c80) (I've added that test also to  the rest of endpoints)

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
iamacook: If "signed in" on another chain, we still want to return counterfactual setups on others. I'd say it's more of a "preventative" test to ensure failure if this is altered in the future. What do you think?

hectorgomezv: I was giving this another thought and cannot find how to implement the desired test. I mean, `chainId` is only used to compose the `authPayload`:
```
      const chain = chainBuilder().build();
      const authPayloadDto = authPayloadDtoBuilder()
        .with('chain_id', chain.chainId)
        .with('signer_address', address)
        .build();
```

And the CF Safes datasource deletion is mocked:
```
counterfactualSafesDataSource.deleteCounterfactualSafesForAccount.mockResolvedValue();
```

So there's no link between the `AuthPayload.chainId` and the `chainId` of the Counterfactual Safes being deleted. If this is altered in the future (i.e.: a correspondence between both `chainId` properties is required), we would need to add a second argument to the datasource function. I've added an assertion on the datasource call in [3d81c8c](https://github.com/safe-global/safe-client-gateway/pull/1795/commits/3d81c8cfa6b5291f0ced0689adb9419856839261)

Please let me know if you think there's another way to test it! Thanks!

iamacook (approved):

PR #1796, opened 2024-08-01
Counterfactual Safes retrieval by account address endpoint

hectorgomezv:
## Summary
This PR adds a route to the Counterfactual Safes storage controller endpoints:
```
GET /v1/accounts/:address/storage/counterfactual-safes
```

The endpoint gets all Counterfactual Safes by the creator's account address.

## Changes
- Add `GET /v1/accounts/:address/storage/counterfactual-safes`

### DIFF
--- a/src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
+++ b/src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
@@ -138,6 +138,59 @@ describe('CounterfactualSafesController', () => {
         });
     });
 
+    it('should fail if the authPayload does not match the URL address', async () => {
+      const address = getAddress(faker.finance.ethereumAddress());
+      const chain = chainBuilder().build();
+      const authPayloadDto = authPayloadDtoBuilder()
+        .with('chain_id', chain.chainId)
+        .with('signer_address', getAddress(faker.finance.ethereumAddress()))
+        .build();
+      const accessToken = jwtService.sign(authPayloadDto);
+
+      await request(app.getHttpServer())
+        .get(
+          `/v1/accounts/${address}/counterfactual-safes/${chain.chainId}/${getAddress(faker.finance.ethereumAddress())}`,
+        )
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(401);
+
+      expect(accountsRepository.getAccount).not.toHaveBeenCalled();
+      expect(
+        counterfactualSafesDataSource.getCounterfactualSafe,
+      ).not.toHaveBeenCalled();
+    });
+
+    it('should fail if the account does not have the CounterfactualSafes data setting enabled', async () => {
+      const address = getAddress(faker.finance.ethereumAddress());
+      const chain = chainBuilder().build();
+      const authPayloadDto = authPayloadDtoBuilder()
+        .with('chain_id', chain.chainId)
+        .with('signer_address', address)
+        .build();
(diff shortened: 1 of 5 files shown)


src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
```diff
@@ -400,6 +400,261 @@ describe('CounterfactualSafesController', () => {
...
+      const authPayloadDto = authPayloadDtoBuilder()
+        .with('chain_id', chain.chainId)
+        .with('signer_address', address)
+        .build();
+      const accessToken = jwtService.sign(authPayloadDto);
+      const account = accountBuilder().build();
+      const accountDataTypes = [
+        accountDataTypeBuilder()
+          .with('name', AccountDataTypeNames.CounterfactualSafes)
+          .with('is_active', true)
```
iamacook, line 416: Nit: we should add a test to ensure it doesn't return if disabled.

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
```diff
@@ -400,6 +400,261 @@ describe('CounterfactualSafesController', () => {
...
+
+      expect(
+        counterfactualSafesDataSource.getCounterfactualSafesForAccount,
+      ).toHaveBeenCalledWith(account);
+      expect(
+        counterfactualSafesDataSource.getCounterfactualSafesForAccount,
+      ).toHaveBeenCalledTimes(1);
+    });
+
+    it('returns 403 if no token is present', async () => {
```
iamacook, line 506: We need to test the authorisation, e.g. if the token doesn't match the address in the URL.

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
```diff
@@ -400,6 +453,316 @@ describe('CounterfactualSafesController', () => {
...
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(403);
+
+      expect(accountsRepository.getAccount).not.toHaveBeenCalled();
+      expect(
+        counterfactualSafesDataSource.getCounterfactualSafesForAccount,
+      ).not.toHaveBeenCalled();
+    });
+
+    it('returns 403 if chain_id is not a valid chain ID', async () => {
```
iamacook, line 711: What do you think about adding a tests to check the validity of a token from a different chain for these endpoints?

iamacook (approved):

src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
hectorgomezv: I think that should be covered by the base case ([link](https://github.com/safe-global/safe-client-gateway/blob/d503281efc6a0d780e810b1c4093668632e6ae0d/src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts#L459)) as the `AuthPayload` chain ID is not connected in any way with the retrieval process, so in fact, the token is valid for any arbitrary chain, wdyt?

If you think an auth-related test is still needed, I think we could include it in [this refactor](https://github.com/safe-global/safe-client-gateway/pull/1799) if you are OK with it.

iamacook: Indeed it does cover it - thanks for the explanation!

iamacook (approved):

PR #1797, opened 2024-08-01
Allow optional signature when registering notifications

iamacook:
## Summary

It should be possible to register for "public" notifications, e.g. incoming/outgoing assets, without a signing key.

This adapts the `POST` `/v2/register/notifications` endpoint to make the `accessToken` optional and, if not present, register for notifications as a non-owner/delegate with no `signer_address` saved. When notifications are dispatched, if no `signer_address` is present, only "public" notifications are therefore dispatch to the related Firebase token.

## Changes

- Make `notification_subscriptions['signer_address']` column nullable
- Allow upsertion of subscriptions without a signer address, adapting the logic to first clear the subscription if the signer address matches or is undefined in order to "align" subscriptions.
- Add/adjust tests accordingly.


src/domain/notifications/notifications.repository.v2.ts
```diff
@@ -111,6 +106,8 @@ export class NotificationsRepositoryV2 implements INotificationsRepositoryV2 {
     chainId: string;
     safeAddress: `0x${string}`;
   }): Promise<Array<NotificationType>> {
+    // TODO: Is this even possible without an account?
+    // We'd need to theoretically need to store the preferences locally first?
```
iamacook, line 110: We need to discuss this with the clients as preferences (having subscribed _without_ a signature) are initially device-Safe related.

src/datasources/notifications/notifications.datasource.ts
```diff
@@ -59,21 +59,32 @@ export class NotificationsDatasource implements INotificationsDatasource {
...
+              WHERE chain_id = ${safe.chainId}
+              AND safe_address = ${safe.address}
+              AND (
+                signer_address = ${args.signerAddress ?? null}
+                OR (
+                  signer_address IS NULL
+                )
+              )
+              AND push_notification_device_id = ${device.id}
+            `;
```
iamacook, line 75: The subscription is still device based so this shouldn't cause issues of "unsubscriptions".

src/routes/notifications/guards/notifications.guard.ts
```diff
@@ -0,0 +1,25 @@
+import { AuthController } from '@/routes/auth/auth.controller';
+import { AuthGuard } from '@/routes/auth/guards/auth.guard';
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { Request } from 'express';
+
+@Injectable()
+export class NotificationsGuard implements CanActivate {
```
hectorgomezv, line 7: Small nit: what about renaming this to a more general term, like `OptionalAuthGuard`, `OptionalAccessTokenGuard` or simply `AccessTokenGuard`. I guess this guard could be used outside the "notifications" scope, wdyt?

src/routes/notifications/guards/notifications.guard.ts
```diff
@@ -0,0 +1,25 @@
...
+  canActivate(context: ExecutionContext): boolean {
+    const request: Request = context.switchToHttp().getRequest();
+
+    const accessToken: string | undefined =
+      request.cookies[AuthController.ACCESS_TOKEN_COOKIE_NAME];
+
+    /**
+     * If there is no access token, we allow the request to proceed as
+     * users can register for "public" notifications having not signed.
+     */
```
hectorgomezv, line 18: As before, what do you think about making this comment more "purpose agnostic"?

src/routes/notifications/notifications.controller.v2.spec.ts
```diff
@@ -767,7 +767,7 @@ describe('Notifications Controller V2 (Unit)', () => {
     });
 
     describe('authentication', () => {
-      it('should return 403 if no token is present', async () => {
+      it('should return "unknown" subscriptions with no token', async () => {
```
hectorgomezv, line 770: Is this test aligned with the description? The test expects the request to fail with a 403 status if I'm not wrong.

src/domain/notifications/notifications.repository.v2.ts
```diff
@@ -144,7 +139,11 @@ export class NotificationsRepositoryV2 implements INotificationsRepositoryV2 {
     chainId: string;
     safeAddress: `0x${string}`;
   }): Promise<void> {
-    return this.notificationsDatasource.deleteSubscription(args);
+    return this.notificationsDatasource.deleteSubscription({
+      deviceUuid: args.deviceUuid,
+      chainId: args.chainId,
+      safeAddress: args.safeAddress,
+    });
```
hectorgomezv, line 146: Another small nit, but this could remain as `this.notificationsDatasource.deleteSubscription(args)`, couldn't it?

hectorgomezv (commented):
👏🏻

src/routes/notifications/guards/notifications.guard.ts
iamacook: I renamed, moved and added an isolated test for it in da9bf557fe0afb1b1eabea3681f878250e7c90ee.

src/routes/notifications/guards/notifications.guard.ts
iamacook: Updated in da9bf557fe0afb1b1eabea3681f878250e7c90ee.

src/routes/notifications/notifications.controller.v2.spec.ts
iamacook: True, here it expects 403. Updated in da9bf557fe0afb1b1eabea3681f878250e7c90ee.

src/domain/notifications/notifications.repository.v2.ts
iamacook: Changed in da9bf557fe0afb1b1eabea3681f878250e7c90ee.

hectorgomezv (approved):

PR #1800, opened 2024-08-05
Implement Accounts creation rate limit

hectorgomezv:
## Summary
This PR introduces a rate limitation to creating User Accounts using the related `AccountsController` endpoint.

## Changes
- Adds a rate limitation controlled by a cache key before proceeding with account creation.

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -6,6 +6,14 @@ export default () => ({
     version: process.env.APPLICATION_VERSION,
     buildNumber: process.env.APPLICATION_BUILD_NUMBER,
   },
+  accounts: {
+    creationRateLimitPeriodSeconds: parseInt(
+      process.env.ACCOUNT_CREATION_RATE_LIMIT_PERIOD_SECONDS ?? `${60}`,
+    ),
+    creationRateLimitCalls: parseInt(
+      process.env.ACCOUNT_CREATION_RATE_LIMIT_CALLS_BY_PERIOD ?? `${1}`,
+    ),
+  },
   amqp: {
     url: process.env.AMQP_URL || 'amqp://localhost:5672',
     exchange: {
--- a/src/datasources/accounts/accounts.datasource.spec.ts
+++ b/src/datasources/accounts/accounts.datasource.spec.ts
@@ -157,7 +157,6 @@ describe('AccountsDatasource tests', () => {
         }),
       ).rejects.toThrow('Rate limit reached');
 
-      expect(mockLoggingService.warn).toHaveBeenCalledTimes(1);
       const { count } = await sql`SELECT id FROM accounts`;
       expect(count).toBe(accountCreationRateLimitCalls);
     });
@@ -192,7 +191,6 @@ describe('AccountsDatasource tests', () => {
         });
       }
 
-      expect(mockLoggingService.warn).not.toHaveBeenCalled();
       const { count } = await sql`SELECT id FROM accounts`;
       expect(count).toBe(accountsToCreate);
     });
--- a/src/datasources/accounts/accounts.datasource.spec.ts
+++ b/src/datasources/accounts/accounts.datasource.spec.ts
@@ -64,7 +64,10 @@ describe('AccountsDatasource tests', () => {
     it('creates an account successfully', async () => {
       const address = getAddress(faker.finance.ethereumAddress());
 
-      const result = await target.createAccount(address);
+      const result = await target.createAccount({
+        address,
+        clientIp: faker.internet.ipv4(),
+      });
 
       expect(result).toStrictEqual({
         id: expect.any(Number),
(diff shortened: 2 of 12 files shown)


src/config/entities/configuration.ts
```diff
@@ -6,6 +6,14 @@ export default () => ({
...
     buildNumber: process.env.APPLICATION_BUILD_NUMBER,
   },
+  accounts: {
+    creationRateLimitPeriodSeconds: parseInt(
+      process.env.ACCOUNT_CREATION_RATE_LIMIT_PERIOD_SECONDS ?? `${60}`,
+    ),
+    creationRateLimitCalls: parseInt(
+      process.env.ACCOUNT_CREATION_RATE_LIMIT_CALLS_BY_PERIOD ?? `${1}`,
+    ),
+  },
```
hectorgomezv, line 16: These defaults allow each IP to create an account per minute. As the account is cross-chain, I thought that one account should be sufficient for most users.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -52,17 +66,21 @@ export class AccountsDatasource implements IAccountsDatasource, OnModuleInit {
     );
   }
 
-  async createAccount(address: `0x${string}`): Promise<Account> {
+  async createAccount(args: {
+    address: `0x${string}`;
+    clientIp: string;
```
hectorgomezv, line 71: As this function's arity increased to 2, I've wrapped the arguments on a type, which caused some additional changes in the PR.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -192,4 +210,33 @@ export class AccountsDatasource implements IAccountsDatasource, OnModuleInit {
...
+   * If the client IP address is invalid, a warning is logged.
+   * If the client IP address is valid and rate limit is reached, a {@link LimitReachedError} is thrown.
+   *
+   * @param clientIp - client IP address.
+   */
+  private async checkCreationRateLimit(clientIp: string): Promise<void> {
+    if (!clientIp || !AccountsDatasource.SIMPLE_IPV4_REGEX.test(clientIp)) {
+      this.loggingService.warn(
+        `Invalid client IP while creating account: ${clientIp}`,
+      );
```
hectorgomezv, line 226: If the IP is not defined or it's not well formatted, the account creation is allowed. This is mostly done to avoid blockers, but I think we need to keep an eye on logs if we start seeing too many warnings here.

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -192,4 +210,33 @@ export class AccountsDatasource implements IAccountsDatasource, OnModuleInit {
...
+      const current = await this.cacheService.increment(
+        CacheRouter.getRateLimitCacheKey(
+          `${AccountsDatasource.ACCOUNT_CREATION_CACHE_PREFIX}_${clientIp}`,
+        ),
+        this.accountCreationRateLimitPeriodSeconds,
+      );
+      if (current > this.accountCreationRateLimitCalls) {
+        this.loggingService.warn(
+          `Limit of ${this.accountCreationRateLimitCalls} reached for IP ${clientIp}`,
+        );
```
hectorgomezv, line 237: As before, I added a log line here also, as getting too many rate limit conditions shouldn't be normal.

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -88,20 +91,126 @@ describe('AccountsDatasource tests', () => {
       );
     });
 
-    it('throws when an account with the same address already exists', async () => {
+    it('creates an account successfully if the clientIp is not a valid IP', async () => {
```
iamacook, line 94:
```suggestion
    it('creates an account successfully if the clientIp is a valid IP', async () => {
```

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -88,20 +91,126 @@ describe('AccountsDatasource tests', () => {
...
+      await expect(
+        target.createAccount({
+          address: getAddress(faker.finance.ethereumAddress()),
+          clientIp,
+        }),
+      ).rejects.toThrow('Rate limit reached');
+
+      expect(mockLoggingService.warn).toHaveBeenCalledTimes(1);
+      const { count } = await sql`SELECT id FROM accounts`;
+      expect(count).toBe(accountCreationRateLimitCalls);
```
iamacook, line 162: Should this allow creation of accounts with the same address? (I won't comment regarding this below.)

src/datasources/accounts/accounts.datasource.spec.ts
```diff
@@ -88,20 +91,126 @@ describe('AccountsDatasource tests', () => {
...
+      }
+
+      await expect(
+        target.createAccount({
+          address: getAddress(faker.finance.ethereumAddress()),
+          clientIp,
+        }),
+      ).rejects.toThrow('Rate limit reached');
+
+      expect(mockLoggingService.warn).toHaveBeenCalledTimes(1);
```
iamacook, line 160: I'm impartial regarding this but I don't think it's imperative we check that logging occurred. (I won't comment regarding this below.)

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -25,7 +26,13 @@ import postgres from 'postgres';
 
 @Injectable()
 export class AccountsDatasource implements IAccountsDatasource, OnModuleInit {
+  private static readonly SIMPLE_IPV4_REGEX = /^(\d{1,3}\.){3}\d{1,3}$/;
```
iamacook, line 29: We could use zod to validate this.

iamacook (commented):
I'm not sure if this should be happening in the datasource. Rate limiting is more guard behaviour to me. What do you think?

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: According to [the unique constraint we have](https://github.com/safe-global/safe-client-gateway/blob/7bfdc2aa4c803fa22b652049cc028630128e9b40/migrations/00001_accounts/index.sql#L18), it shouldn't be allowed (that's why on each account creation `faker` is called to randomize the address).

hectorgomezv:
> I'm not sure if this should be happening in the datasource. Rate limiting is more guard behaviour to me. What do you think?

I don't have a super firm opinion, but I think it makes sense to have the limiter on the datasource for this use case as we have to limit the actual account creations. I guess if we add the rate limiter to the controller layer it wouldn't consider failed attempts for example. For a GET endpoint, I'd be more in favor of limiting the endpoint directly.

Also, I considered that even if we put this on a guard, it would need to have access to the Redis instance, as the rate limiter count needs to be shared among the CGW pods (maybe this could help if we want to do it https://github.com/kkoomen/nestjs-throttler-storage-redis).

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: Removed in [b2e5161](https://github.com/safe-global/safe-client-gateway/pull/1800/commits/b2e51613dcbc65252b8e6f2865027a72adc25644)

src/datasources/accounts/accounts.datasource.ts
hectorgomezv: Added in [b2e5161](https://github.com/safe-global/safe-client-gateway/pull/1800/commits/b2e51613dcbc65252b8e6f2865027a72adc25644)

iamacook:
> I don't have a super firm opinion, but I think it makes sense to have the limiter on the datasource for this use case as we have to limit the actual account creations. I guess if we add the rate limiter to the controller layer it wouldn't consider failed attempts for example. For a GET endpoint, I'd be more in favor of limiting the endpoint directly.
> 
> Also, even if we put this on a Guard, it would need to have access to the Redis instance, as the rate limiter count needs to be shared among the CGW pods (maybe this could help if we want to do it https://github.com/kkoomen/nestjs-throttler-storage-redis).

Considering we want to limit _creations_ and this can happen via various methods - I am happy with keeping it as it is. Although a nit, I think we should add a comment explaining the above.

src/datasources/accounts/accounts.datasource.spec.ts
hectorgomezv: I guess we will see how this goes through time, I don't think user can obfuscate their IPs, but if they do, I think we can start allowing account creation but keep an eye on logging to see if this (a malformed IP reach the service) happens often. So I'd leave it as it is for now, and maybe reconsider it based on the logs.

hectorgomezv:
> Considering we want to limit _creations_ and this can happen via various methods - I am happy with keeping it as it is. Although a nit, I think we should add a comment explaining the above.

Great, thanks! I've included a note in [031aaca](https://github.com/safe-global/safe-client-gateway/pull/1800/commits/031aacadcb8d2347b7780e4ecafa8ece057eb771), please let me know if you think it's not clearer enough, or if you would like to rephrase it.

iamacook (approved):

src/datasources/accounts/accounts.datasource.ts
parius: In my understanding, it cannot happen.

you get this IP address from web server, who gets it from http(s) which is over tpc, where server always has to know the address on the other side.

even with http3, the udp packets would still contain originators ip. To become empty or malformed, it would need to be specially crafted.

So for this case I would propose to throw an error as well

hectorgomezv: Makes sense to me, I've added a ticket to change it: https://github.com/safe-global/safe-client-gateway/issues/1838

Thanks!

src/datasources/accounts/accounts.datasource.ts
```diff
@@ -192,4 +210,38 @@ export class AccountsDatasource implements IAccountsDatasource, OnModuleInit {
...
+        CacheRouter.getRateLimitCacheKey(
+          `${AccountsDatasource.ACCOUNT_CREATION_CACHE_PREFIX}_${clientIp}`,
+        ),
+        this.accountCreationRateLimitPeriodSeconds,
+      );
+      if (current > this.accountCreationRateLimitCalls) {
+        this.loggingService.warn(
+          `Limit of ${this.accountCreationRateLimitCalls} reached for IP ${clientIp}`,
+        );
+        throw new LimitReachedError();
```
parius, line 243: Which http code would a client get?
this Error is not extending HttpException. Is it necessary to implement an ExceptionFilter for this?

hectorgomezv: With this implementation, the client would get a generic 500 status code error.

I've created a PR to extend `HttpException` and set the correct status codes for both accounts and Counterfactual Safes rate limit hit errors on creation: https://github.com/safe-global/safe-client-gateway/pull/1865

Thank you!

PR #1801, opened 2024-08-05
Implement Counterfactual Safes creation rate limit

hectorgomezv:
## Summary
This PR introduces a rate limitation to creating Counterfactual Safes using the related `CounterfactualSafesController` endpoint. This rate limit is applied by taking the `Account` associated with the Counterfactual Safe to calculate the limit threshold.

## Changes
- Adds a rate limitation controlled by a cache key before proceeding with Counterfactual Safes creation.


src/datasources/accounts/counterfactual-safes/counterfactual-safes.datasource.spec.ts
```diff
@@ -88,6 +88,118 @@ describe('CounterfactualSafesDatasource tests', () => {
...
+      const address = getAddress(faker.finance.ethereumAddress());
+      const [account] = await sql<
+        Account[]
+      >`INSERT INTO accounts (address) VALUES (${address}) RETURNING *`;
+      const creationRateLimitCalls = faker.number.int({ min: 5, max: 10 });
+      const createCounterfactualSafes = Array.from(
+        { length: creationRateLimitCalls },
+        (_, i) =>
+          createCounterfactualSafeDtoBuilder()
+            .with('chainId', i.toString())
```
hectorgomezv, line 163: `chainId` is set as a discriminator to avoid the DTO properties collision edge case.

iamacook (commented):
As with the [other implementation for accounts](https://github.com/safe-global/safe-client-gateway/pull/1800#pullrequestreview-2220887161), this seems more guard-like behaviour? It might even allow us to reduce repetition if it was in a guard. What do you think?

hectorgomezv:
> As with the [other implementation for accounts](https://github.com/safe-global/safe-client-gateway/pull/1800#pullrequestreview-2220887161), this seems more guard-like behaviour? It might even allow us to reduce repetition if it was in a guard. What do you think?

I've answered in https://github.com/safe-global/safe-client-gateway/pull/1800#issuecomment-2271229334, we can follow the discussion there if you are happy with it.

iamacook (dismissed):

iamacook (approved):

PR #1809, opened 2024-08-06
Assign default value for `confirmationsRequired`

iamacook:
## Summary

We are occasionally seeing `confirmationsRequired` of multisig transactions returned as `null` by the Transaction Service. As we [expect it to be an integer](https://github.com/safe-global/safe-client-gateway/blob/default-confirmations-required/src/domain/safe/entities/multisig-transaction.entity.ts#L49), this therefore fails validation.

As this causes transactions not to be returned, this implements fallback values should the value be `null`:

- Safe threshold for queued transactions
- Confirmations required (if present) or Safe threshold for historical transactions

Note: the above is a [known bug](https://github.com/safe-global/safe-transaction-service/issues/2170) that should eventually be tackled on the Transaction Service, after which we intend to revert the above.

## Changes

- Fallback to the default values when fetching singular/a list of multisig transactions, or all transactions

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.ts
+++ b/src/datasources/transaction-api/transaction-api.service.ts
@@ -29,6 +29,7 @@ import { AddConfirmationDto } from '@/domain/transactions/entities/add-confirmat
 import { ProposeTransactionDto } from '@/domain/transactions/entities/propose-transaction.dto.entity';
 import { ILoggingService } from '@/logging/logging.interface';
 import { get } from 'lodash';
+import { getAddress } from 'viem';
 
 export class TransactionApi implements ITransactionApi {
   private static readonly ERROR_ARRAY_PATH = 'nonFieldErrors';
@@ -607,6 +608,7 @@ export class TransactionApi implements ITransactionApi {
   }
 
   async getMultisigTransactions(args: {
+    // TODO: safeAddress and to should be `0x${string}`
     safeAddress: string;
     ordering?: string;
     executed?: boolean;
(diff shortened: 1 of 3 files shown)


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -626,28 +631,48 @@ export class TransactionApi implements ITransactionApi {
...
+          });
+          if (hasConfirmationsRequired) {
+            return data;
+          }
+
+          const results = await Promise.all(
+            data.results.map(async (tx) => {
+              return await this._setConfirmationsRequired(tx);
+            }),
+          );
```
hectorgomezv, line 669: Nit: this iterates `data.results` twice when there is an item with `confirmationsRequired === null`, and also executes `_setConfirmationsRequired(tx)` for all the transactions in the page.

No strong requirement, but maybe we can optimize it a bit by:
```javascript
        .then(
          async (data): Promise<Page<MultisigTransaction>> => ({
            ...data,
            results: await Promise.all(
              data.results.map(async (tx) =>
                tx.confirmationsRequired !== null
                  ? tx
                  : await this._setConfirmationsRequired(tx),
              ),
            ),
          }),
        );
```

This would also simplify the `_setConfirmationRequired` implementation because the initial check can be removed (and the input type) as the function would only receive transactions affected by the bug:

And a final adjustment could be also done in the other `_setConfirmationRequired` call in line 705:
```javascript
        .then(async (tx) =>
          transaction.confirmationsRequired !== null
            ? tx
            : await this._setConfirmationsRequired(tx),
        );
```

Again, this is just a nit, feel free to keep the current implementation if you think is clearer!

hectorgomezv (approved):

src/datasources/transaction-api/transaction-api.service.ts
iamacook: I adjusted it in all three instances in 09995bb3079d9c9be8f8cd0a5a654c1cf3295963.

hectorgomezv (approved):

PR #1811, opened 2024-08-06
Generate SDK

iamacook:
## Summary

This generates a TypeScript SDK for all routes of the project from the staging Swagger deployment.

Note: this will be moved to the repo. It was initially done so here with the intention of getting the definitions from the project.


iamacook:
I'm going to close this as generation will move to the SDK repo. itself. Initially, the idea was to get the schema from the project itself but due to env. vars. we are instead scraping it. It theremore makes sense to locate it alongside the SDK generated instead of pushing into a separate repo.

PR #1825, opened 2024-08-15
Add caching layer to RPC requests

iamacook:
## Summary

For each RPC call we currently make a request. As we intend to fetch more data on-chain, e.g. staking details, indexed block timestamp, this implements a caching layer for all RPC requests to reduce this number.

## Changes

- Add `expirationTimeInSeconds.rpc` with a default valut of 15 seconds, modifying via the `EXPIRATION_TIME_RPC_SECONDS` env. var.
- Change `createClient` to create a `PublicClient` with a caching layer on all RPC requests.
  - Add new `rpc_requests` cache key.
- Add relative tests.

### DIFF
--- a/src/datasources/blockchain/blockchain-api.manager.spec.ts
+++ b/src/datasources/blockchain/blockchain-api.manager.spec.ts
@@ -1,11 +1,13 @@
 import { FakeConfigurationService } from '@/config/__tests__/fake.configuration.service';
 import { BlockchainApiManager } from '@/datasources/blockchain/blockchain-api.manager';
 import { FakeCacheService } from '@/datasources/cache/__tests__/fake.cache.service';
+import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 import { chainBuilder } from '@/domain/chains/entities/__tests__/chain.builder';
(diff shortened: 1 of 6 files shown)


src/datasources/blockchain/blockchain-api.manager.spec.ts
```diff
@@ -79,10 +91,54 @@ describe('BlockchainApiManager', () => {
...
+          ok: true,
+          status: 200,
+          json: () => {
+            // Return chain ID
+            return Promise.resolve({
+              result: faker.string.hexadecimal(),
+            });
+          },
+        } as Response);
+      });
```
iamacook, line 119: I'd like feedback on this - ihis goes against how we normally test but I think it's better than mocking viem.

hectorgomezv: I'm OK with spying `fetch` instead of `viem` in this case, I think there's no advantage in one vs. the other, right? Spying `fetch` seems a bit more _e2e-like_. Btw, what do you think about also checking the cache contains the expected key/values here?

hectorgomezv (approved):

hectorgomezv (approved):

PR #1842, opened 2024-08-23
Add StakingStatus to native staking deposit transactions

hectorgomezv:
## Summary
This PR adds `StakingStatus` enum values, and maps this status for a staking `deposit` transaction type, according to its execution timestamp. The implemented mapping follows these rules: 
   * If the deposit transaction is not confirmed, the status is `SignatureNeeded`.
   * If the deposit transaction is confirmed but the deposit execution date is not available, the status is `AwaitingEntry`.
   * If the deposit execution date is available, the status is `AwaitingEntry` if the current date is before the estimated entry time, otherwise, the status is `Validating`.
   * If the status cannot be determined, the status is `Unknown`.

## Changes
- Adds `StakingStatus` values and mappings.

### DIFF
--- a/src/routes/transactions/mappers/common/transaction-info.mapper.ts
+++ b/src/routes/transactions/mappers/common/transaction-info.mapper.ts
@@ -322,9 +322,16 @@ export class MultisigTransactionInfoMapper {
     }
 
     try {
+      const tx = transaction as MultisigTransaction;
+      const isConfirmed =
+        !!tx.confirmations &&
+        tx.confirmations.length >= tx.confirmationsRequired;
+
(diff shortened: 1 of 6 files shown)


src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -322,9 +322,16 @@ export class MultisigTransactionInfoMapper {
     }
 
     try {
+      const tx = transaction as MultisigTransaction;
```
iamacook, line 325: Are we sure this cannot be a `ModuleTransaction`? We could use the validation schema instead of casting it, for example.

PR #1864, opened 2024-08-27
Add annual/monthly fiat rewards calculation

### LINKED ISSUE #1863: Add Native Staking annual/monthly expected fiat rewards
### Description

The CGW aims to provide the expected annual/monthly rewards in fiat (USD) associated with a stake transaction.

### Requirements

Include the following fields in `NativeStakingDepositConfirmationView`:
- `expectedFiatAnnualReward`: it will be calculated as `expectedAnnualReward * nativeCurrencyFiatPrice / nativeCurrencyDecimalPlaces`.
- `expectedFiatMonthlyReward`: it will be calculated as `expectedMonthlyReward * nativeCurrencyFiatPrice / nativeCurrencyDecimalPlaces`.

hectorgomezv:
Closes #1863 

## Summary
This PR adds fiat rewards calculation to Native Staking Transactions confirmation views.

**NOTE:** The fiat prices are given in USD.

## Changes
- Add `expectedFiatAnnualReward` and `expectedFiatMonthlyReward` to `NativeStakingDepositConfirmationView`.
- Adds a new `getNativeCoinPrice` method to `IBalancesApi` and `IBalancesRepository`, to make it possible to retrieve the native coin fiat price in USD.

### DIFF
--- a/src/routes/transactions/transactions-view.service.ts
+++ b/src/routes/transactions/transactions-view.service.ts
@@ -1,4 +1,6 @@
 import { IConfigurationService } from '@/config/configuration.service.interface';
+import { IBalancesRepository } from '@/domain/balances/balances.repository.interface';
+import { IChainsRepository } from '@/domain/chains/chains.repository.interface';
 import { IDataDecodedRepository } from '@/domain/data-decoder/data-decoded.repository.interface';
 import { DataDecoded } from '@/domain/data-decoder/entities/data-decoded.entity';
 import { ComposableCowDecoder } from '@/domain/swaps/contracts/decoders/composable-cow-decoder.helper';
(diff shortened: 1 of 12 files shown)


src/routes/transactions/transactions-view.service.ts
```diff
@@ -287,12 +293,25 @@ export class TransactionsViewService {
       depositExecutionDate: null,
     });
     const value = args.value ? Number(args.value) : 0;
+    const chain = await this.chainsRepository.getChain(args.chainId);
+    const nativeCoinPrice =
```
iamacook, line 297: The `NetworkStats` from Kiln include a `eth_price_usd` value that we could use instead of fetching this. If we cache the response, it should save the extra request. What do you think?

hectorgomezv: Oh, I didn't see that! Excellent point, will change it! Thanks!

PR #1871, opened 2024-08-28
Filter out invalid Chain items

hectorgomezv:
## Summary
When using `buildPageSchema` to validate a `Page<T>` of items, all the items in the `Page<T>` must conform to the `T` schema. Otherwise, the validation of the `Page<T>` fails.

This PR adds a `buildLenientPageSchema`, that filters out the items that do not pass the validation, instead of throwing an error.

## Changes
- Adds `buildLenientPageSchema` function.
- Changes `ChainsRepository.getChains` function to use a _lenient_ page validation, and log out the failing items.

### DIFF
--- a/src/domain/entities/schemas/lenient-page.schema.factory.ts
+++ b/src/domain/entities/schemas/lenient-page.schema.factory.ts
@@ -1,0 +1,21 @@
+import { TypeOf, z } from 'zod';
+import { Page } from '@/domain/entities/page.entity';
+
+/**
+ * Builds a lenient page schema that filters out invalid items from the results array.
+ */
+export function buildLenientPageSchema<T extends z.ZodTypeAny>(
+  itemSchema: T,
+): z.ZodType<Page<z.infer<T>>> {
+  return z.object({
+    count: z.number().nullable(),
+    next: z.string().nullable(),
+    previous: z.string().nullable(),
+    results: z.array(z.any()).transform((items) =>
+      items.flatMap((item) => {
+        const result = itemSchema.safeParse(item);
+        return result.success ? [result.data as TypeOf<T>] : [];
+      }),
+    ),
+  });
+}
--- a/src/routes/chains/chains.controller.spec.ts
+++ b/src/routes/chains/chains.controller.spec.ts
@@ -200,7 +200,7 @@ describe('Chains Controller (Unit)', () => {
       });
     });
 
-    it('Failure: received data is not valid', async () => {
+    it('should exclude items not passing validation', async () => {
       networkService.get.mockResolvedValueOnce({
         data: {
           ...chainsResponse,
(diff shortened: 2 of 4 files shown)


src/domain/entities/schemas/lenient-page.schema.factory.ts
```diff
@@ -0,0 +1,21 @@
...
+import { Page } from '@/domain/entities/page.entity';
+
+/**
+ * Builds a lenient page schema that filters out invalid items from the results array.
+ */
+export function buildLenientPageSchema<T extends z.ZodTypeAny>(
+  itemSchema: T,
+): z.ZodType<Page<z.infer<T>>> {
+  return z.object({
+    count: z.number().nullable(),
```
iamacook, line 11: Nit: this includes any erroneous `results` items in the `count`. Do you think we should reduce it accordingly?

src/domain/entities/schemas/lenient-page.schema.factory.ts
```diff
@@ -0,0 +1,21 @@
...
+    next: z.string().nullable(),
+    previous: z.string().nullable(),
+    results: z.array(z.any()).transform((items) =>
+      items.flatMap((item) => {
+        const result = itemSchema.safeParse(item);
+        return result.success ? [result.data as TypeOf<T>] : [];
```
iamacook, line 17: Do you think we should log details of those items that failed validation? We also don't need to cast `result.data`.

src/routes/chains/chains.controller.spec.ts
```diff
@@ -200,7 +200,7 @@ describe('Chains Controller (Unit)', () => {
       });
     });
 
-    it('Failure: received data is not valid', async () => {
```
iamacook, line 203: We should keep this as there may be issue with non-`results` as well.

src/domain/entities/schemas/lenient-page.schema.factory.ts
hectorgomezv: Sure! Good point!

src/domain/entities/schemas/lenient-page.schema.factory.ts
hectorgomezv: Those are being logged here (on the caller): https://github.com/safe-global/safe-client-gateway/blob/17054280c68c28802ff5ea29131c418a48251b74/src/domain/chains/chains.repository.ts#L41

I thought it was better to let the caller decide about logging than doing it into the schema builder function, wdyt?

iamacook: Makes sense to me!

src/domain/entities/schemas/lenient-page.schema.factory.ts
iamacook: I've addressed this in a [separate PR.](https://github.com/safe-global/safe-client-gateway/pull/1881)

src/routes/chains/chains.controller.spec.ts
iamacook: I added this back in a [separate PR](https://github.com/safe-global/safe-client-gateway/pull/1881).

PR #1892, opened 2024-09-03
Fix coverage reports configuration

hectorgomezv:
## Summary
Since the `rootDir` change in [this PR](https://github.com/safe-global/safe-client-gateway/commit/50d342ef7961fd6f7ee49f1c5b6d9b6d724dc635#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519L85), the coverage reports are taking into consideration all the folders in the root, instead of `/src` only. This PR restores this behavior by ignoring non-relevant folders regarding test coverage.

## Changes
- Adds non-relevant test coverage folders to `coveragePathIgnorePatterns` configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -89,6 +89,11 @@
       "^src/**/*.(t|j)s"
     ],
     "coverageDirectory": "../coverage",
+    "coveragePathIgnorePatterns": [
+      "main.ts",
+      ".+\\/__tests__\\/.+\\.factory.(t|j)s",
+      ".e2e-spec.ts"
+    ],
     "testEnvironment": "node",
     "moduleNameMapper": {
       "^@/abis/(.*)$": "<rootDir>/abis/$1",
--- a/package.json
+++ b/package.json
@@ -86,13 +86,19 @@
       "^.+\\.(t|j)s$": "ts-jest"
     },
     "collectCoverageFrom": [
-      "^src/**/*.(t|j)s"
+      "**/*.(t|j)s"
     ],
     "coverageDirectory": "../coverage",
     "coveragePathIgnorePatterns": [
-      "main.ts",
       ".+\\/__tests__\\/.+\\.factory.(t|j)s",
-      ".e2e-spec.ts"
+      "/abis/",
+      "/assets/",
+      "/coverage/",
+      "/data/",
+      "/dist/",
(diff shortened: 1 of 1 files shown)


iamacook (commented):
The [coverage report](https://github.com/safe-global/safe-client-gateway/pull/1892#issuecomment-2325981401) still shows 46%. Do we first need to merge it?

hectorgomezv:
> The [coverage report](https://github.com/safe-global/safe-client-gateway/pull/1892#issuecomment-2325981401) still shows 46%. Do we first need to merge it?

You're right, I was under the impression that including the `src` reference into the `collectCoverageFrom` property worked, but that's not the case. I reverted to a previous commit where I was _manually_ ignoring some folders: https://github.com/safe-global/safe-client-gateway/pull/1892/commits/c9e9e824824a7270b44f7456d48e67b9619c8977

This is the change: [9de1100](https://github.com/safe-global/safe-client-gateway/pull/1892/commits/9de1100901043c8ad0b4cf9b71f396a8e3978da3)

It's less clear, but works. Another idea would be to just restrict the coverage analysis to `.ts` files (instead of both `.ts` and `.js`). Any preference?

iamacook:
> > The [coverage report](https://github.com/safe-global/safe-client-gateway/pull/1892#issuecomment-2325981401) still shows 46%. Do we first need to merge it?
> 
> You're right, I was under the impression that including the `src` reference into the `collectCoverageFrom` property worked, but that's not the case. I reverted to a previous commit where I was _manually_ ignoring some folders: [c9e9e82](https://github.com/safe-global/safe-client-gateway/commit/c9e9e824824a7270b44f7456d48e67b9619c8977)
> 
> This is the change: [9de1100](https://github.com/safe-global/safe-client-gateway/pull/1892/commits/9de1100901043c8ad0b4cf9b71f396a8e3978da3)
> 
> It's less clear, but works. Another idea would be to just restrict the coverage analysis to `.ts` files (instead of both `.ts` and `.js`). Any preference?

As it's not touched often, I don't think the clarity is a huge concern. I'm still not seeing an updated coverage report though 🙈

iamacook (approved):

PR #1897, opened 2024-09-04
Add KilnApi caching

hectorgomezv:
## Summary
This PR adds a caching layer to `KilnApi`, by switching its `NetworkService` dependency to `CacheFirstDatasource`. All the endpoints in `KilnApi` are now cached after a successful call, for `EXPIRATION_TIME_STAKING_SECONDS` seconds (defaulting to 60 seconds), without a programmatic invalidation mechanism.

## Changes
- Changes the way `KilnApi` interacts with the network by changing `INetworkService` to `CacheFirstDataSource`, and therefore adding caching.
- Adds `CacheDirs` for all the `KilnApi` endpoints to `CacheRouter`.

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -1,9 +1,9 @@
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 
 export class CacheRouter {
-  private static readonly ACCOUNT_KEY = 'account';
   private static readonly ACCOUNT_DATA_SETTINGS_KEY = 'account_data_settings';
   private static readonly ACCOUNT_DATA_TYPES_KEY = 'account_data_types';
+  private static readonly ACCOUNT_KEY = 'account';
   private static readonly ALL_TRANSACTIONS_KEY = 'all_transactions';
   private static readonly AUTH_NONCE_KEY = 'auth_nonce';
   private static readonly BACKBONE_KEY = 'backbone';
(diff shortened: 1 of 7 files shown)


iamacook (approved):

src/datasources/cache/cache.router.ts
```diff
@@ -554,4 +564,49 @@ export class CacheRouter {
...
+   * Calculate cache directory for staking stakes.
+   *
+   * Note: This function hashes the validators public keys to keep the
+   * cache key short and deterministic. The probability of collision is
+   * 0.00000271% for each 1,000,000 sets of validators public keys.
+   *
+   * @param validatorsPublicKeys - Array of validators public keys
+   * @returns {@link CacheDir} - Cache directory
+   */
+  static getStakingStakesCacheDir(
```
hectorgomezv, line 604: To create a cache key for this, I've opted for hashing the `validatorsPublicKeys` to keep the key short (see function documentation), instead of simply concatenating the keys to build the key. Longer keys can harm Redis performance, and the amount of keys this function could receive is not bounded. Lmk if you think there's a better way to handle this.

src/datasources/cache/cache.router.ts
```diff
@@ -554,4 +564,47 @@ export class CacheRouter {
...
+      `${args.chainId}_${this.STAKING_DEFI_VAULT_STATS_KEY}_${args.vault}`,
+      '',
+    );
+  }
+
+  /**
+   * Calculate cache directory for staking stakes.
+   *
+   * Note: This function hashes the validators public keys to keep the
+   * cache key short and deterministic.
```
iamacook, line 598: Nit: I suggest we add a _why_ here for future reference.

iamacook (approved):

src/datasources/cache/cache.router.ts
hectorgomezv: Sure, good point! I've tried to clarify it in [dfefc67](https://github.com/safe-global/safe-client-gateway/pull/1897/commits/dfefc678f5daea42ebc3cf85a592c5f1acd90db1)

Let me know if you think it is not enough or if it's not clear.

iamacook (approved):

PR #1903, opened 2024-09-06
Add Native Staking transaction data decoders

hectorgomezv:
## Summary
As part of the Native Staking transaction mapping implementation in the CGW, this PR adds _local_ decoding for these transactions. 

The point of this PR is to mitigate the risk of not having the Kiln contracts available (or the data-decoding endpoint unreachable) on the Safe Transaction Service (the main source for transaction data decoding). This PR provides a fallback for that scenario.

## Changes
- Adds `KilnDecoder` along with the Kiln Native Staking contract ABIs.
- Adds a fallback to `TransactionsViewService` that tries to decode the data in case the Transaction Service cannot.

### DIFF
--- a/src/domain/staking/contracts/decoders/kiln-decoder.helper.ts
+++ b/src/domain/staking/contracts/decoders/kiln-decoder.helper.ts
@@ -1,0 +1,124 @@
+import { AbiDecoder } from '@/domain/contracts/decoders/abi-decoder.helper';
+import { LoggingService, ILoggingService } from '@/logging/logging.interface';
+import { Inject, Injectable } from '@nestjs/common';
+
+export const KilnAbi = [
+  {
+    inputs: [],
+    name: 'deposit',
+    outputs: [],
+    stateMutability: 'payable',
+    type: 'function',
+  },
+  {
+    inputs: [{ internalType: 'bytes', name: '_publicKeys', type: 'bytes' }],
+    name: 'requestValidatorsExit',
+    outputs: [],
+    stateMutability: 'nonpayable',
+    type: 'function',
+  },
+  {
+    inputs: [{ internalType: 'bytes', name: '_publicKeys', type: 'bytes' }],
+    name: 'batchWithdrawCLFee',
+    outputs: [],
+    stateMutability: 'nonpayable',
+    type: 'function',
+  },
+] as const;
+
+export type KilnRequestValidatorsExitParameters = {
+  name: '_publicKeys';
+  type: 'bytes';
+  value: `0x${string}`;
+  valueDecoded: null;
+};
+export type KilnBatchWithdrawCLFeeParameters =
+  KilnRequestValidatorsExitParameters;
+
+@Injectable()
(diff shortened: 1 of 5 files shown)


src/domain/staking/contracts/decoders/kiln-decoder.helper.ts
```diff
@@ -0,0 +1,124 @@
...
+    }
+    try {
+      const decoded = this.decodeFunctionData({ data });
+      if (decoded.functionName !== 'deposit') {
+        throw new Error('Data is not of deposit type');
+      }
+      return {
+        method: decoded.functionName,
+        parameters: [],
+      };
```
iamacook, line 60: Nit: each decoder is method specific so I don't think we need expose the `functionName`/can just return the parameters. What do you think?

src/domain/staking/contracts/decoders/kiln-decoder.helper.ts
```diff
@@ -0,0 +1,124 @@
...
+] as const;
+
+export type KilnRequestValidatorsExitParameters = {
+  name: '_publicKeys';
+  type: 'bytes';
+  value: `0x${string}`;
+  valueDecoded: null;
+};
+export type KilnBatchWithdrawCLFeeParameters =
+  KilnRequestValidatorsExitParameters;
```
iamacook, line 36: I suggest we keep the decoded values as is and format it to match the backend response where it is used. What do you think?

src/routes/transactions/transactions-view.service.ts
```diff
@@ -61,12 +63,16 @@ export class TransactionsViewService {
...
-        };
+        const { data } = args.transactionDataDto;
+        const fallbackDataDecoded =
+          this.kilnDecoder.decodeDeposit(data) ??
+          this.kilnDecoder.decodeValidatorsExit(data) ??
+          this.kilnDecoder.decodeBatchWithdrawCLFee(data);
+        if (!fallbackDataDecoded) {
+          throw new Error('Transaction data could not be decoded');
+        }
+        return fallbackDataDecoded;
```
iamacook, line 75: Do you think we should assign the fallback value here, or in the mappings? The reason being, `deposit` is a very generic signature and could be returned for non-Kiln interactions.

I suggest we also add a test case for "generic" signatures as well. What do you think?

hectorgomezv: I've moved the fallback decoding to the specific private function for each transaction type in [3fe322c](https://github.com/safe-global/safe-client-gateway/pull/1903/commits/3fe322ce083219a23b36488475b54d71191bc49b)

src/domain/staking/contracts/decoders/kiln-decoder.helper.ts
hectorgomezv: After an offline sync, we decided to keep the `method` property in the return type. I've done so in [8fdaed7](https://github.com/safe-global/safe-client-gateway/pull/1903/commits/8fdaed7ac25cd5458e6c188bd282f238fd9ad3ba)

iamacook (approved):

Issue #1024, opened 2024-01-15
Add support for client timezone information


fmrsabino:
If the user is in a timezone that follows daylight saving, the transaction history grouping is done without taking it into consideration.

# Problem

If a user is in a  UTC+1 timezone (winter) and UTC+2 (summer), a transaction done during daylight savings will be incorrectly grouped when daylight savings are not observed.

When daylight savings is not being observed, the clients send a UTC+1 difference in milliseconds (3600000 ms). The Safe Client Gateway will then create transaction groups taking only the `3600000` shift into consideration.

If the clients use a solution that takes daylight savings into consideration, this might cause some transactions to be rendered with UTC+2 and potentially making the group that was created with UTC+1 to contain transactions of the following/previous day.

# Example

- TX1 done on `Sun Jul 30 2023 22:46:47 UTC+0000`.
- TX2 done on `Sun Jul 30 2023 21:46:47 UTC+0000`
- Client on UTC+1 Winter, UTC+2 Summer
- Client is on Winter timezone

1. Client sends timezone offset of +1 hour
2. CGW computes groups with 1 hour difference.
3. CGW returns a group with July 30 (22:46:47 UTC+0000) with transactions TX1 and TX2
4. Clients, aware of daylight savings, shift the header label to July 31 (00:46:47 UTC+2) that will include TX1 and TX2 while there should be two groups rendered: July 31 with TX1 and July 30 with TX2.

The transaction was therefore shifted to a different day on the client side while it should've been grouped correctly on the CGW side.

# Expected Result

Given the same example as the one above:
1. Client sends timezone offset of +1 hour
2. CGW computes groups with 1 hour difference **and takes into consideration daylight savings**. 
3. CGW returns two groups:
  * July 30 (22:46:47 UTC+0000) with TX1 
  * July 30 (21:46:47 UTC+0000) with TX2
5. Clients, aware of daylight savings, shift the first header label to July 31 (00:46:47 UTC+2) which includes TX1.

DmitryBespalov:
From https://github.com/safe-global/safe-client-gateway/issues/924#issuecomment-2270889249:

> Researching the UTC and DST (daylight saving time), it seems that the best way to resolved the issue with DST is to use the `timezone ID`, like `Europe/Paris` instead of the `timezone_offset` (`+1:00`). This way, backend gets full information about the timezone from the clients, including the DST adjustment. If the timezone ID is not available then we can fall back to the timezone offset.
> 
[...]

PR #1911, opened 2024-09-10
Add client timezone to transaction history

PooyaRaki:
## Summary

Fix an issue that If a user is in a timezone that follows daylight saving, the transaction history grouping is done without taking it into consideration.

For more information:
https://github.com/safe-global/safe-client-gateway/issues/1024

## Changes

- Add a new query parameter named `transactionId` to transaction history
- Group transactions in the history list based on `transactionId`, If not provided falls back to the old `timezone_offset`

### DIFF
--- a/src/routes/transactions/helpers/timezone.helper.ts
+++ b/src/routes/transactions/helpers/timezone.helper.ts
@@ -4,24 +4,21 @@
  * @param {Date} date The date object to be converted.
  * @param {string} timeZone The target timezone (e.g., "Europe/Berlin").
  *
- * @returns {Date | undefined} A new Date object representing the date converted to the specified timezone. If an error occurs returns undefined
+ * @returns {Date} A new Date object representing the date converted to the specified timezone
+ * @throws {RangeError} Throws if an invalid timezone is sent
  */
-export const convertToTimezone = (
-  date: Date,
-  timeZone: string,
-): Date | undefined => {
-  try {
-    const convertedDate = new Intl.DateTimeFormat('en-US', {
-      timeZone,
-      year: 'numeric',
-      month: '2-digit',
-      day: '2-digit',
-    }).format(date);
+export const convertToTimezone = (date: Date, timeZone: string): Date => {
+  const convertedDate = new Intl.DateTimeFormat('en-US', {
+    timeZone,
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+  }).format(date);
 
-    return new Date(convertedDate);
-  } catch {
-    return undefined;
-  }
+  const [month, day, year] = convertedDate.split('/').map(Number);
+  const zeroBasedMonth = month - 1; // JavaScript months are zero-indexed (0 for January, 11 for December), so we subtract 1
+
+  return new Date(Date.UTC(year, zeroBasedMonth, day));
 };
 
 /**
@@ -39,8 +36,7 @@ export const calculateTimezoneOffset = (
   timestamp: Date,
   timezoneOffset: number,
 ): Date => {
-  const date = structuredClone(timestamp);
-  date.setTime(date.getTime() + timezoneOffset);
+  const date = new Date(timestamp.getTime() + timezoneOffset);
 
   return new Date(
     Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate()),
--- a/src/routes/transactions/helpers/timezone.helper.ts
+++ b/src/routes/transactions/helpers/timezone.helper.ts
@@ -1,0 +1,48 @@
+/**
+ * Converts a given date to a specified timezone.
+ *
+ * @param {Date} date The date object to be converted.
+ * @param {string} timeZone The target timezone (e.g., "Europe/Berlin").
+ *
+ * @returns {Date | undefined} A new Date object representing the date converted to the specified timezone. If an error occurs returns undefined
+ */
+export const convertToTimezone = (
+  date: Date,
+  timeZone: string,
+): Date | undefined => {
+  try {
+    const convertedDate = new Intl.DateTimeFormat('en-US', {
+      timeZone,
+      year: 'numeric',
+      month: '2-digit',
+      day: '2-digit',
+    }).format(date);
+
+    return new Date(convertedDate);
+  } catch {
+    return undefined;
+  }
+};
+
+/**
+ * Calculates the local time in UTC based on a provided timestamp and timezone offset.
+ *
+ * This function adjusts the provided timestamp by a given timezone offset (in milliseconds) and returns a new `Date` object
+ * that represents the equivalent UTC date at midnight (00:00:00) for that adjusted timestamp.
+ *
+ * @param {Date} timestamp - The initial date and time to adjust. This will be cloned and modified.
+ * @param {number} timezoneOffset - The offset to apply to the timestamp, in milliseconds. Positive values will move the time forward, and negative values will move it backward.
+ *
+ * @returns {Date} - A new `Date` object representing the calculated UTC date at midnight (00:00:00) for the adjusted timestamp.
+ */
+export const calculateTimezoneOffset = (
+  timestamp: Date,
+  timezoneOffset: number,
+): Date => {
+  const date = structuredClone(timestamp);
+  date.setTime(date.getTime() + timezoneOffset);
+
+  return new Date(
+    Date.UTC(date.getUTCFullYear(), date.getUTCMonth(), date.getUTCDate()),
+  );
+};
--- a/src/routes/transactions/helpers/timezone.helper.ts
+++ b/src/routes/transactions/helpers/timezone.helper.ts
@@ -8,14 +8,18 @@
  * @throws {RangeError} Throws if an invalid timezone is sent
  */
 export const convertToTimezone = (date: Date, timeZone: string): Date => {
-  const convertedDate = new Intl.DateTimeFormat('en-US', {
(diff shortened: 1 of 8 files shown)


src/routes/transactions/helpers/timezone.helper.ts
```diff
@@ -0,0 +1,48 @@
...
+      timeZone,
+      year: 'numeric',
+      month: '2-digit',
+      day: '2-digit',
+    }).format(date);
+
+    return new Date(convertedDate);
+  } catch {
+    return undefined;
+  }
```
iamacook, line 24: I'm apprehensive to swallow the error without feedback. In situ, we are validating the timezone so this shouldn't throw but in the future it may elsewhere so I'd sooner say we not catch it.

src/routes/transactions/transactions.controller.ts
```diff
@@ -233,6 +234,7 @@ export class TransactionsController {
   @Get('chains/:chainId/safes/:safeAddress/transactions/history')
   @ApiQuery({ name: 'timezone_offset', required: false, type: String })
```
iamacook, line 235: Perhaps we should mark this as deprecated. What do you think?

src/validation/entities/schemas/timezone.schema.ts
```diff
@@ -0,0 +1,40 @@
...
+      return true;
+    },
+    { message: 'Invalid Timezone' },
+  );
+
+const isTimezoneEnabled = (): boolean => {
+  return !Intl || !Intl.DateTimeFormat().resolvedOptions().timeZone
+    ? false
+    : true;
+};
```
iamacook, line 30: We aren't checking for `Intl` support elsewhere so I think we can safely remove the initial check, then simplify the rest, e.g.

```suggestion
const isTimezoneEnabled = (): boolean => {
  return !!Intl.DateTimeFormat().resolvedOptions().timeZone
};
```

However, considering support is in place - do we need this function?

hectorgomezv: I agree, I think `Intl` should be always supported, right? It seems it is supported since Node 0.12 (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl)

PooyaRaki: Personally, I find it confusing to have two `!!`. I believe an explicit true, false is way more clear.
 I would say let's keep `Intl` check as well since otherwise if for whatever reason `Intl` is disabled we will get an error instead of `true` or `false`. What do you think?

src/routes/transactions/transactions.controller.ts
```diff
@@ -233,6 +234,7 @@ export class TransactionsController {
...
   @ApiQuery({ name: 'cursor', required: false, type: String })
+  @ApiQuery({ name: 'timezoneId', required: false, type: String })
```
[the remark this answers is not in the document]
PooyaRaki: Since, we already have `timezone_offset`, my original intention was to include the word id explicitly for added clarity. However, I'll change it to timezone, as that's the more standard term when discussing dates.

src/validation/entities/schemas/timezone.schema.ts
PooyaRaki: @hectorgomezv It depends. You can disabled `Intl` when building node binary. more on that:
https://nodejs.org/api/intl.html#options-for-building-nodejs
But the good thing is that the binaries in docker hub have the default behavior i.e. full `Intl` support.
https://github.com/nodejs/docker-node/blob/0c0069246367ac5ac0fc6bca141fb04faaca2f4b/22/alpine3.19/Dockerfile

Actually, that's a very strict check.

src/routes/transactions/helpers/timezone.helper.ts
iamacook: I agree that it should be handled properly, but I wouldn't swallow it for now. I'd favour catching the error where it's called and log it accordingly, for example.

If we favour doing so later, then adding a comment would otherwise be beneficial.

src/validation/entities/schemas/timezone.schema.ts
iamacook: I'm impartial regarding the ternary.

If we are trying to be defensive regarding `Intl`, shouldn't we also add the check everywhere it's used?

hectorgomezv: Oh, I didn't know that, thanks!  Then I'd leave this up to you. 😄

src/routes/transactions/transactions.controller.ts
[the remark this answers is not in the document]
PooyaRaki: Done ✅

src/validation/entities/schemas/timezone.schema.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const isTimezoneValid = (timezone: string): boolean => {
+  try {
+    Intl.DateTimeFormat(undefined, { timeZone: timezone ?? 'INVALID' });
```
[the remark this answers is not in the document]
PooyaRaki: Done ✅

src/routes/transactions/transactions.controller.ts
PooyaRaki: Done ✅

src/validation/entities/schemas/timezone.schema.ts
```diff
@@ -0,0 +1,40 @@
+import { z } from 'zod';
+
+/**
+ * Validates a timezone
+ *
+ * @param {string | undefined} timezone The timezone string to check for validity
+ *
+ * @returns {boolean} Returns 'true' if the timezone is valid, otherwise 'false'
+ */
```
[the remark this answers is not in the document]
PooyaRaki: The comment was for the same function, I added some more words to make it clear. Let me know if you still find it confusing. Done ✅ 
https://github.com/safe-global/safe-client-gateway/pull/1911/commits/5d1d8e8d153712921e9b32103c399c5f45bbd092

src/routes/transactions/helpers/timezone.helper.ts
PooyaRaki: I completely agree that we should avoid hiding the error. Let's discuss error handling and determine the best approach to address this issue.

PooyaRaki: Since the timezone is being validated in the controller, it should not throw an error. For now, the catch block has been removed to allow the method to pass through the general error handler if an issue occurs. ✅ 
https://github.com/safe-global/safe-client-gateway/pull/1911/commits/320e539ce28f1b73cde8ab12dec591f6b6ba5946

src/routes/transactions/helpers/timezone.helper.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @param {number} timezoneOffset - The offset to apply to the timestamp, in milliseconds. Positive values will move the time forward, and negative values will move it backward.
+ *
+ * @returns {Date} - A new `Date` object representing the calculated UTC date at midnight (00:00:00) for the adjusted timestamp.
+ */
+export const calculateTimezoneOffset = (
+  timestamp: Date,
+  timezoneOffset: number,
+): Date => {
+  const date = structuredClone(timestamp);
+  date.setTime(date.getTime() + timezoneOffset);
```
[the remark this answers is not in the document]
PooyaRaki: Thanks for pointing this out. It was a bit of leftover that I moved to the new file. Done ✅ 
https://github.com/safe-global/safe-client-gateway/pull/1911/commits/320e539ce28f1b73cde8ab12dec591f6b6ba5946

src/routes/transactions/helpers/timezone.helper.ts
```diff
@@ -0,0 +1,48 @@
...
+ * @param {string} timeZone The target timezone (e.g., "Europe/Berlin").
+ *
+ * @returns {Date | undefined} A new Date object representing the date converted to the specified timezone. If an error occurs returns undefined
+ */
+export const convertToTimezone = (
+  date: Date,
+  timeZone: string,
+): Date | undefined => {
+  try {
+    const convertedDate = new Intl.DateTimeFormat('en-US', {
```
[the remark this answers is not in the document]
PooyaRaki: In the new version, we are using the date format to fetch the year, month, and day, so consistency is essential. For example `en-US` returns
`12/14/2024 // 14th Dec 2024`

But `en-GB` returns:

`14/12/2024 // 14th Dec 2024`
✅ 
https://github.com/safe-global/safe-client-gateway/commit/320e539ce28f1b73cde8ab12dec591f6b6ba5946

src/routes/transactions/helpers/timezone.helper.ts
```diff
@@ -0,0 +1,44 @@
...
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+  }).format(date);
+
+  const [month, day, year] = convertedDate.split('/').map(Number);
+  const zeroBasedMonth = month - 1; // JavaScript months are zero-indexed (0 for January, 11 for December), so we subtract 1
+
+  return new Date(Date.UTC(year, zeroBasedMonth, day));
+};
```
iamacook, line 22: This works for our usecase but just for future reference: this converts the date but not the time to the given timezone.

src/routes/transactions/helpers/timezone.helper.ts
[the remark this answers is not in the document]
iamacook: We can alternatively `formatToParts` and get the specific `type` but it works nicely as is 👍

iamacook (approved):

PR #1915, opened 2024-09-11
Bump path-to-regexp

iamacook:
Bumps [path-to-regexp](https://github.com/pillarjs/path-to-regexp):
- from 0.1.7 to 0.1.10
- from 0.2.5 to 1.9.0
- from 3.2.0 to 3.3.0


iamacook:
As we install packages immutably, we've instead decided to wait for the relevant packages to update this on their end.

PR #1920, opened 2024-09-12
Add chain-based event filtering

hectorgomezv:
## Summary
This PR adds a check to the `onEvent` callback of `HooksRepositoryWithNotifications/HooksRepository` to check the `chainId` in the event being processed is related to a known `Chain`. If it isn't, a warning log is emitted, instead of letting the service execute the event handler (and fail).

**Note: Ideally the unsupported events should be grouped to reduce the amount of potential logs. A forthcoming PR will take care of the grouping.**

## Changes
- Adds `isSupportedChain` to `IChainsRepository` (and its memorization wrapper `isSupportedChainMemo`to `HooksRepositoryWithNotifications/HooksRepository`). This function checks the validity of the chain in the event being processed.
- Adds the related tests.

### DIFF
--- a/src/domain/hooks/hooks.repository.ts
+++ b/src/domain/hooks/hooks.repository.ts
@@ -73,12 +73,23 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
   }
 
   async onEvent(event: Event): Promise<unknown> {
-    return Promise.allSettled([
-      this.onEventClearCache(event),
-      this.eventNotificationsHelper.onEventEnqueueNotifications(event),
-    ]).finally(() => {
-      this.onEventLog(event);
-    });
+    const isSupportedChainId = await this.isSupportedChainId(event.chainId);
+
+    if (isSupportedChainId) {
+      return Promise.allSettled([
+        this.onEventClearCache(event),
+        this.eventNotificationsHelper.onEventEnqueueNotifications(event),
+      ]).finally(() => {
+        this.onEventLog(event);
+      });
+    }
+  }
+
+  private async isSupportedChainId(chainId: string): Promise<boolean> {
+    return this.chainsRepository
+      .getChain(chainId)
+      .then(() => true)
+      .catch(() => false);
   }
 
   private async onEventClearCache(event: Event): Promise<void[]> {
--- a/src/domain/hooks/hooks.repository.ts
+++ b/src/domain/hooks/hooks.repository.ts
@@ -58,7 +58,9 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
     private readonly eventNotificationsHelper: EventNotificationsHelper,
   ) {
     this.queueName = this.configurationService.getOrThrow<string>('amqp.queue');
-    this.isSupportedChainMemo = memoize(this.isSupportedChain);
+    this.isSupportedChainMemo = memoize(
+      this.chainsRepository.isSupportedChain.bind(this.chainsRepository),
+    );
   }
 
   onModuleInit(): Promise<void> {
@@ -87,18 +89,11 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
       });
     } else {
       this.loggingService.warn({
-        type: 'unsupported_event_chain',
+        type: 'unsupported_chain_event',
         chainId: event.chainId,
         eventType: event.type,
       });
     }
-  }
-
-  private async isSupportedChain(chainId: string): Promise<boolean> {
-    return this.chainsRepository
-      .getChain(chainId)
-      .then(() => true)
-      .catch(() => false);
   }
 
   private async onEventClearCache(event: Event): Promise<void[]> {
@@ -485,7 +480,9 @@ export class HooksRepository implements IHooksRepository {
     private readonly configurationService: IConfigurationService,
   ) {
     this.queueName = this.configurationService.getOrThrow<string>('amqp.queue');
-    this.isSupportedChainMemo = memoize(this.isSupportedChain);
+    this.isSupportedChainMemo = memoize(
+      this.chainsRepository.isSupportedChain.bind(this.chainsRepository),
+    );
   }
 
   onModuleInit(): Promise<void> {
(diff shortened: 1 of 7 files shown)


src/domain/hooks/hooks.repository.ts
```diff
@@ -73,12 +77,28 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
...
+    if (isSupportedChainId) {
+      return Promise.allSettled([
+        this.onEventClearCache(event),
+        this.eventNotificationsHelper.onEventEnqueueNotifications(event),
+      ]).finally(() => {
+        this.onEventLog(event);
+      });
+    } else {
+      this.loggingService.warn({
+        type: 'unsupported_event_chain',
```
iamacook, line 90: If this follows a naming convention we can leave it as is, but `unsupported_chain_event` reads better (also would need changing in the other service).

src/domain/hooks/hooks.repository.ts
```diff
@@ -73,12 +77,28 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
...
+        chainId: event.chainId,
+        eventType: event.type,
+      });
+    }
+  }
+
+  private async isSupportedChain(chainId: string): Promise<boolean> {
```
iamacook, line 97: Nit: I think this should "live" in the chains repository and we then expose a `clearSupportedChain` method for consistency. What do you think?

src/routes/hooks/hooks-cache.controller.spec.ts
```diff
@@ -977,17 +996,28 @@ describe('Post Hook Events for Cache (Unit)', () => {
...
     const data = {
-      chainId: chainId,
+      chainId: chain.chainId,
       ...payload,
     };
+    networkService.get.mockImplementation(({ url }) => {
+      switch (url) {
+        case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
+          return Promise.resolve({
+            data: chainBuilder().with('chainId', chain.chainId).build(),
```
iamacook, line 1014:
```suggestion
            data: chain,
```

src/routes/hooks/hooks-cache.controller.spec.ts
```diff
@@ -1014,17 +1044,28 @@ describe('Post Hook Events for Cache (Unit)', () => {
...
       const data = {
-        chainId: chainId,
+        chainId: chain.chainId,
         ...payload,
       };
+      networkService.get.mockImplementation(({ url }) => {
+        switch (url) {
+          case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
+            return Promise.resolve({
+              data: chainBuilder().with('chainId', chain.chainId).build(),
```
iamacook, line 1062:
```suggestion
              data: chain,
```

src/routes/hooks/hooks-cache.controller.spec.ts
```diff
@@ -1052,18 +1093,30 @@ describe('Post Hook Events for Cache (Unit)', () => {
...
-        chainId: chainId,
+        chainId: chain.chainId,
         ...payload,
       };
 
+      networkService.get.mockImplementation(({ url }) => {
+        switch (url) {
+          case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
+            return Promise.resolve({
+              data: chainBuilder().with('chainId', chain.chainId).build(),
```
iamacook, line 1112:
```suggestion
              data: chain,
```

iamacook (commented):
Out of scope but I think we should consider a way to reduce the duplication across the (non-)notification hook repositories. What do you think?

src/domain/hooks/hooks.repository.ts
```diff
@@ -328,6 +348,10 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
         );
         break;
       case ConfigEventType.CHAIN_UPDATE:
+        // As the chains have been updated, we need to clear the memoized function cache.
+        if (this.isSupportedChainMemo.cache.clear) {
```
PooyaRaki, line 352: Suggestion: What do you think about creating a private method to handle this condition?
The current method is nearly 300 lines long, so splitting it could improve readability. It’s fine for now, but it seems like the entire method could benefit from being refactored.

hectorgomezv: Absolutely! Good point, I think we could move all the code within each `case` to private methods to improve the overall readability. Are you OK if I do that on a separate PR?

src/domain/hooks/hooks.repository.ts
hectorgomezv: Totally agree! Changed in [789cee5](https://github.com/safe-global/safe-client-gateway/pull/1920/commits/789cee533294c8d1e6717e74004882002c737b01)

src/domain/hooks/hooks.repository.ts
hectorgomezv: Moved in [789cee5](https://github.com/safe-global/safe-client-gateway/pull/1920/commits/789cee533294c8d1e6717e74004882002c737b01)

src/domain/hooks/hooks.repository.ts
PooyaRaki: @hectorgomezv Absolutely, that's a good idea.

src/domain/hooks/hooks.repository.ts
hectorgomezv: I didn't move the memoized function to the `ChainsRepository` because I think it could add more complexity (other classes might not need to call a memoized function I think). So I've left the clearing of the memoized function cache responsibility in the `HooksRepository` for now. If memoizing makes sense in other high-load places we could think on memoizing inside the `ChainsRepository` (or even on other repositories), but I tend to think it is out of the scope of the PR, wdyt?

iamacook (approved):

src/domain/hooks/hooks.repository.ts
hectorgomezv: I've created an issue to keep track of it 👍 

https://github.com/safe-global/safe-client-gateway/issues/1926

PooyaRaki (approved):

PR #1921, opened 2024-09-12
Aggregate event logs for unsupported chains

hectorgomezv:


src/domain/hooks/hooks.repository.ts
```diff
@@ -23,11 +28,19 @@ import { IStakingRepository } from '@/domain/staking/staking.repository.interfac
 import { memoize, MemoizedFunction } from 'lodash';
 
 @Injectable()
-export class HooksRepositoryWithNotifications implements IHooksRepository {
+export class HooksRepositoryWithNotifications
+  implements IHooksRepository, OnModuleInit, OnModuleDestroy
+{
   private static readonly HOOK_TYPE = 'hook';
+  private static readonly UNSUPPORTED_CHAIN_EVENT = 'unsupported_chain_event';
+  public static readonly UNSUPPORTED_EVENTS_LOG_INTERVAL = 60 * 1000; // 1 minute
```
hectorgomezv, line 36: We could add this to the configuration if we plan to change it dynamically. I opted for keeping it here for now, but I'm open to externalise the configuration.

iamacook (dismissed):

src/domain/hooks/hooks.repository.ts
```diff
@@ -23,11 +28,19 @@ import { IStakingRepository } from '@/domain/staking/staking.repository.interfac
...
   private readonly queueName: string;
-  private isSupportedChainMemo: ((chainId: string) => Promise<boolean>) &
+  private readonly isSupportedChainMemo: ((
+    chainId: string,
+  ) => Promise<boolean>) &
     MemoizedFunction;
+  private readonly unsupportedEventsStore: Map<string, number> = new Map();
```
iamacook, line 42: As we have two instances and this is stored in memory, won't this mean the log is emitted twice each time?

src/domain/hooks/hooks.repository.spec.ts
```diff
@@ -86,6 +86,8 @@ describe('HooksRepository (Unit)', () => {
   let hooksRepository: HooksRepository;
 
   beforeEach(() => {
+    jest.useFakeTimers();
```
PooyaRaki, line 89: Does it make sense to move it to `beforeAll` ?

src/domain/hooks/hooks.repository.spec.ts
```diff
@@ -100,7 +102,10 @@ describe('HooksRepository (Unit)', () => {
       mockQueuesRepository,
       mockConfigurationService,
     );
-    jest.clearAllMocks();
+  });
+
+  afterEach(() => {
```
PooyaRaki, line 107: It seems like we can move `useRealTimers` to `afterAll`. Is there a reason we are doing it after each test?

src/domain/hooks/hooks.repository.spec.ts
```diff
@@ -195,12 +200,42 @@ describe('HooksRepository (Unit)', () => {
...
+      count: 4,
+    });
+  });
 });
 
 describe('HooksRepositoryWithNotifications (Unit)', () => {
   let hooksRepositoryWithNotifications: HooksRepositoryWithNotifications;
 
   beforeEach(() => {
+    jest.useFakeTimers();
```
PooyaRaki, line 237: Should we enable `fakeTimers` once at the beginning of the test suite using `beforeAll` and disable it again at the end by using `afterAll`?

src/domain/hooks/hooks.repository.ts
iamacook: I moved this to Redis in aea854488ca98389b1d3878f84dad1eadcadb337.

src/domain/hooks/hooks.repository.spec.ts
iamacook: Moved in aea854488ca98389b1d3878f84dad1eadcadb337.

src/domain/hooks/hooks.repository.ts
```diff
@@ -449,9 +507,12 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
 @Injectable()
 export class HooksRepository implements IHooksRepository {
   private static readonly HOOK_TYPE = 'hook';
+  private static readonly UNSUPPORTED_CHAIN_EVENT = 'unsupported_chain_event';
```
iamacook, line 510: The duplication from here on, as well as in the tests will be solved by https://github.com/safe-global/safe-client-gateway/pull/1928.

src/domain/hooks/hooks.repository.ts
hectorgomezv: Thank you for looking into this Aaron! The memory storage was created to alleviate cache usage when receiving a lot of unsupported events in a short period. The log would indeed be emitted N times on each cycle (N being the number of running instances). Still, each instance would print the number of events received by the instance, and monitoring tools can aggregate the numbers if needed. 

Sorry, this approach was highly opinionated and probably should have been discussed beforehand, I'm fine with using Redis for storing a unified counter, but I also think we should use an atomic counter (Redis `INCR`) instead of a plain hash (as implemented on https://github.com/safe-global/safe-client-gateway/commit/aea854488ca98389b1d3878f84dad1eadcadb337) for both consistency (dirty writes) and performance. Wdyt @iamacook @PooyaRaki ?

(Btw, if we think this should be changed, I'd be super happy to tackle it on a separate PR so we can merge this one and unblock the feature 🙂 )

src/domain/hooks/hooks.repository.spec.ts
hectorgomezv: Totally agree in this case! We tend to use `after/beforeEach` to assure test independence and avoid setup interferences when it comes to timers. This is a rule of thumb and I applied that to this test file but it's indeed unnecessary.

src/domain/hooks/helpers/event-cache.helper.ts
```diff
@@ -451,6 +489,28 @@ export class EventCacheHelper {
...
+   */
+  private async _logUnsupportedEvents(): Promise<void> {
+    const chains = await this.chainsRepository.getChains();
+    await Promise.all(
+      chains.results.map(async (chain) => {
+        const cacheKey = CacheRouter.getUnsupportedChainEventCacheKey(
+          chain.chainId,
+        );
+        const count = await this.cacheService.getCounter(cacheKey);
+        if (count) {
```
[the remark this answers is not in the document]
hectorgomezv: Yes, it should always be different from `0` (it should be a non-zero positive integer or `null`). But in case it's `0` (because the underlying implementation changes) then the `if` part shouldn't be executed, right? I mean, if it's `0` then we don't want to emit the log line. Wdyt? In that case the implementation would be accurate.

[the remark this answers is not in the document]
iamacook: Very true. I got confused thinking we were incrementing the counter. Let's leave it as is.

src/domain/hooks/hooks.repository.ts
```diff
@@ -10,7 +10,9 @@ import { EventNotificationsHelper } from '@/domain/hooks/helpers/event-notificat
 import { EventCacheHelper } from '@/domain/hooks/helpers/event-cache.helper';
 
 @Injectable()
-export class HooksRepositoryWithNotifications implements IHooksRepository {
+export class HooksRepositoryWithNotifications
+  implements IHooksRepository, OnModuleInit
```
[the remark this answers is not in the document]
hectorgomezv: I've just found out that `onModuleInit` lifecycle hook is called by NestJS if a class defines a method called `onModuleInit`, regardless of whether the class explicitly implements the `OnModuleInit` interface.

So even if the interface is not mandatory for the hook to work, I think it is a good practice to use it for clarity and to ensure your class conforms to the expected structure, so I've explicitly set both `HooksRepositoryWithNotifications` and `HooksRepository` as `OnModuleInit` implementations in [7d695ee](https://github.com/safe-global/safe-client-gateway/pull/1921/commits/7d695eea60f46d7fa8f684a3251144450cc56371)

iamacook (approved):

PR #1923, opened 2024-09-13
Taking withdrawal time into account when calculating exit status

iamacook:
## Summary

For exit requests, we include a status relative to the estimated exit time. Currently, we state that validators can be withdrawn after only this period. In order to withdraw, however, we need to take into account both the estimated exit and withdrawal times.

This adjusts the relative `status`, taking the extra withdraw period into account before returning the `READY_TO_WITHDRAW` status for validator exit requests.

## Changes

- Add the `estimated_withdrawal_time_seconds` to the `estimatedCompletionTime`
- Update tests accordingly

### DIFF
--- a/src/routes/transactions/mappers/common/native-staking.mapper.ts
+++ b/src/routes/transactions/mappers/common/native-staking.mapper.ts
@@ -293,7 +293,8 @@ export class NativeStakingMapper {
 
     const estimatedCompletionTime =
       transaction.executionDate.getTime() +
-      networkStats.estimated_exit_time_seconds * 1000;
+      networkStats.estimated_exit_time_seconds * 1000 +
+      networkStats.estimated_withdrawal_time_seconds * 1000;
 
     return Date.now() <= estimatedCompletionTime
       ? StakingValidatorsExitStatus.RequestPending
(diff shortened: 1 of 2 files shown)


src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -293,7 +293,8 @@ export class NativeStakingMapper {
 
     const estimatedCompletionTime =
       transaction.executionDate.getTime() +
-      networkStats.estimated_exit_time_seconds * 1000;
+      networkStats.estimated_exit_time_seconds * 1000 +
+      networkStats.estimated_withdrawal_time_seconds * 1000;
```
hectorgomezv, line 297: Nit: I'd also update the function doc from:
```
...
- If the execution date is available, the status is {@link StakingValidatorsExitStatus.RequestPending} if the current date is before the estimated exit time, otherwise the status is {@link StakingValidatorsExitStatus.ReadyToWithdraw}.
   ...
```

To: 

```
...
- If the execution date is available, the status is {@link StakingValidatorsExitStatus.RequestPending} if the current date is before the estimated exit time + the estimated withdrawal time, otherwise the status is {@link StakingValidatorsExitStatus.ReadyToWithdraw}.
   ...
```

(Or something similar)

hectorgomezv (approved):

src/routes/transactions/mappers/common/native-staking.mapper.ts
iamacook: Updated in 321c281cb0332198bdc8ca1c4dde81ee90c7be71.

hectorgomezv (approved):

PR #1925, opened 2024-09-13
Improve caching of a Safe's `Stakes`

### LINKED ISSUE #1914: [Native Staking] Refresh Native Staking rewards cache on transaction execution
### Description

**As a** Safe{Wallet} user
**I want to** have an up-to-date view of my claimable rewards when staking ETH.

### Requirements

- Add an event hook that invalidates all the `Stake` data for the Safe affected by the event.

iamacook:
## Summary

Resolves #1914.

Currently we indefinitely (with a set TTL) cache the `Stakes` response according to `_publicKeys`. It is never cleared when a Safe executes a transaction, meaning that the status will not update if withdrawn.

This modifies the current caching to ensure entries are chain/Safe specific and clear when the Safe in question executes a transaciton.

## Changes

- Modify `staking_stakes` cache key to include the `chainId` and `safeAddress`, moving the `_publicKeys` hash to the cache field.
- Add method to `IStakingApi` to clear the above cache, propagated to `IStakingRepository`.
- Clear the cache on `EXECUTED_MULTISIG_TRANSACTION` events.

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -605,9 +605,9 @@ export class CacheRouter {
    * cache field short and deterministic. Redis and other cache systems
    * may experience performance degradation with long fields.
    *
-   * @param chainId - Chain ID
-   * @param safeAddress - Safe address
-   * @param validatorsPublicKeys - Array of validators public keys
+   * @param {string} args.chainId - Chain ID
+   * @param {string} args.safeAddress - Safe address
+   * @param {string} args.validatorsPublicKeys - Array of validators public keys
    * @returns {@link CacheDir} - Cache directory
    */
   static getStakingStakesCacheDir(args: {
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -591,21 +591,35 @@ export class CacheRouter {
     );
   }
 
+  static getStakingStakesCacheKey(args: {
+    chainId: string;
+    safeAddress: `0x${string}`;
+  }): string {
+    return `${args.chainId}_${CacheRouter.STAKING_STAKES_KEY}_${args.safeAddress}`;
+  }
+
   /**
    * Calculate cache directory for staking stakes.
    *
    * Note: This function hashes the validators' public keys to keep the
-   * cache key short and deterministic. Redis and other cache systems
-   * may experience performance degradation with long keys.
+   * cache field short and deterministic. Redis and other cache systems
+   * may experience performance degradation with long fields.
    *
+   * @param chainId - Chain ID
+   * @param safeAddress - Safe address
    * @param validatorsPublicKeys - Array of validators public keys
    * @returns {@link CacheDir} - Cache directory
    */
-  static getStakingStakesCacheDir(
-    validatorsPublicKeys: Array<`0x${string}`>,
-  ): CacheDir {
+  static getStakingStakesCacheDir(args: {
+    chainId: string;
+    safeAddress: `0x${string}`;
+    validatorsPublicKeys: Array<`0x${string}`>;
+  }): CacheDir {
     const hash = crypto.createHash('sha256');
-    hash.update(validatorsPublicKeys.join('_'));
-    return new CacheDir(`${this.STAKING_STAKES_KEY}_${hash.digest('hex')}`, '');
+    hash.update(args.validatorsPublicKeys.join('_'));
+    return new CacheDir(
+      CacheRouter.getStakingStakesCacheKey(args),
+      hash.digest('hex'),
+    );
   }
 }
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -591,6 +591,13 @@ export class CacheRouter {
     );
   }
 
+  /**
+   * Calculated the chain/Safe-specific cache key of {@link Stake}.
+   *
+   * @param {string} args.chainId - Chain ID
+   * @param {string} args.safeAddress - Safe address
+   * @returns {string} - Cache key
+   */
   static getStakingStakesCacheKey(args: {
     chainId: string;
     safeAddress: `0x${string}`;
--- a/src/datasources/staking-api/kiln-api.service.ts
+++ b/src/datasources/staking-api/kiln-api.service.ts
@@ -1,6 +1,7 @@
 import { IConfigurationService } from '@/config/configuration.service.interface';
 import { CacheFirstDataSource } from '@/datasources/cache/cache.first.data.source';
 import { CacheRouter } from '@/datasources/cache/cache.router';
+import { ICacheService } from '@/datasources/cache/cache.service.interface';
 import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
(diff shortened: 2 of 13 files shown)


src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
```diff
@@ -406,6 +406,7 @@ describe('NativeStakingMapper', () => {
 
       const actual = await target.mapValidatorsExitInfo({
         chainId: chain.chainId,
+        safeAddress: transaction.safe,
```
iamacook, line 409: On face value it looks as though we could access the `safe` property in this method but `transaction` can be `null` (when proposing a transaction). We therefore need explicitly pass the `safeAddress`.

src/domain/hooks/hooks.repository.ts
```diff
@@ -167,6 +167,10 @@ export class HooksRepositoryWithNotifications implements IHooksRepository {
             chainId: event.chainId,
             address: event.address,
           }),
+          this.stakingRepository.clearStakes({
```
hectorgomezv, line 170: Nit: should we also clear on `MODULE_TRANSACTION`?

hectorgomezv (approved):

src/domain/hooks/hooks.repository.ts
iamacook: Good point - yes! A module could execute a staking transaction. Changed in 21667dc4b3e336a0428a1de3527941ffe555a92b.

src/datasources/staking-api/kiln-api.service.ts
```diff
@@ -200,6 +218,14 @@ export class KilnApi implements IStakingApi {
     }
   }
 
+  async clearStakes(safeAddress: `0x${string}`): Promise<void> {
```
PooyaRaki, line 221: What do you think about `clearStakesCache`.
nit: It would be nice to have a JSdoc comment here as well.

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -174,13 +177,15 @@ export class NativeStakingMapper {
    * native staking `batchWithdrawCLFee` transaction.
    *
    * @param args.chainId - the chain ID of the native staking deployment
+   * @param args.safeAddress - the Safe staking
```
PooyaRaki, line 180: nit: We could include the argument type as well.
```suggestion
   * @param {string} args.safeAddress - the Safe staking
```

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -327,20 +333,23 @@ export class NativeStakingMapper {
...
    * @returns the rewards to withdraw from the native staking deployment
    */
-  private async getRewardsFromDataDecoded(
-    data: DataDecoded,
-    chain: Chain,
-  ): Promise<number> {
-    const publicKeys = this.getPublicKeysFromDataDecoded(data);
+  private async getRewardsFromDataDecoded(args: {
+    dataDecoded: DataDecoded;
+    chainId: string;
```
PooyaRaki, line 343: Question: Why is `chain` changed to `chainId` here?

PooyaRaki (changes requested):

src/datasources/cache/cache.router.ts
```diff
@@ -591,21 +591,35 @@ export class CacheRouter {
     );
   }
 
+  static getStakingStakesCacheKey(args: {
```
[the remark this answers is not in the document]
iamacook: Added in 1a78503e8d98bc3f704be5838685dfbbd1b369ad.

As per our discussion, we should slowly aim to include JSDoc for every function from henceforth.

src/datasources/cache/cache.router.ts
```diff
@@ -591,21 +591,35 @@ export class CacheRouter {
...
   /**
    * Calculate cache directory for staking stakes.
    *
    * Note: This function hashes the validators' public keys to keep the
-   * cache key short and deterministic. Redis and other cache systems
-   * may experience performance degradation with long keys.
+   * cache field short and deterministic. Redis and other cache systems
+   * may experience performance degradation with long fields.
    *
+   * @param chainId - Chain ID
```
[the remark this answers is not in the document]
iamacook: Changed in a2ddb8d0a00494f553f85b88871705788d468006.

src/datasources/cache/cache.router.ts
```diff
@@ -591,21 +591,35 @@ export class CacheRouter {
...
+   * @param safeAddress - Safe address
    * @param validatorsPublicKeys - Array of validators public keys
    * @returns {@link CacheDir} - Cache directory
    */
-  static getStakingStakesCacheDir(
-    validatorsPublicKeys: Array<`0x${string}`>,
-  ): CacheDir {
+  static getStakingStakesCacheDir(args: {
```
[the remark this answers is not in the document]
iamacook: An interface seems somewhat excessive as it would not be reused. Do you think it's a pattern we should adopt? We don't tend to user interfaces outside of those for classes.

src/datasources/staking-api/kiln-api.service.ts
```diff
@@ -170,13 +173,28 @@ export class KilnApi implements IStakingApi {
...
 
-  async getStakes(
-    validatorsPublicKeys: Array<`0x${string}`>,
-  ): Promise<Stake[]> {
+  /**
+   * Gets the {@link Stake} for the given {@param args.validatorsPublicKeys}.
+   *
+   * The {@param args.safeAddress} is only used for caching purposes.
+   *
+   * @param args.safeAddress - Safe address
```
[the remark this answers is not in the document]
iamacook: Added in a2ddb8d0a00494f553f85b88871705788d468006.

src/datasources/staking-api/kiln-api.service.ts
iamacook: Changed in a2ddb8d0a00494f553f85b88871705788d468006/1a78503e8d98bc3f704be5838685dfbbd1b369ad. If we think it best, we should also update other cache-clearing methods as we don't have `-Cache` elsewhere.

It shouldn't be done here to keep the diff minimal, but I'd subsequently like to move the comments to the class interface as well.

src/routes/transactions/mappers/common/native-staking.mapper.ts
iamacook: Added in a2ddb8d0a00494f553f85b88871705788d468006.

src/routes/transactions/mappers/common/native-staking.mapper.ts
iamacook: Only `chain.chainId` is used. Passing the entire `Chain` seemed excessive. Should I revert it?

PooyaRaki: All good.
In terms of memory and performance, there shouldn't be any significant differences. JavaScript is pass-by-value, but when dealing with objects, the variable holds a reference to the object rather than the object itself.

PooyaRaki (approved):

PR #1928, opened 2024-09-16
Refactor cache clearing logic into `EventCacheHelper`

### LINKED ISSUE #1926: [Tech debt] Refactor HooksRepository to improve readability
### Description

(Context: https://github.com/safe-global/safe-client-gateway/pull/1920#discussion_r1758546088)

**As a** Safe{Wallet} engineer
**I want to** improve the overall readability of `HooksRepository`
**So** a refactoring of the actions associated with each `case` on `onEventClearCache` could be extracted to a private function. That would also favor reusability, as currently both `HooksRepository` and `HooksRepositoryWithNotifications` share the same logic there.

### Requirements

- Refactor `HooksRepository/HooksRepositoryWithNotifications` to reuse logic via private functions.

iamacook:
## Summary

Resolves #1926

Our `HooksRepository['onEventClearCache']` covers cache clearing based on every event from the Transaction Service and Config. Service and is therefore excessively long. Considering we also have a similar class for notification-enabled hooks, this logic was duplicated.

This adds a new `EventCacheHelper`, breaking down respective event logic into a separate method. It is then used in both the (non-)event based hook repositories.

## Changes

- Extract cache-related logic to `EventCacheHelper`
- Use `EventCacheHelper` in `HooksRepository` and `HooksRepositoryWithNotifications`
- Update tests accordingly


src/domain/hooks/helpers/event-cache.helper.ts
```diff
@@ -0,0 +1,553 @@
...
+  private _logEvent(event: Event): void {
+    this.loggingService.info({
+      type: EventCacheHelper.HOOK_TYPE,
+      eventType: event.type,
+      chainId: event.chainId,
+    });
+  }
+}
+
+@Module({
```
PooyaRaki, line 538: What do you think about having a separate file called event-cache.module for this module? As it makes the module class more accessible and improves the readability of the code

src/domain/hooks/helpers/event-cache.helper.ts
```diff
@@ -0,0 +1,553 @@
...
+    @Inject(LoggingService)
+    private readonly loggingService: ILoggingService,
+  ) {
+    this.isSupportedChainMemo = memoize(
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-call
+      this.chainsRepository.isSupportedChain.bind(this.chainsRepository),
+    );
+  }
+
+  public async onEventClearCache(event: Event): Promise<void[]> {
```
PooyaRaki, line 78: I'm having a hard time understanding this method. What do you think about mapping the event type to the related method instead on having different `case`s? As an example:
```suggestion
  export const EventTypeHandler = {
    TransactionEventType.PENDING_MULTISIG_TRANSACTION: onTransactionEventPendingMultisigTransaction,
    // ...
  }
  public async onEventClearCache(event: Event): Promise<void[]> {
    const eventHandler = this.fetchTransactionEventHandler(event.type);
    promises.push(
        ...eventHandler(event);
    )
  }
```

src/domain/hooks/helpers/event-cache.helper.ts
```diff
@@ -0,0 +1,553 @@
...
+      case ConfigEventType.CHAIN_UPDATE:
+      case ConfigEventType.SAFE_APPS_UPDATE:
+        this._logEvent(event);
+        break;
+      case TransactionEventType.SAFE_CREATED:
+        break;
+    }
+  }
+
+  // Transaction Service events
```
PooyaRaki, line 162: TODO: Split the service into multiple classes, each handling a specific type of event, to adhere to the Single Responsibility Principle.

src/domain/hooks/helpers/event-cache.helper.ts
iamacook: There's a pattern of including the `Module` in less "significant" [helpers](https://github.com/safe-global/safe-client-gateway/blob/refactor-hooks-repository/src/domain/swaps/contracts/decoders/gp-v2-decoder.helper.ts#L783), [repositories](https://github.com/safe-global/safe-client-gateway/blob/refactor-hooks-repository/src/routes/transactions/helpers/swap-order.helper.ts#L159), etc. However, I created a separate module in 9fb9f4ff1ae9ed9cd25837cc165356946b72e5b1 as I agree (`event-cache.helper.module.ts`).

We should reconsider other areas where we follow the aforementioned pattern as well.

src/domain/hooks/helpers/event-cache.helper.ts
iamacook: Added in 9fb9f4ff1ae9ed9cd25837cc165356946b72e5b1.

src/domain/hooks/helpers/event-cache.helper.ts
iamacook: Whilst I understand the approach, experimenting with a secondary class resulted in a lot of boilerplate and it's the [`Event` entity](https://github.com/safe-global/safe-client-gateway/blob/refactor-hooks-repository/src/routes/hooks/entities/schemas/event.schema.ts#L17) is being handled (although it consists of both Config Service and Transaction Service `type`s).

What are your thoughts? Do you think it is essential we separate the class? If so, I'll add a TODO comment for a followup to keep the diff. minimal.

PooyaRaki (approved):

PR #1930, opened 2024-09-16
Remove trailing slashes from Transaction Service URLs

### LINKED ISSUE #1929: Ensure Transaction Service URLs have no trailing slash
### Description

We expect `Chain['transactionService' | 'vpcTransactionService'` to have no trailing slash, though we never enforce this. This means that any request to the Transaction Service (if the relative configuration _has_ a trailing slash) will fail.

### Reproduction steps

1. Add a trailing slash to `transactionService`
2. Fetch anything that uses the Transaction Service, e.g. Safe info 

### Expected behavior

No error occurs.

### Additional information

This was initially discovered on the `/v1/safes` endpoint.

iamacook:
## Summary

Resolves #1929

It is possible that `Chain['transactionService' | 'vpcTransactionService']` have trailing slashes, which breaks requests to the Transaction Service as we expect none to be present.

This refines the incoming values, removing the trailing slash if present.

## Changes

- Remove trailing slashes from `Chain['transactionService' | 'vpcTransactionService']` if present

### DIFF
--- a/src/domain/chains/entities/schemas/__tests__/chain.schema.spec.ts
+++ b/src/domain/chains/entities/schemas/__tests__/chain.schema.spec.ts
@@ -550,6 +550,30 @@ describe('Chain schemas', () => {
       },
     );
 
+    it.each(['transactionService' as const, 'vpcTransactionService' as const])(
+      'accept non-trailing slash %s as is',
+      (field) => {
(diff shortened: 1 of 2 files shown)


PooyaRaki (approved):
LGTM

PR #1936, opened 2024-09-17
Use instance chain ID within `IStakingApi`

iamacook:
## Summary

Instances of `IStakingApi` have are chain-specific and are passed a `chainId`. This value is used in cache keys.

This modifies `KilnApi['getDefiVaultStats']` to use the instance `chainId` instead of using that passed as an argument.

## Changes

- Improve typing of chain identifiers for DeFi vault stat requests
- Fetch stats with `chainId` of instance
- Update tests accordingly

### DIFF
--- a/src/datasources/staking-api/kiln-api.service.spec.ts
+++ b/src/datasources/staking-api/kiln-api.service.spec.ts
@@ -340,14 +340,21 @@ describe('KilnApi', () => {
 
   describe('getDefiVaultStats', () => {
     it('should return the defi vault stats', async () => {
-      const [chain, chainId] = faker.helpers.arrayElement(
-        Object.entries(KilnApi.DefiVaultStatsChains),
+      const chainIds = {
+        eth: 1,
+        arb: 42161,
+        bsc: 56,
+        matic: 137,
+        op: 10,
+      };
+      const [chain, chain_id] = faker.helpers.arrayElement(
+        Object.entries(chainIds) as Array<[keyof typeof chainIds, number]>,
       );
       // Ensure target is created with supported chain
-      createTarget(chainId);
+      createTarget(chain_id.toString());
       const defiVaultStats = defiVaultStatsBuilder()
-        .with('chain', chain as keyof typeof KilnApi.DefiVaultStatsChains)
-        .with('chain_id', +chainId)
+        .with('chain', chain)
+        .with('chain_id', chain_id)
         .build();
       dataSource.get.mockResolvedValue({
         status: 200,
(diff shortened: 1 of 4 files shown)


src/datasources/staking-api/kiln-api.service.spec.ts
```diff
@@ -44,10 +44,8 @@ describe('KilnApi', () => {
   let stakingExpirationTimeInSeconds: number;
   let notFoundExpireTimeSeconds: number;
 
-  beforeEach(() => {
-    jest.resetAllMocks();
-
-    chainId = faker.string.numeric();
+  function createTarget(_chainId = faker.string.numeric()): void {
```
PooyaRaki, line 47: Praise: I like the idea of creating new functions/methods to split the code and be DRY.

src/datasources/staking-api/kiln-api.service.ts
```diff
@@ -239,32 +254,18 @@ export class KilnApi implements IStakingApi {
...
 
-    if (isDeFiSupportedChain(args.chainId)) {
-      const chainIdentifier = chainIdentifiers[args.chainId];
-      return `${chainIdentifier}_${args.vault}`;
+    if (chain) {
+      const chainIdentifier = chain[0];
+      return `${chainIdentifier}_${vault}`;
     } else {
-      throw new Error(`${args.chainId} is not supported for DeFi by Kiln`);
+      throw new Error(`${this.chainId} is not supported for DeFi by Kiln`);
```
PooyaRaki, line 268: Question: Do we show the Error message in reponse? If so, Do we need to include the details? For sure, We need the details for log purposes.

iamacook: Yes, [we create a "response error" from the error thrown](https://github.com/safe-global/safe-client-gateway/blob/kiln-api-instance-chainid/src/datasources/staking-api/kiln-api.service.ts#L186).

PooyaRaki: It seems like we are not showing the error to the user:
[https://github.com/safe-global/safe-client-gateway/blob/d6cb3068ac228fe32962a9c41c6440ab9bc189a1/src/datasources/errors/http-error-factory.ts#L17](http-error-factory)

PooyaRaki (approved):

src/datasources/staking-api/kiln-api.service.ts
iamacook: That error is thrown and caught by the [`GlobalErrorFilter`](https://github.com/safe-global/safe-client-gateway/blob/kiln-api-instance-chainid/src/routes/common/filters/global-error.filter.ts#L0-L1) where it is logged.

PR #1937, opened 2024-09-19
Add missing `ApiQuery` decorators for Swagger

### LINKED ISSUE #1747: Optional query parameters marked as required on Swagger
### Description

We have some query parameters with default values, e.g. `exclude_spam`. These are theoretically optional. On Swagger, however, they are marked as required and also do not show the default value:

image

### Reproduction steps

As an example, observe the balances endpoint marking `trusted` and `exclude_spam` as required fields on Swagger. This also seems to be the case everywhere we use a `DefaultValuePipe`.

### Expected behavior

In the same example, `trusted` and `exclude_param` should not be required but have a note with the default value, e.g.

image

### Additional information

_No response_

iamacook:
## Summary

Resolves #1747

Some optional queries are currently marked a required on Swagger. This adds the missing `ApiQuery` decorators for said queries.

## Changes

- Add missing `ApiQuery` decorators where `Query`s are present.

### DIFF
--- a/src/routes/safes/safes.controller.ts
+++ b/src/routes/safes/safes.controller.ts
@@ -42,7 +42,11 @@ export class SafesController {
     return this.service.getNonces({ chainId, safeAddress });
   }
 
-  @ApiQuery({ name: 'wallet_address', required: false })
+  @ApiQuery({ name: 'wallet_address', required: false, type: String })
+  @ApiQuery({ name: 'currency', required: true, type: String })
+  @ApiQuery({ name: 'safes', required: true, type: String })
+  @ApiQuery({ name: 'trusted', required: false, type: Boolean })
+  @ApiQuery({ name: 'exclude_spam', required: false, type: Boolean })
   @Get('safes')
   async getSafeOverview(
     @Query('currency') currency: string,
--- a/src/routes/balances/balances.controller.ts
+++ b/src/routes/balances/balances.controller.ts
@@ -6,7 +6,7 @@ import {
   ParseBoolPipe,
   Query,
 } from '@nestjs/common';
-import { ApiOkResponse, ApiTags } from '@nestjs/swagger';
+import { ApiOkResponse, ApiQuery, ApiTags } from '@nestjs/swagger';
 import { BalancesService } from '@/routes/balances/balances.service';
 import { Balances } from '@/routes/balances/entities/balances.entity';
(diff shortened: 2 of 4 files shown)


PooyaRaki:
Question: How confident are we that these were the only methods with issues, and that other methods are unaffected?

src/routes/safes/safes.controller.ts
```diff
@@ -42,7 +42,11 @@ export class SafesController {
...
 
-  @ApiQuery({ name: 'wallet_address', required: false })
+  @ApiQuery({ name: 'wallet_address', required: false, type: String })
+  @ApiQuery({ name: 'currency', required: true, type: String })
+  @ApiQuery({ name: 'safes', required: true, type: String })
+  @ApiQuery({ name: 'trusted', required: false, type: Boolean })
+  @ApiQuery({ name: 'exclude_spam', required: false, type: Boolean })
   @Get('safes')
   async getSafeOverview(
     @Query('currency') currency: string,
```
PooyaRaki, line 52: TODO: When parameters are optional, mark them as optional in the code by appending a question mark (?) to the end of the parameter name. This way, Swagger may automatically recognize the parameter as optional. e.g.

```suggestion
    @Query('currency') currency?: string,
```
Implementing this requires additional effort and is outside the scope of the current PR.

PooyaRaki (approved):

src/routes/safes/safes.controller.ts
iamacook: Which parameters are not marked as optional? In this case, `currency` is required.

iamacook:
> Question: How confident are we that these were the only methods with issues, and that other methods are unaffected?

I manually looked through every `Query` decorator to confirm so I am pretty confident. Regarding other Swagger definitions, I am not sure. Usage of the new SDK would highlight any.

src/routes/safes/safes.controller.ts
PooyaRaki: I have seen some instances, but here for example `excludeSpam`. It has a default value though.

iamacook: Perhaps we should reconsider usage of the default pipes, and instead assign a default value manually. This would negate the requirement of `ApiQuery`. I'll merge this for now and we can reconsider this later.

PR #1939, opened 2024-09-20
Only return net `value`

### LINKED ISSUE #1938: Return net claimable staked and rewards amount
### Description

We currently return the `reward` and assumed `value` of staked ETH. This is, however, incorrect.

The currently used value for rewards does not take into account the fees and slashing may have affected the value. We therefore cannot assume the rewards are correct "as is" or that the value is divisble by 32. The value is also not accessable before exit.

### Reproduction steps

1. Stake and request exit.
2. Claim and observe the value present when not available.

### Expected behavior

The net `value` includes the (currently) available, originally staked ETH and rewards minus fee.

### Additional information

A new query parameter for the `/v1/eth/stakes` endpoint (`onchain_v1_include_net_rewards`) will return `net_claimable_consensus_rewards` when `true.

iamacook:
## Summary

Resolves #1938

We currently return an assumed `value` by number of validators, e.g. 1 x validator = 32 ETH. This does not, however, take into account slashing. The value [can be between 31-32 ETH](https://github.com/kilnfi/staking-contracts/blob/84981714f054fd68e61a080ff99c02e3c4303284/src/contracts/ConsensusLayerFeeDispatcher.sol#L74).

The `rewards` we return is also historically accumulative. The user can therefore only claim a partial amount of this.

After discussion with Kiln, we now have access to a new net value that combines to claimable staked ETH and rewards. This is now returned as the `value` of validator exit request/withdrawals and `rewards` removed. They are combined as we don't show one or the other in the UI.

## Changes

- Add `net_claimable_consensus_rewards` to `Stake` schema/entity.
- Add new query to `KilnApi['getStakes']`.
- Remove all instances of `rewards`.
- Return value based on new new `net_claimable_consensus_rewards` value.
- Update tests accordingly.

### DIFF
--- a/src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
+++ b/src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
@@ -419,9 +419,8 @@ describe('NativeStakingMapper', () => {
           estimatedExitTime: networkStats.estimated_exit_time_seconds,
           estimatedWithdrawalTime:
             networkStats.estimated_withdrawal_time_seconds,
-          value: '96000000000000000000', // 3 public keys in the transaction data => 3 validators * 32 eth
+          value: stakes[0].net_claimable_consensus_rewards,
           numValidators: 3, // 3 public keys in the transaction data => 3 validators
-          rewards: stakes[0].rewards,
           tokenInfo: {
             address: NULL_ADDRESS,
             decimals: chain.nativeCurrency.decimals,
@@ -441,8 +440,8 @@ describe('NativeStakingMapper', () => {
         .build();
       const networkStats = networkStatsBuilder().build();
       const stakes = [
-        stakeBuilder().with('rewards', '2').build(),
-        stakeBuilder().with('rewards', '3').build(),
+        stakeBuilder().with('net_claimable_consensus_rewards', '2').build(),
+        stakeBuilder().with('net_claimable_consensus_rewards', '3').build(),
       ];
       const validatorPublicKey = faker.string.hexadecimal({
         length: KilnDecoder.KilnPublicKeyLength * 3,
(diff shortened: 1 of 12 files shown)


src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
```diff
@@ -486,9 +485,11 @@ describe('NativeStakingMapper', () => {
           estimatedExitTime: networkStats.estimated_exit_time_seconds,
           estimatedWithdrawalTime:
             networkStats.estimated_withdrawal_time_seconds,
-          value: '96000000000000000000', // 3 public keys in the transaction data => 3 validators * 32 eth
+          value: (
+            +stakes[0].net_claimable_consensus_rewards! +
+            +stakes[1].net_claimable_consensus_rewards!
+          ).toString(),
```
iamacook, line 491: These could be summed programatically but this is done to ensure the tests break if any adjustments are made. The same for all other instances.

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -353,7 +324,10 @@ export class NativeStakingMapper {
       safeAddress: args.safeAddress,
       validatorsPublicKeys: this.splitPublicKeys(publicKeys),
     });
-    return stakes.reduce((acc, stake) => acc + Number(stake.rewards), 0);
+    return stakes.reduce((acc, stake) => {
+      const netValue = stake.net_claimable_consensus_rewards ?? '0';
```
PooyaRaki, line 328: nit: If `net_claimable_consensus_rewards` is always expected to be a number, could we update the schema to automatically parse it as such?"

PooyaRaki (approved):

PR #1942, opened 2024-09-23
Orm

PooyaRaki:
## Summary

This pull request introduces TypeORM as the Object-Relational Mapping (ORM) solution for the project. TypeORM simplifies database interactions and enhances code maintainability by mapping database tables to application entities.

## Benefits

Reduces boilerplate code and simplifies database operations.
Enhances code readability and maintainability.
Provides built-in support for migrations and schema management.
Improves error handling and validation through TypeORM features.

## Changes

Integrated TypeORM into the project.
Updated database connection configuration to support TypeORM.
Created entity models corresponding to database tables.
Updated unit tests to ensure compatibility with TypeORM.

## Next Steps

Explore additional features like eager loading and query optimizations in future iterations.


src/datasources/notifications/entities/notification-devices.entity.ts
```diff
@@ -0,0 +1,48 @@
...
+  Column,
+  Entity,
+  Unique,
+  OneToMany,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+
+@Entity('push_notification_devices')
+@Unique('device_uuid', ['device_uuid'])
+@Check('device_type', 'device_type IN ("ANDROID", "IOS", "WEB")')
```
iamacook, line 14: Can we set this to the values of the enum via TypeORM?

src/datasources/notifications/entities/notification-devices.entity.ts
```diff
@@ -0,0 +1,48 @@
...
+export class NotificationDevice {
+  @PrimaryGeneratedColumn()
+  id!: number;
+
+  @Column({
+    type: 'varchar',
+    length: 255,
+  })
+  device_type!: string;
```
iamacook, line 23: If we can use the enum in the `Check` decorator, this would then be typed that of the enum.

src/datasources/notifications/entities/notification-type.entity.ts
```diff
@@ -0,0 +1,28 @@
...
+@Unique('name', ['name'])
+export class NotificationType {
+  @PrimaryGeneratedColumn()
+  id!: number;
+
+  @Column({
+    type: 'varchar',
+    length: 255,
+  })
+  name!: string;
```
iamacook, line 20: As with the device, could we ensure that the `name` is of the `NotificationType` enum?

src/datasources/notifications/entities/notification-devices.entity.ts
```diff
@@ -0,0 +1,48 @@
...
+  Entity,
+  Unique,
+  OneToMany,
+  PrimaryGeneratedColumn,
+} from 'typeorm';
+
+@Entity('push_notification_devices')
+@Unique('device_uuid', ['device_uuid'])
+@Check('device_type', 'device_type IN ("ANDROID", "IOS", "WEB")')
+export class NotificationDevice {
```
iamacook, line 15: It can be done in a follow up, but I would be in favour of (on the domain level) validating these entities and, as such, we could then implement the type here, e.g.

```ts
// Datasource
export class NotificationDevice implements z.infer<NotificationDeviceSchema> { ... }

// Domain
function example() {
  const deviceFromDatabase = // ...
  return NotificationDeviceSchema.parse(device)
}
```

src/datasources/notifications/entities/notification-subscription.entity.ts
```diff
@@ -0,0 +1,65 @@
...
+    length: 42,
+  })
+  safe_address!: string;
+
+  @Column({
+    type: 'varchar',
+    nullable: true,
+    length: 42,
+  })
+  signer_address!: string | null;
```
iamacook, line 45: I think we should use the refinement functionality to call viem's `getAddress`, ensuring that these addresses are checksummed. We can then set the type to `0x${string}` as well.

src/app.module.ts
```diff
@@ -121,6 +124,16 @@ export class AppModule implements NestModule {
           // return 500 for files that do not exist instead of a 404
           exclude: ['/(.*)'],
         }),
+        TypeOrmModule.forRootAsync({
+          imports: [ConfigModule],
+          useFactory: (configService: ConfigService) => {
+            return {
+              ...{ autoLoadEntities: true, manualInitialization: true },
```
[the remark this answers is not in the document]
PooyaRaki: I considered it as well, but for now, I think it's too much. Since the app heavily relies on manual initialization, it seems better to hardcode it here. For both cases, since they shouldn't be changed, hardcoding seems like the best approach IMO. What do you think?

src/datasources/notifications/entities/notification-devices.entity.ts
PooyaRaki: Do you mean we should dynamically create the string? TypeOrm `Check` only accepts string. I wouldn't dynamically create the string as that kills the readability here.

src/app.module.ts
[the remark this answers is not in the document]
iamacook: I like the idea of it being centralised, e.g. if we hardcoded these values in our `configuration` file but I am equally impartial. If it makes the most sense to have them here, we can keep it as is.

src/datasources/notifications/entities/notification-devices.entity.ts
PooyaRaki: IMO, The best solution is to define this field as an enum in the database. But, that's outside the scope of this PR.

src/datasources/notifications/entities/notification-devices.entity.ts
iamacook: It would be nice to somehow have them linked. We aren't otherwise certain that our enum and this are the same.

src/datasources/notifications/entities/notification-type.entity.ts
PooyaRaki: Yes, we can make it an ENUM in the database. I think it should be fixed in another PR. What do you think?

src/datasources/notifications/entities/notification-devices.entity.ts
iamacook: That would be preferable. If it's any benefit, we already have the `DeviceType` enum in the project.

src/datasources/notifications/entities/notification-devices.entity.ts
PooyaRaki: How about changing the field type in the database to ENUM? We can handle that in a separate PR.
P.S. I didn't modify the database fields, as it was beyond the scope of this PR.

src/datasources/notifications/entities/notification-type.entity.ts
iamacook: Agreed. We already have a reference enum in the project - `NotificationType`.

src/datasources/notifications/entities/notification-devices.entity.ts
PooyaRaki: Do you agree with me that it should be refactored in another PR?

iamacook: Yes, that's fine with me.

src/config/entities/postgres.config.ts
```diff
@@ -0,0 +1,49 @@
...
+  username: string;
+  password: string;
+  database: string;
+  ssl: {
+    enabled: boolean;
+    caPath: string | undefined;
+    requestCert: boolean;
+    rejectUnauthorized: boolean;
+  };
+}
```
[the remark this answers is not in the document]
PooyaRaki: Unfortunately as the types are different and this is the input type it's not possible to infer this from `DataSource`

src/datasources/notifications/entities/notification-devices.entity.ts
PooyaRaki: I like the idea, but let's do it in the notification PR.

src/datasources/notifications/entities/notification-subscription.entity.ts
PooyaRaki: This will be addressed in a separate PR focused on notifications.

src/config/entities/configuration.ts
```diff
@@ -131,11 +131,13 @@ export default () => ({
       apiKey: process.env.INFURA_API_KEY,
     },
   },
+  typeorm: { autoLoadEntities: true, manualInitialization: true },
```
iamacook, line 134: Nit: I've move these under the `db` property.

src/config/entities/postgres.config.ts
```diff
@@ -0,0 +1,49 @@
...
+  return {
+    type: 'postgres',
+    synchronize: false,
+    host: postgresEnvConfig.host,
+    port: ~~postgresEnvConfig.port,
+    username: postgresEnvConfig.username,
+    password: postgresEnvConfig.password,
+    database: postgresEnvConfig.database,
+    migrations: ['dist/migrations/*.js'],
+    // logging: !isCIContext,
```
iamacook, line 40: If we can't have CI logging, I'd remove this.

src/datasources/db/v2/postgres-database.service.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Fetches the database connection. If the connection is not initialized, it initializes the connection.
+   *
+   * @returns {Promise<DataSource>} The database connection.
+   */
+  public async fetchDatabaseConnection(): Promise<DataSource> {
+    if (!this.dataSource.isInitialized) {
+      this.loggingService.info('PostgresDatabaseService initialized...');
+      await this.dataSource.initialize();
```
iamacook, line 24: Do we need to "de-initialize" this `onModuleDestroy`?

src/config/entities/postgres.config.ts
PooyaRaki: I would say let's keep it, it's like a TODO. I should have been more clear by adding a TODO.

src/datasources/db/v2/postgres-database.service.ts
PooyaRaki: `NestJS` automatically terminates idle connections on shutdown. I suggest we avoid adding onModuleDestroy and stick with the default behavior.

src/config/entities/configuration.ts
PooyaRaki: In my opinion, a better approach than the current one would be to use `db.orm` and `Array<db.databases>`, allowing us the flexibility to add multiple databases in the future and easily iterate over them. Do you believe we should move Typeorm under db?

iamacook: `db.orm` sounds good to me. I wouldn't create and array out of them for now though as we don't have more than one. What do you think?

src/datasources/db/v2/postgres-database.service.ts
iamacook: Unrelated to this PR but a question in general: should we then aim to remove other instances of `onModuleDestroy` that fall under this criteria? We should file an issue if so.

hectorgomezv: Could you provide docs regarding that? I'm searching and I cannot find how exactly NestJS/TypeOrm handle the shutdown. We also need to take Kubernetes into consideration, where a kill/reboot of a pod would happen and only a SIGTERM can be received. `OnModuleDestroy` would guarantee proper connection closing in that case. 

Anyway, we can just monitor the opened connections to get confidence around this.

src/config/entities/configuration.ts
hectorgomezv: `db.orm` also sounds good to me. Regarding the array, no strong opinion, but I'd slightly prefer not having it until it's needed. It can be easily added later if we need to add another db.

src/datasources/db/v2/postgres-database.service.ts
PooyaRaki: @iamacook Well, If they are automatically closed on shut down, we can add an issue to remove the redundant call, As in some cases like here calling the shutdown function manually may cause an error if the connection is already terminated, you need to check whether the connection is active or not before calling the shut down.

PooyaRaki: @hectorgomezv Sure, https://docs.nestjs.com/techniques/database
there is an option that is already deprecated called, `keepConnectionAlive:
`If true, connection will not be closed on the application shutdown (default: false)`

I have also tested the functionality and can confirm that the connection is properly terminated during a graceful shutdown.

We can add `OnModuleDestroy` if you think it's necessary for extra verification, but the connection is already terminated before this hook is triggered.

src/config/entities/configuration.ts
PooyaRaki: @iamacook @hectorgomezv What do you think about moving `orm` under `db` and placing `postgres` under `db.connection`? My concern is that having `orm` and `postgres` under the same parent could lead to ambiguity. For now, `db.connection` will remain an object, but in the future, we can convert it to an array and rename it to `db.connections`(add an **s** to make it plural).

iamacook:
To keep track of our discussion, we need to consider automatic migrations across multiple instances.

src/datasources/db/v2/postgres-database.service.ts
hectorgomezv: Great, no need then to handle that in the `onModuleDestroy` hook, thanks for clarifying!

hectorgomezv (approved):

src/config/entities/configuration.ts
hectorgomezv: Sounds good to me @PooyaRaki, thanks! 👍

src/app.module.ts
```diff
@@ -126,6 +129,21 @@ export class AppModule implements NestModule {
...
+          useFactory: async (configService: ConfigService) => {
+            const typeormConfig = await configService.getOrThrow('typeorm');
+            const postgresConfigObject = postgresConfig(
+              await configService.getOrThrow('db.postgres'),
+            );
+
+            return {
+              ...typeormConfig,
+              ...postgresConfigObject,
+            };
```
iamacook, line 143: Will this throw if the respective configs don't match the expected structure? Reason being, `getOrThrow` returns a "typeless" object and we have no warning until runtime. This is generally why we `getOrThrow` primitives.

PooyaRaki: No, this does not throw in case types are different. On the other hand, we have configuration tests.

iamacook: Good point. In general, any issues would cause a crash before potential issues occur, e.g. migration breaks.

PooyaRaki:
> To keep track of our discussion, we need to consider automatic migrations across multiple instances.

This has been implemented in #1988

iamacook (approved):

PR #1948, opened 2024-09-24
Rename ICacheService get and set methods

hectorgomezv:
## Summary
This PR renames `get` and `set` functions in `ICacheService` to `hGet` and `hSet`.

For historical reasons, the CGW uses _hashes_ to operate the cache, meaning each key holds a _dictionary_/_object_ composed of multiple key-values. This is because this structure provides the flexibility to store various items under the same key and delete them deterministically. (See: https://redis.io/docs/latest/develop/data-types/hashes/)

This PR renames the `get` and `set` functions to make developers aware these functions are not operating over simple keys (`GET`/`SET` in Redis) but over dictionaries or _hashes_ (`HGET`/`HSET` in Redis).

## Changes
- Renames `get` and `set` functions in `ICacheService` to `hGet` and `hSet`.
- Includes https://github.com/safe-global/safe-client-gateway/pull/1953


src/datasources/cache/redis.cache.service.ts
```diff
@@ -31,7 +31,13 @@ export class RedisCacheService
     return this.client.ping();
   }
 
-  async set(
+  async getCounter(key: string): Promise<number | null> {
+    const keyWithPrefix = this._prefixKey(key);
+    const value = await this.client.get(keyWithPrefix);
+    return value ? Number(value) : null;
```
iamacook, line 37: If the key returns a non-numeric string, this could be `NaN`. We should add a respective check. What do you think?

hectorgomezv: Fixed in [1f478d0](https://github.com/safe-global/safe-client-gateway/pull/1948/commits/1f478d011fc9e89cdefdc5a961a7c2e1efb9bb66)

Thanks!

src/datasources/cache/redis.cache.service.ts
```diff
@@ -31,7 +31,12 @@ export class RedisCacheService
     return this.client.ping();
   }
 
-  async set(
+  async getCounter(key: string): Promise<number | null> {
+    const value = await this.client.get(this._prefixKey(key));
+    return value ? Number(value) || null : null;
```
iamacook, line 36: If the cache returns `'0'`, this will return `null`.

hectorgomezv: True 🙈

Fixed and tested in [49b696e](https://github.com/safe-global/safe-client-gateway/pull/1948/commits/49b696e7f63c8d1dea67472f39b3136c122a1ce0)

iamacook (approved):

PR #1953, opened 2024-09-24
Add getCounter and setCounter to ICacheService

hectorgomezv:
## Summary
This PR implements shared counters via the cache layer. It adds functions to set/get numeric counter values via `ICacheService`.

## Changes
- Adds `getCounter` and `setCounter` functions to `ICacheService` and its implementations.
- Adds related testing.

### DIFF
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -1,6 +1,7 @@
 import { ICacheService } from '@/datasources/cache/cache.service.interface';
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 import { ICacheReadiness } from '@/domain/interfaces/cache-readiness.interface';
+import { isNumber } from 'lodash';
 
 export class FakeCacheService implements ICacheService, ICacheReadiness {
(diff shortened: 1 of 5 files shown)


src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -32,6 +33,11 @@ export class FakeCacheService implements ICacheService, ICacheReadiness {
     return Promise.resolve(1);
   }
 
+  getCounter(key: string): Promise<number | null> {
+    const value = this.cache[key];
+    return isNumber(value) ? Promise.resolve(value) : Promise.resolve(null);
```
iamacook, line 38: Nit: if we expect this to only be an integer, we can remove reliance on lodash and use `Number.isInteger` instead.

iamacook (approved):

src/datasources/cache/__tests__/fake.cache.service.ts
hectorgomezv: True, thanks! Done in [a295546](https://github.com/safe-global/safe-client-gateway/pull/1953/commits/a2955464350166bb48010ca21df49066a2ff27cd) (I had to add an extra check as `Number.isInteger` does not narrow the type).

PR #1955, opened 2024-09-24
Compare all deployment type addresses when relaying

### LINKED ISSUE #1954: Unable to create Safe with same address on different network via relayer
### Description

When trying to recreate a Safe with the same address from Celo on Gnosis Chain via the relayer, it fails and returns a 422.

### Reproduction steps

1. Create a 1.3.0 Safe on Celo.
2. Try to replay the creation transaction via the relayer on Gnosis Chain.

### Expected behavior

Relaying works.

### Additional information

This is likely caused by the singleton checks.

iamacook:
## Summary

Resolves #1954

It is possible to get both singular or multiple deployment addresses since https://github.com/safe-global/safe-deployments/pull/668 and, as such, there are respective `get*Deployment` and `get*Deployments` functions.

Since this adjustment in the deployments package, we never took this into account in the project. Therefore, if a chain has both canonical and EIP-155 deployments, for example, we would only be retrieving the first address found.

This adjusts all usage of the deployments package, ensuring we retrieve and compare against all deployment type addresses.

## Changes

- Create helpers to retrieve deployments, and add linting rule to ensure usage of them.
- Modify relaying checks to compare against a list of addresses instead of a singular one.
- Update tests accordingly.

### DIFF
--- a/src/domain/common/utils/deployments.ts
+++ b/src/domain/common/utils/deployments.ts
@@ -1,0 +1,118 @@
+// eslint-disable-next-line no-restricted-imports
+import {
+  getMultiSendCallOnlyDeployments as _getMultiSendCallOnlyDeployments,
+  getMultiSendDeployments as _getMultiSendDeployments,
+  getProxyFactoryDeployments as _getProxyFactoryDeployments,
+  getSafeL2SingletonDeployments as _getSafeL2SingletonDeployments,
+  getSafeSingletonDeployments as _getSafeSingletonDeployments,
+} from '@safe-global/safe-deployments';
+
+/**
+ * Returns a list of official ProxyFactory addresses based on given {@link Filter}.
+ *
+ * @param {string} args.chainId - the chain ID to filter deployments by
+ * @param {string} args.version - the version to filter deployments by
+ *
+ * @returns {Array<`0x${string}`>} - a list of checksummed ProxyFactory addresses
+ */
+export function getProxyFactoryDeployments(args: Filter): Array<`0x${string}`> {
+  return formatDeployments(_getProxyFactoryDeployments, args);
+}
+
+/**
+ * Returns a list of official L1 singleton addresses based on given {@link Filter}.
+ *
+ * @param {string} args.chainId - the chain ID to filter deployments by
+ * @param {string} args.version - the version to filter deployments by
+ *
+ * @returns {Array<`0x${string}`>} - a list of checksummed L1 singleton addresses
+ */
+export function getSafeSingletonDeployments(
+  args: Filter,
+): Array<`0x${string}`> {
+  return formatDeployments(_getSafeSingletonDeployments, args);
+}
+
+/**
+ * Returns a list of official L2 singleton addresses based on given {@link Filter}.
+ *
+ * @param {string} args.chainId - the chain ID to filter deployments by
+ * @param {string} args.version - the version to filter deployments by
+ *
+ * @returns {Array<`0x${string}`>} - a list of checksummed L2 singleton addresses
+ */
+export function getSafeL2SingletonDeployments(
+  args: Filter,
+): Array<`0x${string}`> {
+  return formatDeployments(_getSafeL2SingletonDeployments, args);
+}
+
+/**
+ * Returns a list of official MultiSendCallOnly addresses based on given {@link Filter}.
+ *
+ * @param {string} args.chainId - the chain ID to filter deployments by
+ * @param {string} args.version - the version to filter deployments by
+ *
+ * @returns {Array<`0x${string}`>} - a list of checksummed MultiSendCallOnly addresses
+ */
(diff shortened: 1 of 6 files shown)


src/domain/common/utils/deployments.ts
```diff
@@ -0,0 +1,118 @@
...
+
+type Filter = {
+  chainId: string;
+  version: string;
+};
+
+/**
+ * Helper to remap {@link SingletonDeploymentV2} to a list of checksummed addresses.
+ *
+ * @param getDeployments - deployment getters from @safe-global/safe-deployments
```
PooyaRaki, line 84: Param types missing

iamacook: Added in 842426ee25749a095fff340f0f5722ee9a5990ea.

src/domain/common/utils/deployments.ts
```diff
@@ -0,0 +1,118 @@
...
+ * @param {string} args.chainId - the chain ID to filter deployments by
+ * @param {string} args.version - the version to filter deployments by
+ *
+ * @returns {Array<`0x${string}`>} - a list of checksummed MultiSend addresses
+ */
+export function getMultiSendDeployments(args: Filter): Array<`0x${string}`> {
+  return formatDeployments(_getMultiSendDeployments, args);
+}
+
+type Filter = {
```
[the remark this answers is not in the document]
iamacook: I think separate files for types is quite excessive outside of a class interface. I moved it to the top of the file in 842426ee25749a095fff340f0f5722ee9a5990ea.

src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -228,28 +228,15 @@ export class LimitAddressesMapper {
...
-    const multiSendCallDeployment = getMultiSendDeployment({
-      version: args.version,
-      network: args.chainId,
-    });
-    return (
-      multiSendCallDeployment?.networkAddresses[args.chainId] ===
-        args.address || multiSendCallDeployment?.defaultAddress === args.address
-    );
+    const multiSendCallDeployments = getMultiSendDeployments(args);
+    return multiSendCallDeployments.includes(args.address);
```
hectorgomezv, line 239: Nit: wdyt about simplifying this a bit by:

```typescript
return (
      getMultiSendCallOnlyDeployments(args).includes(args.address) ||
      getMultiSendDeployments(args).includes(args.address)
    );
```

src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -318,19 +298,11 @@ export class LimitAddressesMapper {
...
-    });
+    const safeL1Deployments = getSafeSingletonDeployments(args);
+    const safeL2Deployments = getSafeL2SingletonDeployments(args);
 
-    const isL1Singleton =
-      safeL1Deployment?.networkAddresses[args.chainId] === singleton;
-    const isL2Singleton =
-      safeL2Deployment?.networkAddresses[args.chainId] === singleton;
+    const isL1Singleton = safeL1Deployments.includes(singleton);
+    const isL2Singleton = safeL2Deployments.includes(singleton);
```
hectorgomezv, line 305: Nit: in this case, I think it could be also simplified, and optimized (potentially avoiding a second lookup):

```typescript
return (
      getSafeSingletonDeployments(args).includes(singleton) ||
      getSafeL2SingletonDeployments(args).includes(singleton)
    );
```

Wdyt?

hectorgomezv (approved):

src/domain/relay/limit-addresses.mapper.ts
iamacook: Agreed. Changed in 79fa30cc6d681b6b4f8ceaed1d063464d5ade42e.

src/domain/relay/limit-addresses.mapper.ts
iamacook: Agreed. Changed in 79fa30cc6d681b6b4f8ceaed1d063464d5ade42e.

src/domain/relay/limit-addresses.mapper.ts
```diff
@@ -228,28 +228,14 @@ export class LimitAddressesMapper {
...
 
-    const multiSendCallDeployment = getMultiSendDeployment({
-      version: args.version,
-      network: args.chainId,
-    });
-    return (
-      multiSendCallDeployment?.networkAddresses[args.chainId] ===
-        args.address || multiSendCallDeployment?.defaultAddress === args.address
-    );
+    return getMultiSendDeployments(args).includes(args.address);
```
hectorgomezv, line 238: Nit: we could just

```typescript
return getMultiSendCallOnlyDeployments(args).includes((args.address) || getMultiSendDeployments(args).includes(args.address);
```

And therefore get rid of all the variables and the `if` statement, right? Do you think it would be less readable?

hectorgomezv (approved):

PR #1956, opened 2024-09-24
Add `getTransactionStatus` to staking datasource/domain

iamacook:
## Summary

This adds a new `IStakingApi['getTransactionStatus']` and propagates it into `IStakingRepository`.

Note: it will be used to retrieve `deposit` event logs for `_publicKey` access in a follow up.

## Changes

- Add `TransactionStatusSchema` and infer `TransactionStatus` entity from it.
- Add relevant builders.
- Add Kiln-specific transaction status retrieval.
- Propagate into staking repository.
- Add new `staking_transaction_status` cache key.

### DIFF
--- a/src/datasources/staking-api/entities/transaction-status.entity.ts
+++ b/src/datasources/staking-api/entities/transaction-status.entity.ts
@@ -1,0 +1,41 @@
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { HexSchema } from '@/validation/entities/schemas/hex.schema';
+import { NumericStringSchema } from '@/validation/entities/schemas/numeric-string.schema';
+import { z } from 'zod';
+
+export const TransactionStatusReceiptLogSchema = z.object({
+  address: AddressSchema,
+  topics: z.array(HexSchema),
+  data: HexSchema,
+  blockHash: HexSchema,
+  blockNumber: NumericStringSchema,
+  blockTimestamp: HexSchema,
+  transactionHash: HexSchema,
+  transactionIndex: z.number(),
+  logIndex: z.number(),
+  removed: z.boolean(),
+});
+
+export const TransactionStatusReceiptSchema = z.object({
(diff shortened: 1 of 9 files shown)


src/datasources/staking-api/entities/transaction-status.entity.ts
```diff
@@ -0,0 +1,41 @@
...
+  transactionHash: HexSchema,
+  transactionIndex: z.number(),
+  blockHash: HexSchema,
+  blockNumber: NumericStringSchema,
+  gasUsed: NumericStringSchema,
+  effectiveGasPrice: NumericStringSchema,
+  from: AddressSchema,
+  to: AddressSchema,
+  contractAddress: AddressSchema.nullish().default(null),
+});
```
hectorgomezv, line 34: Nit: Correct me if I am wrong, but taking a look at https://github.com/safe-global/safe-client-gateway/pull/1957, it seems only `txStatus.receipt.logs` is actually used by the consuming class (`NativeStakingMapper`). I know we have some other schemas where we define properties that are not actually used by any class, but, do you think this is worthy? 

I mean, by mapping unused properties coming from 3rd parties responses, we are in fact making our service less resilient: changes on those responses would cause validation errors even if we don't need the changed properties. It also forces us to cover these properties on schema tests and test entity builders. What do you think about just mapping the needed properties when it comes to 3rd parties responses?

hectorgomezv (approved):

src/datasources/staking-api/entities/transaction-status.entity.ts
iamacook: That's correct, of which we only use `data` and `topics` and I agree.

As per your suggestion, I removed the unnecessary validation in dc111525579fc132908f2fddbc4b220c90706aa9.

hectorgomezv: Thanks!

hectorgomezv (approved):

PR #1957, opened 2024-09-24
Get public keys from `DepositEvent`

iamacook:
## Summary

In order to return the status of a given staking transaction, we fetch the `Stake`s from Kiln via validator keys. Generally, this is extracted from the `DataDecoded`. However, `deposit` transactions have no `_publicKeys` parameter.

This fetches the transaction status of a deposit transaction via it's transaction hash and, from it, decodes the public keys of the related validators. With that, the aforementioned status mapping is done.

## Changes

- Add `KilnDecoder['decodeDepositEvent']`.
- Decode public keys from `deposit` transaction status.


src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -347,6 +388,9 @@ export class NativeStakingMapper {
    * Gets the net value (staked + rewards) to withdraw from the native staking deployment
    * based on the length of the publicKeys field in the transaction data.
    *
+   * Note: this can only be used with `validatorsExit` or `batchWithdrawCLFee` transactions
+   * as the have `_publicKeys` field in the decoded data.
```
iamacook, line 392: We should reconsider all the `*FromDataDecoded` methods as they do not work for `deposit` calls. We can get the respective value by means of decoding the transaction data directly.

src/routes/transactions/mappers/common/transaction-info.mapper.ts
```diff
@@ -342,17 +342,11 @@ export class MultisigTransactionInfoMapper {
     }
 
     try {
-      const isConfirmed =
-        'confirmations' in transaction &&
-        !!transaction.confirmations &&
-        transaction.confirmations.length >= transaction.confirmationsRequired;
```
iamacook, line 348: The general `txStatus` signifies if a transaction is awaiting signatures/executions. The `txInfo.status` of a staking transaction is inherently staking specific, e.g. `NOT_STAKED`, `DEPOSIT_IN_PROGRESS`, etc.

src/domain/staking/contracts/decoders/__tests__/kiln-decoder.helper.spec.ts
```diff
@@ -13,6 +14,7 @@ const mockLoggingService = {
   warn: jest.fn(),
 } as jest.MockedObjectDeep<ILoggingService>;
 
+// TODO: Move function encoding to kiln-encoder.builder.ts
```
iamacook, line 17: This is covered in #1959, which I've marked as a draft until this is merged.

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -112,6 +113,47 @@ export class NativeStakingMapper {
...
+          ],
+        });
+      })
+      .filter(<T>(event: T | null): event is T => event !== null);
+
+    if (depositEvents.length === 0) {
+      return [];
+    }
+
+    // There is only one DepositEvent per transaction
```
hectorgomezv, line 153: Nit: Do you think it would be useful to emit a warning log line if this condition is not met for some reason?

hectorgomezv (approved):

src/routes/transactions/mappers/common/native-staking.mapper.ts
iamacook: As the contracts are immutable, this should inherently never happen. However, on the off chance it does, warning makes sense. I added a `warn` and a comment in 0fe91cda00fefc23f7d6f99e1dc17df74d9d4159.

hectorgomezv (dismissed):

hectorgomezv (approved):

PR #1963, opened 2024-09-26
Only return `net_claimable_consensus_rewards` as `value` before withdrawal

### LINKED ISSUE #1951: Post-withdrawal staking `value` resets to 0
### Description

We leverage the net value when mapping exit requests and withdrawals in the transaction list in order to display the `value` of said transaction. After successful withdrawal, however, the net value resets to 0. This means that the transactions show a value of 0.

### Reproduction steps

1. Successfully withdraw and observe exit requests/withdrawals showing a `value` of 0.

### Expected behavior

Successfuly withdrawals do not affect previously mapped values.

### Additional information

_No response_

iamacook:
## Summary

Resolves #1951

We explicitly return the `net_claimable_consensus_rewards` from Kiln's `v1/eth/stakes` endpoint. This is the combined value of staked ETH and current rewards. However, after successful execution of a withdrawal, this value changes to `0`.

This makes the following changes to accomodate for the above:

When exit requests are being proposed, the net claimable rewards theoretically still exists on Kiln's API. It is therefore returned in the confirmation view entity. After execution, it is no longer there. Therefore, it has been removed from the `txInfo` as the value would disappear when moving to the history. (Web will display number of validators instead.)

Withdrawal requests now fetch the net claimable rewards when queued but instead decode the value from the relevant `Withdrawal` events of an executed withdrawal.

## Changes

- Remove `value` from `NativeStakingValidatorsExitTransactionInfo`.
- Implement `Withdrawal` event decoding.
- Move methods shared between transaction/confirmation view mapping to `KilnNativeStakingHelper`.
- Add relative `Withdrawal` encoder/builder.
- Decode `value` of withdrawals after execution.

### DIFF
--- a/src/domain/staking/contracts/decoders/__tests__/kiln-decoder.helper.spec.ts
+++ b/src/domain/staking/contracts/decoders/__tests__/kiln-decoder.helper.spec.ts
@@ -3,6 +3,7 @@ import {
   depositEncoder,
   depositEventEventBuilder,
   requestValidatorsExitEncoder,
+  withdrawalEventBuilder,
 } from '@/domain/staking/contracts/decoders/__tests__/encoders/kiln-encoder.builder';
 import { KilnDecoder } from '@/domain/staking/contracts/decoders/kiln-decoder.helper';
 import { ILoggingService } from '@/logging/logging.interface';
@@ -15,7 +16,6 @@ const mockLoggingService = {
   warn: jest.fn(),
 } as jest.MockedObjectDeep<ILoggingService>;
 
-// TODO: Move function encoding to kiln-encoder.builder.ts
 describe('KilnDecoder', () => {
   let kilnDecoder: KilnDecoder;
 
@@ -102,7 +102,7 @@ describe('KilnDecoder', () => {
   });
 
   describe('decodeDepositEvent', () => {
-    it('decodes a deposit event correctly', () => {
+    it('decodes a DepositEvent correctly', () => {
       const depositEventEvent = depositEventEventBuilder();
       const { data, topics } = depositEventEvent.encode();
 
@@ -114,8 +114,11 @@ describe('KilnDecoder', () => {
       ).toStrictEqual(depositEventEvent.build());
     });
 
-    // Note: we cannot test whether null is returned for a non-DepositEvent
-    // as only DepositEvent is included in the ABI
+    it('returns null if the data is not a DepositEvent', () => {
+      const { data, topics } = withdrawalEventBuilder().encode();
+
+      expect(kilnDecoder.decodeDepositEvent({ data, topics })).toBe(null);
+    });
 
     it('returns null if the data is not a DepositEvent', () => {
       const data = faker.string.hexadecimal({ length: 514 }) as `0x${string}`;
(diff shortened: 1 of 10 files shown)


src/domain/staking/contracts/decoders/__tests__/kiln-decoder.helper.spec.ts
```diff
@@ -114,8 +114,11 @@ describe('KilnDecoder', () => {
       ).toStrictEqual(depositEventEvent.build());
     });
 
-    // Note: we cannot test whether null is returned for a non-DepositEvent
-    // as only DepositEvent is included in the ABI
+    it('returns null if the data is not a DepositEvent', () => {
+      const { data, topics } = withdrawalEventBuilder().encode();
+
+      expect(kilnDecoder.decodeDepositEvent({ data, topics })).toBe(null);
+    });
```
iamacook, line 121: As we now have the `Withdrawal` event in the ABI, this test can be added.

src/routes/transactions/helpers/kiln-native-staking.helper.spec.ts
```diff
@@ -0,0 +1,245 @@
...
+    it('should return a batched `batchWithdrawCLFee` transaction', () => {});
+
+    it('should return null if a `batchWithdrawCLFee` transaction is not from a known staking contract', () => {});
+
+    it('should return null if a batchWithdrawCLFee `deposit` transaction is not from a known staking contract', () => {});
+
+    it('should return null if the transaction is not a `batchWithdrawCLFee` transaction', () => {});
+
+    it('should return null if the transaction batch contains no `batchWithdrawCLFee` transaction', () => {});
+  });
```
iamacook, line 69: This will be added in a follow up as they are out of scope.

src/routes/transactions/helpers/kiln-native-staking.helper.ts
```diff
@@ -104,6 +111,92 @@ export class KilnNativeStakingHelper {
...
+  ): Array<`0x${string}`> {
+    if (
+      !['requestValidatorsExit', 'batchWithdrawCLFee'].includes(
+        dataDecoded.method,
+      )
+    ) {
+      throw new UnprocessableEntityException(
+        `${dataDecoded.method} does not contain _publicKeys`,
+      );
+    }
```
iamacook, line 166: Although we have a comment regarding this, I added the extra check as it threw me off a few times.

src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
```diff
@@ -325,8 +337,6 @@ describe('NativeStakingMapper', () => {
         ])
         .build();
       const transaction = multisigTransactionBuilder()
-        .with('confirmationsRequired', 2) // 2 confirmations required
-        .with('confirmations', [confirmationBuilder().build()]) // only 1 confirmation
```
iamacook, line 329: Confirmations are no longer relevant, hence removing them from these tests.

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -257,6 +260,52 @@ export class NativeStakingMapper {
...
+    const value = txStatus.receipt.logs
+      .map((log) => this.kilnDecoder.decodeWithdrawal(log))
+      .reduce((acc, cur) => {
+        if (cur === null) {
+          return acc;
+        }
+        return acc + cur.rewards;
+      }, BigInt(0));
+
+    return Number(value);
```
hectorgomezv, line 306: Nit: At this point `value` is `BigInt`. It would be likely more correct to have `BigInt` also as the function return type, but I guess that requires an additional, potentially big, refactor. Wdyt? At some point it could be beneficial to rethink some types that we expect to hold big numbers.

src/routes/transactions/helpers/kiln-native-staking.helper.spec.ts
hectorgomezv: Very nit: what do you think about setting them as "todos" instead of skipping the blocks?

Something like:
```typescript
describe('findDepositTransaction', () => {
    it.todo('should return a `deposit` transaction');
```

src/routes/transactions/mappers/common/native-staking.mapper.ts
iamacook: I agree that would make sence but the value returned by `getValueFromDataDecoded` is `number`. Do you think we should return `BigInt` from that instead?

hectorgomezv (approved):

src/routes/transactions/helpers/kiln-native-staking.helper.spec.ts
iamacook: Changed in 055fed1dc06cd51564316680a6c5fb42af0a8f1b.

src/routes/transactions/mappers/common/native-staking.mapper.ts
hectorgomezv: Yes, I think so, but I'd tackle this separately as it could be a major (and risky) refactor. 

In fact, it doesn't seem to make a lot of sense to do the math using `BigInt` and casting the result to `Number` after, the precision is lost anyway:

```javascript
> Number(BigInt(64 * 1e18) + BigInt(1))
64000000000000000000
```

hectorgomezv (approved):

PR #1964, opened 2024-09-26
Return validator addresses in staking entities

iamacook:
## Summary

As it is [not possible to always return the current value](https://github.com/safe-global/safe-client-gateway/pull/1963) for exit requests, we will not show the number of validators exiting in the transaction list of the UI. For individual validators, it's then possible to see their respective balances on the beacon chain explorer.

This returns the validators under a new `validators` property for all staking entities (apart from pre-executed deposits).

## Changes

- Add `validators` property to:
  - Confirmation view: exit requests and withdrawals
  - Transaction info: deposits (after execution), exit requests and withdrawals

### DIFF
--- a/src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
+++ b/src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
@@ -32,7 +32,7 @@ import { KilnNativeStakingHelper } from '@/routes/transactions/helpers/kiln-nati
 import { TransactionFinder } from '@/routes/transactions/helpers/transaction-finder.helper';
 import { NativeStakingMapper } from '@/routes/transactions/mappers/common/native-staking.mapper';
 import { faker } from '@faker-js/faker';
-import { getAddress } from 'viem';
+import { concat, getAddress } from 'viem';
 
 const mockStakingRepository = jest.mocked({
   getDeployment: jest.fn(),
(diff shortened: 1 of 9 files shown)


src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
```diff
@@ -533,9 +625,18 @@ describe('NativeStakingMapper', () => {
...
+          length: KilnDecoder.KilnPublicKeyLength,
+          // Transaction Service returns _publicKeys lowercase
+          casing: 'lower',
+        }),
+        faker.string.hexadecimal({
+          length: KilnDecoder.KilnPublicKeyLength,
+          casing: 'lower',
+        }),
+      ] as Array<`0x${string}`>;
+      const validatorPublicKey = concat(validators);
```
hectorgomezv, line 639: Nit: `validators` is not being expected in the response, I think it would be better to include it:

```typescript
      expect(actual).toEqual(
        expect.objectContaining({
          type: 'NativeStakingWithdraw',
          value: withdrawalEventParams.rewards.toString(),
          tokenInfo: {
            address: NULL_ADDRESS,
            decimals: chain.nativeCurrency.decimals,
            logoUri: chain.nativeCurrency.logoUri,
            name: chain.nativeCurrency.name,
            symbol: chain.nativeCurrency.symbol,
            trusted: true,
          },
          validators,
        }),
      );
```

src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
```diff
@@ -457,9 +543,18 @@ describe('NativeStakingMapper', () => {
...
```
hectorgomezv, line 557: Nit: same here, `validators` is not being checked in the test.

src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
iamacook: Added in 9dd2b7a708e9ba95104617eb3113bdfea4a30455.

src/routes/transactions/mappers/common/native-staking.mapper.spec.ts
iamacook: Added in 9dd2b7a708e9ba95104617eb3113bdfea4a30455.

hectorgomezv (approved):

PR #1967, opened 2024-09-26
Don't assume all RPC responses are strings

iamacook:
## Summary

We temporarily cache RPC requests. Currently, RPC responses are assumed to be strings (`eth_chainId`) and this is not the case, e.g. `eth_getBlockByNumber`

This adjusts the error handling logic to only throw if an error was returned by the RPC, not if the response if a string.

## Changes

- Don't throw if RPC result is not a string.
- Add test coverage

### DIFF
--- a/src/datasources/blockchain/blockchain-api.manager.ts
+++ b/src/datasources/blockchain/blockchain-api.manager.ts
@@ -93,7 +93,7 @@ export class BlockchainApiManager implements IBlockchainApiManager {
       const cache = await this.cacheService.hGet(cacheDir);
 
       if (cache != null) {
-        return cache;
+        return JSON.parse(cache);
       }
 
       const { error, result } = await rpcClient.request({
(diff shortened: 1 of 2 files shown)


src/datasources/blockchain/blockchain-api.manager.ts
```diff
@@ -114,7 +110,7 @@ export class BlockchainApiManager implements IBlockchainApiManager {
 
       await this.cacheService.hSet(
         cacheDir,
-        result,
+        typeof result === 'string' ? result : JSON.stringify(result),
```
hectorgomezv, line 113: Is this behavior difference needed? If the response is not a string (`eth_getBlockByNumber` returns an object) then it get stringfied, but the method is returning the cache key plain content in the line 96 (without a `JSON.parse` call). While this works for strings, it wouldn't work for other objects if I'm not wrong.

Have you considered using `JSON.stringify` and `JSON.parse` regardless of the value type?

iamacook: Good point. Changed in 494f05e15b485ee017aa31dc3d119d7a407138e3.

hectorgomezv (approved):

PR #1973, opened 2024-09-27
Deprecate `/transaction-confirmation` endpoint

### LINKED ISSUE #1927: Investigate using mappers for `transaction-confirmation` endpoint
### Description

We have `TransactionInfo` mappers for the transaction queue/history, built upon multiple Transaction Service endpoint responses. They map a multitude of Safe transaction types, e.g. various settings changes, swaps, etc.

With swaps, we created a new `transaction-confirmation` endpoint responsible for returning transaction info. pre-proposal. On a high level, this works by decoding the transaction data for a subset of swap/staking `TransactionInfo`. As the data structure is relatively similar, we should investigate whether we could reuse the aforementioned mappers to reduce duplication.

### Requirements

The following should be conducted on a singular mapper and, providing it is possible, done mapper-by-mapper.

- Reseach feasibility of the above.
- Providing there is consensys on pre-proposing transactions, either use mappers directly or decoded transaction data and map/pass it accordingly.

### Additional information

It was suggested that transactions could always be proposed and the mappers would then work as is. These transactions are, however, inherently untrusted.

### Time box

1/2 day

iamacook:
## Summary

Resolves #1927

Currently, the `/transaction-confirmation` returns a `TransactionInfo`-like entity for the client to render details about swaps and TWAPs. We equally have the `/preview` endpoint that returns a `TransactionInfo` entity.

This adds swap and staking test coverage to the `/preview` endpoint to ensure usage of it can replace the `/transaction-confirmation` one (migrated from the latter's suite), fixing any found issues.

## Changes

- Decode staking transaction data instead of using `dataDecoded` (because of batches).
- Correct types for `/preview` endpoint.
- Add swap/staking test coverage to `/preview` endpoint.
- Update other tests according.


src/routes/transactions/helpers/kiln-native-staking.helper.ts
```diff
@@ -24,153 +16,51 @@ export class KilnNativeStakingHelper {
...
-        transaction.data.startsWith(toFunctionSelector(args.signature)),
+    value: string;
+  }): { to?: `0x${string}`; data: `0x${string}`; value: string } | null {
+    const selector = toFunctionSelector(
+      KilnNativeStakingHelper.WITHDRAW_SIGNATURE,
+    );
+    return this.transactionFinder.findTransaction(
+      (transaction) => transaction.data.startsWith(selector),
       args,
     );
```
iamacook, line 63: When writing the tests, I realised we also check the deployment when mapping and fallback to a "standard" transaction there. This is therefore unnecessary, and now covered by tests.

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -186,10 +180,9 @@ export class NativeStakingMapper {
       }),
     ]);
     this.validateDeployment(deployment);
-    const publicKeys =
-      this.kilnNativeStakingHelper.getPublicKeysFromDataDecoded(
-        args.dataDecoded,
-      );
+    const publicKeys = this.kilnNativeStakingHelper.splitPublicKeys(
+      this.kilnDecoder.decodeValidatorsExit(args.data)!.parameters[0].value,
```
iamacook, line 184: There is a TODO added to the `KilnDecoder` to simplify this after deprecating the `/transaction-confirmations` endpoint.

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -117,29 +113,29 @@ export class NativeStakingMapper {
...
   private async getDepositPublicKeys(args: {
-    transaction: MultisigTransaction | ModuleTransaction | null;
+    txHash: `0x${string}` | null;
     chainId: string;
-  }): Promise<Array<`0x${string}`>> {
-    if (!args.transaction?.transactionHash) {
-      return [];
+  }): Promise<Array<`0x${string}`> | null> {
+    if (!args.txHash) {
+      return null;
```
iamacook, line 133: I will consult with clients before merge this. If there is no hash, there is technically no keys so this is (currently) incorrect.

src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -283,29 +277,34 @@ export class NativeStakingMapper {
...
         chainId: args.chainId,
         safeAddress: args.safeAddress,
-        dataDecoded: args.dataDecoded,
+        validatorsPublicKeys: args.publicKeys,
       });
+      return stakes.reduce((acc, stake) => {
+        const netValue = stake.net_claimable_consensus_rewards ?? '0';
+        return acc + Number(netValue);
+      }, 0);
     }
```
hectorgomezv, line 303: Nit: I'm OK with changing the way the withdraw value is retrieved/calculated in this PR, but it seems to me a relevant/big change. I'm under the impression this is not tied to the Confirmation View deprecation so what do you think about saying something about the decoding of the staking transaction data (instead of using dataDecoded) in the PR title?

hectorgomezv (approved):

src/routes/transactions/mappers/common/native-staking.mapper.ts
iamacook: After discussion, we've agreed to keep this as is. The preview endpoint does not fetch decoded data in a similar way to the native staking mapper, where we assign fallback values for Kiln contracts. As such, this change is necessary here.

src/routes/transactions/__tests__/controllers/preview-transaction.transactions.controller.spec.ts
```diff
@@ -364,4 +414,3299 @@ describe('Preview transaction - Transactions Controller (Unit)', () => {
         },
       });
   });
+
+  describe('CoW Swap', () => {
```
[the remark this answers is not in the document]
iamacook: Split in 3a2c23943d5673ee8c4f66d5151162ec675af2fc.

iamacook:
At the request of QA, I'm marking this as a draft until the multichain feature is released.

hectorgomezv (approved):

PR #1988, opened 2024-10-01
OrmMigrationAndLogger

PooyaRaki:
## Summary

Introduces a locking mechanism to TypeORM migrations to prevent conflicts when multiple instances attempt to run the same migration simultaneously.

Adds a custom query logger that logs queries to the console without including bound parameters.

## Changes

- Disables migrationsRun in the TypeORM settings.
- Introduces a new query logger to TypeORM.
- Ensures TypeORM migrations run after locking the database during startup.
- If migrations are already running in another instance, it retries a few times to check if they have been completed successfully before throwing an error.


src/config/entities/configuration.ts
```diff
@@ -131,26 +131,36 @@ export default () => ({
...
-          'false',
-        caPath: process.env.POSTGRES_SSL_CA_PATH,
+    migrator: {
+      numberOfRetries: 5,
+      retryAfter: 1 * 1000, // Milliseconds
+    },
+    orm: {
+      migrationsRun: false,
+      autoLoadEntities: true,
+      manualInitialization: true,
```
iamacook, line 142: Nit: the values that are more "universal", I would add relevant env. vars. (`migrator.numberOfRetries`, `migrator.retryAfter`, `orm.migrationsRun` - I think)

src/datasources/db/v2/database-migrator.service.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+  describe('migrate()', () => {
+    it('Should log the start and end of the migration process', async () => {
+      jest.spyOn(loggingService, 'info');
+
+      await databaseMigratorService.migrate();
+
+      expect(loggingService.info).toHaveBeenCalledWith(
+        'Migrations: Running...',
+      );
+      expect(loggingService.info).toHaveBeenCalledWith('Migrations: Finished.');
```
iamacook, line 90: Nit: I'd also check the number of calls.

General remark: this is an interesting edge case in the codebase. Normally we'd use a mock service that has jest mocks in place but the `TestLoggingModule` doesn't.

src/datasources/db/v2/database-migrator.service.ts
```diff
@@ -0,0 +1,157 @@
...
+import { Inject, Injectable } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
+import type { DataSource } from 'typeorm';
+
+interface LockSchema {
+  id: number;
+  name: number;
+}
+
+enum migrationStatus {
```
iamacook, line 15:
```suggestion
enum MigrationStatus {
```

src/datasources/db/v2/database-migrator.service.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+    databaseMigratorService = moduleRef.get<DatabaseMigrator>(DatabaseMigrator);
+    loggingService = moduleRef.get<ILoggingService>(LoggingService);
+  });
+
+  afterAll(async () => {
+    await postgresqlService.destroyDatabaseConnection();
+  });
+
+  describe('migrate()', () => {
+    it('Should log the start and end of the migration process', async () => {
```
iamacook, line 82: We should add other tests that cover locking, retrying and migrating.

src/datasources/db/v2/database-migrator.service.ts
```diff
@@ -0,0 +1,157 @@
...
+  /**
+   * Selects and retrieves locks from the database.
+   *
+   * @param {DataSource} connection The database connection to select locks from.
+   *
+   * @returns {Promise<Array<LockSchema>>} A promise that resolves to an array of LockSchema objects.
+   */
+  private async selectLock(connection: DataSource): Promise<Array<LockSchema>> {
+    return await connection.query('SELECT "id", "status" FROM "_lock";');
+  }
```
iamacook, line 124: Is there a reason [we can't use `pg_advisory_lock`](https://github.com/safe-global/safe-client-gateway/blob/ormMigrationAndLogger/src/datasources/db/v1/postgres-database.migration.hook.ts#L61-L69) here? We could use a similar pattern if so.

src/datasources/db/v2/database-migrator.service.spec.ts
```diff
@@ -0,0 +1,93 @@
+import { join } from 'path';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { Test, type TestingModule } from '@nestjs/testing';
+import { ConfigModule, ConfigService } from '@nestjs/config';
```
iamacook, line 4: We should us `@/config/configuration.module` instead of this directly.

src/datasources/db/v2/database-migrator.service.ts
```diff
@@ -0,0 +1,157 @@
+import { PostgresDatabaseService } from '@/datasources/db/v2/postgres-database.service';
+import {
+  LoggingService,
+  type ILoggingService,
+} from '@/logging/logging.interface';
+import { Inject, Injectable } from '@nestjs/common';
+import { ConfigService } from '@nestjs/config';
```
iamacook, line 7: As before regarding usage of `@/config/configuration.module`.

src/datasources/db/v2/postgres-database.module.spec.ts
```diff
@@ -0,0 +1,73 @@
+import { join } from 'path';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { Test, type TestingModule } from '@nestjs/testing';
+import { ConfigModule, ConfigService } from '@nestjs/config';
```
iamacook, line 4: As before regarding usage of `@/config/configuration.module`.

src/datasources/db/v2/postgres-database.module.ts
```diff
@@ -1,8 +1,11 @@
 import { Module } from '@nestjs/common';
 import { PostgresDatabaseService } from '@/datasources/db/v2/postgres-database.service';
+import { DatabaseMigrator } from '@/datasources/db/v2/database-migrator.service';
+import { ConfigModule } from '@nestjs/config';
```
iamacook, line 4: As before regarding usage of `@/config/configuration.module`.

src/app.module.ts
```diff
@@ -127,9 +129,9 @@ export class AppModule implements NestModule {
         TypeOrmModule.forRootAsync({
           imports: [ConfigModule],
```
iamacook, line 130: This can be done in a separate PR, but we though theoretically only use `@/config/configuration.module` here as well (if possible).

src/datasources/db/v2/postgres-database.service.ts
```diff
@@ -27,6 +45,20 @@ export class PostgresDatabaseService {
...
   }
 
+  /**
+   * Destroys the database connection. If the connection is not initialized, it initializes the connection.
+   *
+   * @returns {Promise<DataSource>} The database connection.
+   */
+  public async destroyDatabaseConnection(): Promise<DataSource> {
+    if (this.dataSource.isInitialized) {
+      this.loggingService.info('PostgresDatabaseService Destroyed...');
```
iamacook, line 55:
```suggestion
      this.loggingService.info('PostgresDatabaseService destroyed...');
```

src/datasources/db/v2/postgresql-logger.service.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  /**
+   * Logs the executed query along with the request URL if available.
+   *
+   * @param {string} query - The SQL query string that is being executed.
+   *
+   * @returns {void} This method does not return anything.
+   */
+  public logQuery(query: string): void {
+    console.log('executing query: ' + query);
```
iamacook, line 63: Nit: I'd put the query in quotation marks for clarity.

src/datasources/db/v2/postgresql-logger.service.ts
```diff
@@ -0,0 +1,65 @@
...
+    const messages = this.prepareLogMessages(logMessage, {
+      highlightSql: false,
+    });
+
+    for (const message of messages) {
+      switch (message.type ?? level) {
+        case 'log':
+        case 'schema-build':
+        case 'migration':
+          console.log(message.message);
```
iamacook, line 22: Looking at the types, this can also theoretically have a `prefix`.

Do you think we move these cases to the bottom of the `switch` and also make it the `default` handling as well? I'd like to think no other `type` would be added but just in case.

src/datasources/db/v2/postgresql-logger.service.ts
```diff
@@ -0,0 +1,65 @@
...
+          }
+          break;
+
+        case 'error':
+        case 'query-error':
+          if (message.prefix) {
+            console.error(message.prefix, message.message);
+          } else {
+            console.error(message.message);
+          }
```
iamacook, line 49: This is quite repetitive but I am also not a fan of the approach to simplify it. What do you think about abstracting it, e.g.

```ts
example(logger: (message?: unknown, ...rest: Array<unknown>) => void, message: LogMessage): void {
  if (message.prefix) {
    logger(message.prefix, message.message)
  } else {
    logger(message.message)
  }
}
```

src/main.ts
```diff
@@ -10,6 +11,9 @@ async function bootstrap(): Promise<void> {
   const applicationPort: string =
     configurationService.getOrThrow('application.port');
 
+  const databaseMigrator = app.get(DatabaseMigrator);
+  await databaseMigrator.migrate();
```
iamacook, line 15: This ties into my comment regarding the locking approach. I think this would make sense "localised" in an `onModuleInit` within the migrator service. It can then be hidden behind a feature flag easier, for example. What do you think?

iamacook (commented):
config

src/datasources/db/v2/postgresql-logger.service.ts
PooyaRaki: This is the implementation by `TypeOrm`, I prefer not to modify it since we only need to override the `logQuery` method.
https://orkhan.gitbook.io/typeorm/docs/logging

src/datasources/db/v2/postgresql-logger.service.ts
PooyaRaki: This is the implementation by TypeOrm, I prefer not to modify it since we only need to override the logQuery method.
https://orkhan.gitbook.io/typeorm/docs/logging

src/main.ts
PooyaRaki: I was considering this approach, but I personally prefer being explicit, especially since we want the migrations to execute on startup. This way, it’s clearer what runs at startup rather than hiding it in `onModuleInit`.

src/datasources/db/v2/database-migrator.service.ts
PooyaRaki: I also tried using pg_advisory_lock, but since these locks are session-based, I found them to be quite unreliable. If the session is terminated, the lock is released, which could potentially cause issues in the database.

src/config/entities/configuration.ts
PooyaRaki: I wouldn't add orm.migrationsRun to the .env file since it should never be set to true unless we remove the migrationLocker functionality. Regarding numberOfRetries and retryAfter, we can include those, but I believe not everything should be configurable, as passing incorrect values could lead to complications.

src/datasources/db/v2/postgresql-logger.service.ts
iamacook: Ok, agreed. Can you please add a comment with that link?

src/datasources/db/v2/postgresql-logger.service.ts
iamacook: Replied [above](https://github.com/safe-global/safe-client-gateway/pull/1988#discussion_r1789834553).

src/datasources/db/v2/postgresql-logger.service.ts
PooyaRaki: Of course.

src/main.ts
iamacook: Whilst I tend to agree, as we are feature-flag reliant and certain database-reliants features are behind some, e.g. accounts, this would prevent startup without enabling the database (which currently isn't the case on prod.)

src/datasources/db/v2/database-migrator.service.ts
iamacook: I see. Can we add a comment explaining that?

PooyaRaki: What about adding a `README` file to the module to explain this information?

iamacook: I am impartial. We don't tend to have localised `README`s but I have no strong opinion.

src/datasources/db/v2/database-migrator.service.ts
```diff
@@ -0,0 +1,157 @@
...
+    let numberOfIterations = 0;
+    const numberOfRetries = await this.configService.getOrThrow(
+      'db.migrator.numberOfRetries',
+    );
+    while (numberOfRetries >= numberOfIterations) {
+      ++numberOfIterations;
+      if (!(await this.lockExists(connection))) {
+        await this.insertLock(connection);
+        await this.runMigration(connection);
+        await this.truncateLocks(connection);
```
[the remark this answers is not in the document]
PooyaRaki: This is intentional. Ideally, errors shouldn't occur, but I wanted to lock the table temporarily in case of an error. This way, a developer will need to manually unlock the table after resolving the issue.I believe it's more reliable to unlock the table manually in the event of an error.

src/datasources/db/v2/postgresql-logger.service.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /**
+   * Logs the executed query along with the request URL if available.
+   *
+   * @param {string} query - The SQL query string that is being executed.
+   *
+   * @returns {void} This method does not return anything.
+   */
+  public logQuery(query: string): void {
```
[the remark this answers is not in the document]
PooyaRaki: Do you think we should only log failed queries?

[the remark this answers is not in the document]
hectorgomezv: Yes, I'm more inclined not to log SQL in staging/production environments by default, but logging the failing ones could also make sense. I'd leave this up to you! 🙂

Have you considered using `ILoggingService` somehow (or even instantiating a new `winston` logger here) to emit logs to the DEBUG level? We could activate the logs at any time by setting the appropriate log level in the staging/prod instances.

src/datasources/db/v2/database-migrator.service.ts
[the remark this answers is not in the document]
hectorgomezv: It makes sense to me, thanks for explaining!

src/datasources/db/v2/database-migrator.service.spec.ts
PooyaRaki: Unfortuantely, Since we need `env` variables, it's not possible to use `configuration.module`.

src/datasources/db/v2/database-migrator.service.ts
PooyaRaki: Unfortunately, Since we need `env` variables, it's not possible to use `configuration.module`.

src/datasources/db/v2/postgres-database.module.spec.ts
PooyaRaki: Unfortunately, Since we need `env` variables, it's not possible to use `configuration.module`.

src/datasources/db/v2/postgres-database.module.ts
PooyaRaki: Unfortunately, Since we need `env` variables, it's not possible to use `configuration.module`.

src/app.module.ts
PooyaRaki: Unfortunately, Since we need `env` variables, it's not possible to use `configuration.module`.

src/config/entities/configuration.ts
PooyaRaki: Added `numberOfRetries` and `retryAfter` to .env https://github.com/safe-global/safe-client-gateway/pull/1988/commits/f7da627dc0505039d12c1605cc646aeda59c9c68

src/datasources/db/v2/database-migrator.service.spec.ts
PooyaRaki: Fixed this one and Added more unit tests https://github.com/safe-global/safe-client-gateway/pull/1988/commits/f7da627dc0505039d12c1605cc646aeda59c9c68

src/datasources/db/v2/database-migrator.service.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/f7da627dc0505039d12c1605cc646aeda59c9c68

src/datasources/db/v2/database-migrator.service.spec.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/f7da627dc0505039d12c1605cc646aeda59c9c68

src/datasources/db/v2/postgres-database.service.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/f7da627dc0505039d12c1605cc646aeda59c9c68

src/datasources/db/v2/postgresql-logger.service.ts
```diff
@@ -0,0 +1,65 @@
+import { Injectable } from '@nestjs/common';
+import { AbstractLogger, type LogLevel, type LogMessage } from 'typeorm';
+
+@Injectable()
+export class PostgresqlLogger extends AbstractLogger {
```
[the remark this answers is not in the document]
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/f7da627dc0505039d12c1605cc646aeda59c9c68

src/datasources/db/v2/postgresql-logger.service.ts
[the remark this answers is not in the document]
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/f7da627dc0505039d12c1605cc646aeda59c9c68

src/datasources/db/v2/database-migrator.service.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/3f9292a5aed4148193ff80241f8e0afba86f902c

package.json
```diff
@@ -95,8 +95,7 @@
       "^.+\\.(t|j)s$": "ts-jest"
     },
     "roots": [
-      "./src",
-      "./migrations"
```
iamacook, line 99: I would suggest we only remove test coverage of now-deprecated migrations when we remove the v1 database integration.

package.json
```diff
@@ -106,7 +105,8 @@
       "index.ts",
       ".+\\/__tests__\\/.+\\.builder.(t|j)s",
       ".+\\/__tests__\\/.+\\.factory.(t|j)s",
-      ".e2e-spec.ts"
+      ".e2e-spec.ts",
+      "dist"
```
iamacook, line 109: Why should we include `dist` in the coverage?

src/app.module.ts
```diff
@@ -131,18 +135,22 @@ export class AppModule implements NestModule {
...
-          useFactory: async (configService: ConfigService) => {
-            const typeormConfig = await configService.getOrThrow('typeorm');
+          useFactory: async (
+            configService: ConfigService,
+            loggingService: ILoggingService,
+          ) => {
+            const typeormConfig = configService.getOrThrow('db.orm');
             const postgresConfigObject = postgresConfig(
-              await configService.getOrThrow('db.postgres'),
+              await configService.getOrThrow('db.connection.postgres'),
```
iamacook, line 144:
```suggestion
             configService.getOrThrow('db.connection.postgres'),
```

src/config/entities/__tests__/configuration.ts
```diff
@@ -23,7 +23,7 @@ export default (): ReturnType<typeof configuration> => ({
   },
   application: {
     isProduction: faker.datatype.boolean(),
-    runMigrations: true,
+    runMigrations: false,
```
iamacook, line 26: I think this will have side effects on targeted messages. cc @hectorgomezv

src/config/entities/__tests__/configuration.ts
```diff
@@ -84,20 +84,32 @@ export default (): ReturnType<typeof configuration> => ({
...
-      database: process.env.POSTGRES_TEST_DB || 'test-db',
-      username: process.env.POSTGRES_TEST_USER || 'postgres',
-      password: process.env.POSTGRES_TEST_PASSWORD || 'postgres',
-      ssl: {
-        enabled: true,
-        requestCert: true,
-        rejectUnauthorized: true,
-        caPath: process.env.POSTGRES_SSL_CA_PATH,
+    migrator: {
+      migrationsExecute: true,
```
iamacook, line 89: Nit: it's semantically clearer with the verb first:

```suggestion
      executeMigrations: true,
```

src/config/entities/__tests__/configuration.ts
```diff
@@ -84,20 +84,32 @@ export default (): ReturnType<typeof configuration> => ({
...
+      numberOfRetries: process.env.DB_TEST_MIGRATIONS_NUMBER_OF_RETRIES ?? 5,
+      retryAfter: process.env.DB_TEST_MIGRATIONS_RETRY_AFTER ?? 1000, // Milliseconds
+    },
+    orm: {
+      autoLoadEntities: true,
+      manualInitialization: true,
+      migrationsRun: false,
```
iamacook, line 96: How does this differ from `db.migrator.migrationsExecute`?

src/config/entities/__tests__/configuration.ts
```diff
@@ -84,20 +84,32 @@ export default (): ReturnType<typeof configuration> => ({
...
+      migrationsTableName: 'migrations_typeorm',
```
iamacook, line 97: Nit: what do you think about prefixing this with an underscore? It's highly unlikely we'll move to another ORM but I suggest we remove `_typeorm` to be more generic as we are with all datasources.

```suggestion
      migrationsTableName: '_migrations_typeorm',
```

src/datasources/db/v2/database-initialize.hook.ts
```diff
@@ -0,0 +1,16 @@
+import { PostgresDatabaseService } from '@/datasources/db/v2/postgres-database.service';
+import { Inject, Injectable, type OnModuleInit } from '@nestjs/common';
+
+@Injectable()
+export class DatabaseInitializeHook implements OnModuleInit {
+  public constructor(
+    @Inject(PostgresDatabaseService)
+    private readonly postgresDatabaseService: PostgresDatabaseService,
+  ) {
+    //
```
iamacook, line 10:
```suggestion
```

src/datasources/db/v2/database-initialize.hook.ts
```diff
@@ -0,0 +1,16 @@
...
+  }
+
+  public async onModuleInit(): Promise<void> {
+    // await this.postgresDatabaseService.initializeDatabaseConnection();
```
iamacook, line 14: Is there a reason this is commented out? If so, I'd add a TODO.

src/datasources/db/v2/database-migrator.service.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+  beforeEach(() => {});
+
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  describe('migrate()', () => {
+    it('Should log the start and end of the migration process', async () => {
+      jest.spyOn(loggingService, 'info');
```
iamacook, line 84: I am impartial regarding this testing method, but we'd generally use a mock logging service consisting of jest mocks. That way we don't need to `spyOn` the service.

src/datasources/db/v2/database-migrator.service.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      expect(loggingService.info).toHaveBeenCalledWith(
+        'Migrations: Running...',
+      );
+      expect(loggingService.info).toHaveBeenCalledWith('Migrations: Finished.');
+      expect(loggingService.info).toHaveBeenCalledTimes(3);
+    });
+
+    it('Should run migration if no lock exists', async () => {
+      jest.spyOn(connection, 'query').mockImplementation(jest.fn());
```
iamacook, line 96: `TestPostgresDatabaseModuleV2` injects a `Datasource` with mocked methods. We therefore don't need to `spyOn` this. You might need to adjust the type of `connection` to be `jest.MockedObjectDeep<DataSource>`.

src/datasources/db/v2/database-migrator.service.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+      await databaseMigratorService.migrate();
+
+      expect(connection.runMigrations).toHaveBeenCalled();
+      expect(connection.query).toHaveBeenCalledWith(insertLockQuery, [1]);
+
+      expect(connection.query).toHaveBeenCalledWith(truncateLockQuery);
+    });
+
+    it('Should throw an error if retries are exhausted', async () => {
+      connection.query = jest.fn().mockResolvedValue([{ id: 1, status: 1 }]);
```
iamacook, line 107: Similar to the above, the methods are mocked already. This means we can use `mockResolvedValue` directly on the property. You might need to adjust the type as above.

src/datasources/db/v2/database-migrator.service.spec.ts
```diff
@@ -0,0 +1,140 @@
...
+
+      await expect(databaseMigratorService.migrate()).rejects.toThrow(
+        'Migrations: Migrations are still running in another instance!',
+      );
+
+      expect(loggingService.info).toHaveBeenCalledWith(
+        'Migrations: Running in another instance...',
+      );
+      expect(loggingService.info).toHaveBeenCalledTimes(2);
```
iamacook, line 116: You're only checking the call value once here, but expecting it to be called twice. We should expect both and use `toHaveBeenNthCalledWith`.

The same can be done in a test below - line 134.

src/datasources/db/v2/database-migrator.service.ts
```diff
@@ -0,0 +1,160 @@
...
+          throw new Error(
+            'Migrations: Migrations are still running in another instance!',
+          );
+        }
+        this.loggingService.info('Migrations: Running in another instance...');
+        const retryAfter = this.configService.getOrThrow<number>(
+          'db.migrator.retryAfter',
+        );
+
+        await new Promise((resolve) => setTimeout(resolve, retryAfter));
```
iamacook, line 74: Do we also need to cleanup this timeout if the app closes? It might cause open handles in our tests?

src/datasources/db/v2/database-migrator.service.ts
```diff
@@ -0,0 +1,160 @@
...
+   *
+   * @param {DataSource} connection The database connection to create the locks table on.
+   *
+   * @returns {Promise<void>} A promise that resolves when the locks table has been created or confirmed to exist.
+   */
+  private async createLockTableIfNotExists(
+    connection: DataSource,
+  ): Promise<void> {
+    await connection.query(
+      'CREATE TABLE IF NOT EXISTS "_lock" ("id" SERIAL NOT NULL, "status" int NOT NULL, PRIMARY KEY ("id"))',
```
iamacook, line 90: I'd suggest moving the table name to a constant, and using it in all methods here. What do you think?

src/datasources/db/v2/postgres-database.module.ts
```diff
@@ -1,8 +1,20 @@
...
-  exports: [PostgresDatabaseService],
+  imports: [ConfigModule],
+  providers: [
+    PostgresDatabaseService,
+    DatabaseMigrator,
+    DatabaseInitializeHook,
+    DatabaseShutdownHook,
+    DatabaseMigrationHook,
+  ],
+  exports: [PostgresDatabaseService, DatabaseMigrator],
```
iamacook, line 18: Can be done in a followup, but if we're adding a suffix to the module, perhaps we should also to all providers as well?

src/datasources/db/v2/postgres-database.service.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    postgresqlService = moduleRef.get<PostgresDatabaseService>(
+      PostgresDatabaseService,
+    );
+  });
+
+  afterAll(async () => {
+    await postgresqlService.destroyDatabaseConnection();
+    await moduleRef.close();
+  });
```
iamacook, line 78: Nit: similar to your comment in the other PR, I would `beforeEach`/`afterEach` this. The tests are then decoupled from each other and we can be sure that the database is always initialised at the start of a test.

For example, when you test the `isInitialised` method, you are reinitialising the database as it is required in the subsequent test.

src/datasources/db/v2/postgres-database.service.spec.ts
```diff
@@ -0,0 +1,169 @@
...
+      jest
+        .spyOn(postgresqlService, 'getRepository')
+        .mockReturnValue(mockRepository);
+
+      const result =
+        await postgresqlService.getRepository<MockEntity>(MockEntity);
+
+      expect(postgresqlService.getRepository).toHaveBeenCalledWith(MockEntity);
+      expect(result).toBe(mockRepository);
+    });
```
iamacook, line 156: Is this test necessary? It's essentially testing `mockReturnValue`.

src/datasources/db/v2/postgres-database.service.ts
```diff
@@ -4,29 +4,70 @@ import {
...
+    return this.dataSource;
+  }
+
+  /**
+   * Checks whether the datasource has been initialized or not
+   *
+   * @returns {boolean} True if the datasource has already been initialized
+   */
+  public isInitialized(): boolean {
+    console.log(this.dataSource.isInitialized);
```
iamacook, line 39:
```suggestion
```

src/datasources/db/v2/postgresql-logger.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  type LoggerOptions,
+  type LogLevel,
+  type LogMessage,
+} from 'typeorm';
+
+@Injectable()
+export class PostgresqlLogger extends AbstractLogger {
+  public constructor(
+    public options?: LoggerOptions,
+    private readonly loggingService?: ILoggingService,
```
iamacook, line 14: We could inject the `loggingService` and not need pass it here.

src/datasources/db/v2/postgresql-logger.service.ts
```diff
@@ -0,0 +1,89 @@
...
+  ) {
+    super(options);
+  }
+  /**
+   * Write log to specific output.
+   *
+   * This is taken from the TypeOrm default implementation, check out {@link https://orkhan.gitbook.io/typeorm/docs/logging}
+   */
+  protected writeLog(
```
iamacook, line 23: This is a follow on to [@hectorgomezv's comment regarding direct usage of `console`](https://github.com/safe-global/safe-client-gateway/pull/1988/files#r1789965643). If we inject the `loggingService` as per my above comment, we could/should use it instead of `console.log`, for example.

PooyaRaki: The `loggingService` is injected and used in `logQuery` method. Since this was taken from the default implementation I didn't change it. Will use logger here as well.

src/config/entities/__tests__/configuration.ts
PooyaRaki: This is used internally by `TypeOrm`. I didn't find it necessary to add a comment about it but LMK if you want me to add a short comment in the cod explaining the difference

src/config/entities/__tests__/configuration.ts
PooyaRaki: Question: Do we have a naming convention? The former emphasizes the action of migration.
I'll fix it soon.

src/config/entities/__tests__/configuration.ts
PooyaRaki: Well the idea was to make it clear who owns the table. Will change the name.

src/datasources/db/v2/database-initialize.hook.ts
PooyaRaki: I've commented this out because the getRepository method in postgres-database.service initializes the connection if it hasn't been initialized yet. This ensures the connection is only initialized once when a call to the database is made.

Do you think we should uncomment it or remove the hook? Uncommenting won’t reinitialize the connection, but it does create a redundant call.

src/datasources/db/v2/database-initialize.hook.ts
PooyaRaki: I've commented this out because the getRepository method in postgres-database.service initializes the connection if it hasn't been initialized yet. This ensures the connection is only initialized once when a call to the database is made.

Do you think we should uncomment it or remove the hook? Uncommenting won’t reinitialize the connection, but it does create a redundant call.

src/datasources/db/v2/postgres-database.module.ts
PooyaRaki: I generalized a bit here. The goal was to create a migrator capable of migrating different databases, but the `PostgresDatabaseService` only handles `PostgreSQL`. For example, we can have:

`DatabaseMigrator`
`PostgresDatabaseService`
`MongoDatabaseService`

src/datasources/db/v2/postgresql-logger.service.ts
iamacook: Not sure if I follow but `postgresConfig` has access to the `LoggingService` in the [relative factory](https://github.com/safe-global/safe-client-gateway/blob/ormMigrationAndLogger/src/app.module.ts#L140), and is passed there.

src/datasources/db/v2/postgres-database.module.ts
iamacook: It was more in reference to the -v2 suffix. It could be later removed. If you think this won't be touched often then we can keep it as is.

src/datasources/db/v2/database-initialize.hook.ts
iamacook: In general, I'd expect to see initialisation here if `onModuleInit` is defined. However, I don't think we should keep redundant code. What do you think?

src/config/entities/__tests__/configuration.ts
iamacook: We can keep the name if you think it's better - it was a suggestion based on the thought that there _may_ be migration in the future.

src/config/entities/__tests__/configuration.ts
iamacook: As far as I understand, this achieves the same thing? If it does, I'd remove one of the two. What do you think?

src/config/entities/__tests__/configuration.ts
iamacook: We don't - it was more grammatical. If you prefer it as is, it doesn't need changing.

package.json
PooyaRaki: `dist ` is now included in `coveragePathIgnorePatterns`

PooyaRaki:
> Yes, this change is due to the fact the database is requested by the controller tests when the v1 database module is attached to the AppModule.
Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

> I'd love some documentation on these properties.
Since these are TypeOrm parameters I didn't add comments. Do you want me to add TypeOrm reference?

src/datasources/db/v2/postgresql-logger.service.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/postgres-database.service.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/1988/commits/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/config/entities/__tests__/configuration.ts
PooyaRaki: Sorry for the confusion earlier. No, it does not do the same thing. This is one of the reasons I suggest we avoid adding certain parameters to Config, as it can make things more confusing. The parameter used in orm pertains to TypeORM. If set to true, TypeORM will automatically run migrations when the module is registered, but it should always be set to false for our current use case. On the other hand, the parameter used by migrator determines whether we intend to run our custom migration logic, which includes locking the database.

src/config/entities/__tests__/configuration.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/app.module.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

package.json
PooyaRaki: The plan was to remove the integration, but that should be done in a separate PR, especially since other modules are now using it.
https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/postgresql-logger.service.ts
PooyaRaki: in `package.json` on line 26 we are passing `orm.js` directly to `TypeOrm`
`"typeorm": "node --env-file=.env -r tsconfig-paths/register ./node_modules/typeorm/cli -d ./dist/src/config/entities/orm.config.js",`
for that reason it's not possible.

src/datasources/db/v2/postgres-database.service.spec.ts
PooyaRaki: The purpose of this test is to ensure that for a given entity, we retrieve the correct repository. Do you think we should refactor or remove it?

src/datasources/db/v2/postgres-database.service.spec.ts
PooyaRaki: Although I believe we shouldn't create a module every time, I made this change for now to maintain consistency.
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/a29425089cb674ccc6dc44526ab72e538d4a7968

src/datasources/db/v2/postgres-database.module.ts
PooyaRaki: Oh, sorry, I read it as 'prefix'. Personally, I prefer not to add a suffix, as we typically don't register different versions of a module. However, I've added the suffix here until we deprecate v1. What do you think?

src/datasources/db/v2/database-migrator.service.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/database-migrator.service.spec.ts
PooyaRaki: In my opinion, a good test should have at most three assertions, but I made this change for the sake of consistency. We should discuss this further later.
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/database-migrator.service.spec.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/database-migrator.service.spec.ts
PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/database-migrator.service.spec.ts
PooyaRaki: Changed for the sake of consistency
https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/database-initialize.hook.ts
PooyaRaki: I added the call to `onModuleInit` because `getRepository` does not invoke initialize if the connection has already been established.
https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/datasources/db/v2/database-initialize.hook.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

src/config/entities/__tests__/configuration.ts
PooyaRaki: Changed the name
https://github.com/safe-global/safe-client-gateway/commit/2ea0e5d8a752fd31cff745ff3490d8fe8134ddc7

migrations/1727702843994-notification_types.ts
```diff
@@ -1,6 +1,7 @@
 import { MigrationInterface, QueryRunner } from 'typeorm';
 
-export class NotificationTypes1726754508107 implements MigrationInterface {
+export class NotificationTypes1727702843994 implements MigrationInterface {
+  name = 'NotificationTypes1727702843994';
```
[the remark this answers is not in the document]
PooyaRaki: I added it for testing purposes. Thank you for pointing that out; I’ve removed it.
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/ff1c5a4aad788e1f0929ba9f5653313d85e8d42c

src/datasources/db/v2/database-migrator.service.ts
iamacook: Resolving after discussion: as we are breaking/throwing, this shouldn't be an issue. The tests have reduced `retryAfter`/`numberOfRetries` values.

PooyaRaki: I think since we are breaking out of the loop either using `break` or `throw` that shouldn't be an issue. For tests I reduced the retryAfter and numberOfRetries .

migrations/1727451367471-notifications_enum.ts
```diff
@@ -0,0 +1,59 @@
...
+      `ALTER TABLE "notification_types" DROP COLUMN "name"`,
+    );
+    await queryRunner.query(
+      `CREATE TYPE "public"."notification_types_name_enum" AS ENUM('CONFIRMATION_REQUEST', 'DELETED_MULTISIG_TRANSACTION', 'EXECUTED_MULTISIG_TRANSACTION', 'INCOMING_ETHER', 'INCOMING_TOKEN', 'MODULE_TRANSACTION', 'MESSAGE_CONFIRMATION_REQUEST')`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "notification_types" ADD "name" "public"."notification_types_name_enum" NOT NULL`,
+    );
+    await queryRunner.query(
+      `ALTER TABLE "notification_types" ADD CONSTRAINT "UQ_1d7eaa0dcf0fbfd0a8e6bdbc9c9" UNIQUE ("name")`,
```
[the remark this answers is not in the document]
PooyaRaki: Well, I Also like unique names. Yes, this is generated by the ORM. I suggest we change it as part of `Notification` PR

package.json
iamacook: Apologies, I misread it. Is this change needed in this PR?

src/app.module.ts
iamacook: It [doesn't seem to have changed](https://github.com/safe-global/safe-client-gateway/blob/ff1c5a4aad788e1f0929ba9f5653313d85e8d42c/src/app.module.ts#L144).

src/config/entities/__tests__/configuration.ts
iamacook: Thank you for the clarification. I now tend to agree re. package-specific logic in the configuration. I'm impartial as to _where_ it's located but I'd suggest adding some documentation as @hectorgomezv suggested.

src/datasources/db/v2/database-migrator.service.spec.ts
iamacook: If your are expecting n calls, I think it's beneficial to also expect _what_ those calls are. For example, the arguments may change later but the number of calls may not. There are other expectations in this suite that are the same as my intial comment. However, if you disagree, you can leave it as is.

src/datasources/db/v2/postgres-database.module.ts
iamacook: I don't _think_ we'll import these providers often but I think it'd be beneficial to add the suffix for now and, as you said, remove it when we deprecate v1.

src/datasources/db/v2/postgres-database.service.spec.ts
iamacook: It wasn't with regards to the module itself, but the database state - ensuring that there was no "bleed" between tests. Before, the initialisation state was dependant on previous tests. Now, tests always start with the database not initialised.

However, it was a nit. If you have a strong opinion, you can go with your preference.

src/datasources/db/v2/postgres-database.service.spec.ts
iamacook: If it's integral, we can keep it. On face value, it's a bit confusing as you mock a function and test that function directly. Could you add a comment?

src/datasources/db/v2/postgresql-logger.service.ts
iamacook: Thank you for the clarification. Do we need a `logger` when running the command? We have a bash-specific configuration without one. Not sure if the [other change](https://github.com/safe-global/safe-client-gateway/pull/1988/files#r1803242982) has any influence either?

If it makes things too complicated though, let's keep it as is.

src/datasources/db/v2/postgres-database.service.spec.ts
PooyaRaki: All good, Let's keep it in `before/afterEach` as you suggested for now.

package.json
PooyaRaki: Well, I added it because we had coverage errors. Not necessarily needed in this PR.

iamacook: Do you mean the errors we solved by tweaking the coverage settings of the repo? If so (and it doesn't reduce the coverage), I think it best to revert this (nit).

src/datasources/db/v2/postgresql-logger.service.ts
PooyaRaki: We are not registering the logger when running the command.

migrations/1727451367471-notifications_enum.ts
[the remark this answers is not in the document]
hectorgomezv: Sounds good to me, thanks!

src/datasources/db/v2/postgresql-logger.service.ts
iamacook: I see. I think this is now [resolved](https://github.com/safe-global/safe-client-gateway/pull/1988/files#r1803242982) then. Apologies for any confusion.

src/config/entities/configuration.ts
```diff
@@ -45,7 +46,7 @@ export default () => ({
     isProduction: process.env.CGW_ENV === 'production',
     // Enables/disables the execution of migrations on startup.
     // Defaults to true.
-    runMigrations: process.env.RUN_MIGRATIONS?.toLowerCase() !== 'false',
+    runMigrations: process.env.RUN_MIGRATIONS?.toLowerCase() === 'false',
```
hectorgomezv, line 49: This is changing the original behavior (and making the comment incosistent), as the default value is no longer `true`.

It should be `!== false` instead of `=== false`, as it is on the `main` branch.

src/config/entities/postgres.config.ts
```diff
@@ -37,7 +40,7 @@ export const postgresConfig = (
     password: postgresEnvConfig.password,
     database: postgresEnvConfig.database,
     migrations: ['dist/migrations/*.js'],
-    // logging: !isCIContext,
+    logger: logger ? new PostgresqlLogger(logger, undefined) : undefined,
```
hectorgomezv, line 43: Nit: the `undefined` second parameter is redundant, isn't it?

src/config/entities/configuration.ts
```diff
@@ -131,26 +132,40 @@ export default () => ({
...
-        // https://nodejs.org/docs/latest-v20.x/api/tls.html#tlscreateserveroptions-secureconnectionlistener
-        rejectUnauthorized:
-          process.env.POSTGRES_SSL_REJECT_UNAUTHORIZED?.toLowerCase() !==
-          'false',
-        caPath: process.env.POSTGRES_SSL_CA_PATH,
+    migrator: {
+      executeMigrations:
+        process.env.DB_MIGRATIONS_EXECUTE?.toLowerCase() !== 'false',
+      numberOfRetries: process.env.DB_MIGRATIONS_NUMBER_OF_RETRIES ?? 5,
+      retryAfter: process.env.DB_MIGRATIONS_RETRY_AFTER ?? 1000, // Milliseconds
```
hectorgomezv, line 140: Nit: besides the comment, could we rename this to `retryAfterMs/DB_MIGRATIONS_RETRY_AFTER_MS`, I think it's good to have the scale in the variable name as we do for other configuration properties. (A person changing this in the k8s/container configuration might not have access to this source code to read the comment and know the scale for some reason)

src/datasources/db/v2/database-migrator.service.ts
hectorgomezv: Naive question maybe: why do we need a retry mechanism? If the lock is taken, couldn't the instance just log that the lock is taken and ignore the migrations?

src/config/entities/__tests__/configuration.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/cfb1f8c009fd9708aef92ed5c597c8170665750f

src/datasources/db/v2/database-initialize.hook.ts
PooyaRaki: Sorry for the confusion, somehow the change was reverted on my end. Done
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/e907e3cc5db5e7b605fce8afb51ea6464a464252

src/app.module.ts
PooyaRaki: Sorry for the confusion. Somehow the change was reverted on my end. Done
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/e907e3cc5db5e7b605fce8afb51ea6464a464252

src/datasources/db/v2/postgres-database.module.ts
PooyaRaki: I suggest we don't add suffix as the module is the face of this component and needs a suffix but the providers are all internally registered so for that reason I believe we don't need suffix. What do you think?

src/config/entities/configuration.ts
hectorgomezv: I saw you already changed this, thanks!

src/datasources/db/v2/postgres-database.module.ts
iamacook: Sounds good to me.

src/config/entities/configuration.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/378fdd1d0c485f4b3dbdba8a8ed0efc33d8a2121

src/config/entities/postgres.config.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/378fdd1d0c485f4b3dbdba8a8ed0efc33d8a2121

src/datasources/db/v2/postgres-database.service.spec.ts
PooyaRaki: I'll proceed with deleting the test if that works for you. @iamacook @hectorgomezv

package.json
PooyaRaki: Removed
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/905b0aa9113a7c34e7325e3b6e91db7c500bf663

src/config/entities/configuration.ts
PooyaRaki: Was about to write. This was a mistake. Fixed here:
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/378fdd1d0c485f4b3dbdba8a8ed0efc33d8a2121

src/datasources/db/v2/postgres-database.service.spec.ts
PooyaRaki: Removed
https://github.com/safe-global/safe-client-gateway/pull/1988/commits/30668354922839bc2e4ecb06d5f93bbd0ce92a15

iamacook (approved):

hectorgomezv (approved):

PR #1989, opened 2024-10-02
Add Targeted Messaging API

hectorgomezv:
## Summary
This PR adds the Targeted Messaging feature's API endpoints and domain logic.

#### GET submission
```
GET /v1/targeted-messaging/outreaches/{outreachId}/chains/{chainId}/safes/{safeAddress}/signers/{signerAddress}/submissions
```
Retrieves a given submission by Outreach, Chain, Safe, and Safe signer. 
- `204 No Content`: The Safe signer is **NOT** targeted for the outreach.
- `400 Bad Request`: The signer address is **NOT** a signer of the Safe.
- `200 OK`:  The Safe signer is targeted. If the submission did happen, `completionDate` would hold the completion date. Otherwise, it would be `null`.

#### CREATE submission
```
POST /v1/targeted-messaging/outreaches/{outreachId}/chains/{chainId}/safes/{safeAddress}/signers/{signerAddress}/submissions
```
- `201 Created`: If the submission was successfully recorded.
- `400 Bad Request`: The payload is invalid or if the Safe signer has already made the submission.
- `400 Bad Request`: The signer address is **NOT** a signer of the Safe.
- `404 Not Found`: The Safe signer is **NOT** targeted for the outreach.

#### Submission schema
```typescript
{
  "outreachId": number,
  "targetedSafeId": number,
  "signerAddress": `0x${string}`,
  "completionDate": Date | null
}
```

## Changes
- Adds GET and POST endpoints to `/targeted-messaging/outreaches/` to create and retrieve submissions.


src/domain/targeted-messaging/targeted-messaging.repository.ts
```diff
@@ -0,0 +1,48 @@
...
+      throw new BadRequestException('The signer is not a Safe owner');
+    }
+    return this.datasource.getSubmission(args);
+  }
+
+  async createSubmission(args: {
+    targetedSafe: TargetedSafe;
+    signerAddress: `0x${string}`;
+  }): Promise<Submission> {
+    return this.datasource.createSubmission(args);
```
iamacook, line 46: I think we should also add an owner check here.

src/routes/targeted-messaging/targeted-messaging.service.ts
```diff
@@ -0,0 +1,79 @@
...
+    const targetedSafe = await this.repository.getTargetedSafe(args);
+    try {
+      const submission = await this.repository.getSubmission({
+        chainId: args.chainId,
+        targetedSafe,
+        signerAddress: args.signerAddress,
+      });
+      return new Submission(
+        submission.outreachId,
+        submission.targetedSafeId,
```
iamacook, line 29: This continues from my `outreachId` comment, the `targetedSafeId` is effectively the Safe address. Do we want to/need to expose this?

This can be tackled in a separate PR if we think necessary.

src/routes/targeted-messaging/targeted-messaging.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+  ParseIntPipe,
+  Post,
+  Res,
+} from '@nestjs/common';
+import { ApiCreatedResponse, ApiOkResponse, ApiTags } from '@nestjs/swagger';
+import { Response } from 'express';
+
+@ApiTags('targeted-messaging')
+@Controller({
+  path: 'targeted-messaging',
```
iamacook, line 28: Do you think we'd have any other form of "messaging" later? We could adjust this to be "/messaging/targets", or drop the "targeted" part.

src/routes/targeted-messaging/targeted-messaging.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+  version: '1',
+})
+export class TargetedMessagingController {
+  constructor(private readonly service: TargetedMessagingService) {}
+
+  @ApiOkResponse({ type: Submission })
+  @Get(
+    'outreaches/:outreachId/chains/:chainId/safes/:safeAddress/signers/:signerAddress/submissions',
```
iamacook, line 36: Nit: as these are the same, we can move the prefix to the `path` property of the `Controller` decorator.

src/routes/targeted-messaging/targeted-messaging.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+  )
+  async getSubmission(
+    @Res() res: Response,
+    @Param('outreachId', new DefaultValuePipe(0), ParseIntPipe)
```
iamacook, line 40: Afaiu, the outreach needs to exist in the database so I don't think we should set a default value. It'll save the extra logic from being called.

src/routes/targeted-messaging/targeted-messaging.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+      throw err;
+    }
+  }
+
+  @ApiCreatedResponse({ type: Submission })
+  @Post(
+    'outreaches/:outreachId/chains/:chainId/safes/:safeAddress/signers/:signerAddress/submissions',
+  )
+  async createSubmission(
+    @Param('outreachId', new DefaultValuePipe(0), ParseIntPipe)
```
iamacook, line 75: As above regarding the default value.

hectorgomezv: Done in [3707402](https://github.com/safe-global/safe-client-gateway/pull/1989/commits/37074023fa76e41a4731987b0fe2b5ba31111eef)

Thanks!

src/routes/targeted-messaging/targeted-messaging.controller.ts
hectorgomezv: Done in [3707402](https://github.com/safe-global/safe-client-gateway/pull/1989/commits/37074023fa76e41a4731987b0fe2b5ba31111eef)

Thanks!

src/routes/targeted-messaging/targeted-messaging.controller.ts
hectorgomezv: Done in [3707402](https://github.com/safe-global/safe-client-gateway/pull/1989/commits/37074023fa76e41a4731987b0fe2b5ba31111eef)

Thanks!

src/routes/targeted-messaging/entities/create-submission.dto.entity.ts
```diff
@@ -0,0 +1,15 @@
+import { ApiProperty } from '@nestjs/swagger';
+import { z } from 'zod';
+
+export const CreateSubmissionDtoSchema = z.object({
+  completed: z.boolean().refine((val) => val === true, {
+    message: 'The value must be true',
+  }),
```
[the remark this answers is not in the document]
hectorgomezv: Added in [3707402](https://github.com/safe-global/safe-client-gateway/pull/1989/commits/37074023fa76e41a4731987b0fe2b5ba31111eef)

Thanks!

src/routes/common/interceptors/cache-control.interceptor.ts
```diff
@@ -19,7 +19,9 @@ export class CacheControlInterceptor implements NestInterceptor {
     return next.handle().pipe(
       tap(() => {
         const response: Response = context.switchToHttp().getResponse();
-        response.header('Cache-Control', 'no-cache');
+        if (!response.headersSent) {
+          response.header('Cache-Control', 'no-cache');
+        }
```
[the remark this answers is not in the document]
hectorgomezv: Done in [3707402](https://github.com/safe-global/safe-client-gateway/pull/1989/commits/37074023fa76e41a4731987b0fe2b5ba31111eef)

Thanks!

src/domain/targeted-messaging/targeted-messaging.repository.ts
```diff
@@ -0,0 +1,48 @@
...
+  }
```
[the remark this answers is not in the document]
hectorgomezv: These only get the submission metadata, not the actual data being submitted. So authentication is not a product requirement in this case.

src/routes/targeted-messaging/targeted-messaging.service.ts
```diff
@@ -0,0 +1,79 @@
...
+
+@Injectable()
+export class TargetedMessagingService {
+  constructor(
+    @Inject(ITargetedMessagingRepository)
+    private readonly repository: ITargetedMessagingRepository,
+  ) {}
+
+  async getSubmission(args: {
+    outreachId: number;
```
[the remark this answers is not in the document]
hectorgomezv: I think it's pretty common to expose entity `ids` at the service URLs. The only _risk_ I could foresee is we are exposing the _count_ of the entities, in the case the `id` is autoincremental. If this is a concern we could add a second `uuid` to each entity, but I think it's a bit overkill given the balance between effort and benefit. 

Regarding the different IDs between environments, I also think it's pretty common and should be expected, and `uuids` wouldn't fix this either. 

We have this scenario in other parts of the system, Safe Apps for example, which have autoincremental ids, and they differ between envs.

Regarding enums, I'm not sure they would scale well for selecting items in a database, and also it's worth considering adding a new item would need a code change/release 🤔 

Wdyt?

src/domain/targeted-messaging/targeted-messaging.repository.ts
hectorgomezv: `getSubmission` is always called before trying to create the submission. There's a test that assures a non-owner cannot create submissions (https://github.com/safe-global/safe-client-gateway/blob/b7acfdbe9a90b04da2795c760caf5d280f13de5c/src/routes/targeted-messaging/targeted-messaging.controller.spec.ts#L395-L396), so if the `getSubmission` code is changed and it stops checking the ownership, the test would fail. 

Happy to change this to explicitly check it on creation though if you prefer, but we would be executing the datasource call twice then. Wdyt?

src/routes/targeted-messaging/targeted-messaging.service.ts
hectorgomezv: I've answered in the other comment, thanks for bringing this up!

src/routes/targeted-messaging/targeted-messaging.controller.ts
hectorgomezv: I don't think so tbh. But it depends on product decisions. As this path was done in alignment with product/frontend, I'd let it as it is for now. We could migrate it later if needed though, wdyt?

src/routes/targeted-messaging/targeted-messaging.service.ts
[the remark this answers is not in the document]
iamacook: Thank you for the explanation. I can't see any issue exposing the ID so I think it's fine.

src/domain/targeted-messaging/targeted-messaging.repository.ts
iamacook: Considering we have the test in place, I think that's sufficient.

src/routes/targeted-messaging/targeted-messaging.controller.ts
iamacook: As it's confirmed, we can leave it as is.

iamacook (approved):

PR #1990, opened 2024-10-04
Add task scheduling

hectorgomezv:
## Summary
This PR introduces `@nestjs/schedule` as a scheduled tasks manager utility.

## Changes
- Adds `@nestjs/schedule`.
- Uses the new tool to execute the unsupported chain events logging in time intervals on `EventCacheHelper`.


src/domain/hooks/helpers/event-cache.helper.ts
```diff
@@ -163,6 +147,29 @@ export class EventCacheHelper implements OnModuleInit, OnModuleDestroy {
...
+          this.loggingService.warn({
+            type: EventCacheHelper.UNSUPPORTED_CHAIN_EVENT,
+            chainId,
+            count,
+          });
+          await this.cacheService.deleteByKey(cacheKey);
+        }
+      }),
+    );
+    this.unsupportedChains = [];
```
hectorgomezv, line 170: There are no changes to the function body, I've changed its position in the file to reflect this is `public` now (following the convention of having public functions before the private ones).

src/app.module.ts
```diff
@@ -49,6 +49,7 @@ import { TransactionsViewControllerModule } from '@/routes/transactions/transact
 import { DelegatesV2Module } from '@/routes/delegates/v2/delegates.v2.module';
 import { AccountsModule } from '@/routes/accounts/accounts.module';
 import { NotificationsModuleV2 } from '@/routes/notifications/v2/notifications.module';
+import { ScheduleModule } from '@nestjs/schedule';
```
iamacook, line 52: I am not opinionated regarding this, but this will be the first datasource we aren't wrapping. What do you think re. this?

src/domain/hooks/helpers/event-cache.helper.ts
```diff
@@ -163,6 +147,29 @@ export class EventCacheHelper implements OnModuleInit, OnModuleDestroy {
     await this.cacheService.increment(cacheKey, MAX_TTL);
   }
 
+  /**
+   * Logs the number of unsupported chain events for each chain and clears the store.
+   * This function is public just for testing purposes.
+   */
+  @Interval(EventCacheHelper.UNSUPPORTED_EVENTS_LOG_INTERVAL)
```
iamacook, line 154: What do you think about importing [the readable value](https://github.com/nestjs/schedule/blob/master/lib/enums/cron-expression.enum.ts#L6) and setting it as the constant? (I didn't know the package accepted milliseconds.)

src/app.module.ts
hectorgomezv: I'd be in favor of not wrapping it, it would be a bit overkill to wrap the annotations part and I don't see a lot of benefit in doing it since this is a NestJS official package. And it seems doubtful to me we replace it, unless we shift the approach completely to something else, like queues, in which case I'd say the API would be different. Wdyt?

src/domain/hooks/helpers/event-cache.helper.ts
hectorgomezv: I've replaced the annotation to `@Cron(CronExpression.EVERY_MINUTE)` I think it's more readable and meaningful as you suggested. Thanks!

[5d51919](https://github.com/safe-global/safe-client-gateway/pull/1990/commits/5d519192550a0547ff1e9f01ed2f460744e0c6a0)

src/app.module.ts
iamacook: Sounds good to me.

Unrelated but I think we should reconsider this with the config. module as well.

iamacook (approved):

iamacook (approved):

PR #2002, opened 2024-10-09
Iterate all chains when fetching all Safes

### LINKED ISSUE #2001: Iterate over chains when getting all Safes by owner
### Description

When calling `getAllSafesByOwner`, we expect the response from the Config Service to [not be longer than one page](https://github.com/safe-global/safe-client-gateway/blob/main/src/domain/safe/safe.repository.ts#L431-L433). As a temporary fix (as this also affected the clients), we [increased the page size o `Chains`](https://github.com/safe-global/safe-config-service/pull/1253). This is, however, not the correct way to paginated. We should, instead iterate of the pages if a `next` value is present.

### Requirements

1. Call `getChains` and determine if there is a `next` value, calling that thereafter and concatenating the `results`.
2. Else return concatenated `results`.

### Additional information

We should consider whether we do this on the chains repository level in case `getChains` is called anywhere without a limit, etc.

After this we should consider reverting the page size of chains (if we have that many configurations saved).

iamacook:
## Summary

Resolves #2001 

When getting all Safes by owner address, we fetch the chain configurations. We don't, however, take into account pagination.

This adds a new `IChainsRepository['getAllChains']` method that fetches subsequent `next` pages and concatenates `results`, called in the `getAllSafesByOwner` method.

## Changes

- Add `IChainsRepository['getAllChains']` implementation
- Call `getAllChains` instead of `getChains` in `getAllSafesByOwner`

### DIFF
--- a/src/domain/chains/chains.repository.spec.ts
+++ b/src/domain/chains/chains.repository.spec.ts
@@ -32,16 +32,17 @@ const mockConfigurationService = jest.mocked({
  * tests are required to cover all edge cases.
  */
 describe('ChainsRepository', () => {
-  let target: ChainsRepository;
+  // According to the limits of the Config Service
+  // @see https://github.com/safe-global/safe-config-service/blob/main/src/chains/views.py#L14-L16
+  const OFFSET = 40;
 
-  const maxLimit = 40;
+  let target: ChainsRepository;
   const maxSequentialPages = 3;
 
   beforeEach(() => {
     jest.resetAllMocks();
 
     mockConfigurationService.getOrThrow.mockImplementation((key) => {
-      if (key === 'safeConfig.chains.maxLimit') return maxLimit;
       if (key === 'safeConfig.chains.maxSequentialPages')
         return maxSequentialPages;
     });
@@ -55,37 +56,46 @@ describe('ChainsRepository', () => {
   });
 
   it('should return all chains across pages below request limit', async () => {
-    const offset = 40;
     const url = faker.internet.url({ appendSlash: true });
     const chains = Array.from(
       {
-        length: maxLimit * (maxSequentialPages - 1), // One page less than request limit
+        length: ChainsRepository.MAX_LIMIT * (maxSequentialPages - 1), // One page less than request limit
       },
       (_, i) => chainBuilder().with('chainId', i.toString()).build(),
     );
-    const pages = chunk(chains, maxLimit).map((results, i, arr) => {
-      const pageOffset = (i + 1) * offset;
+    const pages = chunk(chains, ChainsRepository.MAX_LIMIT).map(
+      (results, i, arr) => {
+        const pageOffset = (i + 1) * OFFSET;
 
-      const previous = ((): string | null => {
-        if (i === 0) {
-          return null;
-        }
-        return limitAndOffsetUrlFactory(maxLimit, pageOffset, url);
-      })();
-      const next = ((): string | null => {
-        if (i === arr.length - 1) {
-          return null;
-        }
-        return limitAndOffsetUrlFactory(maxLimit, pageOffset, url);
-      })();
+        const previous = ((): string | null => {
+          if (i === 0) {
+            return null;
+          }
+          return limitAndOffsetUrlFactory(
+            ChainsRepository.MAX_LIMIT,
+            pageOffset,
+            url,
+          );
+        })();
+        const next = ((): string | null => {
+          if (i === arr.length - 1) {
+            return null;
+          }
+          return limitAndOffsetUrlFactory(
+            ChainsRepository.MAX_LIMIT,
+            pageOffset,
+            url,
+          );
+        })();
 
-      return pageBuilder<Chain>()
-        .with('results', results)
-        .with('count', chains.length)
-        .with('previous', previous)
-        .with('next', next)
-        .build();
-    });
+        return pageBuilder<Chain>()
+          .with('results', results)
+          .with('count', chains.length)
+          .with('previous', previous)
+          .with('next', next)
+          .build();
+      },
+    );
     mockConfigApi.getChains.mockImplementation(
       ({ offset }): Promise<Page<Chain>> => {
         if (offset === 0) {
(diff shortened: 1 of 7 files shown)


src/domain/chains/chains.repository.ts
```diff
@@ -47,6 +48,28 @@ export class ChainsRepository implements IChainsRepository {
...
+      chains.push(...result.results);
+
+      if (!result.next) {
+        break;
+      }
+
+      const url = new URL(result.next);
+      const paginationData = PaginationData.fromLimitAndOffset(url);
+      offset ??= 0;
+      offset += paginationData.offset;
```
hectorgomezv, line 67: I think this is not correct. Considering for example 10 items per page (limit), it would work for the second page (offset was 0, + 10 = 10) and also for the third (previous offset is 10, + 10 = 20), but for the fourth it would take 20 as previous offset and the next URL would contain `?offset=30`, right? So it would set `50` as new offset. 

Maybe my math is wrong but I think we need to set the offset directly to the one extracted from the URL.

```typescript
offset = paginationData.offset;
```

src/domain/chains/chains.repository.ts
```diff
@@ -47,6 +48,28 @@ export class ChainsRepository implements IChainsRepository {
     return valid;
   }
 
+  async getAllChains(): Promise<Array<Chain>> {
+    const chains: Array<Chain> = [];
+
+    let offset: number | undefined;
+
+    while (true) {
```
hectorgomezv, line 56: I think we should be a bit more defensive here: `while(true)` seems scary, and we are relying on the fact the datasource (Config Service in this case) works as expected and does not have a bug affecting the `offset` value. 

What do you think about having some kind of threshold to limit the max number of iterations in the loop? We could define a max number of Chain objects that can be handled in total and divide it by the default `limit` path param.

src/domain/chains/chains.repository.ts
iamacook: Good point. The `offset` is indeed "fixed". I change it in 16fcf76322f95771485d14cb8100fec7de82649b.

src/domain/chains/chains.repository.ts
iamacook: I've set an "upper request limit" (adjustable by env. var.) so that only n amount of subsequent requests are made. We also now fetch the upper limit to hopefully also prevent subsequent requests. As of 6b4f5f7a9a6b1484a514880a028c11374fdfb3a3.

src/domain/chains/chains.repository.ts
```diff
@@ -47,6 +48,27 @@ export class ChainsRepository implements IChainsRepository {
     return valid;
   }
 
+  async getAllChains(): Promise<Array<Chain>> {
```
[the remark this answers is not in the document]
iamacook: 1. I've set the upper limit to 40, customisable via an env. var as of 6b4f5f7a9a6b1484a514880a028c11374fdfb3a3.
2. The results of `getChains` is cached.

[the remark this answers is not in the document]
hectorgomezv: @PooyaRaki just for adding some context: all the requests going to external datasources that are sent via `CacheFirstDatasource` are cached by default.

src/domain/chains/chains.repository.ts
```diff
@@ -16,15 +16,29 @@ import {
 } from '@/domain/indexing/entities/indexing-status.entity';
 import { ILoggingService, LoggingService } from '@/logging/logging.interface';
 import { differenceBy } from 'lodash';
+import { PaginationData } from '@/routes/common/pagination/pagination.data';
+import { IConfigurationService } from '@/config/configuration.service.interface';
 
 @Injectable()
 export class ChainsRepository implements IChainsRepository {
+  private readonly maxLimit: number;
```
hectorgomezv, line 24: Nit: Why is this named `max`? I mean, it seems the default limit we use for all the requests to get the chains. Maybe I'm wrong but I'd expect this to be something like a threshold but it is like a constant. Is there any reason to not use `PaginationData.DEFAULT_LIMIT` for this?

hectorgomezv (approved):

src/domain/chains/chains.repository.ts
iamacook: It mirrors the [`max_limit` of the `Chains` pagination from the Config. Service](https://github.com/safe-global/safe-config-service/blob/main/src/chains/views.py#L16).

I did consider changing the `PaginationData.DEFAULT_LIMIT` instead but that affects all pagination requests across the project. If we did, it would reduce the request by the Web though as they don't provide a `limit` in their request.

What do you think?

hectorgomezv: I meant doing something like:
```typescript
const result = await this.getChains(PaginationData.DEFAULT_LIMIT, offset);
```

I didn't mean to change the `PaginationData.DEFAULT_LIMIT`, that would indeed affect the overall service behavior. Sorry about the lack of clarity in my previous comment 😄

iamacook: The use of `40` was [at the request of @PooyaRaki](https://github.com/safe-global/safe-client-gateway/pull/2002#discussion_r1795022413).

@PooyaRaki, what do you think about this?

PooyaRaki: Changing the default limit could be risky, as it would impact all endpoints. However, since the web isn't specifying any limits, for this particular case, I think it's better to set a maximum limit of for example 40 specifically for this endpoint.

iamacook: Considering the above, I've moved the limit to a constant within the repository as I don't see it changing (for now). I left the sequential page requests in the configuration though.

What are your thoughts?

src/domain/chains/chains.repository.spec.ts
```diff
@@ -0,0 +1,302 @@
...
+          .with('next', next)
+          .build();
+      },
+    );
+    mockConfigApi.getChains.mockImplementation(
+      ({ offset }): Promise<Page<Chain>> => {
+        if (offset === 0) {
+          return Promise.resolve(pages[0]);
+        }
+        if (offset === 40) {
```
PooyaRaki, line 104: What about using:
`offset === offset + ChainsRepository.MAX_LIMIT`
`offset === offset + ChainsRepository.MAX_LIMIT * 2`
This way, if the max limit is changed in the future, the tests won't fail.

However, it might also be intentional for the tests to break if the limit changes, as it could signal that the change needs further attention.

src/domain/chains/chains.repository.spec.ts
```diff
@@ -0,0 +1,302 @@
...
+      chains.map((chain) => {
+        return {
+          ...chain,
+          ensRegistryAddress: getAddress(chain.ensRegistryAddress!),
+        };
+      }),
+    );
+    expect(mockConfigApi.getChains).toHaveBeenCalledTimes(2);
+    expect(mockConfigApi.getChains).toHaveBeenNthCalledWith(1, {
+      limit: 40,
```
PooyaRaki, line 126: Can we use `MAX_LIMIT` instead of 40?

src/domain/chains/chains.repository.spec.ts
iamacook: Changed to use constant in 3b4b46c25b4d282d29daa09f485730ab65e2dc65.

src/domain/chains/chains.repository.spec.ts
iamacook: Changed to use constant in 3b4b46c25b4d282d29daa09f485730ab65e2dc65.

PooyaRaki (approved):
LGTM 🚀

src/domain/chains/chains.repository.ts
hectorgomezv: Sounds good to me! 🚀

hectorgomezv (approved):
Great job!

PR #2007, opened 2024-10-10
Map `saltNonce` of creation transactions

iamacook:
## Summary

The `saltNonce` of creation transactions is now returned by the Transaction Service. This adds the appropriate logic for validating and mapping it across all layers accordingly.

## Changes

- Add property to domain schema, therefore extending type
- Add property to `CreationTransaction` entity
- Add property to `CreationTransactionInfo` entity
- Extend builder and associated test


iamacook:
I'm leaving this as a draft until the Transaction Service has been released to prod.

hectorgomezv (approved):

PR #2008, opened 2024-10-10
Fix Swagger definitions

iamacook:
## Summary

We are observing inconsistencies between our TypeScript types and Swagger schema in our automatically generated SDK. This fixes some of those encountered issues.

## Changes

- All nullable values are defined in `ApiOptionalProperty` decorators, flagged as `nullable` with the correct `type`
- All array-expected values leverage the `isArray` property instead of TypeScript annotations


hectorgomezv (approved):
👏🏻

PR #2010, opened 2024-10-11
Update Targeted Messaging Outreach

hectorgomezv:
## Summary
This PR adds `sourceId`, `type`, `teamName`, `sourceFile`, and `sourceFileProcesedDate` to the Targeted Messaging `Outreach` entity.

## Changes
- Adds a database migration to add the fields.
- Adjusts tests and `TargetedMessagingDatasource.createOutreach`.

### DIFF
--- a/src/datasources/targeted-messaging/entities/outreach.entity.ts
+++ b/src/datasources/targeted-messaging/entities/outreach.entity.ts
@@ -6,6 +6,8 @@ export class Outreach {
   source_id: number;
   type: string;
   team_name: string;
+  source_file: string;
+  source_file_processed_date: string;
   created_at: string;
   updated_at: string;
 
@@ -17,6 +19,8 @@ export class Outreach {
     source_id: number,
     type: string,
     team_name: string,
+    source_file: string,
+    source_file_processed_date: string,
     created_at: string,
     updated_at: string,
   ) {
@@ -27,6 +31,8 @@ export class Outreach {
     this.source_id = source_id;
     this.type = type;
     this.team_name = team_name;
+    this.source_file = source_file;
+    this.source_file_processed_date = source_file_processed_date;
     this.created_at = created_at;
     this.updated_at = updated_at;
   }
--- a/src/datasources/targeted-messaging/entities/outreach.entity.ts
+++ b/src/datasources/targeted-messaging/entities/outreach.entity.ts
@@ -3,6 +3,9 @@ export class Outreach {
   name: string;
   start_date: string;
   end_date: string;
+  source_id: number;
+  type: string;
+  team_name: string;
   created_at: string;
   updated_at: string;
 
@@ -11,6 +14,9 @@ export class Outreach {
     name: string,
     start_date: string,
     end_date: string,
+    source_id: number,
+    type: string,
+    team_name: string,
     created_at: string,
     updated_at: string,
   ) {
@@ -18,6 +24,9 @@ export class Outreach {
     this.name = name;
     this.start_date = start_date;
     this.end_date = end_date;
+    this.source_id = source_id;
+    this.type = type;
+    this.team_name = team_name;
     this.created_at = created_at;
     this.updated_at = updated_at;
   }
--- a/src/datasources/targeted-messaging/entities/outreach.entity.ts
+++ b/src/datasources/targeted-messaging/entities/outreach.entity.ts
@@ -1,28 +1,30 @@
 export class Outreach {
   id: number;
   name: string;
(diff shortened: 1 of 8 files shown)


src/datasources/targeted-messaging/entities/outreach.entity.ts
```diff
@@ -3,6 +3,11 @@ export class Outreach {
   name: string;
   start_date: string;
   end_date: string;
+  source_id: number;
+  type: string;
+  team_name: string;
+  source_file: string;
+  source_file_processed_date: string;
```
iamacook, line 10: Both `source_file` and `source_file_processed_date` can be `null`. I'm not 100% sure, but as it's a `TIMESTAMP`, shouldn't `source_file_processed_date` be a `Date` if defined instead of a `string`?

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
```diff
@@ -60,6 +69,13 @@ export class TargetedMessagingDatasource
       name: outreach.name,
       startDate: new Date(outreach.start_date),
       endDate: new Date(outreach.end_date),
+      sourceId: outreach.source_id,
+      type: outreach.type,
+      teamName: outreach.team_name,
+      sourceFile: outreach.source_file,
+      sourceFileProcessedDate: outreach.source_file_processed_date
+        ? new Date(outreach.source_file_processed_date)
```
iamacook, line 77: Follow on from my type-related comment. Aren't these already `Date`s? (Would stand for `startDate`, `endDate`, `created_at` and `updated_at` as well.

src/domain/targeted-messaging/entities/create-outreach.dto.entity.ts
```diff
@@ -1,9 +1,19 @@
...
+    .int()
+    .gte(1)
+    .lte(DB_MAX_SAFE_INTEGER - 1),
+  type: z.string(),
+  teamName: z.string(),
+  sourceFile: z.string().nullish().default(null),
+  sourceFileProcessedDate: z.coerce.date().nullish().default(null),
 });
 
 export class CreateOutreachDto
```
iamacook, line 19:
```suggestion
export class CreateOutreachDto implements z.infer<CreateOutreachDtoSchema>
```

hectorgomezv: It would be `implements z.infer<typeof CreateOutreachDtoSchema>`, right? That's on the next line (20) (https://github.com/safe-global/safe-client-gateway/blob/eb7c106d7fd453178e976bf756c54c4580cb51e5/src/domain/targeted-messaging/entities/create-outreach.dto.entity.ts#L20)

Maybe I'm not understanding your comment, sorry.

src/datasources/targeted-messaging/entities/outreach.entity.ts
hectorgomezv: I initially thought we would need to thread these values as strings when dealing directly with the `postgres` library, but you are right, it seems we can use `Date` directly so need to parse them twice. I've changed it in [bfb327e](https://github.com/safe-global/safe-client-gateway/pull/2010/commits/bfb327e54f8e3c1713ed4922a6f61d1ed48151c2)

This can also be applied to the other DB-related entities like `DbTargetedSafe` or `DbSubmission`. Will change those in a separate PR as they are not related to this one.

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
hectorgomezv: Absolutely! Answered in https://github.com/safe-global/safe-client-gateway/pull/2010#discussion_r1803861284

hectorgomezv:
I've added `sourceFileChecksum` to the `Outreach` entity as the file integrity should be checked before parsing.

I've updated the PR description accordingly.

src/datasources/targeted-messaging/entities/outreach.entity.ts
iamacook: We should definitely check/adjust for the `Date` coercion accordingly when we move this to the ORM as that uses `pg` instead of `postgres`, cc @PooyaRaki

src/domain/targeted-messaging/entities/create-outreach.dto.entity.ts
iamacook: You're right. My preview window didn't show the next line 🥺

iamacook (approved):

PR #2012, opened 2024-10-14
Return indexing timestamp from Transaction Service

iamacook:
## Summary

We calculate the `lastSync` timestamp of the `/v1/chains/:chainId/about/indexing` endpoint based on the block timestamp. Currently, we request the block timestamp on chain.

The Transaction Service now returns the timestamp of the indexed block alongside the current, ERC-20 and singleton indexed block numbers. This removes reliance on the blockchain call, instead using the returned timestamp instead.

## Changes

- Add new properties to `IndexingStatusSchema`, thus inferring them
- Use timestamps from Transaction Service directly
- Extend builder
- Update test coverage appropriately


iamacook:
I'm marking this as a draft until the Transaction Service is released.

PooyaRaki (approved):

hectorgomezv (approved):

PR #2021, opened 2024-10-15
`close` app `afterEach` test

### LINKED ISSUE #2020: Close the app `afterEach` controller test
### Description

Every controller test initialises and closes the app. In general, `init` is called `beforeEach` and `close` `afterEach`. However, there are some instances whereby the app is closed `afterAll` instead. This means that the instance of all tests prior to the last one are not properly closed.

Every controller test should, therefore, `init` the app `beforeEach` and `close` it `afterEach`.

### Reproduction steps

Observe all instances of `afterAll` in controller tests. Multiple of which `close` the app instead.

### Expected behavior

`close` is called in `afterEach`.

### Additional information

_No response_

iamacook:
## Summary

Resolves #2020

In every controller test, we `init` the app `beforeEach` to have a "clean" instances for testing. Theoretically, we should `close` these before initialising a new one.

Many tests `close` the current instance only `afterAll` tests have run, calling `init` before every test. This moves all `close` calls to `afterEach`.

## Changes

- Move all calls to `close` the app from `afterAll` to `afterEach`

### DIFF
--- a/src/routes/accounts/accounts.controller.spec.ts
+++ b/src/routes/accounts/accounts.controller.spec.ts
@@ -89,12 +89,9 @@ describe('AccountsController', () => {
     jest.useFakeTimers();
   });
 
-  afterEach(() => {
-    jest.useRealTimers();
-  });
-
-  afterAll(async () => {
+  afterEach(async () => {
     await app.close();
+    jest.useRealTimers();
   });
 
   describe('AuthGuard', () => {
--- a/src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
+++ b/src/routes/accounts/counterfactual-safes/counterfactual-safes.controller.spec.ts
@@ -91,7 +91,7 @@ describe('CounterfactualSafesController', () => {
     jest.resetAllMocks();
   });
 
-  afterAll(async () => {
+  afterEach(async () => {
     await app.close();
   });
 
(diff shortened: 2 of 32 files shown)


PooyaRaki:
Question: Why do we need to create a new module `beforeEach`? What about creating a new module `beforeAll`?

iamacook:
> Question: Why do we need to create a new module `beforeEach`? What about creating a new module `beforeAll`?

I'd assume it's to ensure all mocks are cleared correctly/each test has a new app instance.

PooyaRaki:
> > Question: Why do we need to create a new module `beforeEach`? What about creating a new module `beforeAll`?
> 
> I'd assume it's to ensure all mocks are cleared correctly/each test has a new app instance.

In that case, I suggest using jest.clearAllMocks or jest.resetMocks, depending on the use case. My concern is that creating a module before each test can add unnecessary overhead and slow down the tests.

iamacook:
> In that case, I suggest using `jest.clearAllMocks` or `jest.resetMocks`, depending on the use case. My concern is that creating a module before each test can add unnecessary overhead and slow down the tests.

Due to the way in which we architected test modules (using `jest.fn`), the mocks do not successfully clear between each test by having one instance throughout (currently). It would require substantially more effort to achieve that. I do, however, think it's the better approach.

I'm going to close this for now and we can discuss if/how we want to proceed at a later stage.

PR #2024, opened 2024-10-15
Flag spoof events with varied value

### LINKED ISSUE #2023: Flag spoof events with a varied value
### Description

We flag imitation transactions (spoof events) that have the same value as the transaction being imitated. This value can, however, vary slightly.

We should support imitation flag with a varied value +/- a defined tolerance.

### Requirements

Adjust value comparison of imitation checks to allow for a defined tolerance difference.

### Additional information

_No response_

iamacook:
## Summary

Resolves #2023

We flag imitation transactions (spoof events) based on certain criteria, of which a matching value being one of them. This comparison is becoming redundant as the values are beginning to vary compared to the transaction being imitated.

This adjusts the current check, adding a new tolerance. A transaction is now marked as an imitation if the value is +/- value + tolerance of the transaction being imitated. 

## Changes

- Add new tolerance env. var. with default value.
- Include tolerance in value check for flagging spoof events.

### DIFF
--- a/src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
+++ b/src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
@@ -1,14 +1,24 @@
 import { IConfigurationService } from '@/config/configuration.service.interface';
+import { TransactionInfo } from '@/routes/transactions/entities/transaction-info.entity';
 import { TransactionItem } from '@/routes/transactions/entities/transaction-item.entity';
 import {
   isTransferTransactionInfo,
   TransferDirection,
   TransferTransactionInfo,
 } from '@/routes/transactions/entities/transfer-transaction-info.entity';
-import { isErc20Transfer } from '@/routes/transactions/entities/transfers/erc20-transfer.entity';
+import {
+  Erc20Transfer,
+  isErc20Transfer,
+} from '@/routes/transactions/entities/transfers/erc20-transfer.entity';
 import { isSwapTransferTransactionInfo } from '@/routes/transactions/swap-transfer-transaction-info.entity';
 import { Inject } from '@nestjs/common';
-import { formatUnits } from 'viem';
+
+type Erc20TransferTransactionInfo = Omit<
+  TransferTransactionInfo,
+  'transferInfo'
+> & {
+  transferInfo: Erc20Transfer;
+};
 
 export class TransferImitationMapper {
   private static ETH_DECIMALS = 18;
@@ -16,6 +26,7 @@ export class TransferImitationMapper {
   private readonly lookupDistance: number;
   private readonly prefixLength: number;
   private readonly suffixLength: number;
+  private readonly valueTolerance: bigint;
 
   constructor(
     @Inject(IConfigurationService)
@@ -29,6 +40,9 @@ export class TransferImitationMapper {
     );
     this.suffixLength = configurationService.getOrThrow(
       'mappings.imitation.suffixLength',
+    );
+    this.valueTolerance = configurationService.getOrThrow(
+      'mappings.imitation.valueTolerance',
     );
   }
 
@@ -65,13 +79,7 @@ export class TransferImitationMapper {
 
       const txInfo = item.transaction.txInfo;
       // Only transfers can be imitated, of which we are only interested in ERC20s
-      if (
-        !(
-          isTransferTransactionInfo(txInfo) ||
-          isSwapTransferTransactionInfo(txInfo)
-        ) ||
-        !isErc20Transfer(txInfo.transferInfo)
-      ) {
+      if (!this.isErc20TransferTransactionInfo(txInfo)) {
         mappedTransactions.unshift(item);
         continue;
       }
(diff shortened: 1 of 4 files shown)


src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
```diff
@@ -100,1393 +105,2528 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
...
-  function getImitationAddress(address: `0x${string}`): `0x${string}` {
-    // + 2 is to account for the '0x' prefix
-    const prefix = address.slice(0, prefixLength + 2);
-    const suffix = address.slice(-suffixLength);
-    const imitator = `${prefix}${faker.finance.ethereumAddress().slice(prefixLength + 2, -suffixLength)}${suffix}`;
-    return getAddress(imitator);
-  }
-  const chain = chainBuilder().build();
-  const safe = safeBuilder().build();
+  describe('Event spoofing', () => {
```
iamacook, line 108: I'd suggest hiding whitespace to review this. The test content has not drastically changed for the existing tests, but newer value-tolerance-focused tests have been added.

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
```diff
@@ -100,1393 +105,2528 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
...
             },
-            type: 'TRANSACTION',
-          },
-        ]);
-      });
+          ]);
+        });
+    });
+
+    it.todo('Incoming transfer "echo"');
```
iamacook, line 2630: This is in preparation for the aforementioned, forthcoming PR. Considering the length of the file, I may move it to another suite depending on the implementation.

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
```diff
@@ -100,1393 +105,2586 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
...
-  function getImitationAddress(address: `0x${string}`): `0x${string}` {
-    // + 2 is to account for the '0x' prefix
-    const prefix = address.slice(0, prefixLength + 2);
-    const suffix = address.slice(-suffixLength);
-    const imitator = `${prefix}${faker.finance.ethereumAddress().slice(prefixLength + 2, -suffixLength)}${suffix}`;
-    return getAddress(imitator);
-  }
-  const chain = chainBuilder().build();
-  const safe = safeBuilder().build();
+  describe('Event spoofing', () => {
```
iamacook, line 108: I'd suggest hiding whitespace to review this. The test content has not drastically changed for the existing tests, but newer value-tolerance-focused tests have been added.

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
```diff
@@ -100,1393 +105,2586 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
...
+    function parseUnits(value: bigint, decimals: number): bigint {
```
PooyaRaki, line 109: Question: Do we need to use `parseUnits` anywhere else in the code? If that’s the case, I’d suggest creating a helper function for it and importing it in the tests.

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
```diff
@@ -100,1393 +105,2586 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
...
-                  logoUri: null,
-                  name: null,
-                  value: imitationAddress,
+        await request(app.getHttpServer())
+          .get(
+            `/v1/chains/${chain.chainId}/safes/${safe.address}/transactions/history?trusted=false`,
+          )
+          .expect(200)
+          .then(({ body }) => {
+            expect(body.results).toStrictEqual([
```
PooyaRaki, line 1695: Since these expected values and test inputs are making the test file large and difficult to review, do you think it would be a good idea to move them to a separate file? For example:
`transactions-history.imitation-transactions.controller.spec.ts`
`transactions-history.imitation-transactions.mock.ts`

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
```diff
@@ -100,1393 +105,2586 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
...
+
+      it('should not filter out imitation incoming/outgoing transfers of vanity with an intolerant value within the lookup distance', async () => {
+        const results = [
+          valueIntolerantIncomingTransaction,
+          multisigTransaction,
+          valueIntolerantOutgoingTransaction,
+          notImitatedMultisigTransaction,
+          multisigTransaction,
+        ];
+        networkService.get.mockImplementation(({ url }) => {
```
PooyaRaki, line 1941: Nit: I have the same concern here. Could we move these input values elsewhere to keep the focus of the test on the scenarios we’re actually testing?

PooyaRaki (approved):
Approved with some suggestions.

PooyaRaki:
Question: Can we calculate the tolerance dynamically from the recent transactions of the sender?

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
iamacook: No, we use [`parseUnits` from viem](https://viem.sh/docs/utilities/parseUnits.html). This was an edge case where we needed a numerical implementation.

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
iamacook: What do you think about eventually splitting the suites per attack vector? Preferably this would be done after the follow up PR though.

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
iamacook: See https://github.com/safe-global/safe-client-gateway/pull/2024#discussion_r1810773000.

iamacook:
> Question: Can we calculate the tolerance dynamically from the recent transactions of the sender?

We only have the current page for comparison so it's not a sufficient sample size imo. If the page only had high value transactions, for example, there'd be a higher chance of false positives/

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
PooyaRaki: Sure, It makes sense to do it in a separate PR as a clean up

PooyaRaki:
> > Question: Can we calculate the tolerance dynamically from the recent transactions of the sender?
> 
> We only have the current page for comparison so it's not a sufficient sample size imo. If the page only had high value transactions, for example, there'd be a higher chance of false positives/

@iamacook Thanks for the explanation, I'll look into it more to see if there is anything else we can do to make it even stronger.

src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
```diff
@@ -65,13 +79,7 @@ export class TransferImitationMapper {
...
       const txInfo = item.transaction.txInfo;
       // Only transfers can be imitated, of which we are only interested in ERC20s
-      if (
-        !(
-          isTransferTransactionInfo(txInfo) ||
-          isSwapTransferTransactionInfo(txInfo)
-        ) ||
-        !isErc20Transfer(txInfo.transferInfo)
-      ) {
+      if (!this.isErc20TransferTransactionInfo(txInfo)) {
```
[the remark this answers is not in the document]
PooyaRaki: That makes more sense now. I’ll leave it up to you—I’ll resolve the conversation. If you feel the name is accurate, I’m good with the current one.

PR #2026, opened 2024-10-16
Flag "echo" transfers as imitations

### LINKED ISSUE #2025: Flag "echo" incoming transfers
### Description

We flag spoof events as imitation transactions. We don't, however, flag the other address poisoning vector whereby an attacker "echoes" an outgoing transfer with an incoming one of the same token.

We should flag the above transactions if they are under a certain limit.

### Requirements

Flag incoming transfers that following outgoing transfers of the same token address, under a defined limit.

### Additional information

_No response_

iamacook:
## Summary

Resolves #2025

An address poisoning vector we don't flag are incoming transfers of the same asset sent of minimal value, or "echoes". When sending a large amount of an asset, it is common to receive a low value of the same asset from a malicious actor in the hopes you copy their address and send to them instead.

This adds new logic to flag such transactions. If the receipt of a token occurs after it being sent, under a defined limit, it is deemed an "echo" and flagges as such.

## Changes

- Add `mappings.imitation.echoLimit` value
- Flag "echo" transactions

### DIFF
--- a/src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
+++ b/src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
@@ -60,6 +60,8 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
   const suffixLength = 4;
   const valueTolerance = BigInt(1);
   const echoLimit = BigInt(10);
+  const chain = chainBuilder().build();
+  const safe = safeBuilder().build();
 
   beforeEach(async () => {
     jest.resetAllMocks();
@@ -110,9 +112,6 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
   afterAll(async () => {
     await app.close();
   });
-
-  const chain = chainBuilder().build();
-  const safe = safeBuilder().build();
 
   describe('Event spoofing', () => {
     function parseUnits(value: bigint, decimals: number): bigint {


src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
```diff
@@ -113,7 +117,10 @@ export class TransferImitationMapper {
           return false;
         }
 
-        return this.isSpoofedEvent(txInfo, prevTxInfo);
+        return (
+          this.isSpoofedEvent(txInfo, prevTxInfo) ||
+          this.isEchoImitation(txInfo, prevTxInfo)
+        );
```
hectorgomezv, line 123: (This is more a general comment, no directly related to the PR code, and I think it's out of the scope) 

As this `TransferImitationMapper` gets more complex, it is likely getting more computationally expensive. I think it's fine for now, but given this mapper is applied to each group of transactions, and if I'm not wrong it's deterministic, it'd be great to somehow cache/memoize whether a given transaction is an imitation one. Wdyt?

hectorgomezv (approved):
👏🏻👏🏻👏🏻

src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
iamacook: I agree - we should look into caching the mapped entities (potentially across the project). I've noted it down for discussion.

src/routes/transactions/transactions-history.imitation-transactions.controller.spec.ts
```diff
@@ -105,6 +107,9 @@ describe('Transactions History Controller (Unit) - Imitation Transactions', () =
     await app.close();
   });
 
+  const chain = chainBuilder().build();
+  const safe = safeBuilder().build();
```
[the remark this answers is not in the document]
iamacook: Moved in f17faf5668b87975a98654441009908a61a0d205.

hectorgomezv (approved):

PR #2027, opened 2024-10-16
Fix DB initialization on E2E tests

hectorgomezv:
## Summary
This PR adds the SSL certificate path to the e2e tests. Otherwise, these tests would fail as they don't mock the database. As `PostgresDatabaseModule` is now part of the `AppModule`, the actual Postgres container needs to be available on the e2e test execution context.

## Changes
- Adds `POSTGRES_SSL_ENABLED` and `POSTGRES_SSL_CA_PATH` to the `e2e-setup.ts` configuration.

### DIFF
--- a/src/routes/about/__tests__/get-about.e2e-spec.ts
+++ b/src/routes/about/__tests__/get-about.e2e-spec.ts
@@ -1,8 +1,6 @@
 import '@/__tests__/matchers/to-be-string-or-null';
 import { AppModule } from '@/app.module';
 import { CacheKeyPrefix } from '@/datasources/cache/constants';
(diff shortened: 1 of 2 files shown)


src/routes/about/__tests__/get-about.e2e-spec.ts
```diff
@@ -17,8 +15,6 @@ describe('Get about e2e test', () => {
     const moduleRef = await Test.createTestingModule({
       imports: [AppModule.register()],
     })
-      .overrideModule(PostgresDatabaseModule)
```
hectorgomezv, line 20: This mocking was previously added to prevent this initialization error but it's not longer needed.

iamacook (approved):

PR #2034, opened 2024-10-18
Add first Targeted Messaging Outreach

hectorgomezv:
## Changes
- Adds `00008_targeted_messaging_add_outreach_1` database migration inserting an `Outreach` metadata.


migrations/00008_targeted_messaging_add_outreach_1/index.sql
```diff
@@ -0,0 +1,19 @@
...
+    team_name,
+    source_file,
+    source_file_checksum
+) VALUES (
+    1,
+    '',
+    '1970-01-01 00:00:00.000',
+    '1970-01-01 00:00:00.000',
+    '',
+    '',
```
hectorgomezv, line 16: These are just placeholders for non-nullable columns. The actual data will be loaded as soon as the JSON file is processed (at the service startup).

iamacook (approved):

PR #2057, opened 2024-10-25
Improve "echo" detection

iamacook:
## Summary

An "echo" poisoning attempt is when an attacker sends a small amount of a token back to you after sending some to a receipient. The token address is therefore the same, the value low and the recipient similar in vanity.

This fixes the token address comparison as we were previously comparing a token against itself, and it adds a vanity check.

## Changes

- Fix token address comparison
- Add vanity check
- Adjust tests accordingly

### DIFF
--- a/src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
+++ b/src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
@@ -167,9 +167,7 @@ export class TransferImitationMapper {
       return false;
     }
 
-    const refAddress = this.getReferenceAddress(txInfo);
-    const prevRefAddress = this.getReferenceAddress(prevTxInfo);
-    return this.isImitatorAddress(refAddress, prevRefAddress);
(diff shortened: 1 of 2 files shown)


src/routes/transactions/mappers/transfers/transfer-imitation.mapper.ts
```diff
@@ -198,10 +196,12 @@ export class TransferImitationMapper {
...
-    return (
-      // Imitations generally follow high transfer values
-      prevValue >= this.echoLimit && value <= this.echoLimit
-    );
+    const isEchoValue = prevValue >= this.echoLimit && value <= this.echoLimit;
+    if (!isEchoValue) {
+      return false;
+    }
+
+    return this.isImitationAddress(txInfo, prevTxInfo);
```
iamacook, line 204: Whilst this check occurs in both methods, I thought it clearer to include it in both.

hectorgomezv (approved):

PR #2061, opened 2024-10-25
Don't expose rich fragments

iamacook:
## Summary

Rich fragments (human-readable decoding) has an associated feature flag, and is not used by the clients. After confirmation that this can be done, this removes the expose of rich fragments and the associated feature flag.

## Changes

- Don't expose `richDecodedInfo`
- Remove feature flag
- Remove potential `undefined` type throughout project


hectorgomezv (approved):
👏🏻

PR #2080, opened 2024-10-30
Update Node to v22.11.0 (LTS)

iamacook:
## Summary

The latest LTS version of Node [has been released](https://nodejs.org/en/blog/release/v22.11.0) (v22.11.0). As such, this makes the relevant changes to use it in the project.

## Changes

- Use `22.11.0` in all `actions/setup-node`
- Update our `.nvmrc`
- Use "jod" in our Dockerfile
- Update our README


iamacook:
I'm marking this as a draft as the `actions/setup-node` `node-version` is [having issues with "Jod"](https://github.com/actions/setup-node/issues/1153).

iamacook:
> Now that we are upgrading the Node version, I suggest we tag the exact version instead of `lts`

Agreed. As this is currently blocked due to the [workflow compatibility](https://github.com/safe-global/safe-client-gateway/pull/2080#issuecomment-2447013106), I'd suggest pinning the version once unblocked. That way we will have the latest tag pinned.

iamacook:
@PooyaRaki, I've now pinned the version.

hectorgomezv (approved):

PR #2085, opened 2024-11-04
Map `SafeApp['featured']` flag

iamacook:
## Summary

A featured` flag has been [added to the `SafeApp` model](https://github.com/safe-global/safe-config-service/pull/1272) of the Config Service.

This maps the new `featured` flag to the appropriate `SafeApp` entity, under the `featured` property.

## Changes

- Add `featured` to the `SafeAppsSchema`
- Map flag to `SafeApp['featured']`
- Update tests accordingly


hectorgomezv (approved):
LGTM! We have to recall the Config Service needs to be released/deployed before the CGW, otherwise this would make the Safe Apps retrieval fail.

Issue #1731, opened 2024-07-05
Only strictly type fetched data after validation


iamacook:
### Description

We currently fetch data and explicitly type the response, validating only on the domain layer. Whilst the layer separation makes sense, we can't be 100% sure that the type matches the response until _after_ validation. Ideally, we should assume the response is `unknown` until validated.

We should either adjust the types of requests, or validate the response immediately.

### Requirements

There are a couple ways to approach this, each adjusting the architecture of the project:

1. Change reponse types to `unknown` (or `T | unknown`).

The downside of the above is that the types are "lost" and it is hard to then find the corresponding type. A better suggestion would setting all requests as `T | unknown`. This is then a) easier as the generic of the service can be changed and not that of every request and b) we always have the relevant type "in place". Validating the response of this will then "remove" then `unknown` from the union.

2. Immediately validate responses.

We could "hoist" the validation to the datasource layer. If we want to remain "strict", this does not fit our mental model. However, it means that every call returns the expected type but we generally inject the respective repository though. One thing to note is that we _are_ doing something similar with the "incoming data" of DTOs.

PR #2098, opened 2024-11-06
Type `ITransactionApi` responses at "raw"

iamacook:
Partial implementation of #1731

## Summary

We validate API responses on the domain layer, but directly assign the type in the datasources. This means that the response types are not necessarily correct.

This types all the `ITransactionApi` responses as "raw". These responses can therefore not be used directly unless they are validated.

## Changes

- Add `Raw` utility type to `ITransactionApi`
- Include validation where appropriate


hectorgomezv (approved):
👏🏻

PR #2100, opened 2024-11-07
Type `IPricesApi`/`IBalancesApi` responses as "raw"

iamacook:
Partial implementation of #1731

## Summary

We validate API responses on the domain layer, but directly assign the type in the datasources. This means that the response types are not necessarily correct.

This types all request responses of `IPricesApi`/`IBalancesApi` as "raw". These responses are therefore not be used directly unless they are validated.

Not: these are combined as they are heavily interlinked.

## Changes

- Add `Raw` utility type to `IStakingApi`
- Add `Raw` utility type to `IBalancesApi`
- Update validation where appropriate
- Update tests accordingly

### DIFF
--- a/src/routes/balances/balances.controller.spec.ts
+++ b/src/routes/balances/balances.controller.spec.ts
@@ -800,7 +800,7 @@ describe('Balances Controller (Unit)', () => {
       });
     });
 
-    it(`500 error if validation fails`, async () => {
+    it(`503 error if validation fails`, async () => {
       const chainId = '1';
       const safeAddress = getAddress(faker.finance.ethereumAddress());
       const chainResponse = chainBuilder().with('chainId', chainId).build();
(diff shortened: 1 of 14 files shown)


src/routes/balances/balances.controller.spec.ts
```diff
@@ -830,13 +830,13 @@ describe('Balances Controller (Unit)', () => {
...
+        .expect(503)
         .expect({
-          statusCode: 500,
-          message: 'Internal server error',
+          code: 503,
+          message: 'Service unavailable',
         });
 
-      expect(networkService.get.mock.calls.length).toBe(4);
+      expect(networkService.get.mock.calls.length).toBe(3);
```
iamacook, line 839: These will be addressed by [adjusting the `HttpErrorFactory`](https://github.com/safe-global/safe-client-gateway/pull/2100/files#diff-6393ab1bfb621198efe5705d010d4bcd4991e56ff2eb0b2f062942d70346f20aR18).

hectorgomezv: Sounds good to me, and I'm pretty sure you are aware, but just something to have in mind: I'd say we need to be cautious there and avoid exposing the details of a ZodError when the validation failure comes from an error while processing a Datasource response (e.g.: a Transaction Service response), to avoid leaking data through the public route controller 👍🏻

hectorgomezv (approved):

src/routes/balances/balances.controller.spec.ts
iamacook: Totally agree with you!

PR #2104, opened 2024-11-07
Fix jwt algorithm

PooyaRaki:
## Summary
This PR fixes an issue where the JWT algorithm was incorrectly passed in the payload when signing and decoding. The algorithm is now correctly placed in the third parameter as an option, ensuring proper signing/decoding of tokens with the desired algorithms.

## Changes
- Moved the JWT algorithm from the payload to the options parameter in the sign/verify/decode method
- Verified that tokens are now signed/verified/decoded and validated with the specified algorithm

### DIFF
--- a/src/datasources/jwt/jwt.module.ts
+++ b/src/datasources/jwt/jwt.module.ts
@@ -30,7 +30,7 @@ function jwtClientFactory() {
           ...rest,
         },
         options.secretOrPrivateKey,
-        { algorithm: options.algorithm ?? 'HS256' },
+        { algorithm: options.algorithm },
       );
     },
     verify: <T extends object>(
--- a/src/datasources/jwt/jwt.module.ts
+++ b/src/datasources/jwt/jwt.module.ts
@@ -1,4 +1,4 @@
-import jwt from 'jsonwebtoken';
+import jwt, { type Algorithm } from 'jsonwebtoken';
 import { Module } from '@nestjs/common';
 import { JwtService } from '@/datasources/jwt/jwt.service';
 import { IJwtService } from '@/datasources/jwt/jwt.service.interface';
@@ -17,7 +17,7 @@ function jwtClientFactory() {
       },
     >(
       payload: T,
-      options: { secretOrPrivateKey: string },
+      options: { secretOrPrivateKey: string; algorithm?: jwt.Algorithm },
     ): string => {
       // All date-based claims should be second-based NumericDates
       const { exp, iat, nbf, ...rest } = payload;
@@ -30,13 +30,19 @@ function jwtClientFactory() {
           ...rest,
         },
         options.secretOrPrivateKey,
+        { algorithm: options.algorithm ?? 'HS256' },
       );
     },
     verify: <T extends object>(
       token: string,
-      options: { issuer: string; secretOrPrivateKey: string },
+      options: {
+        issuer: string;
+        secretOrPrivateKey: string;
+        algorithms?: Array<Algorithm>;
+      },
     ): T => {
       return jwt.verify(token, options.secretOrPrivateKey, {
+        algorithms: options.algorithms,
         issuer: options.issuer,
         // Return only payload without claims, e.g. no exp, nbf, etc.
         complete: false,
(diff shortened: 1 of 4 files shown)


src/datasources/jwt/jwt.module.ts
```diff
@@ -17,7 +17,7 @@ function jwtClientFactory() {
       },
     >(
       payload: T,
-      options: { secretOrPrivateKey: string },
+      options: { secretOrPrivateKey: string; algorithm?: jwt.Algorithm },
```
iamacook, line 20: You're importing the type.

```suggestion
      options: { secretOrPrivateKey: string; algorithm?:Algorithm },
```

src/datasources/jwt/jwt.module.ts
```diff
@@ -30,24 +30,35 @@ function jwtClientFactory() {
           ...rest,
         },
         options.secretOrPrivateKey,
+        { algorithm: options.algorithm ?? 'HS256' },
```
iamacook, line 33: Nit: if we want to explicitly state the default here (instead of that of the package), I'd move it to a constant. We should also use it in the other methods, `verify` and `decode`.

PooyaRaki: The sign method needs the default value:
https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/sign.js#L98
But the other two don't as they accept an array.

PooyaRaki: Also regarding `const`, since it's the default value I think it's safe to hardcode here.

src/datasources/jwt/jwt.module.ts
PooyaRaki: Removed the import of the algorithm type, as it’s already exported by JWT.
https://github.com/safe-global/safe-client-gateway/pull/2104/commits/fc9cb89a0dce9ab682384c64089723186bee3a5a

src/datasources/jwt/jwt.module.ts
iamacook: > The sign method needs the default value:
https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/sign.js#L98

The default value is set there though?

> Also regarding const, since it's the default value I think it's safe to hardcode here.

If we want to explictly specify it, a constant is much clearer imo.

> But the other two don't as they accept an array.

This was an oversight on my accord. I thought the default value was also assigned for those in the package.

PooyaRaki: I noticed that the default value is also hardcoded in similar cases, like in [jsonwebtoken’s code here](https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/sign.js#L98). I think following a similar pattern works well in our context, but let me know if you feel strongly about using a constant

PooyaRaki: Yes, the default value is set there, but it’s much safer to define it in our code. This way, if the default changes in the future, we won’t be affected.

PooyaRaki: However, the other method, verify, handles different scenarios when no default value is provided.
[Source](https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/verify.js#L132)

iamacook: Looking back at this, what do you think about assigning the default in the service ([where we do the `secretOrPrivateKey`](https://github.com/safe-global/safe-client-gateway/blob/fc9cb89a0dce9ab682384c64089723186bee3a5a/src/datasources/jwt/jwt.service.ts#L32)).

We should also add a test ensuring the default value is used.

I'll leave whether we use a constant or not to your discretion.

PooyaRaki: @iamacook 
Actually, setting the algorithm should be the very last step, as it’s the default for JWT, not specific to our setup.

Regarding testing, how do you think it could cause an issue? Since it’s hardcoded and stable, any break would indicate a bug in jsonwebtoken, not on our end.

PooyaRaki: @iamacook Please see the new changes
https://github.com/safe-global/safe-client-gateway/pull/2104/commits/24d91e768d39bb33c2e0509891aad32e09312257

iamacook (approved):

PR #2107, opened 2024-11-07
Fix flaky GET Creation Transaction controller test

hectorgomezv:
## Summary
This PR fixes a flaky test affecting `get-creation-transaction.transactions.controller.spec`.

An example of an error in this test can be seen in the following CI run: https://github.com/safe-global/safe-client-gateway/actions/runs/11724359558/job/32658109820

The problem was the `app` initialization was called inside a Jest's `beforeAll` hook, instead of a `beforeEach` one. As tests run in parallel, the mocked `networkService` instance was shared among all the tests, and even though `jest.resetAllMocks()` was called on the `beforeEach` hook, the `networkService` mock behaviour was sometimes unexpected due to race conditions between tests' executions.

## Changes
- Ensures the `app` initialization is called inside a Jest's `beforeEach` hook, instead of a `beforeAll` one.

### DIFF
--- a/src/routes/transactions/__tests__/controllers/get-creation-transaction.transactions.controller.spec.ts
+++ b/src/routes/transactions/__tests__/controllers/get-creation-transaction.transactions.controller.spec.ts
@@ -116,9 +116,6 @@ describe('Get creation transaction', () => {
             } as Response),
           );
         default:
-          console.log('************');
-          console.log(url, getChainUrl, chain, getCreationTransactionUrl);
-          console.log('************');
           return Promise.reject(new Error(`Could not match ${url}`));
       }
     });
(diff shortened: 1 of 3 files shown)


src/routes/transactions/__tests__/controllers/get-creation-transaction.transactions.controller.spec.ts
```diff
@@ -117,6 +116,9 @@ describe('Get creation transaction', () => {
             } as Response),
           );
         default:
+          console.log('************');
+          console.log(url, getChainUrl, chain, getCreationTransactionUrl);
+          console.log('************');
```
iamacook, line 121: 👀

hectorgomezv: Ouch 😣 Removed in [552a628](https://github.com/safe-global/safe-client-gateway/pull/2107/commits/552a6285b2bd06e7219b04a2888fb974a3f6df44)

iamacook (approved):

PR #2111, opened 2024-11-08
Return `502` for gateway validation errors

iamacook:
## Summary

When validating the response of a service, we return a `500`. This is a vague response.

This changes server validation-related error codes to `502`.

## Changes

- Change server validation-related errors codes


PooyaRaki (changes requested):
Question: Why do we need to change the response status code to 502?

For validation errors, the appropriate status code is generally 400 Bad Request, as it indicates an issue with the request sent by the client. However, since this is an internal validation error (an issue occurring within our service logic rather than due to a client error), 500 Internal Server Error is more suitable. This code reflects that the error was caused by the server’s internal processing rather than by the client’s input.

hectorgomezv:
[This error](https://github.com/safe-global/safe-client-gateway/actions/runs/11744543866/job/32719898463) happened because the way `AppModule` was initialized in `list-queued-transactions-by-safe.transactions.controller.spec.ts` prevented the injection of the Interceptors/Filters, and therefore the `ZodError` filter wasn't applied. I've created and merged a PR to fix this in https://github.com/safe-global/safe-client-gateway/pull/2127. After that, rebasing this onto `main` will make the tests pass 🙂

> Question: Why do we need to change the response status code to 502?

> For validation errors, the appropriate status code is generally 400 Bad Request, as it indicates an issue with the request sent by the client. However, since this is an internal validation error (an issue occurring within our service logic rather than due to a client error), 500 Internal Server Error is more suitable. This code reflects that the error was caused by the server’s internal processing rather than by the client’s input.

I don't have a strong opinion regarding this, but I think distinguishing 500/generic errors from 502 errors caused by server-side validation can be beneficial 🙂

PooyaRaki (approved):

PooyaRaki (approved):

PR #2132, opened 2024-11-15
Improve swagger

compojoom:
## Summary
I'm continuing to improve the generated swagger output. Most of the time it is the type that was known, but it in the generated swagger output it is set to the generic "string".

## Changes
The easiest way to review the PR would be commit by commit. I tried to make the changes small and I grouped similar changes together.


hectorgomezv (approved):
👏🏻👏🏻👏🏻

PR #2133, opened 2024-11-15
fix: timestamp can’t be null

compojoom:
I don't think that a transaction can have a null timestamp. Looking at what the tx-service returns https://safe-transaction-mainnet.safe.global/#/transactions/safes_all_transactions_list

It seems quite impossible, no?

Same goes for transactionHash.


hectorgomezv:
> It seems quite impossible, no?
The `timestamp` field is computed from several fields depending on the type of transaction that it's being mapped, but you are right, it should never be `null` given the current mappings. Good catch!

I've rebased the PR and solved a conflict before merging it btw.

hectorgomezv (approved):

PR #2149, opened 2024-11-22
Implement AddressBookItem update

hectorgomezv:
## Summary
This PR adds the ability to update an `AddressBookItem` by adding the following endpoint to the service:

```
PUT /accounts/:address/address-books/:chainId/:addressBookItemId
```

The endpoint is behind the `Accounts` scope and requires authentication.

## Changes
- Adds `UPDATE` route for deleting an `AddressBookItem`.


src/domain/accounts/address-books/address-books.repository.ts
```diff
@@ -86,6 +87,40 @@ export class AddressBooksRepository implements IAddressBooksRepository {
...
+   * an AddressBookNotFound error is thrown.
+   */
+  async updateAddressBookItem(args: {
+    authPayload: AuthPayload;
+    address: `0x${string}`;
+    chainId: string;
+    updateAddressBookItemDto: UpdateAddressBookItemDto;
+  }): Promise<AddressBookItem> {
+    if (!args.authPayload.isForSigner(args.address)) {
+      throw new UnauthorizedException();
```
PooyaRaki, line 104: We should send a message as the error message

src/domain/accounts/address-books/address-books.repository.ts
```diff
@@ -86,6 +87,40 @@ export class AddressBooksRepository implements IAddressBooksRepository {
...
+    }
+    await this.checkAddressBooksIsEnabled({
+      authPayload: args.authPayload,
+      address: args.address,
+    });
+    const account = await this.accountsRepository.getAccount({
+      authPayload: args.authPayload,
+      address: args.address,
+    });
+    const addressBook = await this.datasource.getAddressBook({
```
PooyaRaki, line 114: Question: Does it happen that the `account` is not found? If so, we shouldn't try to fetch the address book and update.

src/domain/accounts/address-books/entities/create-address-book-item.dto.entity.spec.ts
```diff
@@ -14,91 +13,6 @@ describe('CreateAddressBookItemDtoSchema', () => {
     expect(result.success).toBe(true);
   });
 
-  it('should not verify a CreateAddressBookItemDto with a shorter name', () => {
```
PooyaRaki, line 17: Why are `CreateAddressBookItemDto` tests deleted?

src/domain/accounts/address-books/entities/update-address-book-item.dto.entity.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    const updateAddressBookItemDto = updateAddressBookItemDtoBuilder().build();
+
+    const result = UpdateAddressBookItemDtoSchema.safeParse(
+      updateAddressBookItemDto,
+    );
+
+    expect(result.success).toBe(true);
+  });
+
+  it('should not verify an UpdateAddressBookItemDto with a string id', () => {
```
PooyaRaki, line 17: It seems to me that we are testing `Zod` validator, right?

src/domain/accounts/address-books/entities/update-address-book-item.dto.entity.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+        code: 'invalid_type',
+        expected: 'number',
+        message: 'Expected number, received string',
+        path: ['id'],
+        received: 'string',
+      },
+    ]);
+  });
+
+  it('should not verify an UpdateAddressBookItemDto with a malformed address', () => {
```
PooyaRaki, line 38: Here as well, it seems like we are testing the external validator library.

src/domain/accounts/address-books/entities/update-address-book-item.dto.entity.spec.ts
```diff
@@ -0,0 +1,70 @@
...
+    expect(!result.success && result.error.issues).toStrictEqual([
+      {
+        code: 'custom',
+        message: 'Invalid address',
+        path: ['address'],
+      },
+    ]);
+  });
+
+  it('should checksum the address of an UpdateAddressBookItemDto', () => {
```
PooyaRaki, line 56: Same as the previous comment.

src/datasources/accounts/address-books/address-books.datasource.ts
```diff
@@ -56,15 +56,15 @@ export class AddressBooksDatasource implements IAddressBooksDatasource {
 
   async updateAddressBookItem(args: {
     addressBook: AddressBook;
```
PooyaRaki, line 58: Why do we need the whole AddressBook Object? I see that the AddressBook Data is stored as a `BYTEA` in the db but we could use native Postgres `Jsonb` for example then use the native json functions of the Postgres to fetch the address book item.

src/domain/accounts/address-books/entities/address-book.entity.spec.ts
```diff
@@ -107,61 +107,6 @@ describe('AddressBookSchema', () => {
     ]);
   });
 
-  it('should not verify an AddressBookItem with a shorter name', () => {
```
PooyaRaki, line 110: Question: Why are these tests deleted?

src/domain/accounts/address-books/address-books.repository.ts
jmealy: it will throw here if there is no account so it won't try to fetch/update the addressbook

src/domain/accounts/address-books/entities/create-address-book-item.dto.entity.spec.ts
jmealy: From a comment in a previous PR: https://github.com/safe-global/safe-client-gateway/pull/2151#issuecomment-2559357775

@iamacook  suggested to just keep the tests for address book names in the address book name schema, so as not to test the same cases multiple times. I didn't remove them in that PR so removing them here.

src/domain/accounts/address-books/entities/update-address-book-item.dto.entity.spec.ts
iamacook: I would keep this as it's important to ensure addresses come in correct.

src/domain/accounts/address-books/entities/update-address-book-item.dto.entity.spec.ts
iamacook: I would keep this as it's important to ensure addresses come in checksummed.

src/domain/accounts/address-books/entities/address-book.entity.spec.ts
jmealy: same as https://github.com/safe-global/safe-client-gateway/pull/2149#discussion_r1910134211

src/datasources/accounts/address-books/address-books.datasource.ts
jmealy: since there will be significant changes as discussed earlier in the kickoff meeting, I will make a note to use jsonb for the AddressBook data field in the v2

katspaugh:
@PooyaRaki is this still in review?

PR #2151, opened 2024-11-22
Add CreateAddressBookItemDto schema tests

hectorgomezv:
## Changes
- Adds missing schema tests for `CreateAddressBookItemDto`.

### DIFF
--- a/src/domain/accounts/address-books/entities/address-book.entity.spec.ts
+++ b/src/domain/accounts/address-books/entities/address-book.entity.spec.ts
@@ -118,7 +118,7 @@ describe('AddressBookSchema', () => {
         code: 'too_small',
         exact: false,
         inclusive: true,
-        message: 'Address Books items names must be at least 3 characters long',
+        message: 'Address book entry names must be at least 3 characters long',
         minimum: 3,
         path: ['data', 0, 'name'],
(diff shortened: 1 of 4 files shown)


iamacook (commented):
The `address` tests look good, but I think we should move most of the `name` tests to a separate spec that only tests `AddressBookItemNameSchema`. What do you think?

I suggest we adjust the error messages of it as well:

- "Address Books items..." -> "Address book entry names...".
- "...and can contain..." -> "...and can only contain...".

jmealy:
> The `address` tests look good, but I think we should move most of the `name` tests to a separate spec that only tests `AddressBookItemNameSchema`. What do you think?
> 
> I suggest we adjust the error messages of it as well:
> 
> * "Address Books items..." -> "Address book entry names...".
> * "...and can contain..." -> "...and can only contain...".

Looks like tests for the name schema already exist in `address-book-item-name.schema.spec.ts`. They cover the same cases as the name tests here.

Changed the error message as suggested and fixed a typo: https://github.com/safe-global/safe-client-gateway/pull/2151/commits/9797af083f8a5ee94262cc9674fb3dc3334c4e57

iamacook (approved):

PR #2165, opened 2024-12-02
Use helper throughout `KilnApi`

iamacook:
## Summary

Kiln returns all responses nested in a `data` property. Within the `KilnApi`, we have a helper that parses this, casting the response as `Raw` accordingly. This is, however, not used consistently throughout the service.

This removes all redundant code, using the helper everywhere.

## Changes

- Move all validation/error handling inside helper
- Use helper in all methods

### DIFF
--- a/src/datasources/staking-api/kiln-api.service.ts
+++ b/src/datasources/staking-api/kiln-api.service.ts
@@ -15,7 +15,7 @@ import type { Stake } from '@/datasources/staking-api/entities/stake.entity';
 import type { TransactionStatus } from '@/datasources/staking-api/entities/transaction-status.entity';
 import type { IStakingApi } from '@/domain/interfaces/staking-api.interface';
 import type { Raw } from '@/validation/entities/raw.entity';
-import { z } from 'zod';
+import { z, ZodError } from 'zod';
 
 export class KilnApi implements IStakingApi {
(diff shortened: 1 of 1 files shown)


src/datasources/staking-api/kiln-api.service.ts
```diff
@@ -281,15 +246,23 @@ export class KilnApi implements IStakingApi {
...
+        }>(args)
+        .then((res) => {
+          // Ensuring response is { data: T }, the data is parsed in the domain
+          return z.object({ data: z.unknown() }).parse(res) as { data: Raw<T> };
+        });
+      return data;
+    } catch (error) {
+      if (error instanceof ZodError) {
+        throw error;
+      }
```
iamacook, line 263: `ZodError`s are used for types and are inherently not `DataSourceError`s.

PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2166#discussion_r1869786237

iamacook: See https://github.com/safe-global/safe-client-gateway/pull/2166#discussion_r1869790236

PooyaRaki (approved):

PR #2166, opened 2024-12-02
Don't create `DataSourceError` from `ZodError`s

iamacook:
## Summary

We now use validation schemas to ensure the incoming `Raw<T>` responses are as expected as it is "safer" than casting. If validation errors are thrown, however, they are therefore caught and thrown as `DataSourceError`s.

This throws any `ZodError`s directly, to be caught by the `ZodErrorFilter`.

## Changes

- Throw `ZodError`s directly within methods that parse against schemas.

### DIFF
--- a/src/datasources/balances-api/zerion-balances-api.service.ts
+++ b/src/datasources/balances-api/zerion-balances-api.service.ts
@@ -38,7 +38,7 @@ import { ILoggingService, LoggingService } from '@/logging/logging.interface';
 import { rawify, type Raw } from '@/validation/entities/raw.entity';
 import { Inject, Injectable } from '@nestjs/common';
 import { getAddress } from 'viem';
-import { z } from 'zod';
+import { z, ZodError } from 'zod';
 
 export const IZerionBalancesApi = Symbol('IZerionBalancesApi');
 
@@ -147,7 +147,7 @@ export class ZerionBalancesApi implements IBalancesApi {
       );
       return this._mapBalances(chainName, zerionBalances.data);
     } catch (error) {
-      if (error instanceof LimitReachedError) {
+      if (error instanceof LimitReachedError || error instanceof ZodError) {
         throw error;
       }
       throw this.httpErrorFactory.from(error);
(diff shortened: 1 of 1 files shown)


src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -147,7 +147,7 @@ export class ZerionBalancesApi implements IBalancesApi {
       );
       return this._mapBalances(chainName, zerionBalances.data);
     } catch (error) {
-      if (error instanceof LimitReachedError) {
+      if (error instanceof LimitReachedError || error instanceof ZodError) {
```
iamacook, line 150: Moving this check into the `HttpErrorFactory` seems excessive as it is only used in a few select places.

PooyaRaki: Thanks for your explanation. Isn't it used in all the places where we validate API responses? If that's the case I would move it to the factory.

src/datasources/balances-api/zerion-balances-api.service.ts
```diff
@@ -210,6 +210,9 @@ export class ZerionBalancesApi implements IBalancesApi {
           zerionCollectibles.data,
         );
       } catch (error) {
+        if (error instanceof ZodError) {
```
PooyaRaki, line 213: The same as the previous comment, Wouldn't it be better to move it to the factory as that's the class that handles error related logic?

src/datasources/balances-api/zerion-balances-api.service.ts
iamacook: Generally, we validate API responses on the domain layer. It is only used in a select few places (touched in this PR) to traverse from a `Raw<T>` type to `T`. I don't think it's therefore worthwhile locating it in the factory. What do you think?

src/datasources/balances-api/zerion-balances-api.service.ts
iamacook: See https://github.com/safe-global/safe-client-gateway/pull/2166#discussion_r1869790236.

PR #2174, opened 2024-12-04
Don't create `DataSourceError`s from `ZodError`s in `SafeBalancesApi`

iamacook:
## Summary

Similarly to https://github.com/safe-global/safe-client-gateway/pull/2166, we are traversing between `Raw<T>` and `T` within the `IBalancesApi` by validating instead of casting.

The `ZerionBalancesApi` was adjusted to not convert `ZodError`s to `DataSourceError`s. This makes the same adjustment for the `SafeBalancesApi`.

## Changes

- Don't wrap `ZodError`s as `DataSourceError`s


iamacook:
To [save exposing a `500`](https://github.com/safe-global/safe-client-gateway/actions/runs/12163747815/job/33923684512?pr=2174), I am moving these changes [into the PR which this supports](https://github.com/safe-global/safe-client-gateway/pull/2111).

PR #2175, opened 2024-12-05
Remove unnecessary `IpSchema`

iamacook:
## Summary

We validate the IP address in order to prevent Account creation abuse using our `IpSchema`. However, this schema re-exports a native `zod` method, with no custom adjustments.

This removes the `IpSchema`, using the validation method directly instead.

## Changes

- Remove `IpSchema` and associated test.


src/datasources/accounts/accounts.datasource.ts
```diff
@@ -247,8 +247,8 @@ export class AccountsDatasource implements IAccountsDatasource, OnModuleInit {
    * @param clientIp - client IP address.
    */
   private async checkCreationRateLimit(clientIp: string): Promise<void> {
-    const { success: isValidIp } = IpSchema.safeParse(clientIp);
-    if (!clientIp || !isValidIp) {
+    const { success: isValidIp } = z.string().ip().safeParse(clientIp);
+    if (!isValidIp) {
```
iamacook, line 251: The `clientIp` check is not needed as a falsy value will fail validation.

PooyaRaki (approved):

PR #2180, opened 2024-12-06
Map `domainHash`/`messageHash` to `MULTISIG` transactions and messages

iamacook:
## Summary

The hashes of an EIP-712 domain/message can be used to verify the legitimacy of a transaction. On the Web interface, these two values are calculated dynamically for each transaction/message.

This calculates/maps a new `typedData`, containing `domainHash` and `messageHash` to the `detailedExecutionInfo` of `MULTISIG` transactions and messages, allowing the client to render it directly.

Note: messages can only be initially proposed _with_ a signature so we need take into consideration whether it influences the viability of this.

## Changes

- Adds new `TypedDataMapper` for dynamically calculating `typedData` hashes.
- Adds/maps new `typedData` properties to `Message` and `MultisigExecutionDetails` entities.
- Adds/updates tests accordingly.


PooyaRaki (approved):

iamacook:
After discussion, we will close this for the following reasons:

1. Both hashes are currently displayed for Ledger-specific reasons that should be solved with the updated module. 
2. This will not work for messages-to-be-proposed. Messages can only be proposed with a signature, from which we calculate hashes.

PR #2189, opened 2024-12-10
Remove timestamp-reference transaction from start of page

iamacook:
## Summary

We group transactions by day, in between which we add date labels.

When fetching the first page of transactions (`offeset=0`), we group/label the result as is. For subsequent pages (`offset>0`), we modify the pagination to request the last transaction of the previous page in order to determine whether we return an initial date label or not, before removing it and mapping the remanining transactions.

Currently, the removal of this "extra" transaction is not assigning correctly due to abstraction into a separate function. This adjusts the logic flow to ensure the "extra" transaction is removed where it is passed to the mapper.

## Changes

- Move `slice` into main mapping function.

### DIFF
--- a/src/routes/transactions/mappers/transactions-history.mapper.ts
+++ b/src/routes/transactions/mappers/transactions-history.mapper.ts
@@ -53,14 +53,25 @@ export class TransactionsHistoryMapper {
     if (transactionsDomain.length == 0) {
       return [];
     }
-    // Must be retrieved before mapping others as we remove it from transactionsDomain
-    const previousTransaction = await this.getPreviousTransaction({
-      offset,
-      transactionsDomain,
-      chainId,
-      safe,
-      onlyTrusted,
-    });
+
+    let previousTransaction: TransactionItem | undefined;
+
+    /**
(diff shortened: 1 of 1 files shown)


src/routes/transactions/mappers/transactions-history.mapper.ts
```diff
@@ -115,8 +121,6 @@ export class TransactionsHistoryMapper {
       args.safe,
       args.onlyTrusted,
     );
-    // Remove first transaction that was requested to get previous day timestamp
-    args.transactionsDomain = args.transactionsDomain.slice(1);
```
PooyaRaki, line 119: Question: AFAI understood, the original `transactionsDomain` array is not mutated here because the `slice` method creates a new array. That's the reason why there first transaction is not removed, Am I right?

PooyaRaki:
It looks good to me 👍 ; I just have one remaining question.

src/routes/transactions/mappers/transactions-history.mapper.ts
iamacook: Correct.

Issue #2181, opened 2024-12-09
[Tech Debt] Optimize parallel requests for token/contracts


hectorgomezv:
### Context 
The async nature of the CGW causes the service to send several requests to the Safe Transaction Service in parallel for the same data, so only after the first one gets a response is the result cached. 

This is especially problematic when the CGW requests the tokens/contracts for all the transactions on a page, and the same token/contract is repeated many times on that page. This causes the CGW to request the same token/contract N times (before receiving any response from the Transaction Service) instead of 1 for each repeated token/contract.

### Requirements

- Collect a Set (no repeated items) before sending the requests to the Safe Transaction Service.
- Wait for the requests in the Set to be fulfilled/rejected. Subsequent calls will be cached.

### Examples

![Screenshot 2024-12-09 at 16 02 33](https://github.com/user-attachments/assets/0093d8f7-b121-4821-a519-5cd68aec7631)
![Screenshot 2024-12-09 at 16 03 23](https://github.com/user-attachments/assets/e8aa1239-82b3-43b1-88f6-f95a81c6561a)

hectorgomezv:
Closing as it was implemented on https://github.com/safe-global/safe-client-gateway/pull/2503 and https://github.com/safe-global/safe-client-gateway/pull/2530

PR #2190, opened 2024-12-11
Add demonstrative `DeduplicationInterceptor` to reduce duplicate requests

iamacook:
⚠️ This is demonstrative and to-be-discussed within the team.

## Summary

Ideates on #2181 and our previous (reverted) [promise registry](https://github.com/safe-global/safe-client-gateway/pull/586) implementation.

Due to the async nature of the project, services can send several requests of the same requests in parallel. Only after one of these succeeds is the result cached.

This implements an interceptor which registers each `Request` promise in a registry. Should a subsequent `Request` happen and the aforementioned registration exist, it is returned instead of making a parallel one.

## Changes

- Create and register `DeduplicationInterceptor` in the app

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -18,6 +18,7 @@ import { CollectiblesModule } from '@/routes/collectibles/collectibles.module';
 import { CommunityModule } from '@/routes/community/community.module';
 import { ContractsModule } from '@/routes/contracts/contracts.module';
 import { DataDecodedModule } from '@/routes/data-decode/data-decoded.module';
+import { DeduplicationInterceptor } from '@/routes/common/interceptors/deduplication.interceptor';
 import { DelegatesModule } from '@/routes/delegates/delegates.module';
 import {
   HooksModule,
@@ -159,6 +160,10 @@ export class AppModule implements NestModule {
         {
           provide: APP_INTERCEPTOR,
           useClass: CacheControlInterceptor,
+        },
+        {
+          provide: APP_INTERCEPTOR,
+          useClass: DeduplicationInterceptor,
         },
         {
           provide: APP_FILTER,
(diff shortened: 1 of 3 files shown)


src/routes/common/interceptors/deduplication.interceptor.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+      .get('/get')
+      .query({ param: 'value' });
+
+    expect(response1.body).toEqual({ message: 'Hello, world!' });
+    expect(response2.body).toEqual({ message: 'Hello, world!' });
+
+    expect(testService.fetchData).toHaveBeenCalledTimes(2);
+  });
+
+  it.todo('test all variations of registry key');
```
iamacook, line 77: Should we approve of the implementation, these tests will be expanded.

src/routes/common/interceptors/deduplication.interceptor.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    ]);
+
+    expect(response1.body).toEqual({ message: 'Hello, world!' });
+    expect(response2.body).toEqual({ message: 'Hello, world!' });
+
+    expect(testService.fetchData).toHaveBeenCalledTimes(1);
+  });
+
+  it('should handle different requests separately', async () => {
+    (testService.fetchData as jest.Mock).mockClear();
```
iamacook, line 64: I'll align this more with our mocking approach should be move foward with it.

src/app.module.ts
```diff
@@ -160,6 +161,10 @@ export class AppModule implements NestModule {
           provide: APP_INTERCEPTOR,
           useClass: CacheControlInterceptor,
         },
+        {
+          provide: APP_INTERCEPTOR,
```
iamacook, line 165: This might need be localised to datasources, otherwise different calls to datasources within a mapping might still make separate requests.

iamacook: I've just experimented some more and this is indeed the case.

PR #2191, opened 2024-12-11
Feat(Chains): add recommendedMasterCopyVersion field to API

katspaugh:


src/routes/safes/safes.service.ts
```diff
@@ -341,14 +341,14 @@ export class SafesService {
 
   private computeVersionState(
     safe: Safe,
-    recommendedSafeVersion: string,
+    recommendedMasterCopyVersion: string,
```
katspaugh, line 344: Renamed this for consistency.

iamacook (approved):

PR #2194, opened 2024-12-16
Allow targeting of all safes for outreach campaigns

jmealy:
## Summary
Allows the creation of outreach campaigns with the option to show to all safes, not requiring a list of safes to target.  Now, if you try to create a submission where targetAll is true, it will create an entry in `targeted_safes` to use, instead of requiring that an entry already exist. 
Also creates a new outreach campaign using this new option.

## Changes
- Adds `target_all` column to the `outreaches` table
- Updates the `createSubmission` to create a targeted safe when `target_all` is true for the outreach.
- Adds a test for this case


src/datasources/targeted-messaging/targeted-messaging.datasource.spec.ts
```diff
@@ -132,6 +133,39 @@ describe('TargetedMessagingDataSource tests', () => {
...
+        name: dto.name,
+        startDate: dto.startDate,
+        endDate: dto.endDate,
+        sourceId: dto.sourceId,
+        type: dto.type,
+        teamName: dto.teamName,
+        sourceFile: dto.sourceFile,
+        sourceFileProcessedDate: null,
+        sourceFileChecksum: null,
+        targetAll: false,
```
PooyaRaki, line 154: The value is already inside the `DTO` builder, so we can fetch it from the builder object.

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
```diff
@@ -101,6 +101,23 @@ export class TargetedMessagingDatasource
     return this.outreachDbMapper.map(dbOutreach);
   }
 
+  async getOutreach(outreachId: number): Promise<Outreach> {
+    const [dbOutreach] = await this.sql<
+      DbOutreach[]
+    >`SELECT * FROM outreaches WHERE id = ${outreachId}`.catch((err) => {
```
PooyaRaki, line 107: Instead of using `*`, we should explicitly specify the columns we want to fetch. Using `*` is unsafe because future changes to the table, such as adding a blob field that we don’t need to fetch here, could lead to performance issues. More about the design pattern:
https://stackoverflow.com/questions/3639861/why-is-select-considered-harmful

src/routes/targeted-messaging/targeted-messaging.service.ts
```diff
@@ -44,27 +45,54 @@ export class TargetedMessagingService {
     }
   }
 
+  private async getOrCreateTargetedSafe(args: {
```
PooyaRaki, line 48: If the survey is public and not targeted, why do we need to call `createTargetedSafes`? Is it possible to simply create a submission once the user has submitted the survey? Creating a separate entity for each Safe seems unnecessary and could overwhelm the database.

src/routes/targeted-messaging/targeted-messaging.service.ts
```diff
@@ -44,27 +45,54 @@ export class TargetedMessagingService {
...
+    safeAddress: `0x${string}`;
+  }): Promise<TargetedSafe> {
+    try {
+      return await this.repository.getTargetedSafe(args);
+    } catch (err) {
+      const outreach = await this.repository.getOutreach({
+        outreachId: args.outreachId,
+      });
+
+      if (!outreach.targetAll) {
```
PooyaRaki, line 59: It seems like if the outreach does not exist we get an error trying to access `targetAll`. If it's intentional I would call the method `getOutreachOrFail` to make it clear that it throws if the entity is not found.

src/routes/targeted-messaging/targeted-messaging.service.ts
```diff
@@ -44,27 +45,54 @@ export class TargetedMessagingService {
...
+        // Safe is not targeted
+        throw err;
+      }
+
+      const [targetedSafe] = await this.repository.createTargetedSafes({
```
PooyaRaki, line 64: As stated before, It would be nice to not create a new entity for each user.

src/domain/targeted-messaging/targeted-messaging.repository.ts
```diff
@@ -45,4 +53,8 @@ export class TargetedMessagingRepository
   }): Promise<Submission> {
     return this.datasource.createSubmission(args);
   }
+
+  async getOutreach(args: { outreachId: number }): Promise<Outreach> {
```
PooyaRaki, line 57: Nit: If we only have one property inside the object we can define it directly:
`async getOutreach(outreachId: number): Promise<Outreach> {`

src/domain/targeted-messaging/targeted-messaging.repository.ts
```diff
@@ -23,6 +25,12 @@ export class TargetedMessagingRepository
     return this.datasource.getTargetedSafe(args);
   }
 
+  async createTargetedSafes(
+    args: CreateTargetedSafesDto,
+  ): Promise<TargetedSafe[]> {
```
PooyaRaki, line 30: Nit: The array types in CGW are usually defined like:
`): Promise<Array<TargetedSafe>> {`

src/domain/targeted-messaging/entities/outreach.entity.ts
```diff
@@ -18,4 +18,5 @@ export const OutreachSchema = RowSchema.extend({
   sourceFile: z.string().nullish().default(null),
   sourceFileProcessedDate: z.coerce.date().nullish().default(null),
   sourceFileChecksum: z.string().nullish().default(null),
+  targetAll: z.boolean().default(false),
```
PooyaRaki, line 21: Since the default is already defined in the entity, can we remove the default here to ensure a value is explicitly provided?

src/domain/targeted-messaging/entities/tests/create-outreach.dto.builder.ts
```diff
@@ -16,5 +16,6 @@ export function createOutreachDtoBuilder(): IBuilder<CreateOutreachDto> {
     .with('teamName', faker.string.alphanumeric(5))
     .with('sourceFile', `${faker.string.alphanumeric(5)}.json`)
     .with('sourceFileProcessedDate', null)
-    .with('sourceFileChecksum', null);
+    .with('sourceFileChecksum', null)
+    .with('targetAll', false);
```
PooyaRaki, line 20: It would be nice to make it random to be able to test different scenarios.

src/datasources/targeted-messaging/targeted-messaging.datasource.spec.ts
```diff
@@ -115,6 +115,7 @@ describe('TargetedMessagingDataSource tests', () => {
         sourceFile: dto.sourceFile,
         sourceFileProcessedDate: null,
         sourceFileChecksum: null,
+        targetAll: false,
```
PooyaRaki, line 118: Here we can access the value from the `DTO`. And making the DTO random would help test more scenarios.

PooyaRaki (changes requested):
Very well done 👏
I left some comments.

src/routes/targeted-messaging/targeted-messaging.service.ts
jmealy: Just because each submission requires a link to an entry in targeted_safes. We could maybe get rid of this requirement and have an option to store the safe address directly in each submission instead?

src/routes/targeted-messaging/targeted-messaging.service.ts
jmealy: This was intended yeah, so if the requested outreach doesn't exist you get this response:
`{  "message": "Outreach not found",  "error": "Unprocessable Entity",  "statusCode": 422}`
So I'll change to getOutreachOrFail

src/routes/targeted-messaging/targeted-messaging.service.ts
PooyaRaki: I see. Since it’s not targeted, I find it confusing to create a new entity in the targeted safe table. Additionally, the name of the method, `createTargetedSafe`, is misleading for me. At first glance, it gives the impression that the method creates a new safe account. What do you think?

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
```diff
@@ -89,7 +90,7 @@ export class TargetedMessagingDatasource
       start_date = ${updateOutreachDto.startDate},
       end_date = ${updateOutreachDto.endDate},
       type = ${updateOutreachDto.type},
-      team_name = ${updateOutreachDto.teamName}
+      team_name = ${updateOutreachDto.teamName}
```
iamacook, line 93: Nit: please avoid such changes.

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
```diff
@@ -101,6 +102,25 @@ export class TargetedMessagingDatasource
...
+      (err) => {
+        this.loggingService.warn(
+          `Error getting outreach: ${asError(err).message}`,
+        );
+        throw new UnprocessableEntityException('Error getting outreach');
+      },
+    );
+
+    if (!dbOutreach) {
+      throw new UnprocessableEntityException('Outreach not found');
```
iamacook, line 118: I think it would make sense to use a `NotFoundException` here.

src/domain/targeted-messaging/entities/create-outreach.dto.entity.ts
```diff
@@ -15,6 +15,7 @@ export const CreateOutreachDtoSchema = z.object({
   sourceFile: z.string().nullish().default(null),
   sourceFileProcessedDate: z.coerce.date().nullish().default(null),
   sourceFileChecksum: z.string().nullish().default(null),
+  targetAll: z.boolean(),
```
iamacook, line 18: Can be done after this PR, but I noticed we have no test coverage for this schema.

src/domain/targeted-messaging/entities/outreach.entity.ts
```diff
@@ -18,4 +18,5 @@ export const OutreachSchema = RowSchema.extend({
   sourceFile: z.string().nullish().default(null),
   sourceFileProcessedDate: z.coerce.date().nullish().default(null),
   sourceFileChecksum: z.string().nullish().default(null),
+  targetAll: z.boolean(),
```
iamacook, line 21: As above regarding test coverage.

src/routes/targeted-messaging/targeted-messaging.controller.spec.ts
```diff
@@ -318,6 +318,79 @@ describe('TargetedMessagingController', () => {
...
+      const chain = chainBuilder().build();
+      const signerAddress = getAddress(faker.finance.ethereumAddress());
+      const safeAddress = getAddress(faker.finance.ethereumAddress());
+      const submission = submissionBuilder().build();
+      const safe = safeBuilder()
+        .with('owners', [
+          getAddress(faker.finance.ethereumAddress()),
+          signerAddress,
+        ])
+        .build();
```
iamacook, line 332: Should `safe.address` be `safeAddress` here?

src/routes/targeted-messaging/targeted-messaging.controller.spec.ts
```diff
@@ -318,6 +318,79 @@ describe('TargetedMessagingController', () => {
         });
     });
 
+    it('should create a submission for a campaign without targeting enabled', async () => {
+      const outreachId = faker.number.int();
+      const chain = chainBuilder().build();
+      const signerAddress = getAddress(faker.finance.ethereumAddress());
+      const safeAddress = getAddress(faker.finance.ethereumAddress());
+      const submission = submissionBuilder().build();
```
iamacook, line 326: Should `submission.build` be `signerAddress` here?

src/routes/targeted-messaging/targeted-messaging.service.ts
iamacook: Nit: this ties in with my aforementioned usage of `NotFoundException` - `422` is validation-specific.

src/routes/targeted-messaging/targeted-messaging.controller.spec.ts
jmealy: true, no need for both

src/datasources/targeted-messaging/targeted-messaging.datasource.spec.ts
```diff
@@ -143,6 +177,7 @@ describe('TargetedMessagingDataSource tests', () => {
         endDate: faker.date.recent(),
         teamName: faker.string.alphanumeric(),
         type: faker.string.alphanumeric(),
+        targetAll: false,
```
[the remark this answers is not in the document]
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/1b72d7acb028af5275397cac1a6b18e63e85e435#diff-ae1f316c87534e60c2f465f11b0301577d8eaf3b6505da5f0bbc49fe31d3fe07L118

src/datasources/targeted-messaging/targeted-messaging.datasource.spec.ts
jmealy: https://github.com/safe-global/safe-client-gateway/commit/1b72d7acb028af5275397cac1a6b18e63e85e435#diff-ae1f316c87534e60c2f465f11b0301577d8eaf3b6505da5f0bbc49fe31d3fe07L154

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
jmealy: https://github.com/safe-global/safe-client-gateway/commit/1b72d7acb028af5275397cac1a6b18e63e85e435#diff-7b46005236ccadec6f3c236cb143960dcc7f8c53afb914e828c907dd112384dbL107

src/routes/targeted-messaging/targeted-messaging.service.ts
jmealy: Use NotFoundException: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/65896bde7d8fef30ed73e817916f179ddf6cf334#diff-7b46005236ccadec6f3c236cb143960dcc7f8c53afb914e828c907dd112384dbL118

Name change: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/65896bde7d8fef30ed73e817916f179ddf6cf334#diff-c00fbe12e870d92628cb543f6ad60251cca6ee18294c7a674612797d9e02cb1fL55

src/domain/targeted-messaging/targeted-messaging.repository.ts
jmealy: https://github.com/safe-global/safe-client-gateway/commit/1b72d7acb028af5275397cac1a6b18e63e85e435#diff-14eea5767397a0c69489890e40793e5400cffb4237b1a6cb426652651659ad9bL57

src/domain/targeted-messaging/targeted-messaging.repository.ts
jmealy: https://github.com/safe-global/safe-client-gateway/commit/1b72d7acb028af5275397cac1a6b18e63e85e435#diff-9bc05b83e026841bc7b3f0dd8c8b4ace8e81ca26aacdf3d56079dd5215316b22L20

src/domain/targeted-messaging/entities/outreach.entity.ts
jmealy: https://github.com/safe-global/safe-client-gateway/commit/1b72d7acb028af5275397cac1a6b18e63e85e435#diff-9057e364f06b991e1f8974eec6c1d72658c8120d2a7b7770fbad66cd7965c4c8L18

src/datasources/targeted-messaging/targeted-messaging.datasource.spec.ts
jmealy: https://github.com/safe-global/safe-client-gateway/commit/1b72d7acb028af5275397cac1a6b18e63e85e435#diff-ae1f316c87534e60c2f465f11b0301577d8eaf3b6505da5f0bbc49fe31d3fe07L196

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/65896bde7d8fef30ed73e817916f179ddf6cf334#diff-7b46005236ccadec6f3c236cb143960dcc7f8c53afb914e828c907dd112384dbL93

src/datasources/targeted-messaging/targeted-messaging.datasource.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/65896bde7d8fef30ed73e817916f179ddf6cf334#diff-7b46005236ccadec6f3c236cb143960dcc7f8c53afb914e828c907dd112384dbL118

src/routes/targeted-messaging/targeted-messaging.controller.spec.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/65896bde7d8fef30ed73e817916f179ddf6cf334#diff-dec390951f4b8861a13a47d81f8e2989f8702c3054dd6562443153fb081f26a7L330

src/routes/targeted-messaging/targeted-messaging.service.ts
jmealy: renamed `createTargetedSafe` to  `addSafeToOutreach`: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/7f93ed57b5881c3474a6fb2f6d0550d583647041

If you're ok to leave the creation of new entries in `targeted_safes` for now like we discussed on slack, we can handle this in the next iteration.

src/datasources/targeted-messaging/targeted-messaging.datasource.spec.ts
```diff
@@ -132,8 +133,40 @@ describe('TargetedMessagingDataSource tests', () => {
...
+
+      await expect(target.getOutreachOrFail(nonExistentId)).rejects.toThrow(
+        'Outreach not found',
+      );
+    });
+  });
+
   describe('updateOutreach', () => {
-    it('should update an outreach successfully', async () => {
+    it.only('should update an outreach successfully', async () => {
```
PooyaRaki, line 169: I think `.only` was just forgotten to be removed.

src/routes/targeted-messaging/targeted-messaging.service.ts
PooyaRaki: Sure, Let's do that in another PR.

src/routes/targeted-messaging/targeted-messaging.service.ts
PooyaRaki: We decided to do it in another PR

migrations/deprecated/00013_targeted_messaging_add_outreach_2/index.sql
```diff
@@ -0,0 +1,21 @@
+INSERT INTO outreaches (
+    source_id,
+    name,
+    start_date,
+    end_date,
+    type,
+    team_name,
+    source_file,
+    source_file_checksum,
+	target_all
```
iamacook, line 10:
```suggestion
    target_all
```

migrations/deprecated/00013_targeted_messaging_add_outreach_2/index.sql
```diff
@@ -0,0 +1,21 @@
...
+) VALUES (
+    2,
+    'micro-campaign',
+    '1970-01-01 00:00:00.000',
+    '1970-01-01 00:00:00.000',
+    '',
+    '',
+    '',
+    '',
```
iamacook, line 19: Is there a reason to not populate these values? Afaiu, we need to add the checksum to target.

src/datasources/targeted-messaging/targeted-messaging.datasource.spec.ts
```diff
@@ -132,6 +133,25 @@ describe('TargetedMessagingDataSource tests', () => {
     });
   });
 
+  describe('getOutreachOrFail', () => {
+    it('gets an outreach successfully', async () => {
+      const dto = createOutreachDtoBuilder().build();
+      const created = await target.createOutreach(dto);
+      console.log('created', created);
```
iamacook, line 140:
```suggestion
```

src/datasources/targeted-messaging/outreach-file-processor.spec.ts
```diff
@@ -138,13 +139,33 @@ describe('OutreachFileProcessor', () => {
...
+  });
+
   it('should not process Outreach data files with bad checksum', async () => {
     const sourceFile = path.resolve(baseDir, fileName);
     const expectedChecksum = faker.string.alphanumeric();
     const created = await targetedMessagingDatasource.createOutreach(
       createOutreachDtoBuilder()
         .with('sourceFile', sourceFile)
         .with('sourceFileChecksum', expectedChecksum)
+        .with('targetAll', false)
```
PooyaRaki, line 168: Question: why is it added? Does it make a difference if it's false or true here?

src/datasources/targeted-messaging/outreach-file-processor.spec.ts
```diff
@@ -204,6 +226,7 @@ describe('OutreachFileProcessor', () => {
         createOutreachDtoBuilder()
           .with('sourceFile', badSourceFile)
           .with('sourceFileChecksum', checksum)
+          .with('targetAll', false)
```
PooyaRaki, line 229: Question: why is it added? Does it make a difference if it's false or true here?

src/datasources/targeted-messaging/outreach-file-processor.spec.ts
```diff
@@ -255,6 +279,7 @@ describe('OutreachFileProcessor', () => {
         .with('sourceId', outreachFile.campaign_id)
         .with('sourceFile', sourceFile)
         .with('sourceFileChecksum', fileChecksum)
+        .with('targetAll', false)
```
PooyaRaki, line 282: As the previous comments: Question: why is it added? Does it make a difference if it's false or true here?

src/datasources/targeted-messaging/outreach-file-processor.spec.ts
jmealy: if targetAll is will not try to process the file, so it wont check the checksum

src/datasources/targeted-messaging/outreach-file-processor.spec.ts
```diff
@@ -173,6 +194,7 @@ describe('OutreachFileProcessor', () => {
         createOutreachDtoBuilder()
           .with('sourceFile', badSourceFile)
           .with('sourceFileChecksum', checksum)
+          .with('targetAll', false)
```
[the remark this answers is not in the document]
jmealy: similar reason to the above

migrations/deprecated/00013_targeted_messaging_add_outreach_2/index.sql
PooyaRaki: @iamacook I think it's because this iteration does not target specific safes and therefor we don't have any file and checksum to validate.

iamacook: If that's the case, I'd exclude this migration.

jmealy: In this case we don't have a file with the targeted safes since we are showing to all safes. And the checksum was for the targeted safes file

src/datasources/targeted-messaging/outreach-file-processor.spec.ts
jmealy: similar here, if it is true it will skip processing the targetedSafes in the test wont be created

migrations/deprecated/00013_targeted_messaging_add_outreach_2/index.sql
iamacook: If the primary goal of this PR is functionality, I’d recommend splitting the creation of the outreach into a separate PR. This would keep the focus narrower and the diff. smaller. For example, if the migration were to fail, it would be easier to pinpoint whether the functionality or migration caused the issue.

Regarding the values: what are your thoughts on making `source_file` and `source_file_checksum` nullable instead of setting them to an empty string? This approach might better align with their intended semantics.

I’d also suggest using more specific dates where possible for clarity.

jmealy: I removed the outreach creation, will do in another PR

iamacook (approved):

src/domain/targeted-messaging/targeted-messaging.repository.ts
```diff
@@ -23,6 +24,16 @@ export class TargetedMessagingRepository
     return this.datasource.getTargetedSafe(args);
   }
 
+  async addSafeToOutreach(args: {
+    outreachId: number;
+    safeAddress: `0x${string}`;
+  }): Promise<Array<TargetedSafe>> {
+    return this.datasource.createTargetedSafes({
```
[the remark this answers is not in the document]
jmealy: I thought it made sense to leave it here as we are creating entries in the table called `targeted_safes` so `createTargetedSafes` to me makes sense for the datasource.
Should we always keep names consistent between `domain` and `datasource`?

src/routes/targeted-messaging/targeted-messaging.controller.spec.ts
```diff
@@ -318,6 +318,77 @@ describe('TargetedMessagingController', () => {
         });
     });
 
+    it('should create a submission for a campaign without targeting enabled', async () => {
```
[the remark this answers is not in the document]
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/5ffc6ced68485bb882d4451ce204b0c9448b3598

src/routes/targeted-messaging/targeted-messaging.service.ts
```diff
@@ -44,27 +45,52 @@ export class TargetedMessagingService {
...
+
   async createSubmission(args: {
     outreachId: number;
     chainId: string;
     safeAddress: `0x${string}`;
     signerAddress: `0x${string}`;
     createSubmissionDto: CreateSubmissionDto;
   }): Promise<Submission> {
-    const targetedSafe = await this.repository.getTargetedSafe(args);
+    const targetedSafe = await this.getOrCreateTargetedSafe(args);
```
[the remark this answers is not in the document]
jmealy: `getSubmission` is run when anybody opens a safe, to check if we should show the popup. So changing this here will result in creating more `targeted_safe` entries than necessary.
If somebody dismisses the popup without ever opening the survey, we don't need to create an entry for them but in this case a `targeted_safe` entry would be created for them.

src/routes/targeted-messaging/targeted-messaging.service.ts
```diff
@@ -44,27 +45,52 @@ export class TargetedMessagingService {
     }
   }
 
+  private async getOrCreateTargetedSafe(args: {
```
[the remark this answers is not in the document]
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2194/commits/5e9d5321f18ec891084cf3d4dfe9a08c63865fab

src/routes/targeted-messaging/targeted-messaging.service.ts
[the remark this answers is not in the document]
PooyaRaki: My suggestion was to change the name of the method. As you already changed the name, I'm going to resolve the thread.

src/domain/targeted-messaging/targeted-messaging.repository.ts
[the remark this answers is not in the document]
PooyaRaki: If it inserts into the `targeted_safe` table, it makes perfect sense to leave it as it is.

PooyaRaki (approved):
LGTM 🚀

PR #2195, opened 2024-12-16
Return the total count of all chains instead of the current page count

PooyaRaki:
## Summary
This PR modifies the functionality to return the total count of all chains rather than limiting the count to only the current page. This ensures better visibility of the complete dataset while preserving the pagination functionality for retrieving individual records.

## Changes
- Updated the API logic to return the total count of all chains in the response.

### DIFF
--- a/src/domain/entities/schemas/page.schema.factory.ts
+++ b/src/domain/entities/schemas/page.schema.factory.ts
@@ -30,9 +30,14 @@ export function buildLenientPageSchema<T extends z.ZodTypeAny>(
       return result.success ? [result.data] : [];
     });
 
+    // Note: @TODO The `results` object includes a `count` property, which the client is not currently using.
+    // We validate the chains and subtract the invalid ones from the count.
+    // However, due to pagination, we can only validate the chains on the current page,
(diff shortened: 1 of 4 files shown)


rmeissner (approved):

src/domain/entities/schemas/page.schema.factory.ts
```diff
@@ -32,7 +32,6 @@ export function buildLenientPageSchema<T extends z.ZodTypeAny>(
 
     return {
       ...data,
-      count: results.length,
```
iamacook, line 35: As this is not used by the client (yet), I think it's fine to ignore the invalid counts but we should address it at some stage (reducing the count). Can you add a comment to reflect this?

iamacook (approved):

src/domain/entities/schemas/page.schema.factory.ts
PooyaRaki: Addressed
https://github.com/safe-global/safe-client-gateway/pull/2195/commits/5f6e7ada9ce2871a7050aa741aeefc01c8910e4f

PR #2210, opened 2024-12-18
Create new outreach campaign

jmealy:
## Summary
Creates a second outreach campaign for collecting user feedback.

## Changes
Creates a migration to insert a new entry into `outreaches`. 
This outreach will be shown to all safes (target_all=true)

### DIFF
--- a/migrations/deprecated/00013_targeted_messaging_add_outreach_2/index.sql
+++ b/migrations/deprecated/00013_targeted_messaging_add_outreach_2/index.sql
@@ -1,0 +1,17 @@
+INSERT INTO outreaches (
+    source_id,
+    name,
+    start_date,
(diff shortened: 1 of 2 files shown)


iamacook (changes requested):
Whilst simple, as this adds a new field/doesn't populate fields that were present in the previous migration (00008), can you add a simple test for it?

You can use [the test for the aforementioned](https://github.com/safe-global/safe-client-gateway/blob/dependabot/npm_and_yarn/typescript-eslint-8.18.1/migrations/deprecated/__tests__/00008_targeted_messaging_add_outreach_1.spec.ts#L0-L1) as refrence.

iamacook (approved):

PR #2213, opened 2024-12-19
Allow Safe nonce to be a number or string

iamacook:
## Summary

The Transaction Service now [returns a string `nonce` for Safes](https://github.com/safe-global/safe-transaction-service/pull/2367). Due to this change, our validation throws - we expect a number.

This updates the validation schema to allow a number or string, coercing to a number so as to not require changes clientside.

Note: we should consider eventually changing this on clientside so as to avoid coercion.

## Changes

- Allow number/string nonce
- Update tests accordingly


PooyaRaki (approved):
LGTM

PR #2215, opened 2024-12-19
Trigger cache invalidation tests by queue

iamacook:
## Summary

Our cache is invalidated in accordance with queue events. Most of the test coverage was linked to our previous POST hooks, and therefore skipped.

This converts the tests to trigger via queue instead of POST calls, unskipping them.

## Changes

- Unskip cache invalidation tests
- Modify tests to trigger via queue

### DIFF
--- a/src/routes/hooks/hooks-cache.spec.ts
+++ b/src/routes/hooks/hooks-cache.spec.ts
@@ -2,7 +2,6 @@ import { faker } from '@faker-js/faker';
 import type { INestApplication } from '@nestjs/common';
 import type { TestingModule } from '@nestjs/testing';
 import { Test } from '@nestjs/testing';
-import request from 'supertest';
 import { TestCacheModule } from '@/datasources/cache/__tests__/test.cache.module';
 import { TestNetworkModule } from '@/datasources/network/__tests__/test.network.module';
 import { chainBuilder } from '@/domain/chains/entities/__tests__/chain.builder';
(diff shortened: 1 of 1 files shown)


src/routes/hooks/hooks-cache.spec.ts
```diff
@@ -98,158 +103,6 @@ describe.skip('Post Hook Events for Cache (Unit)', () => {
...
-          'PENDING_MULTISIG_TRANSACTION',
-          'REORG_DETECTED',
-          'SAFE_APPS_UPDATE',
-          'SAFE_CREATED',
-        ],
-        path: ['type'],
-        message:
-          "Invalid discriminator value. Expected 'CHAIN_UPDATE' | 'DELETED_MULTISIG_TRANSACTION' | 'EXECUTED_MULTISIG_TRANSACTION' | 'INCOMING_ETHER' | 'INCOMING_TOKEN' | 'MESSAGE_CREATED' | 'MODULE_TRANSACTION' | 'NEW_CONFIRMATION' | 'MESSAGE_CONFIRMATION' | 'OUTGOING_ETHER' | 'OUTGOING_TOKEN' | 'PENDING_MULTISIG_TRANSACTION' | 'REORG_DETECTED' | 'SAFE_APPS_UPDATE' | 'SAFE_CREATED'",
-      });
-  });
```
iamacook, line 251: These were removed as we receive no response. If we want to check this, however, we could check [whether an error is logged](https://github.com/safe-global/safe-client-gateway/blob/unskip-cache-tests/src/domain/hooks/hooks.repository.ts#L43).

src/routes/hooks/hooks-cache.spec.ts
```diff
@@ -1128,11 +927,42 @@ describe.skip('Post Hook Events for Cache (Unit)', () => {
...
+    await expect(fakeCacheService.hGet(cacheDir)).resolves.toBeUndefined();
+  });
+
+  it.each(
+    [
+      newDelegateEventBuilder().build(),
+      updatedDelegateEventBuilder().build(),
+      deletedDelegateEventBuilder().build(),
+    ].map((event) => [event.type, event]),
+  )('%s clears delegates', async (_, event) => {
```
iamacook, line 942: These were added as they were missing - initially marked with a TODO at the top.

src/routes/hooks/hooks-cache.spec.ts
PooyaRaki: I think we just need to add a new test to verify that an error is thrown when the event name is valid but is not whitelisted.

src/routes/hooks/hooks-cache.spec.ts
```diff
@@ -308,11 +161,8 @@ describe.skip('Post Hook Events for Cache (Unit)', () => {
       }
     });
 
-    await request(app.getHttpServer())
-      .post(`/hooks/events`)
-      .set('Authorization', `Basic ${authToken}`)
-      .send(data)
-      .expect(202);
+    const cb = queuesApiService.subscribe.mock.calls[0][1];
```
PooyaRaki, line 164: Question: Is there a clearer way to access the queue callback? I’m finding it difficult to understand what indices 0 and 1 represent.

src/routes/hooks/hooks-cache.spec.ts
iamacook: As discussed, there is no easy way to test this and, considering other tests cover this, we will exclude this.

src/routes/hooks/hooks-cache.spec.ts
iamacook: I added a helper to simplify this in e344ba72a663a5bb127a6ab8ad3ec2b550497cb8.

PR #2216, opened 2024-12-19
Trigger notification dispatch tests by queue

iamacook:
## Summary

Our notifications are dispatched in accordance with queue events. Most of the test coverage was linked to our previous POST hooks, and therefore skipped.

This converts the tests to trigger via queue instead of POST calls, unskipping them.

## Changes

- Unskip notification dispatch tests
- Modify tests to trigger via queue

### DIFF
--- a/src/routes/hooks/hooks-notifications.spec.ts
+++ b/src/routes/hooks/hooks-notifications.spec.ts
@@ -64,6 +64,13 @@ import {
 } from '@/routes/hooks/entities/__tests__/delegate-events.builder';
 import type { ConsumeMessage } from 'amqplib';
 
+function getSubscriptionCallback(
+  queuesApiService: jest.MockedObjectDeep<IQueuesApiService>,
+): (msg: ConsumeMessage) => Promise<void> {
+  // First call, second argument
+  return queuesApiService.subscribe.mock.calls[0][1];
+}
+
 // TODO: Migrate to E2E tests as TransactionEventType events are already being received via queue.
 describe('Hook Events for Notifications (Unit)', () => {
   let app: INestApplication<Server>;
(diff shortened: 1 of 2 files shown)


src/routes/hooks/hooks-notifications.spec.ts
```diff
@@ -133,13 +139,14 @@ describe.skip('Post Hook Events for Notifications (Unit)', () => {
       newConfirmationEventBuilder().build(),
       newMessageConfirmationEventBuilder().build(),
       safeCreatedEventBuilder().build(),
+      reorgDetectedEventBuilder().build(),
+      newDelegateEventBuilder().build(),
+      updatedDelegateEventBuilder().build(),
+      deletedDelegateEventBuilder().build(),
```
iamacook, line 145: These were added as they were missing.

src/routes/hooks/hooks-notifications.spec.ts
```diff
@@ -2017,7 +2024,7 @@ describe.skip('Post Hook Events for Notifications (Unit)', () => {
       }
     });
 
-    pushNotificationsApi.enqueueNotification
+    notificationsRepository.enqueueNotification
```
iamacook, line 2027: As we are mocking the repository, it's's not possible to test this as the logic is inside the method.

iamacook: I moved this particular test to a new suite in 963a5557b8dc89b9e21f710b889185523b5d2048. It doesn't mock the repository anymore, but spies on it instead.

We could convert the existing tests to also rely on spying but it goes against the pattern of the project. What do you think?

PooyaRaki: I think it's Ok either way. Let's merge your current implementation.

PooyaRaki (approved):

PR #2223, opened 2024-12-27
Feat: transaction note from tx origin

katspaugh:
## Summary
This PR adds the decoding of multi-sig transaction notes from the transaction-service's `origin` field.
Needed for https://github.com/safe-global/safe-wallet-monorepo/pull/4693

## Changes

* Add a mapper that parses the `origin` string and extracts a note from it
* Add a new top-level `TransactionDetails` property `note` that is sent to the client (optional prop)
* Unit tests for the mapper

### DIFF
--- a/src/routes/transactions/entities/transaction-details/transaction-details.entity.ts
+++ b/src/routes/transactions/entities/transaction-details/transaction-details.entity.ts
@@ -47,5 +47,5 @@ export class TransactionDetails {
   @ApiPropertyOptional({ type: SafeAppInfo, nullable: true })
   safeAppInfo!: SafeAppInfo | null;
   @ApiPropertyOptional({ type: String, nullable: true })
-  note?: string;
+  note!: string | null;
 }
--- a/src/routes/transactions/entities/transaction-details/transaction-details.entity.ts
+++ b/src/routes/transactions/entities/transaction-details/transaction-details.entity.ts
@@ -46,4 +46,6 @@ export class TransactionDetails {
   txHash!: string | null;
   @ApiPropertyOptional({ type: SafeAppInfo, nullable: true })
   safeAppInfo!: SafeAppInfo | null;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  note?: string;
 }
--- a/src/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper.spec.ts
+++ b/src/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper.spec.ts
@@ -1,3 +1,4 @@
+import { faker } from '@faker-js/faker';
 import { multisigTransactionBuilder } from '@/domain/safe/entities/__tests__/multisig-transaction.builder';
 import { MultisigTransactionNoteMapper } from '@/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper';
 
@@ -5,13 +6,17 @@ const mapper = new MultisigTransactionNoteMapper();
 
 describe('Multisig Transaction note mapper (Unit)', () => {
   it('should parse transaction `origin` and return a note', () => {
+    const noteText = faker.lorem.sentence();
     const transaction = multisigTransactionBuilder()
-      .with('origin', '{"name":"{\\"note\\":\\"This is a note\\"}"}')
+      .with(
+        'origin',
+        JSON.stringify({ name: JSON.stringify({ note: noteText }) }),
+      )
       .build();
 
     const note = mapper.mapTxNote(transaction);
 
-    expect(note).toBe('This is a note');
+    expect(note).toBe(noteText);
   });
 
   it('should return undefined if `origin` is not a valid JSON', () => {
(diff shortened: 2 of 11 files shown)


src/routes/transactions/entities/transaction-details/transaction-details.entity.ts
```diff
@@ -46,4 +46,6 @@ export class TransactionDetails {
   txHash!: string | null;
   @ApiPropertyOptional({ type: SafeAppInfo, nullable: true })
   safeAppInfo!: SafeAppInfo | null;
+  @ApiPropertyOptional({ type: String, nullable: true })
+  note?: string;
```
PooyaRaki, line 50: The question mark indicates that the field can be undefined if the value is not provided. However, to ensure the value is explicitly null and for consistency, we should define it as: `note!: string | null;`

src/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper.spec.ts
```diff
@@ -0,0 +1,46 @@
+import { multisigTransactionBuilder } from '@/domain/safe/entities/__tests__/multisig-transaction.builder';
+import { MultisigTransactionNoteMapper } from '@/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper';
+
+const mapper = new MultisigTransactionNoteMapper();
+
+describe('Multisig Transaction note mapper (Unit)', () => {
+  it('should parse transaction `origin` and return a note', () => {
+    const transaction = multisigTransactionBuilder()
+      .with('origin', '{"name":"{\\"note\\":\\"This is a note\\"}"}')
```
PooyaRaki, line 9: Nit: While it might not add significant value here, for the sake of consistency, I would use `faker.js` to generate input data for my tests.

src/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper.ts
```diff
@@ -0,0 +1,23 @@
...
+export class MultisigTransactionNoteMapper {
+  mapTxNote(transaction: MultisigTransaction): string | undefined {
+    let note: string | undefined;
+
+    if (transaction.origin) {
+      try {
+        const origin = JSON.parse(transaction.origin);
+        const parsedName = origin.name && JSON.parse(String(origin.name));
+        if (typeof parsedName.note === 'string') {
+          note = parsedName.note;
```
PooyaRaki, line 14: Nit:
```suggestion
          return parsedName.note;
```

src/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper.ts
```diff
@@ -0,0 +1,23 @@
...
+        }
+      } catch {
+        // Ignore, no note
+      }
+    }
+
+    return note;
```
PooyaRaki, line 21: Nit:
```suggestion
    return undefined;
```

src/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper.ts
```diff
@@ -0,0 +1,23 @@
+import { Injectable } from '@nestjs/common';
+import { MultisigTransaction } from '@/domain/safe/entities/multisig-transaction.entity';
+
+@Injectable()
+export class MultisigTransactionNoteMapper {
+  mapTxNote(transaction: MultisigTransaction): string | undefined {
+    let note: string | undefined;
```
PooyaRaki, line 7:
```suggestion
```

src/routes/transactions/entities/transaction-details/transaction-details.entity.ts
katspaugh: I've made the change but IMHO this convention is a cargo cult that started with the original Rust CGW and the Python-based transaction-service not having an undefined type that we blindly carried it over to this JavaScript project. There's very little value in sending these empty payloads to the client and having to write non-idiomatic null-checking code for them.

PooyaRaki: @katspaugh Thanks for updating the code! I agree that we can revisit it later, but let’s create a task and address it in a separate PR.

src/routes/transactions/mappers/multisig-transactions/multisig-transaction-note.mapper.spec.ts
```diff
@@ -0,0 +1,51 @@
...
+      .build();
+
+    const note = mapper.mapTxNote(transaction);
+
+    expect(note).toBeNull();
+  });
+
+  it('should return undefined if `origin` does not contain a note', () => {
+    const transaction = multisigTransactionBuilder()
+      .with('origin', '{"url":"uniswap.org","name":"Uniswap"}')
```
PooyaRaki, line 34: Nit: Here we could also use fakers.

PooyaRaki (approved):

katspaugh:
Thanks @PooyaRaki! I'll now merge it so that we can test it on staging together with the linked frontend PR.

PooyaRaki (approved):

Issue #1924, opened 2024-09-13
Improve error handling


PooyaRaki:
Our current error handling implementation has room for improvement in providing clear and consistent error messages to the client. Enhancing this aspect will help in troubleshooting issues more effectively and improve overall user experience.

**Objectives:****  <br>****Review and Refactor Error Handling Logic**: Assess the existing error handling mechanisms to identify areas where error processing and communication can be improved.  <br>**Testing and Validation**: Add comprehensive tests to validate the new error handling behavior and confirm that the improvements address the current shortcomings.

**Acceptance Criteria:**

* All error responses follow the standardized format.
* The refactoring does not introduce breaking changes for existing clients.
* Error messages are clear, actionable, and provide sufficient information for debugging.
  * Validation errors should return 4XX for client submitted data.
* Tests cover a variety of error scenarios and confirm correct behavior.

**Example**  <br>If an env. var. is not set, calling `getOrThrow` on the configuration service will thrown an [unknown error](<https://app.datadoghq.eu/logs?query=status%3Aerror%20service%3Astaging-safe-client-nest&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AgAAAZIAPFV-HX0tiAAAAAAAAAAYAAAAAEFaSUFQR2tqQUFBOTJ6ZUJsUWtvWWdBSwAAACQAAAAAMDE5MjAwM2MtOTE1Ni00YzU0LWI5OTktNTcwMjczYmYzOWI2&fromUser=true&index=devstaging&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1726494142100&to_ts=1726580542100&live=true>) that is otherwise hard to debug.

---

<details><summary>🤖 Refined and Investigated by Claude — 2026-07-14</summary>

**Session:** Wallet Hygiene cloud routine, run 2026-07-14

**Labels added:** Tech → Backend

**Investigation notes:**

[...]

PR #2227, opened 2024-12-31
Ensure correct status code is returned in logs

PooyaRaki:
## Summary
This PR ensures that the correct status codes are returned and logged. Previously, an incorrect status code might have been recorded in the logs, which could lead to confusion or misinterpretation during debugging or monitoring.
This change improves the reliability and accuracy of application logs, aiding in better diagnostics and monitoring.

https://github.com/safe-global/safe-client-gateway/issues/1924

## Changes
- Updated the logic to accurately capture and log the status codes returned by the application.
- Added tests to verify that the correct status codes are now logged in various scenarios

### DIFF
--- a/src/routes/common/interceptors/route-logger.interceptor.spec.ts
+++ b/src/routes/common/interceptors/route-logger.interceptor.spec.ts
@@ -16,6 +16,13 @@ import { Server } from 'net';
 import { ValidationPipe } from '@/validation/pipes/validation.pipe';
 import { NumericStringSchema } from '@/validation/entities/schemas/numeric-string.schema';
 
+// We expect 500 instead of the status code of the DataSourceError
+// The reason is that this test webserver does not have logic to map
+// DataSourceErrors to HTTP responses (it is not the goal of this test)
+// The goal of the test is to test that we are logging correctly
+// (see expects below)
+const expectedDatasourceErrorCode = 500;
+
 const mockLoggingService: jest.MockedObjectDeep<ILoggingService> = {
   info: jest.fn(),
   debug: jest.fn(),
@@ -23,7 +30,7 @@ const mockLoggingService: jest.MockedObjectDeep<ILoggingService> = {
   warn: jest.fn(),
 };
 
-class errorWithCode extends Error {
+class ErrorWithCode extends Error {
   private readonly code: number;
 
   public constructor(message: string, code: number) {
@@ -57,12 +64,12 @@ class TestController {
 
   @Get('error-level-info-with-code')
   errorLevelInfoWithCode(): void {
-    throw new errorWithCode('error', 430);
+    throw new ErrorWithCode('error', 430);
   }
 
   @Get('error-level-error-with-code')
   errorLevelErrorWithCode(): void {
-    throw new errorWithCode('error', 530);
+    throw new ErrorWithCode('error', 530);
   }
 
   @Get('server-error-non-http')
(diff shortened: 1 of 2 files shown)


src/routes/common/interceptors/route-logger.interceptor.ts
```diff
@@ -67,6 +69,10 @@ export class RouteLoggerInterceptor implements NestInterceptor {
       statusCode = error.getStatus();
     } else if (error instanceof DataSourceError) {
       statusCode = error.code ?? HttpStatus.INTERNAL_SERVER_ERROR;
+    } else if (error instanceof ZodErrorWithCode) {
+      statusCode = error.code;
+    } else if ('code' in error && isNumber(error.code)) {
+      statusCode = error.code;
```
iamacook, line 75: Nit: we could simplify this by checking for the existence of `code` and assigning it on line 77, e.g.

```ts
statusCode = 'code' in error ? Number(error.code) : HttpStatus.INTERNAL_SERVER_ERROR
```

src/routes/common/interceptors/route-logger.interceptor.spec.ts
```diff
@@ -20,6 +23,15 @@ const mockLoggingService: jest.MockedObjectDeep<ILoggingService> = {
   warn: jest.fn(),
 };
 
+class errorWithCode extends Error {
```
iamacook, line 26:
```suggestion
class ErrorWithCode extends Error {
```

src/routes/common/interceptors/route-logger.interceptor.spec.ts
```diff
@@ -124,6 +154,34 @@ describe('RouteLoggerInterceptor tests', () => {
...
 
+  it('500 Any error triggers error level', async () => {
+    await request(app.getHttpServer())
+      .get('/test/error-level-error-with-code')
+      // We expect 500 instead of the status code of the DataSourceError
+      // The reason is that this test webserver does not have logic to map
+      // DataSourceErrors to HTTP responses (it is not the goal of this test)
+      // The goal of the test is to test that we are logging correctly
+      // (see expects below)
+      .expect(500);
```
iamacook, line 165: Nit: we could move this expected value to a constant with a singular comment to save repetition. What do you think?

src/routes/common/interceptors/route-logger.interceptor.ts
PooyaRaki: I’ve been considering this, but some errors, like `ZodErrors`, might have alphanumeric codes. While we already handle `ZodErrors` correctly, I want to ensure that the error code is strictly numeric in any case.

src/routes/common/interceptors/route-logger.interceptor.spec.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/2227/commits/128b5b96e06745fd2123e74f5f97387e8d4d7d92

src/routes/common/interceptors/route-logger.interceptor.spec.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/2227/commits/128b5b96e06745fd2123e74f5f97387e8d4d7d92

src/routes/common/interceptors/route-logger.interceptor.ts
iamacook: Good point (just looked over [`ZodError`s](https://github.com/colinhacks/zod/blob/main/src/ZodError.ts)). Nonethelesss, we could add the `isNumber` check in the suggestion?

iamacook: If you prefer to keep a condition, we don't need lines 72/73.

iamacook: After discussion, we'lll remain explicit with conditions for each error type.

iamacook (approved):

PR #2228, opened 2024-12-31
Enhance Logs with Color-Coding and Improved Readability

PooyaRaki:
## Summary
This PR introduces color-coding and formatting to logs to improve their readability and make debugging more efficient.

## Changes
- Add prretyPrint and color to log messages
- Introduced the `LOG_PRETTY_COLORIZE` environment variable to toggle pretty logs on or off, allowing flexibility in different environments.

### DIFF
--- a/.env.sample
+++ b/.env.sample
@@ -136,7 +136,10 @@
 # JWT_SECRET=
 
 # Log level for the service.
-# LOG_LEVEL=
+# LOG_LEVEL=q
+
+# Make logs prrety and colorize
+# LOG_PRETTY_COLORIZE=
 
 # Targeted Messaging
 # TARGETED_MESSAGING_FILE_STORAGE_TYPE=
--- a/.env.sample
+++ b/.env.sample
@@ -136,7 +136,7 @@
 # JWT_SECRET=
 
 # Log level for the service.
-# LOG_LEVEL=q
+# LOG_LEVEL=
 
 # Make logs prrety and colorize
 # LOG_PRETTY_COLORIZE=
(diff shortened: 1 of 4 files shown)


.env.sample
```diff
@@ -136,7 +136,10 @@
 # JWT_SECRET=
 
 # Log level for the service.
-# LOG_LEVEL=
+# LOG_LEVEL=q
```
iamacook, line 139: I think this was added by mistake.

src/logging/logging.module.ts
```diff
@@ -30,9 +30,13 @@ const LoggerTransports = Symbol('LoggerTransports');
 function winstonTransportsFactory(
   configurationService: IConfigurationService,
 ): Transport[] | Transport {
+  const prettyColorize =
+    configurationService.getOrThrow<boolean>('log.prettyColorize');
```
iamacook, line 34: Afaiu, this is purely for local development benefit so I don't think a feature flag is necessary. We can detect that the project is running locally by checking `CGW_ENV` is not `'production'` or `'staging'`. What do you think?

PooyaRaki: What do you think about starting using the standard `NODE_ENV`?

.env.sample
PooyaRaki: Yes, that was a mistake, fixed here
https://github.com/safe-global/safe-client-gateway/pull/2228/commits/5ae86dbfe95416b883024beafa1052caadf3f5a1

src/logging/logging.module.ts
PooyaRaki: Another use case is the ability to temporarily enable this feature for debugging purposes on the staging server. But I wouldn't take that into account for now as it's highly unlikely to be enabled on staging.

iamacook: We wouldn't need to enable this on staging as Datadog prettifies on our behalf.

Considering we have other log-specific config in place (`level` and `silent`), keeping it where it is makes sense - I initially thought it was a feature flag. I'll resolve this.

iamacook (approved):

iamacook (approved):

PR #2238, opened 2025-01-06
Remove `FakeCacheService` and `TestCacheModule`

### LINKED ISSUE #1960: [Tech Debt] Remove FakeCacheService
### Description

`FakeCacheService` was introduced to mimic the behavior of a Cache layer while running unit tests. One of its main goals was to avoid the need for having a running cache instance (Redis) while running the unit tests. This is not true anymore since the CGW does require a running Redis instance for some test suites (e.g.: `redis.cache.service.spec.ts`).

Moreover, there are no strict guarantees that the actual Redis behavior is mimicked by `FakeCacheService`, making this class a risk since bugs can be introduced unnoticed while using this class on tests.

### Requirements

- Remove `FakeCacheService`.
- Replace `FakeCacheService` usages on test classes by `RedisCacheService`.

### Additional information

_No response_

iamacook:
## Summary

Resolves #1960

We mock our (Redis-based) cache service throughout our tests, despite Redis being a requirement.

This removes our `FakeCacheService` and `TestCacheModule` which builds on top of it, updating the tests accordingly.

## Changes

tbd


iamacook:
I'm going to close this, with the intention to use it as a reference in the future. This implementation relies on the same Redis instance across tests, making proper sanitisation challenging, with no straightforward approach to achieve it. Additionally, there are complications with Redis client/service instantiation when using faked timers.

PR #2244, opened 2025-01-07
Add `to` to pending/executed multisig events

iamacook:
## Summary

A new `to` field is to-be-added to the pending/executed multisig events [on the Transaction Service](https://github.com/safe-global/safe-events-service/pull/331). This adds them to the codebase, validating them and propagating them across tests.

## Changes

- Add `to` to `TransactionEventType.PENDING_MULTISIG_TRANSACTION` and `TransactionEventType.EXECUTED_MULTISIG_TRANSACTION` `type` events
- Update tests accordingly

### DIFF
--- a/src/routes/hooks/entities/schemas/__tests__/executed-transaction.schema.spec.ts
+++ b/src/routes/hooks/entities/schemas/__tests__/executed-transaction.schema.spec.ts
@@ -42,44 +42,51 @@ describe('ExecutedTransactionEventSchema', () => {
     );
   });
 
-  it('should not allow a non-address address', () => {
-    const executedTransactionEvent = executedTransactionEventBuilder()
-      .with('address', faker.string.sample() as `0x${string}`)
-      .build();
+  it.each(['to' as const, 'address' as const])(
+    'should not allow a non-address %s',
+    (field) => {
+      const executedTransactionEvent = executedTransactionEventBuilder()
+        .with(field, faker.string.sample() as `0x${string}`)
+        .build();
 
-    const result = ExecutedTransactionEventSchema.safeParse(
-      executedTransactionEvent,
-    );
+      const result = ExecutedTransactionEventSchema.safeParse(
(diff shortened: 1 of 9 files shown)


src/routes/hooks/entities/schemas/__tests__/pending-transaction.schema.spec.ts
```diff
@@ -42,44 +42,51 @@ describe('PendingTransactionEventSchema', () => {
     );
   });
 
-  it('should not allow a non-address address', () => {
-    const pendingTransactionEvent = pendingTransactionEventBuilder()
-      .with('address', faker.string.sample() as `0x${string}`)
-      .build();
+  it.each(['to' as const, 'address' as const])(
```
PooyaRaki, line 45: Nit: To improve test reliability, we should aim to minimize the use of dynamically generated test inputs. I noticed that this test suite already contains quite a few. While it’s not necessary to revise all existing instances, I believe it would be better to avoid adding new ones whenever possible.

src/routes/hooks/entities/schemas/__tests__/executed-transaction.schema.spec.ts
```diff
@@ -42,44 +42,51 @@ describe('ExecutedTransactionEventSchema', () => {
     );
   });
 
-  it('should not allow a non-address address', () => {
-    const executedTransactionEvent = executedTransactionEventBuilder()
-      .with('address', faker.string.sample() as `0x${string}`)
-      .build();
+  it.each(['to' as const, 'address' as const])(
```
PooyaRaki, line 45: Nit: To improve test reliability, we should aim to minimize the use of dynamically generated test inputs. I noticed that this test suite already contains quite a few. While it’s not necessary to revise all existing instances, I believe it would be better to avoid adding new ones whenever possible.

PooyaRaki (approved):

PR #2245, opened 2025-01-07
Return relevant error code in logger interceptor

PooyaRaki:
## Summary
his PR ensures that the correct status codes are returned and logged. Previously, an incorrect status code might have been recorded in the logs, which could lead to confusion or misinterpretation during debugging or monitoring.
This change improves the reliability and accuracy of application logs, aiding in better diagnostics and monitoring.

## Changes
Updated the logic to accurately capture and log the status codes returned by the application.


src/routes/common/interceptors/route-logger.interceptor.ts
```diff
@@ -64,17 +63,13 @@ export class RouteLoggerInterceptor implements NestInterceptor {
...
-    } else if ('code' in error && isNumber(error.code)) {
+    let statusCode: number = HttpStatus.INTERNAL_SERVER_ERROR;
+    if ('code' in error && isNumber(error.code)) {
       statusCode = error.code;
-    } else {
-      statusCode = HttpStatus.INTERNAL_SERVER_ERROR;
+    } else if (error instanceof HttpException) {
+      statusCode = error.getStatus();
+    } else if (error instanceof ZodError) {
+      statusCode = HttpStatus.UNPROCESSABLE_ENTITY;
```
iamacook, line 72: Lines 66-68 makes this redundant.

PooyaRaki: Line 66 is the default value i.e. `INTERNAL_SERVER_ERROR`. Unfortunately, `ZodError`s do not have a code. And since we use the `parse` method we cannot easily convert them to `ZodErrorWithCode`.

iamacook: I missed the error diff. 🙈 (I though it was still 500).

I don't think a client error (422) makes sense for `ZodError`s as they are generally thrown based on datasources. What do you think about 502?

PooyaRaki: Very good point 👍 . But what if we use them to verify database results in the future for example?
For now, I totally agree that 502 is better as we use the `parse` method only to verify the data fetched from the datasource.

iamacook: > what if we use them to verify database results in the future for example?

We should parse, throwing a `ZodErrorWithCode`.

PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/2245/commits/81b7c99c98b439dd2ace97a9e4ffeaae6d81fbf9
Also added a comment about why we are throwing 422

iamacook (approved):

src/routes/common/interceptors/route-logger.interceptor.ts
PooyaRaki: Since the PR is approved, I'm going to resolve this thread.

PR #2248, opened 2025-01-13
Add `IPortfolio` datasource with Octav implementation

iamacook:
## Summary

This adds the datasource layer for the forthcoming portfolio feature, implementing Octav.

Note: the `Portfolio` entity will be populated after designs are confirmed and a client-side scaffolding has been completed as not every value from the API if/will be used.

## Changes

- Adds `IPortfolio` interface
- Adds blank `Portfolio` entity
- Implements `IPortfolio` with Octav's API
- Adds validated with URI and API keys config.

### DIFF
--- a/src/config/entities/schemas/configuration.schema.ts
+++ b/src/config/entities/schemas/configuration.schema.ts
@@ -23,6 +23,7 @@ export const RootConfigurationSchema = z
     INFURA_API_KEY: z.string(),
     JWT_ISSUER: z.string(),
     JWT_SECRET: z.string(),
+    PORTFOLIO_API_KEY: z.string(),
     PUSH_NOTIFICATIONS_API_PROJECT: z.string(),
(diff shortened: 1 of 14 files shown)


src/config/entities/schemas/configuration.schema.ts
```diff
@@ -23,6 +23,7 @@ export const RootConfigurationSchema = z
     INFURA_API_KEY: z.string(),
     JWT_ISSUER: z.string(),
     JWT_SECRET: z.string(),
+    PORTFOLIO_API_KEY: z.string(),
```
iamacook, line 26: We will need to add dummy values to the environments before merging this.

src/domain/portfolio/entities/portfolio.entity.ts
```diff
@@ -0,0 +1,5 @@
+import { z } from 'zod';
+
+export const PortfolioSchema = z.unknown();
```
iamacook, line 3: This will be populated after scaffolding the UI as we may not use every field from the API.

PR #2258, opened 2025-01-14
Remove validation of `fullFeeAmount`

iamacook:
## Summary

We currently validate the presence of `fullFeeAmount` in `Order`s. This field is [no longer returned by CoW](https://github.com/cowprotocol/services/pull/3223) and, as such, breaks Swap decoding.

As we are not reliant on the field, it can be safely removed. As such, validation is no longer included and all presence of the field removed accross the codebase.

Note: `solverFee` has also been removed from mock data as that is also not used (though not validated)

## Changes

- Remove validation of `fullFeeAmount`
- Remove mock `solveFee` values
- Propagate across codebase


PooyaRaki (approved):
LGTM

PR #2260, opened 2025-01-15
Add `IPortfolioRepository` domain and implementation

iamacook:
## Summary

This adds the domain layer for the forthcoming portfolio feature.

A new `IPortfolioRepositry`, including a method to get the portfolio for a specified address. The implementation of which fetches a portfolio and validates it according to the Octav-based `PortfolioSchema`.

## Changes

- Add `IPortfolioRepositry` with implementation
- Implement and test `PortfolioSchema`
- Add builders for `Portfolio` and constituent entities

PR #2261, opened 2025-01-15
Add `/v1/chains/:chainId/safes/:safeAddress/portfolio` route

iamacook:
⚠️ Branches off #2260

## Summary

This adds the route layer for the forthcoming portfolio feature.

A new feature flag (`features.portfolio`) enabled `RoutesController` has been added with the following route:

- `/v1/chains/:chainId/safes/:safeAddress/portfolio`

The return type of which is a newly chain-specific mapped routes list, mapped by the `PortfolioMapper`.

## Changes

- Add test covered `PortfolioMapper` with mapped entities
- Add test covered `PortfolioController`, with associated service
- Include controller and service in a `PortfolioModule` behind a feature flag


src/routes/portfolio/mappers/portfolio.mapper.spec.ts
```diff
@@ -0,0 +1 @@
+it.todo('PortfolioMapper');
```
iamacook, line 1: The mapping is being reference in local development. After confirmed, tests will be added and this marked for review.

src/routes/portfolio/portfolio.controller.spec.ts
```diff
@@ -0,0 +1 @@
+it.todo('PortfolioController');
```
iamacook, line 1: As before regarding local development before tests.

src/domain/portfolio/entities/portfolio.entity.ts
```diff
@@ -84,7 +85,7 @@ export const ProtocolPositionType = [
 ] as const;
 
 export const ProtocolPositionsSchema = z.record(
-  z.enum(ProtocolPositionType),
+  z.enum([...ProtocolPositionType, 'UNKNOWN']).catch('UNKNOWN'),
```
iamacook, line 88: I think we should add logging here but I'm not sure on the best approach as we'd need to inject the logging service into the entity. To keep the footprint focused, I will address this in a follow up PR.

4player90 (approved):

iamacook:
I'm going to close this as we are awaiting clarity on provider details. We will reopen this if/when necessary. cc @kirkkonen

PR #2262, opened 2025-01-15
Add `User` entity

usame-algan:
## Summary

Adds a new `Users` table 

## Changes

- Create `User` entity
- Create `userBuilder` to be used in future tests


migrations/1733761866546-cleanNotificationData.ts
usame-algan, line 1: Looks like this was removed after running the migration. Should we restore it in this PR and remove it later?

.envrc
```diff
@@ -0,0 +1 @@
+dotenv .env
```
iamacook, line 1: Not sure if you meant to include this?

migrations/1733761866546-cleanNotificationData.ts
iamacook: This should not be removed.

src/app.module.ts
```diff
@@ -77,6 +78,7 @@ export class AppModule implements NestModule {
     return {
       module: AppModule,
       imports: [
+        UserRepositoryModule,
```
iamacook, line 81: This shouldn't yet be imported as you are working on the datasource. The `AppModule` should only import route layer modules.

src/datasources/users/entities/__tests__/users.entity.db.builder.ts
```diff
@@ -0,0 +1,14 @@
...
+import type { UUID } from 'crypto';
+import type { User } from '@/datasources/users/entities/users.entity.db';
+import { UserStatus } from '@/domain/users/entities/user.entity';
+
+export function userBuilder(): IBuilder<User> {
+  return new Builder<User>()
+    .with('id', faker.string.uuid() as UUID)
+    .with('status', faker.helpers.enumValue(UserStatus))
+    .with('created_at', new Date())
+    .with('updated_at', new Date());
```
iamacook, line 13: Nit: we could use faker to have different dates here: `created_at` in the past and `updated_at` thereafter.

src/domain/users/users.repository.module.ts
```diff
@@ -0,0 +1,9 @@
+import { Module } from '@nestjs/common';
+import { TypeOrmModule } from '@nestjs/typeorm';
+import { User } from '@/datasources/users/entities/users.entity.db';
+import { PostgresDatabaseModuleV2 } from '@/datasources/db/v2/postgres-database.module';
+
+@Module({
+  imports: [PostgresDatabaseModuleV2, TypeOrmModule.forFeature([User])],
+})
+export class UserRepositoryModule {}
```
iamacook, line 9: I would move this to a separate PR in which the repository is implemented, e.g. with methods for retrieving `User`s.

usame-algan: This is needed to generate the migration file and create the database table. Would you suggest also doing that in a separate PR?

iamacook: Yes. I would scaffold the structure there as well, e.g. create a controller and domain module (that can be empty) as the repository would not be imported here, rather via the aforementioned.

.envrc
usame-algan: It is needed for direnv to work so it loads all the .env variables into the terminal whenever we are in the root folder. Otherwise the configuration.ts is not able to detect `process.env` variables and it breaks locally. I think there is also a way in NestJS to do this via the `ConfigModule` (https://docs.nestjs.com/techniques/configuration). Maybe something we can look into separately cc @PooyaRaki

iamacook: Is this [feature-related](https://github.com/safe-global/safe-client-gateway/blob/main/src/app.module.ts#L65-L67)? If this is a general issue (and not related to the entity in question), this should be done in a separate PR.

usame-algan: Its not feature related but it doesn't work locally unless we hard-code the config which is cumbersome. This solves it for whoever is using direnv. The .envrc file was already there but empty. How are you handling .env variables locally to make them work?

iamacook (approved):

src/datasources/users/entities/__tests__/users.entity.db.builder.ts
usame-algan: Lets do it once we start writing tests with this builder

src/datasources/users/entities/users.entity.db.ts
```diff
@@ -0,0 +1,37 @@
...
+  @Column({
+    type: 'timestamp with time zone',
+    default: () => 'CURRENT_TIMESTAMP',
+  })
+  created_at!: Date;
+
+  @Column({
+    type: 'timestamp with time zone',
+    default: () => 'CURRENT_TIMESTAMP',
+    onUpdate: 'CURRENT_TIMESTAMP',
```
PooyaRaki, line 34: It won't update the timestamp, please have a look at notification code for example. In PostgreSQL you need to create a database function and trigger to update the timestamp on update.

PooyaRaki: FYI:
https://github.com/safe-global/safe-client-gateway/blob/main/migrations/1727701600427-update_timestamp_trigger.ts
https://github.com/safe-global/safe-client-gateway/blob/main/migrations/1727701873513-notification_update_updated_at.ts

You need to write the queries manually, but you can reuse the ones from the notification by simply updating the names.

src/datasources/users/entities/users.entity.db.ts
```diff
@@ -0,0 +1,37 @@
...
+@Unique('id', ['id'])
+export class User implements z.infer<typeof UserSchema> {
+  @PrimaryGeneratedColumn('uuid')
+  id!: UUID;
+
+  @Column({
+    type: 'enum',
+    enum: UserStatus,
+  })
+  status!: UserStatus;
```
PooyaRaki, line 23: Since we will search based on user status, it should be stored as an integer. Searching an integer column is significantly more efficient than searching a text column.

PR #2264, opened 2025-01-15
Add `wallet` entity

jmealy:
## Summary
Adds a new entity for Wallet. Each entry represents a signer wallet that can be used to sign in for a User.

## Changes
- Creates `Wallet` entity
- Creates walletBuilder

### DIFF
--- a/src/datasources/users/entities/__tests__/wallets.entity.db.builder.ts
+++ b/src/datasources/users/entities/__tests__/wallets.entity.db.builder.ts
@@ -3,12 +3,13 @@ import { Builder } from '@/__tests__/builder';
 import { faker } from '@faker-js/faker/.';
 import type { Wallet } from '@/datasources/users/entities/wallets.entity.db';
 import { userBuilder } from '@/datasources/users/entities/__tests__/users.entity.db.builder';
+import { getAddress } from 'viem';
 
 export function walletBuilder(): IBuilder<Wallet> {
   return new Builder<Wallet>()
     .with('id', faker.number.int())
     .with('user', userBuilder().build())
-    .with('address', faker.string.hexadecimal({ length: 42 }) as `0x${string}`)
+    .with('address', getAddress(faker.finance.ethereumAddress()))
     .with('created_at', new Date())
     .with('updated_at', new Date());
 }
--- a/src/datasources/users/entities/__tests__/wallets.entity.db.builder.ts
+++ b/src/datasources/users/entities/__tests__/wallets.entity.db.builder.ts
@@ -1,0 +1,14 @@
+import type { IBuilder } from '@/__tests__/builder';
+import { Builder } from '@/__tests__/builder';
+import { faker } from '@faker-js/faker/.';
+import type { UUID } from 'crypto';
+import type { Wallet } from '@/datasources/users/entities/wallets.entity.db';
+
+export function walletBuilder(): IBuilder<Wallet> {
+  return new Builder<Wallet>()
+    .with('id', faker.string.uuid() as UUID)
+    .with('user_id', faker.string.uuid() as UUID)
+    .with('address', faker.string.hexadecimal({ length: 42 }) as `0x${string}`)
+    .with('created_at', new Date())
+    .with('updated_at', new Date());
+}
--- a/src/datasources/users/entities/__tests__/wallets.entity.db.builder.ts
+++ b/src/datasources/users/entities/__tests__/wallets.entity.db.builder.ts
@@ -1,13 +1,13 @@
 import type { IBuilder } from '@/__tests__/builder';
 import { Builder } from '@/__tests__/builder';
 import { faker } from '@faker-js/faker/.';
-import type { UUID } from 'crypto';
 import type { Wallet } from '@/datasources/users/entities/wallets.entity.db';
+import { userBuilder } from '@/datasources/users/entities/__tests__/users.entity.db.builder';
 
 export function walletBuilder(): IBuilder<Wallet> {
   return new Builder<Wallet>()
-    .with('id', faker.string.uuid() as UUID)
-    .with('user_id', faker.string.uuid() as UUID)
+    .with('id', faker.number.int())
+    .with('user', userBuilder().build())
     .with('address', faker.string.hexadecimal({ length: 42 }) as `0x${string}`)
     .with('created_at', new Date())
     .with('updated_at', new Date());
--- a/src/datasources/users/entities/wallets.entity.db.ts
+++ b/src/datasources/users/entities/wallets.entity.db.ts
@@ -6,6 +6,7 @@ import {
   PrimaryGeneratedColumn,
   ManyToOne,
   JoinColumn,
+  Index,
 } from 'typeorm';
 import { z } from 'zod';
 import { getAddress } from 'viem';
@@ -17,7 +18,7 @@ export const WalletSchema = RowSchema.extend({
 });
 
 @Entity('wallets')
-@Unique(['user'])
+@Unique(['user', 'address'])
 export class Wallet implements z.infer<typeof WalletSchema> {
   @PrimaryGeneratedColumn()
   id!: number;
@@ -28,6 +29,7 @@ export class Wallet implements z.infer<typeof WalletSchema> {
   @JoinColumn({ name: 'user_id' })
   user!: User;
 
+  @Index({ unique: true })
   @Column({
     type: 'varchar',
     length: 42,
--- a/src/datasources/users/entities/wallets.entity.db.ts
+++ b/src/datasources/users/entities/wallets.entity.db.ts
@@ -1,12 +1,11 @@
 import { RowSchema } from '@/datasources/db/v1/entities/row.entity';
-import { UuidSchema } from '@/validation/entities/schemas/uuid.schema';
-import type { UUID } from 'crypto';
 import {
   Column,
   Entity,
   Unique,
   PrimaryGeneratedColumn,
   ManyToOne,
+  JoinColumn,
 } from 'typeorm';
 import { z } from 'zod';
 import { getAddress } from 'viem';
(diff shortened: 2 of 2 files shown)


src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -0,0 +1,60 @@
...
+  id: UuidSchema,
+  user_id: UuidSchema,
+  address: AddressSchema,
+});
+
+@Entity('wallets')
+@Unique('id', ['id'])
+export class Wallet implements z.infer<typeof WalletSchema> {
+  @PrimaryGeneratedColumn('uuid')
+  id!: UUID;
```
iamacook, line 26: Why do we want to use a UUID as the primary key? Are we going to expose this value? I'd suggest using an auto-incrementing value if not.

src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -0,0 +1,60 @@
...
+
+  @Column({ type: 'uuid' })
+  @ManyToOne(() => User, (user) => user.id, {
+    onDelete: 'CASCADE',
+  })
+  user_id!: UUID;
```
iamacook, line 32: Similar to the above, this might need changing in accordance with https://github.com/safe-global/safe-client-gateway/pull/2262#discussion_r1916946152.

src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -0,0 +1,60 @@
...
+    default: () => 'CURRENT_TIMESTAMP',
+  })
+  created_at!: Date;
+
+  @Column({
+    type: 'timestamp with time zone',
+    default: () => 'CURRENT_TIMESTAMP',
+    onUpdate: 'CURRENT_TIMESTAMP',
+  })
+  updated_at!: Date;
```
iamacook, line 59: https://github.com/safe-global/safe-client-gateway/pull/2262#discussion_r1916954522

src/datasources/users/entities/wallets.entity.db.ts
jmealy: With auto incrementing IDs we would expose how many accounts have been created, but I'm not sure if this matters

PooyaRaki: If you want to use both `integer` and `UUID`, you can make the primary key an integer and add a unique `uuid` column for cases where the `ID` shouldn’t be exposed. For foreign keys, stick to using the `integer` field instead of the `UUID`, since `UUID`s are bigger and can make indexes larger, which will slow things down.

src/datasources/users/entities/wallets.entity.db.ts
jmealy: changed to auto-incrementing id https://github.com/safe-global/safe-client-gateway/pull/2264/commits/11fb7054197d840a8030cb31ea5fc771d0e755c1

src/datasources/users/entities/wallets.entity.db.ts
jmealy: changed User to use auto incrementing ID as well: https://github.com/safe-global/safe-client-gateway/pull/2264/commits/11fb7054197d840a8030cb31ea5fc771d0e755c1

src/datasources/users/entities/__tests__/wallets.entity.db.builder.ts
```diff
@@ -0,0 +1,14 @@
...
+import { Builder } from '@/__tests__/builder';
+import { faker } from '@faker-js/faker/.';
+import type { Wallet } from '@/datasources/users/entities/wallets.entity.db';
+import { userBuilder } from '@/datasources/users/entities/__tests__/users.entity.db.builder';
+
+export function walletBuilder(): IBuilder<Wallet> {
+  return new Builder<Wallet>()
+    .with('id', faker.number.int())
+    .with('user', userBuilder().build())
+    .with('address', faker.string.hexadecimal({ length: 42 }) as `0x${string}`)
```
iamacook, line 11:
```suggestion
import { getAddress } from 'viem';

export function walletBuilder(): IBuilder<Wallet> {
  return new Builder<Wallet>()
    .with('id', faker.number.int())
    .with('user', userBuilder().build())
    .with('address', getAddress(faker.finance.ethereumAddress()))
```

jmealy: What's the benifit of `faker.string.hexadecimal({ length: 42 })` over `faker.finance.ethereumAddress()` ?

iamacook: There is no difference, but I favour `faker.finance.ethereumAddress` as it's more declarative.

Wrapping either in `getAddress` checksums the address and returns a `0x${string}` though. This [matches the entity](https://github.com/safe-global/safe-client-gateway/blob/11fb7054197d840a8030cb31ea5fc771d0e755c1/src/datasources/users/entities/wallets.entity.db.ts#L34-L41).

jmealy: sorry I read it the wrong way around, makes sense!

src/datasources/users/entities/wallets.entity.db.ts
PooyaRaki: @jmealy Since PostgreSQL does not have a direct `ON UPDATE` clause for columns, we should add a Trigger for it. Please refer to the Notification migration, as a similar approach is used there.

src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -0,0 +1,56 @@
...
+  @PrimaryGeneratedColumn()
+  id!: number;
+
+  @ManyToOne(() => User, (user) => user.id, {
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({ name: 'user_id' })
+  user!: User;
+
+  @Column({
```
PooyaRaki, line 31: Looks like we’ll be searching for wallets by their address. If that’s the case, I’d index this column to improve search performance.

src/datasources/users/entities/wallets.entity.db.ts
jmealy: yes understood 👍 but this PR is just for the entity, not the migration, so I have nowhere to add it yet. I can add it after this PR is merged and and a migration is generated

src/datasources/users/entities/wallets.entity.db.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2264/commits/acfafcf29c0aec34ad405eb30839910d18a9257d

src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -0,0 +1,58 @@
...
+import { getAddress } from 'viem';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { User } from '@/datasources/users/entities/users.entity.db';
+
+export const WalletSchema = RowSchema.extend({
+  address: AddressSchema,
+});
+
+@Entity('wallets')
+@Unique(['user', 'address'])
```
PooyaRaki, line 21: This will make the combination of the two unique.
```suggestion
@Unique(['user'])
@Unique( ['address'])
```

src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -0,0 +1,58 @@
...
+  @PrimaryGeneratedColumn()
+  id!: number;
+
+  @ManyToOne(() => User, (user) => user.id, {
+    onDelete: 'CASCADE',
+  })
+  @JoinColumn({ name: 'user_id' })
+  user!: User;
+
+  @Index({ unique: true })
```
PooyaRaki, line 32: Since it's been defined as a unique column above, the database sysmt will automatically create an index for it, therefor we don't need to add an `@index` decorator here.

src/datasources/users/entities/wallets.entity.db.ts
PooyaRaki: Ok, IMO it's better to add the migration in the same PR, but I'm ok if you do it in another PR.

src/datasources/users/entities/wallets.entity.db.ts
usame-algan: Would this mean that a user can't appear twice in different rows?
According to the [spec](https://www.notion.so/safe-global/Users-can-create-an-account-15e8180fe5738029ad5dc9141672bbd1?pvs=4#1628180fe5738002bf6ef8f793c6f6f4) a user can have multiple wallet entries so I think we need to make sure that the same user and wallet combination must be unique.

usame-algan: Although the current constraint would allow different users to have the same address. Maybe it just needs to be `@Unique(['address'])` then so that address is unique but users can appear multiple times in the table?

PooyaRaki: @usame-algan My comment was more about the usage of the `@unique` decorator. Sure, if the logic needs the address field to be the only unique column, we should remove `User` from the constraint.

src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -0,0 +1,56 @@
...
+  ManyToOne,
+  JoinColumn,
+} from 'typeorm';
+import { z } from 'zod';
+import { getAddress } from 'viem';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { User } from '@/datasources/users/entities/users.entity.db';
+
+export const WalletSchema = RowSchema.extend({
+  address: AddressSchema,
```
PooyaRaki, line 16: We should also add `User` to the schema.

src/datasources/users/entities/wallets.entity.db.ts
PooyaRaki: I'll leave it to @iamacook to resolve the conversation.

iamacook (approved):

PooyaRaki (approved):
LGTM

iamacook (approved):

PR #2266, opened 2025-01-16
Add `/v2/owners/:ownerAddress/safes` to gracefully handle errors

iamacook:
## Summary

When getting all Safes by owner (`/v1/owners/:ownerAddress/safes`), every Transaction Service is called. Should one or more throw, the request will as well.

This adds a new endpoint (`/v2/owners/:ownerAddress/safes`) which returns `null` for every chain which the Transaction Service threw for.

## Changes

- Add new scaffolding for v2 endpoint, deprecating previous.
- Wrap Transaction Service requests in `Promise.allSettled`.
- Construct all Safe lists from successful requests, logging for errorneous ones.


katspaugh (approved):
Nice. It's good that it still logs when it fails for a given chain.

src/domain/safe/safe.repository.ts
```diff
@@ -427,7 +427,8 @@ export class SafeRepository implements ISafeRepository {
     ownerAddress: `0x${string}`;
   }): Promise<{ [chainId: string]: Array<string> }> {
     const chains = await this.chainsRepository.getAllChains();
-    const allSafeLists = await Promise.all(
+    // Gracefully handle errors in case singular Transaction Service throws
```
PooyaRaki, line 430: Nit: The code is self-explanatory, I think we can remove the comment

PooyaRaki:
I believe we should either return all of a user’s Safes or none at all. As a user, if some of my Safes are missing from a request, I would assume that all have been returned and the missing ones are lost.
The relevance of the above comment depends on the specific use case of the endpoint.

katspaugh:
@PooyaRaki I think the proposed change is good for two reasons:
* In case of an accidental misconfiguration of one of the chains, the other chains will still work
* Isolating configs per chain will also help isolating issues – when the whole endpoint is down, you have to to triage from scratch

PooyaRaki:
@katspaugh I agree with you, but I think we should inform users when we’re unable to fetch data from a specific chain due to ongoing issues. Imagine a power user with assets on three different chains—if they see assets on two chains but not the third, they’d likely be confused. If we give them a heads-up about the issue and suggest trying again later, that’s fine. Otherwise, it could be misleading and might hurt their trust in the platform.

iamacook:
@PooyaRaki, I agree it's important to prioritise clarity for users. However, this approach aligns with the [existing behaviour](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/safes/safes.service.ts#L191-L193) of endpoints that fetch data from every Transaction Service.

> Imagine a power user with assets on three different chains—if they see assets on two chains but not the third, they’d likely be confused. If we give them a heads-up about the issue and suggest trying again later, that’s fine. Otherwise, it could be misleading and might hurt their trust in the platform.

As we're fetching data from every chain, if a particular Transaction Service throws, we lack clarity as to whether the user has Safes on that chain, so it's potentially confusing either way.

One alternative could be to return `null` for the affected chain(s) and explicitly display a warning that Safes couldn't be fetched there.

What are your thoughts? cc @katspaugh

src/domain/safe/safe.repository.ts
iamacook: Removed in 602a36172bb713810549dcddbce20ed883ec4378.

iamacook:
> Regarding returning `null`, what do we return if the user doesn’t have a Safe on a chain?

An empty array indicates no Safes are owned on a given chain when a request is successful. This PR, however, excludes the chain in question if the request fails.

I'd like to emphasise that returning `null` would strictly indicate that a request failed, leaving us uncertain whether Safes are (or aren't) owned on the affected chain.

> we should display a warning indicating that Safes couldn’t be loaded there. Regarding returning null, what do we return if the user doesn’t have a Safe on a chain

Let's see what @katspaugh thinks about this, as implementing such changes would require client updates, whereas this PR is intended as a quick fix.

PooyaRaki:
I’d be happy to approve it, provided we ensure that our users are properly notified about the error. As an end user, I wouldn’t want to see my assets disappear simply because certain chains are experiencing issues.

src/routes/owners/owners.controller.ts
```diff
@@ -23,12 +22,39 @@ export class OwnersController {
...
-  async getAllSafesByOwner(
+  @ApiOkResponse({
+    schema: {
+      type: 'object',
+      additionalProperties: {
+        type: 'array',
+        items: { type: 'string' },
+      },
+    },
+  })
```
iamacook, line 33: The types were previously incorrect.

This cannot be set in a class as the dynamic properties are top-level.

iamacook:
> An empty array if no safes or `null` if a chain threw sounds good. That'd be a breaking change so we'll have to adjust the frontend.

I've added a v2 endpoint implementing the above. Please lmk what you think, @PooyaRaki @katspaugh

src/domain/safe/safe.repository.ts
```diff
@@ -449,6 +449,45 @@ export class SafeRepository implements ISafeRepository {
...
+          `Failed to fetch Safe owners. chainId=${chainId}`,
+        );
+      }
+
+      result[chainId] =
+        allSafeList.status === 'fulfilled'
+          ? allSafeList.value.safeList.safes
+          : // Transaction Service threw; could own Safes or not
+            null;
+    }
```
katspaugh, line 486: The same check twice.
```suggestion
      if (allSafeList.status === 'fulfilled') {
        result[chainId] = allSafeList.value.safeList.safes
      } else {
        result[chainId] = null
        this.loggingService.warn(
          `Failed to fetch Safe owners. chainId=${chainId}`,
        );
      }
    }
```

iamacook: Changed in 2e27da02514d37577621737f2a9d10968172e559.

katspaugh (approved):
Thanks!

src/routes/owners/owners.controller.ts
```diff
@@ -23,12 +22,39 @@ export class OwnersController {
...
+    schema: {
+      type: 'object',
+      additionalProperties: {
+        type: 'array',
+        items: { type: 'string' },
+        nullable: true,
+      },
+    },
+  })
+  @Get('/v2/owners/:ownerAddress/safes')
```
PooyaRaki, line 53: The v2 endpoint should be stored in another controller, ideally the folder structure would look like:
`src/routes/owners/v1/owners.controller.ts`
`src/routes/owners/v2/owners.controller.ts`

But for now, if you think restructuring the whole component is outside the scope of this PR we could just create a new v2 file in the same directory:
`src/routes/owners/owners.controller.v1.ts`
`src/routes/owners/owners.controller.v2.ts`

The ideal solution is the first option, but I’m also satisfied with the second.

iamacook: As per discussion, this was adjusted in 912117016c32895b9bd01a41c8ac3a7730ead883:

There are now versioned controllers: `OwnersControllerV1` and `OwnersControllerV2` (`owners.controller.v*`.ts). As the service/repository methods are controller agnostic, however, they retain the `deprecated__` prefix, with JSDoc markers.

iamacook:
Do you think we should copy this behaviour to [other endpoints](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/safes/safes.service.ts#L191-L193) that query all Transaction Services?

PR #2272, opened 2025-01-17
Enhance Service Liveliness Check with Redis and AMQP Monitoring

PooyaRaki:
## Summary
This PR improves the service liveliness check by incorporating Redis and AMQP (e.g., RabbitMQ) health monitoring. These enhancements help detect connectivity issues early, ensuring better system reliability.

## Changes
- Implemented a `Redis` liveliness check.
- Implemented a `Amqp` liveliness check.
- Refactored Existing Health Check Logic
- Updated Unit Tests
- Ensured that failures in either Redis or AMQP are correctly reflected in the liveliness probe.

### DIFF
--- a/src/domain/health/health.repository.ts
+++ b/src/domain/health/health.repository.ts
@@ -10,6 +10,7 @@ import {
   IQueueReadiness,
   QueueReadiness,
 } from '@/domain/interfaces/queue-readiness.interface';
+import { HealthCheckError } from '@/domain/health/entities/healthError.entity';
 
 @Injectable()
 export class HealthRepository implements IHealthRepository {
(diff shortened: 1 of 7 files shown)


src/domain/health/health.repository.ts
```diff
@@ -19,19 +20,51 @@ export class HealthRepository implements IHealthRepository {
...
   async isReady(): Promise<HealthEntity> {
+    try {
+      this.isAmqpHealthy();
+      await this.isRedisHealthy();
+
+      return HealthEntity.READY;
+    } catch (error) {
+      if (error instanceof HealthCheckError) {
+        this.loggingService.warn(error.message);
+      }
```
iamacook, line 47: Have you checked if these errors aren't already logged?

src/routes/health/health.service.ts
```diff
@@ -18,14 +18,23 @@ export class HealthService {
...
+    switch (healthMetric) {
       case HealthEntity.READY:
         return new Health(HealthStatus.OK);
       case HealthEntity.NOT_READY:
         throw new ServiceUnavailableException(new Health(HealthStatus.KO));
       default:
-        this.loggingService.error(`Readiness status ${readiness} not handled`);
+        this.loggingService.error(
+          `Readiness status ${healthMetric} not handled`,
+        );
```
iamacook, line 37: Nit: why was this change needed?

src/domain/health/health.repository.ts
```diff
@@ -19,19 +20,51 @@ export class HealthRepository implements IHealthRepository {
...
+
+      return HealthEntity.NOT_READY;
+    }
+  }
```
iamacook, line 51: These methods are the same as each other. Why do we need two?

PooyaRaki: One is the readiness of the service, and the other is liveliness. Even though they are similar, I believe we should have separate methods for each since readiness and liveliness serve different purposes.

src/routes/health/health.service.ts
PooyaRaki: This line was already there; I just changed the structure of the code.

src/domain/health/health.repository.ts
PooyaRaki: I didn't change the logs; I just changed the structure of the code.

PooyaRaki:
> Have we confirmed these changes with DevOps?
DevOps isn’t involved here since this is our internal service logic to check the liveliness of all services, not just CGW.

src/domain/health/health.repository.ts
iamacook: Oversight on my behalf: didn't see we the error wasn't being thrown further.

iamacook:
> DevOps isn’t involved here since this is our internal service logic to check the liveliness of all services, not just CGW.

Afaiu, the endpoint is automatically checked and logged accordingly. Considering the recent spike in Redis-related logs, this may also cause an increase here.

src/domain/health/health.repository.ts
PooyaRaki: I strongly believe they should as separate as possible from each other even though they are exactly the same, but I added some new methods. Please have a look at the new changes

PooyaRaki:
> Afaiu, the endpoint is automatically checked and logged accordingly. Considering the recent spike in Redis-related logs, this may also cause an increase here.

The endpoint is automatically checked, logged and restarts the service if the services are not alive, It's not that it logs multiple times each second.

iamacook (approved):

PR #2274, opened 2025-01-20
Read rejectUnauthorized Database Requests from Environment Variables in Tests

PooyaRaki:
## Summary
This PR introduces the ability to toggle rejectUnauthorized on or off based on an environment variable. This allows for greater flexibility in test environments while ensuring secure database connections when necessary.

## Changes
- Enable rejectUnauthorized to be controlled via an environment variable.


iamacook:
As per discussion, the CA path was not correctly set for "standard" tests. I've set it in https://github.com/safe-global/safe-client-gateway/pull/2275, so this should no longer be required.

PR #2276, opened 2025-01-20
Remove transaction confirmation view endpoint

iamacook:
## Summary

Transaction confirmation views (`/v1/chains/:chainId/safes/:safeAddress/views/transaction-confirmation`) inherently mirrored transaction previews. As such, we previously [deprecated it](https://github.com/safe-global/safe-client-gateway/pull/1973).

As the endpoint is [no longer used](https://github.com/safe-global/safe-wallet-monorepo/pull/4783), we can safely remove the implementation. This does that.

## Changes

- Remove transaction confirmation view: module, controller, service and entities
- Simplify `KilnDecoder` methods that were dependant on the above
- Update tests accordingly


src/routes/transactions/mappers/common/native-staking.mapper.ts
```diff
@@ -181,7 +181,7 @@ export class NativeStakingMapper {
     ]);
     this.validateDeployment(deployment);
     const publicKeys = this.kilnNativeStakingHelper.splitPublicKeys(
-      this.kilnDecoder.decodeValidatorsExit(args.data)!.parameters[0].value,
+      this.kilnDecoder.decodeValidatorsExit(args.data)!,
```
PooyaRaki, line 184: This is not the concern of this PR but it's usually not a good practice to add `!` to the variable when the function return type can be `null` even though the only case the return type can be null is when the data to decode is not validator exit

PooyaRaki (approved):

src/routes/transactions/mappers/common/native-staking.mapper.ts
iamacook: I agree, and it shouldn't be used. However, at this stage the `data` is that of `requestValidatorsExit`. We can update it in a follow up PR though.

PR #2282, opened 2025-01-21
Add wallet scaffolding and migrations

usame-algan:
## Summary

Scaffolds wallet migrations and repository

## Changes

- Adds `create_wallets` migration
- Adds `update_wallet_timestamp_trigger` migration
- Adds empty `WalletsRepository`
- Imports `WalletsRepository` in `UserRepositoryModule`


src/domain/users/wallets/wallets.repository.interface.ts
```diff
@@ -0,0 +1,5 @@
+export const IWalletsRepository = Symbol('IWalletsRepository');
```
iamacook, line 1: Do you think wallet CRUD will occur outside of the user domain? If so, I think we should move it top-level and create a `WalletsRepositoryModule` as well.

src/datasources/users/entities/wallets.entity.db.ts
```diff
@@ -19,15 +19,18 @@ export const WalletSchema = RowSchema.extend({
 });
 
 @Entity('wallets')
-@Unique(['address'])
+@Unique('UQ_wallet_address', ['address'])
```
PooyaRaki, line 22: Praise: Thank you for thoughtfully naming the constraints. 💯

src/domain/users/wallets/wallets.repository.interface.ts
usame-algan: Good question! I think this is something that still needs to be discussed. As of now I believe Wallet CRUD will only happen in the context of users e.g. creating a user and wallet, adding a wallet to an existing user and removing a wallet from a user. On the other hand User CRUD will happen both on its own e.g. deleting a user and in the context of organisations e.g. adding a member to an org (which possibly needs to create an inactive user).

usame-algan: For now I would keep it as part of the user domain as its not planned that wallets will be used independent of users. If that happens at a later stage e.g. if we want to use wallets for notifications then we can still extract it.

iamacook (approved):

4player90 (approved):

4player90 (approved):

PR #2283, opened 2025-01-22
Serialize `Error` messages for logs

iamacook:
## Summary

We [pass `unknown` messages](https://github.com/safe-global/safe-client-gateway/blob/cf41694e757e32ab637fe28e1ae770cb0a40e0fc/src/logging/logging.service.ts#L56) to our logger.

Currently, winston only formats [using the `json`](https://github.com/safe-global/safe-client-gateway/blob/main/src/logging/logging.module.ts#L39) [formatter](https://github.com/winstonjs/logform?tab=readme-ov-file#json). This means that `Error` instances serialise to an empty object and are not shown in our logs. The [`error` formatter](https://github.com/winstonjs/logform?tab=readme-ov-file#errors) solves this, serialising the `message` of the instance and returning that instead of an empty object.

This combines the `json` and `error` formatters to successful log both JSON and `Error` instances.

## Changes

- Add `error` formatter to transport

### DIFF
--- a/src/logging/logging.module.ts
+++ b/src/logging/logging.module.ts
@@ -11,7 +11,7 @@ import { RequestScopedLoggingService } from '@/logging/logging.service';
  * @param transports - the logger transports to be used in the winston instance
  * @param configurationService - the configuration service to retrieve service-level settings
  */
-function winstonFactory(
+export function winstonFactory(
   transports: Array<Transport> | Transport,
   configurationService: IConfigurationService,
 ): winston.Logger {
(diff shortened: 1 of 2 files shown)


src/logging/logging.module.ts
```diff
@@ -36,7 +36,7 @@ function winstonTransportsFactory(
     level: configurationService.getOrThrow<string>('log.level'),
     format: prettyColorize
       ? winston.format.prettyPrint({ colorize: true })
-      : winston.format.json(),
+      : winston.format.combine(winston.format.errors(), winston.format.json()),
```
iamacook, line 39: It is also [possible to access the stack](https://github.com/winstonjs/logform?tab=readme-ov-file#errors) of an `Error` should we want to in the future.

PooyaRaki (approved):

PR #2284, opened 2025-01-22
Automatically remove database lock when migration fails

PooyaRaki:
## Summary
This PR introduces an automatic mechanism to remove database locks when a migration fails. Previously, failed migrations could leave the database in a locked state, requiring manual intervention. This enhancement ensures that the lock is cleaned up

## Changes
- Removes locks when the failure is confirmed.
- Adjusts tests

### DIFF
--- a/src/datasources/db/v2/database-migrator.service.ts
+++ b/src/datasources/db/v2/database-migrator.service.ts
@@ -152,14 +152,20 @@ export class DatabaseMigrator {
    * @returns {Promise<void>} An empty promise that resolves when the migration is complete.
    */
   private async runMigration(connection: DataSource): Promise<void> {
-    const migrations = await connection.runMigrations({ transaction: 'all' });
+    try {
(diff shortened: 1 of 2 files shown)


src/datasources/db/v2/database-migrator.service.ts
```diff
@@ -64,6 +64,7 @@ export class DatabaseMigrator {
         break;
       } else {
         if (numberOfIterations === numberOfRetries) {
+          await this.truncateLocks(connection);
```
iamacook, line 67: If the lock exists due to ongoing migrations in another instance, truncating it here would effectively invalidate the purpose of the lock. Is this intended?

PooyaRaki: Since running services is not asynchronous, it’s safe to remove the lock. Additionally, after five retries, the migrations should have either completed successfully or failed. However, to be more precise, I modified the code to remove the lock only in case of a migration error.

iamacook (approved):

PR #2286, opened 2025-01-23
Validate `data` of `EXECUTED_MULTISIG_TRANSACTION` event

iamacook:
## Summary

A new (nullable) `data` field has been added to the `EXECUTED_MULTISIG_TRANSACTION` event [on the Transaction Service](https://github.com/safe-global/safe-transaction-service/pull/2392).

This adds the appropriate validation of the property, should we want to use it in the future.

## Changes

- Add `data` to validation schema


src/datasources/push-notifications-api/entities/firebase-notification.entity.ts
```diff
@@ -3,5 +3,5 @@ export type FirebaseNotification = {
     title?: string;
     body?: string;
   };
-  data?: Record<string, string>;
+  data?: Record<string, unknown>;
```
iamacook, line 6: This needed updating as `data` can `null`.

PooyaRaki: data is already optional. Since Firebase only accepts string values, it looks like the type was defined as an object with string values to ensure that only strings are passed to Firebase.

iamacook: I've adjusted the schema to return `0x${string} | undefined` so as to not modify this type.

PooyaRaki (approved):

PR #2288, opened 2025-01-23
Fix(Swagger): TransactionInfo as a union type

katspaugh:
## Summary

Swagger had the wrong type for TransactionInfo in TransactionDetails.

I extracted a union schema into a shared function which is now used in the TransactionDetails, Transaction and TransactionPreview entities.


iamacook:
As per discussion, I've adjusted the implementation to instead [extend a base entity](https://github.com/safe-global/safe-client-gateway/pull/2289). I will therefore close this in favour of that.

PR #2289, opened 2025-01-23
Add and extend `BaseTransaction` entity

iamacook:
## Summary

Superseeds #2288.

Swagger does not correctly define the `txInfo` of `TransactionDetails`, and has duplicate definitions for `TransactionPreview` and `Transaction`.

All of the above have been simplified into a new `BaseTransaction` entity, which is extended from. It correctly implements and defines `txInfo`.

Below are screenshots from Swagger demonstrating the implementation (observe the correct `txInfo`):

image

image

image

## Changes

- Add `BaseTransaction` entity
- Extend `TransactionDetails`, `TransactionPreview` and `Transaction` from it


katspaugh:
@PooyaRaki could you pls review this? (I'm blocked by it to proceed with the refactoring I mentioned yesterday.)

PooyaRaki (approved):
I’ll approve the PR, but for future reference: Hiding an object’s properties within its parent can make the code harder to follow. This becomes even more challenging as more properties are added to the parent object, requiring developers to jump between types to understand its structure.

PR #2291, opened 2025-01-24
Fix(Balances): make decimals non-nullable

katspaugh:
## Summary

Token decimals being always a number has been the assumption on the client for quite some time and a lot of code would have to be changed if we were to start treating it as nullable.

## Changes

So I've adjusted the OpenAPI type and also added a fallback value (although it's probably never null in reality).

### DIFF
--- a/src/datasources/balances-api/safe-balances-api.service.ts
+++ b/src/datasources/balances-api/safe-balances-api.service.ts
@@ -17,12 +17,12 @@ import { Chain } from '@/domain/chains/entities/chain.entity';
 import { rawify, type Raw } from '@/validation/entities/raw.entity';
 import { AssetPricesSchema } from '@/datasources/balances-api/entities/asset-price.entity';
 import { ZodError } from 'zod';
+import { DEFAULT_DECIMALS } from '@/domain/tokens/entities/schemas/token.schema';
 
 @Injectable()
 export class SafeBalancesApi implements IBalancesApi {
   private readonly defaultExpirationTimeInSeconds: number;
   private readonly defaultNotFoundExpirationTimeSeconds: number;
-  private static readonly DEFAULT_DECIMALS = 18;
   private static readonly HOLESKY_CHAIN_ID = '17000';
 
   constructor(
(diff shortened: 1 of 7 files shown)


src/datasources/balances-api/safe-balances-api.service.ts
```diff
@@ -198,6 +198,11 @@ export class SafeBalancesApi implements IBalancesApi {
         const fiatBalance = this._getFiatBalance(price, balance);
         return {
           ...balance,
+          token: {
+            ...balance.token,
+            decimals:
+              balance.token?.decimals ?? SafeBalancesApi.DEFAULT_DECIMALS,
+          },
```
iamacook, line 205: As there are [multiple](https://github.com/safe-global/safe-client-gateway/blob/decimals/src/datasources/balances-api/zerion-balances-api.service.ts#L46) `IBalanceApi`s, we should assign the default value [during validation](https://github.com/safe-global/safe-client-gateway/blob/decimals/src/domain/tokens/entities/schemas/token.schema.ts#L8). It then doesn't need to be done in both places and will solve [the `MultisigTransactionExecutionDetailsMapper` CI error](https://github.com/safe-global/safe-client-gateway/actions/runs/12952281091/job/36129259361?pr=2291#step:6:14).

katspaugh: Nice, thanks for the pointer!

iamacook (approved):

src/domain/tokens/entities/schemas/__tests__/token.schema.spec.ts
```diff
@@ -26,14 +26,14 @@ describe('TokenSchema', () => {
...
-  it('should allow undefined decimals', () => {
+  it('should fallback to default decimals', () => {
     const token = tokenBuilder().build();
     // @ts-expect-error - inferred type doesn't allow optional properties
     delete token.decimals;
 
     const result = TokenSchema.safeParse(token);
 
-    expect(result.success && result.data.decimals).toBe(null);
+    expect(result.success && result.data.decimals).toBe(18);
```
PooyaRaki, line 36: Nit: I would add a new `const` is test named `DEFAULT_DECIMALS`.

PooyaRaki (approved):

PooyaRaki (approved):

PR #2294, opened 2025-01-27
Scaffold org entity, domain, route and service

usame-algan:
## Summary

Scaffolds the organisations entity datasource, migrations, repository, route and service and includes the module on the app level with a feature flag.

## Changes

- Create `organisations.entity.db.ts` TypeORM entity
- Create `organisation.entity.ts` domain entity
- Create empty `organisations.repository.ts` and `organisations.repository.interface.ts`
- Create empty `organisations.controller.ts` and `organisations.service.ts`
- Create `organisations.repository.module.ts` and `organisations.module.ts`
- Import `organisations.module.ts` in `app.module.ts`
- Create `create_organisations` and `update_organisations_timestamp_trigger` migrations

### DIFF
--- a/src/datasources/organisations/entities/organisations.entity.db.ts
+++ b/src/datasources/organisations/entities/organisations.entity.db.ts
@@ -1,0 +1,32 @@
+import { Column, Entity, Index, PrimaryGeneratedColumn } from 'typeorm';
+import {
+  OrganisationStatus,
+  Organisation as DomainOrganisation,
+} from '@/domain/organisations/entities/organisation.entity';
+
+@Entity('organisations')
+export class Organisations implements DomainOrganisation {
+  @PrimaryGeneratedColumn({ primaryKeyConstraintName: 'PK_org_id' })
+  id!: number;
+
+  @Column({ type: 'varchar', length: 255 })
+  name!: string;
+
+  @Index('idx_org_status')
+  @Column({
+    type: 'integer',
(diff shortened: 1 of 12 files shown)


src/datasources/organisations/entities/organisations.entity.db.ts
```diff
@@ -0,0 +1,32 @@
...
+  id!: number;
+
+  @Column({ type: 'varchar', length: 255 })
+  name!: string;
+
+  @Index('idx_org_status')
+  @Column({
+    type: 'integer',
+  })
+  status!: OrganisationStatus;
```
iamacook, line 19: This is out of the scope of this PR, but a thought I had re. numeric enums. What do you think about leveraging the `transform` method to convert this to <> from a string?

iamacook (approved):

src/datasources/organisations/entities/organisations.entity.db.ts
usame-algan: Yes, agree handling strings rather than integers would make the code easier to read while still having the benefit of storing and indexing it more efficiently. @PooyaRaki do you have an opinion on using transformers for the enums we create?

iamacook: As we're using an enum, it should be fine as is for now but it's something we can consider in the future.

PR #2296, opened 2025-01-27
Add wallet to user

jmealy:
## Summary
Implements a method on the domain layer to add an additional wallet to an existing User.

## Changes
- adds `addWalletToUser` function.
- adds test for `addWalletToUser`


src/domain/users/users.repository.interface.ts
```diff
@@ -1,5 +1,6 @@
 import type { AuthPayload } from '@/domain/auth/entities/auth-payload.entity';
 import type { User, UserStatus } from '@/domain/users/entities/user.entity';
+import type { Wallet } from '@/domain/users/entities/wallet.entity';
```
iamacook, line 3: Nit: I'd only reference this type if we validate against it. At the moment, we're "using" the database entity.

When access datasources, we [normally parse the incoming data](https://github.com/safe-global/safe-client-gateway/blob/main/src/domain/chains/chains.repository.ts#L46), but since the inclusion of TypeORM, this is not the case with database entities.

I have no strong opinion re. this as everything should be typesafe but perhaps we should consider using these schemas as they are also previously defined [for notifications](https://github.com/safe-global/safe-client-gateway/blob/main/src/datasources/notifications/entities/notification-devices.entity.db.ts#L15-L19), for example.

@PooyaRaki, what do you think?

src/domain/users/users.repository.ts
```diff
@@ -46,4 +46,43 @@ export class UsersRepository implements IUsersRepository {
...
+      async (entityManager: EntityManager) => {
+        const walletRepository = entityManager.getRepository(Wallet);
+
+        const authenticatedWallet = await walletRepository.findOne({
+          where: { address: args.authPayload.signer_address },
+          relations: { user: true },
+        });
+
+        if (!authenticatedWallet?.user) {
+          throw new Error('User not found');
```
iamacook, line 64:
```suggestion
          throw new NotFoundException('User not found');
```

src/domain/users/users.repository.ts
```diff
@@ -46,4 +46,43 @@ export class UsersRepository implements IUsersRepository {
...
+  async addWalletToUser(args: {
+    newSignerAddress: `0x${string}`;
+    authPayload: AuthPayload;
+  }): Promise<Pick<Wallet, 'id'>> {
+    return await this.postgresDatabaseService.transaction(
+      async (entityManager: EntityManager) => {
+        const walletRepository = entityManager.getRepository(Wallet);
+
+        const authenticatedWallet = await walletRepository.findOne({
+          where: { address: args.authPayload.signer_address },
```
iamacook, line 59: Nit: it took me a little bit to deduce the authentication as `chain_id` and `signer_address` of `AuthPayload` can be `undefined`.

If I understand correctly, this is handled by the presence of `authenticatedWallet`? What do you think about throwing a 401 if none exsists?

src/domain/users/users.repository.ts
```diff
@@ -46,4 +46,43 @@ export class UsersRepository implements IUsersRepository {
...
+        if (walletAlreadyExists) {
+          throw new ConflictException(
+            'A wallet with the same address already exists',
+          );
+        }
+
+        const walletInsertResult = await walletRepository.insert({
+          user: authenticatedWallet.user,
+          address: args.newSignerAddress,
+        });
```
iamacook, line 82: As the `address` is unique, insertion throws when trying to add a duplicate entry. Would there be benefit to handling the error instead of first checking the presence? cc @PooyaRaki

src/domain/users/users.repository.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2296/commits/4f7e8cc2b7bf7798281c1207400e863bf0e8bd7a

src/domain/users/users.repository.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2296/commits/4f7e8cc2b7bf7798281c1207400e863bf0e8bd7a

src/domain/users/users.repository.ts
PooyaRaki: @iamacook Very good point. We should always catch the error unless checking the presence is necessary. This way we will execute 1 less query.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +65,33 @@ describe('UsersRepository', () => {
       expect(result).toEqual({ id: mockUser.id });
     });
   });
+
+  describe('addWalletToUser', () => {
```
iamacook, line 69: We should also add tests coverage for each thrown error.

src/domain/users/users.repository.ts
```diff
@@ -46,4 +51,47 @@ export class UsersRepository implements IUsersRepository {
...
+    newSignerAddress: `0x${string}`;
+    authPayload: AuthPayload;
+  }): Promise<Pick<Wallet, 'id'>> {
+    return await this.postgresDatabaseService.transaction(
+      async (entityManager: EntityManager) => {
+        const walletRepository = entityManager.getRepository(Wallet);
+
+        if (!args.authPayload.signer_address) {
+          throw new UnauthorizedException();
+        }
```
iamacook, line 65: Nit: let's move this outside of the transaction.

src/domain/users/users.repository.interface.ts
PooyaRaki: @iamacook I have no strong opinion here, I think it doesn't hurt to have this type but if it's not used anywhere atm then we can remove it.

iamacook: Considering we have a pattern of inferring types from schemas, it would be nice to keep/parse against them but if we don't then I can see benefit in removing them.

src/domain/users/users.repository.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2296/commits/c70aa972e85ca4c7d96117f77a19e50f8a86263c

src/domain/users/users.repository.spec.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2296/commits/6f59db25211414b678f461018edee617bbbb58a1

src/domain/users/users.repository.ts
jmealy: https://github.com/safe-global/safe-client-gateway/pull/2296/commits/c70aa972e85ca4c7d96117f77a19e50f8a86263c

src/domain/users/users.repository.interface.ts
jmealy: @iamacook  Do I understand correctly:  you would prefer to get the Wallet type from the database entity here instead of the zod schema, because the purpose of the latter is to parse against, which we are not doing here?

iamacook: Correct. Only after parsing should we use the schema type.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+    it('should add a wallet to a user', async () => {
+      const authPayloadDto = authPayloadDtoBuilder().build();
+      const authPayload = new AuthPayload(authPayloadDto);
+
+      const newSignerAddressMock = faker.finance.ethereumAddress();
```
iamacook, line 79:
```suggestion
      const newSignerAddressMock = getAddress(faker.finance.ethereumAddress());
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+      mockWalletRepository.findOne.mockResolvedValueOnce(mockWallet);
+      mockWalletRepository.insert.mockResolvedValue({
+        identifiers: [{ id: mockWallet.id }],
+        generatedMaps: [{ id: 1 }],
+        raw: jest.fn(),
+      });
+
+      const result = await usersRepository.addWalletToUser({
+        authPayload,
+        newSignerAddress: newSignerAddressMock as `0x${string}`,
```
iamacook, line 90: When using `getAddress`, we can remove the cast here.

```suggestion
        newSignerAddress: newSignerAddressMock,
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+        user: mockWallet.user,
+        address: newSignerAddressMock,
+      });
+
+      expect(result).toEqual({ id: mockWallet.id });
+    });
+
+    it('should throw an UnauthorizedException if the auth payload is empty', async () => {
+      const authPayload = new AuthPayload();
+      const newSignerAddressMock = faker.finance.ethereumAddress();
```
iamacook, line 103:
```suggestion
      const newSignerAddressMock = getAddress(faker.finance.ethereumAddress());
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+
+      await expect(
+        usersRepository.addWalletToUser({
+          authPayload,
+          newSignerAddress: newSignerAddressMock as `0x${string}`,
```
iamacook, line 108:
```suggestion
          newSignerAddress: newSignerAddressMock,
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+        }),
+      ).rejects.toThrow(UnauthorizedException);
+    });
+
+    it('should throw a NotFoundException if the user is not found', async () => {
+      const authPayloadDto = authPayloadDtoBuilder().build();
+      const authPayload = new AuthPayload(authPayloadDto);
+
+      const newSignerAddressMock = faker.finance.ethereumAddress();
```
iamacook, line 117:
```suggestion
      const newSignerAddressMock = getAddress(faker.finance.ethereumAddress());
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+      mockWalletRepository.findOne.mockResolvedValueOnce(null);
+
+      await expect(
+        usersRepository.addWalletToUser({
+          authPayload,
+          newSignerAddress: newSignerAddressMock as `0x${string}`,
```
iamacook, line 123:
```suggestion
          newSignerAddress: newSignerAddressMock,
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+      const mockWallet = walletBuilder().build();
+      mockWalletRepository.findOne.mockResolvedValueOnce(mockWallet);
+      mockWalletRepository.insert.mockRejectedValue(
+        new QueryFailedError('', [], new Error()),
+      );
+
+      await expect(
+        usersRepository.addWalletToUser({
+          authPayload,
+          newSignerAddress: newSignerAddressMock as `0x${string}`,
```
iamacook, line 142:
```suggestion
          newSignerAddress: newSignerAddressMock,
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+
+      const newSignerAddressMock = faker.finance.ethereumAddress();
+      mockWalletRepository.findOne.mockResolvedValueOnce(null);
+
+      await expect(
+        usersRepository.addWalletToUser({
+          authPayload,
+          newSignerAddress: newSignerAddressMock as `0x${string}`,
+        }),
+      ).rejects.toThrow(NotFoundException);
```
iamacook, line 125: Nit: let's add check of the error message.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +70,78 @@ describe('UsersRepository', () => {
...
+      mockWalletRepository.insert.mockRejectedValue(
+        new QueryFailedError('', [], new Error()),
+      );
+
+      await expect(
+        usersRepository.addWalletToUser({
+          authPayload,
+          newSignerAddress: newSignerAddressMock as `0x${string}`,
+        }),
+      ).rejects.toThrow(ConflictException);
```
iamacook, line 144: Nit: let's add check of the error message.

src/domain/users/users.repository.ts
```diff
@@ -46,4 +51,42 @@ export class UsersRepository implements IUsersRepository {
...
+            user: authenticatedWallet.user,
+            address: args.newSignerAddress,
+          });
+          return { id: walletInsertResult.identifiers[0].id };
+        } catch (error) {
+          if (error instanceof QueryFailedError) {
+            throw new ConflictException(
+              `A wallet with the same address already exists`,
+            );
+          }
```
iamacook, line 86: We are assuming that every query failure denotes a unique violation. We should also check the error code, for which [unique violation is 23505](https://www.postgresql.org/docs/current/errcodes-appendix.html).

src/domain/users/users.repository.ts
```diff
@@ -46,4 +51,45 @@ export class UsersRepository implements IUsersRepository {
...
+        try {
+          const walletInsertResult = await walletRepository.insert({
+            user: authenticatedWallet.user,
+            address: args.newSignerAddress,
+          });
+          return { id: walletInsertResult.identifiers[0].id };
+        } catch (error) {
+          if (
+            error instanceof QueryFailedError &&
+            error.driverError.code === '23505'
```
iamacook, line 84: Nit: we should consider creating either an enum or adding a comment for such error codes.

iamacook (approved):

src/domain/users/users.repository.ts
jmealy: I added a comment for now

PooyaRaki: I would create the enum sooner then the code will look like:
`error.driverError.code === DatabaseErrorCode.CONFLICT`

iamacook: Might be best to follow their docs, e.g. `PostgresErrorCodes.UniqueViolation`. What do you think?

jmealy: Added the enum in `/datasources/db/v2/postgres-error-codes.ts`. Lmk if there is a better place for it
 https://github.com/safe-global/safe-client-gateway/pull/2296/commits/81781c145894725cb6cab3995c16379054ec5f97

iamacook (approved):

katspaugh:
@iamacook @PooyaRaki could we update and merge this while James is away?

iamacook:
Closing as superseeded by https://github.com/safe-global/safe-client-gateway/pull/2315/commits/4a10e412e74c248ea0662b322daf9e6cc90005b3.

PR #2309, opened 2025-01-28
Create token-specific entities

### LINKED ISSUE #2307: Create native, ERC-20 and ERC-721 entities
### Description

We currently [parse all token types the same](https://github.com/safe-global/safe-client-gateway/blob/main/src/domain/tokens/entities/schemas/token.schema.ts). This results in a merged entity that doesn't follow standards, e.g. ERC-721 tokens should have 0 decimals, but are returned as `null`.

We should have explicit validation schemas and associated entities based on the [token type](https://github.com/safe-global/safe-client-gateway/blob/main/src/domain/tokens/entities/token.entity.ts#L4). These should have the correct fallback values, should anything be returned as `null`, e.g. native token having 18 decimals, ERC-721 having 0 decimals, etc.

### Requirements

- [`TokenSchema`](https://github.com/safe-global/safe-client-gateway/blob/main/src/domain/tokens/entities/schemas/token.schema.ts) should be split into three: `NativeTokenSchema`, `Erc20TokenSchem`, `Erc721TokenSchema` with correct fallback values.
- Swagger-specific entities created for each of the above, tagged correctly.

### Additional information

- [Initial change](https://github.com/safe-global/safe-client-gateway/pull/2291) that [needed reverting](https://github.com/safe-global/safe-client-gateway/pull/2306) due to ERC-721 decimal issues.
- [Discussion](https://5afe.slack.com/archives/C04MJJM9S3U/p1738061419922679) regarding this proposal.

iamacook:
## Summary

Resolves #2307

We currently parse native, ERC-20 and ERC-721 tokens as one, meaning we cannot safely assume decimals should they not exist.

This splits `TokenSchema` into three specific schemas per token type with the following assumptions:

- Native tokens always have defined decimals.
- ERC-20 tokens have 0 decimals, should they not be specified.
- ERC-721 tokens [have 0 decimals](https://eips.ethereum.org/EIPS/eip-721#backwards-compatibility), should they not be specified.

From the parsed token, token-specific entities are mapped for Swagger.

## Changes

- Split `TokenSchema` into `NativeTokenSchema`, `Erc20TokenSchema` and `Erc721TokenSchema`
- Infer types and create respective builders for each
- Map Swagger-specific entities according to types
- Update tests accordingly


src/domain/tokens/entities/token.entity.ts
```diff
@@ -1,10 +1,45 @@
...
+  decimals: z.number(),
+});
+
+const Erc20TokenSchema = BaseTokenSchema.extend({
+  type: z.literal('ERC20'),
+  decimals: z.number().catch(DEFAULT_ERC20_DECIMALS),
+});
+
+const Erc721TokenSchema = BaseTokenSchema.extend({
+  type: z.literal('ERC721'),
```
iamacook, line 27: String `type`s are now preferred due to type complaints when using different enums from the domain vs. route layers.

src/domain/tokens/entities/token.entity.ts
```diff
@@ -1,10 +1,45 @@
-import type { TokenSchema } from '@/domain/tokens/entities/schemas/token.schema';
-import type { z } from 'zod';
-
-export enum TokenType {
-  Erc721 = 'ERC721',
-  Erc20 = 'ERC20',
-  NativeToken = 'NATIVE_TOKEN',
-}
+import { z } from 'zod';
```
iamacook, line 1: This is the "core" change, from which everything else implemented.

hectorgomezv (approved):
👏🏻

hectorgomezv (approved):

PR #2310, opened 2025-01-28
Implement user domain logic

usame-algan:
## Summary

Implements functions on the user domain

## Changes

- Add `getUser`
- Add `deleteUser`
- Add `deleteWalletFromUser`

### DIFF
--- a/src/domain/users/users.repository.spec.ts
+++ b/src/domain/users/users.repository.spec.ts
@@ -35,6 +35,12 @@ describe('UsersRepository', () => {
     jest.clearAllMocks();
 
     usersRepository = new UsersRepository(mockPostgresDatabaseService);
+    mockPostgresDatabaseService.getRepository.mockImplementation((entity) => {
+      if (entity === User) return Promise.resolve(mockUserRepository);
+      if (entity === Wallet) return Promise.resolve(mockWalletRepository);
+
+      return Promise.reject(new Error('Invalid entity'));
+    });
     mockPostgresDatabaseService.transaction.mockImplementation((fn) =>
       fn(mockEntityManager),
     );
@@ -175,8 +181,8 @@ describe('UsersRepository', () => {
           walletAddress: addressToRemove,
         }),
       ).rejects.toThrow(
-        new NotFoundException(
-          `A wallet with address ${addressToRemove} does not exist for the current user`,
+        new ConflictException(
+          `User could not be removed from wallet. Wallet=${addressToRemove}`,
         ),
       );
     });
@@ -186,9 +192,9 @@ describe('UsersRepository', () => {
     it('should throw an UnauthorizedException if the auth payload is empty', async () => {
       const authPayload = new AuthPayload();
 
-      await expect(usersRepository.getUser(authPayload)).rejects.toThrow(
-        UnauthorizedException,
-      );
+      await expect(
+        usersRepository.getUserWithWallets(authPayload),
+      ).rejects.toThrow(UnauthorizedException);
     });
 
     it('should return user information and associated wallets', async () => {
@@ -213,19 +219,26 @@ describe('UsersRepository', () => {
       mockWalletRepository.findOne.mockResolvedValueOnce(
         mockAuthenticatedWallet,
       );
-      mockWalletRepository.findBy.mockResolvedValueOnce([
-        mockAuthenticatedWallet,
-        mockAdditionalWallet,
+      mockWalletRepository.find.mockResolvedValueOnce([
+        {
+          id: mockAuthenticatedWallet.id,
+          address: mockAuthenticatedWallet.address,
+        },
+        {
+          id: mockAdditionalWallet.id,
+          address: mockAdditionalWallet.address,
+        },
       ]);
 
-      const result = await usersRepository.getUser(authPayload);
+      const result = await usersRepository.getUserWithWallets(authPayload);
 
       expect(mockWalletRepository.findOne).toHaveBeenCalledWith({
         where: { address: authPayload.signer_address },
         relations: { user: true },
       });
-      expect(mockWalletRepository.findBy).toHaveBeenCalledWith({
-        user: mockAuthenticatedWallet.user,
+      expect(mockWalletRepository.find).toHaveBeenCalledWith({
+        select: ['id', 'address'],
+        where: { user: mockAuthenticatedWallet.user },
       });
       expect(result).toEqual({
         id: mockUser.id,
@@ -246,9 +259,9 @@ describe('UsersRepository', () => {
     it('should throw an UnauthorizedException if no authenticated wallet is defined', async () => {
       const authPayload = new AuthPayload();
 
-      await expect(usersRepository.getUser(authPayload)).rejects.toThrow(
-        UnauthorizedException,
-      );
+      await expect(
+        usersRepository.getUserWithWallets(authPayload),
+      ).rejects.toThrow(UnauthorizedException);
     });
 
     it('should throw a NotFoundException if the user is not found', async () => {
@@ -257,9 +270,9 @@ describe('UsersRepository', () => {
 
       mockWalletRepository.findOne.mockResolvedValueOnce(null);
 
-      await expect(usersRepository.getUser(authPayload)).rejects.toThrow(
-        new NotFoundException('User not found'),
-      );
+      await expect(
+        usersRepository.getUserWithWallets(authPayload),
+      ).rejects.toThrow(new NotFoundException('User not found'));
     });
   });
 
@@ -298,9 +311,7 @@ describe('UsersRepository', () => {
       mockUserRepository.delete.mockResolvedValueOnce({ affected: 0, raw: {} });
 
       await expect(usersRepository.deleteUser(authPayload)).rejects.toThrow(
-        new NotFoundException(
-          `A user for wallet ${walletAddress} does not exist.`,
-        ),
+        new NotFoundException(`Could not delete user. Wallet=${walletAddress}`),
       );
 
       expect(mockUserRepository.delete).toHaveBeenCalledWith({
--- a/src/domain/users/users.repository.ts
+++ b/src/domain/users/users.repository.ts
@@ -53,60 +53,54 @@ export class UsersRepository implements IUsersRepository {
     );
   }
 
-  async getUser(authPayload: AuthPayload): Promise<{
+  async getUserWithWallets(authPayload: AuthPayload): Promise<{
     id: User['id'];
     status: User['status'];
-    wallets: Array<{ address: Wallet['address']; id: Wallet['id'] }>;
+    wallets: Array<Pick<Wallet, 'id' | 'address'>>;
   }> {
     this.assertSignerAddress(authPayload);
 
-    return await this.postgresDatabaseService.transaction(
-      async (entityManager: EntityManager) => {
-        const walletRepository = entityManager.getRepository(Wallet);
+    const walletRepository =
+      await this.postgresDatabaseService.getRepository(Wallet);
 
-        const authenticatedWallet = await walletRepository.findOne({
-          where: { address: authPayload.signer_address },
-          relations: { user: true },
-        });
+    const authenticatedWallet = await walletRepository.findOne({
+      where: { address: authPayload.signer_address },
+      relations: { user: true },
+    });
 
-        if (!authenticatedWallet?.user) {
-          throw new NotFoundException('User not found');
-        }
+    if (!authenticatedWallet?.user) {
+      throw new NotFoundException('User not found');
+    }
 
-        const wallets = await walletRepository.findBy({
-          user: authenticatedWallet.user,
-        });
-
-        return {
-          id: authenticatedWallet.user.id,
-          status: authenticatedWallet.user.status,
-          wallets: wallets.map((wallet) => ({
-            id: wallet.id,
-            address: wallet.address,
-          })),
-        };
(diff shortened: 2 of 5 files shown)


src/domain/users/users.repository.ts
```diff
@@ -46,4 +51,78 @@ export class UsersRepository implements IUsersRepository {
...
+        const authenticatedWallet = await walletRepository.findOne({
+          where: { address: authPayload.signer_address },
+          relations: { user: true },
+        });
+
+        if (!authenticatedWallet?.user) {
+          throw new NotFoundException('User not found');
+        }
+
+        // TODO: Check that it also deletes all wallets of that user
```
usame-algan, line 112: We added a `cascade` in the wallet entity join column `onDelete`. Does this work in both directions or do we need to add it to the user entity as well?

PooyaRaki: A CASCADE foreign key ensures that when a parent record is deleted, all associated child records are also removed. However, deleting a child record does not affect the parent. In your case, the `Wallet` is a child of the `User`. This means that if a `User` is deleted, all associated `Wallets` will be removed as well. However, deleting a `Wallet` will not impact the `User`.

usame-algan: Alternatively, we could also explicitely delete all wallets:

`await walletRepository.delete({ user: authenticatedWallet.user })`

usame-algan: Thanks for the explanation! Then I suggest we leave it up to the cascade to handle it instead of explicitely removing wallets when deleting a user. We can test this functionality once we have the endpoints.

PooyaRaki: @usame-algan Yes, exactly we should leave it up to the cascade to handle as that's why we defined the FK as `CASCADE`

src/domain/users/users.repository.ts
```diff
@@ -46,4 +52,116 @@ export class UsersRepository implements IUsersRepository {
...
+        await userRepository.delete(authenticatedWallet.user);
+      },
+    );
+  }
+
+  async deleteWalletFromUser(args: {
```
iamacook, line 119: What do you think about splitting this function a bit as there's quite a bit going on? e.g. `removeWalletFromUser` and `deleteWallet`

src/domain/users/users.repository.ts
```diff
@@ -46,4 +52,116 @@ export class UsersRepository implements IUsersRepository {
...
+
+        if (!authenticatedWallet?.user) {
+          throw new NotFoundException('User not found');
+        }
+
+        const userWalletCount = await walletRepository.count({
+          where: { user: authenticatedWallet.user },
+        });
+
+        if (userWalletCount <= 1) {
```
iamacook, line 148:
```suggestion
        if (userWalletCount === 1) {
```

src/domain/users/users.repository.ts
```diff
@@ -46,4 +52,116 @@ export class UsersRepository implements IUsersRepository {
...
+          where: { address: authPayload.signer_address },
+          relations: { user: true },
+        });
+
+        if (!authenticatedWallet?.user) {
+          throw new NotFoundException('User not found');
+        }
+
+        // TODO: Check that it also deletes all wallets of that user
+        await userRepository.delete(authenticatedWallet.user);
```
iamacook, line 114: We should also check that an entity was `affected`.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,115 @@ describe('UsersRepository', () => {
       expect(result).toEqual({ id: mockUser.id });
     });
   });
+
```
iamacook, line 75: We should also add test coverage for `getUser` and `deleteUser`.

src/domain/users/users.repository.ts
```diff
@@ -46,4 +52,116 @@ export class UsersRepository implements IUsersRepository {
...
+            'Cannot delete the last wallet of a user',
+          );
+        }
+
+        const deleteResult = await walletRepository.delete({
+          address: args.walletAddress,
+          user: authenticatedWallet.user,
+        });
+
+        if (!deleteResult.affected) {
```
iamacook, line 159: Nit: I would be more explicit here for reability. I initially thought this was a `boolean`.

```suggestion
        if (deleteResult.affected === 0) {
```

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,115 @@ describe('UsersRepository', () => {
...
+    });
+
+    it('should throw an ConflictException if the user tries to remove the currently authenticated wallet', async () => {
+      const authPayloadDto = authPayloadDtoBuilder().build();
+      const authPayload = new AuthPayload(authPayloadDto);
+
+      await expect(
+        usersRepository.deleteWalletFromUser({
+          authPayload,
+          walletAddress: authPayload.signer_address as `0x${string}`,
```
iamacook, line 138: Nit: casting can result in undesired side effects. I'd suggest throwing further up if `undefined` and removing this.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,115 @@ describe('UsersRepository', () => {
...
+
+      await expect(
+        usersRepository.deleteWalletFromUser({
+          authPayload,
+          walletAddress: addressToRemove,
+        }),
+      ).rejects.toThrow(BadRequestException);
+    });
+
+    it('should throw an ConflictException if the user tries to remove the currently authenticated wallet', async () => {
```
iamacook, line 131: Nit: let's move this case one up to follow the logic in the function.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,115 @@ describe('UsersRepository', () => {
...
+      const authPayload = new AuthPayload(authPayloadDto);
+
+      const addressToRemove = getAddress(faker.finance.ethereumAddress());
+
+      const mockAuthenticatedWallet = walletBuilder().build();
+      mockWalletRepository.findOne.mockResolvedValueOnce(
+        mockAuthenticatedWallet,
+      );
+      mockWalletRepository.delete.mockResolvedValue({ affected: 0, raw: {} });
+      mockWalletRepository.count.mockResolvedValue(2);
```
iamacook, line 172: This should successfully delete as there are >1 wallets attached to the user. This test should have a count of `1`, and a second test be added to test successful deletion should the count be >1.

src/domain/users/users.repository.ts
```diff
@@ -46,4 +52,116 @@ export class UsersRepository implements IUsersRepository {
...
+        const authenticatedWallet = await walletRepository.findOne({
+          where: { address: authPayload.signer_address },
+          relations: { user: true },
+        });
+
+        if (!authenticatedWallet?.user) {
+          throw new NotFoundException('User not found');
+        }
+
+        const wallets = await walletRepository.findBy({
```
PooyaRaki, line 78: These two queries seem a bit confusing. Why not simply fetch all `wallets` by `address` instead of retrieving the user first and then looking up `wallets` by the `user`? Since the foreign key is set to `cascade`, there shouldn’t be any orphaned `wallets`, right?

src/domain/users/users.repository.ts
PooyaRaki: The type of `affected` is `number | null`, so I’d leave it as is to account for both cases.

src/domain/users/users.repository.ts
```diff
@@ -46,4 +52,116 @@ export class UsersRepository implements IUsersRepository {
...
+            id: wallet.id,
+            address: wallet.address,
+          })),
+        };
+      },
+    );
+  }
+
+  async deleteUser(authPayload: AuthPayload): Promise<void> {
+    if (!authPayload.signer_address) {
```
PooyaRaki, line 95: Nit: Since we’re checking the address in multiple places, I’d suggest creating a `validateSignerAddress` function to handle validation and throw an error if needed.

src/domain/users/users.repository.spec.ts
usame-algan: Its strange because we are setting up the mock data ourselves and know that the signer address is there. What do you think about using `!` instead?

src/domain/users/users.repository.spec.ts
usame-algan: In this case we want to check that even though other wallets exist (count > 1) the wallet we are trying to delete does not so essentially testing the `deleteResult.affected` case at the very end.

src/domain/users/users.repository.ts
usame-algan: We need to fetch all wallets that belong to a given user but our starting point is the access token which contains one of the users wallet. Is there a way to retrieve all other wallets of that user without fetching the user?

src/domain/users/users.repository.ts
PooyaRaki, line 114: If we define the opposite direction of the relationship (i.e., OneToMany in the User entity), we can combine both queries:
`await userRepository.delete({ wallet: { address: authPayload.signer_address } });`

However, the key point is that we should analyze the queries generated by the ORM in both scenarios to determine which approach is more efficient.

src/domain/users/users.repository.ts
usame-algan: As discussed we might include the user id in the access token at a later time ([Notion task](https://www.notion.so/safe-global/Extend-auth-verify-to-include-user-id-1848180fe57380baac95fd117d3dce3c)) which would reduce the queries so lets revisit it then.

PooyaRaki: There is a way to do that, but in my opinion, it would make things even more complicated. I’ll go ahead and resolve the thread for now.

src/domain/users/users.repository.spec.ts
usame-algan: Used a mock address instead

src/domain/users/users.repository.ts
usame-algan: I added the OneToMany column to the user entity and update the repository functions and tests.

src/domain/users/users.repository.ts
usame-algan: Do you have a suggestion on where to keep this function?

src/domain/users/users.repository.ts
usame-algan: https://github.com/safe-global/safe-client-gateway/pull/2310/commits/1e7660d1ec6956f61c98b2b0417c6869479a9b1f

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,231 @@ describe('UsersRepository', () => {
...
+  describe('getUser', () => {
+    it('should return user information and associated wallets', async () => {
+      const authPayloadDto = authPayloadDtoBuilder().build();
+      const authPayload = new AuthPayload(authPayloadDto);
+
+      const mockUser = userBuilder().with('id', 1).build();
+
+      const mockAuthenticatedWallet = walletBuilder()
+        .with('user', mockUser)
+        .with('address', authPayload.signer_address!)
```
iamacook, line 194: Nit: if we pass an address to `authPayloadDtoBuilder`, we can use that directly and not need to use a bang.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,231 @@ describe('UsersRepository', () => {
...
+        }),
+      ).rejects.toThrow(
+        new NotFoundException(
+          `A wallet with address ${addressToRemove} does not exist for the current user`,
+        ),
+      );
+    });
+  });
+
+  describe('getUser', () => {
```
iamacook, line 185: We should also test when there is no presence of `authPayload.signer_address`.

src/domain/users/users.repository.ts
iamacook: It can be a private function in this repository.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,231 @@ describe('UsersRepository', () => {
...
+
+      mockWalletRepository.findOne.mockResolvedValueOnce(null);
+
+      await expect(usersRepository.getUser(authPayload)).rejects.toThrow(
+        new NotFoundException('User not found'),
+      );
+    });
+  });
+
+  describe('deleteUser', () => {
```
iamacook, line 255: We should also test when there is no presence of `authPayload.signer_address`.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,231 @@ describe('UsersRepository', () => {
       expect(result).toEqual({ id: mockUser.id });
     });
   });
+
+  describe('deleteWalletFromUser', () => {
```
iamacook, line 76: We should also test when there is no presence of `authPayload.signer_address`.

src/domain/users/users.repository.spec.ts
```diff
@@ -63,4 +72,231 @@ describe('UsersRepository', () => {
...
+      });
+
+      expect(mockWalletRepository.delete).toHaveBeenCalledWith({
+        user: { id: mockUser.id },
+        address: addressToRemove,
+      });
+    });
+
+    it('should throw an UnauthorizedException if the auth payload is empty', async () => {
+      const authPayload = new AuthPayload();
```
iamacook, line 100: Nit: I think it'd be clearer if we use the `authPayloadBuilder` and explicitly define `signer_address` as `undefined`.

iamacook (approved):

PR #2322, opened 2025-02-04
fix: add integration tests for user domain

iamacook:
## Summary

We test the user (and wallet) domain via controller tests. This adds integration tests for the repositories of the aforementioned controller, covering database-relevant cases.

## Changes

- Make `Wallet['user_id']` non-nullable
- Add integration test for `UserRepository`
- Add integration test ffor `WalletRepository`

PR #2328, opened 2025-02-05
refactor: create generic entity repository

iamacook:
⚠️ Branches off #2322

## Summary

We we progress with development, we are starting to see that entity-specific repositories require wrappers around the `find` and `findOne` methods of TypeORM.

This creates a new base class, `EntityRepository` which accepts a generic of a given entity (`T`). It exposes methods of the previously mentioned pattern:

```ts
getRepository(): Promise<Repository<T>>

findOneOrFail(args: FindOneOptions<T>): Promise<T>

findOne(args: FindOneOptions<T>): Promise<T | null>

findOrFail(args?: FindManyOptions<T>): Promise<[T, ...Array<T>]>

find(args?: FindManyOptions<T>): Promise<Array<T>>
```

## Changes

- Adds a new base `EntityRepository` class
- Extends `Wallet` and `User` repositories from base class
- Adds/updates test coverage accordingly


src/domain/wallets/wallets.repository.ts
```diff
@@ -5,69 +5,24 @@ import type {
...
-    select?: FindOptionsSelect<Wallet>;
-    relations?: FindOptionsRelations<Wallet>;
-  }): Promise<Array<Wallet>> {
-    const walletRepository =
-      await this.postgresDatabaseService.getRepository(Wallet);
-
-    return await walletRepository.find(args);
+    readonly postgresDatabaseService: PostgresDatabaseService,
+  ) {
+    super(postgresDatabaseService, Wallet);
```
PooyaRaki, line 25: I like the idea of creating a new repository to customize some of the `TypeOrm` repository methods, even though there is some overlap with the existing `TypeOrm` repository. 👏

That said, I have a few points to consider:
- The `entity` property is inherited from the parent class, which isn’t immediately obvious. At first glance, it was unclear where it originated.
- The `postgresqlDatabaseService` is only used here, solely to pass it to the parent, which seems unnecessary.

TypeOrm provides a decorator called `InjectRepository`. My suggestion is to create a new decorator—something like `InjectEntityRepository`—that takes the required parameters. We could then use the decorated property instead of relying on the `entity` prop from the parent.

Here’s a rough idea of what that might look like in pseudo-code:
```ts
@InjectEntityRepository(postgresDatabaseService, Wallet)
private readonly walletRepository: IEntityRepository<Wallet>;
```

iamacook: As per our discussion, we will revisit the approach to instantiate/inject the refactored repository after completion of the (user-)organizations so as to progress with development of org. onboarding.

As a general note: these methods are heavily tied to datasource actions and, following the current pattern of the project, we will move these to the datasource layer (taking inspiration from the cache, for example).

Therefore, the "TODO" list for this is to:

- Investigate the possibility of (due to async repository) and potentially implement decorator injection.
- Add generic CRUD methods, sans transaction-specific ones as they are very specific (for now).
- Remove existing generic methods from user, wallet and (user-)organization repositories.
- Keep domain-specific methods in respective repositories, including auth. checks, moving generic ones to the datasource layer.

As we intend to address this later, I will mark this as a draft.

PR #2329, opened 2025-02-06
feature: Add organization repository functionality

PooyaRaki:
## Summary
This PR Introduces a new repository for managing organizations, including methods for creating, finding, updating, and deleting organizations. Adjustments made to user organization relationships and repository interfaces to support these operations.

## Changes
- Add CRUD to Organizations repository
- Change findByAddress method name in user repository

PR #2330, opened 2025-02-06
feture: Implement organization management DTOs and controller methods

PooyaRaki:
## Summary
Introduce DTOs and controller and service methods for creating, retrieving, updating, and deleting organizations, enhancing the organization management functionality.

This PR depends on #2329

## Changes
- Added Organizations controller
- Added Organizations service
- Added Organization CRUD DTOs


src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+  @ApiUnauthorizedResponse({ description: 'Signer address not provided' })
+  public async create(
+    @Body(new ValidationPipe(CreateOrganizationSchema))
+    body: CreateOrganizationDto,
+    @Auth() authPayload: AuthPayload,
+  ): Promise<CreateOrganizationResponse> {
+    return await this.organizationsService.create({
+      authPayload,
+      name: body.name,
+      status: OrganizationStatus.ACTIVE,
```
iamacook, line 64: Nit: I'd move this to the service.

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+export class OrganizationsController {
+  public constructor(
+    private readonly organizationsService: OrganizationsService,
+  ) {
+    //
+  }
+
+  @Post()
+  @ApiOkResponse({
+    description: 'Organizations created',
```
iamacook, line 51:
```suggestion
    description: 'Organizations created',
    type: CreateOrganizationResponse
```

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+
+  @Patch('/:id')
+  @ApiOkResponse({
+    description: 'Organization updated',
+  })
+  @ApiForbiddenResponse({ description: 'Forbidden resource' })
+  @ApiUnauthorizedResponse({
+    description: 'Signer address not provided OR User is unauthorized',
+  })
+  public async update(
```
iamacook, line 106:
```suggestion
  @ApiNotFoundResponse({ description: 'User not found.' })
  public async update(
```

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+    @Param('id', ParseIntPipe, new ValidationPipe(RowSchema.shape.id))
+    id: number,
+    @Auth() authPayload: AuthPayload,
+  ): Promise<GetOrganizationResponse> {
+    return await this.organizationsService.getOne(id, authPayload);
+  }
+
+  @Patch('/:id')
+  @ApiOkResponse({
+    description: 'Organization updated',
```
iamacook, line 100:
```suggestion
    description: 'Organization updated',
    type: UpdateOrganizationResponse
```

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+  @ApiUnauthorizedResponse({ description: 'Signer address not provided' })
+  public async get(
+    @Auth() authPayload: AuthPayload,
+  ): Promise<Array<GetOrganizationResponse>> {
+    return await this.organizationsService.get(authPayload);
+  }
+
+  @Get('/:id')
+  @ApiOkResponse({
+    description: 'Organization found',
```
iamacook, line 83:
```suggestion
    description: 'Organization found',
    type: GetOrganizationResponse
```

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+    return await this.organizationsService.create({
+      authPayload,
+      name: body.name,
+      status: OrganizationStatus.ACTIVE,
+    });
+  }
+
+  @Get()
+  @ApiOkResponse({
+    description: 'Organizations found',
```
iamacook, line 70:
```suggestion
    description: 'Organizations found',
    type: GetOrganizationResponse,
    isArray: true,
```

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+  @Delete('/:id')
+  @ApiResponse({
+    description: 'Organization deleted',
+    status: HttpStatus.NO_CONTENT,
+  })
+  @ApiForbiddenResponse({ description: 'Forbidden resource' })
+  @ApiUnauthorizedResponse({
+    description: 'Signer address not provided OR User is unauthorized',
+  })
+  @HttpCode(HttpStatus.NO_CONTENT)
```
iamacook, line 128:
```suggestion
  @ApiNotFoundResponse({ description: 'User not found.' })
  @HttpCode(HttpStatus.NO_CONTENT)
```

src/routes/organizations/entities/create-organization.dto.entity.ts
```diff
@@ -0,0 +1,14 @@
+import { Organization } from '@/datasources/organizations/entities/organizations.entity.db';
```
iamacook, line 1: I know previously discussed usage of the database entity directly, but I think we should reconsider it as this breaks the route <> domain <> datasource principle. When we hoist the basic methods to the datasource layer, it might become even more clear.

src/routes/organizations/entities/create-organizations.dto.entity.ts
```diff
@@ -0,0 +1,11 @@
+import { Organization } from '@/datasources/organizations/entities/organizations.entity.db';
```
iamacook, line 1: As before re. usage of the datasource entity on the domain layer.

src/routes/organizations/entities/create-organizations.dto.entity.ts
```diff
@@ -0,0 +1,11 @@
...
+import { ApiProperty } from '@nestjs/swagger';
+import { number } from 'zod';
+
+export class CreateOrganizationResponse {
+  @ApiProperty({ type: String })
+  public readonly name!: Organization['name'];
+
+  @ApiProperty({ type: number })
```
iamacook, line 9: `z.number` is a validation method. (I'm fairly certain we don't need specify the `type` of primitives though.)

```suggestion
  @ApiProperty({ type: Number })
```

src/routes/organizations/entities/get-organization.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+  UserOrganizationStatus,
+} from '@/domain/users/entities/user-organization.entity';
+import { UserStatus } from '@/domain/users/entities/user.entity';
+import { ApiProperty } from '@nestjs/swagger';
+
+class UserDto {
+  @ApiProperty({ type: Number })
+  public id!: User['id'];
+
+  @ApiProperty({ enum: UserStatus, type: Number })
```
iamacook, line 16: The API spec. specifies that this should be the key value.

We could use the [new helper](https://github.com/safe-global/safe-client-gateway/blob/952b8420c6c1a76e481329b8154cccc60e7afa83/src/domain/common/utils/enums.ts) to retrieve this or, considering the clients want to interface purely with the keys, `transform` directly `to`/`from` on the entity. What do you think?

src/routes/organizations/entities/get-organization.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+  public status!: User['status'];
+}
+
+class UserOrganizationsDto {
+  @ApiProperty({ type: Number })
+  public id!: UserOrganization['id'];
+
+  @ApiProperty({ type: Number, enum: UserOrganizationRole })
```
iamacook, line 24: As before regarding the key value in the spec.

src/routes/organizations/entities/get-organization.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+  public role!: UserOrganization['role'];
+
+  @ApiProperty({ type: Number, enum: UserOrganizationStatus })
+  public status!: UserOrganization['status'];
```
iamacook, line 28: As before regarding the key value in the spec.

src/routes/organizations/entities/get-organization.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+
+  @ApiProperty({ type: Date })
+  public created_at!: UserOrganization['created_at'];
+
+  @ApiProperty({ type: Date })
+  public updated_at!: UserOrganization['updated_at'];
```
iamacook, line 34: Nit: these should be camelcase.

src/routes/organizations/entities/get-organization.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+export class GetOrganizationResponse {
+  @ApiProperty({ type: Number })
+  public id!: Organization['id'];
+
+  @ApiProperty({ type: String })
+  public name!: Organization['name'];
+
+  @ApiProperty({ type: Number, enum: OrganizationStatus })
```
iamacook, line 47: As before regarding the key value in the spec.

src/routes/organizations/entities/update-organization.dto.entity.ts
```diff
@@ -0,0 +1,18 @@
...
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+
+export class UpdateOrganizationDto {
+  @ApiPropertyOptional({ type: String })
+  public readonly name?: Organization['name'];
+
+  @ApiPropertyOptional({
+    enum: OrganizationStatus,
+  })
+  public readonly status?: OrganizationStatus;
```
iamacook, line 12: As before regarding the key value in the spec.

src/routes/organizations/entities/get-organization.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+import { User } from '@/datasources/users/entities/users.entity.db';
+import { OrganizationStatus } from '@/domain/organizations/entities/organization.entity';
+import {
+  UserOrganizationRole,
+  UserOrganizationStatus,
+} from '@/domain/users/entities/user-organization.entity';
+import { UserStatus } from '@/domain/users/entities/user.entity';
+import { ApiProperty } from '@nestjs/swagger';
+
+class UserDto {
```
iamacook, line 12: We could `implement` those properties from the `User` to ensure further type safety. The same could be said for the other DTOs here but I won't mark them.

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+  CreateOrganizationDto,
+  CreateOrganizationSchema,
+} from '@/routes/organizations/entities/create-organization.dto.entity';
+import { GetOrganizationResponse } from '@/routes/organizations/entities/get-organization.dto.entity';
+import {
+  UpdateOrganizationDto,
+  UpdateOrganizationResponse,
+} from '@/routes/organizations/entities/update-organization.dto.entity';
+import { ValidationPipe } from '@/validation/pipes/validation.pipe';
+import { RowSchema } from '@/datasources/db/v1/entities/row.entity';
```
iamacook, line 38: Nit: I'd add a new line after imports.
```suggestion
import { RowSchema } from '@/datasources/db/v1/entities/row.entity';

```

src/routes/organizations/organizations.service.ts
```diff
@@ -1 +1,169 @@
...
+        },
+      },
+    });
+
+    if (!organization) {
+      throw new UnauthorizedException(
+        'User is unauthorized. SignerAddress= ' + signerAddress,
+      );
+    }
+  }
```
iamacook, line 168: For visibility: as previously discussed, these kind of assertions should happen on the domain layer. We will address this later.

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -1,3 +1,546 @@
...
+      await request(app.getHttpServer())
+        .post('/v1/organizations')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(403)
+        .expect({
+          statusCode: 403,
+          message: 'Forbidden resource',
+          error: 'Forbidden',
+        });
+    });
```
iamacook, line 151: This is the same as the previous test - it should be testing `401`. I think this can only be tested in an integration.

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -1,3 +1,546 @@
...
+      await request(app.getHttpServer())
+        .get('/v1/organizations')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(403)
+        .expect({
+          statusCode: 403,
+          message: 'Forbidden resource',
+          error: 'Forbidden',
+        });
+    });
```
iamacook, line 300: This is the same as the previous test - it should be testing `401`. I think this can only be tested in an integration.

src/routes/organizations/entities/get-organization.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+  public id!: Organization['id'];
+
+  @ApiProperty({ type: String })
+  public name!: Organization['name'];
+
+  @ApiProperty({ type: Number, enum: OrganizationStatus })
+  public status!: Organization['status'];
+
+  @ApiProperty()
+  public user_organizations!: Array<UserOrganizationsDto>;
```
iamacook, line 51: Nit: this should be camelcase.

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -1,3 +1,546 @@
...
```
iamacook, line 411: This is the same as the previous test - it should be testing `401`. I think this can only be tested in an integration.

src/routes/organizations/organizations.controller.ts
```diff
@@ -1,6 +1,136 @@
...
+  @Patch('/:id')
+  @ApiOkResponse({
+    description: 'Organization updated',
+  })
+  @ApiForbiddenResponse({ description: 'Forbidden resource' })
+  @ApiUnauthorizedResponse({
+    description: 'Signer address not provided OR User is unauthorized',
+  })
+  public async update(
+    @Body() payload: UpdateOrganizationDto,
```
iamacook, line 107: We should validate this.

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -1,3 +1,546 @@
...
+  });
+
+  describe('PATCH /organizations/:id', () => {
```
iamacook, line 414: We should also add an auth test.

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -1,3 +1,546 @@
...
+          }),
+        );
+    });
+
+    it.todo(
+      'Should throw a 401 if user does not have access to update an organization',
+    ); // @todo Needs membership logic
+  });
+
+  describe('DELETE /organizations/:id', () => {
```
iamacook, line 475: We should also add an auth test.

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -1,3 +1,546 @@
...
+      await request(app.getHttpServer())
+        .delete(`/v1/organizations/${organizationId}`)
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(403)
+        .expect({
+          statusCode: 403,
+          message: 'Forbidden resource',
+          error: 'Forbidden',
+        });
+    });
```
iamacook, line 544: This is testing the guard - it should be testing `401`. I think this can only be tested in an integration.

src/routes/organizations/organizations.service.ts
```diff
@@ -1 +1,169 @@
...
+  ) {}
+
+  public async create(args: {
+    name: Organization['name'];
+    status: Organization['status'];
+    authPayload: AuthPayload;
+  }): Promise<CreateOrganizationResponse> {
+    this.assertSignerAddress(args.authPayload);
+    const { id: userId } = await this.userRepository.findByWalletAddressOrFail(
+      args.authPayload.signer_address,
```
usame-algan, line 29: According to the [spec](https://www.notion.so/safe-global/Data-flows-1858180fe5738004a4adc09d03ae4b7a?pvs=4#1858180fe57380b9a661dfeb8b79c010), if a user account for the given wallet doesn't exist yet we should create one when creating an org. I see multiple possible ways we can achieve this:

1. Create a user here first if it doesn't exist yet
2. A new org endpoint for `createWithUser` -> Don't think we need this endpoint then
3. Handle it in the frontend -> First send a request to the user creation endpoint and then to this endpoint

We should prevent orphaned entries so I am more in favour for 1 or 2 because we can rollback on the repository if necessary. Any other ideas?

src/routes/organizations/organizations.controller.ts
PooyaRaki: Since everything should go through the controller, I think the best place for it is in the controller.

src/routes/organizations/entities/create-organization.dto.entity.ts
PooyaRaki: As discussed, we have this pattern in other components like `users` but I agree with you. Let's change it in another PR and for all the modules.

src/routes/organizations/entities/create-organizations.dto.entity.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: Since the DTO is fetching the types from `User` it's type safe. Could you explain how we can improve the code?

src/routes/organizations/organizations.service.ts
PooyaRaki: In my opinion it should be in the service file but as long as it's not inside the repo I'm happy.

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: We’re testing for a `403 Forbidden` response triggered by `AuthPayload`. The test expects `signer_address` to be empty because it’s defined as `nullable`. While `signer_address` should not be `nullable` in practice, we’re testing this scenario because the parameter is optional. I don't think it should be moved to the integration tests as it's checking the controller response and also we currently don’t have separate integration test suites for component interactions—only integration tests for the database are in place and we keep such tests in the controller tests as it's testing controller.

https://github.com/safe-global/safe-client-gateway/blob/69dd3ea77068f764cbf2f98404d223e584a16148/src/routes/users/users.controller.spec.ts#L134

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: Could you please tell me what test exactly is missing? Please have a look at these two tests:
https://github.com/safe-global/safe-client-gateway/blob/feature/organizationRoute/src/routes/organizations/organizations.controller.spec.ts#L526

https://github.com/safe-global/safe-client-gateway/blob/feature/organizationRoute/src/routes/organizations/organizations.controller.spec.ts#L498

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: Could you please tell me what tests are missing? Please have a look at these two tests:
https://github.com/safe-global/safe-client-gateway/blob/feature/organizationRoute/src/routes/organizations/organizations.controller.spec.ts#L444

https://github.com/safe-global/safe-client-gateway/blob/feature/organizationRoute/src/routes/organizations/organizations.controller.spec.ts#L472

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/entities/create-organizations.dto.entity.ts
PooyaRaki: The zod method was mistakenly imported. When fetching a variable type from another type, we need to explicitly specify the type.

src/routes/organizations/organizations.controller.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/organizations.controller.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/organizations.controller.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/organizations.controller.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/organizations.controller.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/organizations.controller.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/organizations.controller.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: Since these are in snake_case in the database entity, using camelCase here would be confusing. One alternative is to use camelCase in the database entity while explicitly defining the column names to keep them in snake_case. wdyt?

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/organizations.controller.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/2330/commits/c79d0ad56bc7917eb478754c994bb08fec48a743

src/routes/organizations/organizations.service.ts
PooyaRaki: We need the `create` method because there may be `users` who haven’t created an `organization` yet. I’m happy to create a new method to combine the `createUser` and `createOrganization` endpoints, but ideally, we should avoid such cases whenever possible.

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: We previously discussed keeping them as numbers since the clients have access to our types/enums. I’m happy to transform the values if needed, but ideally, this shouldn’t be done at the database entity layer since we might need the raw values later.

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: Please see my previous comment.

src/routes/organizations/entities/get-organization.dto.entity.ts
iamacook: It was my understanding that we agreed to consume/return keys. This is consistent with other enums in the project.

I think this further highlighst the important of following our existing database <> domain pattern, whereby repositories map values from datasources. For now, I suggest we use my aforementioned helper.

src/routes/organizations/entities/get-organization.dto.entity.ts
iamacook: https://github.com/safe-global/safe-client-gateway/pull/2330#discussion_r1950331982

src/routes/organizations/entities/get-organization.dto.entity.ts
iamacook: I see youre point, but I disagree. We follow snakecase as a PostgreSQL naming convention, and the database is only a datasource - it shouldn't dictate internal logic, just as the service shouldn't influence the database structure. This is pattern is [consistent across datasources](https://github.com/safe-global/safe-client-gateway/blob/main/src/routes/transactions/mappers/common/native-staking.mapper.ts#L94-L97).

src/routes/organizations/entities/get-organization.dto.entity.ts
iamacook: https://github.com/safe-global/safe-client-gateway/pull/2330#discussion_r1950331982

src/routes/organizations/entities/get-organization.dto.entity.ts
iamacook: It is type safe - it was a suggestion to further improve it. If we implement the existing type, missing/mismatching types would error:

image

I'll leave this up to your discretion.

src/routes/organizations/organizations.service.ts
iamacook: > as long as it's not inside the repo I'm happy

This flouts the idea that route layer never interfaces with the datasource.

If it's paramount this be called on the route layer, I suggest the helper be defined on the domain layer at least. What do you think? (Again, this will be addressed later.)

src/routes/organizations/organizations.controller.spec.ts
iamacook: As the highlighted tests check for un `undefined` `signer_address`, I assumed it was in reference to `assertSignerAddress`, which [throws a 401](https://github.com/safe-global/safe-client-gateway/blob/c79d0ad56bc7917eb478754c994bb08fec48a743/src/routes/organizations/organizations.service.ts#L140).

This is why I suggested moving it to the integration tests - since, as you mentioned `signer_address` can bever be `undefined` due to the guard in place.

The defined scenario is already covered in multiple tests within the guard, auth and SiWe repositories. However, if we want to keep this, I have no issue.

src/routes/organizations/organizations.controller.spec.ts
iamacook: https://github.com/safe-global/safe-client-gateway/pull/2330#discussion_r1950332721

src/routes/organizations/entities/get-organization.dto.entity.ts
iamacook: https://github.com/safe-global/safe-client-gateway/pull/2330#discussion_r1950332352

src/routes/organizations/organizations.controller.spec.ts
iamacook: https://github.com/safe-global/safe-client-gateway/pull/2330#discussion_r1950332721

src/routes/organizations/organizations.controller.spec.ts
iamacook: All tests include access tokens. We should test the case when there is a missing one.

src/routes/organizations/organizations.controller.spec.ts
iamacook: https://github.com/safe-global/safe-client-gateway/pull/2330#discussion_r1950332721

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/blob/feature/organizationRoute/src/routes/organizations/organizations.controller.spec.ts#L93

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/blob/feature/organizationRoute/src/routes/organizations/organizations.controller.spec.ts#L93

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: Let's keep the test. We can clean them up once the `signer_address` is not `nullable`.

src/routes/organizations/organizations.controller.spec.ts
iamacook: [Other tests](https://github.com/safe-global/safe-client-gateway/blob/feature/organizationRoute/src/routes/organizations/organizations.controller.spec.ts#L125-L134) explicitly check the access token.

src/routes/organizations/entities/create-organization.dto.entity.ts
iamacook: As discussed, we will investigate this later as we can theoretically trust the datasource entities as is. However, we may want to consider keeping schema-inferred entities on the domain layer for things such as unknown enum values (as we store them as integers).

As such, I'll resolve this and we can revisit/discuss this in the [proposed refactor PR](https://github.com/safe-global/safe-client-gateway/pull/2328) we have open.

src/routes/organizations/entities/get-organization.dto.entity.ts
iamacook: Summarising our discussion: TypeORM can map the property on the entity level for us. (I will also change this on the user organisation PR as well.)

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/entities/update-organization.dto.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

src/routes/organizations/entities/get-organization.dto.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/2330/commits/b296b0796c3ea8941ab67b93582208482b5341d5

iamacook (approved):

iamacook (approved):

PR #2342, opened 2025-02-12
chore: add staging CoW Swap API

iamacook:
## Summary

This adds the CoW Swap API URLs for staging so as to be earlier aware of any API changes.

## Changes

- Add condiitonal API for all CoW Swap API URLs

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -384,11 +384,26 @@ export default () => ({
   },
   swaps: {
     api: {
-      1: 'https://api.cow.fi/mainnet',
-      100: 'https://api.cow.fi/xdai',
-      8453: 'https://api.cow.fi/base',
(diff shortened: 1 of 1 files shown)


src/config/entities/configuration.ts
```diff
@@ -384,11 +384,26 @@ export default () => ({
...
     api: {
-      1: 'https://api.cow.fi/mainnet',
-      100: 'https://api.cow.fi/xdai',
-      8453: 'https://api.cow.fi/base',
-      42161: 'https://api.cow.fi/arbitrum_one',
-      11155111: 'https://api.cow.fi/sepolia',
+      1:
+        process.env.CGW_ENV !== 'production'
+          ? 'https://barn.api.cow.fi/mainnet'
+          : 'https://api.cow.fi/mainnet',
```
iamacook, line 390: This is inverse so as to default to prod. should the env. var. not be present.

PR #2344, opened 2025-02-12
feat: Add create organization endpoint with user handling

PooyaRaki:
## Summary
This PR introduces an endpoint to create an organization, handling cases where the user exists or not. Implement user retrieval and creation logic within the organization service.

## Changes
- Introduced a new endpoint to create a user if they don’t already exist when a new organization is created.
- Added unit tests for the new endpoint
- Added a new `findByWallet` method to `UserRepository`

### DIFF
--- a/src/routes/organizations/organizations.controller.spec.ts
+++ b/src/routes/organizations/organizations.controller.spec.ts
@@ -188,6 +188,95 @@ describe('OrganizationController', () => {
     });
   });
 
+  describe('POST /v1/organizations/create-with-user', () => {
+    it('Should create an organization when user exists', async () => {
+      const authPayloadDto = authPayloadDtoBuilder().build();
+      const accessToken = jwtService.sign(authPayloadDto);
+      const organizationName = faker.company.name();
+
+      await request(app.getHttpServer())
+        .post('/v1/users/wallet')
+        .set('Cookie', [`access_token=${accessToken}`]);
+
+      await request(app.getHttpServer())
+        .post('/v1/organizations/create-with-user')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .send({ name: organizationName })
+        .expect(201)
+        .expect(({ body }) =>
+          expect(body).toEqual({
+            id: expect.any(Number),
+            name: organizationName,
+          }),
+        );
+    });
+
+    it('Should create an organization with user does not exist', async () => {
+      const authPayloadDto = authPayloadDtoBuilder().build();
+      const accessToken = jwtService.sign(authPayloadDto);
+      const organizationName = faker.company.name();
+
+      await request(app.getHttpServer())
+        .post('/v1/organizations/create-with-user')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .send({ name: organizationName })
+        .expect(201)
+        .expect(({ body }) =>
+          expect(body).toEqual({
+            id: expect.any(Number),
+            name: organizationName,
+          }),
+        );
+    });
+
+    it('should return a 403 if not authenticated', async () => {
+      await request(app.getHttpServer())
+        .post('/v1/organizations/create-with-user')
+        .expect(403)
+        .expect({
+          statusCode: 403,
+          message: 'Forbidden resource',
+          error: 'Forbidden',
+        });
+    });
+
+    it('Should return a 403 if the AuthPayload is empty', async () => {
+      const authPayloadDto = authPayloadDtoBuilder()
+        .with('signer_address', undefined as unknown as `0x${string}`)
+        .build();
+      const accessToken = jwtService.sign(authPayloadDto);
+
+      await request(app.getHttpServer())
+        .post('/v1/organizations/create-with-user')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(403)
+        .expect({
+          statusCode: 403,
+          message: 'Forbidden resource',
+          error: 'Forbidden',
+        });
+    });
+
+    it('Should return a 422 if no name is provided', async () => {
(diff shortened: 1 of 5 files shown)


src/routes/organizations/organizations.controller.ts
```diff
@@ -69,6 +70,26 @@ export class OrganizationsController {
...
+  public async createWithUser(
+    @Body(new ValidationPipe(CreateOrganizationSchema))
+    body: CreateOrganizationDto,
+    @Auth() authPayload: AuthPayload,
+  ): Promise<CreateOrganizationResponse> {
+    return await this.organizationsService.createWithUser({
+      authPayload,
+      name: body.name,
+      status: getEnumKey(OrganizationStatus, OrganizationStatus.ACTIVE),
+      userStatuus: getEnumKey(UserStatus, UserStatus.ACTIVE),
```
iamacook, line 89: There's a spelling mistake (I'll only mark it here but will need to be propagated).

```suggestion
      status: getEnumKey(OrganizationStatus, OrganizationStatus.ACTIVE),
      userStatus: getEnumKey(UserStatus, UserStatus.ACTIVE),
```

iamacook: Nit: the string values are sufficient here, and are type safe.

src/routes/organizations/organizations.controller.ts
```diff
@@ -69,6 +70,26 @@ export class OrganizationsController {
     });
   }
 
+  @Post('/create-with-user')
+  @ApiOkResponse({
+    description: 'Organizations created',
+    type: CreateOrganizationResponse,
+  })
+  @ApiForbiddenResponse({ description: 'Forbidden resource' })
+  @ApiUnauthorizedResponse({ description: 'Signer address not provided' })
```
iamacook, line 79:
```suggestion
  @ApiUnauthorizedResponse({ description: 'Signer address not provided' })
  @ApiConflictResponse({ description: 'Wallet already exists' })
```

src/routes/organizations/organizations.controller.ts
```diff
@@ -69,6 +70,26 @@ export class OrganizationsController {
     });
   }
 
+  @Post('/create-with-user')
+  @ApiOkResponse({
+    description: 'Organizations created',
+    type: CreateOrganizationResponse,
```
iamacook, line 76: @usame-algan, would you benefit from having the new user ID in the response?

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -188,6 +188,95 @@ describe('OrganizationController', () => {
     });
   });
 
+  describe('POST /v1/organizations/create-with-user', () => {
```
iamacook, line 191: We should add a conflict test for existing wallets.

src/routes/organizations/organizations.service.ts
```diff
@@ -33,6 +34,36 @@ export class OrganizationsService {
...
+        authPayload: args.authPayload,
+      });
+
+      userId = user.id;
+    }
+
+    return await this.organizationsRepository.create({
+      userId,
+      ...args,
+    });
```
iamacook, line 64: I think would benefit from being in a transaction. If the org. creation fails, the user/wallet will still be created.

src/routes/organizations/organizations.service.ts
```diff
@@ -1,4 +1,5 @@
 import type { Organization } from '@/datasources/organizations/entities/organizations.entity.db';
+import { User } from '@/datasources/users/entities/users.entity.db';
```
iamacook, line 2: Nit: we should use the domain entity on the route layer.

src/routes/organizations/organizations.controller.ts
PooyaRaki: It won't create a user if the user already exists therefore no conflict will happen.

src/routes/organizations/organizations.controller.spec.ts
PooyaRaki: It won't create a user if the user already exists therefore no conflict will happen.

src/routes/organizations/organizations.service.ts
PooyaRaki: Yes, but we need to change both user and organization create method. I'd do that later in a separate PR as it needs some effort.

src/routes/organizations/organizations.controller.ts
iamacook: `createWithUser` > `createWithWallet` > `assertWalletDoesNotExist` will throw a 409 if the wallet already exists.

src/routes/organizations/organizations.controller.spec.ts
iamacook: Please see my [previous comment](https://github.com/safe-global/safe-client-gateway/pull/2344#discussion_r1952486694).

src/routes/organizations/organizations.controller.ts
PooyaRaki: Fixed the typo. Let's fetch it from the enum.
https://github.com/safe-global/safe-client-gateway/pull/2344/commits/8ffc8a53505c566f8d26e322f155516353c23c9a

src/routes/organizations/organizations.controller.ts
usame-algan: Don't think we use the user id in any of the requests so it would be fine not returning it.

src/routes/organizations/organizations.service.ts
iamacook: Agreed. This can be addressed in the refactor proposal that moved methods to the datasource.

iamacook (approved):

PR #2345, opened 2025-02-13
feat: only allow adding wallets with signature

iamacook:
## Summary

This modifies the current endpoint for adding a `Wallet` to a `User` to expect a SiWe `message` and `signature`, which must pass validation in order to succeed.

The `POST` `/v1/wallet/:walletAddress` endpoint now becomes `/v1/wallet/add`, which requires the original access token cookie, as well as the following request:

```ts
type Request = {
  message: string;
  signature: `0x${string}`
}
```

The `message` is parsed according to EIP-4362 (Sign-In with Ethereum) and the `signature` validated accordingly. The recovered address is that which a `Wallet` is created for.

As with other endpoints, the access token cookie is responsible for finding the `User` - the `address` of the token.

## Changes

- Rename `POST` `/v1/wallet/:walletAddress` to `/v1/wallet/add`, and expect/validate payload
- Add `SiweMessageSchema`, responsible for parsing a string `message` to an object
- Add `ISiweRepository['getValidatedSiweMessage']`, responsible for returning a parsed/validated string `message` as an object

### DIFF
--- a/src/domain/siwe/siwe.repository.ts
+++ b/src/domain/siwe/siwe.repository.ts
@@ -2,14 +2,16 @@ import { ISiweApi } from '@/domain/interfaces/siwe-api.interface';
 import { IConfigurationService } from '@/config/configuration.service.interface';
 import { ISiweRepository } from '@/domain/siwe/siwe.repository.interface';
 import { LoggingService, ILoggingService } from '@/logging/logging.interface';
-import { Inject, Injectable } from '@nestjs/common';
+import { Inject, Injectable, UnauthorizedException } from '@nestjs/common';
 import { verifyMessage } from 'viem';
 import {
   generateSiweNonce,
   parseSiweMessage,
+  SiweMessage,
   validateSiweMessage,
 } from 'viem/siwe';
 import { IBlockchainApiManager } from '@/domain/interfaces/blockchain-api.manager.interface';
+import { SiweMessageSchema } from '@/domain/siwe/entities/siwe-message.entity';
 
 @Injectable()
 export class SiweRepository implements ISiweRepository {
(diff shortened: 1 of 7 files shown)


src/domain/siwe/siwe.repository.ts
```diff
@@ -45,6 +47,36 @@ export class SiweRepository implements ISiweRepository {
...
+      throw new UnauthorizedException('Invalid nonce');
+    }
+
+    await this.siweApi.clearNonce(result.data.nonce);
+
+    const isValidSignature = await verifyMessage({
+      message: args.message,
+      signature: args.signature,
+      address: result.data.address,
+    }).catch(() => false);
```
PooyaRaki, line 70: Nit: to make it more readable we should use `try/catch`

src/routes/users/users.service.ts
```diff
@@ -31,9 +35,16 @@ export class UsersService {
 
   public async addWalletToUser(args: {
     authPayload: AuthPayload;
-    walletAddress: `0x${string}`;
+    siweDto: SiweDto;
```
PooyaRaki, line 38: Nit: Since we already have a guard in the controller, it’s slightly confusing why we’re authenticating again here. I would move this logic to the controller. Additionally, we could create a new guard, such as `SiweGuard`, to use alongside `AuthGuard` for better DRYness and readability. If we add the `SiweGuard` do we need the `AuthGuard`?

PooyaRaki (approved):

src/domain/siwe/siwe.repository.ts
iamacook: As per discussed, this won't be addressed to unblock development.

src/routes/users/users.service.ts
iamacook: > Since we already have a guard in the controller, it’s slightly confusing why we’re authenticating again here

As mentioned, locating this in the service matches the pattern of other user org.-related authentication.

>If we add the SiweGuard do we need the AuthGuard?

The `AuthGuard` is responsible for the connected signer, e.g. the "logged in" `ADMIN`, whereas the payload is from the wallet-to-be-added so both logic is required.

iamacook:
As agreed, we will merge this without addressing the nits to unblock development.

PR #2349, opened 2025-02-14
fix: add tests for enum utils

iamacook:
## Summary

We use numeric enums for performance reasons in our database entities, the values of which are stored as `integer`s. The values are converted to/from their respective keys when interacting with the database.

This adds relevant tests to the helpers, as well as fixing a small issue discovered when writing them.

## Changes

- Throw for `undefined` values in database enum transformer
- Add enum util tests

### DIFF
--- a/src/domain/common/utils/__tests__/enum.spec.ts
+++ b/src/domain/common/utils/__tests__/enum.spec.ts
@@ -1,0 +1,127 @@
+import {
+  databaseEnumTransformer,
+  getEnumKey,
+  getStringEnumKeys,
+} from '@/domain/common/utils/enum';
+
+enum NumericEnum {
+  A,
+  B,
+}
+
+enum NumericEnumExplicit {
+  A = 0,
+  B = 1,
+}
+
+enum StringEnum {
+  A = 'A',
+  B = 'B',
+}
+
+describe('enum utils', () => {
+  describe('getEnumKey', () => {
+    it('should return the key of a numeric enum with no explicit values', () => {
+      const result = getEnumKey(NumericEnum, 1);
+
+      expect(result).toEqual('B');
+    });
+
+    it('should return the key of a numeric enum with explicit values', () => {
+      const result = getEnumKey(NumericEnumExplicit, 1);
+
+      expect(result).toEqual('B');
+    });
+
+    it('should throw for invalid enum values', () => {
+      expect(() => getEnumKey(NumericEnum, 2)).toThrow('Invalid enum value: 2');
+    });
+
+    it('should return the key of a string enum', () => {
+      // Function is inherently typed for number enums
+      const result = getEnumKey(StringEnum, 'B' as unknown as number);
+
+      expect(result).toEqual('B');
+    });
+  });
+
+  describe('databaseEnumTransformer', () => {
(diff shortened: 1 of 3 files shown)


src/domain/common/utils/__tests__/enum.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+    it('should return the key of a string enum', () => {
+      // Function is inherently typed for number enums
+      const result = getEnumKey(StringEnum, 'B' as unknown as number);
+
+      expect(result).toEqual('B');
+    });
+  });
+
+  describe('databaseEnumTransformer', () => {
```
PooyaRaki, line 48: It looks like we’re primarily testing `getEnumKey` here. While I understand the importance of testing this method, I would suggest limiting it to one or two test cases per transformer method, as they all follow the same pattern as the `getEnumKey` tests.

src/domain/common/utils/enum.ts
```diff
@@ -24,8 +24,24 @@ export const databaseEnumTransformer = <T extends NumericEnum>(
   enumObj: T,
 ): ValueTransformer => {
   return {
-    to: (value: keyof typeof enumObj) => enumObj[value],
-    from: (value: number): keyof T => getEnumKey(enumObj, value),
+    to: <K extends keyof T>(key: K): T[K] => {
+      const value = enumObj[key];
+
+      if (value === undefined) {
```
PooyaRaki, line 30: Do you think we might ever have an enum with `undefined` values? A database column may not always be `non-nullable`.

src/domain/common/utils/__tests__/enum.spec.ts
```diff
@@ -0,0 +1,127 @@
...
+
+      expect(result).toEqual('B');
+    });
+
+    it('should throw for invalid enum values', () => {
+      expect(() => getEnumKey(NumericEnum, 2)).toThrow('Invalid enum value: 2');
+    });
+
+    it('should return the key of a string enum', () => {
+      // Function is inherently typed for number enums
```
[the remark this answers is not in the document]
iamacook: Changed in 4b7fe32d46a1452df98f770f047adf748b42d247.

src/domain/common/utils/__tests__/enum.spec.ts
iamacook: Whilst true, I would keep it in case we adjust the internal logic of the transformer. What do you think?

src/domain/common/utils/enum.ts
iamacook: That's hard to predict but I don't think so. We could address that if/when necessary.

src/domain/common/utils/__tests__/enum.spec.ts
PooyaRaki: Sounds good. :thumbs up

src/domain/common/utils/enum.ts
PooyaRaki: I'll resolve it for now. But having `non-nullable` enums is very common and we definitely need to fix it soon.

PooyaRaki (approved):

PR #2351, opened 2025-02-17
refactor: issue token from validated message

iamacook:
⚠️ Branches off #2345

## Summary

We have to methods for Sign-In with Ethereum messages. This refactors the existing to use the simpler method implemented for org. onboard.

The `ISiweRepository` is now only responsible for parsing and validating a submitted `message`/`signature`. The `nonce` of which is compared against cached ones. As the upper limit for token expiration is auth-specific, that is now handled in the `AuthService`.

1. User calls `/nonce` and one is generated, cached and returned to the user.
2. User calls `/verify` and the message is parsed, validated, nonce compared, signature verified and then passed to the `AuthService`. The service ensures an `expirationTime` does not exceed the maxium and issues the token.

Note: on-chain signature verification was removed as it is not being used.

## Changes

- Remove unused code from `ISiweRepository`.
- Move `expirationTime` check to `AuthService`.
- Add `iat` claim to token.
- Update tests accordingly.


src/domain/siwe/siwe.repository.ts
```diff
@@ -75,105 +55,4 @@ export class SiweRepository implements ISiweRepository {
...
-    if (isValidEoaSignature) {
-      return true;
-    }
-
-    // Else, verify hash on-chain using ERC-6492 for smart contract accounts
-    const blockchainApi = await this.blockchainApiManager.getApi(args.chainId);
-    return blockchainApi.verifySiweMessage({
-      message: args.message,
-      signature: args.signature,
-    });
```
iamacook, line 177: This was not being used, not would work with Safe signatures in its current form.

src/domain/auth/auth.repository.interface.ts
```diff
@@ -12,6 +12,7 @@ export interface IAuthRepository {
     options?: {
       exp?: Date;
       nbf?: Date;
+      iat?: Date;
```
iamacook, line 15: Although not needed, we have an `issuedAt` date.

PooyaRaki (dismissed):

PooyaRaki (approved):

PR #2359, opened 2025-02-18
fix: add decorator options to `GetOrganizationResponse['userOrganizations']`

iamacook:
## Summary

As the clients generate types based on our Swagger scheme, we must explicitly define the decorator `type` as the TypeScript type is not correctly inferred.

`GetOrganizationResponse['userOrganizations']` is currently inferred as a `string` by default. This adds the relevant `type`, marking it as an array in accordance with the response.

## Changes

- Add `type` and `isArray` to `GetOrganizationResponse['userOrganizations']`


usame-algan (approved):
LGTM

PooyaRaki (approved):

PR #2365, opened 2025-02-19
fix: add OrganizationSafesRepository tests

hectorgomezv:
## Summary
This adds integration tests for the repositories of the `OrganizationSafe` entity.

## Changes
- Adds `OrganizationSafesRepository` integration tests.

## Summary by CodeRabbit

- **Tests**
	- Introduced a comprehensive suite to validate organization safe management.
	- Ensured operations such as creation, retrieval, updating, and deletion work as expected.
	- Verified correct handling of timestamps and input validations to improve overall system reliability.
	- Enhanced test coverage for error handling and repository behavior with non-existent safes.
	- Utilized mock services and random data generation for diverse testing scenarios.

### DIFF
--- a/src/domain/organizations/organizations-safe.repository.spec.ts
+++ b/src/domain/organizations/organizations-safe.repository.spec.ts
@@ -365,7 +365,7 @@ describe('OrganizationSafesRepository', () => {
   });
 
   describe('findByOrganizationId', () => {
-    it('should return found organizations', async () => {
+    it('should return found organization Safes', async () => {
       const orgSafes = faker.helpers.multiple(
         () => ({
           chainId: faker.string.numeric(),
@@ -390,7 +390,7 @@ describe('OrganizationSafesRepository', () => {
       );
     });
 
-    it('should return empty array if no organizations found', async () => {
+    it('should return empty array if no organization Safes found', async () => {
       await expect(
         orgSafesRepo.findByOrganizationId(
           faker.number.int({ max: DB_MAX_SAFE_INTEGER }),
@@ -455,9 +455,58 @@ describe('OrganizationSafesRepository', () => {
   });
 
   describe('find', () => {
-    it.todo('should return found organizations');
+    it('should return found organization Safes', async () => {
+      const orgSafes = faker.helpers.multiple(
+        () => ({
+          chainId: faker.string.numeric(),
+          address: getAddress(faker.finance.ethereumAddress()),
+        }),
+        { count: { min: 2, max: 5 } },
+      );
+      const org = await dbOrgRepo.insert({
+        status: faker.helpers.arrayElement(
+          getStringEnumKeys(OrganizationStatus),
+        ),
+        name: faker.word.noun(),
+      });
+      const orgId = org.identifiers[0].id as Organization['id'];
+      await orgSafesRepo.create({
+        organizationId: orgId,
+        payload: orgSafes,
+      });
 
-    it.todo('should return empty array if no organizations found');
+      await expect(
+        orgSafesRepo.find({
+          where: { organization: { id: orgId } },
+        }),
+      ).resolves.toEqual(
+        expect.arrayContaining(
+          orgSafes.map(({ chainId, address }) => ({
+            id: expect.any(Number),
+            chainId,
+            address,
+            createdAt: expect.any(Date),
+            updatedAt: expect.any(Date),
+          })),
+        ),
+      );
+    });
+
+    it('should return empty array if no organization Safes found', async () => {
+      const org = await dbOrgRepo.insert({
+        status: faker.helpers.arrayElement(
+          getStringEnumKeys(OrganizationStatus),
+        ),
+        name: faker.word.noun(),
+      });
+      const orgId = org.identifiers[0].id as Organization['id'];
+
+      expect(
+        await orgSafesRepo.find({
+          where: { organization: { id: orgId } },
+        }),
+      ).toEqual([]);
+    });
   });
 
   describe('delete', () => {
--- a/src/domain/organizations/organizations-safe.repository.spec.ts
+++ b/src/domain/organizations/organizations-safe.repository.spec.ts
@@ -657,6 +657,27 @@ describe('OrganizationSafesRepository', () => {
           ],
         }),
       ).rejects.toThrow(new NotFoundException('Organization has no Safes.'));
+    });
+
+    it('should delete found OrganizationSafes and ignore not found', async () => {
+      const orgSafes = faker.helpers.multiple(
+        () => ({
+          chainId: faker.string.numeric(),
+          address: getAddress(faker.finance.ethereumAddress()),
+        }),
+        { count: { min: 2, max: 5 } },
+      );
+      const org = await dbOrgRepo.insert({
+        status: faker.helpers.arrayElement(
+          getStringEnumKeys(OrganizationStatus),
+        ),
+        name: faker.word.noun(),
+      });
+      const orgId = org.identifiers[0].id as Organization['id'];
+      await orgSafesRepo.create({
+        organizationId: orgId,
+        payload: orgSafes,
+      });
 
       // Some are found
       await expect(
--- a/src/domain/organizations/organizations-safe.repository.spec.ts
+++ b/src/domain/organizations/organizations-safe.repository.spec.ts
@@ -510,16 +510,170 @@ describe('OrganizationSafesRepository', () => {
   });
 
   describe('delete', () => {
-    it.todo('should delete an OrganizationSafe');
+    it('should delete an OrganizationSafe', async () => {
+      const org = await dbOrgRepo.insert({
+        status: faker.helpers.arrayElement(
+          getStringEnumKeys(OrganizationStatus),
+        ),
+        name: faker.helpers.arrayElement([
+          faker.word.noun(),
+          faker.word.noun(),
+        ]),
+      });
+      const orgId = org.identifiers[0].id as Organization['id'];
+      const orgSafe = await dbOrgSafeRepo.insert({
+        chainId: faker.string.numeric(),
+        address: getAddress(faker.finance.ethereumAddress()),
+        organization: {
+          id: orgId,
+        },
+      });
+      const orgSafeId = orgSafe.identifiers[0].id as OrganizationSafe['id'];
 
-    it.todo('should delete multiple OrganizationSafes');
+      const orgSafeBefore = await orgSafesRepo.findOrFail({
+        where: { id: orgSafeId },
+      });
+      expect(orgSafeBefore).toEqual(
+        expect.arrayContaining([
+          expect.objectContaining({
+            id: orgSafeId,
+          }),
+        ]),
+      );
 
-    it.todo(
-      'should throw NotFoundException if provided OrganizationSafe is not found',
-    );
+      await orgSafesRepo.delete({
+        organizationId: orgId,
+        payload: [
+          {
+            chainId: orgSafeBefore[0].chainId,
+            address: orgSafeBefore[0].address,
+          },
+        ],
+      });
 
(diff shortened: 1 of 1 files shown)


src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+          address: getAddress(faker.finance.ethereumAddress()),
+        }),
+        { count: { min: 2, max: 5 } },
+      );
+      const org = await dbOrgRepo.insert({
+        status: faker.helpers.arrayElement(
+          getStringEnumKeys(OrganizationStatus),
+        ),
+        name: faker.word.noun(),
+      });
```
iamacook, line 381: `Organization`s can only exist with an `ACTIVE` `ADMIN`. Although it's not necessary on an entity-level, I think we should ensure there are such members present. What do you think? (I won't mark other occurances.)

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+      const orgId = org.identifiers[0].id as Organization['id'];
+      await orgSafesRepo.create({
+        organizationId: orgId,
+        payload: orgSafes,
+      });
```
iamacook, line 386: Should we rely on another method here, or insert the entry via the database repo?

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+
+      await expect(
+        orgSafesRepo.findOrFail({
+          where: { organization: { id: orgId } },
+        }),
+      ).resolves.toEqual(
+        expect.arrayContaining(
```
iamacook, line 428: We don't need to use `expect.arrayContaining` here.

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+        organizationId: orgId,
+        payload: orgSafes,
+      });
+
+      await expect(
+        orgSafesRepo.find({
+          where: { organization: { id: orgId } },
+        }),
+      ).resolves.toEqual(
+        expect.arrayContaining(
```
iamacook, line 483: We don't need to use `expect.arrayContaining` here.

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+  describe('delete', () => {
+    it('should delete an OrganizationSafe', async () => {
+      const org = await dbOrgRepo.insert({
+        status: faker.helpers.arrayElement(
+          getStringEnumKeys(OrganizationStatus),
+        ),
+        name: faker.helpers.arrayElement([
+          faker.word.noun(),
+          faker.word.noun(),
+        ]),
```
iamacook, line 521: Nit: `faker.word.noun` achieves the same as picking it from a randomised array.
```suggestion
        name: faker.word.noun(),
```

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+          id: orgId,
+        },
+      });
+      const orgSafeId = orgSafe.identifiers[0].id as OrganizationSafe['id'];
+
+      const orgSafeBefore = await orgSafesRepo.findOrFail({
+        where: { id: orgSafeId },
+      });
+      expect(orgSafeBefore).toEqual(
+        expect.arrayContaining([
```
iamacook, line 537: We don't need to use `expect.arrayContaining` here.

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+      const orgSafe = await dbOrgSafeRepo.findOneOrFail({
+        where: {
+          id: insertOrgSafeResult.identifiers[0].id as OrganizationSafe['id'],
+        },
+      });
+
+      expect(orgSafe.address).toEqual(checksummedAddress);
+    });
+
+    it('should retrieve non-checksummed addresses, checksummed', async () => {
```
PooyaRaki, line 235:
```suggestion
    it('should update non-checksummed addresses, checksummed', async () => {
```

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+        organization: {
+          id: orgId,
+        },
+      });
+      const orgSafeId = orgSafe.identifiers[0].id as OrganizationSafe['id'];
+
+      const orgSafeBefore = await orgSafesRepo.findOrFail({
+        where: { id: orgSafeId },
+      });
+      expect(orgSafeBefore).toEqual(
```
PooyaRaki, line 536: I expect the assertion to be the final step in a test. There’s no need to check `orgSafeBefore`; it’s sufficient to verify that after delete, the length has changed or, as you’ve done here, check for an error indicating that no `orgSafes` remain.

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+        ),
+        name: faker.word.noun(),
+      });
+      const orgId = org.identifiers[0].id as Organization['id'];
+      await orgSafesRepo.create({
+        organizationId: orgId,
+        payload: orgSafes,
+      });
+      const orgSafeBefore = await orgSafesRepo.findByOrganizationId(orgId);
+      expect(orgSafeBefore).toHaveLength(orgSafes.length);
```
PooyaRaki, line 581: https://github.com/safe-global/safe-client-gateway/pull/2365/files#r1963356458

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+            {
+              chainId,
+              address: getAddress(faker.finance.ethereumAddress()),
+            },
+          ],
+        }),
+      ).rejects.toThrow('Organization has no Safes.');
+    });
+
+    it('should throw NotFoundException if none of the provided OrganizationSafes is found', async () => {
```
PooyaRaki, line 625: I would additionally check the behavior when any of the provided OrganizationSafes is not found.

src/domain/organizations/organizations-safe.repository.spec.ts
iamacook: Yes, but I don't think it's a hard requirement as it's not enforced on the entity. Do you think it's worthwhile?

hectorgomezv: I'm open to add it, but I also think that as it's not enforced on the entity, we would be pushing domain/business rules into the datasource/repository if we added that constraint here, making the test more dependant and less flexible. But as I said, I'll be happy to add it if you think we should.

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: I've changed `orgSafesRepo.create` by `dbOrgSafeRepo.insert`, and therefore an iteration through `orgSafes` is needed. The order of the insertion depends on the promises execution so `expect.arrayContaining` the test would become flaky.

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: I've changed `orgSafesRepo.create` by `dbOrgSafeRepo.insert` here :)

src/domain/organizations/organizations-safe.repository.spec.ts
PooyaRaki: I totally agree with keeping concerns separate. We should get the logic i.e authorization, authentication out of the repository file. Since the logic is here now, it makes sense to test the logic as well but I'm happy with not adding it to every test. At least one test that checks the whole logic would be very beneficial.

hectorgomezv: Thanks for the discussion, guys. Then if we are in an agreement I'll leave these tests as they are. We can review the logic separation in repository test later as a follow-up technical debt task :)

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: Thanks, I've moved the assertion to the end of the test as you suggested. I changed the assertion object to check that `expect(orgSafeBefore).toHaveLength(1);` instead, I think it's clearer, and I prefer to check the item was there before the deletion. Changed in [219808a](https://github.com/safe-global/safe-client-gateway/pull/2365/commits/219808a85cb3e08a769de9ac1b54a8b6c8b44a86)

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: Done in [219808a](https://github.com/safe-global/safe-client-gateway/pull/2365/commits/219808a85cb3e08a769de9ac1b54a8b6c8b44a86)

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: Removed in [219808a](https://github.com/safe-global/safe-client-gateway/pull/2365/commits/219808a85cb3e08a769de9ac1b54a8b6c8b44a86)

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: Changed in [219808a](https://github.com/safe-global/safe-client-gateway/pull/2365/commits/219808a85cb3e08a769de9ac1b54a8b6c8b44a86)

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: Removed in [219808a](https://github.com/safe-global/safe-client-gateway/pull/2365/commits/219808a85cb3e08a769de9ac1b54a8b6c8b44a86)

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -0,0 +1,679 @@
...
+        ),
+        name: faker.word.noun(),
+      });
+      const orgId = org.identifiers[0].id as Organization['id'];
+
+      await expect(
+        orgSafesRepo.findOrFail({
+          where: { organization: { id: orgId } },
+        }),
+      ).rejects.toThrow('Organization has no Safes.');
```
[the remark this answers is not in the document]
hectorgomezv: Good point, I've changed the test expectations in [219808a](https://github.com/safe-global/safe-client-gateway/pull/2365/commits/219808a85cb3e08a769de9ac1b54a8b6c8b44a86) to include `NotFoundException`, thanks!

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: What would be the expected behavior in that case? Should we fail if one or more of the provided OrganizationSafes are not found? @PooyaRaki @iamacook 

The current implementation only fails when **all** are non-existent, because it relies in `findOrFail`, and this method has the inner logic:

```
if (organizationSafes.length === 0) {
  throw new NotFoundException('Organization has no Safes.');
}
```

PooyaRaki: The current implementation fails fast i.e. as soon as one of the Safes in request is not found the service throws.

PooyaRaki: The current implementation fails if no org is found, even if one org is found it tries to delete that one org.

PR #2373, opened 2025-02-20
feat: add `/v1/auth/logout` to clear cookie

iamacook:
## Summary

After a token is issued, it remains set until it the cookie expires. This adds a new `POST` `/v1/auth/logout` endpoint which clears the cookie value and sets it to expire meaning the user is no longer authenticated.

## Changes

- Add `POST` `/v1/auth/logout` route that clears cookie

### DIFF
--- a/src/routes/auth/auth.controller.spec.ts
+++ b/src/routes/auth/auth.controller.spec.ts
@@ -560,4 +560,47 @@ describe('AuthController', () => {
       await expect(cacheService.hGet(cacheDir)).resolves.toBe(undefined);
     });
   });
+
+  describe('POST /v1/auth/logout', () => {
+    it('should log a signer out', async () => {
+      const privateKey = generatePrivateKey();
+      const signer = privateKeyToAccount(privateKey);
+      const nonceResponse = await request(app.getHttpServer()).get(
+        '/v1/auth/nonce',
+      );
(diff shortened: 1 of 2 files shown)


src/routes/auth/auth.controller.spec.ts
```diff
@@ -560,4 +560,47 @@ describe('AuthController', () => {
...
+          message,
+          signature,
+        })
+        .expect(200);
+      await request(app.getHttpServer())
+        .post('/v1/auth/logout')
+        .expect(200)
+        .expect(({ headers }) => {
+          const setCookie = headers['set-cookie'].toString();
+          // access_token has no value and it is set to expire
```
PooyaRaki, line 599: Nit: I'd remove the comment as it is expected that the token is empty at this point.

iamacook: I think it's beneficial as I was initially confused as to why the cookie "key" remained.

PooyaRaki:
Unrelated to this PR but since we are still working on authentication:
I'd remove the `max-age` from the cookie to make it non-persistent because of security reasons.
FYI: [Owasp Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html#expire-and-max-age-attributes)

PooyaRaki (approved):

PR #2393, opened 2025-02-26
feat: refine messageHash verification on message creation

hectorgomezv:
## Summary
This refines incoming off-chain messages before sending its payload to the Safe Transaction Service.

## Changes
- Adds new refinement helper
- Calls helper in messages repository
- Includes feature flag for the refinement

### DIFF
--- a/src/domain/common/utils/safe.ts
+++ b/src/domain/common/utils/safe.ts
@@ -1,5 +1,6 @@
 import semverSatisfies from 'semver/functions/satisfies';
 import { hashMessage, hashTypedData, type TypedDataDefinition } from 'viem';
+import { MessageSchema } from '@/domain/messages/entities/message.entity';
 import type { MultisigTransaction } from '@/domain/safe/entities/multisig-transaction.entity';
 import type { Safe } from '@/domain/safe/entities/safe.entity';
 
@@ -11,12 +12,16 @@ const MESSAGE_PRIMARY_TYPE = 'SafeMessage';
 export function getSafeMessageMessageHash(args: {
   chainId: string;
   safe: Safe;
-  message: string | TypedDataDefinition;
+  message: string | Record<string, unknown>;
 }): `0x${string}` {
+  if (!args.safe.version) {
+    throw new Error('Safe version is required');
+  }
   try {
-    if (!args.safe.version) {
-      throw new Error('Safe version is required');
-    }
+    const message = MessageSchema.shape.message.parse(args.message) as
+      | string
+      | TypedDataDefinition;
+
     return hashTypedData({
       domain: _getSafeDomain({
         chainId: args.chainId,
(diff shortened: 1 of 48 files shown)


src/domain/common/utils/safe.ts
```diff
@@ -1,11 +1,56 @@
...
+            : hashTypedData(args.message),
+      },
+    });
+  } catch {
+    throw new Error('Failed to hash message data');
+  }
+}
+
+// eslint-disable-next-line @typescript-eslint/explicit-function-return-type
+function _getSafeMessageDomain(args: { chainId: string; safe: Safe }) {
```
iamacook, line 41: This is the same as `_getSafeTxDomain`. Let's rename it (and in the tests) and use it instead.

src/domain/common/utils/safe.ts
```diff
@@ -1,11 +1,56 @@
...
+            type: 'bytes',
+          },
+        ],
+      },
+      message: {
+        message:
+          typeof args.message === 'string'
+            ? hashMessage(args.message)
+            : hashTypedData(args.message),
+      },
```
iamacook, line 33: Nit: I'd create a function similar to `_getSafeTxTypesAndMessage` for testing purposes, but this can be done in a follow up.

src/domain/messages/messages.repository.ts
```diff
@@ -52,28 +57,38 @@ export class MessagesRepository implements IMessagesRepository {
...
-    return transactionService.postMessage({
+    const response = await transactionService.postMessage({
       safeAddress: args.safeAddress,
       message: args.message,
       safeAppId: args.safeAppId,
       signature: args.signature,
       origin: args.origin,
     });
+    const message = MessageSchema.parse(response);
+    this.messageVerifier.verifyMessageHash({
```
iamacook, line 75: We should verify before the `POST` otherwise the check is worthless.

src/domain/common/utils/safe.ts
hectorgomezv: Renamed to `_getSafeDomain` and reused in both places in [4139428](https://github.com/safe-global/safe-client-gateway/pull/2393/commits/41394284d223fbef894bae20b2362e375678601d)

src/domain/messages/messages.repository.ts
hectorgomezv: My intention was to check the hash the Transaction Service returns is consistent with the payload sent. Does this make sense to you? Otherwise which hash would we check? The user request does not include the hash if I'm not wrong.

src/domain/messages/helpers/message-verifier.helper.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (!isOwner) {
+      throw new BadGatewayException('Invalid signature');
+    }
+  }
+
+  public async recoverSignerAddress(args: {
+    messageHash: `0x${string}`;
+    signature: `0x${string}`;
+  }): Promise<`0x${string}`> {
+    // TODO: Throw is v is eth_sign and feature disabled
```
iamacook, line 126: We should create a separate signature helper and reuse the logic from the transaction verifier. Let's do it in a follow up though.

iamacook (approved):

PR #2398, opened 2025-02-27
fix: categorize propose  errors

hectorgomezv:
## Summary

This PR unmasks errors incoming from the Transaction Service when `this.contractsRepository.getContract` throws an error during delegate calls trustability tests.

## Changes
- Removes _catch-all_ block from `SafeRepository.proposeTransaction`.
- Adds related tests.

### DIFF
--- a/src/domain/safe/safe.repository.ts
+++ b/src/domain/safe/safe.repository.ts
@@ -538,15 +538,11 @@ export class SafeRepository implements ISafeRepository {
       if (!this.isTrustedDelegateCallEnabled) {
         throw new UnprocessableEntityException('Delegate call is disabled');
       }
-      try {
-        const contract = await this.contractsRepository.getContract({
-          chainId: args.chainId,
-          contractAddress: args.proposeTransactionDto.to,
-        });
-        if (!contract.trustedForDelegateCall) {
-          throw new Error('Delegate call is disabled');
-        }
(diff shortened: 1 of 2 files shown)


PooyaRaki (approved):

src/domain/safe/safe.repository.ts
```diff
@@ -538,15 +538,11 @@ export class SafeRepository implements ISafeRepository {
...
-        });
-        if (!contract.trustedForDelegateCall) {
-          throw new Error('Delegate call is disabled');
-        }
-      } catch {
+      const contract = await this.contractsRepository.getContract({
+        chainId: args.chainId,
+        contractAddress: args.proposeTransactionDto.to,
+      });
+      if (contract && !contract.trustedForDelegateCall) {
```
iamacook, line 545: If a `contract` doesn't exist, this will allow delegate calls.

iamacook (changes requested):

src/domain/safe/safe.repository.ts
hectorgomezv: If the contract does not exist, I think the flow should stop there and the service would throw an error. So no further call to the TXS propose endpoint. But given this is unclear, and it's just a detail (the error message), I'm closing the PR to avoid distractions.

PR #2422, opened 2025-03-04
refactor: create `SafeSignature` entity

iamacook:
## Summary

For every confirmation refined, we reference the owner and type of a signature. This is the same for transactions and messages.

This moves shared logic into a new `SafeSignature` entity, [taking inspiration from `safe-eth-py`](https://github.com/safe-global/safe-eth-py/blob/main/safe_eth/safe/safe_signature.py).

It takes the `hash` signed and `signature`, and exposes the `r`, `s`, `v`, `signatureType` and `owner` properties of them.

## Changes

- Remove existing signature helpers
- Create `SafeSignature` entity
- Use entity in place for transaction/message verification


src/domain/common/entities/safe-signature.ts
```diff
@@ -0,0 +1,127 @@
...
+function toRecoveryBit(signature: `0x${string}`): number {
+  const v = parseInt(signature.slice(-2), 16);
+  if (v === 0 || v === 1) {
+    return v;
+  }
+  if (v === 27 || v === 28) {
+    return v - 27;
+  }
+  throw new Error('Invalid v value');
+}
```
iamacook, line 127: This is the same as viem's implementation, only without a dynamic import so as not need `async`/`await`.

src/routes/transactions/helpers/transaction-verifier.helper.spec.ts
```diff
@@ -305,40 +304,9 @@ describe('TransactionVerifierHelper', () => {
...
 
-        await expect(
-          target.verifyApiTransaction({ chainId, safe, transaction }),
-        ).resolves.not.toThrow();
-      });
-
-      it.each([
-        SignatureType.ApprovedHash as const,
-        SignatureType.ContractSignature as const,
-      ])('should validate a %s signature', async (signatureType) => {
```
iamacook, line 316: As these signatures are now in the builder, they are covered by the `it.each(Object.values(SignatureType))('...` changes.

src/routes/transactions/helpers/transaction-verifier.helper.spec.ts
```diff
@@ -556,29 +524,16 @@ describe('TransactionVerifierHelper', () => {
...
-          expect(mockLoggingRepository.error).toHaveBeenCalledTimes(1);
-          expect(mockLoggingRepository.error).toHaveBeenNthCalledWith(1, {
-            message: 'Could not recover address',
-            chainId,
-            safeAddress: safe.address,
-            safeVersion: safe.version,
-            safeTxHash: transaction.safeTxHash,
-            signature: transaction.confirmations![0].signature,
-            type: 'TRANSACTION_VALIDITY',
-          });
```
iamacook, line 581: This is now inherently part of the `SafeSignature`.

src/domain/common/entities/safe-signature.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  private _owner = memoize(
+    () => {
+      try {
+        switch (this.signatureType) {
+          case SignatureType.ContractSignature:
+          case SignatureType.ApprovedHash: {
+            return uint256ToAddress(this.r);
+          }
+          case SignatureType.EthSign: {
```
hectorgomezv, line 63: Nit: should we document _why_ is it needed to subtract 4 to the `v` value when `SignatureType` is `EthSign`?

PR #2428, opened 2025-03-06
feat: add pgAdmin service to Docker Compose configuration

PooyaRaki:
## Summary
This PR Introduces a pgAdmin service in the Docker Compose setup to facilitate database management. This includes configuration for environment variables, ports, and network settings.

## Changes
- Added PgAdmin to the docker compose file.


docker-compose.yml
```diff
@@ -88,3 +103,11 @@ services:
       APPLICATION_PORT: ${APPLICATION_PORT-3000}
     depends_on:
       - web
+
+volumes:
+  local_pgdata:
```
hectorgomezv, line 108: Why is this volume needed? It seems it's not used anywhere.

Offtopic: maybe we should make `db` and `db-test` containers to use volumes instead, wdyt?

docker-compose.yml
```diff
@@ -88,3 +103,11 @@ services:
...
+  pgadmin-data:
+
+networks:
+  backend-cgw:
+    driver: bridge
```
hectorgomezv, line 113: Why a custom network to communicate the pgAdmin container with the test-db one is needed here?

PooyaRaki: If they are not on the same network, Docker containers cannot communicate with each other. For example, the `pgAdmin` container won’t be able to ping/connect to the `database` container.

docker-compose.yml
PooyaRaki: It was a leftover, I removed it. https://github.com/safe-global/safe-client-gateway/pull/2428/commits/a8fb5cdb9bb76ef97423479c294ae010ac13b9f0
I didn’t quite understand that. What is the benefit of using names volumes for both db and db-test here?

docker-compose.yml
hectorgomezv: It should work without specifying any custom network. All the containers are started in the same _default_ network (something like `safe-client-gateway_default`, you can check with `docker network ls`), and all should be accessible to each other by name. 

I've just tried after removing this `networks: ...` part and I can still access `db` and `db-test` with no errors. If you want to create a custom network to communicate the `pgAdmin` container with `db-test` one I'm not against, but I think it shouldn't be necessary unless there's an specific need :)

hectorgomezv (approved):

db_config/test/pg_hba.conf
```diff
@@ -1,4 +1,5 @@
 # Allow any user on the local system to connect
 # TYPE  DATABASE        USER            ADDRESS                             METHOD
 local   all             all                                                 trust
-hostssl all             all             all                                 md5
+host    all             all             all                                 md5
+hostssl all             all             all                                 md5
```
hectorgomezv, line 5: As discussed async, this change shouldn't be necessary and it's not related to the scope of the PR IMO. 🙂

PR #2430, opened 2025-03-06
feat: optimize trusted contracts retrieval

hectorgomezv:
## Summary
After [the addition of a new filter](https://github.com/safe-global/safe-transaction-service/pull/2448) that makes it possible to get the complete list of trusted for delegate call contracts in the Transaction Service API, the Client Gateway can take advantage of it by caching that list and comparing proposals and transaction mappings against the list, instead of checking each contract on demand.

This PR adds a new function to retrieve the list, and checks against that list instead of calling `ContractRepository.getContract` in both `TransactionDataMapper` and `SafeRepository`.

## Changes
- Adds the `ContractRepository['isTrustedForDelegateCall']` function.
- Adds an `FF_TRUSTED_FOR_DELEGATE_CALL_CONTRACTS_LIST` feature flag.
- Adjusts the related tests.

### DIFF
--- a/src/datasources/transaction-api/transaction-api.service.spec.ts
+++ b/src/datasources/transaction-api/transaction-api.service.spec.ts
@@ -526,7 +526,10 @@ describe('TransactionApi', () => {
   describe('getTrustedForDelegateCallContracts', () => {
     it('should return the trusted for delegate call contracts received', async () => {
       const contractPage = pageBuilder()
-        .with('results', [contractBuilder().build(), contractBuilder().build()])
+        .with('results', [
+          contractBuilder().with('trustedForDelegateCall', true).build(),
+          contractBuilder().with('trustedForDelegateCall', true).build(),
+        ])
         .build();
       const getTrustedForDelegateCallContractsUrl = `${baseUrl}/api/v1/contracts/`;
       const cacheDir = new CacheDir(`${chainId}_trusted_contracts`, '');
--- a/src/datasources/transaction-api/transaction-api.service.spec.ts
+++ b/src/datasources/transaction-api/transaction-api.service.spec.ts
@@ -523,6 +523,72 @@ describe('TransactionApi', () => {
     });
   });
 
+  describe('getTrustedForDelegateCallContracts', () => {
+    it('should return the trusted for delegate call contracts received', async () => {
+      const contractPage = pageBuilder()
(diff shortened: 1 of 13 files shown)


iamacook (approved):

src/datasources/transaction-api/transaction-api.service.spec.ts
```diff
@@ -523,6 +523,72 @@ describe('TransactionApi', () => {
     });
   });
 
+  describe('getTrustedForDelegateCallContracts', () => {
+    it('should return the trusted for delegate call contracts received', async () => {
+      const contractPage = pageBuilder()
+        .with('results', [contractBuilder().build(), contractBuilder().build()])
```
iamacook, line 529: Nit:
```suggestion
        .with('results', [
          contractBuilder().with('trustedForDelegateCall', true).build(),
          contractBuilder().with('trustedForDelegateCall', true).build(),
        ])
```

src/domain/contracts/contracts.repository.ts
```diff
@@ -19,4 +34,50 @@ export class ContractsRepository implements IContractsRepository {
...
+
+  private async isIncludedInTrustedForDelegateCallContractsList(args: {
+    chainId: string;
+    contractAddress: `0x${string}`;
+  }): Promise<boolean> {
+    const trustedContracts = await this.getTrustedForDelegateCallContracts(
+      args.chainId,
+    );
+    return trustedContracts.results.some((contract) =>
+      isAddressEqual(contract.address, args.contractAddress),
```
iamacook, line 69: Althought theoretically correct, I think we should still check the flag to be safe.
```suggestion
      isAddressEqual(contract.address, args.contractAddress) &&
        contract.trustedForDelegateCall,```

hectorgomezv: Agreed. Even if it's not needed theoretically, we can check the flag to avoid the case the filter is not working properly on the Transaction Service for some reason. Thanks. Changed in [7e3b393](https://github.com/safe-global/safe-client-gateway/pull/2430/commits/7e3b393d2742e73aa195cc9f47298be406c860ad)

iamacook: Should we want to expose pages of contracts later on, we are also covered, e.g. we call a hypothetical `getContracts` method on the domain, passing the query there instead.

iamacook (approved):

PR #2432, opened 2025-03-07
fix: add test coverage for `messageHash` generation

iamacook:
## Summary

We are missing test coverage for our method used to generate `messageHash`es for refinement. This adds coverage for `getSafeMessageMessageHash` accordingly.

Note: exact example tests are scaffolded but not completed at the moment. These will be tackled later with `safeTxHash` examples as well.

## Changes

- Add test coverage for `getSafeMessageMessageHash`

### DIFF
--- a/src/domain/common/utils/__tests__/safe.spec.ts
+++ b/src/domain/common/utils/__tests__/safe.spec.ts
@@ -1,11 +1,14 @@
 import { faker } from '@faker-js/faker';
-import { getAddress } from 'viem';
+import { getAddress, type TypedDataDomain } from 'viem';
 import { Builder } from '@/__tests__/builder';
+import { multisigTransactionBuilder } from '@/domain/safe/entities/__tests__/multisig-transaction.builder';
 import { Operation } from '@/domain/safe/entities/operation.entity';
 import { safeBuilder } from '@/domain/safe/entities/__tests__/safe.builder';
 import {
   _getSafeDomain,
   _getSafeTxTypesAndMessage,
+  getBaseMultisigTransaction,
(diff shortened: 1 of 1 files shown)


src/domain/common/utils/__tests__/safe.spec.ts
```diff
@@ -46,9 +86,264 @@ function safeTxHashMultisigTransactionBuilder(): IBuilder<BaseMultisigTransactio
...
+        operation: multisigTransaction.operation,
+        safeTxGas: multisigTransaction.safeTxGas,
+        baseGas: multisigTransaction.baseGas,
+        gasPrice: multisigTransaction.gasPrice,
+        gasToken: multisigTransaction.gasToken,
+        refundReceiver: multisigTransaction.refundReceiver,
+        nonce: multisigTransaction.nonce,
+      });
+    });
+  });
```
iamacook, line 108: This is included here as it is a simple test.

src/domain/common/utils/__tests__/safe.spec.ts
```diff
@@ -46,9 +86,264 @@ function safeTxHashMultisigTransactionBuilder(): IBuilder<BaseMultisigTransactio
...
+          ]);
+
+          expect(() =>
+            getSafeMessageMessageHash({ chainId, message, safe }),
+          ).not.toThrow();
+        },
+      );
+    });
+
+    describe.skip('examples', () => {
```
iamacook, line 173: As mentioned in the desc., these will be added later (as well as those for `safeTxHash`es) because they take some time to find.

hectorgomezv (approved):

PR #2434, opened 2025-03-07
fix: ensure API signatures are even in length

iamacook:
## Summary

We currently validate that transaction signatures from our API are hex and a denomination fo 130 and that message signatures are hex.

Whilst we can be sure that incoming signatures from clients are a denomination of 130 (because the interface does not currently support proposal of contract signatures), we cannot be sure that the Transaction Service hasn't indexed contract signatures with dynamic parts.

This creates a new `HexBytesSchema` which ensures a string is hex and even in length, aka "byte-aligned", applying it to the validation of transaction/message signatures from our API.

## Changes

- Add and use `HexBytesSchema`

### DIFF
--- a/src/validation/entities/schemas/signature.schema.ts
+++ b/src/validation/entities/schemas/signature.schema.ts
@@ -1,4 +1,4 @@
-import { HexSchema } from '@/validation/entities/schemas/hex.schema';
+import { HexBytesSchema } from '@/validation/entities/schemas/hexbytes.schema';
 
 // This does not take dynamic parts into account but we can safely
 // apply it to proposed signatures as we do not support contract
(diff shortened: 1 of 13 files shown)


hectorgomezv (approved):

src/validation/entities/schemas/signature.schema.ts
```diff
@@ -1,10 +1,13 @@
-import { HexSchema } from '@/validation/entities/schemas/hex.schema';
+import { HexBytesSchema } from '@/validation/entities/schemas/hexbytes.schema';
 
+// This does not take dynamic parts into account but we can safely
+// apply it to proposed signatures as we do not support contract
+// signatures in the inferface
```
hectorgomezv, line 5:
```suggestion
// signatures in the interface
```

(Nit, sorry)

hectorgomezv (approved):

hectorgomezv (approved):

PR #2446, opened 2025-03-12
fix: avoid duplicate OrganizationSafes

hectorgomezv:
## Changes
- Adds a new `UNIQUE INDEX` to the `organization_safes` table, affecting `['chainId', 'address']` fields.
- Adds a new general-purpose `UniqueConstraintError` to the codebase, with an attached default `409 - Conflict` status code and its related `ExceptionFilter`.
- Adjusts the related test cases.


src/datasources/errors/unique-constraint-error.ts
```diff
@@ -0,0 +1,21 @@
+import { ConflictException } from '@nestjs/common';
+import { QueryFailedError } from 'typeorm';
+
+// Ref: https://www.postgresql.org/docs/current/errcodes-appendix.html
```
PooyaRaki, line 4: Nit: I wouldn't add a comment here as the code is redable and the name of the variable suggests that it's Postgres error code

src/domain/organizations/organizations-safe.repository.ts
```diff
@@ -35,7 +39,15 @@ export class OrganizationSafesRepository
       address: safe.address,
     }));
 
-    await organizationSafeRepository.insert(safesToInsert);
+    try {
+      await organizationSafeRepository.insert(safesToInsert);
+    } catch (err) {
```
PooyaRaki, line 44: It seems that if the error isn't a UniqueConstraintError, we are ignoring it and not throwing it.

src/domain/organizations/organizations-safe.repository.ts
```diff
@@ -35,7 +39,15 @@ export class OrganizationSafesRepository
       address: safe.address,
     }));
 
-    await organizationSafeRepository.insert(safesToInsert);
+    try {
+      await organizationSafeRepository.insert(safesToInsert);
```
PooyaRaki, line 43: I'm not sure if TypeOrm uses database transactions behind the scenes for the insert method. If it doesn't, we might end up persisting the Safes before encountering an error. Is this the expected behavior?

PooyaRaki (dismissed):

migrations/1741709533269-add-organizations-safe-unique-index.ts
```diff
@@ -0,0 +1,20 @@
...
+  implements MigrationInterface
+{
+  name = 'AddOrganizationsSafeUniqueIndex1741709533269';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DELETE FROM organization_safes WHERE id NOT IN (SELECT MIN(id) FROM organization_safes GROUP BY chain_id, address);`,
+    );
+    await queryRunner.query(
+      `CREATE UNIQUE INDEX idx_OS_unique_chain_id_address ON organization_safes (chain_id, address);`,
```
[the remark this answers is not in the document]
hectorgomezv: Thanks, I didn't participate in that naming convention but theoretically that would allow conflicts, right? I mean, two tables can have an index affecting the same column names. And also, neither `user` nor `organizations` (taken from `UQ_user_organizations`) are column names, shouldn't it be `user_id` and `organization_id`?

I'm not against the `UQ_` prefix, but I'd prefer to include the table name to avoid conflicts, and also the full name of each column, so, I'm renaming this one to `UQ_organizations_safes_chain_id_address`, wdyt?

src/datasources/errors/unique-constraint-error.ts
hectorgomezv: I agree, I thought this code was from Postgres but it seems it's not vendor specific: https://docs.oracle.com/en/database/oracle/oracle-database/19/drdas/error-code.html

Thanks!

src/domain/organizations/organizations-safe.repository.ts
hectorgomezv: This [is translated to](https://github.com/typeorm/typeorm/blob/ba47b6237984ec15c67fe0587f4e56a77ae01903/src/entity-manager/EntityManager.ts#L672) a single `INSERT` statement, so it's atomic by default.

migrations/1741709533269-add-organizations-safe-unique-index.ts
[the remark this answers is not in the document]
hectorgomezv: [41864ba](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/41864babbcaf4762f607547570faec8143784a17)

src/datasources/errors/unique-constraint-error.ts
hectorgomezv: I've removed the comment and I've changed the variable name in [41864ba](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/41864babbcaf4762f607547570faec8143784a17)

src/domain/organizations/organizations-safe.repository.ts
hectorgomezv: Good catch! Fixed in [41864ba](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/41864babbcaf4762f607547570faec8143784a17)

src/datasources/errors/unique-constraint-error.ts
```diff
@@ -0,0 +1,21 @@
...
+  err: unknown,
+): err is QueryFailedError & { driverError: { code: string } } {
+  return (
+    err instanceof QueryFailedError &&
+    'code' in err.driverError &&
+    err.driverError.code === PG_UNIQUE_CONSTRAINT_ERROR_CODE
+  );
+}
+
+export class UniqueConstraintError extends ConflictException {
```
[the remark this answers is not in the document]
hectorgomezv: I've removed the filter in [41864ba](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/41864babbcaf4762f607547570faec8143784a17)

src/routes/organizations/organization-safes.controller.ts
```diff
@@ -48,6 +50,7 @@ export class OrganizationSafesController {
   ) {}
 
   @Post()
+  @UseFilters(UniqueConstraintExceptionFilter)
```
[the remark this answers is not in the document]
hectorgomezv: I've removed the filter in [41864ba](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/41864babbcaf4762f607547570faec8143784a17)

src/domain/relay/exception-filters/unique-constraint.exception-filter.ts
```diff
@@ -0,0 +1,21 @@
...
+import {
+  Catch,
+  ExceptionFilter,
+  ArgumentsHost,
+  HttpStatus,
+} from '@nestjs/common';
+import { UniqueConstraintError } from '@/datasources/errors/unique-constraint-error';
+
+@Catch(UniqueConstraintError)
+export class UniqueConstraintExceptionFilter implements ExceptionFilter {
```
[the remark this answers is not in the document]
hectorgomezv: I've removed the filter in [41864ba](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/41864babbcaf4762f607547570faec8143784a17)

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -365,6 +365,48 @@ describe('OrganizationSafesRepository', () => {
...
+          orgSafesRepo.create({
+            organizationId: orgId,
+            payload: [{ chainId, address }],
+          }),
+          orgSafesRepo.create({
+            organizationId: orgId,
+            payload: [{ chainId, address }],
+          }),
+        ]),
+      ).rejects.toThrow(
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [41864ba](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/41864babbcaf4762f607547570faec8143784a17)

src/datasources/errors/unique-constraint-error.ts
PooyaRaki: Having `PQ` prefix was a very good idea even though it's not only Postgres that uses the same error code. I'll leave it up to you whether we should add the `PQ_` again or not and I'll resolve it.

migrations/1741709533269-add-organizations-safe-unique-index.ts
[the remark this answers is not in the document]
PooyaRaki: The convention is actually in the `TypeOrm` documentation even though it does not talk about naming conventions, please see it [here](https://orkhan.gitbook.io/typeorm/docs/decorator-reference#unique).
The Unique decorator accepts entity names, from the docs:
> You must specify the entity field names (not database column names) as arguments

That's why it's not `user_id` in `UQ_user_organizations`

But, I agree we should change that to:
`UQ_UO_user_organizations `

Back to this PR, I'd suggest we rename the constraint to:
`UQ_OS_chainId_address`

What do you think?

src/domain/organizations/organizations-safe.repository.spec.ts
```diff
@@ -403,9 +404,7 @@ describe('OrganizationSafesRepository', () => {
             payload: [{ chainId, address }],
           }),
         ]),
-      ).rejects.toThrow(
-        'An OrganizationSafe with the same chainId and address already exists.',
-      );
+      ).rejects.toThrow(UniqueConstraintError);
```
PooyaRaki, line 407: Thanks for implementing my suggestion. I would still recommend keeping the message to ensure that, throughout the entire process of adding a new safe to an organization, it's clear that the duplicate is related to the organization's safe and not something earlier in the process.

src/datasources/errors/unique-constraint-error.ts
```diff
@@ -0,0 +1,20 @@
+import { ConflictException } from '@nestjs/common';
+import { QueryFailedError } from 'typeorm';
+
+const UNIQUE_CONSTRAINT_ERROR_CODE = '23505';
+
+export function isUniqueConstraintError(
```
PooyaRaki, line 6: Nit: I would move it to another helper to keep my error files concise and consistent, containing only a single class.

src/domain/organizations/organizations-safe.repository.ts
```diff
@@ -35,7 +39,16 @@ export class OrganizationSafesRepository
...
+    } catch (err) {
+      if (isUniqueConstraintError(err)) {
+        throw new UniqueConstraintError(
+          'An OrganizationSafe with the same chainId and address already exists.',
```
PooyaRaki, line 47: We usually provide more information in the error message, please see [here](https://github.com/safe-global/safe-client-gateway/blob/41864babbcaf4762f607547570faec8143784a17/src/domain/organizations/organizations.repository.ts#L123-L124)

src/domain/organizations/organizations-safe.repository.spec.ts
hectorgomezv: Absolutely! Added back in: [d7158d9](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/d7158d98ea1babae933f4284aad684e9f6ea6980)

src/datasources/errors/unique-constraint-error.ts
hectorgomezv: I moved the helper to a new file `is-unique-constraint-error.helper.ts` in [d7158d9](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/d7158d98ea1babae933f4284aad684e9f6ea6980)

src/domain/organizations/organizations-safe.repository.ts
hectorgomezv: In this case, since the INSERT is affecting multiple rows, the only way to access that information is to use the `driverError.detail` string. I included that in the message directly, as I think parsing it would be overkill, and might break if the `driverError.detail` message format is changed. ([d7158d9](https://github.com/safe-global/safe-client-gateway/pull/2446/commits/d7158d98ea1babae933f4284aad684e9f6ea6980))

src/datasources/errors/helpers/is-unique-constraint-error.helper.ts
```diff
@@ -0,0 +1,13 @@
+import { QueryFailedError } from 'typeorm';
+
+const UNIQUE_CONSTRAINT_ERROR_CODE = '23505';
+
+export function isUniqueConstraintError(
+  err: unknown,
+): err is QueryFailedError & { driverError: { code: string; detail: string } } {
```
iamacook, line 7: Nit: `driverError` extends an `Error` so we can't be certain that `detail` exists:

```suggestion
): err is QueryFailedError & {
  driverError: { code: typeof UNIQUE_CONSTRAINT_ERROR_CODE; detail?: string };
} {
```

iamacook (commented):
This looks good and I won't block it but what do you think about rather adding an `ExceptionFilter` for `QueryFailedError`s? We could then extend it according to other error codes and not need to explicitly check `isUniqueConstraintError` on every query.

iamacook (approved):

PR #2453, opened 2025-03-13
fix: update the UserOrganization name on invitation acceptance

hectorgomezv:
## Changes
- Updates the `UserOrganization` `name` when accepting an invitation to an Organization.
- Adjusts the related test case.


src/domain/users/user-organizations.repository.ts
```diff
@@ -229,25 +226,12 @@ export class UsersOrganizationsRepository
     const userOrg = org.userOrganizations[0];
 
     await this.postgresDatabaseService.transaction(async (entityManager) => {
-      await this.updateStatus({
-        userOrgId: userOrg.id,
+      await entityManager.update(DbUserOrganization, userOrg.id, {
```
PooyaRaki, line 229: Nit: We created a new function to make the code more redable and actually following SOLID principles split the code into small functions that take care of a single responsibility, So I wouldn't remove the `updateStatus` method.

src/domain/users/user-organizations.repository.integration.spec.ts
```diff
@@ -774,7 +774,7 @@ describe('UserOrganizationsRepository', () => {
   });
 
   describe('acceptInvite', () => {
-    it('should accept an invite to an organization, setting the user organization and user to ACTIVE', async () => {
+    it('should accept an invite to an organization, setting the user organization and user to ACTIVE, and overwriting the name', async () => {
```
PooyaRaki, line 777: I would create a new test to check `name` update, it helps split the responsibilities between tests and keep tests as small as possible with less assertions.

src/domain/users/user-organizations.repository.ts
hectorgomezv: I agree, but given the change, that function would be called from just one place, as the other call would need to update both the `status` and the `name` field instead. So I thought having a private function called from just one place does not make much sense, and even more so if the private function is a one-liner.

src/domain/users/user-organizations.repository.integration.spec.ts
hectorgomezv: I agree, thanks! I kept the original test case and added a second one to verify the name change in [6b06c87](https://github.com/safe-global/safe-client-gateway/pull/2453/commits/6b06c875c51772d8214527446630fe46446a7f5e)

PooyaRaki: Thanks for adding the new test looks good to me but a small comment:
Now that the descriptions are the same, I would rename the test on line 855 to:
`should accept an invite to an organization and override the name`
I suggest the change because all the scenarios are tested in the newly added test and in this test we can only assert the `name` is updated to keep the test precise and small.

src/domain/users/user-organizations.repository.ts
PooyaRaki: I see, but as I see in the code we would still call it in two places but we needed to change the title of the method so that it reflects that it does not only update the status. As it's a nit and the new code is quite readable I'll resolve this comment.

iamacook (approved):

PR #2454, opened 2025-03-13
fix: check the membership of the user when listing members

hectorgomezv:
## Changes
- Adds an additional check to verify the user is actually a member of the organization before getting the list of members.
- Adjusts the existing test cases.

### DIFF
--- a/src/domain/users/user-organizations.repository.integration.spec.ts
+++ b/src/domain/users/user-organizations.repository.integration.spec.ts
@@ -1430,9 +1430,10 @@ describe('UserOrganizationsRepository', () => {
       ).rejects.toThrow('User not found.');
     });
 
-    it('should throw an error if the org does not exist', async () => {
+    it('should throw an error if the user is not a member of the organization', async () => {
       const authPayloadDto = authPayloadDtoBuilder().build();
       const userStatus = faker.helpers.arrayElement(UserStatusKeys);
+      const orgName = faker.word.noun();
       const user = await dbUserRepo.insert({
         status: userStatus,
       });
@@ -1440,17 +1441,18 @@ describe('UserOrganizationsRepository', () => {
         user: user.generatedMaps[0],
         address: authPayloadDto.signer_address,
       });
-      const orgId = faker.number.int({
-        min: 69420,
-        max: DB_MAX_SAFE_INTEGER,
+      const org = await dbOrgRepo.insert({
+        name: orgName,
+        status: 'ACTIVE',
       });
+      const orgId = org.generatedMaps[0].id;
 
       await expect(
         userOrgRepo.findAuthorizedUserOrgsOrFail({
           authPayload: new AuthPayload(authPayloadDto),
           orgId,
         }),
-      ).rejects.toThrow('Organization not found.');
+      ).rejects.toThrow('The user is not a member of the organization.');
     });
   });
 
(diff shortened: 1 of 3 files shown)


src/domain/users/user-organizations.repository.integration.spec.ts
```diff
@@ -1430,27 +1430,29 @@ describe('UserOrganizationsRepository', () => {
       ).rejects.toThrow('User not found.');
     });
 
-    it('should throw an error if the org does not exist', async () => {
+    it('should throw an error if the user is not a member of the organization', async () => {
```
hectorgomezv, line 1433: As the membership check (i.e., whether a `UserOrganization` for the `userId` and `organizationId` pair exists) is done before getting the organization data, checking the organization existence does not make sense anymore. It should exist if a `UserOrganization` for the organization exists.

src/routes/organizations/user-organizations.controller.spec.ts
```diff
@@ -1228,9 +1228,10 @@ describe('UserOrganizationsController', () => {
         });
     });
 
-    it('should throw a 404 if the organization does not exist', async () => {
+    it('should throw a 401 if the user is not a member of the organization', async () => {
```
hectorgomezv, line 1231: Same as https://github.com/safe-global/safe-client-gateway/pull/2454/files#r1994230892

src/domain/users/user-organizations.repository.ts
```diff
@@ -254,9 +254,20 @@ export class UsersOrganizationsRepository
...
+      await this.postgresDatabaseService.getRepository(DbUserOrganization);
+    const userOrganization = await userOrganizationRepository.findOne({
+      where: { user: { id: user.id }, organization: { id: args.orgId } },
+    });
+
+    if (!userOrganization) {
+      throw new UnauthorizedException(
+        'The user is not a member of the organization.',
+      );
+    }
```
iamacook, line 270: We can use `IUsersOrganizationsRepository['findOneOrFail']` here.

hectorgomezv: I agree, but that also implies the error would be 404 (because  `IUsersOrganizationsRepository['findOneOrFail']` throws a 404) if the user is not a member of the organization. Unless a `try-catch` is present in `IUsersOrganizationsRepository['findAuthorizedUserOrgsOrFail']`. 

I'm doing the latter to maintain the 401, as I think it's more accurate/meaningful given the use case. 

[70c9f36](https://github.com/safe-global/safe-client-gateway/pull/2454/commits/70c9f36f85791b4311e2647afb32c1b64b8bbc5b)

I'm a bit unsure of the value these `findOneOrFail` are providing (they just relay their arguments to the TypeORM repository), and they force the caller to remap errors sometimes. Having so many layers of redirections to the actual query sent to the db seems to me a bit prone to error also, and a bit _too-DRY_ 😄 But all this is opinionated of course!

iamacook: Should we deem the error message a requirement then I would avoid rethrowing (https://github.com/safe-global/safe-client-gateway/pull/2454/commits/70c9f36f85791b4311e2647afb32c1b64b8bbc5b).

I agree with your perspective re. the helpers - they are wrapping existing methods. I suggest we address this separately though.

iamacook (approved):

PR #2456, opened 2025-03-14
fix: distinguish the name of an Organization from its creator

hectorgomezv:
## Summary
Previously to this PR, the organization name was reused as the name of the user who created the organization (its first admin). This PR changes the default name to `${organizationName} creator` instead, in order to distinguish the organization name fro the user name.

## Changes
- Sets the default name to `${organizationName} creator` for the user who creates the organization.

### DIFF
--- a/src/domain/organizations/organizations.repository.ts
+++ b/src/domain/organizations/organizations.repository.ts
@@ -41,7 +41,7 @@ export class OrganizationsRepository implements IOrganizationsRepository {
 
     // @todo Move to UserOrganizationsRepository
     const userOrganization = new UserOrganization();
-    userOrganization.name = args.name;
+    userOrganization.name = `${organization.name} creator`;
     userOrganization.role = getEnumKey(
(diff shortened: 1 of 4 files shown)


src/domain/organizations/organizations.repository.ts
```diff
@@ -41,7 +41,7 @@ export class OrganizationsRepository implements IOrganizationsRepository {
 
     // @todo Move to UserOrganizationsRepository
     const userOrganization = new UserOrganization();
-    userOrganization.name = args.name;
+    userOrganization.name = `${organization.name} creator`;
```
iamacook, line 44: I'm not against this but do you think we should be handling internationalisation somehow?

hectorgomezv: To my understanding, internationalization is not on the table right now, but indeed we should adapt this if the supported clients does internationalize their UIs 👍

iamacook (approved):

PR #2468, opened 2025-03-18
fix: swagger message type

compojoom:
## Summary
The generated api had typed the message as object, which is too broad. We know that the message can either be a string or Eip712TypedData message.

With the applied changes when generating the RTK we would get
```
export type TypedDataDomain = {
  name?: string | null
  version?: string | null
  chainId?: object | null
  verifyingContract?: string | null
  salt?: object | null
}
export type TypedDataTypes = {
  name: string
  type: string
}
export type Eip712TypedData = {
  domain: TypedDataDomain
  types: {
    [key: string]: TypedDataTypes[]
  }
  message: {
    [key: string]: any
  }
}

export type Message = {
  messageHash: string
  status: string
  logoUri?: string | null
  name?: string | null
  message: string | Eip712TypedData
  creationTimestamp: number
  modifiedTimestamp: number
  confirmationsSubmitted: number
  confirmationsRequired: number
  proposedBy: AddressInfo
  confirmations: string[]
  preparedSignature?: string | null
  origin?: string | null
}
```
which is way closer to the truth, than the old `message: object`

### DIFF
--- a/src/routes/messages/entities/message.entity.ts
+++ b/src/routes/messages/entities/message.entity.ts
@@ -1,4 +1,9 @@
-import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import {
+  ApiExtraModels,
+  ApiProperty,
+  ApiPropertyOptional,
+  getSchemaPath,
+} from '@nestjs/swagger';
 import { AddressInfo } from '@/routes/common/entities/address-info.entity';
 import { MessageConfirmation } from '@/routes/messages/entities/message-confirmation.entity';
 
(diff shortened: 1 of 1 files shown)


src/routes/messages/entities/message.entity.ts
```diff
@@ -7,6 +12,53 @@ export enum MessageStatus {
...
+
+  constructor(name: string, type: string) {
+    this.name = name;
+    this.type = type;
+  }
+}
+
+type TypedMessageTypes = Record<string, Array<TypedDataTypes>>;
+
+class EIP712TypedData {
```
iamacook, line 35: Nit: we should implement the types from `abitype`, e.g. `TypedData`. The argument types of `hashTypedData` will help guide.

src/routes/messages/entities/message.entity.ts
```diff
@@ -16,7 +68,9 @@ export class Message {
   logoUri: string | null;
   @ApiPropertyOptional({ type: String, nullable: true })
   name: string | null;
-  @ApiProperty()
+  @ApiProperty({
+    oneOf: [{ type: 'string' }, { $ref: getSchemaPath(EIP712TypedData) }],
+  })
```
iamacook, line 73: I can't comment on the specific line (74) but we should be using the above types instead of `Record<string, unknown>`.

src/routes/messages/entities/message.entity.ts
```diff
@@ -7,6 +12,53 @@ export enum MessageStatus {
...
+  constructor(
+    domain: TypedDataDomain,
+    types: Record<string, Array<TypedDataTypes>>,
+    message: Record<string, unknown>,
+  ) {
+    this.domain = domain;
+    this.types = types;
+    this.message = message;
+  }
+}
```
iamacook, line 59: Nit: I'd move these to their own entity file (can be in this folder) as they are not necessarily `Message`-related.

iamacook (commented):
It needn't be included in this PR but we should also include validation of the message to adhere to the type strictness. Currently the validation matches the existing types.

iamacook:
Superseeded by https://github.com/safe-global/safe-client-gateway/pull/2482.

PR #2484, opened 2025-03-20
fix: require accepted membership when requesting a Space metadata

hectorgomezv:
## Changes
- Adds a `status: 'ACTIVE'` requirement to the user requesting space(s) data.
- Adds related test cases.


src/domain/organizations/organizations.repository.ts
```diff
@@ -141,7 +141,7 @@ export class OrganizationsRepository implements IOrganizationsRepository {
       await this.postgresDatabaseService.getRepository(UserOrganization);
 
     const userOrganizations = await userOrganizationRepository.find({
-      where: { user: { id: args.userId } },
+      where: { user: { id: args.userId }, status: 'ACTIVE' },
```
iamacook, line 144: As this is somewhat of a generic function, do you think it'd be better to pass the `status` as an argument instead? This prevents finding `INVITED` user organisations.

src/routes/spaces/spaces.service.ts
```diff
@@ -136,7 +136,7 @@ export class SpacesService {
     const space = await this.spacesRepository.findOneOrFail({
       where: {
         id,
-        userOrganizations: { user: { id: userId } },
+        userOrganizations: { user: { id: userId, status: 'ACTIVE' } },
```
iamacook, line 139: Nit: what do you think about renaming this function to be explicit, e.g. `getActiveSpace`?

src/domain/organizations/organizations.repository.ts
hectorgomezv: It's a really good point, thanks! Besides modifying the filter to include `INVITED`,  I've done some changes to the service calls the repository.
- The service now uses the generic `find` function, as `findByUserId` actually had a lot of internal logic so I thought it wasn't a pure _get-by-id_ function.
- I deleted `findByUserId`, `findOneByUserId`, and `findByUserIdOrFail`. They weren't used anywhere. I don't see the point of having unused wrappers over generic repositories. I've deleted the tests associated with these functions also.
- I've removed the `select` filters, for the sake of simplicity. If there's a specific reason for omitting `createdAt` and `updatedAt` fields I'd revert this change, but I think there isn't, and filtering adds unnecessary complexity.

Feel free to let me know if you think those changes aren't making sense.

src/routes/spaces/spaces.service.ts
hectorgomezv: Good point as well, I've modified that function name, and also `get` to `getActiveOrInvitedSpaces` to be more explicit in both cases.

src/routes/organizations/organizations.controller.spec.ts
```diff
@@ -310,6 +310,8 @@ describe('OrganizationController', () => {
               id: expect.any(Number),
               name: firstOrganizationName,
               status: getEnumKey(OrganizationStatus, OrganizationStatus.ACTIVE),
+              createdAt: expect.any(String),
+              updatedAt: expect.any(String),
```
iamacook, line 314: Are all of these new values required by the clients? I don't see mention of them being needed.

src/routes/organizations/organizations.service.ts
```diff
@@ -114,7 +99,7 @@ export class OrganizationsService {
     return await this.organizationsRepository.findOneOrFail({
       where: {
         id,
-        userOrganizations: { user: { id: userId } },
+        userOrganizations: { user: { id: userId, status: 'ACTIVE' } },
```
iamacook, line 102: Nit: as this now only returns `ACTIVE` `Organization`s, let's rename it to be more explicit.

src/routes/organizations/organizations.controller.spec.ts
hectorgomezv: This is a consequence of what I commented above _I've removed the select filters, for the sake of simplicity. If there's a specific reason for omitting createdAt and updatedAt fields I'd revert this change, but I think there isn't, and filtering adds unnecessary complexity._. 

I've removed the lengthy `select` clause so `createdAt` and `updatedAt` are included in the response, as the are for other objects (in the same response, there are `createdAt` and `updatedAt` timestamps for `userOrganizations` and `user` objects). If there's a strong requirement for including these timestamps I'd include/mask them on mapper, not in the `select` clause, wdyt?

In addition to that, I think the client do expect to have those fields: https://github.com/safe-global/safe-client-gateway/blob/5e268731369fc451a63c0cb06d54f2139bd12ed0/src/routes/organizations/entities/get-organization.dto.entity.ts#L41

src/routes/organizations/organizations.service.ts
hectorgomezv: Agreed, changed in [7c984a6](https://github.com/safe-global/safe-client-gateway/pull/2484/commits/7c984a6922ac4c537919938537314c5599e4a7a7)

iamacook (approved):

PR #2486, opened 2025-03-21
feat: meaningful duplicated invitation error messages

hectorgomezv:
## Changes
- Adds a clearer `0x... is already in this space or has a pending invite` error message when an admin duplicates invitations to the same Space.

### DIFF
--- a/src/domain/users/user-organizations.repository.ts
+++ b/src/domain/users/user-organizations.repository.ts
@@ -17,6 +17,7 @@ import type {
   FindOptionsWhere,
   FindOptionsRelations,
   FindManyOptions,
+  EntityManager,
 } from 'typeorm';
 import type { IUsersOrganizationsRepository } from '@/domain/users/user-organizations.repository.interface';
 import type { Organization } from '@/domain/organizations/entities/organization.entity';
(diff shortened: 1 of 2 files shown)


src/domain/users/user-organizations.repository.ts
```diff
@@ -174,6 +170,19 @@ export class UsersOrganizationsRepository
     return invitations;
   }
 
+  private async createUserAndWallet(args: {
```
hectorgomezv, line 173: Not strictly related, but I've extracted some code from `UsersOrganizationsRepository['inviteUsers']` as it's getting large and harder to read.

src/domain/users/user-organizations.repository.ts
```diff
@@ -117,51 +120,44 @@ export class UsersOrganizationsRepository
...
+            status: 'INVITED',
+            invitedBy: adminAddress,
+          });
+        } catch (err) {
+          if (isUniqueConstraintError(err)) {
+            throw new UniqueConstraintError(
+              `${userToInvite.address} is already in this space or has a pending invite.`,
+            );
+          }
+          throw err;
```
iamacook, line 156: We should prioritise a filter for these error types, but in a follow up.

iamacook (approved):

src/domain/users/user-organizations.repository.ts
hectorgomezv: Agreed, will note this!

PR #2493, opened 2025-03-24
fix: improve swagger message type

compojoom:
## Summary
MessageItem has only a type equal to "MESSAGE" and string is too broad.

Message.confirmation was ending up as sting[] as in reality it is a MessageConfirmation[] type.

### DIFF
--- a/src/routes/messages/entities/message.entity.ts
+++ b/src/routes/messages/entities/message.entity.ts
@@ -37,7 +37,7 @@ export class Message {
   confirmationsRequired: number;
   @ApiProperty()
   proposedBy: AddressInfo;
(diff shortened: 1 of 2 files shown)


src/routes/messages/entities/message.entity.ts
```diff
@@ -37,7 +37,7 @@ export class Message {
   confirmationsRequired: number;
   @ApiProperty()
   proposedBy: AddressInfo;
-  @ApiProperty()
+  @ApiProperty({ type: [MessageConfirmation] })
```
iamacook, line 40:
```suggestion
  @ApiProperty({ type: MessageConfirmation, isArray: true })
```

iamacook (approved):

PR #2500, opened 2025-03-26
feat: limit the number of Safes per Space

hectorgomezv:
## Summary
This PR limits the number of Safes that can be added to the same Space. If the sum of the already present Safes in the Space and the requested Safes to add surpasses this limit, then no Safes are added to the Space, and an error is thrown.

## Changes
- Adds a `maxSafesPerSpace` limit to `SpaceSafesRepository`.
- Adds related test cases.

### DIFF
--- a/src/config/entities/__tests__/configuration.ts
+++ b/src/config/entities/__tests__/configuration.ts
@@ -270,6 +270,10 @@ export default (): ReturnType<typeof configuration> => ({
   safeWebApp: {
     baseUri: faker.internet.url({ appendSlash: false }),
   },
+  spaces: {
+    maxSafesPerSpace: faker.number.int({ min: 5, max: 10 }),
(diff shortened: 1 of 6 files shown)


src/config/entities/__tests__/configuration.ts
```diff
@@ -270,6 +270,10 @@ export default (): ReturnType<typeof configuration> => ({
   safeWebApp: {
     baseUri: faker.internet.url({ appendSlash: false }),
   },
+  spaces: {
+    maxSafesPerSpace: faker.number.int({ min: 5, max: 10 }),
+    maxInvites: faker.number.int({ min: 5, max: 10 }),
```
hectorgomezv, line 275: This is not strictly related with the PR, but I thought it would be nice to group all the spaces-related configuration.

iamacook (approved):

PR #2503, opened 2025-03-26
feat: warm token/contracts cache

hectorgomezv:
## Summary
This PR implements a pre-fetch step for tokens/contracts when getting the transactions queue/history.
The addresses from the transactions involved in the queue/history are extracted, de-duplicated, and fetched beforehand: this stores their values (or their absence), so subsequent mappers have this data available in the cache.

## Changes
- Adds a `prefetchAddressInfos` function to `QueuedItemsMapper` and `TransactionsHistoryMapper`.
- Adjusts the related tests.

PR #2506, opened 2025-03-27
refactor: cache in-flight network requests

iamacook:
## Summary

Due to the nature of the Client Gateway, we fetch and map data from multiple datasources. As mapping can occur simultaneously, there can be duplicate network requests for the same data.

As we investigate improving [cache warming](https://github.com/safe-global/safe-client-gateway/pull/2503) for [token requests](https://github.com/safe-global/safe-client-gateway/issues/2181), this demonstrates a cache for in-flight requests.

Promises of network requests are stored in an in-memory cache, returned if there is a parallel request. After the request resolves/rejects, it is cleared from the cache.

## Changes

- Add caching mechanism to `fetch` factory

### DIFF
--- a/src/datasources/network/network.module.ts
+++ b/src/datasources/network/network.module.ts
@@ -25,7 +25,7 @@ function fetchClientFactory(
     'httpClient.requestTimeout',
   );
 
-  return async <T>(
+  const request = async <T>(
     url: string,
     options: RequestInit,
   ): Promise<NetworkResponse<T>> => {
@@ -54,6 +54,27 @@ function fetchClientFactory(
       status: response.status,
       data,
     };
+  };
+
+  /**
+   * A cache to prevent multiple in-flight requests for the same data.
+   */
+  const cache: Record<string, Promise<NetworkResponse<unknown>>> = {};
+
+  return async <T>(
+    url: string,
+    options: RequestInit,
+  ): Promise<NetworkResponse<T>> => {
+    // Naive key for simplicity but JSON.stringify is not stable
+    const key = JSON.stringify({ url, ...options });
+
+    if (!(key in cache)) {
+      cache[key] = request(url, options).finally(() => {
+        delete cache[key];
+      });
+    }
+
+    return cache[key];
   };
 }
 
--- a/src/datasources/network/network.module.spec.ts
+++ b/src/datasources/network/network.module.spec.ts
@@ -108,4 +108,89 @@ describe('NetworkModule', () => {
 
     expect(fetchMock).toHaveBeenCalledTimes(1);
   });
+
+  it('caches GET requests based on URL and options', async () => {
+    const json = fakeJson();
+    const response = {
+      ok: true,
+      json: () => Promise.resolve(json),
+    } as Response;
+    fetchMock.mockResolvedValue(response);
+
+    const url = faker.internet.url({ appendSlash: false });
+    const options = { method: 'GET' };
+
+    void fetchClient(url, options);
+    await fetchClient(url, options);
+
+    expect(fetchMock).toHaveBeenCalledTimes(1);
+  });
+
+  it.each(['POST', 'DELETE'])(
+    'caches %s requests based on URL and options',
+    async (method) => {
+      const json = fakeJson();
+      const response = {
+        ok: true,
+        json: () => Promise.resolve(json),
+      } as Response;
+      fetchMock.mockResolvedValue(response);
+
+      const url = faker.internet.url({ appendSlash: false });
+      const options = {
+        method,
+        body: JSON.stringify({ example: 'data' }),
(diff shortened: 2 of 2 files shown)


src/datasources/network/network.module.ts
```diff
@@ -55,6 +55,27 @@ function fetchClientFactory(
...
+    const key = JSON.stringify({ url, ...options });
+
+    if (!(key in cache)) {
+      cache[key] = request(url, options).finally(() => {
+        delete cache[key];
+      });
+    }
+
+    return cache[key];
+  };
```
iamacook, line 78: We could abstract this to a wrapper function but it is included for simplicity.

src/datasources/network/network.module.ts
```diff
@@ -55,6 +55,27 @@ function fetchClientFactory(
...
+   * A cache to prevent multiple in-flight requests for the same data.
+   */
+  const cache: Record<string, Promise<NetworkResponse<unknown>>> = {};
+
+  return async <T>(
+    url: string,
+    options: RequestInit,
+  ): Promise<NetworkResponse<T>> => {
+    // Naive key for simplicity but JSON.stringify is not stable
+    const key = JSON.stringify({ url, ...options });
```
iamacook, line 69: If the keys of an object are different in order, this key will be different, but I think this is a sufficient starting point.

src/datasources/network/network.module.ts
iamacook: Split in 197e26611d8cb172cb8698dd2cd6b3c15271ab17.

src/datasources/network/network.module.ts
iamacook: The key should be stable, according to your suggestion, as of 197e26611d8cb172cb8698dd2cd6b3c15271ab17.

src/datasources/network/network.module.ts
```diff
@@ -21,11 +21,14 @@ export type FetchClient = <T>(
 function fetchClientFactory(
```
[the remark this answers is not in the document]
iamacook: Added in 197e26611d8cb172cb8698dd2cd6b3c15271ab17.

src/datasources/network/network.module.ts
```diff
@@ -57,6 +74,72 @@ function fetchClientFactory(
...
+}
+
+function sortRequestInit(requestInit: RequestInit): RequestInit {
+  if (typeof requestInit.body !== 'string') {
+    return sortObject(requestInit);
+  }
+
+  const sortedBody = sortObject(JSON.parse(requestInit.body));
+  return sortObject({
+    ...requestInit,
```
PooyaRaki, line 138: Why do we need to sort the whole object? My concern is about the performance even though the JavaScript `sort` function has a time complexity of `O(n log n)`

iamacook: As per discussion, a review comment was removed regarding a more stable key (`JSON.stringify` results in a different string depending on the key order).

To initially test the implementation, we have decided to revert to this naive implementation in 52fcdb189e55579f31fbf21c96766bccb9460143 and intend to visit it thereafter.

src/domain/common/utils/utils.ts
```diff
@@ -25,3 +28,7 @@ export function truncateAddress(
 ): `0x${string}` {
   return `${address.slice(0, length + 2)}...${address.slice(-length)}` as `0x${string}`;
 }
+
+export function hashString(value: string): string {
```
PooyaRaki, line 32: Nit: It should be clear which hashing algorithm is used just by looking at the function. To improve clarity, we could either include the algorithm in the function signature or rename the function to reflect the algorithm used.

src/datasources/network/network.module.ts
```diff
@@ -57,6 +74,63 @@ function fetchClientFactory(
...
+function getCacheKey(url: string, requestInit?: RequestInit): string {
+  if (!requestInit) {
+    return hashString(url);
+  }
+
+  // JSON.stringify does not produce a stable key but initially
+  // use a naive implementation for testing the implementation
+  // TODO: Revisit this and use a more stable key
+  const key = JSON.stringify({ url, ...requestInit });
+  return hashString(key);
```
PooyaRaki, line 131: Nit: In this case, the specific hashing algorithm isn’t critical, since our goal is to generate the shortest possible string. Therefore, we can use algorithms like SHA-1 that produce relatively smaller outputs.

PooyaRaki (approved):

src/domain/common/utils/utils.ts
iamacook: Renamed in d7b5297ddbefac6c43bfa41e1d66645500a22a0d.

PooyaRaki (approved):

PR #2511, opened 2025-03-28
feat: map batched transfers

### LINKED ISSUE #2352: Return token info. and/or parsed value for batched ERC-20 transfers
### Description

Note: after this is finished, we should create a relevant task for the clients.

The UI currently maps batched transactions as is, lacking clarity and requiring manual checks to ensure the correct amount is being sent in the case of ERC-20 transfers, etc.

Image

It should ideally include the parsed value and/or the token info. as historical transactions, for example, may mean the asset is currently not held (and the balance endpoint not be referenced for the decimal value).

We should consider returning the token info., or the parsed value alongside the decoded data so that it can be rendered as a parsed value.

As to whether we traverse data decoded or create a new data structure is to also be decided.

### Requirements

- Decide on whether we return the token info. and/or the parsed value alongside transfers.
- Traverse data decoded and include above value.

### Additional information

_No response_

iamacook:
## Summary

Resolves #2352

Batched transactions are mapped "as is" by the client, meaning that rich token data is missing. This extracts and maps native coin, ERC-20 and ERC-721 transfers, returning them in the response.

## Changes

- Extract and map transfers within batches


iamacook:
Superseeded by https://github.com/safe-global/safe-client-gateway/pull/2537.

PR #2512, opened 2025-03-28
feat: add in-memory caching layer for Token prices

hectorgomezv:
## Summary
This PR implements an in-memory cache mechanism for ERC20 token prices, aiming to reduce the number of network calls to the Redis server. This mechanism could be extended to other parts of the service.

## Changes
- Adds the `cache-manager` library dependency ([recommended by NestJS](https://docs.nestjs.com/techniques/caching)).
- Adds this in-memory cache layer to `CoingeckoApiService`.
- Unifies debug-level logging for cache/memory hits/misses on the `LogType` enum.
- Adds related tests.


src/datasources/balances-api/coingecko-api.service.ts
```diff
@@ -369,4 +414,41 @@ export class CoingeckoApi implements IPricesApi {
       this.notFoundPriceTtlSeconds + CoingeckoApi.NOT_FOUND_TTL_RANGE_SECONDS,
     );
   }
+
+  /**
+   * Gets the in-memory cache key for the given cacheDir.
+   */
+  private getMemoryKey(cacheDir: CacheDir): string {
+    return `${cacheDir.key}:${cacheDir.field}`;
+  }
```
iamacook, line 423: Nit: shall we move this to the `CacheRouter`?

src/datasources/balances-api/coingecko-api.service.ts
```diff
@@ -262,14 +286,26 @@ export class CoingeckoApi implements IPricesApi {
...
       } else {
-        this.loggingService.debug({ type: 'cache_miss', key, field });
+        this.logMemoryMiss(cacheDir.key);
+        const cacheItem = await this.cacheService.hGet(cacheDir);
+        if (cacheItem != null) {
+          this.logCacheHit(cacheDir);
+          await this.inMemoryCache.set(
+            this.getMemoryKey(cacheDir),
+            cacheItem,
+            this.pricesTtlSeconds * 1_000,
```
iamacook, line 303: If the value from Redis is has a remaining TTL of `1`, this will be cached for a further `this.pricesTtlSeconds * 1_000`. Is there any way to get the remaining TTL from Redis and set it as that? Otherwise we could doubling the TTL of these values.

yarn.lock
```diff
@@ -7175,6 +7206,15 @@ __metadata:
   languageName: node
   linkType: hard
 
+"keyv@npm:^5.3.1":
```
iamacook, line 7209: General comment: It's interesting that cache-manger uses this. Keyv also [supports Redis](https://keyv.org/docs/storage-adapters/redis/) so maybe we could simplify this logic (in the future).

iamacook (commented):
This needn't be addressed here but we now have a two-level cache implementation, and it's manually implementated in a datasource.

Whilst I understand the reasoning for the isolated approach, we should consider integrating it into our cache service to not lose context.

src/datasources/balances-api/coingecko-api.service.ts
hectorgomezv: This value is in milliseconds, that's the reason we multiply it by 1000. The cache times should be the same for Redis and memory in this case. Will include a comment to document this!

hectorgomezv: Added in [0cf8239](https://github.com/safe-global/safe-client-gateway/pull/2512/commits/0cf8239b6f15217eb2401d7ed840318c4b741771)

src/datasources/balances-api/coingecko-api.service.ts
hectorgomezv: Moved in [0cf8239](https://github.com/safe-global/safe-client-gateway/pull/2512/commits/0cf8239b6f15217eb2401d7ed840318c4b741771)

src/datasources/balances-api/coingecko-api.service.spec.ts
```diff
@@ -390,6 +438,122 @@ describe('CoingeckoAPI', () => {
...
+          'x-cg-pro-api-key': coingeckoApiKey,
+        },
+        params: {
+          contract_addresses: firstTokenAddress,
+          vs_currencies: lowerCaseFiatCode,
+        },
+      },
+    });
+    // mockInMemoryCache.get should have been called 3 times
+    expect(mockInMemoryCache.get).toHaveBeenCalledTimes(3);
```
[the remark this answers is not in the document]
hectorgomezv: Added in [0cf8239](https://github.com/safe-global/safe-client-gateway/pull/2512/commits/0cf8239b6f15217eb2401d7ed840318c4b741771)

src/datasources/balances-api/coingecko-api.service.spec.ts
```diff
@@ -390,6 +438,122 @@ describe('CoingeckoAPI', () => {
...
+        return Promise.resolve(
+          JSON.stringify({
+            [secondTokenAddress]: { [lowerCaseFiatCode]: secondPrice },
+          }),
+        );
+      }
+      return Promise.resolve(undefined);
+    });
+    mockInMemoryCache.get.mockImplementation((key) => {
+      if (key.includes(thirdTokenAddress)) {
```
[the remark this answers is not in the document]
hectorgomezv: Changed in [0cf8239](https://github.com/safe-global/safe-client-gateway/pull/2512/commits/0cf8239b6f15217eb2401d7ed840318c4b741771)

hectorgomezv:
> This needn't be addressed here but we now have a two-level cache implementation, and it's manually implementated in a datasource.
Whilst I understand the reasoning for the isolated approach, we should consider integrating it into our cache service to not lose context.

Totally agree. In fact, I was planning to extend this 2-layer approach to other areas of the service, primarily to entities that do not change often over time (and therefore are not invalidated by event hooks), that are accessed on high rates, and have small payloads. I thought we could start with token prices as they tick all these boxes.

iamacook (approved):

PR #2513, opened 2025-03-28
feat: unify names validation

hectorgomezv:
## Summary
This PR unifies `name` validation for some domain entities:
- Spaces
- Members
- Accounts (deprecated version)
- Address Books (deprecated version)

## Changes
- Adds validation to `Space` and `Member` names, via a new `NameSchema`.
- Adds tests for the new schema.
- Removes custom name validators for the deprecated `Account` and `AddressBookItem` entities.


src/domain/spaces/entities/space.entity.ts
```diff
@@ -21,7 +22,7 @@ export const SpaceSchema: z.ZodType<
     safes?: DbSpace['safes'];
   }
 > = RowSchema.extend({
-  name: z.string().max(255),
+  name: NameSchema,
```
hectorgomezv, line 25: This is one of the meaningful changes in the PR: `Space` names where plain strings.

src/domain/users/entities/member.entity.ts
```diff
@@ -31,7 +32,7 @@ export const MemberSchema: z.ZodType<
 > = RowSchema.extend({
   user: z.lazy(() => UserSchema),
   space: z.lazy(() => SpaceSchema),
-  name: z.string(),
+  name: NameSchema,
```
hectorgomezv, line 35: This is the other meaningful change in the PR: `Member` names were plain strings.

src/domain/spaces/entities/space.entity.ts
iamacook: The database entity has a max limit of 255, whereas `NameSchema` has 20. Should we also change the schema, or increase the schema? We need to take into consideration that there may already be names in the database that are >20 characters in length.

hectorgomezv: I've added the restriction to the database schema in [92dd3f0](https://github.com/safe-global/safe-client-gateway/pull/2513/commits/92dd3f0c910ff4220f233a85bce8817ca49c888f)

In addition to that, I've centralized the min/max name values into `name.schema.ts`, and extended the allowed chars to 30. I think it's a better balance between security and UX. Given that non-ascii and non-alphanumeric characters are disallowed, the 30-characters limitation should be OK, wdyt?.

iamacook: I think the restrictions make sense. Thank you for making the changes.

iamacook (approved):

PR #2518, opened 2025-04-01
feat: return 24h balance change

iamacook:
## Summary

Our prices provider returns a 24h balance change but we are not exposing it. This adds the relevant logic to include it in balance requests under a new `fiatBalance24hChange` field.

## Changes

- Convert `AssetPriceSchema` to be dynamic, based on a provided currency code
- Map and return `{currency}_24h_change` from CoinGecko, with a default of `null` with Zerion/Transaction Service
- Map change onto balances
- Update test coverage accordingly

### DIFF
--- a/src/datasources/balances-api/coingecko-api.service.ts
+++ b/src/datasources/balances-api/coingecko-api.service.ts
@@ -3,7 +3,7 @@ import { IConfigurationService } from '@/config/configuration.service.interface'
 import { IPricesApi } from '@/datasources/balances-api/prices-api.interface';
 import {
   AssetPrice,
-  AssetPriceSchema,
+  getAssetPriceSchema,
 } from '@/datasources/balances-api/entities/asset-price.entity';
 import { CacheFirstDataSource } from '../cache/cache.first.data.source';
 import { CacheRouter } from '../cache/cache.router';
@@ -30,6 +30,14 @@ import { Cache } from 'cache-manager';
 import { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 import { LogType } from '@/domain/common/entities/log-type.entity';
 
+/**
+ * TODO: Refactor away the return of currency codes from public methods, e.g.
+ *
+ * {                                  {
+ *   usd: number;                       price: number;
+ *   usd_24h_change: number;    =>      price_24h_change: number;
+ * }                                  }
+ */
 @Injectable()
 export class CoingeckoApi implements IPricesApi {
   /**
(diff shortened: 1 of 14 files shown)


src/datasources/balances-api/coingecko-api.service.ts
```diff
@@ -26,6 +26,14 @@ import { Chain } from '@/domain/chains/entities/chain.entity';
...
 import { rawify, type Raw } from '@/validation/entities/raw.entity';
 
+/**
+ * TODO: Refactor away the return of currency codes from public methods, e.g.
+ *
+ * {                                  {
+ *   usd: number;                       price: number;
+ *   usd_24h_change: number;    =>      price_24h_change: number;
+ * }                                  }
+ */
```
iamacook, line 36: Due to the complexity of this change, I suggest we follow up with this change.

src/domain/balances/entities/balance.entity.ts
```diff
@@ -23,6 +23,11 @@ export const Erc20BalanceSchema = z.object({
 
 const FiatSchema = z.object({
   fiatBalance: z.string().nullish().default(null),
+  fiatBalance24hChange: z.coerce
+    .string()
+    .nullable()
+    // Transaction Service does not return this field
```
hectorgomezv, line 29: Nit: the Transaction Service doesn't return any of the fields in FiatSchema. We can move this comment to FiatSchema itself, but I'm not sure it's relevant here (ideally, this class is agnostic regarding the origin of the balances data).

src/domain/balances/entities/__tests__/balance.builder.ts
```diff
@@ -6,10 +6,15 @@ import type { Balance } from '@/domain/balances/entities/balance.entity';
...
-    .with('fiatConversion', null);
+  return (
+    new Builder<Balance>()
+      .with('tokenAddress', getAddress(faker.finance.ethereumAddress()))
+      .with('token', balanceTokenBuilder().build())
+      .with('balance', faker.string.numeric())
+      .with('fiatBalance', null)
+      .with('fiatConversion', null)
+      // TODO: Separate entity into base and extended to avoid doing this
+      // Transaction Service does not return this field and it will default to null
```
hectorgomezv, line 17: Same as https://github.com/safe-global/safe-client-gateway/pull/2518/files#r2022886229

src/domain/balances/entities/balance.entity.ts
iamacook: True - this was a remnant from development. Removed in a2226a9401f4bb00fc7cf36cae976c022790e3d6.

hectorgomezv (approved):

src/domain/balances/entities/__tests__/balance.builder.ts
iamacook: Removed in d716924a75e14ee9d8cbb28e3a12f0923c181789.

hectorgomezv (approved):

PR #2525, opened 2025-04-03
fix: define `fiatBalance24hChange` correctly

iamacook:
## Summary

`Balance['fiatBalance24hChange']` is a string but it is does not have a `type` on Swagger. This adds a `String` `type`.

## Changes

- Add `String` `type` to decorator


iamacook:
After discussing this with the team, I am going to mark this as a draft for the time being. All numerical values are returned as strings so we might want to reconsider this. We will go ahead with UI implementation and decide thereafter.

We will either way need to fix the entity as the decorator is currently missing a `type`.

iamacook:
After discussion, we return floats elsewhere as strings. As such, this keeps the type but corrects the Swagger definition.

hectorgomezv (approved):

PR #2534, opened 2025-04-11
fix: error handling in notification services

PooyaRaki:
## Summary
Enhance error handling in notification services to provide clearer error messages.
Currently, the Firebase error is not logged properly because there is no `message` property in the error thrown so for that reason `HttpErrorFactory` logs a generic error message. This PR logs the error message from the Firebase directly.

## Changes
- Logs the error message of Firebase endpoint call

### DIFF
--- a/src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
+++ b/src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
@@ -21,6 +21,7 @@ import {
   FirebaseOauth2Token,
   FirebaseOauth2TokenSchema,
 } from '@/datasources/push-notifications-api/entities/firebase-oauth2-token.entity';
+import { asError } from '@/logging/utils';
 
 @Injectable()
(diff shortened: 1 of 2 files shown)


src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
```diff
@@ -127,13 +128,9 @@ export class FirebaseCloudMessagingApiService implements IPushNotificationsApi {
...
       });
     } catch (error) {
-      /**
-       * TODO: Error handling based on `error.details[i].reason`, e.g.
-       * - expired OAuth2 token
-       * - stale FCM token
-       * - don't expose the error to clients, logging on domain level
-       */
-      throw this.httpErrorFactory.from(error);
+      const errorMessage = asError(error).message;
```
PooyaRaki, line 131: I added this line solely for consistency with other services; otherwise, we would be able to catch the error in the domain layer.

hectorgomezv (approved):

iamacook (approved):

PR #2535, opened 2025-04-11
fix: preserve numerical string decimals

iamacook:
## Summary

When converting to numerical strings, we use the local formatting (`en-US`). This reduces the number of decimals places that can be returned, notably for fiat balances returned by prices providers.

This pivots from using `toLocaleString` to `Intl.NumberFormat` with the maximum fraction digits possible. Therefore, balances are now returned with greater accuracy.

## Changes

- Use `Intl.NumberFormat`.

### DIFF
--- a/src/domain/common/utils/utils.ts
+++ b/src/domain/common/utils/utils.ts
@@ -2,13 +2,14 @@
 // @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#maximumfractiondigits
 const MAX_MAXIMUM_FRACTION_DIGITS = 100;
 
-export function getNumberString(value: number): string {
+const formatter = new Intl.NumberFormat('en-US', {
   // Prevent scientific notation
-  const formatter = new Intl.NumberFormat('en-US', {
-    notation: 'standard',
-    useGrouping: false,
-    maximumFractionDigits: MAX_MAXIMUM_FRACTION_DIGITS,
-  });
+  notation: 'standard',
+  useGrouping: false,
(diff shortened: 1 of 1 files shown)


hectorgomezv (approved):

src/domain/common/utils/utils.ts
```diff
@@ -1,9 +1,16 @@
-export function getNumberString(value: number): string {
+// We use the maximum value in order to preserve all decimals
+// @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#maximumfractiondigits
+const MAX_MAXIMUM_FRACTION_DIGITS = 100;
+
+const formatter = new Intl.NumberFormat('en-US', {
   // Prevent scientific notation
```
PooyaRaki, line 6: Nit: We could remove the comment and the notation as the default value is `standard`.

PooyaRaki (approved):

PooyaRaki:
QQ: Have we checked how the UI displays the price when there are multiple leading zeros?

src/domain/common/utils/utils.ts
iamacook: As this was there prior, I'm more in favour of remaining explicit and keeping it.

iamacook:
> QQ: Have we checked how the UI displays the price when there are multiple leading zeros?

It [formats](https://github.com/safe-global/safe-wallet-monorepo/blob/631db0cead5316a50173b351323c57b8a68ac637/packages/utils/src/utils/formatters.ts#L38) according to [our spec.](https://github.com/5afe/safe/wiki/How-to-format-amounts).

PR #2536, opened 2025-04-14
refactor: improve error handling in notification services

PooyaRaki:
## Summary
This PR refactors the handling of Firebase error messages to enable proper logging. Firebase errors are now mapped explicitly, as the actual error details are found in data.error_description rather than the standard message property.

## Changes
- Maps `Firebase` error message

### DIFF
--- a/src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
+++ b/src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
@@ -4,7 +4,6 @@ import {
   CacheService,
   ICacheService,
 } from '@/datasources/cache/cache.service.interface';
-import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
 import {
   NetworkService,
   INetworkService,
@@ -21,7 +20,8 @@ import {
   FirebaseOauth2Token,
   FirebaseOauth2TokenSchema,
 } from '@/datasources/push-notifications-api/entities/firebase-oauth2-token.entity';
-import { asError } from '@/logging/utils';
+import { get } from 'lodash';
+import { NetworkResponseError } from '@/datasources/network/entities/network.error.entity';
 
 @Injectable()
 export class FirebaseCloudMessagingApiService implements IPushNotificationsApi {
@@ -34,6 +34,8 @@ export class FirebaseCloudMessagingApiService implements IPushNotificationsApi {
   private static readonly DefaultIosNotificationTitle = 'New Activity';
   private static readonly DefaultIosNotificationBody =
     'New Activity with your Safe';
+
+  private static readonly ERROR_ARRAY_PATH = 'data.error_description';
 
   private readonly baseUrl: string;
   private readonly project: string;
@@ -49,7 +51,6 @@ export class FirebaseCloudMessagingApiService implements IPushNotificationsApi {
     private readonly cacheService: ICacheService,
     @Inject(IJwtService)
     private readonly jwtService: IJwtService,
-    private readonly httpErrorFactory: HttpErrorFactory,
   ) {
     this.baseUrl = this.configurationService.getOrThrow<string>(
       'pushNotifications.baseUri',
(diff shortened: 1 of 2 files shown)


src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
```diff
@@ -128,13 +129,7 @@ export class FirebaseCloudMessagingApiService implements IPushNotificationsApi {
...
       });
     } catch (error) {
-      /**
-       * @todo Handle error properly
-       * We should consider NotificationRespository when handling errors because the logic is parially handled there.
-       */
-      const errorMessage = asError(error).message;
-
-      throw new Error(errorMessage);
+      throw this.mapError(error);
```
iamacook, line 132: Nit: can we keep mention of `NotificationRespository`? Not using the `HttpErrorFactory` goes against the datasource pattern so it's good reference for when we come to improve error handling.

src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
```diff
@@ -196,4 +191,19 @@ export class FirebaseCloudMessagingApiService implements IPushNotificationsApi {
       algorithm: 'RS256',
     });
   }
+
+  private mapError(error: unknown): unknown {
```
iamacook, line 195: Nit: we should add tests for this.

src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
PooyaRaki: Well, IMHO it's unnecessary as we are mapping the errors like other datasources but as you requested I'll add the todo comment again.

src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
PooyaRaki: As it's a private method we should test it through `enqueueNotification`. Unfortunately, the method does not have a high test coverage i.e. not all scenarios like error cases are tested, please have a look at the test file here here https://github.com/safe-global/safe-client-gateway/blob/7d43ca92a263735d430bf64c0fc3fdaec1dd3de8/src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.spec.ts
On the other hand, it is implicitly tested by [src/routes/hooks/hooks-notifications.spec.ts](https://github.com/safe-global/safe-client-gateway/blob/b289d4480da7d323ac3e6d4c95d65e3813ef86e7/src/routes/hooks/hooks-notifications.spec.ts)
IMO, adding coverage to the current file is out of the scope of the current PR but I'm happy to add some todo tests tests for it, what do you think?

src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
PooyaRaki: Done ✅ https://github.com/safe-global/safe-client-gateway/pull/2536/commits/66b04e4756aa1da57f06cb75f16a6791a63cfec5

src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
iamacook: I'm always in favour of adding test coverage, but I don't think it's overtly necessary if you prefer not to. We could add an extra case to the [service suite](https://github.com/safe-global/safe-client-gateway/blob/main/src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.spec.ts) - similar to how we test Django error mapping [in the `TransactionApi`](https://github.com/safe-global/safe-client-gateway/blob/main/src/datasources/transaction-api/transaction-api.service.spec.ts#L137).

PooyaRaki: Done https://github.com/safe-global/safe-client-gateway/pull/2536/commits/63caf742157ea2670478979c75d28c1846403eed ✅

iamacook (approved):

PR #2551, opened 2025-04-28
feat: map vault `redeem` transactions

iamacook:
## Summary

Resolves https://github.com/safe-global/wallet-private-tasks/issues/2

This adds mapping of vault-specific `redeem` transactions, refactoring mapping of `deposit` accordingly.

The staking datasource has been extended to retrieve DeFi stakes, and Morpho extra rewards, required for determining the balance of staking (additional) rewards. These are included/validated in their respective repository.

## Changes

- Add new cache keys
- Extend `KilnApi` with new endpoints for DeFi stakes and Morpho extra rewards:
  - Add/infer tyes from validation schema for the above
  - Add respective methods to the `StakingRepository`, validating against said schemas
- Refactor `mapDepositInfo` and related entities
- Add `mapWithdrawInfo`, called in `MultisigTransactionInfoMapper`


src/routes/transactions/mappers/common/vault-transaction.mapper.ts
```diff
@@ -33,43 +40,156 @@ export class VaultTransactionMapper {
...
+        defiVaultStats,
+      });
+
+    return new VaultDepositTransactionInfo({
+      value: getNumberString(value),
+      fee,
+      baseNrr: defiVaultStats.nrr,
+      tokenInfo: new TokenInfo({ ...token, trusted: true }),
+      vaultInfo: this.mapVaultInfo({ deployment, defiVaultStats }),
+      currentReward: '0',
```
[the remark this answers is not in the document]
iamacook: If you previously desposited the asset, there may already be accrued awards. I thought it would be confusing to return previous rewards when depositing though.

I can't see this value in the design yet - do you know if it'll be shown in the transaction details? If so, it would make more sense to return the current rewards.

src/domain/staking/staking.repository.ts
```diff
@@ -93,6 +102,40 @@ export class StakingRepository implements IStakingRepository {
...
   }
 
+  public async getDefiVaultStake(args: {
+    chainId: string;
+    safeAddress: `0x${string}`;
+    vault: `0x${string}`;
+  }): Promise<DefiVaultStake> {
+    const stakingApi = await this.stakingApiFactory.getApi(args.chainId);
+    const defiStakes = await stakingApi.getDefiVaultStakes(args);
+    // Cannot be >1 contract deployed at the same address so return first element
```
usame-algan, line 112: Not sure I understand this comment. We are fetching the deposits on a specific vault (1 address) for a specific safe so the response array always contains exactly one element right?

src/domain/staking/staking.repository.ts
```diff
@@ -93,6 +102,40 @@ export class StakingRepository implements IStakingRepository {
...
+    return (
+      DefiVaultStakesSchema.parse(defiStakes)
+        // Should not return non-matching chain_id but just in case
+        .filter((stake) => {
+          return stake.chain_id === Number(args.chainId);
+        })[0]
```
usame-algan, line 118: nit: We could also check that `stake.owner === args.safeAddress`

src/domain/staking/staking.repository.ts
```diff
@@ -93,6 +102,40 @@ export class StakingRepository implements IStakingRepository {
...
+  }): Promise<Array<DefiMorphoExtraReward>> {
+    const stakingApi = await this.stakingApiFactory.getApi(args.chainId);
+    const defiMorphoExtraRewards = await stakingApi.getDefiMorphoExtraRewards(
+      args.safeAddress,
+    );
+    return (
+      DefiMorphoExtraRewardsSchema.parse(defiMorphoExtraRewards)
+        // Should not return non-matching chain_id but just in case
+        .filter((reward) => {
+          return reward.chain_id === Number(args.chainId);
```
usame-algan, line 134: nit: Similar to above we could also check `reward.wallet === args.safeAddress`

usame-algan (approved):
Thanks!

src/domain/staking/staking.repository.ts
iamacook: I checked the API and it only returns data for the chain/address so this is not necessary. I removed it entirely in 879497c177b4a19f1143158159b018d4cabcca38.

src/domain/staking/staking.repository.ts
iamacook: I checked the API and it only returns data for the chain/address so this is not necessary. I removed it entirely in 879497c177b4a19f1143158159b018d4cabcca38.

usame-algan (approved):
LFG 🔥

PR #2554, opened 2025-04-28
fix: Log Firebase auth and send error message

PooyaRaki:
## Summary
In a previous [PR](https://github.com/safe-global/safe-client-gateway/pull/2536), we added the ability to directly log error messages from the Firebase `Auth` endpoint. However, the error structure for `sendMessage` differs from that of the `Auth` endpoint. In this PR, we update the logging to handle and log both error structures.

## Changes
- Logs the Firebase `sendMessage` endpoint errors.
- Added a helper to get the first available key in an array.

### DIFF
--- a/src/domain/common/utils/array.ts
+++ b/src/domain/common/utils/array.ts
@@ -1,0 +1,16 @@
+import type { GetFieldType } from 'lodash';
+import get from 'lodash/get';
+
+export function getFirstAvailable<TObject, TPath extends string>(
+  sourceObject: TObject,
+  paths: Array<string>,
+): GetFieldType<TObject, TPath> | undefined {
+  for (const path of paths) {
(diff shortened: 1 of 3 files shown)


src/domain/common/utils/array.ts
```diff
@@ -0,0 +1,16 @@
...
+  paths: Array<string>,
+): GetFieldType<TObject, TPath> | undefined {
+  for (const path of paths) {
+    const value = get(sourceObject, path);
+    if (value !== undefined && value !== null) {
+      return value;
+    }
+  }
+
+  return undefined;
```
iamacook, line 15: Nit: we don't need this as it otherwise inherently returns `undefined`.

iamacook (approved):

PR #2558, opened 2025-04-29
fix: use dynamic TTL for in-memory price cache

iamacook:
## Summary

We use a dynamic TTL when caching token prices - certain tokens have high refresh rates.

When populating the cache, this is [set correctly](https://github.com/safe-global/safe-client-gateway/blob/fix/in-memory-token-cache/src/datasources/balances-api/coingecko-api.service.ts#L376). When populating the in-memory cache, however, it is currently set to the defaul price TTL. This adjusts the logic in question to set the dynamic TTL

## Changes

- Populate in-memory cache according to token

### DIFF
--- a/src/datasources/balances-api/coingecko-api.service.spec.ts
+++ b/src/datasources/balances-api/coingecko-api.service.spec.ts
@@ -66,6 +66,7 @@ describe('CoingeckoAPI', () => {
   const coingeckoApiKey = faker.string.sample();
   const pricesTtlSeconds = faker.number.int();
   const nativeCoinPricesTtlSeconds = faker.number.int();
+  const highRefreshRateToken = faker.finance.ethereumAddress().toLowerCase();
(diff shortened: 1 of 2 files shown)


schmanu (approved):
Good catch! 
Should we add a test case to the Coingecko API test?

PooyaRaki (approved):

PooyaRaki:
> Good catch! Should we add a test case to the Coingecko API test?

Very good point about testing @schmanu 👍 
Based on the code, I feel like it needs some refactoring to be able to effectively test the TTL as it's hard coded(not injected, even though we read it from the config service) in the method itself. But anyway, I agree with testing it.

iamacook:
@schmanu, @PooyaRaki, I added test coverage by tweaking an existing case in f95f4dc2974428499317594f3703c7dcf7b0e046.

schmanu (approved):
Thanks for adding the test case :)

PR #2559, opened 2025-04-30
feat: use Decoder Service for decoding data

### LINKED ISSUE #2533: Migrate transaction's decodedData source to Safe Decoder Service
### Description

**Context**
The Safe Transaction Service is not the proper source for `decodedData` on transactions. The CGW needs to change its source for transactions' decoded data to the new Safe Decoder Service.

### Requirements

Change the affected mappings:
- Multisig Transactions
- Transactions Queue
- Transactions History
- Get a transaction by ID.

### Additional information

_No response_

iamacook:
## Summary

Resolves #2533

The `dataDecoded` of a transaction entity is decoded by the Transaction Service, whereas we should be using the Decoder Service.

This removes validation of the aforementioned `dataDecoded`, replacing usage of it with that fetched from the Decoder Service.

## Changes

- Do not validated, ergo remove, `dataDecoded` from multisig, module and creation transactions.
- Replace all usage of `transaction.dataDecoded` with that fetched from the Decoder Service.
- Update tests accordingly.

### DIFF
--- a/src/domain/data-decoder/v2/data-decoder.repository.ts
+++ b/src/domain/data-decoder/v2/data-decoder.repository.ts
@@ -10,12 +10,16 @@ import {
   Contract,
   ContractPageSchema,
 } from '@/domain/data-decoder/v2/entities/contract.entity';
+import { Transaction } from '@/domain/safe/entities/transaction.entity';
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
+import { asError } from '@/logging/utils';
 
 @Injectable()
 export class DataDecoderRepository implements IDataDecoderRepository {
   constructor(
     @Inject(IDataDecoderApi)
     private readonly dataDecoderApi: IDataDecoderApi,
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
   ) {}
 
   public async getDecodedData(args: {
@@ -33,5 +37,59 @@ export class DataDecoderRepository implements IDataDecoderRepository {
   }): Promise<Page<Contract>> {
     const contracts = await this.dataDecoderApi.getContracts(args);
     return ContractPageSchema.parse(contracts);
+  }
+
+  public async getTransactionDataDecoded(args: {
+    chainId: string;
+    transaction: Transaction;
+  }): Promise<DataDecoded | null> {
+    const data = this.getDataDecodedData(args.transaction);
+
+    if (!data || data === '0x') {
+      return null;
+    }
+
+    try {
+      return await this.getDecodedData({
+        chainId: args.chainId,
+        data,
+        to: this.getDataDecodedTo(args.transaction),
+      });
+    } catch (e) {
+      this.loggingService.warn(
+        `Error decoding transaction data: ${asError(e).message}`,
+      );
+      return null;
+    }
+  }
+
+  private getDataDecodedData(transaction: Transaction): `0x${string}` | null {
+    // Multisig transaction
+    if ('data' in transaction) {
+      return transaction.data;
+    }
+    // Creation
+    if ('setupData' in transaction) {
+      return transaction.setupData;
+    }
+
+    throw Error('Unrecognized transaction type');
+  }
+
+  private getDataDecodedTo(transaction: Transaction): `0x${string}` {
+    // Multisig transaction
+    if ('to' in transaction) {
+      return transaction.to;
+    }
+    // Native transfer
+    if ('from' in transaction) {
+      return transaction.from;
+    }
+    // Creation
+    if ('factoryAddress' in transaction) {
+      return transaction.factoryAddress;
+    }
+
+    throw Error('Unrecognized transaction type');
   }
 }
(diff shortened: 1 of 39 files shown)


src/routes/transactions/transactions.service.ts
```diff
@@ -212,17 +241,22 @@ export class TransactionsService {
...
-              safeInfo,
-            ),
-            ConflictType.None,
+      domainTransactions.results.map(async (domainTransaction) => {
+        const dataDecoded =
+          await this.dataDecoderRepository.getTransactionDataDecoded({
+            chainId: args.chainId,
+            transaction: domainTransaction,
+          });
+        return new MultisigTransaction(
```
PooyaRaki, line 250: Nit: This should ideally happen outside the `Promise.all` in another loop. There are other instance of it in the same file but I won't mark them separately.

src/domain/data-decoder/v2/data-decoder.repository.ts
```diff
@@ -10,12 +10,16 @@ import {
   Contract,
```
PooyaRaki, line 10: Even though `data-decoder-api.service` is not part of this PR we should cache the requests to the decoder service.

src/routes/transactions/mappers/transaction-preview.mapper.ts
```diff
@@ -26,14 +26,23 @@ export class TransactionPreviewMapper {
...
     previewTransactionDto: PreviewTransactionDto,
   ): Promise<TransactionPreview> {
+    const multisigTransaction = {
+      safe: safe.address,
+      to: previewTransactionDto.to,
+      value: previewTransactionDto.value,
+      data: previewTransactionDto.data,
+      operation: previewTransactionDto.operation,
+      // Keep type safety as only the above are available in previewTransactionDto
+    } as MultisigTransaction;
```
PooyaRaki, line 36: Why can't we use Partial types instead of casting?

src/routes/transactions/entities/txs-multisig-transaction.entity.ts
```diff
@@ -15,8 +15,8 @@ export class TXSMultisigTransaction implements DomainMultisigTransaction {
   value: string;
   @ApiProperty()
   data: `0x${string}` | null;
-  @ApiProperty()
-  dataDecoded: DataDecoded | null;
+  // @ApiProperty()
+  // dataDecoded: DataDecoded | null;
```
PooyaRaki, line 19: Why is it not removed?

src/routes/transactions/entities/txs-creation-transaction.entity.ts
```diff
@@ -17,8 +16,8 @@ export class TXSCreationTransaction implements DomainCreationTransaction {
   setupData: `0x${string}` | null;
   @ApiProperty()
   saltNonce: string | null;
-  @ApiProperty()
-  dataDecoded: DataDecoded | null;
+  // @ApiProperty()
+  // dataDecoded: DataDecoded | null;
```
PooyaRaki, line 20: Why is it not removed?

src/routes/transactions/transactions.service.ts
```diff
@@ -153,14 +170,21 @@ export class TransactionsService {
           chainId: args.chainId,
           safeTransactionHash: args.txId,
         });
-        const safe = await this.safeRepository.getSafe({
-          chainId: args.chainId,
-          address: tx.safe,
-        });
+        const [safe, dataDecoded] = await Promise.all([
```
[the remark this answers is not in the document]
iamacook: As per discussion, we agree that this is a beneficial change but our target library does not support `Promise.any`. We will consider changing this separately.

src/routes/transactions/transactions.service.ts
iamacook: I split the decoding, mapping and instantiation into separate iterations in 4a396f4658e990c1b95eecbbdd04a48aca8ff2f1.

src/domain/data-decoder/v2/data-decoder.repository.ts
iamacook: Good point, I agree. I suggest we do so separately to keep the diff. small. I will open a separate PR after this.

src/routes/transactions/mappers/transaction-preview.mapper.ts
iamacook: As per discussion, we should `Pick` the required properties but, as this has an affect on the mapper and requires more type changes, we will do this separately.

src/routes/transactions/entities/txs-multisig-transaction.entity.ts
iamacook: This is a redundancy. Removed in 4a396f4658e990c1b95eecbbdd04a48aca8ff2f1.

src/routes/transactions/entities/txs-creation-transaction.entity.ts
iamacook: This is a redundancy. Removed in 4a396f4658e990c1b95eecbbdd04a48aca8ff2f1.

PooyaRaki (approved):

PR #2526, opened 2025-04-07
feat: map vault deposit transactions

hectorgomezv:
## Summary
This PR implements mappings for ERC-4626 deposit transactions.

## Changes
- Adds a `VaultDepositTransactionInfo` and its mappings in the routes layer.
- Extends `DefiVaultStats` and its schema entity with additional properties.
- Adds a new `KilnVaultHelper` to decode ERC-4626 deposits.
- Adds related test cases.

PR #2570, opened 2025-05-05
fix: cache Kiln deployments relative to env.

iamacook:
## Summary

As vault deployments are only deployed on Base, we [dynamically use the testnet configuration of Kiln](https://github.com/safe-global/safe-client-gateway/pull/2526#discussion_r2031467964). This dynamic value is not taken into account when caching, however.

If the deployments are initially fetched on a network other than Base, the prod. deployments will be cached meaning that any subsequent deployment requests on Base are returned from the prod. cache and vice versa. This includes the dynamic URL in the cache key so as to distinguish the both.

## Changes

- Include Kiln URL in cache key for deployments

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -596,8 +596,10 @@ export class CacheRouter {
     );
   }
 
-  static getStakingDeploymentsCacheDir(): CacheDir {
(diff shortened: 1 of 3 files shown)


usame-algan (approved):
Code looks good. Lets see if our assumption for this fix is correct 🙏

PR #2572, opened 2025-05-05
feat(amqp): add heartbeat and reconnect time configuration options

PooyaRaki:
## Summary
This PR introduces configuration options for the AMQP heartbeat interval and reconnect time, improving connection stability and management. Previously, these values were not configurable, which could result in the client prematurely closing the connection.

## Changes
- Explicitly pass `heartbeatIntervalInSeconds` to the RabbitMQ connection.
- Explicitly pass `reconnectTimeInSeconds` to the RabbitMQ connection.


src/datasources/queues/queues-api.module.ts
```diff
@@ -25,8 +25,17 @@ function queueConsumerFactory(
     configurationService.getOrThrow<string>('amqp.exchange.mode');
   const queue = configurationService.getOrThrow<string>('amqp.queue');
   const prefetch = configurationService.getOrThrow<number>('amqp.prefetch');
+  const heartbeatIntervalInSeconds = configurationService.getOrThrow<number>(
+    'amqp.heartbeatIntervalInSeconds',
+  );
+  const reconnectTimeInSeconds = configurationService.getOrThrow<number>(
+    'amqp.reconnectTimeInSeconds',
+  );
```
iamacook, line 33: Although I would prefer explicitly setting the default in our config, should we default to `null`, this generic is incorrect:

```suggestion
  const heartbeatIntervalInSeconds = configurationService.getOrThrow<number | null>(
    'amqp.heartbeatIntervalInSeconds',
  );
  const reconnectTimeInSeconds = configurationService.getOrThrow<number | null>(
    'amqp.reconnectTimeInSeconds',
  );
```

PooyaRaki: Already Fixed in a recent change. https://github.com/safe-global/safe-client-gateway/pull/2572/commits/247f9773c2be34488e62bc9cbd58092e38cd279a

src/config/entities/configuration.ts
```diff
@@ -59,6 +59,12 @@ export default () => ({
       process.env.AMQP_PREFETCH != null
         ? parseInt(process.env.AMQP_PREFETCH)
         : 100,
+    heartbeatIntervalInSeconds: process.env.AMQP_HEARBEAT_INTERVAL_SECONDS
+      ? +process.env.AMQP_HEARBEAT_INTERVAL_SECONDS
+      : null,
+    reconnectTimeInSeconds: process.env.AMQP_RECONNECT_TIME_SECONDS
+      ? +process.env.AMQP_RECONNECT_TIME_SECONDS
+      : null,
```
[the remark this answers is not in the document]
PooyaRaki: I would leave them to the library, if we want to change them then we can do so using the environment variables.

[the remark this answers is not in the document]
PooyaRaki: As per discussion, I change the default values ✅ 
https://github.com/safe-global/safe-client-gateway/pull/2572/commits/c626df450bd489d25689f19b5cc11a9e46923831

src/config/entities/__tests__/configuration.ts
```diff
@@ -33,6 +33,8 @@ export default (): ReturnType<typeof configuration> => ({
     exchange: { name: faker.string.sample(), mode: faker.string.sample() },
     queue: faker.string.sample(),
     prefetch: faker.number.int(),
+    heartbeatIntervalInSeconds: 60,
+    reconnectTimeInSeconds: 5,
```
iamacook, line 37: Nit: do these need to be hardcoded in test environments?

iamacook (approved):

src/config/entities/__tests__/configuration.ts
PooyaRaki: It’s best to avoid making them random, as that could potentially lead to issues.

PR #2574, opened 2025-05-07
fix: map `performance_fee` to vault `fee`

iamacook:
## Summary

We currently map the `product_fee` of a deployment as a Vault's `fee`. This is, however, incorrect and shouldn't relied upon. The `performance_fee` stats is the fee that should be shown.

This adds validation, and therefore infers into the type, the `performance_fee` to the `DefiVaultStats` and maps it to the `fee` of mapped Vault transactions.

## Changes

- Add validation of `performance_fee`, inferring it into the type
- Switch mapping of `fee` to be `performance_fee`


usame-algan (approved):
LGTM. I can see the `performance_fee` is now returned on the testnet API from Kiln through the network-stats endpoint

Screenshot 2025-05-07 at 15 23 24

iamacook:
> LGTM. I can see the `performance_fee` is now returned on the testnet API from Kiln through the network-stats endpoint

This is now on [mainnet](https://docs.api.kiln.fi/v1.8/reference/getdefinetworkstats) as well so I will now merge this.

PR #2577, opened 2025-05-12
fix: use mainnet staking for Base on prod.

iamacook:
## Summary

Due to Kiln deployment restrictions, we temporarily set Base to use their testnet deployments on prod. As we are now able to successfully test this by using a separate Transaction Service deployment, this removes the workaround code.

Note: we need to remove the `STAKING_BASE_PRODUCTION_ACTIVE` env. var. as well.

## Changes

- Remove usage of Kiln testnet deployment on prod. from `StakingApiManager`
- Remove all associated config.

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -600,8 +600,7 @@ export class CacheRouter {
     );
   }
 
-  // TODO: remove passing url and the associated configuration once Base is
-  // fully migrated to the Kiln mainnet API.
+  // Kiln uses different endpoints for mainnet/testnet
   static getStakingDeploymentsCacheDir(url: string): CacheDir {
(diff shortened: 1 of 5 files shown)


src/datasources/cache/cache.router.ts
```diff
@@ -600,8 +600,7 @@ export class CacheRouter {
     );
   }
 
-  // TODO: remove passing url and the associated configuration once Base is
-  // fully migrated to the Kiln mainnet API.
+  // Kiln uses different endpoints for mainnet/testnet
```
PooyaRaki, line 603: QQ: Why is the TODO removed but nothing in the code is removed ?

PooyaRaki (approved):

src/datasources/cache/cache.router.ts
iamacook: After further thought, it makes sense to keep the parameter as we would otherwise use the same cache entry for both endpoints.

Issue #2578, opened 2025-05-12
Add priority high to push notifications


compojoom:
### Description

Right now the push notifications are not sent with priority high, which means that Android won't deliver them to the app if it is in background mode.

> Background messages only work if the message priority is set to 'high' android.app.BackgroundServiceStartNotAllowedException: Not allowed to start service Intent

### Requirements

Add priority:high to push notifications sent to android.

### Additional information

https://firebase.google.com/docs/cloud-messaging/android/message-priority
https://developer.android.com/about/versions/oreo/background.html

PR #2579, opened 2025-05-12
feat: add priority:high to android push notification

compojoom:
## Summary
Ads a priority high flag on the push notification sent to android.

Resolves https://github.com/safe-global/safe-client-gateway/issues/2578

## Changes
By adding high priority to the push notification Android passes the notification to the app even when it is in the background.

### DIFF
--- a/src/datasources/push-notifications-api/entities/firebase-notification.entity.ts
+++ b/src/datasources/push-notifications-api/entities/firebase-notification.entity.ts
@@ -24,3 +24,9 @@ export type FireabaseNotificationApn = {
     };
   };
 };
+
+export type FirebaseAndroidMessageConfig = {
+  android: {
+    priority: 'high' | 'normal';
+  };
+};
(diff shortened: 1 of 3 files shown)


src/datasources/push-notifications-api/entities/firebase-notification.entity.ts
```diff
@@ -24,3 +24,9 @@ export type FireabaseNotificationApn = {
     };
   };
 };
+
+export type FirebaseAndroidMessageConfig = {
+  android: {
+    priority: 'high';
```
PooyaRaki, line 30: Nit:
```suggestion
    priority: 'normal' | 'high';
```

src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts
```diff
@@ -103,6 +104,20 @@ export class FirebaseCloudMessagingApiService implements IPushNotificationsApi {
     };
   }
 
+  /**
+   * Returns the Android message config for the notification.
+   *
+   * To ensure that Android delivers the notification to the app, even when it is in the background,
```
PooyaRaki, line 110: Nit: This line seems excessive since the method name is about Android config in general not only the priority and we might add more fields to it in future. If you feel it's extremely necessary to keep it I'd suggest we move it to line `117`.
I’d remove the comment since the priority field already clearly indicates its purpose.

PooyaRaki (approved):

iamacook (approved):

PR #2586, opened 2025-05-15
fix: also decode Vault `withdraw` transactions

iamacook:
## Summary

Currently, we decode the `redeem` methods of Vault iteractions (full withdrawal of assets). When partially withdrawing assets, however, the `withdraw` method is used instead.

This applies the existing decoding logic to `withdraw` transactions as the function signature matches that of `redeem`. It therefore also decodes to `VaultRedeemTransactionInfo` but this is irrelevant client-side.

## Changes

- Fallback to decoding `withdraw` transactions of Vaults
- Rename methods accordingly


usame-algan (approved):
LGTM 👍

PR #2587, opened 2025-05-15
fix: reduce caching on staging Base

iamacook:
## Summary

We reduce caching for Hoodi (prev. Holesky) as the Transaction Service is not maintained by us, meaning we receive no hooks used for invalidation.

As we need to temporarily test with Base on staging, this propagates the cache reduction for non-production Base.

## Changes

- Reduce caching of balances/Transaction Service on non-production Base.


usame-algan (approved):
Thanks!

PR #2588, opened 2025-05-16
feat: Add ORM query cache key and Redis integration for notifications tests

PooyaRaki:
## Summary
This PR introduces a cache key for ORM queries and update the notification repository integration tests to utilize Redis for caching, enhancing performance and scalability.

## Changes
- Added a new cache key for the Orm query cache.
- Changed the cache engine of notification integration tests to Redis

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -71,6 +71,7 @@ export class CacheRouter {
   private static readonly UNSUPPORTED_CHAIN_EVENT = 'unsupported_chain_event';
   private static readonly ZERION_BALANCES_KEY = 'zerion_balances';
   private static readonly ZERION_COLLECTIBLES_KEY = 'zerion_collectibles';
+  private static readonly ORM_QUERY_CACHE_KEY = 'orm_query_cache';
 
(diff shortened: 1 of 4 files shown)


src/datasources/cache/cache.router.ts
```diff
@@ -787,4 +788,21 @@ export class CacheRouter {
...
+   * @param {string} safeAddress - Safe address
+   *
+   * @returns {string} - Cache key
+   */
+  static getOrnCacheKey(
+    prefix: string,
+    chainId: string,
+    safeAddress: `0x${string}`,
+  ): string {
+    return `${CacheRouter.ORM_QUERY_CACHE_KEY}:${prefix}:${chainId}:${safeAddress}`;
```
usame-algan, line 806: This cache key has a different structure than the existing one (: instead of -). Could this cause issues once released or will it simply ignore existing cache entries and create new ones?

PooyaRaki: I’ve updated the cache key for consistency with other parts of the app.
This will cause existing cache entries to be ignored and new ones to be created.
The old entries will be automatically deleted once they expire.

PR #2594, opened 2025-05-22
fix: prevent earn/staking cache collision

iamacook:
## Summary

The datasources of stake and earn use the same API with differing API keys. The caches for each are, however, the same. This results in cache collision with data of one request being returned for the other.

This adds a new domain-specific field to the respective `CacheDir`s.

## Changes

- Add new `cacheType` field to the the cache routing.
- Instantiate/cache separate `KilnApi`s with the aforementioned field.
- Update tests accordingly.

### DIFF
--- a/src/datasources/cache/cache.router.ts
+++ b/src/datasources/cache/cache.router.ts
@@ -640,72 +640,65 @@ export class CacheRouter {
     );
   }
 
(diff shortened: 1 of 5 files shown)


usame-algan (approved):
LGTM. The usage of `field` for the `CacheDir` should work as Redis hget uses it to access a different key even if the keys are the same.

PR #2598, opened 2025-05-26
feat: add a delete notifications endpoint

### LINKED ISSUE #2597: Add endpoint for unregistering multiple push-notification subscriptions
### Description

Currently the endpoint that allows for the subscribing to push notifications allows us to send an array of chainIds and safes to be registered. This way we can pass an array of safes for the subscription and don't have to do multiple network requests per chain. That's very convinient. Sadly the unregister endpoint is not that flexible and there we need to create multiple requests if we want to unregister the safe across multiple chains. 

### Requirements

Create an endpoint that can unregister multiple safes from push-notifications at once. Just like the register endpoint. 

### Additional information

_No response_

compojoom:
## Summary
The endpoint allows the deletion of multiple safe push notification subscriptions at once, similar to the register endpoint where one can provide multiple safes to subscribe to.
fixes #2597

### DIFF
--- a/src/routes/notifications/v2/notifications.controller.ts
+++ b/src/routes/notifications/v2/notifications.controller.ts
@@ -2,6 +2,10 @@ import { AuthPayload } from '@/domain/auth/entities/auth-payload.entity';
 import { UpsertSubscriptionsDto } from '@/routes/notifications/v2/entities/upsert-subscriptions.dto.entity';
 import { UpsertSubscriptionsDtoSchema } from '@/domain/notifications/v2/entities/upsert-subscriptions.dto.entity';
 import { NotificationsServiceV2 } from '@/routes/notifications/v2/notifications.service';
+import {
+  DeleteSubscriptionsDto,
+  DeleteSubscriptionsDtoSchema,
+} from '@/domain/notifications/v2/entities/delete-subscriptions.dto.entity';
 import { Auth } from '@/routes/auth/decorators/auth.decorator';
 import { AuthGuard } from '@/routes/auth/guards/auth.guard';
 import { AddressSchema } from '@/validation/entities/schemas/address.schema';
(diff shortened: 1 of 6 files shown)


src/routes/notifications/v2/notifications.controller.ts
```diff
@@ -73,6 +77,32 @@ export class NotificationsControllerV2 {
...
+  })
+  @ApiBody({ type: DeleteSubscriptionsDto })
+  deleteSubscriptions(
+    @Param('deviceUuid', new ValidationPipe(UuidSchema)) deviceUuid: UUID,
+    @Body(new ValidationPipe(DeleteSubscriptionsDtoSchema))
+    deleteSubscriptionsDto: DeleteSubscriptionsDto,
+  ): Promise<void> {
+    return this.notificationsService.deleteSubscriptions({
+      deviceUuid,
+      safes: deleteSubscriptionsDto.safes.map((safe) => ({
```
PooyaRaki, line 99: Nit: For more readability we can extract this `map` to a `const`

src/routes/notifications/v2/notifications.service.ts
```diff
@@ -38,6 +38,19 @@ export class NotificationsServiceV2 {
     await this.notificationsRepository.deleteSubscription(args);
   }
 
+  async deleteSubscriptions(args: {
+    deviceUuid: UUID;
+    safes: Array<{ chainId: string; safeAddress: `0x${string}` }>;
+  }): Promise<void> {
+    for (const safe of args.safes) {
+      await this.notificationsRepository.deleteSubscription({
```
PooyaRaki, line 46: I suggest we avoid calling `deleteSubscription` in a loop and instead create a new repository method with a custom database query. Currently, passing 10 subscriptions results in at least 10 queries—possibly 20, depending on how we aak the ORM to load the device relation. By writing a dedicated method, we can fetch all related devices in a single query and remove all safes for those devices in another, reducing it to just two queries total.

PooyaRaki (commented):
Since the tests will most likely change after applying the suggested refactor, I haven't reviewed them.

PooyaRaki:
I'll close the PR for now. Please feel free to reopen it with a fix.

PR #2613, opened 2025-06-13
feat(Spaces): member alias functionality 

tmjssz:
Resolves [EN-68](https://linear.app/safe-global/issue/EN-68/spaces-be-give-users-the-possibility-to-rename-at-any-time)

## Summary

This PR implements a member alias feature that allows users to set and update their display names within spaces. The feature includes database schema changes, API endpoint, validation, and comprehensive testing coverage.

### Database
- **Migration**: Added `alias` column to `members` table with proper indexing
- **Entity**: Updated member database entity with nullable alias field

### Functionality
- **New Endpoint**: `PATCH /:spaceId/members/alias` for updating the member alias of the requesting user
- **Repository**: Added `updateAlias` method with proper user validation. Users can only update their own alias using authenticated user ID

### Testing
- **Repository Tests**: tests covering success/error cases
- **Controller Tests**: tests for API endpoint
- **Existing Tests**: Updated all member expectations to include `alias: null`

### Breaking Changes
- **Database**: Requires running new migration
- **API**: New endpoint available, member responses include alias field


migrations/1749755837793-add-member-alias.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddMemberAlias1749755837793 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+                  ALTER TABLE members ADD COLUMN alias VARCHAR(255);
```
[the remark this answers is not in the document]
PooyaRaki: @tmjssz Do we want users to be able to add an alias that is longer than name? Otherwise the length should be equal to `src/datasources/users/entities/member.entity.db.ts`

src/routes/spaces/members.controller.spec.ts
```diff
@@ -1981,4 +1984,248 @@ describe('MembersController', () => {
...
+        .expect(201);
+
+      const createSpaceResponse = await request(app.getHttpServer())
+        .post('/v1/spaces')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .send({ name: spaceName })
+        .expect(201);
+      const spaceId = createSpaceResponse.body.id;
+
+      await request(app.getHttpServer())
```
PooyaRaki, line 2007: It seems like we haven't added the user to the space, so we shouldn't be able to update an alias for a member that is not part of the space.
We're updating the alias of the admin here.

src/domain/users/members.repository.ts
```diff
@@ -305,6 +305,39 @@ export class MembersRepository implements IMembersRepository {
...
   }
 
+  public async updateAlias(args: {
+    authPayload: AuthPayload;
+    spaceId: Space['id'];
+    alias: Member['alias'];
+  }): Promise<void> {
+    this.assertSignerAddress(args.authPayload);
+
+    const wallet = await this.walletsRepository.findOne(
```
PooyaRaki, line 315: Since the find criteria is deeply nested, it’s unclear what the final query looks like. We need to ensure the ORM-generated query is efficient—otherwise, we should consider splitting it into multiple simpler queries.

src/domain/users/members.repository.ts
```diff
@@ -305,6 +305,39 @@ export class MembersRepository implements IMembersRepository {
...
+    if (!wallet) {
+      throw new NotFoundException(
+        'User not found or not a member of the space.',
+      );
+    }
+
+    const membersRepository =
+      await this.postgresDatabaseService.getRepository(DbMember);
+    const updateResult = await membersRepository.update(
+      { user: { id: wallet.user.id }, space: { id: args.spaceId } },
```
PooyaRaki, line 332: Since the find criteria is deeply nested, it’s unclear what the final query looks like. We need to ensure the ORM-generated query is efficient—otherwise, we should consider splitting it into multiple simpler queries.

src/datasources/users/entities/member.entity.db.ts
```diff
@@ -22,6 +22,7 @@ import {
 @Unique('UQ_members', ['user', 'space'])
 @Index('idx_members_name', ['name'])
 @Index('idx_members_role_status', ['role', 'status'])
+@Index('idx_members_alias', ['alias'])
```
PooyaRaki, line 25: Are we filtering members by alias or use this column in any `WHERE`, `JOIN`, `ORDER BY`, `GROUP BY` clauses? If not we don't need to index it.

src/routes/spaces/members.controller.spec.ts
tmjssz: Added another test case for when the user is a member, but no admin: be8d22d4e8fa4a3ab43513afa3dcceb462b7292a

src/domain/users/members.repository.ts
tmjssz: I replaced the complex query by a simple `findByWalletAddressOrFail` because the other conditions are already covered by the update query below. Commit: de263254bd4d6c5ca5cfedf4923395a459f98a71

src/domain/users/members.repository.ts
tmjssz: I don't think it makes sense to split this one into multiple queries. But I simplified the other query above instead.

PooyaRaki: Well, We don't necessarily need to split, but it's enough to check the generated queries to make sure they're efficient.

tmjssz: Here’s the generated query:
`UPDATE "members" SET "alias" = $1 WHERE ("user_id" = $2 AND "space_id" = $3)`
Looks good to me, or would you adapt anything?

PR #2628, opened 2025-06-27
feat: Remove signature management cache every hour, reduce event cache clear time to 30 seconds

PooyaRaki:
## Summary
This PR addresses a memory leak issue where the service's memory usage gradually increases. The root cause is unbounded memoization caches that accumulate entries without proper cleanup, leading to continuous memory growth in long-running instances.

## Changes
- Add hourly cleanup for SafeSignature cache: Clear the memoized signature owner cache every hour to prevent unbounded growth from cryptographic signature processing.
- Reduce EventCacheHelper cache cleanup interval: Change memoization cache clearing from every minute to every 30 seconds to reduce peak memory usage during high event processing periods.


LucieFaire (approved):
👍

PR #2631, opened 2025-07-01
fix: default contract.name to <empty string> if null

LucieFaire:
## Changes

- Default contract.name to`''` if the decoder service returned `null`

### DIFF
--- a/src/routes/contracts/contracts.controller.spec.ts
+++ b/src/routes/contracts/contracts.controller.spec.ts
@@ -149,7 +149,7 @@ describe('Contracts controller', () => {
         .expect(503);
     });
 
-    it('should not get a validation error if name is null', async () => {
+    it('Should pass validation if name is null', async () => {    
       const chain = chainBuilder().build();
       const contract = contractBuilder().build();
       const contractPage = pageBuilder()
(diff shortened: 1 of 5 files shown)


src/routes/contracts/mappers/contract.mapper.spec.ts
```diff
@@ -32,6 +32,13 @@ describe('Contract Mapper', () => {
     });
   });
 
+  it('should return name = "" if its null', () => {
```
PooyaRaki, line 35: Nit:
```suggestion
  it('Should return name = "" if it is null', () => {
```

src/routes/contracts/mappers/contract.mapper.ts
```diff
@@ -12,7 +12,7 @@ export class ContractMapper {
       : null;
     return {
       address: contract.address,
-      name: contract.name,
+      name: contract.name ?? '',
```
PooyaRaki, line 15: We don't need to change it here if we define a default value in the validation.

src/routes/contracts/contracts.controller.spec.ts
```diff
@@ -149,11 +149,46 @@ describe('Contracts controller', () => {
         .expect(503);
     });
 
-    it('should get a validation error', async () => {
+    it('Should pass validation if name is null', async () => {
```
PooyaRaki, line 152: Nit: it would be nice to have the same test for `displayName`

PooyaRaki (approved):

PR #2642, opened 2025-07-04
fix: Invalidates notification subscription cache when removing a subscription/device

PooyaRaki:
## Summary
This PR adds cache invalidation for notification subscriptions when a subscription or device is removed.

Previously, removing a subscription or device did not invalidate the associated cache. This could result in stale subscription data being served, leading to inconsistencies in push notifications. This fix explicitly clears the relevant cache entries on removal, ensuring consistency across notification states.

## Changes
- Adds cache invalidation logic to `notificationRepository.deleteDevice`
- Adds cache invalidation logic to `notificationRepository.deleteSubscription`
- Adds new tests


compojoom (approved):
keep the good work! 👍

PR #2647, opened 2025-07-14
fix(test): flaky tests

LucieFaire:
## Summary

 The fix addresses the root cause: database timestamps are set by the database server and may not align perfectly with application-captured timestamps due to network latency and processing time.


src/domain/spaces/spaces.repository.integration.spec.ts
```diff
@@ -151,11 +149,12 @@ describe('SpacesRepository', () => {
 
       expect(createdAt).toEqual(updatedAt);
 
-      expect(createdAt.getTime()).toBeGreaterThanOrEqual(before);
-      expect(createdAt.getTime()).toBeLessThanOrEqual(after);
+      // Verify that the timestamps are recent (within the last 5 seconds)
+      expect(Math.abs(createdAt.getTime() - before)).toBeLessThan(5000);
```
tmjssz, line 153: I think it would be good to verify that createdAt falls within the expected range as well.
Also might be helpful to extract the 5000 as a const.

tmjssz (commented):
Thanks for fixing it ✨ just a minor suggestion

src/domain/spaces/spaces.repository.integration.spec.ts
LucieFaire: [06f2f2e](https://github.com/safe-global/safe-client-gateway/pull/2647/commits/06f2f2e6de48809a7218893625fd315abcec330a)

tmjssz: Ah sorry, I missed that 👍

src/domain/spaces/spaces.repository.integration.spec.ts
```diff
@@ -151,11 +150,13 @@ describe('SpacesRepository', () => {
 
       expect(createdAt).toEqual(updatedAt);
 
-      expect(createdAt.getTime()).toBeGreaterThanOrEqual(before);
-      expect(createdAt.getTime()).toBeLessThanOrEqual(after);
+      // Verify that the timestamps are recent (within the last 5 seconds)
```
PooyaRaki, line 153: I find it hard to understand how the previous code was not functioning as expected. We got the current timestamp(before) and inserted an entity and then got the current timestamp(after) again and checked the timestamps inserted in the database against those.

LucieFaire: oh, I tried to find it, but I'm not sure I can see the pipelines that I have restarted. It does happen on my local machine though, from time to time.

PooyaRaki: My suggestion would be to fix that in a following PR if that happens again.

LucieFaire: what changed would you say should be excluded from this PR?

PooyaRaki: Thanks 👍

LucieFaire:
> Since we are fixing the flaky tests: Could you please look at [this](https://github.com/safe-global/safe-client-gateway/actions/runs/16413092535/job/46372631352?pr=2647) one as well?

I looked at it, and noticed that there was a duplicated module initiated. Otherwise, i have not found what could have caused it. Let's see if it happens in the future, but `faker.string.sample()` should never produce a valid token

PR #2651, opened 2025-07-16
feat: add signerAddress prop to DELETE subscriptions endpoint

compojoom:
## Summary
Resolves https://linear.app/safe-global/issue/COR-336/expose-signeraddress-to-the-delete-subscriptions-endpoint

adds a signerAddress prop to the delete subscriptions endpoint. This allows us to delete just a single signerAddress.

With this change the endpoint behaves as follows
- no signer address provided -> drop everything row related to the safeAddress and deviceId
- signerAddress provided -> drop only the rows that match the signerAddress and safeAddress and deviceId
- signerAddress set to null -> drop only the rows where signerAddress is NULL and SafeAddress and DeviceId match

### DIFF
--- a/src/domain/notifications/v2/entities/__tests__/delete-all-subscriptions.dto.builder.ts
+++ b/src/domain/notifications/v2/entities/__tests__/delete-all-subscriptions.dto.builder.ts
@@ -17,12 +17,7 @@ export function deleteAllSubscriptionsDtoBuilder(): IBuilder<DeleteAllSubscripti
           chainId: faker.string.numeric(),
           deviceUuid: faker.string.uuid() as UUID,
           safeAddress: getAddress(faker.finance.ethereumAddress()),
-          ...(faker.datatype.boolean() && {
-            signerAddress: faker.helpers.arrayElement([
-              getAddress(faker.finance.ethereumAddress()),
-              null,
-            ]),
-          }),
+          // signerAddress is undefined by default - use .with() to add it on demand
         };
       },
     ),
--- a/src/domain/notifications/v2/entities/__tests__/delete-all-subscriptions.dto.builder.ts
+++ b/src/domain/notifications/v2/entities/__tests__/delete-all-subscriptions.dto.builder.ts
@@ -17,6 +17,12 @@ export function deleteAllSubscriptionsDtoBuilder(): IBuilder<DeleteAllSubscripti
           chainId: faker.string.numeric(),
           deviceUuid: faker.string.uuid() as UUID,
           safeAddress: getAddress(faker.finance.ethereumAddress()),
+          ...(faker.datatype.boolean() && {
+            signerAddress: faker.helpers.arrayElement([
+              getAddress(faker.finance.ethereumAddress()),
+              null,
+            ]),
+          }),
         };
       },
     ),
--- a/src/domain/notifications/v2/entities/delete-all-subscriptions.dto.entity.ts
+++ b/src/domain/notifications/v2/entities/delete-all-subscriptions.dto.entity.ts
@@ -10,6 +10,7 @@ export const DeleteAllSubscriptionsDtoSchema = z.object({
         chainId: NumericStringSchema,
         deviceUuid: UuidSchema,
         safeAddress: AddressSchema,
+        signerAddress: AddressSchema.nullable().optional(),
       }),
     )
     .min(1, 'At least one subscription is required'),
--- a/src/domain/notifications/v2/entities/delete-all-subscriptions.dto.entity.ts
+++ b/src/domain/notifications/v2/entities/delete-all-subscriptions.dto.entity.ts
@@ -10,7 +10,7 @@ export const DeleteAllSubscriptionsDtoSchema = z.object({
         chainId: NumericStringSchema,
         deviceUuid: UuidSchema,
         safeAddress: AddressSchema,
-        signerAddress: AddressSchema.nullable().optional(),
+        signerAddress: AddressSchema.nullish(),
       }),
     )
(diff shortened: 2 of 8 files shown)


src/domain/notifications/v2/entities/delete-all-subscriptions.dto.entity.ts
```diff
@@ -10,6 +10,7 @@ export const DeleteAllSubscriptionsDtoSchema = z.object({
         chainId: NumericStringSchema,
         deviceUuid: UuidSchema,
         safeAddress: AddressSchema,
+        signerAddress: AddressSchema.nullable().optional(),
```
LucieFaire, line 13: can be replaced with `nullish()` - it is an equivalent

src/domain/notifications/v2/notifications.repository.integration.spec.ts
```diff
@@ -1187,5 +1187,254 @@ describe('NotificationsRepositoryV2', () => {
...
+      });
+
+      expect(subscriptions).toHaveLength(1);
+
+      // Update the subscription to have a signerAddress
+      await notificationSubscriptionRepository.update(subscriptions[0].id, {
+        signer_address: signerAddress,
+      });
+
+      const deleteAllSubscriptionsDto = [
```
LucieFaire, line 1224: could we please also use the builder in these tests

src/domain/notifications/v2/notifications.repository.spec.ts
```diff
@@ -645,5 +646,185 @@ describe('NotificationsRepositoryV2', () => {
         where: [],
       });
     });
+
+    it('Should include signerAddress in where conditions when provided', async () => {
+      const signerAddress = getAddress(faker.finance.ethereumAddress());
+      const deleteAllSubscriptionsDto = [
```
LucieFaire, line 652: builder ? :)

src/domain/notifications/v2/notifications.repository.ts
```diff
@@ -430,19 +430,40 @@ export class NotificationsRepositoryV2 implements INotificationsRepositoryV2 {
...
+    const whereConditions = args.subscriptions.map((subscription) => {
+      const baseCondition = {
+        chain_id: subscription.chainId,
+        safe_address: subscription.safeAddress,
+        push_notification_device: {
+          device_uuid: subscription.deviceUuid,
+        },
+      };
+
+      // Handle signerAddress: undefined (omitted) vs null (explicit) vs string (specific address)
```
LucieFaire, line 449: could  be simplified:
```
return subscription.signerAddress === undefined
  ? baseCondition
  : {
      ...baseCondition,
      signer_address:
        subscription.signerAddress === null
          ? IsNull()
          : subscription.signerAddress,
    };
```
And maybe we can leave one comment before describing all cases in one place

LucieFaire (commented):
looks legit, but left some small comments :)

src/domain/notifications/v2/notifications.repository.interface.ts
```diff
@@ -43,6 +43,7 @@ export interface INotificationsRepositoryV2 {
       chainId: string;
       deviceUuid: UUID;
       safeAddress: `0x${string}`;
+      signerAddress?: `0x${string}` | null;
```
PooyaRaki, line 46: We should only add `?` or `null` otherwise the type would be
```
`0x${string}` | null | undefined
```
It should ideally either be nullable or undefined.

src/domain/notifications/v2/notifications.repository.integration.spec.ts
```diff
@@ -1187,5 +1188,260 @@ describe('NotificationsRepositoryV2', () => {
...
+          push_notification_device: {
+            device_uuid: upsertSubscriptionsDto.deviceUuid as UUID,
+          },
+        },
+      });
+
+      expect(subscriptions).toHaveLength(1);
+
+      // Update the subscription to have a signerAddress
+      await notificationSubscriptionRepository.update(subscriptions[0].id, {
```
PooyaRaki, line 1221: We should insert a subscription with a `signer_address` from the beginning to test the actual flow instead of updating it here.

src/routes/notifications/v2/entities/delete-all-subscriptions.dto.entity.ts
```diff
@@ -11,6 +11,19 @@ export class DeleteAllSubscriptionItemDto {
 
   @ApiProperty()
   public readonly safeAddress!: `0x${string}`;
+
+  @ApiProperty({
+    type: 'string',
+    required: false,
+    nullable: true,
+    description:
```
PooyaRaki, line 19: Thanks for the thorough explanation. :clap

src/domain/notifications/v2/notifications.repository.interface.ts
PooyaRaki: Thanks for the explanation. The logic is quite complex.

src/domain/notifications/v2/entities/__tests__/delete-all-subscriptions.dto.builder.ts
```diff
@@ -17,6 +17,12 @@ export function deleteAllSubscriptionsDtoBuilder(): IBuilder<DeleteAllSubscripti
           chainId: faker.string.numeric(),
           deviceUuid: faker.string.uuid() as UUID,
           safeAddress: getAddress(faker.finance.ethereumAddress()),
+          ...(faker.datatype.boolean() && {
```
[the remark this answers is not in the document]
compojoom: removed it, now it needs to be explicitly set

PooyaRaki (approved):

PR #2680, opened 2025-08-15
fix: Invert positions feature flag to be enabled by default

usame-algan:
## Summary

Changes `FF_ZERION_POSITIONS` to `FF_ZERION_POSITIONS_DISABLED` and inverts the logic so the feature is turned on by default and can be turned off via env variable


LucieFaire (approved):
👍

PR #2682, opened 2025-08-19
refactor(address-book): Update name validation to use dynamic max length constant

PooyaRaki:
## Summary

fixes https://linear.app/safe-global/issue/COR-489/spaces-back-end-and-front-end-have-diff-validation-for-the-name-field

This PR refactors `NameSchema` to support a dynamic max length. Instead of being tied to a single hardcoded limit, the schema can now be configured with different maximum lengths depending on context.

## Changes
- Extracted the common name validation logic into a base schema.
- Introduced a factory function to generate a `NameSchema` with a configurable max length.
- Updated existing `NameSchema` in the `AddressBookItemSchema` entity to use the new dynamic version with a different max length, `ADDRESS_BOOK_NAME_MAX_LENGTH`.
- update database to respect the new maxLength on the name field

### DIFF
--- a/migrations/1755621780124-addressBookItemLength.ts
+++ b/migrations/1755621780124-addressBookItemLength.ts
@@ -1,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class AddressBookItemLength1755621780124 implements MigrationInterface {
+  name = 'AddressBookItemLength1755621780124';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `ALTER TABLE "space_address_book_items" ALTER COLUMN "name" TYPE character varying(50)`,
+    );
+  }
+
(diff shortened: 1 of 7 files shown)


usame-algan (approved):
Looks good! I made sure that the db migration ran and was able to insert an entry with a name of length 50

migrations/1755621780124-addressBookItemLength.ts
```diff
@@ -0,0 +1,17 @@
+import { MigrationInterface, QueryRunner } from 'typeorm';
```
usame-algan, line 1: On localdev I get a lint warning here since these are only used as types.

LucieFaire (approved):

migrations/1755621780124-addressBookItemLength.ts
compojoom: I'm running yarn lint and it is fine for me?

PR #2684, opened 2025-08-21
fix: migrate getTrustedForDelegateCallContracts endpoint

LucieFaire:
## Summary

[COR-321](https://linear.app/safe-global/issue/COR-321/migrate-gettrustedfordelegatecallcontracts-to-decoder-service-endpoint)

## Changes
- migrated `getTrustedForDelegateCallContracts` endpoint from Tx Service to Decoder Service
- added a new endpoint in `DataDecoderApi`
- removed unused endpoint from `TransactionApi`
- removed the hoodi chain logic
- updated tests

### DIFF
--- a/src/datasources/data-decoder-api/data-decoder-api.service.ts
+++ b/src/datasources/data-decoder-api/data-decoder-api.service.ts
@@ -11,16 +11,10 @@ import { CacheRouter } from '@/datasources/cache/cache.router';
 
 @Injectable()
 export class DataDecoderApi implements IDataDecoderApi {
-  private static readonly HOODI_CHAIN_ID = '560048';
-  // TODO: Remove after Vault decoding has been released
-  private static readonly BASE_CHAIN_ID = '8453';
-
   private readonly baseUrl: string;
   private readonly defaultNotFoundExpirationTimeSeconds: number;
   private readonly defaultExpirationTimeSeconds: number;
   private readonly contractNotFoundExpirationTimeSeconds: number;
-  private readonly hoodiExpirationTimeSeconds: number;
-  private readonly isProduction: boolean;
 
   constructor(
     @Inject(IConfigurationService)
@@ -32,13 +26,6 @@ export class DataDecoderApi implements IDataDecoderApi {
     this.baseUrl = this.configurationService.getOrThrow<string>(
       'safeDataDecoder.baseUri',
     );
-    this.isProduction = this.configurationService.getOrThrow<boolean>(
-      'application.isProduction',
-    );
-    this.hoodiExpirationTimeSeconds =
-      this.configurationService.getOrThrow<number>(
-        'expirationTimeInSeconds.hoodi',
-      );
     this.defaultNotFoundExpirationTimeSeconds =
       this.configurationService.getOrThrow<number>(
         'expirationTimeInSeconds.notFound.default',
@@ -71,15 +58,12 @@ export class DataDecoderApi implements IDataDecoderApi {
     }
   }
 
-  //TODO expiration time as it used to be with hoodi?
   public async getContracts(args: {
     address: `0x${string}`;
     chainId: string;
     limit?: number;
     offset?: number;
   }): Promise<Raw<Page<Contract>>> {
-    const { notFoundExpireTimeSeconds, expireTimeSeconds } =
-      this.setCacheExpirationTimes(args.chainId, true);
     try {
       const url = `${this.baseUrl}/api/v1/contracts/${args.address}`;
       return await this.dataSource.get<Page<Contract>>({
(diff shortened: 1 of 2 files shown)


src/datasources/transaction-api/transaction-api.service.ts
```diff
@@ -91,10 +88,6 @@ export class TransactionApi implements ITransactionApi {
         this.configurationService.getOrThrow<number>(
           'expirationTimeInSeconds.notFound.token',
         );
-      this.contractNotFoundExpirationTimeSeconds =
```
LucieFaire, line 94: @PooyaRaki I don't know whether we still need this setting for Decoder endpoint or if the default value is enough. Lets sync on Monday

src/datasources/data-decoder-api/data-decoder-api.service.ts
```diff
@@ -11,8 +11,16 @@ import { CacheRouter } from '@/datasources/cache/cache.router';
 
 @Injectable()
 export class DataDecoderApi implements IDataDecoderApi {
+  private static readonly HOODI_CHAIN_ID = '560048';
+  // TODO: Remove after Vault decoding has been released
+  private static readonly BASE_CHAIN_ID = '8453';
```
usame-algan, line 16: Vault decoding has been released for a while now. Looking at the git history it looks like we needed this to reduce caching time on Base. Is it still necessary?

src/datasources/data-decoder-api/data-decoder-api.service.ts
```diff
@@ -72,4 +130,32 @@ export class DataDecoderApi implements IDataDecoderApi {
...
+        };
+  }
+
+  private isHoodyChain(chainId: string): boolean {
+    return (
+      chainId === DataDecoderApi.HOODI_CHAIN_ID ||
+      // TODO: Remove after Vault decoding has been released
+      (!this.isProduction && chainId === DataDecoderApi.BASE_CHAIN_ID)
+    );
+  }
```
usame-algan, line 160: Same here I guess. I found this: https://github.com/safe-global/safe-client-gateway/pull/2575
@compojoom was this needed for native staking and is it still needed?

src/domain/contracts/contracts.repository.ts
```diff
@@ -40,12 +35,12 @@ export class ContractsRepository implements IContractsRepository {
...
     contractAddress: `0x${string}`;
-  }): Promise<DataDecoderContract> {
+  }): Promise<Contract> {
     const contracts = await this.dataDecoderApi.getContracts({
       address: args.contractAddress,
-      chainIds: [args.chainId],
+      chainId: args.chainId,
     });
-    const { count, results } = DataDecoderContractPageSchema.parse(contracts);
+    const { count, results } = ContractPageSchema.parse(contracts);
```
usame-algan, line 43: Its interesting that both decoder service and transaction service return an array even though its always only one element. Do you know if its possible for them to return more than one result?

usame-algan (approved):
Code looks good. The caching exceptions for hoodi and base might be outdated so lets check that. From my tests, the endpoint still returns the same as before 👍

src/domain/contracts/contracts.repository.ts
LucieFaire: For one specific contract: Tx service always returns 1 contract back, but the Decoder service is multichain so it can return the same contract for many chains.
For trusted contracts: both return multiple contracts, might even be multiple pages. But again Decoder service is multichain, so its even more values :)

src/datasources/data-decoder-api/data-decoder-api.service.ts
LucieFaire: thank you for pointing it out, i will verify

src/domain/contracts/contracts.repository.ts
usame-algan: So when we call `getContracts` and pass a chainId we limit the response to just one element instead of multiple? I think in that case it is fine to access the first element in the array that is returned but I wonder if there is a future opportunity to cache data that is already returned but not needed e.g. for multichain.

LucieFaire: yeah I think so, we would just need to change the mechanics of fetching it from the cache

src/datasources/data-decoder-api/data-decoder-api.service.ts
LucieFaire: @PooyaRaki do you by chance remember if the vault decoding was implemented? I checked the git history but did not find anything related

PooyaRaki: @LucieFaire Yes, vault decoding is implemented. We can remove it and ensure that it works as expected afterwards.

LucieFaire: [92e8e75](https://github.com/safe-global/safe-client-gateway/pull/2684/commits/92e8e755ed5c67c27e515d7355ad5bdf52456ce5)

PR #2685, opened 2025-08-27
fix: Remove network limitation for positions api service

usame-algan:
## Summary

Resolves [COR-534](https://linear.app/safe-global/issue/COR-534/update-zerion-chainid-list-for-top-10-networks)

## Changes

- Removes `_getChainName` from positions api service
- Instead uses the chain name that is passed down

### DIFF
--- a/src/datasources/positions-api/zerion-positions-api.service.ts
+++ b/src/datasources/positions-api/zerion-positions-api.service.ts
@@ -1,5 +1,4 @@
 import { IConfigurationService } from '@/config/configuration.service.interface';
-import { ChainAttributes } from '@/datasources/balances-api/entities/provider-chain-attributes.entity';
 import {
   ZerionAttributes,
   ZerionBalance,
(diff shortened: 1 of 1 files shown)


LucieFaire (approved):
Looks good, and I agree with the overhead of maintaining this list. 
But in case it is important to keep it, we might as well extract it to a `env` config object and have a default preselected. I dont have strong preferences, but as both require a new deployment anyway, we can keep the configuration and make it env-specific.

usame-algan:
> But in case it is important to keep it, we might as well extract it to a env config object and have a default preselected. I dont have strong preferences, but as both require a new deployment anyway, we can keep the configuration and make it env-specific.

I am not sure if we need to have this granularity in the future but right now we already have multiple feature flags:

- Frontend feature flag per chain coming from the config service to turn positions on/off on the frontend
- CGW .env feature flag to turn the positions endpoint etc. on/off which is not chain specific
- This chains list that allows turning it on/off on a per chain basis in the CGW but this achieves the same as the frontend feature flag.

So for now I am for removing this and in case we need this granularity in the future e.g. we want to have positions for certain chains on mobile but not on web or something else we could reintroduce logic like this.

PR #2687, opened 2025-09-01
feat: Update API documentation

PooyaRaki:
## Summary
This PR fills gaps in our `Swagger/OpenAPI` spec by adding missing annotations and documentation so every exposed endpoint is now discoverable, accurately typed, and consistently described in the API reference.

## Changes
- Added missing @swagger/OpenAPI annotations for previously undocumented endpoints.


LucieFaire:
@claude review

LucieFaire (approved):
I scanned through and it looks fine to me. Maybe there are some places where we can be more concrete with the summary, but all in all looks good

PR #2697, opened 2025-09-03
feat: add nonce field to CSV export

LucieFaire:
## Changes

* Add a new field `nonce` in the CSV export file


src/modules/csv-export/v1/entities/__tests__/transaction-export.builder.ts
```diff
@@ -27,7 +27,8 @@ export function transactionExportBuilder(): IBuilder<TransactionExport> {
     .with('executedAt', faker.date.recent())
     .with('note', faker.lorem.sentence())
     .with('transactionHash', faker.string.hexadecimal({ length: 64 }) as Hex)
-    .with('contractAddress', getAddress(faker.finance.ethereumAddress()));
+    .with('contractAddress', getAddress(faker.finance.ethereumAddress()))
+    .with('nonce', faker.number.int({ min: 0, max: 100 }).toString());
```
PooyaRaki, line 31: Nit: Is there a reason the maximum nonce is 100?

src/modules/csv-export/v1/entities/transaction-export.entity.ts
```diff
@@ -22,6 +22,7 @@ export const TransactionExportSchema = z
     note: z.string().nullable(),
     transactionHash: HexSchema,
     contractAddress: AddressSchema.nullable(),
+    nonce: z.string().nullable(),
```
PooyaRaki, line 25: QQ: Why is the nonce nullable?

LucieFaire: it can be null according to the Transaction Service spec

src/modules/csv-export/v1/entities/__tests__/transaction-export.builder.ts
LucieFaire: no just a random number, i will remove the constraint

PooyaRaki (approved):

PR #2699, opened 2025-09-05
fix(positions): remove position value aggregation

vseehausen:
## Summary

We noticed multiple inconsistencies and decided to simplify the logic and just display positions 1-to-1, like in the Zerion API.

## Changes

* removes the aggregation of single positions per protocol
* reworks sum of fiat value

### DIFF
--- a/src/domain/common/utils/utils.ts
+++ b/src/domain/common/utils/utils.ts
@@ -2,9 +2,9 @@ import { createHash } from 'crypto';
 import type { BinaryLike } from 'crypto';
 import type { Address } from 'viem';
 
-// We use the maximum value in order to preserve all decimals
+// We use a high value to preserve decimals (max allowed is 20)
 // @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#maximumfractiondigits
-const MAX_MAXIMUM_FRACTION_DIGITS = 100;
+const MAX_MAXIMUM_FRACTION_DIGITS = 20;
 
 const formatter = new Intl.NumberFormat('en-US', {
   // Prevent scientific notation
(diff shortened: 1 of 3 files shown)


usame-algan:
> We noticed multiple inconsistencies

What are the inconsistencies that these changes solve? Please provide an example on how to reproduce.

src/domain/common/utils/utils.ts
```diff
@@ -2,9 +2,9 @@ import { createHash } from 'crypto';
 import type { BinaryLike } from 'crypto';
 import type { Address } from 'viem';
 
-// We use the maximum value in order to preserve all decimals
+// We use a high value to preserve decimals (max allowed is 20)
 // @see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#maximumfractiondigits
-const MAX_MAXIMUM_FRACTION_DIGITS = 100;
+const MAX_MAXIMUM_FRACTION_DIGITS = 20;
```
usame-algan, line 7: Why is this change necessary?

src/routes/positions/positions.service.spec.ts
```diff
@@ -161,13 +161,20 @@ describe('PositionsService', () => {
...
+        expect.objectContaining({
+          position_type: PositionType.deposit,
+          balance: '10',
+          fiatBalance: '30',
+        }),
+        expect.objectContaining({
+          position_type: PositionType.deposit,
+          balance: '5',
+          fiatBalance: '15',
+        }),
```
usame-algan, line 176: On Zerion, multiple positions of the same type are aggregated. Why don't we want to do this anymore?

vseehausen:
Closed in favor of https://github.com/safe-global/safe-client-gateway/pull/2700.

PR #2700, opened 2025-09-08
feat(positions): remove custom logic

vseehausen:
## Summary

We noticed multiple inconsistencies and decided to simplify the logic and just display positions 1-to-1, like in the Zerion API.

## Changes

* removes the aggregation of single positions per protocol
* reworks the sum of fiat value

### DIFF
--- a/src/domain/common/utils/utils.ts
+++ b/src/domain/common/utils/utils.ts
@@ -2,9 +2,9 @@ import { createHash } from 'crypto';
 import type { BinaryLike } from 'crypto';
 import type { Address } from 'viem';
 
-// We use the maximum value in order to preserve all decimals
+// We use a high value to preserve decimals (max allowed is 20)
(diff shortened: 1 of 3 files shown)


usame-algan:
Please check my comments in the other PR as this is a duplication.

katspaugh:
@vseehausen and I just tested together carefully comparing the old and new responses along with Zerion's raw response. This branch fixes (at least) two issues:
* Wrong token balances in Balancer pools (possibly other positions as well) due to summation.
* Missing tokens in Balancer pool token pairs–we were only showing one token of the pair, which was probably intentional but doesn't seem right.

Approved the PR. ✅

usame-algan:
Does this change require any frontend changes? If so can you link a PR where this can be tested on the frontend?

vseehausen:
And this is how the Balancer positions look like.

Screenshot 2025-09-09 at 18 07 59

very similar to the Zerion UI:

Screenshot 2025-09-09 at 18 09 11

to me this is better than before. We should tackle the position merging in a separate PR.

LucieFaire (approved):

PR #2707, opened 2025-09-10
chores: update MAX_MAXIMUM_FRACTION_DIGITS

vseehausen:
## Summary

resolves an error occurring in runtime:
```
const formatter = new Intl.NumberFormat('en-US', {
                  ^

RangeError: maximumFractionDigits value is out of range.
```

This showed up in other PRs, which could take more time to merge, so I extracted this change to be discussed separately.

## Changes

updates maximumFractionDigits to 20.


katspaugh (approved):
Let's merge this already and be done with it. Not worth such a long discussion.

PooyaRaki (approved):
Since it'll unblock you and helps moving forward faster and since the tests are green, I'll approve

PR #2715, opened 2025-09-25
Feat(Positions): Filter out dust positions from Zerion data

katspaugh:
Part of https://linear.app/safe-global/issue/GRO-72/correct-portfolio-view-at-safe-specifically-avoid-double-accounting

## Summary
- filter Zerion-derived positions below $0.01 before grouping and drop empty protocols
- add helper to detect dust positions and reuse filtered data for protocol totals
- extend positions service tests to cover dust filtering and mock number formatting

## Testing
- yarn test src/routes/positions/positions.service.spec.ts

------
https://chatgpt.com/codex/tasks/task_b_68d4f9a04390832fb84977f12cd1004e

### DIFF
--- a/src/routes/positions/positions.service.spec.ts
+++ b/src/routes/positions/positions.service.spec.ts
@@ -9,7 +9,7 @@ import type { Chain } from '@/domain/chains/entities/chain.entity';
 import type { Address } from 'viem';
 
 jest.mock('@/domain/common/utils/utils', () => ({
-  getNumberString: (value: number) => value.toString(),
+  getNumberString: (value: number): string => value.toString(),
 }));
 
 const positionsRepoMock = jest.mocked({
--- a/src/routes/positions/positions.service.spec.ts
+++ b/src/routes/positions/positions.service.spec.ts
@@ -8,6 +8,10 @@ import { NULL_ADDRESS } from '@/routes/common/constants';
 import type { Chain } from '@/domain/chains/entities/chain.entity';
 import type { Address } from 'viem';
 
+jest.mock('@/domain/common/utils/utils', () => ({
+  getNumberString: (value: number) => value.toString(),
+}));
+
 const positionsRepoMock = jest.mocked({
   getPositions: jest.fn(),
 } as jest.MockedObjectDeep<IPositionsRepository>);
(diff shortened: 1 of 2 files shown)


src/routes/positions/positions.service.ts
```diff
@@ -142,4 +152,12 @@ export class PositionsService {
...
     };
   }
+
+  private _isDust(position: PositionEntry): boolean {
+    const fiatBalance = Number(position.fiatBalance);
+    if (Number.isNaN(fiatBalance)) {
+      return true;
+    }
+    return Math.abs(fiatBalance) < PositionsService.DUST_THRESHOLD_USD;
+  }
```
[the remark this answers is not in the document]
katspaugh: I think it's a very edge case, so let's ignore it.

src/routes/positions/positions.service.ts
```diff
@@ -61,23 +61,33 @@ export class PositionsService {
     return groupBy(positions, (position) => position.protocol ?? 'unknown');
   }
 
+  private static readonly DUST_THRESHOLD_USD = 0.01;
```
LucieFaire, line 64: could we move this to the top of the file?

LucieFaire (approved):
pre-approving but would like these 2 comments addressed 👍

src/routes/positions/positions.service.ts
katspaugh: 🫡

src/routes/positions/positions.service.spec.ts
```diff
@@ -8,6 +8,10 @@ import { NULL_ADDRESS } from '@/routes/common/constants';
 import type { Chain } from '@/domain/chains/entities/chain.entity';
 import type { Address } from 'viem';
 
+jest.mock('@/domain/common/utils/utils', () => ({
```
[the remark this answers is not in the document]
LucieFaire: oh we had this before in some PR, where codex added it everywhere. I dont think we need to mock it, cant we use the actual implementation?

vseehausen:
Can we increase UX by filtering out dust positions at the frontend and optionally including them? Similar to the default tokens/all tokens dropdown?

vseehausen:
Good point. For tokens it's a requirement by QA to see custom tokens. So I wonder if this applies here, too. 

The query would be nice. Can exclude dust positions by default. 

Is this something for the v2 endpoint?

katspaugh:
Screenshot 2025-09-25 at 16 36 10

@LucieFaire @PooyaRaki any idea why this test is failing? Seems unrelated.

src/routes/positions/positions.service.ts
```diff
@@ -142,4 +152,12 @@ export class PositionsService {
       application_metadata: position.application_metadata,
     };
   }
+
+  private _isDust(position: PositionEntry): boolean {
+    const fiatBalance = Number(position.fiatBalance);
+    if (Number.isNaN(fiatBalance)) {
+      return true;
+    }
+    return Math.abs(fiatBalance) < DUST_THRESHOLD_USD;
```
PooyaRaki, line 161:
```suggestion
    return Number.isNaN(fiatBalance) || Math.abs(fiatBalance) < DUST_THRESHOLD_USD;
```

src/routes/positions/positions.service.ts
```diff
@@ -14,6 +14,8 @@ import { z } from 'zod';
 import { PositionType } from '@/domain/positions/entities/position-type.entity';
 import type { Address } from 'viem';
 
+const DUST_THRESHOLD_USD = 0.01;
```
PooyaRaki, line 17: Nit: We usually move these constants to a separate file to keep them centralized.

src/routes/positions/positions.service.ts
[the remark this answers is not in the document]
PooyaRaki: @katspaugh Why do you think it's a very edge case? What happens if the provider doesn't have give us a value for the fiatBalance ?

src/routes/positions/positions.service.ts
```diff
@@ -64,20 +66,28 @@ export class PositionsService {
...
+
+    if (!filteredPositions.length) {
+      return null;
+    }
+
+    const nameGroups = this._groupByName(filteredPositions);
     const positionGroups = Object.entries(nameGroups).map(([name, group]) =>
       this._mapPositionGroup(name, group),
     );
     // Calculate fiat total from all individual positions
```
PooyaRaki, line 82: Nit: Since we’re already modifying `fiatTotal` in this PR, it might make sense to move it into a separate small function for cleaner code. Based on the earlier comment, it seems like a strong candidate for its own function.

katspaugh:
Guys, we gotta stop nit-picking and move on. We can't review each PR for several days.

src/routes/positions/positions.service.ts
katspaugh: I think this is out of scope for this PR. We'll be working further on the Portfolio, and I'll keep your suggestion in mind for the future.

src/routes/positions/positions.service.ts
katspaugh: Happy to move it (again). Which file should I move it to?

PooyaRaki: If there is no config file for this module, we can create one e.g.
https://github.com/safe-global/safe-client-gateway/blob/3ca6e1c2908dc16ec6b30db7c22ada0af04e8774/src/domain/common/entities/jobs.constants.ts

PR #2720, opened 2025-09-30
feat: Detect delegateCall

LucieFaire:
## Changes
- extended the `extractContracts` method to provide the operation type to determine if `delegateCall`
- extended the `verifyContract` method to check the `delegateCall` trustworthiness
- unit test coverage


src/modules/safe-shield/contract-analysis/contract-analysis.service.ts
```diff
@@ -92,45 +99,52 @@ export class ContractAnalysisService {
...
-   * @returns The analysis result.
+   * @param args.isDelegateCall - Whether the contract is called via delegateCall.
+   * @returns A pair of analysis results: [verification result, delegateCall result (if available)].
    */
   async verifyContract(args: {
     chainId: string;
     contract: Address;
-  }): Promise<ContractAnalysisResult> {
+    isDelegateCall: boolean;
+  }): Promise<[ContractAnalysisResult, ContractAnalysisResult | undefined]> {
```
tmjssz, line 131: I think we could make it more type-safe like this:
```suggestion
  }): Promise<[ContractAnalysisResult, AnalysisResult<'UNEXPECTED_DELEGATECALL'> | undefined]> {
```

If you also update the signature of `checkDelegateCall` accordingly and use a generic for `mapToAnalysisResult`, it should work:
```ts
  private mapToAnalysisResult<T extends ContractStatus>(args: {
    type: T;
    interactions?: number;
    name?: string;
  }): AnalysisResult<T> {
```

src/modules/safe-shield/contract-analysis/contract-analysis.service.ts
```diff
@@ -173,6 +191,20 @@ export class ContractAnalysisService {
...
+   * Checks if a delegateCall is unexpected based on the contract's trust status.
+   * @param isDelegateCall - Whether the call is a delegateCall.
+   * @param contract - The contract details (if available).
+   * @returns The analysis result (if available).
+   */
+  private checkDelegateCall(
+    isDelegateCall: boolean,
+    contract: Contract | undefined,
+  ): ContractAnalysisResult | undefined {
+    if (!isDelegateCall || contract?.trustedForDelegateCall) return undefined;
```
tmjssz, line 204: This logic won’t produce a warning if the contract object is undefined. I think in that case, we should return an UNEXPECTED_DELEGATECALL warning as a default. WDYT?

src/modules/safe-shield/contract-analysis/contract-analysis.service.ts
LucieFaire: [50659a7](https://github.com/safe-global/safe-client-gateway/pull/2720/commits/50659a78acaf8bf4c3c00d2ed1dff66e656b2a40)

tmjssz (approved):
🙌

PR #2721, opened 2025-10-01
refactor(safe-shield): Enhanced transaction decoding via TransactionsService

tmjssz:
## Summary
Updates the Safe Shield module to use the `TransactionsService` for transaction decoding instead of relying on the `DataDecodedModule`.

By leveraging the `previewTransaction` call, we also gain access to additional details. So this change enables accurate decoding of bridge transactions and lays the groundwork for [COR-565](https://linear.app/safe-global/issue/COR-565/bridge-analysis).

### DIFF
--- a/src/modules/safe-shield/utils/extraction.utils.spec.ts
+++ b/src/modules/safe-shield/utils/extraction.utils.spec.ts
@@ -81,6 +81,21 @@ describe('extraction.utils', () => {
       expect(result).toEqual([]);
     });
 
+    it('ignores transactions without data payload', () => {
+      const result = extractContracts(
+        [
+          createTransaction({
+            data: undefined,
+          }),
+        ],
+        mockErc20Decoder,
+      );
+
+      expect(result).toEqual([]);
+      expect(mockErc20Decoder.helpers.isTransfer).not.toHaveBeenCalled();
+      expect(mockErc20Decoder.helpers.isTransferFrom).not.toHaveBeenCalled();
+    });
+
     it('ignores transactions without decoded data', () => {
       const result = extractContracts(
         [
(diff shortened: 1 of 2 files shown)


src/modules/safe-shield/safe-shield.service.ts
```diff
@@ -43,128 +35,61 @@ export class SafeShieldService {
...
-   * Analyzes contracts in a transaction, including inner calls if it's a multiSend.
-   *
-   * @param args - Analysis parameters
-   * @param args.chainId - The chain ID
-   * @param args.safeAddress - The Safe address
-   * @param args.to - The transaction contract address
-   * @param args.data - The transaction data (may contain multiSend)
-   * @returns Map of contract addresses to their analysis results
-   */
-  async analyzeContracts(args: {
```
LucieFaire, line 79: where did it go?

src/modules/safe-shield/utils/recipient-extraction.utils.ts
```diff
@@ -27,29 +28,28 @@ export function extractRecipients(
...
-    dataDecoded?.parameters?.[2].value === '0x'
-  ) {
-    return getAddress(dataDecoded?.parameters?.[0].value as string);
+  if (isExecTransaction(tx) && tx.dataDecoded.parameters[2].value === '0x') {
+    return getAddress(tx.dataDecoded.parameters[0].value);
   }
 
   // ERC-20 transfer
-  if (erc20Decoder.helpers.isTransfer(data)) {
+  if (!!data && erc20Decoder.helpers.isTransfer(data)) {
```
LucieFaire, line 42: why do we need to check for existence of data? Wouldnt the second part return false if data isnt there?

tmjssz: It wouldn’t return false, it would actually fail. That’s because the function `isTransfer` expects data to be defined and [calls .startsWith on it](https://github.com/safe-global/safe-client-gateway/blob/feat%2Fbridge-analysis/src/domain/contracts/decoders/abi-decoder.helper.ts#L39). Without the existence check, if data is undefined, you’d get a runtime error like “can’t call .startsWith of undefined”.

src/modules/safe-shield/utils/recipient-extraction.utils.ts
```diff
@@ -27,29 +28,28 @@ export function extractRecipients(
  * @returns The recipient address or undefined if the transaction is not a transfer.
  */
 export function extractRecipient(
```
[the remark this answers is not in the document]
tmjssz: Done -> 3abf667879e4ea53ada3b66c85fbff4a9e3b7bd2

PR #2730, opened 2025-10-02
feat: Allow migration from unsupported to supported Safe mastercopies

katspaugh:
Resolves https://linear.app/safe-global/issue/REVE-391/opbnb-safes-are-not-usable-with-wallet-due-to-the-unofficial

## Summary

Enables Safes with unsupported base contracts to migrate to official mastercopies via both relay transactions and normal transaction proposals.

## Problem

Previously, Safes with unsupported mastercopies were blocked from:
1. **Transaction proposals**: The Transaction Service returns `version: null` for unsupported Safes, causing `getSafeTxHash()` to fail

This created a catch-22 where users couldn't migrate their Safes to supported mastercopies.

### Transaction Proposals (`src/domain/common/utils/safe.ts`)
- When `Safe.version` is `null`, attempts to detect version from the transaction
- For `changeMasterCopy` transactions, extracts the target mastercopy address
- Looks up which version that mastercopy corresponds to
- Uses the detected version for safeTxHash calculation

## Key Changes

**Transaction Proposals:**
- `getVersionFromMastercopy()` - helper to detect Safe version from mastercopy address
- `detectVersionFromMigrationTransaction()` - extracts version from changeMasterCopy calls
- `getSafeTxHash()` - handles null version by detecting migration transactions

## Testing

- Added 5 comprehensive tests for null version migration scenarios
- All 987 tests passing ✅

## Example Flow

**Before:**
```
Safe with unofficial mastercopy (version: null)
  ↓
Propose changeMasterCopy transaction
  ↓
❌ Error: "Safe version is required"
```

**After:**
```
Safe with unofficial mastercopy (version: null)
  ↓
Propose changeMasterCopy to official singleton
  ↓
Detect target version from mastercopy address
  ↓
✅ Calculate safeTxHash using target version
  ↓
✅ Transaction proposal succeeds
```

## Related Issues

Fixes migration path for Safes with unsupported base contracts.

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/src/domain/common/utils/__tests__/safe.spec.ts
+++ b/src/domain/common/utils/__tests__/safe.spec.ts
@@ -730,7 +730,10 @@ describe('Safe', () => {
         .with('to', migrationContract)
         .with('value', '0')
         .with('operation', Operation.DELEGATE)
-        .with('data', migrateToL2Encoder().with('l2Singleton', l2Singleton).encode())
+        .with(
+          'data',
+          migrateToL2Encoder().with('l2Singleton', l2Singleton).encode(),
+        )
         .build();
 
       // Should not throw and should calculate hash using detected version
@@ -761,7 +764,10 @@ describe('Safe', () => {
         .with('to', unofficialMigrationContract)
         .with('value', '0')
         .with('operation', Operation.DELEGATE)
-        .with('data', migrateToL2Encoder().with('l2Singleton', l2Singleton).encode())
+        .with(
+          'data',
+          migrateToL2Encoder().with('l2Singleton', l2Singleton).encode(),
+        )
         .build();
 
       expect(() =>
@@ -786,9 +792,7 @@ describe('Safe', () => {
       }
 
       const migrationContract = faker.helpers.arrayElement(migrationContracts);
-      const unofficialL2Singleton = getAddress(
-        faker.finance.ethereumAddress(),
-      );
+      const unofficialL2Singleton = getAddress(faker.finance.ethereumAddress());
 
       const safe = safeBuilder()
         .with('address', safeAddress)
@@ -801,7 +805,9 @@ describe('Safe', () => {
         .with('operation', Operation.DELEGATE)
         .with(
           'data',
-          migrateToL2Encoder().with('l2Singleton', unofficialL2Singleton).encode(),
+          migrateToL2Encoder()
+            .with('l2Singleton', unofficialL2Singleton)
+            .encode(),
         )
         .build();
 
--- a/src/domain/common/utils/__tests__/safe.spec.ts
+++ b/src/domain/common/utils/__tests__/safe.spec.ts
@@ -14,6 +14,11 @@ import {
 import { typedDataBuilder } from '@/routes/messages/entities/__tests__/typed-data.builder';
 import type { BaseMultisigTransaction } from '@/domain/common/utils/safe';
 import type { IBuilder } from '@/__tests__/builder';
+import { migrateToL2Encoder } from '@/domain/contracts/__tests__/encoders/safe-to-l2-migration-encoder.builder';
+import {
+  getSafeL2SingletonDeployments,
+  getSafeToL2MigrationDeployments,
+} from '@/domain/common/utils/deployments';
 
 // Audited versions only
 const DOMAIN_WITHOUT_CHAIN_ID_VERSIONS = [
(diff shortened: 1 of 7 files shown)


katspaugh:
I've tested this locally, it works. `FF_TRUSTED_DELEGATE_CALL` must be set to true.

moisses89 (approved):
Looks good to me, just a minor comment.

src/domain/common/utils/safe.ts
```diff
@@ -111,6 +133,78 @@ export function getSafeTxHash(args: {
...
+  transaction: BaseMultisigTransaction;
+}): string | null {
+  // Delegate call to SafeToL2Migration.migrateToL2 (v1.3.0+)
+  return detectDelegateMigration(args);
+}
+
+/**
+ * Detects delegate call migrations to SafeToL2Migration.migrateToL2 (v1.3.0+)
+ */
+function detectDelegateMigration(args: {
```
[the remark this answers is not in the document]
katspaugh: I suggest we leave it like this because I already tested this version and renaming will require me to test it again.

src/domain/common/utils/safe.ts
```diff
@@ -111,6 +133,78 @@ export function getSafeTxHash(args: {
...
+  chainId: string;
+  safeAddress: Address;
+  transaction: BaseMultisigTransaction;
+}): string | null {
+  // Delegate call to SafeToL2Migration.migrateToL2 (v1.3.0+)
+  return detectDelegateMigration(args);
+}
+
+/**
+ * Detects delegate call migrations to SafeToL2Migration.migrateToL2 (v1.3.0+)
```
PooyaRaki, line 157: Nit: `@returns` and `@param` are missing

src/domain/common/utils/safe.ts
```diff
@@ -111,6 +133,78 @@ export function getSafeTxHash(args: {
...
+ */
+function detectDelegateMigration(args: {
+  chainId: string;
+  transaction: BaseMultisigTransaction;
+}): string | null {
+  try {
+    // Check if operation is DELEGATECALL
```
PooyaRaki, line 164: The next line is crystal clear, so we can remove the comment.

src/domain/common/utils/safe.ts
```diff
@@ -111,6 +133,78 @@ export function getSafeTxHash(args: {
...
+
+    const isOfficialMigrationContract = migrationContracts.some(
+      (addr) => addr.toLowerCase() === args.transaction.to.toLowerCase(),
+    );
+
+    if (!isOfficialMigrationContract) {
+      return null;
+    }
+
+    // Check if transaction has data
```
PooyaRaki, line 183: The comment is misleading. We are checking if transaction does not have data. Also the code is clear we can remove the comment.

src/domain/common/utils/__tests__/safe.spec.ts
```diff
@@ -681,4 +686,138 @@ describe('Safe', () => {
       },
     );
   });
+
+  describe('getSafeTxHash with null version (migration)', () => {
+    const chainId = '1'; // Ethereum Mainnet
```
PooyaRaki, line 691: No need for an extra comment

src/domain/common/utils/__tests__/safe.spec.ts
```diff
@@ -681,4 +686,138 @@ describe('Safe', () => {
...
+    it('should calculate safeTxHash for migrateToL2 delegate call when version is null', () => {
+      const targetVersion = '1.3.0';
+      const safeAddress = getAddress(faker.finance.ethereumAddress());
+      const migrationContracts = getSafeToL2MigrationDeployments({
+        chainId,
+        version: '1.4.1',
+      });
+
+      // Skip if no migration contract for this chain
+      if (migrationContracts.length === 0) {
```
PooyaRaki, line 715: Nit: Why can't we use `!migrationContracts.length`

src/domain/common/utils/safe.ts
```diff
@@ -111,6 +133,78 @@ export function getSafeTxHash(args: {
...
+    }
+
+    // Check if target is an official SafeToL2Migration contract
+    const migrationContracts = getSafeToL2MigrationDeployments({
+      chainId: args.chainId,
+      version: '1.4.1', // SafeToL2Migration was introduced in 1.4.1
+    });
+
+    const isOfficialMigrationContract = migrationContracts.some(
+      (addr) => addr.toLowerCase() === args.transaction.to.toLowerCase(),
```
PooyaRaki, line 176: Can we use `getAddress` from Viem to checksum the addresses?

src/domain/common/utils/deployments.ts
```diff
@@ -111,3 +125,38 @@ function formatDeployments(
...
+): string | null {
+  const versions = ['1.4.1', '1.3.0', '1.2.0', '1.1.1', '1.0.0'];
+
+  for (const version of versions) {
+    const l1Singletons = getSafeSingletonDeployments({ chainId, version });
+    const l2Singletons = getSafeL2SingletonDeployments({ chainId, version });
+
+    const found =
+      l1Singletons.some(
+        (addr) => addr.toLowerCase() === mastercopyAddress.toLowerCase(),
```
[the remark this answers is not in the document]
katspaugh: Because we just want to make sure they are equal and not internally consistent. Why waste CPU cycles?

src/domain/common/utils/safe.ts
katspaugh: There's no need to check the sum here, only to check if the two addresses are the same.

katspaugh:
@PooyaRaki I'm sorry but I'm not going to address all the nit-picks. It will take another day to apply these insignificant style changes and they test again. If we want to adhere to a certain style, it should be enforced by prettier and eslint, as well as guidelines for AI in AGENTS.md. Nit-picking like this will just slow us down.

src/domain/common/utils/deployments.ts
```diff
@@ -111,3 +125,38 @@ function formatDeployments(
...
+ * @param {string} chainId - the chain ID
+ * @param {Address} mastercopyAddress - the mastercopy address to check
+ *
+ * @returns {string | null} - the Safe version if found, null otherwise
+ */
+export function getVersionFromMastercopy(
+  chainId: string,
+  mastercopyAddress: Address,
+): string | null {
+  const versions = ['1.4.1', '1.3.0', '1.2.0', '1.1.1', '1.0.0'];
```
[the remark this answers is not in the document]
katspaugh: The list of versions can actually be taken from safe-deployments. I'll adjust this.

PooyaRaki:
@katspaugh 
> @PooyaRaki I'm sorry but I'm not going to address all the nit-picks. It will take another day to apply these insignificant style changes and then test everything again. If we want to adhere to a certain style, it should be enforced by prettier and eslint, as well as guidelines for AI in AGENTS.md. Nit-picking like this will just slow us down.

As always, I’ll leave nitpicks to your discretion to fix, but the non-nitpick issues need to be resolved before I can approve the PR.

src/domain/common/utils/safe.ts
katspaugh: Added.

src/domain/common/utils/deployments.ts
```diff
@@ -7,6 +7,8 @@ import {
   getSafeSingletonDeployments as _getSafeSingletonDeployments,
   getSafeToL2MigrationDeployments as _getSafeToL2MigrationDeployments,
 } from '@safe-global/safe-deployments';
+// eslint-disable-next-line no-restricted-imports
```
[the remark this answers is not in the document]
katspaugh: Removed, lint still passes. AI was hallucinating probably.

PR #2739, opened 2025-10-07
fix: update swagger type for operation - it can be only 0 and 1

compojoom:
## Summary
The swagger type outputs "operation:number", where in reality operation can only be 0 = call and 1 = delegate.

## Changes
Ads the necessary swagger apiProperty to specify the operation type

### DIFF
--- a/src/routes/data-decode/entities/data-decoded.entity.ts
+++ b/src/routes/data-decode/entities/data-decoded.entity.ts
@@ -23,7 +23,11 @@ class BaseDataDecoded implements DomainBaseDataDecoded {
 }
 
 class MultiSend implements DomainMultiSend {
-  @ApiProperty({ enum: Operation })
+  @ApiProperty({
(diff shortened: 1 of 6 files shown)


src/routes/data-decode/entities/data-decoded.entity.ts
```diff
@@ -23,7 +23,11 @@ class BaseDataDecoded implements DomainBaseDataDecoded {
 }
 
 class MultiSend implements DomainMultiSend {
-  @ApiProperty({ enum: Operation })
+  @ApiProperty({
+    enum: Operation,
+    enumName: 'Operation',
+    description: 'Operation type: 0 for CALL, 1 for DELEGATE',
```
PooyaRaki, line 29: Nit: I’d recommend removing the numbers from the description. If anything changes in the future or we add another operation, it’ll be easier to maintain without risking an overlooked update.

PR #2751, opened 2025-10-21
fix: swagger type wrong for chainName and data

compojoom:
## Summary
- fix the type for the data prop on the data-decoded object. Swagger was returning this as object, but the more appropriate type here is string, as we are dealing with hex strings
- chainName was typed as Number in swagger, but was actually String in the code


src/routes/chains/entities/balances-provider.entity.ts
```diff
@@ -2,7 +2,7 @@ import { BalancesProvider as DomainBalancesProvider } from '@/domain/chains/enti
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 
 export class BalancesProvider implements DomainBalancesProvider {
-  @ApiPropertyOptional({ type: Number, nullable: true })
+  @ApiPropertyOptional({ type: String, nullable: true })
```
PooyaRaki: Shall we change the decorator to `@ApiProperty` since the prop is not optional but `nullable`.

PR #2757, opened 2025-10-24
fix: multiple wrong swagger types

compojoom:
## Summary
- fix: dataDecodedParameter value can only be string
looking through the codebase we assume that value is either string or an Array of Strings. Since we didn’t type the swagger ApiProperty we ended up with “object” type which is not correct.
- fix: executedFeeToken is from type TokenInfo
- fix: propose endpoint returns TransactionDetails and not Transaction

### DIFF
--- a/src/routes/transactions/transactions.controller.ts
+++ b/src/routes/transactions/transactions.controller.ts
@@ -766,7 +766,7 @@ export class TransactionsController {
       'Transaction proposal including recipient, value, data, and initial signature',
   })
   @ApiOkResponse({
-    type: Transaction,
+    type: TransactionDetails,
     description: 'Transaction proposed successfully',
   })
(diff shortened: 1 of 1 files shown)


src/routes/transactions/entities/swaps/twap-order-info.entity.ts
```diff
@@ -162,7 +162,6 @@ export class TwapOrderTransactionInfo
   executedFee: string | null;
 
   @ApiProperty({
```
PooyaRaki: Does Swagger infer types from TS definitions? I would expect it to infer an empty object rather than the full `TokenInfo` object.

katspaugh (approved):
👍

PooyaRaki (approved):

PR #2758, opened 2025-10-24
fix(safe-shield): analyze interactions with only executed txs

LucieFaire:
## Changes
- Check contract interactions in counterparty analysis only with executed transactions, otherwise the current transaction counts as well.


tmjssz (approved):
Nice catch!

PR #2768, opened 2025-10-28
fix: change description for failed threat analysis

LucieFaire:
## Summary

Change the description message for failed threat analysis.


LucieFaire:
I have bypassed the review as it is a one-liner change with no effect on logic in it.

PR #2769, opened 2025-10-29
fix: Caching relay count failing due to non-int ttlSeconds value

akshay-ap:
## Summary

The incremented value of relay count is not being stored in cache due to `ttlSeconds` being a non-integer value.
The PR fixes the issue by using `Math.floor` function so that `ttlSeconds` is an integer.

### DIFF
--- a/src/domain/relay/relayers/no-fee-campaign.relayer.ts
+++ b/src/domain/relay/relayers/no-fee-campaign.relayer.ts
@@ -169,9 +169,10 @@ export class NoFeeCampaignRelayer implements IRelayer {
       const currentCount = await this.getRelayCount(args);
       const incremented = currentCount + 1;
 
-      const ttlSeconds =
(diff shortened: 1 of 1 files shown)


src/domain/relay/relayers/no-fee-campaign.relayer.ts
```diff
@@ -169,9 +169,10 @@ export class NoFeeCampaignRelayer implements IRelayer {
       const currentCount = await this.getRelayCount(args);
       const incremented = currentCount + 1;
 
-      const ttlSeconds =
+      const ttlSeconds = Math.floor(
         this.relayConfiguration[parseInt(args.chainId)].endsAtTimeStamp -
```
PooyaRaki: QQ: Is `endsTimeStamp` in seconds or milliseconds?
If it's in milliseconds, you're subtracting a seconds from a milliseconds value.

PR #2796, opened 2025-11-07
feat(safe-shield): extend contract analysis to include name/logoUrl

LucieFaire:
## Summary [COR-817](https://linear.app/safe-global/issue/COR-817/backend-extend-the-contract-analysis-response)

Extend contract analysis response to include name and logoUrl of the contract (when available)

## Changes
- Added new optional fields: `name` and `logoUrl` to the `contracts` part of the counterparty analysis response.
- Example response:
```
{
  "recipient": {},
  "contract": {
    "0x123": {
      "CONTRACT_VERIFICATION": [
        {
          "severity": "OK",
          "type": "VERIFIED",
          "title": "Verified contract",
          "description": "This contract is verified as \"BLA BLA\"."
        }
      ],
      "name": "BAL BLA",
      "logoUrl": "https://example.logo.comÏ",
      "CONTRACT_INTERACTION": [
        {
          "severity": "OK",
          "type": "KNOWN_CONTRACT",
          "title": "Known contract",
          "description": "You have already interacted with this contract."
        }
      ]
    }
  }
}
```

### DIFF
--- a/src/modules/safe-shield/entities/__tests__/analysis-responses.entity.spec.ts
+++ b/src/modules/safe-shield/entities/__tests__/analysis-responses.entity.spec.ts
@@ -35,14 +35,9 @@ describe('Analysis Response Schemas', () => {
       });
 
       it('should validate response with multiple addresses', () => {
-        const multiAddressResponse = recipientAnalysisResponseBuilder()
-          .with(getAddress(faker.finance.ethereumAddress()), {
-            isSafe: true,
-            RECIPIENT_INTERACTION: [recipientAnalysisResultBuilder().build()],
-            RECIPIENT_ACTIVITY: [recipientAnalysisResultBuilder().build()],
-            BRIDGE: [recipientAnalysisResultBuilder().build()],
-          })
-          .build();
+        const responseA = recipientAnalysisResponseBuilder().build();
+        const responseB = recipientAnalysisResponseBuilder().build();
+        const multiAddressResponse = { ...responseA, ...responseB };
 
         const result =
           RecipientAnalysisResponseSchema.safeParse(multiAddressResponse);
@@ -124,15 +119,9 @@ describe('Analysis Response Schemas', () => {
       });
 
       it('should validate response with multiple addresses', () => {
-        const multiAddressResponse = contractAnalysisResponseBuilder()
-          .with(getAddress(faker.finance.ethereumAddress()), {
-            logoUrl: faker.image.url(),
-            name: faker.company.name(),
-            CONTRACT_VERIFICATION: [contractAnalysisResultBuilder().build()],
-            CONTRACT_INTERACTION: [contractAnalysisResultBuilder().build()],
-            DELEGATECALL: [contractAnalysisResultBuilder().build()],
-          })
-          .build();
+        const responseA = contractAnalysisResponseBuilder().build();
+        const responseB = contractAnalysisResponseBuilder().build();
+        const multiAddressResponse = { ...responseA, ...responseB };
 
         const result =
           ContractAnalysisResponseSchema.safeParse(multiAddressResponse);
--- a/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
+++ b/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
@@ -382,6 +382,7 @@ describe('ContractAnalysisService', () => {
               modified: faker.date.recent(),
             })
             .with('displayName', name)
+            .with('logoUrl', undefined)
             .build(),
         ])
         .build();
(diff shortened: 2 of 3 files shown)


src/modules/safe-shield/contract-analysis/contract-analysis.service.ts
```diff
@@ -158,49 +154,59 @@ export class ContractAnalysisService {
...
+        return { ...withDelegate(undefined) };
       }
 
-      resolvedContract = results[0];
-      const { abi, displayName, name: rawName } = resolvedContract;
-      const type = abi ? 'VERIFIED' : 'NOT_VERIFIED';
-      const name = displayName || rawName;
+      const resolved = results[0];
+      const name = resolved.displayName || resolved.name || undefined;
+      const logoUrl = resolved.logoUrl || undefined;
```
PooyaRaki: If we made default value for logoUrl `undefined` then here we could:

```ts
      const logoUrl = resolved?.logoUrl;
```

The same applies to name.

src/modules/safe-shield/entities/__tests__/analysis-responses.entity.spec.ts
```diff
@@ -126,6 +126,8 @@ describe('Analysis Response Schemas', () => {
       it('should validate response with multiple addresses', () => {
         const multiAddressResponse = contractAnalysisResponseBuilder()
           .with(getAddress(faker.finance.ethereumAddress()), {
+            logoUrl: faker.image.url(),
```
PooyaRaki: Instead of duplicating the whole object here we could create another `contractAnalysisResponseBuilder` object and merge them to have multiple responses. We should do the same for other test cases.

src/modules/safe-shield/entities/dtos/counterparty-analysis.dto.ts
```diff
@@ -43,6 +43,18 @@ export class ContractAnalysisResultDto extends AnalysisResultDto<
 export class ContractAnalysisDto
   implements GroupedAnalysisResults<ContractAnalysisResult>
 {
+  @ApiPropertyOptional({
+    description: 'Logo URL for the contract',
+    example: 'https://example.com/logo.png',
+  })
+  logoUrl?: string;
```
PooyaRaki: Nit: Access modifier is should be defined e.g. `public readonly`

src/modules/safe-shield/entities/__tests__/analysis-responses.entity.spec.ts
```diff
@@ -178,9 +182,28 @@ describe('Analysis Response Schemas', () => {
...
+
+        expect(result.success && result.data).toStrictEqual(
+          responseWithoutMetadata,
+        );
+      });
+
       it('should reject invalid status group', () => {
         const invalidStatusGroupResponse = contractAnalysisResponseBuilder()
           .with(getAddress(faker.finance.ethereumAddress()), {
+            logoUrl: faker.image.url(),
```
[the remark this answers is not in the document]
clovisdasilvaneto: the test case `should reject invalid status group`. It has nothing to do with validation of logoUrl and name 🤔

src/modules/safe-shield/entities/__tests__/analysis-responses.entity.spec.ts
clovisdasilvaneto: This way is easier to read and simpler IMO.

clovisdasilvaneto:
@PooyaRaki  helper was moved, about the point you had regarding why do we have the `checkDelegateCall` and `isDelegateCall` I would rather address it with @LucieFaire when she is back in another PR instead of blocking the PR only because of it, since the FE is depending on this and this topic is not critical IMO.

src/modules/safe-shield/entities/__tests__/analysis-responses.entity.spec.ts
PooyaRaki: Well, that's mostly subjective but since we heavily use builders and as per DRY principles, we should have reusable components as much as possible.

src/modules/safe-shield/contract-analysis/contract-analysis.service.ts
```diff
@@ -257,6 +264,17 @@ export class ContractAnalysisService {
     return this.mapToAnalysisResult({ type: 'UNEXPECTED_DELEGATECALL' });
   }
 
+  /**
+   * Helper to include delegatecall result if applicable.
+   */
+  private withDelegate(
+    isDelegateCall: boolean,
+    resolved?: Contract,
```
PooyaRaki: The name of the parameter should be `contract`

PR #2800, opened 2025-11-13
feat(portfolio): fix trusted parameter on tokens from zerion

vseehausen:
## Summary

Fixes a "bug", where some tokens are not verified, but should not be marked as spam.

### DIFF
--- a/src/modules/portfolio/datasources/zerion-portfolio-api.service.ts
+++ b/src/modules/portfolio/datasources/zerion-portfolio-api.service.ts
@@ -218,7 +218,7 @@ export class ZerionPortfolioApi implements IPortfolioApi {
               position.attributes.name,
             logoUri: position.attributes.fungible_info.icon?.url ?? '',
             chainId,
-            trusted: position.attributes.fungible_info.flags?.verified ?? false,
+            trusted: !(position.attributes.flags.is_trash ?? false),
             type:
               address === null ||
               address === '0x0000000000000000000000000000000000000000'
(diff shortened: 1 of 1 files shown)


src/modules/portfolio/datasources/zerion-portfolio-api.service.ts
```diff
@@ -218,7 +218,7 @@ export class ZerionPortfolioApi implements IPortfolioApi {
               position.attributes.name,
             logoUri: position.attributes.fungible_info.icon?.url ?? '',
             chainId,
-            trusted: position.attributes.fungible_info.flags?.verified ?? false,
+            trusted: !(position.attributes.flags.is_trash ?? false),
```
vseehausen: Tokens should only be marked as spam, when they have the `is_trash` attribute set to true.

src/modules/portfolio/datasources/zerion-portfolio-api.service.ts
```diff
@@ -363,7 +363,7 @@ export class ZerionPortfolioApi implements IPortfolioApi {
             name: position.attributes.fungible_info.name ?? '',
             logoUri: position.attributes.fungible_info.icon?.url ?? '',
             chainId,
-            trusted: position.attributes.fungible_info.flags?.verified ?? false,
+            trusted: !(position.attributes.flags.is_trash ?? false),
```
vseehausen: This is the token info for a position, which inherits spam/trash attribute from the position.

PR #2926, opened 2026-02-16
Refactor/performanceImprovements

PooyaRaki:
## Summary

Performance improvements across several hot paths: Redis pipeline batching, parallelized async calls, in-memory key derivation cache, single-pass chain parsing, and batched SQL operations.

### Redis pipelining (`redis.cache.service.ts`)
- Replaced sequential `hSet` + `expire` and `unlink` + `hSet` + `expire` calls with `multi().exec()` pipelines to reduce round-trips
- `deleteByKey` validates all three pipeline results (`unlink`, `hSet`, `expire`) and **logs errors** if the invalidation marker fails (does not throw — cache invalidation is best-effort)
- `deleteByKey` uses `.expire()` without `NX` so repeated invalidations refresh the TTL instead of being a no-op
- Cleanup `unlink` in `hSet` error path now logs a warning instead of silently swallowing errors

### Subscription batching (`notifications.repository.ts`)
- Consolidated `deletePreviousSubscriptions` into a single SQL `DELETE ... OR ...` instead of one query per safe
- Batched `removeGetSubscribersBySafeCache` and `deleteAllSubscriptions` cache invalidations into single calls
- Added `.max(100)` to `safes` array in `UpsertSubscriptionsDto` to prevent unbounded SQL query generation

### Parallelized hot endpoints
- **`getMultisigTransactions`** (`transactions.service.ts`): `getSafe()` and `prefetchAddressInfos()` now run concurrently via `Promise.all`. `prefetchAddressInfos` is wrapped in `.catch()` so cache-warming failure does not abort the request
- **`getSafeOverview`** (`safes.service.ts`): `getSafe()` and `getBalances()` start concurrently after chain resolves; `getTransactionQueue()` starts as soon as `getSafe()` resolves, overlapping with any remaining `getBalances()` time. Added `balancesPromise.catch(() => {})` to prevent unhandled rejection when `getSafe` rejects before `balancesPromise` is awaited

### Derived key cache (`encryption.ts`)
- Added a bounded in-memory LRU cache (max 50 entries) for `scryptSync` results to avoid repeated CPU-intensive derivation for the same `(key, salt)` pair
- True LRU eviction: entries are promoted on cache hit via delete-and-reinsert
- Cache key uses length-prefix to prevent collisions
- `MAX_DERIVED_KEY_CACHE_SIZE` is defined in `encryption.ts` (not in `datasources/`) to respect the domain→datasources layer boundary
- Exported `clearDerivedKeyCache()` for testing and key rotation

### Portfolio deduplication (`zerion-portfolio-api.service.ts`)
- Pre-computed a `Map<networkName, chainId>` to eliminate redundant cache/API lookups for duplicate networks when building token balances and app positions
- Extracted `_resolvePositionMeta` to deduplicate token resolution logic shared between `_buildTokenBalances` and `_buildAppPositions`
- Consolidated `_calculatePositionsBalance` and `_calculateTotalBalance` into a single method

### Chain data optimization (`chains.repository.ts`)
- **Single-pass chain parsing**: Replaced the two-pass flow (`LenientBasePageSchema.parse` then `ChainSchema.parse`) with a single pass using `ChainSchema.safeParse` per item via shared `parseChainPage` helper
- **Parallel chain fetching**: `getAllChains()` fetches the first page, uses its `count` to calculate remaining pages, then fetches them all in parallel via `Promise.all` with arithmetic offsets (multiples of `MAX_LIMIT`)

### Documentation (`AGENTS.md`, `CLAUDE.md`)
- Added `CLAUDE.md` referencing `AGENTS.md` for Claude Code auto-loading
- Added `yarn env:validate` to the pre-commit checklist (stop if it fails)
- Added batch commit guidelines for grouping related changes into logical commits
- Reinforced license header requirement on modified files

### Test updates
- Updated integration tests for `getSafeOverview` to match new parallelized network call order
- Set explicit `count` on `chainsPage1` in owners v2/v3 integration tests for `getAllChains` parallel fetch with offset-based pagination
- Updated `NotificationsRepositoryV2` tests to validate batched deletion logic
- Updated `RedisCacheService` tests to mock and validate `multi()` pipeline commands
- Added tests for derived key caching with `afterEach` cache cleanup
- Added SPDX license identifiers to modified source files

## Files changed (17)

| File | Changes |
|------|---------|
| `AGENTS.md` | Pre-commit checklist additions, batch commit guidelines |
| `CLAUDE.md` | New — references AGENTS.md |
| `src/datasources/cache/constants.ts` | SPDX license header |
| `src/datasources/cache/redis.cache.service.ts` | Pipeline batching, result validation, cleanup logging |
| `src/datasources/cache/redis.cache.service.key-prefix.spec.ts` | Updated tests for pipeline mocks |
| `src/domain/common/utils/encryption.ts` | LRU derived key cache, clearDerivedKeyCache export |
| `src/domain/common/utils/__tests__/encryption.spec.ts` | Cache tests with afterEach cleanup |
| `src/modules/chains/domain/chains.repository.ts` | Single-pass parsing + parallel fetch |
| `src/modules/notifications/domain/v2/entities/upsert-subscriptions.dto.entity.ts` | `.max(100)` on safes array |
| `src/modules/notifications/domain/v2/notifications.repository.ts` | Batched deletion |
| `src/modules/notifications/domain/v2/notifications.repository.spec.ts` | Updated tests |
| `src/modules/portfolio/datasources/zerion-portfolio-api.service.ts` | Network map dedup, _resolvePositionMeta extraction |
| `src/modules/safe/routes/safes.service.ts` | Parallel getSafeOverview |
| `src/modules/safe/routes/safes.controller.overview.integration.spec.ts` | Updated integration tests |
| `src/modules/transactions/routes/transactions.service.ts` | Parallel getSafe + prefetch with catch |
| `src/modules/owners/routes/owners.controller.v2.integration.spec.ts` | Offset-based pagination mocks |
| `src/modules/owners/routes/owners.controller.v3.integration.spec.ts` | Offset-based pagination mocks |

## Test plan

- [x] All existing tests pass with updated assertions
- [x] New tests added for derived key caching with cache cleanup
- [x] Integration tests updated for parallelized call ordering
- [x] Type check passes (0 new errors)
- [ ] Manual smoke test on staging for overview and transaction endpoints

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -42,7 +42,15 @@ Before creating **EACH** commit, you MUST run the following commands in sequence
 
    If there are any remaining lint errors, fix them manually before proceeding.
 
-3. **Run tests:**
+3. **Validate environment configuration:**
+
+   ```bash
+   yarn env:validate
+   ```
+
+   **Stop immediately if this fails.** It means the configuration schema is broken and must be fixed before proceeding.
+
+4. **Run tests:**
 
    ```bash
    yarn test
@@ -59,12 +67,23 @@ The correct workflow for making commits is:
 # 2. Run quality checks
 yarn format
 yarn lint --fix
+yarn env:validate  # STOP if this fails
 yarn test
 
 # 3. Only after all checks pass, commit
 git add <files>
 git commit -m "Your commit message"
 ```
+
+### Batch Related Changes
+
+When committing multiple files, **group related changes into logical commits** rather than committing all files at once or one file at a time. Each commit should represent a single coherent change:
+
+- Files that fix the same issue or implement the same feature go together
+- A source file and its corresponding test file go in the same commit
+- Unrelated changes get separate commits even if made at the same time
+
+For example, if you fix a bug in `encryption.ts`, update its test, and also fix an unrelated issue in `redis.cache.service.ts`, make two commits — not one and not three.
 
 ## License Headers
 
@@ -81,7 +100,7 @@ For `.md` files the hook uses a multi-line HTML comment style (`<!--| | -->`), s
 
 This is enforced by a `pre-commit` hook (`insert-license` from `Lucas-C/pre-commit-hooks`) and a `license-headers` CI job. The canonical header text lives in `LICENSE_HEADER.txt`.
 
-Every file you create or modify in a PR **MUST** have the correct license header. Add it yourself — do not rely on the pre-commit hook to do it for you. If the pre-commit hook adds it automatically, include that change in your commit.
+Every file you create or modify in a PR **MUST** have the correct license header. **Always add the license header yourself to every modified file** — do not rely on the pre-commit hook to do it for you. If the pre-commit hook adds it automatically, include that change in your commit.
 
 ## Important Notes
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,4 +1,5 @@
 <!--
   SPDX-License-Identifier: FSL-1.1-MIT
  -->
+
 Follow all instructions in AGENTS.md strictly.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,4 @@
+<!--
+  SPDX-License-Identifier: FSL-1.1-MIT
+ -->
+Follow all instructions in AGENTS.md strictly.
--- a/src/datasources/cache/constants.ts
+++ b/src/datasources/cache/constants.ts
@@ -18,11 +18,3 @@ export const CacheKeyPrefix = Symbol('CacheKeyPrefix');
  * Ref: https://github.com/redis/redis/blob/cc244370a2d622d5d2fec5573ae87de6c59ed3b9/src/expire.c#L573
  */
 export const MAX_TTL = Number.MAX_SAFE_INTEGER - 1;
-
-/**
- * Maximum number of entries in the in-memory derived-key cache used by the
- * encryption utilities. Oldest entries are evicted (FIFO) once this limit
- * is reached, bounding memory usage while avoiding repeated scrypt calls for
- * frequently used (key, salt) pairs.
- */
-export const MAX_DERIVED_KEY_CACHE_SIZE = 50;
--- a/src/datasources/cache/constants.ts
+++ b/src/datasources/cache/constants.ts
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
 /**
  * The {@link CacheKeyPrefix} is meant to be used whenever a prefix should be added to the keys
  * registered in the cache.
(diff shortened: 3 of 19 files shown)


PooyaRaki:
@claude review

vseehausen (approved):

LucieFaire (commented):
It would make it easier if in the future split these into multiple PRs, as there are a lot of changes that are not related to each other, and after merge they will be squashed into 1 big change

PooyaRaki:
Investigated all three items:

#### 1. `deleteByKey` doesn't handle `exec()` rejection — Already handled

`deleteByKey` already has a `try/catch` around `exec()` (lines 108-123). If `exec()` throws, it logs the error and returns 0. The concern about "unlink succeeded but hSet/expire failed at transport level" doesn't apply — Redis `MULTI/EXEC` sends the pipeline as an atomic batch over a single round-trip. If the transport fails, `exec()` throws and none of the individual results are available; there's no partial execution at the transport level. Individual command failures within a *successful* `exec()` are returned as error objects in the results array, which lines 125-137 already handle.

#### 2. `getSafe` called twice in `getSafeOverview` — False premise

`getTransactionQueue` does **not** call `getSafe()`. It accepts a `safe: Safe` object directly and uses `safe.address`/`safe.nonce` from it (see `safe.repository.ts:261-279`). The second network call to `/api/v1/safes/{address}` visible in the integration test comes from `isSafe()` in `BalancesApiManager.getApi()` (`balances-api.manager.ts:69`), which checks whether to use `SafeBalancesApi` vs `ZerionBalancesApi` when `counterfactualBalances` is enabled. This uses a separate cache key (`getIsSafeCacheDir` vs `getSafeCacheDir`) and is a pre-existing pattern on `main`, not introduced by this branch.

#### 3. Missing expire failure test — Already exists

The test at line 162-174 in `redis.cache.service.key-prefix.spec.ts` already does exactly this: `exec` returns `[1, 1, 'unexpected']` and verifies the error log for the expire check branch.

PooyaRaki:
@claude review

PooyaRaki:
The observation is technically accurate but the suggested fix (throwing or returning an error) would be counterproductive. Here's why log-and-continue is the correct behavior:

**What happens when the marker is missing:** The cache key was just deleted by `unlink`. The next `CacheFirstDataSource.get()` finds no cached value, makes a fresh network call, and `_shouldBeCached()` returns `true` (condition 1: "no invalidation marker found") — so the **fresh** data gets cached. This is the correct outcome, not stale data.

**The actual risk is narrow:** The marker only guards against a race where a network request was already in-flight *before* `deleteByKey`, and that in-flight response tries to write stale data *after* the delete. Without the marker, that stale write succeeds. This is a brief timing window, and the previous implementation on `main` had the same exposure — `hSet` for the marker could fail independently there too.

**Why throwing would be worse:** The delete *did* succeed. Callers (webhook handlers, cache invalidation flows) would incorrectly treat it as a failure, potentially retrying or propagating errors for an operation that accomplished its primary goal. The invalidation marker is a best-effort race-condition guard, not a correctness requirement.

LucieFaire (approved):

LucieFaire (approved):

PR #3018, opened 2026-04-09
feat: migrate queue operations from TX service to queue service

PooyaRaki:
## Summary

  Migrate queue-related Safe operations — transaction proposals/confirmations/deletions, queued-transaction reads, messages, and delegates — from the Safe Transaction Service to the new **Safe Queue Service**, gated behind the `FF_QUEUE_SERVICE` feature flag. When the flag is
  disabled, behaviour is unchanged: every queue path falls back to the existing tx-service routing.

  Alongside the migration, message Safe App identity is now resolved by **origin URL** instead of a client-supplied `safeAppId`.

  ## Changes

  ### Queue service client (new `src/modules/queue/`)
  - `IQueue` interface + `QueueService` implementation covering propose / confirm / delete, queued-transaction reads, batch multisig fetch, messages (get-by-hash, list-by-safe, create, add-signature), and delegates
  - Zod entity schemas (multisig transaction, message, delegate, origin) and tx/message mappers
  - `parseOrigin` / `buildOrigin` helpers to translate between the tx-service origin JSON and the queue's `originName` / `originUrl` fields
  - Namespaced cache keys (`queue_…`) to avoid collisions with tx-service cache entries, plus a dedicated circuit-breaker key wired on **every** queue call
  - `getMultisigTransactionsBatch` chunks `safe_tx_hash` queries (50/call) to stay under proxy/WAF URL limits

  ### Auth & transport
  - `getQueueAuthHeaders` injects a Bearer API key against the public Queue Service in development; `USE_QUEUE_SERVICE_VPC_URL` switches to VPC-internal routing
  - `NetworkService` gains a `PATCH` method (used by delegate updates)

  ### Feature flag & routing
  - `FF_QUEUE_SERVICE` (read as `features.queueService`) gates all routing inline in the repositories
  - **Queued/pending reads** — `SafeRepository._getTransactionQueue` routes the queue list to the queue service
  - **Executed-tx origin enrichment** — executed transactions are still read from the tx-service; origin metadata is enriched from the queue via `bindQueueOrigins` (batch) / `bindQueueOrigin` (single), with failures logged and swallowed for graceful degradation
  - **`getMultiSigTransaction` is queue-first** when enabled: the queue is authoritative for not-yet-executed transactions; for executed ones the tx-service is the source of truth with the queue origin overlaid
  - **Transaction writes** — propose / confirm / delete routed to the queue service
  - **Delegates** — new delegates **v3** module (list / create / **update** / delete with body-based payloads); v2 repository routes to the queue service
  - **Messages** — `MessagesRepository` routes get-by-hash, list-by-safe, create, and add-signature to the queue service, filtering out any cross-chain records the queue returns

  ### SafeApp resolution by origin URL
  - Message domain entity no longer carries `safeAppId`
  - `SafeAppInfoMapper` moved to `src/modules/safe-apps/mappers/`; the `SafeAppInfo` route entity now exposes `id`
  - `MessageMapper` resolves Safe App identity by looking up the origin URL through `SafeAppsRepository`
  - `CreateMessageDto.safeAppId` is marked `@deprecated` — still accepted for backward compatibility but ignored

  ### Code-review hardening
  - Origin overlay is applied **only when the queue actually has origin metadata**, so a not-yet-backfilled queue entry no longer clobbers an existing tx-service origin with `null`
  - Queue ordering is passed as a typed `nonceOrder: 'asc' | 'desc'` (replacing a fragile substring check); the queue can only sort by nonce, so `getTransactionQueueByModified`'s `-modified` ordering degrades to nonce-desc — documented at the call site, and the cache field keys on
  `nonceOrder`
  - `getMultisigTransactionsBatch` uses `Promise.allSettled`, so one failing chunk no longer discards the whole batch (partial enrichment preserved)
  - `QueueService.proposeTransaction` returns the response body, consistent with the other queue writes
  - `getMessagesBySafe` adjusts the page `count` when cross-chain messages are filtered out
  - `parseOrigin` guards against valid-but-non-object JSON

  ### Configuration
  New environment variables:
  - `FF_QUEUE_SERVICE` — enables queue service routing (default `false`)
  - `QUEUE_SERVICE_BASE_URI` — public queue service base URL
  - `QUEUE_SERVICE_API_KEY` — Bearer token for the public queue service in development
  - `USE_QUEUE_SERVICE_VPC_URL` — switch to VPC-internal routing (skips API-key auth)


PooyaRaki:
@claude review

PooyaRaki:
@claude review

PooyaRaki:
Thanks for the second pass. Went through each item — addressed two with clarifying comments and pushed back on the rest with verification. Details below.
  
  ✅ #3 — getMessagesBySafe count is per-page — Agreed it's best-effort. Added a comment making that explicit (only corrects rows filtered on the current page; can still over-report if count is an all-pages total; cross-chain results should be rare since the queue is queried with the
  chain id). Logic unchanged.

  ✅ #7 — label ?? '' — This is intentional parity, not a silent coercion. The domain DelegateSchema.label is z.string(), and the existing tx-service v2 path (DelegatePageSchema.parse) already produces '' for unlabeled delegates. Coercing the queue's nullable label to '' makes both
  backends represent "no label" identically for the label= filter and consumers. Added a comment to document it.

  ❌ #2 — floating promises / void — This fix would break the build here. The repo lints with biome, not typescript-eslint: noVoid is "error" (prefixing with void is a lint error) while noFloatingPromises is "warn". Running biome lint on both files reports no warnings, because the
  attached .catch() rejection handler already satisfies the rule. So the current code is correct and adding void would regress it. Left as-is.

  ❌ #8 — parallelize getSafe in the non-executed path — There's a real data dependency: getSafe is called with address: queue.safe, and the safe address only comes from the queue response (the caller passes safeTransactionHash, not an address). So it can't be issued in parallel with
  the queue call.

  ❌ safeAppId nit (! → ?) — Doesn't compile. CreateMessageDto implements z.infer<typeof CreateMessageDtoSchema>, and .default(null) makes the schema's output type number | null (required). Making the property optional fails the implements check: TS2416: 'number | null | undefined' 
  is not assignable to 'number | null'. The .nullish().default(null) shape applies to the input type, not the z.infer output the class implements.

  ⚠️  #1 — prod auth with useVpcUrl=false — This is a deliberate, explicitly-tested design. Worth noting the failure mode is bounded: a prod misconfig (USE_QUEUE_SERVICE_VPC_URL unset) hits the public endpoint unauthenticated and is rejected with a 401 — it fails loudly, it doesn't
  silently leak. I'd rather not flip tested security behavior in this PR. If we want a guard, I'd go with your option (a): a startup assertion that in production useVpcUrl === true || apiKey != null. Happy to add it as a follow-up if we agree on that shape.

  ⚠️  #4 / #6 — executed-item eventual-consistency race — Not a regression: pre-migration getMultiSigTransaction always read the tx-service and 404'd until it caught up — same behavior. The suggested fallback (mapQueueToMultisigTransaction on tx-service failure) would return a tx
  flagged isExecuted: true but with executionDate/blockNumber/executor/fee all null, which is arguably worse for clients than a brief retryable error. Keeping "tx-service is source of truth for executed txs" as a conscious tradeoff rather than returning partial execution metadata.

  ⚠️  #5 — propose verifies against tx-service — Agreed it's transitional. Called out in the PR description; will track removal of the tx-service verifier lookup as a follow-up once the tx-service is decommissioned.

  Pushed in 3cbbf67dc (the #3 and #7 comments)

PooyaRaki:
@claude review again, there are new changes as well

PooyaRaki:
Bug 1 — note stripped from executed-tx origin overlay
  
  Fixed. All three overlay sites (bindQueueOrigins, bindQueueOrigin, and the executed branch of resolveMultisigTransaction) now pass queue.notes through to buildOrigin, so the note survives once a transaction executes — consistent with the queued-tx mapper path. Added a
  safe.repository.spec.ts case asserting the note is preserved on the executed overlay (the queue builder defaults txHash non-null, so it exercises the executed branch).

  Bug 2 — clearMultisigTransactions skipped queue clear on tx-service failure

  Fixed. Both invalidations now run independently via Promise.allSettled, so a tx-service hiccup no longer blocks the queue clear. If either rejects we throw a generic Error('Failed to clear multisig transactions') — no chain/safe/error detail leaked — which keeps webhook retries
  working.

  Nit — origin.helper.ts:71-75 note === null ternary

  Leaving this as-is. The premise that an undefined note falls into the "include note" branch doesn't hold: the parameter has a default of = null, and JS coerces an undefined argument to that default, so inside the function note is only ever string | null. buildOrigin(a, b, 
  undefined) makes note === null true and hits the no-note branch — the {"note":undefined} emission can't occur, and the current check is already correct.

  The other nits (mapper:38, queue.service.ts:94, the lost early return) were noted as already-clean — no change needed.

PooyaRaki:
@claude review

PooyaRaki:
@claude review again,

1. clearMultisigTransaction (singular) & clearAllExecutedTransactions retain the original fragility
  
  Fixed in 456cbb71e. Mirrored the Promise.allSettled + generic-error shape from the plural method to both siblings — the tx-service and queue clears now run independently, and a rejection in either surfaces a generic error so the webhook retries the whole invalidation. Messages are
  method-specific ('Failed to clear all executed transactions' / 'Failed to clear multisig transaction').

  2. getMultiSigTransactionWithNoCache fetches the Safe twice on the non-executed path

  Acknowledged, leaving as-is. The duplicate is real, but getSafe is cached so the second call is a cache hit, and the fix would mean threading the resolved Safe out of resolveMultisigTransaction — which getMultiSigTransaction also depends on — changing a shared internal contract to
  save a cache-hit round-trip. The coupling cost outweighs the benefit on this path. Happy to revisit if profiling shows it matters.

  3. Note-preservation test only covered the queue-first executed branch

  Fixed in 456cbb71e. Added a bindQueueOrigins (transaction-history batch) test asserting the note survives the overlay — the production-hot path you flagged — and extended the buildQueueEntityForSafe helper with an optional notes field.

  Verification: tsc --noEmit clean; safe.repository.spec.ts 45/45.

PooyaRaki:
@claude review

src/datasources/network/fetch.network.service.ts
```diff
@@ -107,6 +107,41 @@ export class FetchNetworkService implements INetworkService {
     });
   }
 
+  async patch<T>(args: {
+    url: string;
+    data?: object;
+    networkRequest?: NetworkRequest;
+  }): Promise<NetworkResponse<T>> {
+    const url = this.buildUrl(args.url, args.networkRequest?.params);
+    this.logRequest(url, 'PATCH');
```
compojoom, line 116: why are we logging here? I don't see us logging in the other methods.

PooyaRaki: We do log in each method:
https://github.com/safe-global/safe-client-gateway/blob/d606f6d82be1ed6dbd5443a8b3bb5ef850e9fb1f/src/datasources/network/fetch.network.service.ts#L36
https://github.com/safe-global/safe-client-gateway/blob/d606f6d82be1ed6dbd5443a8b3bb5ef850e9fb1f/src/datasources/network/fetch.network.service.ts#L62
https://github.com/safe-global/safe-client-gateway/blob/d606f6d82be1ed6dbd5443a8b3bb5ef850e9fb1f/src/datasources/network/fetch.network.service.ts#L151

src/modules/delegate/domain/v3/delegates.v3.repository.ts
```diff
@@ -0,0 +1,175 @@
...
+
+  async updateDelegate(args: {
+    chainId: string;
+    safeAddress: Address | null;
+    delegate: Address;
+    delegator: Address;
+    signature: string;
+    label: string;
+  }): Promise<void> {
+    if (!this.queueServiceEnabled) {
```
compojoom: nothing to change, but for the protocol I think that `if(this.queuServiceEnabled)` is way easier to read than the negation form.

src/__tests__/testing-module.ts
```diff
@@ -87,6 +89,10 @@ export function createTestModule(
         originalModule: TxAuthNetworkModule,
         testModule: TestTxAuthNetworkModule,
       },
+      {
+        originalModule: QueueModule,
+        testModule: TestQueueModule,
+      },
```
LucieFaire, line 95: this can be very easily confused with `QueuesModule`. Can we maybe use a more descriptive name or rename the other one? Cause we have queue queue and we have Queue service which is like an datasource and not an actual queue

src/datasources/circuit-breaker/circuit-breaker.keys.ts
```diff
@@ -33,4 +34,14 @@ export class CircuitBreakerKeys {
...
     return CircuitBreakerKeys.SERVICE_PREFIX.DATA_DECODER_SERVICE;
   }
+
+  /**
+   * Generates a circuit breaker key for the queue service
+   *
+   * @param chainId - The chain ID to scope the circuit breaker to
+   * @returns Circuit breaker key in the format: `queue-service-{chainId}`
+   */
+  static getQueueServiceKey(chainId: string): string {
```
LucieFaire: isnt queue-service as decoder-service chain agnostic?

src/modules/queue/queue.interface.ts
```diff
@@ -0,0 +1,139 @@
...
+import type { Page } from '@/domain/entities/page.entity';
+import type { Delegate } from '@/modules/delegate/domain/entities/delegate.entity';
+import type { QueueMessage } from '@/modules/queue/entities/message.entity';
+import type { QueueMultisigTransactionEntity } from '@/modules/queue/entities/multisig-transaction.entity';
+import type { ProposeTransactionDto } from '@/modules/transactions/domain/entities/propose-transaction.dto.entity';
+import type { Raw } from '@/validation/entities/raw.entity';
+
+export const IQueue = Symbol('IQueue');
+
+export interface IQueue {
```
LucieFaire: Im really not sure about calling it just `IQueue` when its a service and not a queue :)

LucieFaire (commented):
first run, and more of a generic qs on naming conventions for the new service integration

src/modules/queue/queue.interface.ts
PooyaRaki: Renamed `IQueue` → `IQueueService` (interface + DI symbol) in `634c1646b`, so it reads as the service client it is rather than a queue. The `QueueService` class, `src/modules/queue/` dir, and `Queue*` entity names are unchanged.

src/__tests__/testing-module.ts
PooyaRaki: Agreed on the collision with the `message-queue` `QueuesModule`. Renamed `QueueModule` → `QueueServiceModule` and `TestQueueModule` → `TestQueueServiceModule` in `634c1646b`.

src/datasources/circuit-breaker/circuit-breaker.keys.ts
PooyaRaki: Correct — the Queue Service is a single deployment for all chains (one `queueService.baseUri`, chain passed as a request param), exactly like the `data-decoder` service, unlike the per-chain `transactionApiManager.getApi(chainId)`. Made `getQueueServiceKey()` chain-agnostic in `634c1646b`

src/modules/delegate/domain/v3/delegates.v3.repository.ts
PooyaRaki: Flipped to the positive if (`this.queueServiceEnabled`) form across the repository and dropped the else blocks (guard-clause + early return), so it reads consistently with `getDelegates/clearDelegates`. `634c1646b`.

src/__tests__/testing-module.ts
compojoom: shouldn't we have the Transaction in the name? It would make it eaiser to distinguish from "MessageQueue", "NotificationQueue" etc. 

QueueService is too generic?

PooyaRaki: Well, we can, but then it just turns into a train of words:

`TransactionQueueModule`
`TestQueueTransactionModule`

PooyaRaki:
I split this into several smaller PRs therefor I'll close this one!

PR #3019, opened 2026-04-10
feat: Relay fee preview

akshay-ap:
Related to: https://linear.app/safe-global/issue/PLA-1237/gtf-cgw-preview-relay-fee-api

### Summary

Integrates with the Fee Service to support "Pay with Safe" — a relay mode where users pay gas via the Fee Service rather than a free quota. Introduces a new `RelayFeeRelayer` strategy and a dedicated `fees` module for fee estimation.

### New API

`POST /v1/chains/:chainId/fees/:safeAddress/preview`

Returns estimated relay fees (gas params, USD cost, pricing context) for a transaction before it is proposed. Requires Pay with Safe to be enabled for the chain (`RELAY_FEE_CHAIN_IDS`). Lives in a dedicated `FeesModule` — intentionally separate from the relay execution domain.

### Changes to existing API

- `GET /v1/chains/:chainId/relay/:safeAddress` — extended with optional `safeTxHash` query param. Used by `RelayFeeRelayer` to check per-transaction eligibility via the Fee Service. Other relayer types ignore it. Fully backward compatible.

### Relay domain

- **`RelayFeeRelayer`** — new `IRelayer` implementation backed by the Fee Service. Takes highest priority in `RelayManager` when the chain is in `RELAY_FEE_CHAIN_IDS`. Uses `IFeeServiceApi.canRelay()` for eligibility checks before relaying; throws `RelayDeniedError` (→ 403) if denied.
- **`RelayManager`** — updated relayer priority: `RelayFeeRelayer` → `DailyLimitRelayer` → `NoFeeCampaignRelayer` → `DailyLimitRelayer` (fallback).
- **`RelayDeniedError` / `RelayDeniedExceptionFilter`** — new error and filter for relay denial by the Fee Service.

### Fee Service datasource (`src/datasources/fee-service-api/`)

`FeeServiceApiService` implementing `IFeeServiceApi`:
- `canRelay` — per-transaction relay eligibility check
- `getRelayFees` — fee estimation with Redis caching (keyed by chain + safe + SHA-256 hash of tx params; TTL: `RELAY_FEE_PREVIEW_TTL_SECONDS`)
- `isPayWithSafeEnabled` — chain enablement check

Cache helpers added to `CacheRouter`: `getRelayFeePreviewCacheKey` and `getRelayFeePreviewCacheDir`.

### Configuration

| Variable | Description |
|---|---|
| `RELAY_FEE_CHAIN_IDS` | Comma-separated chain IDs with Pay with Safe enabled |
| `FEE_SERVICE_BASE_URI` | Base URI for the Fee Service API |
| `RELAY_FEE_PREVIEW_TTL_SECONDS` | Cache TTL for fee preview responses (`0` = disabled) |


akshay-ap:
@claude review

akshay-ap:
> 4. Potential cache poisoning with feePreviewTtlSeconds: 0
fee-service-api.service.ts:130-134 — When feePreviewTtlSeconds is 0, the response is still written to the cache via hSet. Depending on the cache implementation, a TTL of 0 could mean "cache forever" instead of "don't cache." Consider skipping the hSet call entirely when TTL is 0.

`RedisCacheService` already handles it gracedfully

akshay-ap:
> transactions.service.ts:771-778 — The manual mapping from feePreviewDto to request is redundant since FeePreviewTransactionDto already has the exact same fields as TxFeesRequest. Could pass feePreviewDto directly.

This is intentional in case, CGW response deviates from Fee Service response in future

akshay-ap:
> 1. Missing response validation for canRelay
fee-service-api.service.ts:79 — The canRelay method casts the external API response with as unknown as { result: boolean; reason?: string } without any schema validation. Every other external response in this PR is validated with Zod (e.g., TxFeesResponseSchema.parse). An unexpected response shape from the fee service would silently produce incorrect behavior (e.g., result being undefined which is falsy). Add a Zod schema for the canRelay response.

> fee-service-api.service.ts:38-41 — baseUri is read twice from config: once directly as relay.fee.baseUri and once as part of relay.fee. Could derive it from this.relayFeeConfiguration.baseUri instead.


added: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/8564ac03b84a5566c298d1dc8ee4845c5d2441a9

> 2. Sequential relay eligibility checks in relay()
relay-fee.relayer.ts:97-114 — The relay() method checks each limit address sequentially with for...of + await. If there are multiple addresses, these are independent calls that could be parallelized with Promise.all to reduce latency. The existing DailyLimitRelayer and NoFeeCampaignRelayer have the same pattern, so this may be intentional, but worth noting.

NA

>3. Hardcoded value: '0' in relay eligibility checks
relay-fee.relayer.ts:53,103 — In canRelay() (line 53), the value is always hardcoded to '0', and in relay() (line 103), it's also '0'. If the fee service uses the value parameter for eligibility decisions, this could produce incorrect results. The relay() method has access to args.to and args.data but not the actual transaction value — consider whether value should be plumbed through the IRelayer interface

NA

> 6. Missing error handling for cache operations in getRelayFees
fee-service-api.service.ts:108-113 — If this.cacheService.hGet() throws (e.g., Redis connection failure), the entire request fails. Consider wrapping cache reads in a try-catch so that a cache failure falls through to the API call gracefully (similar to how DailyLimitRelayer handles increment failures).

NA

PooyaRaki:
@claude review

akshay-ap:
> Minor
relay-fee.relayer.ts:90 — data parameter is typed as Address (inherited from interface). It should be Hex since it's transaction calldata, not an address. This is a pre-existing issue in the IRelayer interface (relayer.interface.ts:38).
fee-service-api.service.ts:124 — feeData as unknown as TxFeesResponse cast before TxFeesResponseSchema.parse() is unnecessary. The schema parse already handles type validation. Removing it would be cleaner.
Integration test (fee-preview.transactions.controller.integration.spec.ts:135-145) mocks networkService.get for chain/safe responses but these GET calls don't appear to be made by the getFeePreview flow (only isPayWithSafeEnabled + getRelayFees which is a POST). Verify these mocks are actually needed or if they're dead code in the test.

Good catch! Fixed: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/df09febcea59c6b8124ef1cbfb025b288cfe690b

src/config/entities/configuration.ts
```diff
@@ -540,6 +540,13 @@ export default () => ({
           ) ?? [],
       },
     },
+    fee: {
+      enabledChainIds: process.env.RELAY_FEE_CHAIN_IDS?.split(',') ?? [],
+      baseUri: process.env.FEE_SERVICE_BASE_URI ?? '',
```
PooyaRaki: We should validate these in the `configuration.schema.ts` file. If you want to make them nullable you can define them as nullable but the values should be validated.

akshay-ap:
> 1. relay() skips chain enablement check — can relay on disabled chains
relay-fee.relayer.ts:86-121 — Both canRelay() (line 46) and getRelaysRemaining() (line 136) guard against disabled chains by checking enabledChainIds, but relay() does not. If RelayManager.getRelayer() returns the RelayFeeRelayer for a chain and the config is subsequently changed (or there's a mismatch), relay() will proceed to call the fee service and Gelato without the chain guard. Add the same enabledChainIds check at the top of relay().


The issue is invalid. Here's why:

RelayManager.getRelayer() (line 50) gates RelayFeeRelayer behind the exact same enabledChainIds check:

```
if (this.relayFeeConfiguration.enabledChainIds.includes(chainId)) {
  return this.relayFeeRelayer;
}
```

RelayFeeRelayer.relay() is only ever reachable through `RelayRepository.relay() → RelayManager.getRelayer(chainId).relay(args)`. If the chain is not in enabledChainIds, getRelayer returns `DailyLimitRelayer` or `NoFeeCampaignRelayer` instead - `RelayFeeRelayer.relay()` is never called.

The "mismatch after config change" scenario is not a concern because RelayManager reads enabledChainIds once at construction time from the same `IConfigurationService` instance that `RelayFeeRelayer` would use.


> 2. relay() validates via RelaySchema.parse but canRelay inside relay() does not short-circuit on first denial reason
relay-fee.relayer.ts:107-108 — When canRelay is denied in the relay flow, you construct RelayLimitReachedError(address, 0, 0) with hardcoded zeros. This loses the denial reason from the fee service. The error message will say "Relay limit reached for 0x... | current: 0 | limit: 0" which is misleading — it's not a limit issue, it's a fee-engine denial. Consider a more specific error type or including the reason string.

https://github.com/safe-global/safe-client-gateway/pull/3019/commits/2d2b5c294b149b321607cfe6937fd4c80a7cdbb4

> 3. FeePreviewTransactionDtoSchema doesn't validate data as hex
fee-preview-transaction.dto.schema.ts:6-9 — The schema extends TransactionBaseSchema which likely validates to, value, data, and operation. However, data is typed as Hex in the DTO but the schema does not independently verify hex format for data. If TransactionBaseSchema doesn't enforce hex, invalid data could reach the fee service. Worth verifying TransactionBaseSchema covers this.

The issue is invalid. TransactionBaseSchema already validates data against HexSchema at line 10. The concern was already covered — no fix needed.

> 4. Logging safe addresses in plain text
fee-service-api.service.ts:110-112,136-139 — Debug/info logs include safeAddress and chainId in cleartext. This is likely fine for internal logs, but ensure your log aggregation doesn't expose these externally.

This is not a code issue — it's an operational/infrastructure concern about log aggregation configuration, not something to address in the source code.

> 5. No test for getRelaysRemaining when fee service throws
relay-fee.relayer.spec.ts — Tests cover the happy and denied paths for getRelaysRemaining, but there's no test for when feeServiceApi.canRelay throws an error (e.g., network failure). The current code will propagate the error unhandled. The existing DailyLimitRelayer has similar behavior, but it's worth being explicit about error propagation behavior in tests.

good catch! 

https://github.com/safe-global/safe-client-gateway/pull/3019/commits/2bfe8ee55d99a84777008c135f384e89dbea6d6c

> 6. RelayFeeRelayer.relay() passes args.data to canRelay but canRelay sends data: '0x' in its own path
relay-fee.relayer.ts:53-58 vs 98-104 — The canRelay() method (public) always sends data: '0x' as a "lightweight eligibility check", but inside relay(), the actual transaction data is sent. This means canRelay() and relay() may get different answers from the fee service for the same transaction. If the fee service makes decisions based on data content, the canRelay check in the controller could return true while the relay() check later returns false. This could confuse consumers.

good catch! 

https://github.com/safe-global/safe-client-gateway/pull/3019/commits/2bfe8ee55d99a84777008c135f384e89dbea6d6c

akshay-ap:
@PooyaRaki I have addressed the comments by Claude and also pushed new commits.

src/modules/transactions/routes/transactions.controller.ts
```diff
@@ -609,6 +613,55 @@ export class TransactionsController {
     });
   }
 
+  @ApiOperation({
```
akshay-ap: this should go in separate `fee` module. Will push changes in new commit.

akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/9545191b769b124c400e5a66c33441a639229da0

src/modules/relay/routes/relay.controller.ts
```diff
@@ -134,7 +144,12 @@ export class RelayController {
     @Param('chainId') chainId: string,
     @Param('safeAddress', new ValidationPipe(AddressSchema))
     safeAddress: Address,
+    @Query('safeTxHash') safeTxHash?: string,
```
akshay-ap: Instead of adding optional `safeTxHash` parameter, shall we add a dedicated api to check if transaction can be relayed by `safeTxHash`?
Front-end will have to use the new api for GTF in this case.

src/datasources/fee-service-api/fee-service-api.service.ts
```diff
@@ -0,0 +1,137 @@
...
+    @Inject(IConfigurationService)
+    private readonly configurationService: IConfigurationService,
+    private readonly httpErrorFactory: HttpErrorFactory,
+    @Inject(CacheService)
+    private readonly cacheService: ICacheService,
+    @Inject(LoggingService)
+    private readonly loggingService: ILoggingService,
+  ) {
+    this.relayFeeConfiguration =
+      this.configurationService.getOrThrow('relay.fee');
```
LucieFaire, line 34: is it expected that you do not validate if everything within that configuration is present?

src/datasources/fee-service-api/fee-service-api.service.ts
```diff
@@ -0,0 +1,137 @@
...
+    }
+  }
+
+  /**
+   * Gets relay fees from the fee service with internal caching.
+   * 1. Checks cache for existing fee response
+   * 2. On cache miss: calls the external fee service API
+   * 3. Stores the response in cache with configured TTL
+   */
+  async getRelayFees(args: {
```
LucieFaire, line 67: what if instead of dealing with cache yourself you use `CacheFirstDataSource` ?

src/modules/fees/routes/__tests__/fee-preview.fees.controller.integration.spec.ts
```diff
@@ -0,0 +1,167 @@
...
+    const safeAddress = getAddress(faker.finance.ethereumAddress());
+    const feePreviewDto = {
+      to: getAddress(faker.finance.ethereumAddress()),
+      value: '1000000000000000000',
+      data: '0x',
+      operation: Operation.CALL,
+      gasToken: '0x0000000000000000000000000000000000000000',
+      numberSignatures: 2,
+    };
+    const mockFeeResponse = {
```
LucieFaire: nit: as its used in multiple test you can also create a builder instead of recreating it everywhere

src/modules/fees/routes/entities/fee-preview-response.entity.ts
```diff
@@ -0,0 +1,6 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+export {
+  TxDataResponse as FeePreviewTxData,
+  PricingContextSnapshot as FeePreviewPricingContext,
+  TxFeesResponse as FeePreviewResponse,
+} from '@/modules/transactions/domain/entities/relay-fee/tx-fees-response.dto';
```
LucieFaire: do we really need this file?

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,144 @@
...
+   * @param args.chainId - Chain ID
+   * @param args.address - Safe address to check
+   * @param args.safeTxHash - Safe transaction hash for relay-fee eligibility
+   * @returns Result with currentCount=0 and limit=1 when eligible, or all zeros when denied
+   */
+  async canRelay(args: {
+    chainId: string;
+    address: Address;
+    safeTxHash?: string;
+  }): Promise<{ result: boolean; currentCount: number; limit: number }> {
```
LucieFaire: nit: maybe worth to have an interface for this return type

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,144 @@
...
+        const feeServiceResult = await this.feeServiceApi.canRelay({
+          chainId: args.chainId,
+          safeTxHash: args.safeTxHash,
+        });
+
+        if (!feeServiceResult.canRelay) {
+          this.loggingService.info(`relay-fee relay denied for ${address}`);
+          throw new RelayDeniedError(address);
+        }
+      }
```
LucieFaire: hm, something is not right here, we make N number of calls to `canRelay` but all of them are identical as `address` is not used. Think we can optimize:
```
if (args.safeTxHash) {
      const feeServiceResult = await this.feeServiceApi.canRelay({
        chainId: args.chainId,
        safeTxHash: args.safeTxHash,
      });

      if (!feeServiceResult.canRelay) {
        const [address] = relayAddresses;
        this.loggingService.info(`relay-fee relay denied for ${address}`);
        throw new RelayDeniedError(address);
      }
    }
```

src/modules/transactions/routes/transactions.controller.ts
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
```
LucieFaire, line 1: hm dont see any change in this file for license header

src/modules/transactions/routes/transactions.service.ts
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
```
LucieFaire, line 1: same here

src/modules/transactions/transactions.module.ts
```diff
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
```
LucieFaire, line 1: and here

LucieFaire (commented):
left some small comments

src/modules/fees/routes/entities/fee-preview-response.entity.ts
akshay-ap: The purpose is have naming of request/response dtos relevant to the api i.e. `*/preview`.
As of now, `*/preview` API uses same request/response as Fee service api so exports are simply renamed.
In future, if both deviate, we can create new interfaces/classes here.

src/modules/transactions/routes/transactions.controller.ts
akshay-ap: There are changes in this file in some of the previous commits in this PR and hence this header was added. We can squash the commits, force push and remove the header if we are strict with it.

src/modules/transactions/routes/transactions.service.ts
akshay-ap: There are changes in this file in some of the previous commits in this PR and hence this header was added. We can squash the commits, force push and remove the header if we are strict with it.

src/modules/transactions/transactions.module.ts
akshay-ap: There are changes in this file in some of the previous commits in this PR and hence this header was added. We can squash the commits, force push and remove the header if we are strict with it.

src/datasources/fee-service-api/fee-service-api.service.ts
akshay-ap: I skipped adding validation in constructor because fee related params are optional for CGW to work.
The defaults are empty values.

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: good catch!

akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/35bf52b789ace3add7982a87552266207c829e2d

Also, added a new error `RelayTxDeniedError` because we cannot reuse `RelayDeniedError` here after the change.

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/ece491ba4f6140feaaef7d95fc7f95eba93a4abe

src/modules/relay/domain/exception-filters/relay-tx-denied.exception-filter.ts
```diff
@@ -0,0 +1,22 @@
...
+import {
+  Catch,
+  ExceptionFilter,
+  ArgumentsHost,
+  HttpStatus,
+} from '@nestjs/common';
+import { RelayTxDeniedError } from '@/modules/relay/domain/errors/relay-tx-denied.error';
+
+@Catch(RelayTxDeniedError)
+export class RelayTxDeniedExceptionFilter implements ExceptionFilter {
```
LucieFaire: Im not sure why we need this filter and the other one as NestJS's built-in exception handling would already return a 403 for them, but if we do want to have them we can merge them in 1: 
```
@Catch(RelayDeniedError, RelayTxDeniedError)
```
cause they do the same

LucieFaire (commented):
one more

src/modules/relay/domain/exception-filters/relay-tx-denied.exception-filter.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/6921dabdb5e9f6867ff16405a738a9f39ed7ce00

Good suggestion!

src/datasources/fee-service-api/fee-service-api.service.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/8f3c2e76377fb87f942931e65cdfb22554202e10

src/modules/fees/routes/fees.service.ts
```diff
@@ -0,0 +1,43 @@
...
+      operation: args.feePreviewDto.operation,
+      gasToken: args.feePreviewDto.gasToken,
+      numberSignatures: args.feePreviewDto.numberSignatures,
+    };
+
+    const txFeesResponse = await this.feeServiceApi.getRelayFees({
+      chainId: args.chainId,
+      safeAddress: args.safeAddress,
+      request,
+    });
```
LucieFaire: I think we can just pass it directly as `request` is 1-1 as `feePreviewDto`
```
const txFeesResponse = await this.feeServiceApi.getRelayFees({
      chainId: args.chainId,
      request: args.feePreviewDto,
    });
```

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,142 @@
...
+import { RelayFeeConfiguration } from '@/modules/relay/domain/entities/relay.configuration';
+
+@Injectable()
+export class RelayFeeRelayer implements IRelayer {
+  private readonly relayFeeConfiguration: RelayFeeConfiguration;
+
+  constructor(
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
+    @Inject(IConfigurationService) configurationService: IConfigurationService,
+    private readonly limitAddressesMapper: LimitAddressesMapper,
```
LucieFaire: i think `limitAddressesMapper` is not used anywhere here

LucieFaire (approved):
2 teeny tiny remarks, otherwise great job! thank you for addressing all comments 💪

src/modules/fees/routes/fees.service.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/a4653e3b9029f7900ecc97e13c7ded829e731678

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/a4653e3b9029f7900ecc97e13c7ded829e731678

src/modules/fees/routes/entities/__tests__/fee-preview-transaction.dto.builder.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { IBuilder } from '@/__tests__/builder';
+import { Builder } from '@/__tests__/builder';
+import type { FeePreviewTransactionDto } from '@/modules/fees/routes/entities/fee-preview-transaction.dto.entity';
+import { Operation } from '@/modules/safe/domain/entities/operation.entity';
+
+export function feePreviewTransactionDtoBuilder(): IBuilder<FeePreviewTransactionDto> {
+  return new Builder<FeePreviewTransactionDto>()
+    .with('to', getAddress(faker.finance.ethereumAddress()))
+    .with('value', '0')
+    .with('data', '0x')
```
PooyaRaki: Should we send some random hex data here? We can use Faker to create one.

src/modules/fees/routes/entities/fee-preview-transaction.dto.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+  typeof FeePreviewTransactionDtoSchema
+> {
+  @ApiProperty()
+  to!: Address;
+
+  @ApiProperty()
+  value!: string;
+
+  @ApiProperty()
+  data!: Hex;
```
PooyaRaki: We should be consistent. Either use `!` everywhere or assign the values in the constructor method.
I know there are some inconstancies in the service but at least in the PRs from now on we should be consistent.

src/modules/fees/routes/fees.controller.ts
```diff
@@ -0,0 +1,72 @@
...
+      'Fee preview with transaction data, relay cost, and pricing context',
+  })
+  @ApiBadRequestResponse({
+    description:
+      'Invalid transaction data or Pay with Safe not available for this chain',
+  })
+  @HttpCode(200)
+  @Post(':safeAddress/preview')
+  async getFeePreview(
+    @Param('chainId') chainId: string,
```
PooyaRaki: Should be validated.

src/modules/relay/domain/errors/relay-denied.error.ts
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { ForbiddenException } from '@nestjs/common';
+import type { Address } from 'viem';
+
+export class RelayDeniedError extends ForbiddenException {
```
PooyaRaki, line 5: For consistency, should be `RelayDeniedException`

src/modules/relay/domain/errors/relay-tx-denied.error.ts
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { ForbiddenException } from '@nestjs/common';
+import type { Hex } from 'viem';
+
+export class RelayTxDeniedError extends ForbiddenException {
```
PooyaRaki, line 5: The class name should end with `Exception` instead of `Error`

src/modules/relay/routes/relay.controller.ts
PooyaRaki: It's fine to have an optional parameter for now since it's only one, but it should be validated.
In future if we have more parameters to check we should create a new API.

src/modules/relay/routes/relay.controller.ts
```diff
@@ -134,7 +144,12 @@ export class RelayController {
     @Param('chainId') chainId: string,
```
PooyaRaki: It should be validated here.

src/modules/relay/routes/relay.controller.integration.spec.ts
```diff
@@ -2846,6 +2846,17 @@ describe('Relay controller', () => {
...
 
+      it('should accept safeTxHash query parameter and return relay limits', async () => {
+        const safeAddress = faker.finance.ethereumAddress();
+        const safeTxHash = faker.string.hexadecimal({ length: 64 });
+        await request(app.getHttpServer())
+          .get(
+            `/v1/chains/${chainId}/relay/${safeAddress}?safeTxHash=${safeTxHash}`,
+          )
+          .expect(200)
+          .expect({ remaining: 5, limit: 5 });
```
PooyaRaki: Shouldn't we read the remaining from the config or if you're mocking them from a constant defined in this file?
The tests will break if you change the remaining and limit in the configuration.

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,140 @@
...
+   * @returns Relay result from the relay API
+   */
+  async relay(args: {
+    version: string;
+    chainId: string;
+    to: Address;
+    data: Hex;
+    gasLimit: bigint | null;
+    safeTxHash?: Hex;
+  }): Promise<Relay> {
```
PooyaRaki, line 90: QQ: Should we also check here if the fee relay is available or do we assume that it's already checked when this method is called?

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,140 @@
...
+  }): Promise<{ remaining: number; limit: number }> {
+    if (!this.relayFeeConfiguration.enabledChainIds.includes(args.chainId)) {
+      return { remaining: 0, limit: 0 };
+    }
+
+    // For relay-fee, the FeeService API is the authority on relay eligibility.
+    // We report a simplified view: remaining=1 if eligible, 0 if not.
+    const feeServiceResult = await this.feeServiceApi.canRelay({
+      chainId: args.chainId,
+      safeTxHash: args.safeTxHash ?? '',
```
PooyaRaki: QQ: `arg.safeTxHash` is already nullable so why do we need `?? ''`? canRelay also accepts a nullable param.

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,140 @@
...
+    safeTxHash?: Hex;
+  }): Promise<Relay> {
+    if (args.safeTxHash) {
+      const feeServiceResult = await this.feeServiceApi.canRelay({
+        chainId: args.chainId,
+        safeTxHash: args.safeTxHash,
+      });
+
+      if (!feeServiceResult.canRelay) {
+        this.loggingService.info(
```
PooyaRaki: Why is it an info level log? Shouldn't it be an error? We should also define a type for the error log. You may check the `log-type.entity.ts` file.

PooyaRaki (commented):
Great job @akshay-ap 💯 
I left a couple of small comments to improve it even further.

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: Relay manager takes care of selecting the appropriate relayer so the checks can be skipped here.
We will have duplicate checks if we add them here too.

https://github.com/safe-global/safe-client-gateway/blob/cc1d87fec2712a00c8c3df424bb3bd5c1acdb6a9/src/modules/relay/domain/relay.repository.ts#L24

https://github.com/safe-global/safe-client-gateway/blob/cc1d87fec2712a00c8c3df424bb3bd5c1acdb6a9/src/modules/relay/domain/relay.manager.ts#L38

src/config/entities/configuration.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/cc1d87fec2712a00c8c3df424bb3bd5c1acdb6a9

src/modules/fees/routes/entities/__tests__/fee-preview-transaction.dto.builder.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/9b70582ee507ea42179a2f28b044927493cc9415

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: Fee service interface requires `safeTxHash` as non-nullable.

https://github.com/safe-global/safe-client-gateway/blob/9b70582ee507ea42179a2f28b044927493cc9415/src/domain/interfaces/fee-service-api.interface.ts#L15

I think will will do the work: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/22164072f51596c4a1ceb562a8fe7a635b60296f

We don't need additional validation as fee service will reject the request if `safeTxHash` is invalid

src/modules/relay/routes/relay.controller.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/22164072f51596c4a1ceb562a8fe7a635b60296f

src/modules/relay/domain/errors/relay-tx-denied.error.ts
akshay-ap: All the classes in this `/error` are errors and extend `*Exception`. Why this specific class should end with `Exception`?

src/config/entities/configuration.ts
PooyaRaki: Well, I meant we don't probably need ` ?? ''` if it's defined as nullable but it's a small nitpick I'll resolve it for now.

src/modules/relay/domain/errors/relay-tx-denied.error.ts
PooyaRaki: You're right. I'll resolve it. Ideally they should have been `*Exception` but for consistency you're approach is the better solution here 👍 .

src/modules/relay/routes/relay.controller.ts
PooyaRaki: It's still not validated in the latest commit.

src/modules/relay/routes/relay.controller.integration.spec.ts
akshay-ap: Good catch!
https://github.com/safe-global/safe-client-gateway/pull/3019/commits/74134802c3c54dad6b9500be303021bc41eb683e

src/modules/relay/domain/errors/relay-tx-denied.error.ts
akshay-ap: can be worked on separately: https://github.com/safe-global/safe-client-gateway/issues/3025

src/modules/relay/domain/errors/relay-denied.error.ts
akshay-ap: can be worked on separately: https://github.com/safe-global/safe-client-gateway/issues/3025

src/config/entities/configuration.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/90e44e3a057885d27aa7d22238544ded88dbaf15

src/modules/fees/routes/entities/fee-preview-transaction.dto.entity.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/a7cf84a213926f46f7533d6b609411670789a274

src/modules/fees/routes/fees.controller.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/6bdafaf8669cd6947b6b31bb584cd7785fb3e36e

src/modules/relay/routes/relay.controller.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/2a46744860d85930df5cc188a7c78b6fb881679f

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/156b4a111ecdc002e29f032177a23b9846c34e09

yagopv:
`RelayManager.getRelayer()` (src/modules/relay/domain/relay.manager.ts) selects the relayer based on `chainId`. If the chain is in `RELAY_FEE_CHAIN_IDS`, it always returns `RelayFeeRelayer` regardless of whether `safeTxHash` is present

Then in `RelayFeeRelayer.relay()` (src/modules/relay/domain/relayers/relay-fee.relayer.ts), the fee service eligibility check is wrapped in `if (args.safeTxHash)`. When no hash is provided, the method skips the check entirely and calls `relayApi.relay()` directly

This means a `POST /relay` request on a fee enabled chain without `safeTxHash` in the body will:
- Bypass the fee service eligibility check
- Bypass the daily limit relayer (since `RelayManager` already chose `RelayFeeRelayer`)
- Execute the relay against Gelato with no controls at all

Is this intentional ?

akshay-ap:
> `RelayManager.getRelayer()` (src/modules/relay/domain/relay.manager.ts) selects the relayer based on `chainId`. If the chain is in `RELAY_FEE_CHAIN_IDS`, it always returns `RelayFeeRelayer` regardless of whether `safeTxHash` is present
> 
> Then in `RelayFeeRelayer.relay()` (src/modules/relay/domain/relayers/relay-fee.relayer.ts), the fee service eligibility check is wrapped in `if (args.safeTxHash)`. When no hash is provided, the method skips the check entirely and calls `relayApi.relay()` directly
> 
> This means a `POST /relay` request on a fee enabled chain without `safeTxHash` in the body will:
> 
> * Bypass the fee service eligibility check
> * Bypass the daily limit relayer (since `RelayManager` already chose `RelayFeeRelayer`)
> * Execute the relay against Gelato with no controls at all
> 
> Is this intentional ?

Good catch! Looks like a bug. Let me review if it got missed in the tests and add if so. It is expected that Daily limit relayer is not called for chains enabled for `Relay fee`. But, it is not expected to bypass validation logic in `relay-fee.relayer.ts` if safeTxHash is missing.

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,143 @@
...
+      return { result: false, currentCount: 0, limit: 0 };
+    }
+
+    const feeServiceResult = await this.feeServiceApi.canRelay({
+      chainId: args.chainId,
+      safeTxHash: args.safeTxHash,
+    });
+
+    if (!feeServiceResult.canRelay) {
+      this.loggingService.error({
```
yagopv: I think this is more as a warn or an info log as i see it as a normal bussiness log right?. If we treat it as an error, it can trigger events in the Datadog monitoring system

akshay-ap: @PooyaRaki wdyt? We might get false alerts in CGW. Are you ok with `error` or `warn`?

PooyaRaki (approved):

src/modules/relay/domain/relayers/relay-fee.relayer.ts
PooyaRaki: Makes sense to change this one to `warn`. The other one should remain an error since that actually throws an error after logging.
The text is also a bit confusing, It doesn't say why it was denied.

akshay-ap: > The text is also a bit confusing, It doesn't say why it was denied.

Fee-service does not give message on denial reason yet. Will propose another PR for this once we have if from fee service.

akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3019/commits/91e8441f4e3463b12df58aa8ba0d07214d223803

PooyaRaki (approved):

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,145 @@
...
+    to: Address;
+    data: Hex;
+    gasLimit: bigint | null;
+    safeTxHash?: Hex;
+  }): Promise<Relay> {
+    if (!args.safeTxHash) {
+      throw new RelayTxDeniedError(args.safeTxHash);
+    }
+
+    const feeServiceResult = await this.feeServiceApi.canRelay({
```
yagopv, line 95: Does the fee service verify that the approved safeTxHash matches the data ?

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -0,0 +1,145 @@
...
+    if (!feeServiceResult.canRelay) {
+      this.loggingService.error({
+        type: LogType.TxRelayEligibility,
+        message: `relay-fee relay denied for ${args.safeTxHash}`,
+      });
+      throw new RelayTxDeniedError(args.safeTxHash);
+    }
+
+    const relayResponse = await this.relayApi
+      .relay(args)
```
yagopv, line 109: `IRelayApi.relay()` in `relay-api.interface.ts`) only accepts { chainId, to, data }, but here we pass all the args (including version, gasLimit, safeTxHash)

src/config/entities/configuration.ts
```diff
@@ -540,6 +540,13 @@ export default () => ({
           ) ?? [],
       },
     },
+    fee: {
+      enabledChainIds: process.env.RELAY_FEE_CHAIN_IDS?.split(',') ?? [],
+      baseUri: process.env.FEE_SERVICE_BASE_URI,
+      feePreviewTtlSeconds: parseInt(
+        process.env.RELAY_FEE_PREVIEW_TTL_SECONDS ?? `${0}`,
```
yagopv, line 547: The default value of 0 in Redis indicates that the key never expires right ?

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: We are tackling it in another PR.

src/config/entities/configuration.ts
akshay-ap: In deployments it is set to `60`. But, lets update it as you suggest to default `10` to avoid cache never expiring.

akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3032

PR #3026, opened 2026-04-16
feat: private address book contacts and request-to-add workflow

katspaugh:
> Private contacts bloom in silence,
> requests queue for admin eyes,
> approve and they join the space.

Part of https://linear.app/safe-global/issue/WA-1973/redesign-spaces-address-book-to-match-prototype

## Summary

- Non-admin Space members can create **private contacts** visible only to them
- Members can **request to add** private contacts to the shared space address book
- Admins see a **pending requests** list and can **approve** or **reject** each request
- Approved contacts are upserted into the existing `space_address_book_items` table

## What it solves

Currently only admins can add contacts to the space address book. Members have read-only access. This feature gives members a way to manage their own contacts within a space and propose additions to the shared book.

### New tables (migration)
- `user_address_book_items` — per-user private contacts scoped to a space (unique on `space_id, creator_id, address`)
- `address_book_requests` — pending approval requests with `PENDING/APPROVED/REJECTED` status enum

### New endpoints
| Method | Path | Auth | Description |
|--------|------|------|-------------|
| `GET` | `/v1/spaces/:spaceId/address-book/private` | Member+ | Get caller's private contacts |
| `PUT` | `/v1/spaces/:spaceId/address-book/private` | Member+ (SIWE) | Upsert private contacts |
| `DELETE` | `/v1/spaces/:spaceId/address-book/private/:address` | Member+ (SIWE) | Delete private contact |
| `GET` | `/v1/spaces/:spaceId/address-book/requests` | Member+ | Pending requests (admin: all, member: own) |
| `POST` | `/v1/spaces/:spaceId/address-book/requests` | Member+ (SIWE) | Request to add private contact to space |
| `PUT` | `/v1/spaces/:spaceId/address-book/requests/:id/approve` | Admin (SIWE) | Approve request → upserts to shared book |
| `PUT` | `/v1/spaces/:spaceId/address-book/requests/:id/reject` | Admin (SIWE) | Reject request → contact stays private |

### Architecture
Follows the **counterfactual safes pattern** — separate user-scoped table with `creator_id` FK, `assertMember()`/`assertAdmin()` for authorization, no changes to existing address book endpoints.

## How to test it

- [ ] Run migration: `yarn migration:run`
- [ ] Private contacts CRUD as member (create, read, update, delete)
- [ ] Verify private contacts are not visible to other members
- [ ] Create request from private contact → appears in pending
- [ ] Admin approves → contact appears in shared address book
- [ ] Admin rejects → contact stays private
- [ ] Member cannot approve/reject requests

## Visual summary

```mermaid
flowchart TD
    M[Member] -->|PUT /private| PC[user_address_book_items]
    M -->|POST /requests| R[address_book_requests<br/>status: PENDING]
    A[Admin] -->|PUT /approve| R
    R -->|approved| S[space_address_book_items<br/>shared contacts]
    A -->|PUT /reject| R2[address_book_requests<br/>status: REJECTED]
```

## Checklist

- [x] Type-check passes (`npx tsc --noEmit`)
- [x] E2E tests written (16 tests across 2 spec files)
- [x] No changes to existing address book endpoints
- [x] Follows existing patterns (DDD, repository interfaces, Swagger decorators)


katspaugh:
@claude review

src/modules/spaces/domain/address-books/address-book-requests.repository.ts
```diff
@@ -0,0 +1,105 @@
...
+      space: { id: args.spaceId },
+      requestedBy: { id: args.requestedById },
+      address: args.item.address,
+      name: args.item.name,
+      chainIds: args.item.chainIds,
+      status: 'PENDING',
+    });
+    const insertedId = result.identifiers[0].id;
+    return this.findOneOrFail({ id: insertedId, spaceId: args.spaceId });
+  }
```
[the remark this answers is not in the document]
katspaugh: Handled in d873d6de — `create()` now wraps the insert and maps `isUniqueConstraintError` → `UniqueConstraintError` (409), matching the `counterfactual-safes.repository` pattern.

src/modules/spaces/routes/address-book-requests.service.ts
```diff
@@ -0,0 +1,248 @@
...
+  private async mapRequestItem(
+    request: AddressBookRequest,
+  ): Promise<AddressBookRequestItemDto> {
+    let requestedByAddress = '';
+    if (request.requestedBy?.id) {
+      const wallets = await this.walletsRepository.findByUser(
+        request.requestedBy.id,
+      );
+      requestedByAddress = wallets[0]?.address ?? '';
+    }
```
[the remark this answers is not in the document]
katspaugh: Handled in d873d6de. Persisted the requester wallet as a new `requested_by_wallet` column on `address_book_requests`, populated at creation from `authPayload.signer_address`. The listing now reads it straight from the row (no per-request wallet query) and `IWalletsRepository` is no longer injected into the service.

src/modules/spaces/routes/address-book-requests.service.ts
```diff
@@ -0,0 +1,248 @@
...
+      const privateContact =
+        await this.privateRepository.findOneBySpaceCreatorAndAddress({
+          spaceId,
+          creatorId: requesterId,
+          address: request.address,
+        });
+      if (privateContact?.createdBy) {
+        createdByOverride = privateContact.createdBy;
+      }
+    }
```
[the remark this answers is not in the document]
katspaugh: Handled in d873d6de. `requested_by_wallet` is now persisted on the request row and used as `createdByOverride` at approval time, so attribution is stable even if the requester later edits or deletes the private contact. The approve path no longer re-fetches the private contact.

src/modules/spaces/domain/address-books/address-book-items.repository.interface.ts
```diff
@@ -36,6 +37,7 @@ export interface IAddressBookItemsRepository {
     authPayload: AuthPayload;
     spaceId: Space['id'];
     addressBookItems: UpsertAddressBookItemsDto['items'];
+    createdByOverride?: `0x${string}`;
   }): Promise<Array<AddressBookDbItem>>;
```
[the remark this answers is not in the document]
katspaugh: Retyped as `Address` from `viem` in d873d6de (both the interface and the implementation).

katspaugh:
Second-pass review addressed in ff2cfa36.

**Bugs / Correctness**
- *Race in approve:* `approve()` and `reject()` now go through `transitionFromPending`, an atomic conditional `UPDATE ... WHERE status='PENDING'`. Only one concurrent caller can win the transition; if the following `upsertMany` throws, `revertToPending` puts the request back so it can be retried. Removes the double-approval window and the "500 → inconsistent state" risk.
- *Missing maxItems on private contacts:* `UserAddressBookItemsRepository` now injects `IConfigurationService` and enforces `spaces.addressBooks.maxItems` per `(space, creator)` in `checkItemsLimit`, mirroring the shared repo.
- *Double query in upsertMany:* `updateExisting` returns the addresses it already loaded; `getExistingAddresses` removed.

**Security**
- *Duplicate request guard:* already fixed in d873d6de via `isUniqueConstraintError` → `UniqueConstraintError` (409).
- *`requestedBy` leaks wallet:* per your note, acceptable — no change.

**Performance**
- *N+1 wallet lookup:* already fixed in d873d6de — requester wallet is persisted as `requested_by_wallet` and read straight from the row.
- *`isSpaceAdmin` exception-for-control-flow:* added `isAdmin(spacesRepository, spaceId, userId): Promise<boolean>` in `space-assert.utils`; `assertAdmin` delegates to it; service calls the boolean version directly.

**Minor**
- *`lastUpdatedBy` misleading:* new `UserAddressBookDto` / `UserAddressBookItemDto` without `lastUpdatedBy`; wired into the private controller + service, e2e assertions updated. The `user_address_book_items` table is single-author by design, so no field to map.
- *`reviewedBy` wallet vs `requested_by` FK:* pushing back — different purposes. `reviewed_by` is an audit snapshot of the SIWE signer at review time (same pattern as `created_by`/`last_updated_by` on `space_address_book_items`); `requested_by` is an FK because we query "requests by user X", with `requested_by_wallet` added as its matching audit snapshot. Both sides have the same (FK + wallet) story now. Happy to revisit if there's a concrete scenario where the current layout breaks auditing.

katspaugh:
## PR Briefing


https://github.com/user-attachments/assets/fea1ae15-c0d7-460a-b8ee-72fcbfd40541



This PR introduces private address book contacts and a request-to-add approval workflow for Safe Spaces. Non-admin members can now create private contacts (stored in the new `user_address_book_items` table) visible only to themselves, and can submit promotion requests (stored in `address_book_requests`) for admin review. Admins can approve or reject these requests via new endpoints in `AddressBookRequestsController`; approval upserts the contact into the shared `space_address_book_items` table. The feature is implemented following existing DDD patterns with new repository interfaces, services, controllers, DTOs, and a TypeORM migration, and is covered by 16 e2e tests across two new spec files.

<details>

## Summary

This PR introduces private address book contacts and a request-to-add approval workflow for Safe Spaces. Non-admin members can now create private contacts (stored in the new `user_address_book_items` table) visible only to themselves, and can submit promotion requests (stored in `address_book_requests`) for admin review. Admins can approve or reject these requests via new endpoints in `AddressBookRequestsController`; approval upserts the contact into the shared `space_address_book_items` table. The feature is implemented following existing DDD patterns with new repository interfaces, services, controllers, DTOs, and a TypeORM migration, and is covered by 16 e2e tests across two new spec files.

## Architecture notes

Two new DB tables (`user_address_book_items`, `address_book_requests`) are added via a single migration, each with their own TypeORM entities, repository interfaces/implementations, and NestJS service/controller pairs (`UserAddressBookController`/`UserAddressBookService` and `AddressBookRequestsController`/`AddressBookRequestsService`). The approval path in `AddressBookRequestsService.approve` calls `transitionFromPending` then `upsertAddressBookItems` with a new optional `createdByOverride` parameter added to `IAddressBookItemsRepository`, attributing the shared contact to the original requester rather than the approving admin. Authorization reuses `assertMember`/`assertAdmin` helpers, with `space-assert.utils.ts` refactored to expose a non-throwing `isAdmin` boolean helper. All new providers and entities are wired into `SpacesModule`.

## Ripple risks

Analyzed 4 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 21 file(s), 29 changed export(s), 0 uncovered callsite(s) in 4.2s._

## Walkthrough outline

_21 steps across 7 chapters, approximately 5 min 45s._

### Why this change

1. **motivation** · `src/modules/spaces/routes/utils/space-assert.utils.ts` L10-30 `isAdmin`
   Right now only admins can write to the shared space address book — members are read-only. This PR cracks that open by giving members a private lane and a way to propose contacts for admin approval. The first clue is here: assertAdmin gets refactored to expose a non-throwing isAdmin boolean, because the new request endpoints need to branch on role rather than blow up.

2. **motivation** · `src/modules/spaces/routes/utils/space-assert.utils.ts` L10-30 `isAdmin`
   Here is the full picture. Members write to their own private table and optionally promote a contact into a pending request. Admins review that queue and either approve — which upserts into the shared book — or reject, leaving the contact private. Nothing touches the existing shared address book endpoints.

### New database tables

3. **change** · `migrations/1776556800000-user-address-book-items.ts` L1-40 `UserAddressBookItems1776556800000`
   The migration is the foundation. It creates two new tables in one shot: user_address_book_items for private contacts and address_book_requests for the approval queue. Both get update triggers and indexes, and both have unique constraints to prevent duplicate entries per member per address.

4. **change** · `src/modules/spaces/datasources/entities/user-address-book-item.entity.db.ts` L18-60 `UserAddressBookItem`
   The TypeORM entity for private contacts maps that table. Notice the unique constraint on space, creator, and address — one private entry per member per address per space. Both Space and User FKs use CASCADE delete, so removing a user silently wipes all their private contacts.

5. **change** · `src/modules/spaces/datasources/entities/address-book-request.entity.db.ts` L23-70 `AddressBookRequest`
   The request entity stores who asked, what address and name they want promoted, and the current status as an integer via a transformer. One thing to flag: updatedAt has update:false, meaning TypeORM will never refresh it on save — so the reviewed timestamp won't move when an admin approves or rejects.

### The private contacts feature

6. **change** · `src/modules/spaces/domain/address-books/entities/user-address-book-item.entity.ts` L14-35 `UserAddressBookItemSchema`
   Up in the domain layer, the Zod schema defines the shape the service layer works with. It has both a creator relation and a createdBy address string — two representations of the same fact. Make sure these stay in sync through the ORM mapping.

7. **change** · `src/modules/spaces/domain/address-books/user-address-book-items.repository.ts` L14-80 `UserAddressBookItemsRepository`
   The repository implementation enforces a configurable max-items limit per user per space via checkItemsLimit. The upsert flow opens a transaction for updates and inserts, but then calls findBy outside that transaction — so if the transaction rolls back, the caller still gets a result without an error.

8. **change** · `src/modules/spaces/routes/user-address-book.controller.ts` L38-90 `UserAddressBookController`
   The controller exposes three endpoints under /v1/spaces/:spaceId/address-book/private — GET, PUT, and DELETE. One quick note: DELETE is decorated with ApiNoContentResponse but there is no HttpCode(204) decorator, so NestJS will actually return 200.

9. **change** · `src/modules/spaces/routes/user-address-book.service.ts` L13-60 `UserAddressBookService`
   The service enforces assertMember for all three operations and adds a SIWE check on writes. Notice findAll skips the SIWE check — reads are allowed with regular session auth. The mapItems helper manually spreads fields, which means any new field added to the entity in future needs a matching line here.

### The request-to-add workflow

10. **change** · `src/modules/spaces/domain/address-books/entities/address-book-request.entity.ts` L14-43 `AddressBookRequestStatus`
   The domain entity for requests defines the PENDING, APPROVED, REJECTED status enum. The status is stored as an integer in the DB but the domain uses string keys — the databaseEnumTransformer handles the round-trip, so verify the enum ordering never shifts.

11. **change** · `src/modules/spaces/domain/address-books/address-book-requests.repository.ts` L18-80 `AddressBookRequestsRepository`
   The repository has a key method: transitionFromPending, which does an atomic conditional update — it only flips status if the current value is PENDING. It returns a boolean so callers can detect a race where two admins try to act simultaneously. The plain updateStatus method does not scope by spaceId, which is an inconsistency worth noting.

12. **change** · `src/modules/spaces/routes/address-book-requests.service.ts` L29-90 `AddressBookRequestsService`
   This is the heart of the approval workflow. The approve method calls transitionFromPending, then upserts into the shared book, and if that upsert fails it calls revertToPending as compensation. The problem: if revertToPending itself throws, the request is stuck in APPROVED with no contact written — a split-brain state with no alerting.

13. **change** · `src/modules/spaces/domain/address-books/address-book-items.repository.ts` L23-60 `AddressBookItemsRepository`
   To attribute the shared contact to the original requester rather than the approving admin, upsertAddressBookItems gains an optional createdByOverride parameter. When provided it substitutes the createdBy field. Note that lastUpdatedBy still records the admin — that asymmetry is intentional but worth a reviewer eyeball.

14. **change** · `src/modules/spaces/routes/address-book-requests.controller.ts` L38-90 `AddressBookRequestsController`
   The controller wires four endpoints: GET list, POST create, PUT approve, PUT reject. All are behind AuthGuard. The PR description says approve and reject require SIWE wallet auth, but there is no dedicated SIWE guard visible here — that enforcement lives inside the service's assertSiwe call.

### Consumers that had to adapt

15. **ripple** · `src/modules/spaces/spaces.module.ts` L38-90 `SpacesModule`
   SpacesModule is the wiring harness. Two new entities, two new repository pairs, and two new controller-service pairs all get registered here. Both new repository interfaces are also added to exports — double-check that downstream modules actually need them, or you are leaking internals.

16. **ripple** · `src/modules/spaces/routes/entities/address-book-request.dto.entity.ts` L40-55 `CreateAddressBookRequestSchema`
   CreateAddressBookRequestDto only validates an Ethereum address — name and chainIds are not in the creation payload. That means the service must look them up from the existing private contact record. Verify that lookup is enforced server-side and cannot be bypassed.

17. **ripple** · `src/modules/spaces/routes/utils/space-assert.utils.ts` L10-30 `isAdmin`
   Coming back to the refactored space-assert utils — isAdmin returns false both when the space is not found and when the user is not an admin. Callers cannot distinguish those two cases, which could mask a 404 scenario as a plain permission denial.

### What can go wrong

18. **risk** · `migrations/1776556800000-user-address-book-items.ts` L1-40 `UserAddressBookItems1776556800000`
   The unique constraint on address_book_requests covers space, requester, and address — regardless of status. So a member whose request was rejected cannot re-submit for the same address without a manual DB delete. Decide whether that is the intended UX or whether status should be part of the unique key.

19. **risk** · `src/modules/spaces/routes/address-book-requests.controller.e2e-spec.ts` L1-50 `AddressBookRequestsController`
   The e2e tests share a single app instance with no database cleanup between describe blocks. The assertion that an admin sees exactly two requests assumes no state leaked from earlier tests. Also, inviteMember never simulates the member accepting the invitation, so tests may pass with looser membership semantics than production enforces.

20. **risk** · `src/modules/spaces/routes/address-book-requests.service.ts` L29-60 `AddressBookRequestsService`
   There is no database transaction wrapping the transitionFromPending plus upsertAddressBookItems pair in approve. The manual revertToPending compensation is fragile — a process crash between those two calls leaves the system in an inconsistent state with no recovery path.

### The bottom line

21. **wrap** · `src/modules/spaces/spaces.module.ts` L38-90 `SpacesModule`
   This PR cleanly extends the spaces address book without touching existing endpoints — members get private contacts, admins get an approval queue, and the shared book stays the source of truth — but the approval path's manual compensation logic and the rejected-request re-submission block are the two things worth resolving before merge.

## Reviewer checklist

- [ ] Verify the approve/reject flow is atomic: `AddressBookRequestsService.approve` calls `transitionFromPending` and then `upsertAddressBookItems` without a wrapping transaction — confirm the manual `revertToPending` compensation is sufficient and that a failure in `revertToPending` itself cannot leave a request in APPROVED state with no contact written.
- [ ] Check that the `address_book_requests` unique constraint on `(space_id, requested_by, address)` ignores status, meaning a member whose request was REJECTED cannot re-submit for the same address without manual DB intervention — decide if this is the intended UX or if status should be part of the key.
- [ ] Confirm `updatedAt` columns on both `AddressBookRequest` and `UserAddressBookItem` TypeORM entities have `update: false` set — if these columns are intended to track review/modification time they will never be updated by TypeORM saves, which is likely a bug.
- [ ] Audit the e2e tests in both spec files to confirm `inviteMember` actually results in an accepted MEMBER-role membership (not just an INVITED state) before tests exercise member-only endpoints, and that database state is isolated between test cases to prevent bleed-through.
- [ ] Validate that `createdByOverride` in `upsertAddressBookItems` is only reachable via the admin-approval code path and that the value (sourced from `request.requestedByWallet`) is validated as a checksummed Ethereum address before being persisted, preventing spoofed creator attribution.
- [ ] Check that the DELETE endpoint in `UserAddressBookController` returns HTTP 204 (needs `@HttpCode(HttpStatus.NO_CONTENT)`) rather than the default 200, and that attempting to delete a non-existent or another user's private contact returns an appropriate 404/403 rather than a silent no-op.

## File-by-file

### `migrations/1776556800000-user-address-book-items.ts`

Adds migration `UserAddressBookItems1776556800000` creating two new tables: `user_address_book_items` (per-user private contacts scoped to a space with unique constraint on `space_id, creator_id, address`) and `address_book_requests` (approval workflow with `status SMALLINT` and unique constraint on `space_id, requested_by, address`), each with update triggers and indexes.

_Enables the private contacts and request-to-add workflow by persisting user-scoped contacts and their promotion requests separately from the shared space address book._

**Risks:**
- status stored as SMALLINT with no CHECK constraint or enum type — magic numbers risk; no DB-level enforcement of valid status values (0/1/2)
- UQ_ABR_space_requester_address prevents a member from requesting the same address twice regardless of status — a rejected request cannot be re-submitted without manual DB cleanup
- reviewed_by is nullable VARCHAR(42) with no FK to users — referential integrity for the reviewer is not enforced
- down() drops tables with CASCADE which could silently drop dependent objects if any are added later
- update_updated_at() procedure is assumed to already exist; migration will fail if run before the procedure-creating migration
- chain_ids uses VARCHAR(32) ARRAY — no validation on individual element format at DB level

### `src/modules/spaces/datasources/entities/address-book-request.entity.db.ts`

Adds a new TypeORM entity `AddressBookRequest` mapped to the `address_book_requests` table, with columns for space/user FKs, address, name, chainIds, status (enum-stored-as-integer), requestedByWallet, reviewedBy, and timestamps.

_Defines the persistence model for the member-to-admin request workflow, enabling members to propose contacts for the shared space address book._

**Risks:**
- status is stored as integer via databaseEnumTransformer but the domain type uses string keys — verify round-trip serialization is correct and no off-by-one in enum ordering
- updatedAt has `update: false`, meaning it will never be updated by TypeORM on save — likely a bug if the intent is to track when a request was reviewed/approved/rejected
- Unique constraint UQ_ABR_space_requester_address allows a member to re-request the same address only once; verify whether REJECTED requests should allow re-submission (may need status in the unique key or a delete-on-reject strategy)
- chainIds is varchar(32) array — no validation that individual chain ID strings fit within 32 chars or that the array is non-empty
- reviewedBy stores only the wallet address of the reviewer, not a FK to User — inconsistent with requestedBy pattern and loses referential integrity
- No migration file visible in this diff — confirm the corresponding migration creates the table, indexes, and enum type correctly

### `src/modules/spaces/datasources/entities/user-address-book-item.entity.db.ts`

Adds a new TypeORM entity `UserAddressBookItem` mapped to the `user_address_book_items` table, with a unique constraint on `(space, creator, address)` and foreign keys to `Space` and `User` with CASCADE deletes.

_Defines the persistence layer for per-user private contacts scoped to a space, enabling members to manage their own address book entries independently of the shared space address book._

**Risks:**
- `updatedAt` has `update: false`, meaning it will never be updated on upsert — likely a bug if the column is intended to track modification time
- `createdBy` stores a raw address string (varchar 42) alongside the `creator` FK relation — dual-source of truth that could diverge if not kept in sync
- `chainIds` is a varchar(32) array — 32 chars may be too short for some chain ID strings or future multi-chain scenarios; verify constraint is sufficient
- No `nullable` specified on `name` column — defaults to nullable in TypeORM unless explicitly set, which may differ from intended schema
- CASCADE delete on both Space and User FKs means deleting a user silently removes all their private contacts across all spaces — confirm this is the desired behavior

### `src/modules/spaces/domain/address-books/address-book-items.repository.interface.ts`

Added optional `createdByOverride?: Address` parameter to the `upsertAddressBookItems` method signature in `IAddressBookItemsRepository`, plus added `Address` import from viem and an FSL license header.

_Likely allows the approve-request flow to upsert shared address book items attributed to the original requester's address rather than the admin performing the approval._

**Risks:**
- Check that all implementations of IAddressBookItemsRepository handle the new optional parameter correctly and don't silently ignore it when it matters
- Verify the override address is validated/sanitized before being persisted to prevent spoofing of creator attribution
- Confirm the override is only used in the admin-approval code path and not accidentally exposed to non-admin callers

### `src/modules/spaces/domain/address-books/address-book-items.repository.ts`

Added optional `createdByOverride?: Address` parameter to `upsertAddressBookItems` and the private `_upsertAddressBookItems` method, which substitutes the `createdBy` field instead of using `authPayload.signer_address` when provided.

_Likely allows the request-approval flow to upsert shared address book items attributed to the original requester (member) rather than the approving admin._

**Risks:**
- Verify that `createdByOverride` is validated as a checksummed/valid EVM address before being persisted, since it bypasses the auth-derived address
- Check all callers of `upsertAddressBookItems` to ensure `createdByOverride` is only supplied in the approval path and not accidentally exposed to untrusted input
- Confirm `lastUpdatedBy` intentionally remains the approving admin's address and this asymmetry is desired

### `src/modules/spaces/domain/address-books/address-book-requests.repository.interface.ts`

Adds a new repository interface `IAddressBookRequestsRepository` with methods for CRUD and status-transition operations on `AddressBookRequest` entities, including `findBySpaceId`, `findBySpaceAndRequester`, `findOneOrFail`, `create`, `updateStatus`, `transitionFromPending`, and `revertToPending`.

_Defines the contract for managing address book approval requests, enabling members to propose contacts and admins to approve/reject them._

**Risks:**
- Both `updateStatus` and `transitionFromPending` can update status — overlapping responsibilities may cause confusion or inconsistent state if both are used; clarify which is authoritative
- `transitionFromPending` returns `boolean` (success/failure) while `updateStatus` returns `void` — inconsistent error signaling between two similar operations
- `revertToPending` has no guard on current status — could allow reverting already-approved/rejected requests if not enforced at implementation level
- The `status` parameter in `findBySpaceId` and `findBySpaceAndRequester` uses `keyof typeof AddressBookRequestStatus` but `updateStatus` also uses this type — verify the enum has the expected keys PENDING/APPROVED/REJECTED

### `src/modules/spaces/domain/address-books/address-book-requests.repository.ts`

Adds `AddressBookRequestsRepository` implementing `IAddressBookRequestsRepository` with methods: `findBySpaceId`, `findBySpaceAndRequester`, `findOneOrFail`, `create`, `updateStatus`, `transitionFromPending`, and `revertToPending` — all backed by TypeORM against the `DbAddressBookRequest` entity.

_Provides the data-access layer for the new address-book request workflow, enabling members to submit requests and admins to approve/reject them with atomic status transitions._

**Risks:**
- `updateStatus` does not scope by `spaceId`, allowing any request ID to be updated without verifying it belongs to the target space — unlike `transitionFromPending` which does scope by space
- `revertToPending` is exposed but has no corresponding interface method or service call shown; unclear if it is dead code or a test helper that leaked into production
- `transitionFromPending` returns a boolean but callers must check it to detect race conditions (concurrent approve/reject); verify the service layer handles `false` return
- Unique constraint on `create` catches duplicate requests per address, but there is no check preventing a request for an address already present in `space_address_book_items`
- `result.identifiers[0].id` in `create` will throw unchecked if the insert somehow returns no identifiers
- No transaction wrapping between `insert` and the subsequent `findOneOrFail` in `create`, risking inconsistent reads under concurrent deletes

### `src/modules/spaces/domain/address-books/entities/address-book-request.entity.ts`

Adds a new `AddressBookRequestSchema` Zod entity and `AddressBookRequestStatus` enum (PENDING/APPROVED/REJECTED) defining the shape of address book approval requests, including fields for `space`, `requestedBy`, `requestedByWallet`, `chainIds`, `address`, `name`, `status`, and nullable `reviewedBy`.

_Establishes the domain entity for the request-to-add workflow, allowing members to propose contacts for the shared space address book._

**Risks:**
- status field uses string enum keys ('PENDING','APPROVED','REJECTED') but the numeric enum values (0,1,2) are never used — verify DB column type matches string representation
- reviewedBy stores only the wallet Address, not a User FK — no audit trail linking reviewer to a User entity
- chainIds is Array<string> with no validation on individual chain ID format or non-empty constraint
- No `requestedByWallet` uniqueness or deduplication constraint visible at schema level — duplicate pending requests for same address possible
- z.lazy() used for SpaceSchema and UserSchema but comment says it's for recursion — these aren't recursive, so the lazy wrapping may be unnecessary overhead

### `src/modules/spaces/domain/address-books/entities/user-address-book-item.entity.ts`

Adds a new Zod schema `UserAddressBookItemSchema` and `UserAddressBookItem` type representing a per-user private address book contact scoped to a space, with fields for `space`, `creator`, `createdBy`, `chainIds`, `address`, and `name`.

_Defines the domain entity for the new private contacts feature, enabling members to store contacts visible only to themselves within a space._

**Risks:**
- Both `creator` (User relation) and `createdBy` (Address) exist — verify they are not redundant and understand the distinction between them
- The comment about explicit ZodType due to recursion may be misleading — check if true recursion exists or if this is cargo-culted from another entity
- No validation on `chainIds` array elements beyond `z.string()` — consider whether chain ID format should be validated
- Missing `spaceId`/`creatorId` FK fields that the DB table likely has — confirm ORM mapping handles this or that they are intentionally omitted from the domain entity

### `src/modules/spaces/domain/address-books/user-address-book-items.repository.interface.ts`

Adds a new repository interface `IUserAddressBookItemsRepository` with four methods: `findBySpaceAndCreator`, `findOneBySpaceCreatorAndAddress`, `upsertMany`, and `deleteByAddress`, all scoped by `spaceId` and `creatorId` for per-user private address book items.

_Defines the contract for persisting and retrieving private contacts that are visible only to the creating member within a space._

**Risks:**
- No `findById` method — callers cannot look up a single item by primary key without knowing the address
- upsertMany takes `signerAddress` separately from `creatorId`; unclear if these are validated to match or could diverge
- deleteByAddress returns void with no indication of whether a row was actually deleted (no 404 signaling)
- Interface has no method to delete all private contacts for a user/space (e.g., on membership removal)

### `src/modules/spaces/domain/address-books/user-address-book-items.repository.ts`

Adds `UserAddressBookItemsRepository` implementing `IUserAddressBookItemsRepository` with methods `findBySpaceAndCreator`, `findOneBySpaceCreatorAndAddress`, `upsertMany`, `deleteByAddress`, plus private helpers `updateExisting` and `checkItemsLimit`.

_Provides the data-access layer for per-user private contacts scoped to a space, enforcing a configurable max-items limit per user per space._

**Risks:**
- upsertMany opens a transaction for updates/inserts but then calls repository.findBy outside the transaction — if the transaction fails the outer findBy still executes and returns stale/empty data without error
- checkItemsLimit counts existing rows inside the transaction but after updateExisting already ran; if updateExisting deletes nothing, the count is correct, but the limit check does not account for items being updated (not new) — verify no double-counting edge case
- updateExisting issues one UPDATE per existing item in a loop (N+1 pattern) — could be a performance issue for large batches
- No explicit handling when deleteByAddress finds no matching row — silent no-op may mask bugs if caller passes wrong address
- maxItems config key 'spaces.addressBooks.maxItems' must exist or getOrThrow throws at startup — verify it is defined in all config profiles including test
- address comparison uses isAddressEqual (checksummed) but DB stores addresses as plain strings — ensure DB values are always checksummed or comparison may miss matches

### `src/modules/spaces/routes/address-book-requests.controller.e2e-spec.ts`

Adds a new e2e test file `address-book-requests.controller.e2e-spec.ts` with 7 integration tests covering POST (create request), GET (list requests), PUT approve, and PUT reject endpoints for the `AddressBookRequestsController`, using shared utility functions `createSpace`, `inviteMember`, and `createPrivateContact`.

_Validates the full request-to-add workflow including authorization boundaries (member vs admin), state transitions (PENDING→APPROVED/REJECTED), and side effects (approved contact appears in shared address book, rejected contact stays private)._

**Risks:**
- inviteMember does not call POST /v1/users/wallet for the member before inviting — member wallet may not be registered, which could cause silent failures or false positives if the endpoint doesn't require it
- inviteMember does not accept the invitation (no PATCH/PUT accept step), so the member may still be in INVITED state rather than MEMBER — tests may pass vacuously if membership checks are lenient
- No test for duplicate request submission (same address requested twice by same member) — potential unique constraint violation not covered
- No test verifying that approving a request also updates the request status to APPROVED in the requests list
- createPrivateContact uses PUT which expects 200 but the controller description says it's an upsert — no assertion on the response body to confirm the contact was actually created
- Tests share a single app instance across all describe blocks with no database cleanup between tests — state from one test could bleed into another (e.g., the 'admin sees both' assertion in GET tests assumes exactly 2 requests exist)

### `src/modules/spaces/routes/address-book-requests.controller.ts`

Adds a new NestJS controller `AddressBookRequestsController` with four endpoints under `/v1/spaces/:spaceId/address-book/requests`: GET (list pending), POST (create request), PUT approve, and PUT reject, all guarded by `AuthGuard`.

_Implements the request-to-add workflow allowing non-admin members to propose contacts for the shared space address book while admins approve or reject them._

**Risks:**
- approve/reject endpoints use PUT but have no idempotency guard at the controller level — relies entirely on service layer to enforce PENDING-only transitions
- no rate limiting or duplicate-request prevention visible at controller level for POST /requests
- spaceId from path param is not cross-validated against the requestId in the service call — a user could supply a requestId belonging to a different space
- approve and reject return void with 200 OK but ApiOkResponse has no body type, which may cause inconsistent client handling
- SIWE (wallet authentication) requirement mentioned in Swagger descriptions but not enforced by a dedicated guard — only AuthGuard is applied

### `src/modules/spaces/routes/address-book-requests.service.ts`

Adds `AddressBookRequestsService` with four public methods: `findPending`, `createRequest`, `approve`, and `reject`, orchestrating across `IAddressBookRequestsRepository`, `IUserAddressBookItemsRepository`, `IAddressBookItemsRepository`, `IMembersRepository`, and `ISpacesRepository`.

_Implements the core business logic for the member-to-admin approval workflow, allowing non-admin members to propose private contacts for the shared space address book._

**Risks:**
- Compensation logic in `approve` reverts to PENDING on upsert failure, but if `revertToPending` itself throws, the request is left in APPROVED state with no contact written — no outer error handling or alerting for this split-brain scenario
- `findPending` calls `isAdmin` via `spacesRepository` after already calling `assertMember` via `membersRepository` — two separate DB round-trips that could be collapsed, and the two repositories may have inconsistent views of membership/role
- `createRequest` does not check for a duplicate PENDING request for the same address/space/user, so a member could spam multiple pending requests for the same contact
- `approve` fetches the request with `findOneOrFail` then immediately calls `transitionFromPending` — the intermediate fetch is unused for validation beyond existence, creating a redundant DB call and a TOCTOU window
- `createdByOverride: request.requestedByWallet` passes the original requester's wallet to `upsertMany`; reviewers should verify the `upsertMany` signature accepts and correctly uses this override rather than silently ignoring it
- No transaction wraps the `transitionFromPending` + `upsertMany` pair, so the compensation pattern is manual and fragile under concurrent failures

### `src/modules/spaces/routes/entities/address-book-request.dto.entity.ts`

Adds three new DTO classes and a Zod schema: `AddressBookRequestItemDto` (single request shape with id, name, address, chainIds, requestedBy, status, timestamps), `AddressBookRequestsDto` (wrapper with spaceId and data array), and `CreateAddressBookRequestDto` backed by `CreateAddressBookRequestSchema` validating only an Ethereum address.

_Defines the API contract for the new address-book request workflow, enabling members to propose private contacts for promotion to the shared space address book._

**Risks:**
- CreateAddressBookRequestSchema only validates `address`; name and chainIds are not part of the creation payload, meaning they must be sourced from the existing private contact record — verify this lookup is enforced server-side
- status field uses a plain string union rather than a shared enum or const — risk of drift if the DB enum values change
- AddressBookRequestsDto.spaceId is typed as string but space IDs elsewhere may be numbers — verify consistency
- No @IsNotEmpty or additional Zod refinements on CreateAddressBookRequestDto beyond AddressSchema — confirm AddressSchema is strict enough (checksum, length, etc.)
- AddressBookRequestItemDto has no @Expose/@Transform decorators; confirm serialization/class-transformer pipeline won't leak unexpected fields

### `src/modules/spaces/routes/entities/space-address-book.dto.entity.ts`

Two new DTO classes added to space-address-book.dto.entity.ts: `UserAddressBookItemDto` (with name, address, chainIds, createdBy, createdAt, updatedAt fields) and `UserAddressBookDto` (with spaceId and data array), mirroring the existing `SpaceAddressBookItemDto`/`SpaceAddressBookDto` pattern for private contacts.

_These DTOs define the response shape for the new private address book endpoints scoped to individual users rather than the shared space._

**Risks:**
- UserAddressBookItemDto reuses AddressBookDbItem field types but lacks validation decorators (e.g., @IsString, @IsEthereumAddress) — if used as input DTO elsewhere, no validation is enforced
- chainIds typed as isArray:true with type String but no explicit array validation; inconsistent with how other DTOs may handle this
- createdAt/updatedAt have no explicit type in @ApiProperty() — Swagger may infer incorrectly
- spaceId in UserAddressBookDto is typed as String but no format constraint; consider whether it should be numeric to match space ID conventions used elsewhere

### `src/modules/spaces/routes/user-address-book.controller.e2e-spec.ts`

Adds a new e2e test file `user-address-book.controller.e2e-spec.ts` with 9 tests covering GET/PUT/DELETE on `/v1/spaces/:spaceId/address-book/private`, plus utility helpers `createSpace`, `inviteMember`, and `createPrivateContact`.

_Validates that private address book contacts are scoped per-user (not visible to other members), that both admins and members can create/update/delete their own private contacts, and that unauthenticated access is rejected._

**Risks:**
- inviteMember creates the member JWT but never registers the member user via POST /v1/users/wallet — if user registration is required before invite acceptance, the member token may not be valid for subsequent requests
- inviteMember sends an invite but does not simulate the member accepting it; if acceptance is required to gain MEMBER role, tests using memberAccessToken may silently pass with wrong authorization semantics
- DELETE test only verifies the contact is gone via GET but does not assert the response body of the DELETE itself (e.g. 200 with empty body vs structured response)
- No test covers attempting to delete another user's private contact (cross-user delete authorization gap)
- No test covers unauthenticated PUT or DELETE — only GET 403 is tested
- createPrivateContact uses faker.helpers.multiple with random count (1–5) for chainIds, making assertions order-sensitive if the API doesn't sort them

### `src/modules/spaces/routes/user-address-book.controller.ts`

Adds a new `UserAddressBookController` with three endpoints (`GET`, `PUT`, `DELETE`) under `/v1/spaces/:spaceId/address-book/private` for managing per-user private address book entries scoped to a space.

_Enables non-admin space members to maintain their own private contacts without touching the shared space address book._

**Risks:**
- DELETE returns `void` but is decorated with `@ApiNoContentResponse` — NestJS will return 200 by default, not 204; needs `@HttpCode(HttpStatus.NO_CONTENT)` decorator
- PUT upsert uses `AuthGuard` but PR description says it requires SIWE (wallet auth) — verify the guard enforces SIWE and not just session auth
- GET also uses only `AuthGuard` — confirm member-only access is enforced inside `service.findAll` and not just at the guard level
- No rate limiting or throttling decorators on write endpoints (PUT/DELETE)
- Address param validation uses `AddressSchema` but no checksum normalization is visible — duplicate entries with different casing could be possible depending on DB layer

### `src/modules/spaces/routes/user-address-book.service.ts`

New `UserAddressBookService` class added with `findAll`, `upsertMany`, and `deleteByAddress` methods that manage per-user private address book items scoped to a space, using `IUserAddressBookItemsRepository` and `IMembersRepository`.

_Implements the service layer for private (member-scoped) address book contacts, enforcing SIWE authentication for writes and member-level authorization for all operations._

**Risks:**
- SIWE check is duplicated in `upsertMany` and `deleteByAddress` but not in `findAll` — verify reads intentionally allow non-SIWE auth
- No validation that `address` param in `deleteByAddress` is a valid/checksummed Ethereum address before passing to repository
- `signerAddress` from `authPayload` is passed to repository upsert but its purpose/usage in the repository is unclear — verify it's not silently ignored or misused
- No error handling if `repository.deleteByAddress` deletes 0 rows (non-existent address) — should this return 404?
- `mapItems` spreads all fields manually rather than returning the entity directly — risk of missing new fields added to `UserAddressBookItem` in future

### `src/modules/spaces/routes/utils/space-assert.utils.ts`

Refactored `assertAdmin` by extracting a new `isAdmin` function that returns `boolean` instead of throwing; `assertAdmin` now delegates to `isAdmin` and throws `ForbiddenException` if it returns false.

_Likely enables callers to branch on admin status without catching exceptions, needed for the new request endpoints where admins see all requests and members see only their own._

**Risks:**
- isAdmin returns false when space not found AND when user is not admin — callers cannot distinguish these two cases
- assertAdmin behavior is preserved but now makes two async calls if used alongside isAdmin in the same request — check for redundant DB queries
- No test coverage change visible for the refactored isAdmin path

### `src/modules/spaces/spaces.module.ts`

SpacesModule is extended to register two new entities (UserAddressBookItem, AddressBookRequest), two new repository pairs (UserAddressBookItemsRepository/IUserAddressBookItemsRepository, AddressBookRequestsRepository/IAddressBookRequestsRepository), and two new controller/service pairs (UserAddressBookController/Service, AddressBookRequestsController/Service).

_Wires the new private address book and request-to-add workflow into the NestJS DI container so all new endpoints and data access layers are available within the SpacesModule._

**Risks:**
- Both new interfaces are added to the exports array — verify downstream modules that import SpacesModule actually need these exports, or that exporting them doesn't leak internal implementation details
- UserAddressBookItem and AddressBookRequest entities must have corresponding TypeORM migrations already applied before this module loads, otherwise startup will fail
- Check that forwardRef circular dependencies with AuthModule and UsersModule still resolve correctly with the added providers

---
_Generated by prbrief • model: claude-sonnet-4-6 • 23 LLM calls • $0.37 • 179.2s_

</details>

<sub>Generated locally by <a href="https://github.com/anthropics/claude-code">prbrief</a>. · `safe-global/safe-client-gateway@1fea5f4ffb15`</sub>

migrations/1776556800000-user-address-book-items.ts
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class UserAddressBookItems1776556800000 implements MigrationInterface {
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // Private contacts table
+    await queryRunner.query(
+      `CREATE TABLE user_address_book_items (
+        id SERIAL NOT NULL,
```
vseehausen, line 9: As always, I recommend UUID.

migrations/1776556800000-user-address-book-items.ts
```diff
@@ -0,0 +1,67 @@
...
+        space_id INTEGER NOT NULL,
+        creator_id INTEGER NOT NULL,
+        created_by VARCHAR(42) NOT NULL,
```
LucieFaire, line 12: what is the difference between `creator_id` and `created_by`?

migrations/1776556800000-user-address-book-items.ts
```diff
@@ -0,0 +1,67 @@
...
+        FOR EACH ROW EXECUTE PROCEDURE update_updated_at();`,
+    );
+
+    // Address book requests table
+    await queryRunner.query(
+      `CREATE TABLE address_book_requests (
+        id SERIAL NOT NULL,
+        space_id INTEGER NOT NULL,
+        requested_by INTEGER NOT NULL,
+        requested_by_wallet VARCHAR(42) NOT NULL,
```
LucieFaire, line 37: why do we need both? cant we just map it to a member or a user?

LucieFaire: also what about the email-based users? they wont have wallet address

src/modules/spaces/domain/address-books/user-address-book-items.repository.ts
```diff
@@ -0,0 +1,159 @@
...
+          name: patch.name,
+          chainIds: patch.chainIds,
+        });
+      }
+    }
+
+    return existing.map((item) => item.address);
+  }
+
+  private async checkItemsLimit(args: {
```
vseehausen: This function should be called `assertItemsLimitNotReached` or something with `assert`, when it throws.

vseehausen (approved):
I have a request to change a function name (sorry, but it's worth it), after that LGTM.

src/modules/spaces/domain/address-books/address-book-requests.repository.interface.ts
```diff
@@ -0,0 +1,56 @@
...
+    id: AddressBookRequest['id'];
+    spaceId: Space['id'];
+    toStatus: 'APPROVED' | 'REJECTED';
+    reviewedBy: Address;
+  }): Promise<boolean>;
+
+  revertToPending(args: {
+    id: AddressBookRequest['id'];
+    spaceId: Space['id'];
+  }): Promise<void>;
```
LucieFaire: do we need all three methods? Seems to me we could also just use `updateStatus`, no? Cause right now its actually dead code

src/modules/spaces/domain/address-books/user-address-book-items.repository.ts
katspaugh: Renamed to `assertItemsLimitNotReached` in 15d59555 to match the existing `assert*` convention (`assertAdmin`, `assertMember`).

migrations/1776556800000-user-address-book-items.ts
katspaugh: Kept `SERIAL` here for consistency with the sibling tables in the spaces domain (`space_address_book_items`, `counterfactual_safes` — same migration wave) and because the FKs point at `spaces.id` / `users.id` which are also `SERIAL`. Happy to switch as part of the wider wa-1583 sequential→UUID migration so everything flips together.

src/modules/spaces/domain/address-books/user-address-book-items.repository.ts
```diff
@@ -0,0 +1,159 @@
...
+      const patch = args.items.find((i) =>
+        isAddressEqual(i.address, item.address),
+      );
+      if (patch) {
+        await repository.update(item.id, {
+          name: patch.name,
+          chainIds: patch.chainIds,
+        });
+      }
+    }
```
LucieFaire: maybe we can have Promise.all here as updates are independent

src/modules/spaces/routes/entities/address-book-request.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+  @ApiProperty({ type: String })
+  public address!: string;
+
+  @ApiProperty({ type: String, isArray: true })
+  public chainIds!: Array<string>;
+
+  @ApiProperty({ type: String })
+  public requestedBy!: string;
+
+  @ApiProperty({ enum: ['PENDING', 'APPROVED', 'REJECTED'] })
```
LucieFaire: cant we use the enum directly?

src/modules/spaces/routes/entities/address-book-request.dto.entity.ts
```diff
@@ -0,0 +1,52 @@
...
+});
+
+export class CreateAddressBookRequestDto implements z.infer<
+  typeof CreateAddressBookRequestSchema
+> {
+  @ApiProperty({
+    type: String,
+    description: 'Address of the private contact to request adding to space',
+  })
+  public readonly address!: `0x${string}`;
```
LucieFaire: can be `Address`

migrations/1776556800000-user-address-book-items.ts
LucieFaire: It would be great if we account for wallet/email users from the beginning

src/modules/spaces/domain/address-books/address-book-requests.repository.interface.ts
katspaugh: Good catch — `updateStatus` was dead code (`transitionFromPending` + `revertToPending` cover all flows). Removed in 5e0998d2.

src/modules/spaces/domain/address-books/user-address-book-items.repository.ts
katspaugh: Switched to `Promise.all` in 5e0998d2 — updates are independent row-level writes so they can run in parallel.

src/modules/spaces/routes/entities/address-book-request.dto.entity.ts
katspaugh: Done in 5e0998d2 — now uses `keyof typeof AddressBookRequestStatus` with `getStringEnumKeys` for the Swagger enum, matching how the domain entity exposes it.

src/modules/spaces/routes/entities/address-book-request.dto.entity.ts
katspaugh: Retyped as `Address` from `viem` in 5e0998d2.

migrations/1776556800000-user-address-book-items.ts
katspaugh: Two different concerns:

- `creator_id` (INTEGER → `users.id`) — ownership FK. Powers cascade deletes, the unique constraint `(space_id, creator_id, address)`, and the "who can read/edit this row" authorization check (we scope every query by `creator_id`).
- `created_by` (VARCHAR(42) wallet) — audit/attribution, same field as on `space_address_book_items.created_by`. It's the wallet that SIWE-signed the write, so the UI can show "created by 0x…" consistently across the private and shared address book. It's also what we carry over into `space_address_book_items.created_by` when a request gets approved.

So `creator_id` answers "whose row is it" and `created_by` answers "which wallet authored it". Mirrors the existing shared-book schema.

migrations/1776556800000-user-address-book-items.ts
katspaugh: Waiting for https://github.com/safe-global/safe-client-gateway/pull/3056 to be merged to main for this.

If #3056 lands first:                                                                                 
  - space_address_book_items.created_by becomes integer (user id), with SIWE guards removed             
  - My branch rebases on top and gets a much cleaner diff: drop user_address_book_items.created_by      
  entirely (it's redundant — creator_id already identifies the user), drop                              
  address_book_requests.requested_by_wallet (same — requested_by user id is enough), drop SIWE guards on
   the private endpoints and request creation                                                           
  - My createdByOverride at approval time becomes a user id taken straight from requested_by, no wallet 
  lookup

LucieFaire (approved):

PR #3029, opened 2026-04-17
Revert "fix: drop stale counterfactual_safes table before recreating it"

katspaugh:
Reverts safe-global/safe-client-gateway#3027

### DIFF
--- a/migrations/1775500000000-create-counterfactual-safes.ts
+++ b/migrations/1775500000000-create-counterfactual-safes.ts
@@ -3,12 +3,6 @@ import type { MigrationInterface, QueryRunner } from 'typeorm';
 
 export class CreateCounterfactualSafes1775500000000 implements MigrationInterface {
   public async up(queryRunner: QueryRunner): Promise<void> {
-    // Drop the orphaned table left over from the deprecated Knex "accounts"
-    // module (removed in #2804). That schema is incompatible with the one
-    // below and its FK target (accounts) no longer exists.
-    await queryRunner.query(
-      `DROP TABLE IF EXISTS counterfactual_safes CASCADE;`,
-    );
     await queryRunner.query(
       `CREATE TABLE counterfactual_safes (
         id SERIAL NOT NULL,


0xAryanShah (approved):

katspaugh:
[Open the PR briefing](https://viewer.popcorn.report/view?briefing_id=pb_d3be832f0c9252125dd244e0)

**This PR reverts #3027, which had added a `DROP TABLE IF EXISTS counterfactual_safes CASCADE` statement to the `up` method of the `CreateCounterfactualSafes1775500000000` TypeORM migration.** The revert removes that drop statement, restoring the migration to its original form that only creates the table. The motivation for the revert is not explicitly stated in the PR but likely relates to unintended side effects such as data loss or failures in environments that do not have the legacy stale table. The underlying problem—an orphaned `counterfactual_safes` table with an incompatible schema from the old Knex module blocking the migration—remains unresolved by this revert.

▶ **[Watch trailer](https://eyfiffagztwbokkxesiq.supabase.co/storage/v1/object/public/published-assets/safe-global/safe-client-gateway/pr-3029/12c3c4f10eb3c60feae37167ba30719ef7b0a18e/video.mp4)** · 📺 **[Open interactive briefing](https://viewer.popcorn.report/view?briefing_id=pb_d3be832f0c9252125dd244e0)** · 📋 **Review checklist below**

### What's worth a closer look

| Signal | Reviewer attention |
|---|---|
| 🟡 Attention | Confirm that all deployment environments (production, staging, CI) do not have a stale `counterfactual_safes` table that would cause the `CREATE TABLE` statement to fail. |
| 🟡 Attention | Verify that a follow-up migration or manual remediation plan exists to handle environments where the legacy Knex-managed table may still be present. |
| 🟡 Attention | Check that the revert does not leave any partially applied state from PR #3027 in environments where that migration had already run. |

<details>
<summary><b>Full briefing — architecture, contracts, file map</b></summary>

## 🍿 Summary

This PR reverts #3027, which had added a `DROP TABLE IF EXISTS counterfactual_safes CASCADE` statement to the `up` method of the `CreateCounterfactualSafes1775500000000` TypeORM migration. The revert removes that drop statement, restoring the migration to its original form that only creates the table. The motivation for the revert is not explicitly stated in the PR but likely relates to unintended side effects such as data loss or failures in environments that do not have the legacy stale table. The underlying problem—an orphaned `counterfactual_safes` table with an incompatible schema from the old Knex module blocking the migration—remains unresolved by this revert.

## 🏗️ Architecture notes

The change touches a single TypeORM migration file (`migrations/1775500000000-create-counterfactual-safes.ts`) within the safe-client-gateway service. The migration's `up` method no longer defensively drops the legacy table before creating it, meaning the `CREATE TABLE counterfactual_safes` statement will fail with a 'relation already exists' error if the old Knex-managed table is still present in the database. No other subsystems, data flows, or API contracts are directly affected by this revert.

## 🌊 Ripple risks

Analyzed 1 changed export. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 1 file(s), 1 changed export(s), 0 uncovered callsite(s) in 0.8s._

## 🎬 Walkthrough outline

_10 steps across 4 chapters, approximately 3 min 35s._

### Why this revert happened

1. **🎯 motivation** · `migrations/1775500000000-create-counterfactual-safes.ts` L1-5 `CreateCounterfactualSafes1775500000000`
   We are looking at a single-file revert. PR 3027 added a defensive drop-table statement to this TypeORM migration, and now PR 3029 is pulling that back out. The most likely reason is that the drop caused unintended data loss or broke environments that never had the legacy table in the first place.

2. **🎯 motivation** · `migrations/1775500000000-create-counterfactual-safes.ts` L1-5 `CreateCounterfactualSafes1775500000000`
   A bit of history: the old Knex module left behind an orphaned counterfactual safes table with an incompatible schema. PR 3027 tried to clear it by dropping the table before recreating it. That fix worked for some environments but introduced new problems elsewhere, so the team is reverting while they figure out a cleaner solution.

### The single line that changed

3. **🔧 change** · `migrations/1775500000000-create-counterfactual-safes.ts` L4-10 `CreateCounterfactualSafes1775500000000`
   Here is the class declaration for the migration. The timestamp in the name, 1775500000000, pins this as a specific point in the migration sequence. Nothing about the class signature changed; the revert only touches what is inside the up method.

4. **🔧 change** · `migrations/1775500000000-create-counterfactual-safes.ts` L10-20 `CreateCounterfactualSafes1775500000000`
   This is the up method, the part that runs when the migration is applied forward. In PR 3027 the very first line here was a DROP TABLE IF EXISTS counterfactual safes CASCADE call. That line is now gone, and the method goes straight to creating the table.

5. **🔧 change** · `migrations/1775500000000-create-counterfactual-safes.ts` L20-35 `CreateCounterfactualSafes1775500000000`
   The rest of the up method, the actual CREATE TABLE block with all its columns and constraints, is completely untouched. This is a surgical revert of exactly one statement, nothing more.

6. **🔧 change** · `migrations/1775500000000-create-counterfactual-safes.ts` L35-50 `CreateCounterfactualSafes1775500000000`
   The down method, which rolls the migration back by dropping the table, is also unchanged. So rollback behavior is identical to what it was before PR 3027 ever landed.

### What breaks if the old table is still there

7. **💥 risk** · `migrations/1775500000000-create-counterfactual-safes.ts` L10-20 `CreateCounterfactualSafes1775500000000`
   The core risk is straightforward: any environment that still has the old Knex-managed counterfactual safes table will now hit a relation already exists error the moment this migration runs. The defensive drop is gone, so there is no safety net.

8. **💥 risk** · `migrations/1775500000000-create-counterfactual-safes.ts` L10-20 `CreateCounterfactualSafes1775500000000`
   The original problem that PR 3027 was solving is now unresolved. If the team needs to handle the stale table, the right path is probably a separate migration that runs before this one, or a manual remediation step in affected environments, rather than embedding a destructive drop inside this migration.

9. **💥 risk** · `migrations/1775500000000-create-counterfactual-safes.ts` L10-20 `CreateCounterfactualSafes1775500000000`
   Before merging, reviewers should confirm that CI, staging, and any production-like environments do not have the legacy table present. If they do, this revert will cause a deployment failure rather than a data-loss incident, which may be the lesser evil, but it still needs a follow-up plan.

### Bottom line

10. **🍿 wrap** · `migrations/1775500000000-create-counterfactual-safes.ts` L4-50 `CreateCounterfactualSafes1775500000000`
   In short, this PR removes a risky drop-before-create from the counterfactual safes migration, trading the danger of accidental data loss for the danger of a migration failure in environments with a legacy table, and the team still needs a proper long-term fix for that orphaned schema.

## ✅ Reviewer checklist

- [ ] Confirm that all deployment environments (production, staging, CI) do not have a stale `counterfactual_safes` table that would cause the `CREATE TABLE` statement to fail.
- [ ] Verify that a follow-up migration or manual remediation plan exists to handle environments where the legacy Knex-managed table may still be present.
- [ ] Check that the revert does not leave any partially applied state from PR #3027 in environments where that migration had already run.
- [ ] Ensure the migration's `down` method is still consistent and correctly drops the table without unintended side effects.
- [ ] Confirm the reason the original fix (#3027) was problematic is documented, either in this PR description or a linked issue, to prevent re-introducing the same approach.

## 🗂️ File-by-file

### `migrations/1775500000000-create-counterfactual-safes.ts`

Removed the `DROP TABLE IF EXISTS counterfactual_safes CASCADE` statement from the `up` migration in `CreateCounterfactualSafes1775500000000`, reverting PR #3027's fix that dropped an orphaned legacy table before recreating it.

_Likely reverted because the drop-before-create approach caused issues (e.g., data loss, migration failures in environments without the stale table, or unintended side effects from CASCADE)._

**💥 Risks:**
- If any environment still has the old incompatible `counterfactual_safes` table from the Knex module, the `CREATE TABLE` will now fail with a 'relation already exists' error
- Reverting means the original problem (orphaned table with incompatible schema blocking migration) is unresolved for affected deployments
- Check whether a separate migration or manual remediation step is planned to handle the stale table scenario
- Verify that CI/CD and staging environments do not have the legacy table present before merging

---
_Generated by prbrief • model: claude-sonnet-4-6 • 3 LLM calls • $0.06 • 45.8s_

</details>

<sub>Generated by <a href="https://popcorn.report">Popcorn Review</a> · `safe-global/safe-client-gateway@12c3c4f10eb3`</sub>

katspaugh:
@PooyaRaki do you want this merged or should I close it?

PR #3030, opened 2026-04-17
feat: migrate from ESLint to Biome 2.0

PooyaRaki:
## What it solves

Replaces both ESLint and Prettier with Biome 2.4.12 as the unified linting and formatting tool for faster, single-tool code quality checks.

Resolves: https://linear.app/safe-global/issue/WA-1985/migrate-from-eslint-to-biome-in-safe-client-gateway

## How this PR fixes it

- Removes `eslint`, `eslint-config-prettier`, `typescript-eslint`, and `prettier` dependencies entirely
- Removes `.prettierrc`, `.prettierignore`, and `eslint.config.mjs`
- Adds `@biomejs/biome` 2.4.12 with a `biome.json` covering both linting and formatting:
  - **Formatting:** single quotes, trailing commas, 2-space indent, 80-char line width (matching previous Prettier settings)
  - **Linting:** rules across `correctness`, `suspicious`, `complexity`, `style`, `performance`, and selected `nursery` groups — chosen to mirror the previous `typescript-eslint` recommended-type-checked coverage (e.g. `noExplicitAny`, `noFloatingPromises`, `noMisusedPromises`, `useImportType`-equivalent behavior, restricted-imports for `@safe-global/safe-deployments`)
  - **Overrides:** test files (`*.spec.ts`, `*.integration.spec.ts`, `*.e2e-spec.ts`, `__tests__/**`) relax the strictest rules (`noExplicitAny`, `noConsole`, `noNonNullAssertion`, cognitive complexity, etc.); a few legacy modules opt out of `noStaticOnlyClass` / `noExcessiveCognitiveComplexity`
  - **`useImportType` is intentionally `off`** — NestJS dependency injection relies on runtime metadata emitted from `import` statements (e.g. `@Inject()`, constructor parameter types). Biome's `useImportType` would rewrite those to `import type`, which strips the runtime metadata and breaks DI. The previous ESLint `@typescript-eslint/consistent-type-imports` rule had the same caveat in NestJS projects.
- Updates `package.json` scripts: `lint` / `lint-check` use `biome check`, `format` / `format-check` use `biome format`
- Merges the separate `prettier` and `es-lint` CI jobs into a single `biome` job that runs both `format-check` and `lint-check`
- Adds SPDX license headers to all modified files (including spec files merged in from main during rebases)

**Note:** The "Contribution Rule" ruleset needs updating — replace `es-lint` and `prettier` with `biome` in required status checks.

## How to test it

1. Run `yarn format-check` — should pass with no errors
2. Run `yarn lint-check` — should pass with no errors
3. Run `yarn lint` and `yarn format` — should apply fixes and exit cleanly
4. Verify CI pipeline passes (the single `biome` job replaces both `es-lint` and `prettier`)

## Screenshots

N/A - no UI changes

## Checklist

- [ ] I've tested the branch on mobile 📱
- [ ] I've documented how it affects the analytics (if at all) 📊
- [ ] I've written a unit/e2e test for it (if applicable) 🧑‍💻

---

## CLA signature

With the submission of this Pull Request, I confirm that I have read and agree to the terms of the [Contributor License Agreement](https://safe.global/cla).


PooyaRaki:
@claude review

PooyaRaki:
@claude review

biome.json
```diff
@@ -0,0 +1,184 @@
...
+        "noUnusedImports": "error",
+        "noUnusedFunctionParameters": "error",
+        "noUnusedPrivateClassMembers": "error",
+        "noInvalidUseBeforeDeclaration": "error",
+        "noUndeclaredVariables": "error",
+        "noVoidTypeReturn": "error",
+        "noConstantCondition": "error",
+        "noSelfAssign": "error"
+      },
+      "suspicious": {
```
LucieFaire, line 45: could we also add `no-deprecated` analog?
https://biomejs.dev/linter/rules/no-deprecated-imports/

biome.json
```diff
@@ -0,0 +1,184 @@
...
+        "noUselessThisAlias": "error",
+        "useSimplifiedLogicExpression": "error",
+        "useFlatMap": "error",
+        "useOptionalChain": "error",
+        "noArguments": "error",
+        "noCommaOperator": "error",
+        "useNumericLiterals": "error"
+      },
+      "style": {
+        "useImportType": "off",
```
LucieFaire, line 109: why is this one off? for eslint we had it as `error`

biome.json
```diff
@@ -0,0 +1,184 @@
...
+        "noUselessLoneBlockStatements": "error",
+        "noUselessThisAlias": "error",
+        "useSimplifiedLogicExpression": "error",
+        "useFlatMap": "error",
+        "useOptionalChain": "error",
+        "noArguments": "error",
+        "noCommaOperator": "error",
+        "useNumericLiterals": "error"
+      },
+      "style": {
```
LucieFaire, line 100: shall we also add 'no-restricted-imports` as in esLint?
https://biomejs.dev/linter/rules/no-restricted-imports/

biome.json
PooyaRaki: This is a known caveat with Biome’s useImportType rule in NestJS projects. Even when an imported class looks like it is only used as a type, NestJS may still need it as a runtime value because dependency injection depends on emitted TypeScript metadata. Using import type in those cases can break DI.
The other option would be to enable the rule but ignore it in all those cases which make us end up with a ton of Biome ignores.

biome.json
PooyaRaki: Added https://github.com/safe-global/safe-client-gateway/pull/3030/changes/bd53411cf7e0e488b794cc376e9dc22b9da45827#diff-2bc8a1f5e9380d5a187a4e90f11b4dd36c3abad6aea44c84be354a4f44cdec55
It was missing in version 2.0.0 of the library, I upgraded to the latest version.

biome.json
PooyaRaki: Added https://github.com/safe-global/safe-client-gateway/pull/3030/changes/bd53411cf7e0e488b794cc376e9dc22b9da45827#diff-2bc8a1f5e9380d5a187a4e90f11b4dd36c3abad6aea44c84be354a4f44cdec55

LucieFaire (approved):

PooyaRaki:
@claude review

PooyaRaki:
Thanks for the 3rd-pass review. Responses inline:

**1. `speckit.implement.md` ESLint/Prettier references — declining.**
That file is a vendored generic spec-kit slash-command template (added in `61198e86 chore: set up spec-kit`), not project-specific config. The detection block is a generic catalog so the same command works for any project — ESLint, Prettier, Biome, or otherwise. Removing ESLint/Prettier detection would degrade the template for any non-Biome project where this slash command is used; the upstream-correct fix would be to *add* Biome detection alongside, but that's out of scope for this migration PR.

**2. `parseRelayRules` discarded validation — out of scope, will track separately.**
Confirmed the bug: `parsed.every(...)` returns a boolean that's discarded, so invalid rules pass silently. It originates from PR #2773. The function body wasn't actually refactored in this PR — the only changes to `configuration.ts` are lint-driven (`parseInt` → `Number.parseInt` for `useNumberNamespace`, and removing one stale `eslint-disable` comment). Filing as a separate ticket rather than expanding this PR's blast radius.

**3. `useImportType: "off"` rationale — done.**
Added an explicit note to the PR description explaining the NestJS DI caveat (DI relies on runtime metadata emitted from `import` statements; `import type` strips it and breaks `@Inject()` / constructor injection). Same caveat applied to the previous `@typescript-eslint/consistent-type-imports` rule. `biome.json` doesn't carry comments in our setup, so the PR description is the durable place for it.

**4. `maxAllowedComplexity: 35` tracking issue — declining.**
The previous ESLint config had **no** cognitive-complexity rule at all, so 35 in this PR is a *new* constraint, not a relaxed one. A tracker to ratchet down further would be speculative scope (YAGNI) — if a future contributor wants to tighten it, they'll surface natural targets via violations during refactoring.

Net: 3 done (description), 1, 2, 4 declined with reasoning above. Happy to revisit any of these if the rationale doesn't land.

src/modules/alerts/routes/alerts.controller.integration.spec.ts
```diff
@@ -173,10 +174,10 @@ describe('Alerts Controller', () => {
             .build();
           const timestamp = Date.now().toString();
           // TODO: Check value of threshold and owner in email
-          // eslint-disable-next-line @typescript-eslint/no-unused-vars
+          // biome-ignore lint/correctness/noUnusedVariables: intentionally unused in test
```
vseehausen, line 177: In a refactor task, we could check if those lines are still required, or if we can have a lintable version of this code

src/modules/notifications/domain/v2/entities/notification.entity.ts
```diff
@@ -9,11 +9,11 @@ import type { PendingTransactionEvent } from '@/modules/hooks/routes/entities/sc
 
 export enum NotificationType {
   CONFIRMATION_REQUEST = 'CONFIRMATION_REQUEST', // TransactionEventType.PENDING_MULTISIG_TRANSACTION
-  DELETED_MULTISIG_TRANSACTION = TransactionEventType.DELETED_MULTISIG_TRANSACTION,
-  EXECUTED_MULTISIG_TRANSACTION = TransactionEventType.EXECUTED_MULTISIG_TRANSACTION,
-  INCOMING_ETHER = TransactionEventType.INCOMING_ETHER,
-  INCOMING_TOKEN = TransactionEventType.INCOMING_TOKEN,
-  MODULE_TRANSACTION = TransactionEventType.MODULE_TRANSACTION,
+  DELETED_MULTISIG_TRANSACTION = 'DELETED_MULTISIG_TRANSACTION',
```
vseehausen: Noting we are moving from type to string here.

vseehausen (approved):
Nice! Great to have the speed performance. The new code style is more clean, and more readable. There were quite some file changes that only comprise reorders of imports, but fine. I didn't find any anomalies or logic changes when scimming the changes.

src/modules/notifications/domain/v2/entities/notification.entity.ts
PooyaRaki: Restored in ca0c6ec36a96e47a1fff8c3ca5bdd184c65f9b4c — the five members shared with `TransactionEventType` (`DELETED_MULTISIG_TRANSACTION`, `EXECUTED_MULTISIG_TRANSACTION`, `INCOMING_ETHER`, `INCOMING_TOKEN`, `MODULE_TRANSACTION`) reference the source enum again, with a file-level `biome-ignore-all` for `lint/style/useLiteralEnumMembers` so future Biome runs don't undo it.

`CONFIRMATION_REQUEST` and `MESSAGE_CONFIRMATION_REQUEST` are intentionally left as string literals — their values differ from the corresponding `TransactionEventType` members (`PENDING_MULTISIG_TRANSACTION` and `MESSAGE_CREATED`), so they were already plain strings before this PR. The trailing comments document the conceptual mapping.

src/modules/alerts/routes/alerts.controller.integration.spec.ts
PooyaRaki: Out of scope for this PR — leaving as a refactor follow-up.

vseehausen (approved):

PR #3031, opened 2026-04-20
fix: address book response type

0xAryanShah:
## Summary                                                                                                                                        
- Fix Swagger typing for `createdAt` and `updatedAt` in address book endpoint — was rendering as `object` (`{}`) instead of `string` (`date-time`)
- Root cause: `@ApiProperty()` without explicit `type` can't resolve lookup types (`AddressBookDbItem['createdAt']`) via reflection, falling back to `Object`                                                                                                                                       
                                                                                                                                                    
  ## Test plan                                                                                                                                      
  - [ ] Verify Swagger schema shows `"type": "string", "format": "date-time"` for `createdAt`/`updatedAt` in address book response
  - [x] Confirm autogenerated types from CGW match actual API response


LucieFaire (approved):

PooyaRaki (approved):

vmosharova (approved):
Nice! Thank you!

PR #3033, opened 2026-04-20
fix: update relay method call to include only required parameters

akshay-ap:
This pull request makes a minor update to the `RelayFeeRelayer` class, specifically how the `relay` method is called. The change ensures that only the required properties (`chainId`, `to`, and `data`) from `args` are passed to the `relay` API, rather than the entire `args` object.

The change was suggested here: https://github.com/safe-global/safe-client-gateway/pull/3019#discussion_r3099857329


LucieFaire (approved):
Could you make sure there is a test case that checks that only required data is provided? Thank you

akshay-ap:
@LucieFaire can you please approve again?

LucieFaire (approved):

PR #3035, opened 2026-04-21
feat(authentication): capture and expose authenticated user's email

vseehausen:

This PR implements the first backend slice of email identity visibility for authenticated users. It stores a verified OIDC email once in CGW and exposes the stored value on `/v1/auth/me`, so clients can render the authenticated identity without expanding the CGW JWT payload.

## Summary
- widen Auth0 scope and token parsing to accept verified email claims
- persist verified OIDC email on first login with write-once semantics and duplicate-email conflict handling
- expose stored email on `/v1/auth/me` without putting email into the CGW JWT
- add focused test coverage for verified, unverified, and missing-email cases

## Notes
- Email is not added to the CGW JWT. `/me` reads persisted email from the database for OIDC sessions.
- HTTP integration specs are not runnable in this sandbox because `supertest` hits `listen EPERM: operation not permitted 0.0.0.0`.
- The existing Postgres-backed `users.repository.integration.spec.ts` harness also failed here before the new assertions ran, with `EntityMetadataNotFoundError: No metadata for "Wallet" was found`.

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -59,7 +59,7 @@ export default () => ({
       redirectUri: process.env.AUTH0_REDIRECT_URI,
       audience: process.env.AUTH0_API_AUDIENCE,
       signingSecret: process.env.AUTH0_SIGNING_SECRET,
-      scope: process.env.AUTH0_SCOPE || 'openid',
+      scope: process.env.AUTH0_SCOPE || 'openid email profile',
     },
     rateLimit: {
       max: parseInt(process.env.AUTH_RATE_LIMIT_MAX ?? `${5}`),
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -59,7 +59,7 @@ export default () => ({
       redirectUri: process.env.AUTH0_REDIRECT_URI,
       audience: process.env.AUTH0_API_AUDIENCE,
       signingSecret: process.env.AUTH0_SIGNING_SECRET,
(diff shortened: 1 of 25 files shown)


src/config/entities/configuration.ts
```diff
@@ -59,7 +59,7 @@ export default () => ({
       redirectUri: process.env.AUTH0_REDIRECT_URI,
       audience: process.env.AUTH0_API_AUDIENCE,
       signingSecret: process.env.AUTH0_SIGNING_SECRET,
-      scope: process.env.AUTH0_SCOPE || 'openid',
+      scope: process.env.AUTH0_SCOPE || 'openid email profile',
```
0xAryanShah: QQ: do we really require the profile scope??

src/modules/users/datasources/entities/users.entity.db.ts
```diff
@@ -39,6 +39,15 @@ export class User implements DomainUser {
   })
   extUserId!: string | null;
 
+  @Index('users_email_key', { unique: true })
```
0xAryanShah:
```suggestion
  @Index('users_email_key', {
    unique: true,
    where: '"email" IS NOT NULL',
  })
```

src/modules/auth/routes/auth.service.spec.ts
```diff
@@ -284,6 +289,55 @@ describe('AuthService', () => {
     });
   });
 
+  describe('getUserSession', () => {
```
0xAryanShah, line 292: Add test for : should reject with ForbiddenException.

PR #3036, opened 2026-04-22
feat(authentication): capture and expose authenticated user's email

vseehausen:

This PR implements the first backend slice of email identity visibility for authenticated users. It stores a verified OIDC email once in CGW and exposes the stored value on `/v1/auth/me`, so clients can render the authenticated identity without expanding the CGW JWT payload.

## Summary
- widen Auth0 scope and token parsing to accept verified email claims
- persist verified OIDC email on first login with write-once semantics and duplicate-email conflict handling
- expose stored email on `/v1/auth/me` without putting email into the CGW JWT
- add focused test coverage for verified, unverified, and missing-email cases


src/config/entities/configuration.spec.ts
```diff
@@ -0,0 +1,18 @@
...
+  afterEach(() => {
+    process.env.AUTH0_SCOPE = originalAuth0Scope;
+  });
+
+  it('should default AUTH0_SCOPE to openid email', () => {
+    delete process.env.AUTH0_SCOPE;
+
+    const result = configuration();
+
+    expect(result.auth.auth0.scope).toBe('openid email');
```
LucieFaire: we dont need this test, its an optional value, lets remove it. Also for future ref we keep config tests in `src/config/configuration.validator.spec.ts`

src/config/entities/configuration.ts
```diff
@@ -59,7 +59,7 @@ export default () => ({
       redirectUri: process.env.AUTH0_REDIRECT_URI,
       audience: process.env.AUTH0_API_AUDIENCE,
       signingSecret: process.env.AUTH0_SIGNING_SECRET,
-      scope: process.env.AUTH0_SCOPE || 'openid',
+      scope: process.env.AUTH0_SCOPE || 'openid email',
```
LucieFaire: lets not update the default as this should be provided on demand in prod/staging or for local dev of auth

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,50 @@
...
+} from '@/datasources/network/network.service.interface';
 import { IJwtService } from '@/datasources/jwt/jwt.service.interface';
 import { ILoggingService, LoggingService } from '@/logging/logging.interface';
 import type { Auth0Token } from '@/modules/auth/oidc/auth0/domain/entities/auth0-token.entity';
 import { Auth0TokenSchema } from '@/modules/auth/oidc/auth0/domain/entities/auth0-token.entity';
 import { Inject, Injectable, UnauthorizedException } from '@nestjs/common';
 import { JsonWebTokenError } from 'jsonwebtoken';
+import { z } from 'zod';
+
+const AUTH0_ID_TOKEN_ALGORITHM = 'RS256' as const;
```
LucieFaire: we have a dedicated file with algorithms, lets move it there

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,50 @@
...
+  n: z.string().min(1).optional(),
+  e: z.string().min(1).optional(),
+  x5c: z.array(z.string().min(1)).optional(),
+});
+
+const Auth0JwksSchema = z.object({
+  keys: z.array(Auth0JwkSchema),
+});
+
+type Auth0Jwk = z.infer<typeof Auth0JwkSchema>;
```
LucieFaire: please put it to a separate `entity` file as done for other schemas

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,50 @@
...
 
 @Injectable()
 export class Auth0TokenVerifier {
```
LucieFaire, line 16: now that we have logic in place to deal with `id_token` lets remove logic with `access_token` as its redundant to check both. Also auth0 suggests to validate and use `id_token` for our use case, so that would be more correct.

src/modules/auth/oidc/auth0/domain/auth0.repository.spec.ts
```diff
@@ -20,7 +21,12 @@ describe('Auth0Repository', () => {
   beforeEach(() => {
     jest.resetAllMocks();
 
-    target = new Auth0Repository(auth0ApiMock, auth0TokenVerifierMock);
+    target = new Auth0Repository(auth0ApiMock, auth0TokenVerifierMock, {
+      info: jest.fn(),
+      warn: jest.fn(),
+      error: jest.fn(),
+      debug: jest.fn(),
```
LucieFaire: lets create a proper mock for logging service

src/modules/auth/oidc/auth0/domain/auth0.repository.ts
```diff
@@ -22,7 +28,64 @@ export class Auth0Repository implements IAuth0Repository {
     code: string,
   ): Promise<Auth0Token> {
     const response = await this.auth0Api.exchangeAuthorizationCode(code);
-    const { access_token } = Auth0TokenResponseSchema.parse(response);
-    return this.auth0TokenVerifier.verifyAndDecode(access_token);
+    const { access_token, id_token } = Auth0TokenResponseSchema.parse(response);
```
LucieFaire: lets not deal with `access_token` anymore

LucieFaire (commented):
initial review , havent gone through everything :)

LucieFaire:
@claude review

vseehausen:
@claude review

src/datasources/jwt/jwt.constants.ts
```diff
@@ -2,3 +2,4 @@
 import type { Algorithm } from 'jsonwebtoken';
 
 export const JWT_ALGORITHM: Algorithm = 'HS256';
+export const AUTH0_ID_TOKEN_ALGORITHM: Algorithm = 'RS256';
```
LucieFaire: Lets rename them to be 
`JWT_HS_ALGORITHM` and `JWT_RS_ALGORITHM` as they are not strict to auth0 token or anything else, just names

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,52 @@
...
+  Auth0TokenHeader,
+} from '@/modules/auth/oidc/auth0/domain/entities/auth0-token.entity';
+import {
+  Auth0TokenHeaderSchema,
+  Auth0TokenSchema,
+} from '@/modules/auth/oidc/auth0/domain/entities/auth0-token.entity';
 import { Inject, Injectable, UnauthorizedException } from '@nestjs/common';
 import { JsonWebTokenError } from 'jsonwebtoken';
 
+const AUTH0_ID_TOKEN_SIGNING_KEY_CACHE_TTL_MS = 60 * 60 * 1_000;
```
LucieFaire: lets move it to configuration  and env so we can adjust it if needed

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,52 @@
...
+
+type CachedSigningKey = {
+  publicKey: string;
+  expiresAt: number;
+};
+
+type ResolvedSigningKey = {
+  publicKey: string;
+  fromCache: boolean;
+};
```
LucieFaire: why do you need 2 types for the same fields?

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,52 @@
...
+
 @Injectable()
 export class Auth0TokenVerifier {
   private readonly issuer: string;
-  private readonly audience: string;
-  private readonly signingSecret: string;
+  private readonly idTokenAudience: string;
```
LucieFaire: why change the name? We only have 1 audience so lets keep it as it was, no?

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,52 @@
...
+  private readonly idTokenSigningKeys = new Map<string, CachedSigningKey>();
```
LucieFaire: nit: just `signingKeys` ?

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -25,31 +55,163 @@ export class Auth0TokenVerifier {
...
-    );
-    this.signingSecret = this.configurationService.getOrThrow<string>(
-      'auth.auth0.signingSecret',
+    this.idTokenAudience = this.configurationService.getOrThrow<string>(
+      'auth.auth0.clientId',
     );
   }
 
-  verifyAndDecode(accessToken: string): Auth0Token {
+  public async verifyAndDecodeIdToken(idToken: string): Promise<Auth0Token> {
```
LucieFaire: nit, I would keep the name of the func as it was, we only deal with one token anyway

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -25,31 +55,163 @@ export class Auth0TokenVerifier {
...
+    if (cachedKey && cachedKey.expiresAt > Date.now()) {
+      this.loggingService.debug(
+        `Auth0: using cached JWKS signing key for kid ${kid}`,
+      );
+      return { publicKey: cachedKey.publicKey, fromCache: true };
+    }
+
+    if (cachedKey) {
+      this.idTokenSigningKeys.delete(kid);
+    }
```
LucieFaire: The explicit cache eviction is redundant — `refreshIdTokenSigningPublicKey` overwrites the entry via .set() anyway. Can be removed

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -25,31 +55,163 @@ export class Auth0TokenVerifier {
...
+    if (alg !== AUTH0_ID_TOKEN_ALGORITHM) {
+      throw new UnauthorizedException('Invalid id token');
+    }
+  }
+
+  private verifyIdTokenWithSigningKey(
+    idToken: string,
+    signingPublicKey: string,
+  ): Auth0Token {
+    const decoded = this.jwtService.decode<{ sub: string }>(idToken, {
```
LucieFaire: nit: i'd be explicit and pass <Auth0Token> to `decode` instead of only `{sub: string}` as now we expect more claims returned

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,22 +1,52 @@
...
+  expiresAt: number;
+};
+
+type ResolvedSigningKey = {
+  publicKey: string;
+  fromCache: boolean;
+};
+
 @Injectable()
 export class Auth0TokenVerifier {
```
LucieFaire, line 16: lets add JSDoc comments for the methods in this class

LucieFaire: I would also rename the methods, as they are a bit too verbose IMHO:

Current | Suggested
-- | --
verifyAndDecodeIdToken | verifyAndDecode
getIdTokenSigningPublicKey | getSigningKey
refreshIdTokenSigningPublicKey | refreshSigningKey
assertIdTokenAlgorithm | assertAlgorithm
verifyIdTokenWithSigningKey | verifyWithSigningKey
decodeIdTokenHeader | decodeHeader
fetchJsonWebKeySet | fine as-is
toPemPublicKey | toPem

Same for the fields — `idTokenSigningKeys` → `signingKeys`, `idTokenAudience` → `audience`.

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -25,31 +55,163 @@ export class Auth0TokenVerifier {
...
+            throw new UnauthorizedException('Invalid id token');
+          }
+
+          throw retryError;
+        }
       }
+
       throw error;
     }
   }
```
LucieFaire: I would rewrite it to not have nested checks:
```
  public async verifyAndDecodeIdToken(idToken: string): Promise<Auth0Token> {
    const header = this.decodeIdTokenHeader(idToken);
    const signingKey = await this.getIdTokenSigningPublicKey(header);

    try {
      return this.verifyIdTokenWithSigningKey(idToken, signingKey.publicKey);
    } catch (error) {
      if (!(error instanceof JsonWebTokenError)) {
        throw error;
      }
      this.loggingService.debug(
        `Auth0: id token JWT verification failed: ${error.message}`,
      );
      if (!signingKey.fromCache) {
        throw new UnauthorizedException('Invalid id token');
      }
    }

    // Retry with refreshed JWKS (only reached when cached key failed)
    const refreshedKey = await this.refreshIdTokenSigningPublicKey(header);

    try {
      return this.verifyIdTokenWithSigningKey(idToken, refreshedKey);
    } catch (error) {
      if (error instanceof JsonWebTokenError) {
        this.loggingService.debug(
          `Auth0: id token JWT verification failed after JWKS refresh: ${error.message}`,
        );
        throw new UnauthorizedException('Invalid id token');
      }
      throw error;
    }
  }
```

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -25,31 +55,163 @@ export class Auth0TokenVerifier {
...
+      return Auth0TokenHeaderSchema.parse(
+        JSON.parse(Buffer.from(encodedHeader, 'base64url').toString('utf8')),
+      );
+    } catch {
+      throw new UnauthorizedException('Invalid id token');
+    }
+  }
+
+  private async fetchJsonWebKeySet(): Promise<Auth0Jwks> {
+    const response = await this.networkService.get<unknown>({
```
LucieFaire: we should not use `unknown` and please move the URL into a static const

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -25,31 +55,163 @@ export class Auth0TokenVerifier {
...
+      })
+        .export({ format: 'pem', type: 'spki' })
+        .toString();
+    }
+
+    const certificate = jwk.x5c?.[0];
+    if (certificate) {
+      const body = certificate.match(/.{1,64}/g)?.join('\n') ?? certificate;
+      return `-----BEGIN CERTIFICATE-----\n${body}\n-----END CERTIFICATE-----\n`;
+    }
```
LucieFaire: why did we decide to deal with JWKS resolution ourselves? Why dont we use npm packages like jwks-rsa (Auth0's own) or jose?
There caching, rotation, rate limiting, PEM conversion all handled.

I would strongly suggest to use one of them instead of manual hand-rolling

LucieFaire (commented):
We should also remove `AUTH0_SIGNING_SECRET` from ArgoCD **after** we merge this into dev  (please create a ticket) as it is not used.

Also I would address Claude comments, yet maybe they will not matter if we use a dedicated package to deal with JWKS

vseehausen:
@claude review

vseehausen:
> 1. Extra DB round-trip on every /auth/me for OIDC users
auth.service.ts:121-123 — findEmailById executes a separate DB query on every getUserSession call for OIDC users. Since /auth/me is likely called on every page load, consider caching the email briefly (e.g. in the user lookup already performed) or including it in the JWT claims to avoid the per-request DB hit.

We accept this roundtrip in favor of adding it to the JWT claims.

vseehausen:
@claude review

vseehausen:
> 2. AUTH0_SCOPE default is still 'openid' — docs say to set AUTH0_SCOPE for email claims
configuration.ts:61 — The default 'openid' is correct per AGENTS.md ("do not change config defaults"). However, AUTH.md:152 says to set AUTH0_SCOPE to include email. Ensure deployment documentation or runbooks clearly call out that AUTH0_SCOPE=openid email must be set in environments where email capture is needed, since the default won't request it.

That is fine, as the email auth feature is an opt-in feature, and we keep the default configuration minimal.

> 3. UserEmailAlreadyInUseError surfaces as 409 to the user during login
oidc-auth.controller.ts:221-227 — When persistVerifiedEmail throws UserEmailAlreadyInUseError, the controller re-throws it, returning HTTP 409 to the browser callback. This is documented and intentional, but it's worth noting this is a potential email enumeration vector: an attacker who controls an Auth0 identity with a verified email can probe whether that email is already registered in CGW by observing the 409 vs redirect response. The risk is mitigated by the rate limiter on the OIDC controller, but consider whether logging + redirect with a generic error would be more appropriate than exposing the conflict to the client.

Yes, that is true, but the risk of email enumaration with stolen identies on Auth0 is low.

The other comments will be fixed.

vseehausen:
@claude review again

vseehausen:
> **Claude finished @vseehausen's task** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25003545094)
> ### Code Review
> 
>     * [x]  Read repo guidelines and changed files
> 
>     * [x]  Analyze code quality, security, and correctness
> 
>     * [x]  Post review feedback
> 
> 
> Implementation has matured significantly. The `jose`/`createRemoteJWKSet` switch, write-once email semantics, and test coverage are solid. Two remaining items:
> 
> **1. `InternalServerErrorException` from `persistVerifiedEmail` propagates as 500 during login** `oidc-auth.controller.ts:221-235` — The `catch` block only checks for `UserEmailAlreadyInUseError`. If `persistVerifiedEmail` throws `InternalServerErrorException` (the `affected === 0` + re-read-still-null path at `users.repository.ts:293`), it falls through to the generic catch and redirects with `authentication_failed`. This is fine, but `oidc-auth.service.ts:191` calls `persistVerifiedEmail` with `await` and doesn't catch it — so the error propagates up. Confirm this is intentional: an edge-case race in email persistence kills the entire login. Consider whether catching + logging (and allowing login to proceed without email) would be more resilient.
> 
> **2. `Auth0TokenResponseSchema` no longer validates `expires_in` or `scope`** `auth0-token-response.entity.ts:4-6` — Clean and minimal now (only `id_token`). However, Auth0 token responses still include `access_token`, `token_type`, `expires_in`, and `scope` fields. Parsing with `z.object({ id_token: ... })` without `.passthrough()` or `.strict()` means those extra fields are silently stripped. This is correct behavior, but worth a quick sanity check that nothing downstream expects the full response shape.
> 
> Overall: clean, well-tested, and ready to merge once the above are confirmed as intentional.
> 
> | Branch: `wa-2026-cgw-capture-expose-authenticated-users-email`

Yes, both are intentional.

src/config/entities/configuration.spec.ts
vseehausen: removed in deff5e0f92cf2ddf423b5cf1357013349ecef7e9

src/datasources/jwt/jwt.constants.ts
vseehausen: 77709ae2df07b31d61df5e76526645cd43fe80bd

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: The config for the new jose library: https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-a13ec7a4d04b743754a74ef5f390ded5f6b6aeb41e68de735adba4f15b9df7ed

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-97691cfdb71cf790d2f4502be16e9f0b3a96bd182003b606e9d53e9bceb2564d

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: Yes, that's been a residue from when we had both tokens, removed here https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-97691cfdb71cf790d2f4502be16e9f0b3a96bd182003b606e9d53e9bceb2564d

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: same https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-97691cfdb71cf790d2f4502be16e9f0b3a96bd182003b606e9d53e9bceb2564d

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: Now done like this with jose https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-97691cfdb71cf790d2f4502be16e9f0b3a96bd182003b606e9d53e9bceb2564dR42

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: All updated.

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: Rewritten in favor of https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-97691cfdb71cf790d2f4502be16e9f0b3a96bd182003b606e9d53e9bceb2564dR42

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: Moved the logic into it's own entity: https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-7eacc3173f5a88d819c5f6786ce4f6c1044cb9ca35f69b98b91e10997225a059

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
vseehausen: Sorry, I understood we have the preference for not adding a package. I added [jose](https://github.com/panva/jose) now https://github.com/safe-global/safe-client-gateway/pull/3036/changes/77709ae2df07b31d61df5e76526645cd43fe80bd#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

src/modules/auth/oidc/auth0/domain/auth0-id-token-jwks.factory.ts
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { IConfigurationService } from '@/config/configuration.service.interface';
+import { AUTH0_JWKS_PATH } from '@/modules/auth/oidc/auth0/auth0.constants';
+import type { IAuth0IdTokenJwks } from '@/modules/auth/oidc/auth0/domain/auth0-id-token-jwks.interface';
+import { createRemoteJWKSet } from 'jose';
+
+export function auth0IdTokenJwksFactory(
```
LucieFaire, line 7: As `Auth0TokenVerifier` already has `IConfigurationService` injected and already reads the required variables, I would just call `createRemoteJWKSet` there directly:

```
// In Auth0TokenVerifier constructor — no factory needed
this.jwks = createRemoteJWKSet(new URL(AUTH0_JWKS_PATH, this.issuer), {
  cacheMaxAge: configurationService.getOrThrow('auth.auth0.jwksCacheMaxAgeMs'),
  cooldownDuration: configurationService.getOrThrow('auth.auth0.jwksCooldownMs'),
});
```

That eliminates:
- auth0-id-token-jwks.factory.ts]
- auth0-id-token-jwks.interface.ts
- IAuth0IdTokenJwks provider block in auth0.module.ts
- The awkward dual-import IAuth0IdTokenJwks / type IAuth0IdTokenJwks as Auth0IdTokenJwks in the verifier

src/modules/auth/oidc/auth0/datasources/auth0-api.service.spec.ts
```diff
@@ -72,14 +72,10 @@ describe('Auth0Api', () => {
...
   describe('exchangeAuthorizationCode', () => {
-    it('should exchange an authorization code for tokens', async () => {
+    it('should exchange an authorization code for an ID token', async () => {
       const code = faker.string.alphanumeric(32);
       const tokenResponse = {
-        access_token: faker.string.alphanumeric(),
-        refresh_token: faker.string.alphanumeric(),
         id_token: faker.string.alphanumeric(),
-        token_type: 'Bearer',
-        expires_in: faker.number.int({ min: 60, max: 3_600 }),
```
LucieFaire, line 82: ideally we always try to match the real response to the test one, to make sure we deal with everything as we should and nothing is accidentally missed or added

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.spec.ts
```diff
@@ -1,134 +1,256 @@
...
-const jwtServiceMock = {
-  decode: jest.fn(),
-} as jest.MockedObjectDeep<IJwtService>;
+import jwt from 'jsonwebtoken';
 
 const loggingServiceMock = {
   debug: jest.fn(),
+  info: jest.fn(),
+  warn: jest.fn(),
+  error: jest.fn(),
```
LucieFaire, line 16: are they all in use?

package.json
```diff
@@ -62,6 +62,7 @@
     "cache-manager": "6.4.1",
     "cookie-parser": "1.4.7",
     "csv-stringify": "^6.7.0",
+    "jose": "^6.2.2",
```
LucieFaire, line 65: lets use a static version w/o ^

AGENTS.md
```diff
@@ -83,6 +83,27 @@ This is enforced by a `pre-commit` hook (`insert-license` from `Lucas-C/pre-comm
...
+- Does every changed file belong to this ticket?
+- Is every changed file in the correct directory for this repo?
+- Did any config default change?
+- If yes, was that explicitly required by the ticket?
+- Is there any duplicate or transitional logic that should now be removed?
+- Do docs describe the actual implementation?
+- Is there one clear source of truth in the code?
+- Would a reviewer reasonably ask why a file is here, why a default changed, or why two paths still exist?
+- If yes, the PR is not ready.
+
```
LucieFaire, line 106: unrelated to PR, please exclude to a separate PR

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.spec.ts
```diff
@@ -1,134 +1,256 @@
...
-      };
-      jwtServiceMock.decode.mockReturnValue(decoded);
-
-      const result = target.verifyAndDecode(accessToken);
-
-      expect(result.sub).toBe(decoded.sub);
-      expect(result.iat).toEqual(new Date(now * 1_000));
-      expect(result.nbf).toEqual(new Date(now * 1_000));
-      expect(result.exp).toEqual(new Date((now + 3600) * 1_000));
-      expect(jwtServiceMock.decode).toHaveBeenCalledTimes(1);
```
LucieFaire, line 62: lets bring back checks for claims as they are important too

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,54 +1,64 @@
...
         issuer: this.issuer,
         audience: this.audience,
-        secretOrPrivateKey: this.signingSecret,
-        algorithms: [JWT_ALGORITHM],
+        algorithms: [JWT_RS_ALGORITHM],
       });
-      return Auth0TokenSchema.parse(decoded);
+      const token = Auth0TokenSchema.parse(payload);
+      this.loggingService.debug(
+        `Auth0: id token verified successfully for sub ${token.sub}`,
```
LucieFaire, line 51: dont think we need the log here

src/modules/auth/oidc/auth0/domain/auth0-token.verifier.ts
```diff
@@ -1,54 +1,64 @@
...
+      );
+      return token;
     } catch (error) {
-      if (error instanceof JsonWebTokenError) {
+      if (error instanceof errors.JOSEError || error instanceof z.ZodError) {
         this.loggingService.debug(
-          `Auth0: JWT verification failed: ${error.message}`,
+          `Auth0: id token verification failed: ${error.message}`,
```
LucieFaire, line 57: nit but i think we can keep JWT here instead of id token as its the same

src/modules/auth/oidc/auth0/domain/auth0.repository.spec.ts
```diff
@@ -54,36 +66,55 @@ describe('Auth0Repository', () => {
   });
 
   describe('authenticateWithAuthorizationCode', () => {
-    it('should exchange the code and verify the returned access token', async () => {
+    it('should exchange the code and use claims from the verified id token', async () => {
       const code = faker.string.alphanumeric(32);
-      const accessToken = faker.string.alphanumeric(64);
-      const decodedToken = {
+      const idToken = faker.string.alphanumeric(64);
+      const decodedIdToken = {
```
LucieFaire, line 72: nit: really dont have to say Id everywhere :)

src/modules/auth/oidc/auth0/domain/auth0.repository.spec.ts
```diff
@@ -96,21 +127,16 @@ describe('Auth0Repository', () => {
       expect(auth0TokenVerifierMock.verifyAndDecode).not.toHaveBeenCalled();
     });
 
-    it('should propagate token verifier errors', async () => {
+    it('should propagate id token verifier errors', async () => {
```
LucieFaire, line 130: we only have 1 token, we dont have to explicitly state everywhere its an id one

src/modules/auth/oidc/auth0/domain/auth0.repository.ts
```diff
@@ -21,8 +27,12 @@ export class Auth0Repository implements IAuth0Repository {
...
     code: string,
   ): Promise<Auth0Token> {
+    this.loggingService.debug(
+      'Auth0: exchanging authorization code for tokens',
+    );
     const response = await this.auth0Api.exchangeAuthorizationCode(code);
-    const { access_token } = Auth0TokenResponseSchema.parse(response);
-    return this.auth0TokenVerifier.verifyAndDecode(access_token);
+    const { id_token } = Auth0TokenResponseSchema.parse(response);
+    this.loggingService.debug('Auth0: received id_token from token exchange');
```
LucieFaire, line 35: I would not log so many things

src/modules/auth/oidc/auth0/datasources/entities/auth0-token-response.entity.ts
```diff
@@ -2,12 +2,7 @@
 import { z } from 'zod';
 
 export const Auth0TokenResponseSchema = z.object({
-  access_token: z.string().min(1),
-  refresh_token: z.string().min(1).optional(), //not included atm
   id_token: z.string().min(1),
-  token_type: z.literal('Bearer'),
```
LucieFaire, line 8: i would keep it to make sure its the right type

vseehausen:
Closed to split it up into two PRs. First one is https://github.com/safe-global/safe-client-gateway/pull/3065.

PR #3037, opened 2026-04-22
feat(authentication): return email for active members

vseehausen:
Resolves: https://linear.app/safe-global/issue/WA-2027/cgw-members-returns-email-for-active-members

This PR exposes stored user emails on the members endpoint for active memberships only.

## Summary

- add email to the members response user DTO
- return stored user email for ACTIVE members
- redact email as null for invited members until invite acceptance
- add e2e coverage for active email exposure and invited email redaction

## Notes

- The current invite flow remains wallet-address based; this PR does not add email invites. It simply checks for status `ACTIVE` to return emails.


vseehausen:
@claude review

vseehausen:
> **Claude finished @vseehausen's task** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25316634423)
> ### Code Review
> 
> Overall this is a clean, well-scoped PR. Two items worth addressing:
> 
> **1. `MemberUser` type constraint is out of sync with its fields**
> 
> `MemberUser` at `members.dto.entity.ts:14` declares `implements Pick<User, 'id'>` but now also carries `status` and `email`. Consider updating it to `Pick<User, 'id' | 'status' | 'email'>` so the compiler enforces the contract between the DTO and the domain type.
> 
> **2. `getSelfMembership` doesn't apply email redaction**
> 
> `members.service.ts:89-96` — the `/v1/spaces/:spaceId/membership` endpoint returns a single `MemberDto` with the user object, but it passes the repository result through directly without the `email: member.status === 'ACTIVE' ? ... : null` guard. If a user hits `GET /membership` while their status is `INVITED`, the raw email from the DB would be exposed. Apply the same redaction or extract a shared helper.
> 
> Everything else looks good — the e2e tests cover both the happy path (active email returned) and the redaction path (invited email nulled), and the service-level mapping is straightforward.

1. addressed in 719a0be01ca6d0ee8e1bc90cd811d704af9c600d
2. That's fine. Users see their own invitation only when they authenticate with their own email or address (which is not linked to an email), so there is no data leakage.

@claude review again

src/modules/spaces/routes/entities/members.dto.entity.ts
```diff
@@ -11,12 +11,15 @@ import {
 } from '@/modules/users/domain/entities/user.entity';
 import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
 
-class MemberUser implements Pick<User, 'id'> {
+class MemberUser implements Pick<User, 'id' | 'status' | 'email'> {
```
LucieFaire: we are not exposing status through User, dont think we need it

src/modules/spaces/routes/entities/members.dto.entity.ts
```diff
@@ -11,12 +11,15 @@ import {
...
   @ApiProperty({ type: Number })
   id!: User['id'];
 
   @ApiProperty({ enum: getStringEnumKeys(UserStatus) })
   status!: keyof typeof UserStatus;
+
+  @ApiPropertyOptional({ type: String, nullable: true })
```
LucieFaire: I wonder if we really need this? Right now Members only contain names of users, so if we add email shouldnt we also add wallet then?

src/modules/spaces/routes/entities/members.dto.entity.ts
vseehausen: We actually expose it already here: https://github.com/safe-global/safe-client-gateway/blob/main/src/modules/spaces/routes/entities/members.dto.entity.ts#L19

But I can remove it from this PR.

LucieFaire: you can keep it, they do resolve to the same type, but it could have been different, as we dont say `User['status']` but rather `UserStatus`

src/modules/spaces/routes/entities/members.dto.entity.ts
vseehausen: Good question. I would personally move on with this PR, as the ticket and the design are done for it.

src/modules/spaces/routes/entities/members.dto.entity.ts
vseehausen: Good argument, I removed it, so it stays decoupled. ece3f67fb2040ecf53ef70502cd0654a8af87a05

src/modules/spaces/routes/entities/members.dto.entity.ts
```diff
@@ -11,12 +11,15 @@ import {
...
 
-class MemberUser implements Pick<User, 'id'> {
+class MemberUser implements Pick<User, 'id' | 'email'> {
   @ApiProperty({ type: Number })
   id!: User['id'];
 
   @ApiProperty({ enum: getStringEnumKeys(UserStatus) })
   status!: keyof typeof UserStatus;
+
+  @ApiPropertyOptional({ type: String, nullable: true })
```
LucieFaire: Can this field be both optional (so undefined) and nullable?

src/modules/spaces/routes/members.controller.e2e-spec.ts
```diff
@@ -1288,6 +1302,111 @@ describe('MembersController', () => {
...
+              address: invitedAddress,
+              name: invitedName,
+            },
+          ],
+        })
+        .expect(201);
+      const invitedUser =
+        await usersRepository.findByWalletAddressOrFail(invitedAddress);
+      // Seed an invited email to verify it is still redacted below.
+      await postgresDatabaseService.transaction(async (entityManager) => {
```
LucieFaire: The e2e test having to reach into the DB to set up state that the API doesn't expose is a smell — it's testing a unit-level concern at the wrong layer. The e2e test for GET /members already verifies the response shape with email: null for invited members. I would suggest to create a unit test for `MembersService.findAllBySpaceId` and test DB seeding there

src/modules/spaces/routes/entities/members.dto.entity.ts
vseehausen: Thanks! a17c3889

src/modules/spaces/routes/members.controller.e2e-spec.ts
vseehausen: Thanks! a17c3889

LucieFaire (approved):

LucieFaire (approved):

PR #3055, opened 2026-04-24
feat: share counterfactual safes across users via junction table

katspaugh:
> Safes shared by salt,
> one prediction, many homes,
> rows unchained by one.

## Summary

When two users independently register the same counterfactual safe (same init params, same predicted address, same chain) — a routine scenario when a safe is copied between spaces and later extended to a new chain — the backend currently rejects the second user with a 409 conflict. This PR makes the write path idempotent on matching init params and turns the canonical row into a shared identity that many users can reference.

## What changed

- New `counterfactual_safe_users` junction table keyed on `(counterfactual_safe_id, user_id)`; migrated data backfilled from `creator_id`.
- `creator_id` is retained on `counterfactual_safes` as first-creator audit metadata.
- `POST /v1/users/counterfactual-safes` is now idempotent per `(chainId, address)` with exact init-params equality; it returns 409 only when the submitted init params differ from the stored row.
- `GET /v1/users/counterfactual-safes` queries through the junction, so a user sees every CF safe they are associated with, not just the ones they authored.
- `DELETE /v1/users/counterfactual-safes` removes only the caller's association and leaves the canonical row in place for other users and space links.

## Before / after

\`\`\`mermaid
flowchart LR
  subgraph Before
    A1[User A POST cfSafe X] --> B1[(counterfactual_safes<br/>creator_id = A)]
    A2[User B POST cfSafe X] -->|409 Conflict| B1
  end
  subgraph After
    C1[User A POST cfSafe X] --> D1[(counterfactual_safes<br/>creator_id = A)]
    C1 --> E1[(counterfactual_safe_users<br/>A ↔ X)]
    C2[User B POST cfSafe X<br/>matching params] --> D1
    C2 --> E2[(counterfactual_safe_users<br/>B ↔ X)]
    C3[User B POST cfSafe X<br/>mismatched params] -->|409 Conflict| D1
  end
\`\`\`

## Tests

- 22/22 repository integration tests pass against a real Postgres, covering the new idempotent-match, params-mismatch-409, two-user-share, shared-visibility-after-second-attach, and delete-only-unlinks paths.
- Existing controller e2e tests continue to exercise the POST/GET/DELETE surface; the new 201-on-match and 409-on-mismatch behaviors are covered end-to-end.

## Migration

- Forward-only: creates `counterfactual_safe_users`, adds indexes, and backfills associations for every row that has a non-null `creator_id`. No data is dropped; no existing column changes type.
- Safe to roll back: the new table is dropped in the down migration; the canonical table is untouched.

### DIFF
--- a/migrations/1777636800000-create-counterfactual-safe-users.ts
+++ b/migrations/1777636800000-create-counterfactual-safe-users.ts
@@ -1,9 +1,7 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import type { MigrationInterface, QueryRunner } from 'typeorm';
 
-export class CreateCounterfactualSafeUsers1777636800000
-  implements MigrationInterface
-{
+export class CreateCounterfactualSafeUsers1777636800000 implements MigrationInterface {
   public async up(queryRunner: QueryRunner): Promise<void> {
     await queryRunner.query(
       `CREATE TABLE counterfactual_safe_users (
--- a/migrations/1777636800000-create-counterfactual-safe-users.ts
+++ b/migrations/1777636800000-create-counterfactual-safe-users.ts
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CreateCounterfactualSafeUsers1777636800000
+  implements MigrationInterface
+{
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE counterfactual_safe_users (
+        id SERIAL NOT NULL,
+        counterfactual_safe_id INTEGER NOT NULL,
+        user_id INTEGER NOT NULL,
+        created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+        updated_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(),
+        CONSTRAINT PK_CFSU_id PRIMARY KEY (id),
+        CONSTRAINT UQ_CFSU_cf_safe_user UNIQUE (counterfactual_safe_id, user_id),
+        CONSTRAINT FK_CFSU_cf_safe_id FOREIGN KEY (counterfactual_safe_id) REFERENCES counterfactual_safes(id) ON DELETE CASCADE,
+        CONSTRAINT FK_CFSU_user_id FOREIGN KEY (user_id) REFERENCES users(id) ON DELETE CASCADE
+      );
+      CREATE INDEX IDX_CFSU_user_id ON counterfactual_safe_users(user_id);`,
+    );
+    await queryRunner.query(
+      `CREATE TRIGGER update_counterfactual_safe_users_updated_at
+        BEFORE UPDATE ON counterfactual_safe_users
+        FOR EACH ROW EXECUTE PROCEDURE update_updated_at();`,
+    );
+    await queryRunner.query(
+      `INSERT INTO counterfactual_safe_users (counterfactual_safe_id, user_id)
+        SELECT id, creator_id FROM counterfactual_safes
+        WHERE creator_id IS NOT NULL
+        ON CONFLICT DO NOTHING;`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DROP TABLE IF EXISTS counterfactual_safe_users CASCADE;`,
+    );
+  }
+}
--- a/src/modules/counterfactual-safes/counterfactual-safes.module.ts
+++ b/src/modules/counterfactual-safes/counterfactual-safes.module.ts
@@ -4,6 +4,7 @@ import { UsersModule } from '@/modules/users/users.module';
 import { SpacesModule } from '@/modules/spaces/spaces.module';
 import { PostgresDatabaseModuleV2 } from '@/datasources/db/v2/postgres-database.module';
 import { CounterfactualSafe } from '@/modules/counterfactual-safes/datasources/entities/counterfactual-safe.entity.db';
+import { CounterfactualSafeUser } from '@/modules/counterfactual-safes/datasources/entities/counterfactual-safe-user.entity.db';
 import { CounterfactualSafesRepository } from '@/modules/counterfactual-safes/domain/counterfactual-safes.repository';
 import { ICounterfactualSafesRepository } from '@/modules/counterfactual-safes/domain/counterfactual-safes.repository.interface';
 import { CounterfactualSafesController } from '@/modules/counterfactual-safes/routes/counterfactual-safes.controller';
@@ -16,7 +17,7 @@ import { TypeOrmModule } from '@nestjs/typeorm';
 @Module({
   imports: [
     PostgresDatabaseModuleV2,
-    TypeOrmModule.forFeature([CounterfactualSafe]),
+    TypeOrmModule.forFeature([CounterfactualSafe, CounterfactualSafeUser]),
     forwardRef(() => AuthModule),
     forwardRef(() => UsersModule),
     forwardRef(() => SpacesModule),
--- a/src/modules/counterfactual-safes/datasources/entities/counterfactual-safe-user.entity.db.ts
+++ b/src/modules/counterfactual-safes/datasources/entities/counterfactual-safe-user.entity.db.ts
@@ -1,0 +1,57 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { CounterfactualSafe } from '@/modules/counterfactual-safes/datasources/entities/counterfactual-safe.entity.db';
+import { User } from '@/modules/users/datasources/entities/users.entity.db';
+import {
+  Column,
+  Entity,
+  ManyToOne,
+  JoinColumn,
+  PrimaryGeneratedColumn,
+  Unique,
+} from 'typeorm';
+
+@Entity('counterfactual_safe_users')
+@Unique('UQ_CFSU_cf_safe_user', ['counterfactualSafe', 'user'])
+export class CounterfactualSafeUser {
+  @PrimaryGeneratedColumn({
+    primaryKeyConstraintName: 'PK_CFSU_id',
+  })
+  public readonly id!: number;
+
+  @ManyToOne(
+    () => CounterfactualSafe,
+    (counterfactualSafe: CounterfactualSafe) => counterfactualSafe.id,
+    { onDelete: 'CASCADE', nullable: false },
+  )
+  @JoinColumn({
+    name: 'counterfactual_safe_id',
+    foreignKeyConstraintName: 'FK_CFSU_cf_safe_id',
+  })
+  public readonly counterfactualSafe!: CounterfactualSafe;
+
+  @ManyToOne(() => User, (user: User) => user.id, {
+    onDelete: 'CASCADE',
+    nullable: false,
+  })
+  @JoinColumn({
+    name: 'user_id',
+    foreignKeyConstraintName: 'FK_CFSU_user_id',
+  })
+  public readonly user!: User;
+
+  @Column({
+    name: 'created_at',
+    type: 'timestamp with time zone',
+    default: () => 'CURRENT_TIMESTAMP',
+    update: false,
+  })
+  public readonly createdAt!: Date;
+
+  @Column({
+    name: 'updated_at',
+    type: 'timestamp with time zone',
+    default: () => 'CURRENT_TIMESTAMP',
+    update: false,
+  })
+  public readonly updatedAt!: Date;
+}
--- a/src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
+++ b/src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
@@ -499,7 +499,7 @@ describe('CounterfactualSafesRepository', () => {
   });
 
   describe('delete', () => {
-    it('should remove the user association and the canonical row when no other associations remain', async () => {
+    it('should remove the user association but keep the canonical row for audit trail', async () => {
       const user = await dbUserRepo.insert({ status: 'ACTIVE' });
       const userId = user.identifiers[0].id as User['id'];
       const payload = buildCounterfactualSafePayload();
--- a/src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
+++ b/src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
@@ -12,6 +12,7 @@ import { Wallet } from '@/modules/wallets/datasources/entities/wallets.entity.db
 import { Space } from '@/modules/spaces/datasources/entities/space.entity.db';
 import { SpaceSafe } from '@/modules/spaces/datasources/entities/space-safes.entity.db';
 import { CounterfactualSafe } from '@/modules/counterfactual-safes/datasources/entities/counterfactual-safe.entity.db';
+import { CounterfactualSafeUser } from '@/modules/counterfactual-safes/datasources/entities/counterfactual-safe-user.entity.db';
 import { CounterfactualSafesRepository } from '@/modules/counterfactual-safes/domain/counterfactual-safes.repository';
 import { counterfactualSafeBuilder } from '@/modules/counterfactual-safes/datasources/entities/__tests__/counterfactual-safe.entity.db.builder';
 import { UniqueConstraintError } from '@/datasources/errors/unique-constraint-error';
@@ -69,6 +70,7 @@ describe('CounterfactualSafesRepository', () => {
   let dbUserRepo: Repository<User>;
   let dbWalletRepo: Repository<Wallet>;
   let dbCounterfactualSafeRepo: Repository<CounterfactualSafe>;
+  let dbCounterfactualSafeUserRepo: Repository<CounterfactualSafeUser>;
 
   const testDatabaseName = faker.string.alpha({
     length: 10,
@@ -83,7 +85,15 @@ describe('CounterfactualSafesRepository', () => {
       database: testDatabaseName,
     }),
     migrationsTableName: testConfiguration.db.orm.migrationsTableName,
-    entities: [CounterfactualSafe, Member, Space, SpaceSafe, User, Wallet],
+    entities: [
+      CounterfactualSafe,
+      CounterfactualSafeUser,
+      Member,
+      Space,
+      SpaceSafe,
+      User,
+      Wallet,
+    ],
   });
 
   beforeAll(async () => {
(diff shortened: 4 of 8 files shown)


katspaugh:
## PR Briefing

[Open the PR briefing](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-client-gateway%2Fpr-3055%2F6057d28770d9%2Fbundle.json)

▶ **[Open the interactive review](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-client-gateway%2Fpr-3055%2F6057d28770d9%2Fbundle.json)**
▶ **[Watch the walkthrough video](https://github.com/katspaugh/popcorn-review-artifacts/releases/download/prbrief-safe-global-safe-client-gateway-pr-3055-6057d28770d9/video-github.mp4)** (direct MP4)

This PR replaces the single-owner model for counterfactual safes with a many-to-many junction table (`counterfactual_safe_users`) so multiple users can share a single canonical `counterfactual_safes` row. A new migration creates the junction table, backfills associations from the existing `creator_id` column (which is retained for audit purposes), and the `CounterfactualSafesRepository` is rewritten to make `POST /v1/users/counterfactual-safes` idempotent on matching init params, returning 409 only when submitted params differ from the stored row. `GET` now queries through the junction table so users see all safes they are associated with, and `DELETE` removes only the caller's junction row rather than the canonical record. The `CounterfactualSafeUser` entity, module registration, service layer, interface, and e2e/integration tests are all updated to reflect the new semantics.

<details>

## Summary

This PR replaces the single-owner model for counterfactual safes with a many-to-many junction table (`counterfactual_safe_users`) so multiple users can share a single canonical `counterfactual_safes` row. A new migration creates the junction table, backfills associations from the existing `creator_id` column (which is retained for audit purposes), and the `CounterfactualSafesRepository` is rewritten to make `POST /v1/users/counterfactual-safes` idempotent on matching init params, returning 409 only when submitted params differ from the stored row. `GET` now queries through the junction table so users see all safes they are associated with, and `DELETE` removes only the caller's junction row rather than the canonical record. The `CounterfactualSafeUser` entity, module registration, service layer, interface, and e2e/integration tests are all updated to reflect the new semantics.

## Architecture notes

The change introduces a `counterfactual_safe_users` junction table with FK references to `counterfactual_safes` and `users`, turning a one-to-many (creator→safe) relationship into many-to-many. The `CounterfactualSafesRepository.create` method now runs a find-then-insert transaction with `associateUser` using `orIgnore`, and `findByUserId` issues a raw-column JOIN against the junction table. The `CounterfactualSafesService` passes `userId` (not `creatorId`) to all repository calls. The migration is forward-only with a DROP TABLE down path, meaning a rollback after deployment would break GET and DELETE at runtime.

## Ripple risks

Analyzed 4 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 8 file(s), 6 changed export(s), 0 uncovered callsite(s) in 1.9s._

## Walkthrough outline

_19 steps across 6 chapters, approximately 4 min 15s._

### Why this change

1. **motivation** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L21-40 `CounterfactualSafesRepository`
   Here is the problem this PR solves. When two users independently register the same counterfactual safe — same chain, same predicted address, same init params — the old code would hit a unique constraint and return a 409 conflict to the second user. That is wrong: both users have a legitimate claim to the same canonical safe row.

2. **motivation** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L21-40
   The fix is a many-to-many junction table. One canonical row lives in counterfactual safes, and each user gets their own row in counterfactual safe users pointing at it. Let's see how that plays out across the codebase.

### The junction table foundation

3. **change** · `migrations/1777636800000-create-counterfactual-safe-users.ts` L4-30 `CreateCounterfactualSafeUsers1777636800000`
   The migration creates the counterfactual safe users table with foreign keys to both counterfactual safes and users, and a unique constraint on the pair so no duplicate associations can sneak in. It also wires up the update timestamp trigger that the rest of the schema already uses.

4. **change** · `migrations/1777636800000-create-counterfactual-safe-users.ts` L31-55
   Right after creating the table, the migration backfills it. Every existing row in counterfactual safes that has a non-null creator id gets an association row inserted for that creator, using on conflict do nothing so the backfill is safe to re-run. The creator id column itself stays on the canonical table as an audit trail of who first registered the safe.

5. **change** · `src/modules/counterfactual-safes/datasources/entities/counterfactual-safe-user.entity.db.ts` L13-45 `CounterfactualSafeUser`
   Here is the TypeORM entity that maps to that new table. It has two many-to-one relations — one to CounterfactualSafe and one to User — and a unique decorator covering both relation fields so the ORM enforces the same constraint at the application layer.

6. **change** · `src/modules/counterfactual-safes/counterfactual-safes.module.ts` L17-35 `CounterfactualSafesModule`
   The module now registers CounterfactualSafeUser alongside CounterfactualSafe in the TypeORM feature list, which makes the junction repository injectable. Without this line the new entity would be invisible to the dependency injection container.

### Rewriting the repository

7. **change** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L21-50 `CounterfactualSafesRepository`
   Now the heart of the PR: the repository rewrite. The constructor now injects a second TypeORM repository for CounterfactualSafeUser alongside the existing one for CounterfactualSafe. Everything else flows from that second injection.

8. **change** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L51-100 `CounterfactualSafesRepository`
   The create method now runs inside a transaction. It first does a find-one lookup for an existing row matching chain and address. If one exists and the init params match exactly, it skips the insert and jumps straight to associating the current user. If the params differ, it throws a 409 conflict — that is the only remaining conflict case.

9. **change** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L101-130
   The private associate user helper inserts a row into counterfactual safe users with or-ignore semantics, so if the calling user is already associated — say they re-submit the same safe — the insert is silently skipped and the operation returns cleanly. That is what makes the POST endpoint idempotent.

10. **change** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L131-165
   Find by user id now joins through the junction table. It selects canonical safe rows where a matching junction row exists for the requesting user, so a user sees every safe they are associated with regardless of who first created it.

11. **change** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L166-200
   Delete is now a soft unlink. It removes only the caller's row from counterfactual safe users and leaves the canonical safe row intact for any other users still associated with it. The old behavior would have deleted the canonical row outright, which would have broken other users' references.

### Consumers that had to adapt

12. **ripple** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.interface.ts` L15-35 `ICounterfactualSafesRepository`
   The repository interface reflects the rename: find by creator id is now find by user id, and the delete method's parameter object uses user id instead of creator id. Any other implementation of this interface — including mocks in tests — must be updated to match.

13. **ripple** · `src/modules/counterfactual-safes/routes/counterfactual-safes.service.ts` L11-45 `CounterfactualSafesService`
   The service layer is a thin pass-through change. The calls to find by creator id and delete now pass user id instead of creator id, aligning with the renamed interface. The authenticated user identity comes from the same helper as before, so no auth logic changed.

14. **ripple** · `src/modules/counterfactual-safes/routes/__tests__/counterfactual-safes.controller.e2e-spec.ts` L1-40
   The end-to-end tests now cover three POST scenarios: same user re-submitting returns 201, a different user with matching params also returns 201 and can see the safe via GET, and a different user with mismatched params still gets 409. These replace the old single duplicate-conflict test.

15. **ripple** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts` L1-50
   The integration spec adds CounterfactualSafeUser to the test data source and cleans up junction rows before canonical rows in after-each, respecting the foreign key order. New test cases cover two-user sharing, shared visibility after a second attach, and the delete-only-unlinks path against a real Postgres instance.

### What can go wrong

16. **risk** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L51-100
   There is a time-of-check-to-time-of-use race in the create method. Two concurrent POSTs for the same chain and address can both pass the find-one check before either inserts, and the second insert will hit the database unique constraint rather than the application-level 409 path. The error surfaced to the caller may be a raw database error rather than the expected conflict response.

17. **risk** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L101-130
   The associate user helper uses or-ignore, which silently swallows all insert errors on the junction table — not just duplicate-key violations. A foreign key failure or schema mismatch would be swallowed too, making those bugs invisible at runtime. Worth a close look before merging.

18. **risk** · `migrations/1777636800000-create-counterfactual-safe-users.ts` L55-75
   The down migration drops the junction table with cascade but does not restore any data. If this migration is rolled back after the application code is already deployed, the GET and DELETE endpoints will fail at runtime because they query a table that no longer exists. Plan the rollback window carefully.

### The bottom line

19. **wrap** · `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts` L21-50 `CounterfactualSafesRepository`
   Bottom line: this PR turns counterfactual safe ownership from a single-creator lock into a shared association model, making the POST endpoint idempotent on matching params and letting DELETE cleanly unlink one user without disturbing others — a solid foundation for cross-space safe sharing.

## Reviewer checklist

- [ ] Verify the `create` transaction is safe under concurrent POSTs for the same `(chainId, address)`: the non-locking `findOne` followed by `insert` has a TOCTOU race where two simultaneous requests can both miss the existing row and collide on the unique constraint, which is no longer caught by the error handler.
- [ ] Confirm `isSameAddressList` (used in `isSameInitParams`) is order-insensitive or that the API contract guarantees owners are always submitted in a canonical order, since two registrations with the same owners in different order will currently produce a 409 conflict.
- [ ] Check that `associateUser` using `orIgnore` does not silently swallow FK violations or other unexpected insert errors on `counterfactual_safe_users` beyond the intended duplicate-key case.
- [ ] Ensure the `CounterfactualSafeUser` entity is registered in the global TypeORM entity list (e.g., `PostgresDatabaseModuleV2`) and not only in `CounterfactualSafesModule`, otherwise the table will be unknown to the ORM at runtime.
- [ ] Validate the down migration is safe: dropping `counterfactual_safe_users` with CASCADE after the new application code is deployed will break GET and DELETE endpoints — confirm the rollback procedure includes a coordinated code revert.
- [ ] Add or confirm e2e coverage for DELETE after two users share a safe (assert user1's association survives after user2 deletes), and verify GET for user1 still returns the safe after user2 attaches — both gaps are noted in the current test suite.

## File-by-file

### `migrations/1777636800000-create-counterfactual-safe-users.ts`

Adds a new migration `CreateCounterfactualSafeUsers1777636800000` that creates the `counterfactual_safe_users` junction table with FK constraints to `counterfactual_safes` and `users`, an `updated_at` trigger, and backfills existing rows from `counterfactual_safes.creator_id`.

_Enables multiple users to share a single canonical counterfactual safe row, replacing the implicit single-owner model with an explicit many-to-many association._

**Risks:**
- Backfill uses `ON CONFLICT DO NOTHING` — verify no valid associations are silently skipped if `creator_id` is non-null but duplicated across rows
- Down migration drops the table with CASCADE but does not restore any data; if rolled back after the application code is deployed, GET/DELETE endpoints will break
- No index on `counterfactual_safe_id` column — queries filtering by safe rather than user will do full scans; the unique constraint covers the pair but a dedicated index on `counterfactual_safe_id` alone may be needed
- Timestamp `1777636800000` is in the future (May 2026); verify migration ordering is intentional and won't conflict with other pending migrations
- The `update_updated_at` procedure is assumed to already exist — migration will fail if it doesn't; no guard or creation step is included

### `src/modules/counterfactual-safes/counterfactual-safes.module.ts`

Imports `CounterfactualSafeUser` entity and registers it alongside `CounterfactualSafe` in `TypeOrmModule.forFeature` within `CounterfactualSafesModule`.

_Makes the new junction table entity available for injection as a TypeORM repository within the module's providers._

**Risks:**
- Verify `CounterfactualSafeUser` entity is also registered in the database connection's entity list (e.g., PostgresDatabaseModuleV2 or ormconfig), otherwise TypeORM won't recognize the table.
- Check that no other module independently registers `CounterfactualSafeUser` causing duplicate repository conflicts.

### `src/modules/counterfactual-safes/datasources/entities/counterfactual-safe-user.entity.db.ts`

New TypeORM entity `CounterfactualSafeUser` added for the `counterfactual_safe_users` junction table, with `ManyToOne` relations to `CounterfactualSafe` and `User`, a unique constraint on `(counterfactualSafe, user)`, and `createdAt`/`updatedAt` timestamp columns.

_This junction table enables multiple users to share a single canonical counterfactual safe row, replacing the previous single-owner model._

**Risks:**
- updatedAt has `update: false` so it will never be updated after insert — likely intentional but should be confirmed
- The `@Unique` decorator references relation objects `['counterfactualSafe', 'user']` rather than column names `['counterfactual_safe_id', 'user_id']` — verify TypeORM resolves this correctly to the FK columns
- No `updated_at` trigger or auto-update mechanism exists; if audit trail of changes is needed this is insufficient
- CounterfactualSafe entity needs a corresponding `OneToMany` back-reference or the relation target `counterfactualSafe.id` used in ManyToOne callback is misleading

### `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts`

Integration tests for `CounterfactualSafesRepository` are updated to cover the new junction-table model: `CounterfactualSafeUser` entity is added to the test data source, `afterEach` cleanup deletes junction rows first, `findByCreatorId` tests are renamed/rewritten to `findByUserId`, and new test cases cover idempotent re-submission, cross-user sharing, and delete-only-unlinks semantics.

_The tests validate that the repository's new shared-identity behavior (idempotent create, junction-based visibility, association-only delete) works correctly against a real Postgres schema including the new migration._

**Risks:**
- Delete test 'should remove the user association and the canonical row when no other associations remain' asserts canonical row is kept (length 1) but the comment says 'intentionally kept (audit trail)' — verify this matches the actual repository implementation and is not a test bug masking a missing cascade delete
- afterEach cleanup deletes junction rows before canonical rows; if FK constraints are DEFERRABLE or cascade is set differently, order may not matter, but confirm the migration's FK definition matches this assumed dependency order
- The 'should throw UniqueConstraintError when (chainId, address) collides with different init params' test uses two different users — confirm the repository actually compares init params and throws UniqueConstraintError (not a raw DB constraint error) when params differ
- No test verifies that `findByUserId` does NOT return safes the user has no association with (only the 'empty array' case covers a user with zero associations, not a user who exists but is unassociated with a specific safe that exists)
- The 'delete multiple associations' test only checks junction rows are empty but does not assert canonical rows still exist, potentially missing a regression if delete accidentally cascades

### `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.interface.ts`

Renamed `creatorId` to `userId` in two interface method signatures: `findByCreatorId` (renamed to `findByUserId`) and the `delete` method's parameter object.

_Reflects the shift from creator-only ownership to shared user associations via the junction table, ensuring the interface contract matches the new multi-user semantics._

**Risks:**
- Verify all implementations of ICounterfactualSafesRepository are updated to match the renamed method and parameter
- Check that no callers still pass `creatorId` as a named property to `delete` — structural typing could silently accept the old shape if the field name wasn't enforced
- Confirm `findByCreatorId` is fully removed/replaced and not still referenced anywhere under the old name

### `src/modules/counterfactual-safes/domain/counterfactual-safes.repository.ts`

Rewrites `CounterfactualSafesRepository`: `create` now runs inside a transaction that upserts via find-then-insert and calls `associateUser` (inserting into `CounterfactualSafeUser` with `orIgnore`); `findByCreatorId` is renamed `findByUserId` and queries through the junction table; `delete` now removes only the caller's `CounterfactualSafeUser` row rather than the canonical `CounterfactualSafe` row.

_Enables multiple users to share a single canonical counterfactual safe row, making POST idempotent on matching init params and returning 409 only on param mismatch, while DELETE becomes a soft-unlink that preserves the row for other users._

**Risks:**
- isSameInitParams does a strict string equality on owners array elements but `create` normalizes incoming owners via getAddress while the stored row may have been inserted before normalization — verify the DB always stores checksummed addresses
- isSameAddressList is order-sensitive; two registrations with the same owners in different order will be treated as mismatched params and throw 409
- The transaction in `create` does a non-locking `findOne` then `insert`, creating a TOCTOU race: two concurrent POSTs for the same (chainId, address) can both see no existing row and both attempt insert, causing a DB unique-constraint error that is no longer caught
- delete builds a dynamic OR clause with positional parameter names (chainId0, address0, …); if args.payload is empty the WHERE clause is an empty string which is invalid SQL — no guard against empty payload
- associateUser uses `orIgnore` which silently swallows all insert errors on CounterfactualSafeUser, not just duplicate-key; any FK violation or schema error will be silently ignored
- findByUserId uses a raw column name string 'cfsu.counterfactual_safe_id = cfs.id' and 'cfsu.user_id' — these must exactly match the DB column names; a mismatch will fail at runtime rather than compile time
- The old `delete` path called `findOrFail` which threw NotFoundException; the new path also throws NotFoundException but only checks junction rows — a user who was never associated but requests deletion of a valid CF safe will get NotFoundException rather than a more descriptive error

### `src/modules/counterfactual-safes/routes/__tests__/counterfactual-safes.controller.e2e-spec.ts`

Replaces the 'duplicate chainId+address → 409' test with three new e2e tests: same-user idempotent re-submit returns 201, different-user with mismatched init params returns 409, and different-user with matching init params returns 201 and is visible via GET. Also renames the delete test description and casts `cfSafe.chainId` to `string` to fix TypeScript indexing.

_Validates the new shared-identity semantics end-to-end: idempotent writes, conflict only on param mismatch, and cross-user visibility through the junction table._

**Risks:**
- The 'second user attach' test only checks getResponse.body.safes[cfSafe.chainId].length === 1 for user2, but does not verify user1 still sees the safe — regression in user1 visibility is untested here
- No test covers DELETE behavior after two users share a safe (i.e., that deleting user2's association leaves user1's intact)
- The idempotent re-submit test uses the same user/token for both POSTs but does not assert the response body or that no duplicate row was created
- wallet creation requests (POST /v1/users/wallet) are not awaited with .expect(), so failures there would be silently swallowed

### `src/modules/counterfactual-safes/routes/counterfactual-safes.service.ts`

Renames the parameter passed to `findByCreatorId` (now `findByUserId`) and `delete` from `creatorId: userId` to `userId` in `CounterfactualSafesService`, aligning the service layer with the new junction-table-based repository methods.

_Ensures GET and DELETE operations query/remove by the authenticated user's junction-table association rather than the legacy `creator_id` column, enabling shared counterfactual safe visibility and non-destructive unlink on delete._

**Risks:**
- Verify `findByUserId` in the repository actually queries the junction table and not just a renamed `creator_id` filter
- Confirm `delete` in the repository only removes the junction row and does not cascade-delete the canonical `counterfactual_safes` row
- Check that `getAuthenticatedUserIdOrFail` still returns the correct user identity after the junction-table migration

---
_Generated by prbrief • model: claude-sonnet-4-6 • 10 LLM calls • $0.19 • 88.6s_

</details>

<sub>Generated by <a href="https://popcorn.report">Popcorn Review</a>. · `safe-global/safe-client-gateway@6057d28770d9`</sub>

clovisdasilvaneto (approved):

PR #3056, opened 2026-04-24
migrating addressBook createdBy updatedBy to be userId

0xAryanShah:
## Summary  [Linear](https://linear.app/safe-global/issue/WA-1912/address-book-migrate-createdbylastupdatedby-to-support-oidc-users)                                                                                                                                        

- Migrate address book item `createdBy`/`lastUpdatedBy` columns from wallet addresses (`varchar(42)`) to user IDs (`integer`)
- Resolve user identity (wallet address or email) at the service layer so the API returns meaningful attribution instead of raw user IDs
- Fix `createdByOverride` bug where approved address book requests were attributed to the approving admin instead of the original requester
- Fix pre-existing `@OneToMany` inverse side bug on `User.wallets` (`wallet.id` → `wallet.user`)
- Fix pre-existing `walletBuilder` property name bug (`created_at` → `createdAt`)

## Changes

**Migration** (`1777033895000-address-book-attribution-to-user-id.ts`):
- Preflight check that aborts the migration if any wallet address can't be resolved
- Backfill `created_by`/`last_updated_by` from wallet addresses to user IDs via `LOWER()` case-insensitive join
- Down migration drops constraints/indexes first, uses deterministic `ORDER BY` for wallet selection

**Entity + Schema**:
- `createdBy`/`lastUpdatedBy` changed to `number` type
- Zod schema updated to `z.number().int()`

**Repository**:
- `createdByOverride` type changed from `Address` to `number`
- `createNewAddressBookItems` now uses `args.createdByOverride ?? args.userId`

**Service + DTO**:
- New `buildUserIdentityMap`: loads users + wallets in parallel, resolves identity as wallet address (preferred) or email (fallback)
- DTO adds `createdByUserId`/`lastUpdatedByUserId` (`number`), `createdBy`/`lastUpdatedBy` now show identity string or `"Deleted user"`
- `UserAddressBookItemDto` fixed to reference `UserAddressBookItem` types instead of `AddressBookDbItem`

**Request approval flow**:
- `createdByOverride` now passes `request.requestedBy.id` (user ID) instead of `request.requestedByWallet` (Address)


0xAryanShah:
@claude review

0xAryanShah:
@claude review again, addressed previous review comments:
1) Intentional design decision
2) Already handled. The DatabaseMigrator service runs migrations with { transaction: 'all' }
3) Not possible in practice. The wallets table has a UNIQUE constraint on address, so one address maps to exactly one user_id.
4) Fixed : [2fa1fba](https://github.com/safe-global/safe-client-gateway/pull/3056/commits/2fa1fbad238d8b5b59929f8e7d9fa2eb566c4d47)
5) invited_by is a separate migration : [Linear](https://linear.app/safe-global/issue/WA-1913/invitedby-resolve-null-conflation-for-oidc-admin-invitations)
6) Added: [2fa1fba](https://github.com/safe-global/safe-client-gateway/pull/3056/commits/2fa1fbad238d8b5b59929f8e7d9fa2eb566c4d47)

katspaugh:
## PR Briefing

[Open the PR briefing](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-client-gateway%2Fpr-3056%2F2fa1fbad238d%2Fbundle.json)

▶ **[Open the interactive review](https://viewer.popcorn.report/view?bundle_url=https%3A%2F%2Fkatspaugh.github.io%2Fpopcorn-review-artifacts%2Fbriefings%2Fsafe-global%2Fsafe-client-gateway%2Fpr-3056%2F2fa1fbad238d%2Fbundle.json)**
▶ **[Watch the walkthrough video](https://github.com/katspaugh/popcorn-review-artifacts/releases/download/prbrief-safe-global-safe-client-gateway-pr-3056-2fa1fbad238d/video-github.mp4)** (direct MP4)

This PR migrates `space_address_book_items.created_by` and `last_updated_by` columns from `varchar(42)` wallet addresses to `integer` user IDs, unblocking OIDC-authenticated users from writing address book entries. A new TypeORM migration (`AddressBookAttributionToUserId1777033895000`) handles the schema change with a pre-flight guard, backfill via a `wallets` table JOIN, column drop, and rename. The `isSiwe()` guards previously blocking OIDC users in `AddressBooksService` and `AddressBookItemsRepository` are removed, with attribution now sourced from `getAuthenticatedUserIdOrFail(authPayload)`. This is a breaking API change: `createdBy`/`lastUpdatedBy` in GET responses change from wallet address strings like `'0xAbc...'` to numeric user ID strings like `'42'`, requiring coordinated client updates.

<details>

## Summary

This PR migrates `space_address_book_items.created_by` and `last_updated_by` columns from `varchar(42)` wallet addresses to `integer` user IDs, unblocking OIDC-authenticated users from writing address book entries. A new TypeORM migration (`AddressBookAttributionToUserId1777033895000`) handles the schema change with a pre-flight guard, backfill via a `wallets` table JOIN, column drop, and rename. The `isSiwe()` guards previously blocking OIDC users in `AddressBooksService` and `AddressBookItemsRepository` are removed, with attribution now sourced from `getAuthenticatedUserIdOrFail(authPayload)`. This is a breaking API change: `createdBy`/`lastUpdatedBy` in GET responses change from wallet address strings like `'0xAbc...'` to numeric user ID strings like `'42'`, requiring coordinated client updates.

## Architecture notes

The change touches four layers: the database migration, the TypeORM entity (`AddressBookItem`), the Zod DB schema (`AddressBookDbItemSchema`), and the service/repository layer (`AddressBooksService`, `AddressBookItemsRepository`). The DTO (`SpaceAddressBookItemDto`) is explicitly typed as `string` and decoupled from the DB entity type, meaning future DB type changes won't automatically propagate to the API contract. Attribution resolution is centralized: `getAuthenticatedUserIdOrFail` is called once in `upsertAddressBookItems` and the resolved `userId` is threaded into both `createNewAddressBookItems` and `updateExistingAddressBookItems`. No foreign key constraint is added on the new integer columns referencing `users.id`, so orphaned user IDs can accumulate silently.

## Ripple risks

Analyzed 6 changed exports. No untouched external callers found — all usages appear to be inside the PR diff.

_Analyzed 9 file(s), 7 changed export(s), 0 uncovered callsite(s) in 1.8s._

## Walkthrough outline

_19 steps across 6 chapters, approximately 4 min 15s._

### Why this change

1. **motivation** · `src/modules/spaces/routes/address-books.service.ts` L1-20 `AddressBooksService`
   Here is the problem this PR is solving. OIDC-authenticated users, those who log in with email rather than a wallet, were blocked from writing to the address book because the attribution columns stored wallet addresses, and email-only users simply have no wallet address to store. The fix is to swap those columns from wallet address strings to numeric user IDs, which every authenticated user has regardless of login method.

2. **motivation** · `src/modules/spaces/routes/address-books.service.ts` L1-20 `AddressBooksService`
   This diagram shows the before-and-after flow for a write operation. Previously, OIDC users hit a forbidden wall at the service layer. After this PR, both SIWE and OIDC users reach the repository, and attribution is resolved through a shared helper that returns a numeric user ID.

### The database migration

3. **change** · `migrations/1777033895000-address-book-attribution-to-user-id.ts` L1-20 `AddressBookAttributionToUserId1777033895000`
   The migration class is the foundation of everything else in this PR. It carries a timestamp well into the future, so double-check that your migration runner orders by timestamp and that this does not collide with anything already in flight. The class name makes the intent clear: attribution columns are moving to user ID.

4. **change** · `migrations/1777033895000-address-book-attribution-to-user-id.ts` L21-45
   The up migration starts with a pre-flight guard. It checks whether any row in space_address_book_items has a created_by or last_updated_by wallet address that cannot be resolved to a user ID via the wallets table. If even one row fails resolution, the migration throws loudly rather than silently producing nulls. This is a good defensive pattern.

5. **change** · `migrations/1777033895000-address-book-attribution-to-user-id.ts` L46-75
   After the guard, the migration adds two new integer columns, backfills them by joining against the wallets table, drops the old varchar columns, renames the new ones, and finally enforces NOT NULL. These are separate statements with no explicit transaction wrapping, so a failure mid-sequence could leave the table in a partially migrated state.

6. **change** · `migrations/1777033895000-address-book-attribution-to-user-id.ts` L76-110
   The down migration reverses the process, but it is explicitly lossy. For users who authenticated via OIDC and have no wallet, the rollback leaves those attribution columns as NULL. Any downstream code that expects a NOT NULL varchar after a rollback will break, so the team has accepted this as a one-way door.

7. **change** · `migrations/1777033895000-address-book-attribution-to-user-id.ts` L46-60
   One subtle point worth flagging: the pre-flight subquery uses OR logic to detect unresolvable rows. A row that fails only one of the two columns will still trigger the guard, which is the correct behavior, but the SQL deserves a careful read to confirm it handles every edge case, especially rows where one column is already null.

### Entity and schema updates

8. **change** · `src/modules/spaces/datasources/entities/address-book-item.entity.db.ts` L17-45 `AddressBookItem`
   The TypeORM entity reflects the schema change directly. The created_by and last_updated_by columns are now typed as number and declared as integer columns, and the database address transformer that used to normalize wallet address casing has been removed entirely. This is the source of truth that TypeORM uses to generate queries.

9. **change** · `src/modules/spaces/domain/address-books/entities/address-book-item.db.entity.ts` L12-30 `AddressBookDbItemSchema`
   The Zod schema for the DB entity mirrors the entity change. The created_by and last_updated_by fields move from the address schema type to z.number().int(), so any value coming out of the database that is not a plain integer will now fail validation at the Zod boundary rather than silently passing through as a string.

10. **change** · `src/modules/spaces/routes/entities/space-address-book.dto.entity.ts` L5-30 `SpaceAddressBookItemDto`
   The API-facing DTO is deliberately decoupled from the DB entity type. The created_by and last_updated_by fields are now explicitly typed as string in the DTO, and the Swagger descriptions are updated to say user ID of the creator and last updater. This means the API contract is stable even if the DB column type changes again in the future, but it also means drift between the two layers is now invisible to the compiler.

### Service and repository rewiring

11. **change** · `src/modules/spaces/routes/address-books.service.ts` L1-50 `AddressBooksService`
   Back in the service, the two isSiwe guard calls are simply gone. The forbidden exception import is also removed. In their place, the mapping from DB item to DTO now calls toString on the numeric created_by and last_updated_by values, converting the integer user ID into the string the DTO contract expects.

12. **change** · `src/modules/spaces/domain/address-books/address-book-items.repository.ts` L18-70 `AddressBookItemsRepository`
   In the repository, the isSiwe guards are also removed from both create and update paths. The key change is that getAuthenticatedUserIdOrFail is called once at the top of upsertAddressBookItems, and the resolved numeric user ID is then threaded down into both createNewAddressBookItems and updateExistingAddressBookItems as a plain number argument.

13. **ripple** · `src/modules/spaces/domain/address-books/entities/__tests__/address-book-item.db.builder.ts` L11-35 `addressBookItemBuilder`
   The test builder is updated to match. The created_by and last_updated_by fields now generate random integers using faker's number generator bounded by the database max safe integer constant, replacing the old ethereum address generation. Any test that previously asserted these fields look like wallet addresses will need to be updated.

14. **ripple** · `src/modules/spaces/domain/address-books/address-book-items.repository.integration.spec.ts` L1-50
   The integration tests gain two new helpers: createOidcUser and createSpaceAsOidcAdmin. These stand up the database state needed to exercise the OIDC write path end to end. Two new test cases cover OIDC upsert and OIDC delete, confirming that the full stack works without a wallet address in the picture.

### What can go wrong

15. **risk** · `migrations/1777033895000-address-book-attribution-to-user-id.ts` L46-75
   The migration has no explicit transaction wrapping the backfill, the NOT NULL enforcement, and the column drop. If the process crashes between the backfill and the NOT NULL statement, the table is left with nullable integer columns and no constraint. On a large table this window is real, and a partial failure here is hard to recover from cleanly.

16. **risk** · `src/modules/spaces/datasources/entities/address-book-item.entity.db.ts` L17-45 `AddressBookItem`
   There is no foreign key constraint added on the new integer columns referencing users dot id. This means if a user is deleted, their ID can remain in created_by and last_updated_by indefinitely with no referential integrity error. Orphaned user IDs will accumulate silently unless the application layer cleans them up.

17. **risk** · `src/modules/spaces/routes/address-books.service.ts` L1-50 `AddressBooksService`
   The toString call on created_by and last_updated_by in the DTO mapping will throw at runtime if either value is null or undefined. The migration enforces NOT NULL, but if the migration has not run yet in a given environment, or if a rollback has occurred, null values are possible and this code path will crash.

18. **risk** · `src/modules/spaces/routes/entities/space-address-book.dto.entity.ts` L5-30 `SpaceAddressBookItemDto`
   This is a breaking API change with no versioning guard. Any client reading created_by or last_updated_by as a wallet address string will now receive a numeric string like forty-two instead of a hex address. The DTO has no validation decorator enforcing the new format, so malformed values could slip through without a clear error signal to the caller.

### Closing the loop

19. **wrap** · `src/modules/spaces/domain/address-books/address-book-items.repository.ts` L18-70 `AddressBookItemsRepository`
   Bottom line: this PR successfully unblocks OIDC users from address book writes by migrating attribution from wallet addresses to numeric user IDs across the migration, entity, schema, service, and repository layers, but it ships a breaking API change and a migration without transaction safety, so coordinate client updates and validate the migration carefully before deploying to production.

## Reviewer checklist

- [ ] Verify the migration's pre-flight SQL logic correctly catches rows where either `created_by` OR `last_updated_by` cannot be resolved to a user ID, and confirm the backfill + NOT NULL steps are wrapped in a transaction to prevent partial failure leaving nullable columns.
- [ ] Confirm no remaining `isSiwe()` guards or `signer_address` usages exist in the `deleteAddressBookItems` path or any other address-book write path not covered by this diff.
- [ ] Check that `item.createdBy` and `item.lastUpdatedBy` are guaranteed non-null before `.toString()` is called in `AddressBooksService`, and that `z.number().int()` in `AddressBookDbItemSchema` correctly rejects null/undefined values post-migration.
- [ ] Validate that the `Number(authPayload.sub)` conversion in integration test helpers cannot silently produce `NaN` (e.g., if `sub` is a non-numeric OIDC subject), and that `createSpaceAsOidcAdmin` reliably extracts `id` from `generatedMaps[0]`.
- [ ] Confirm that downstream API clients consuming `createdBy`/`lastUpdatedBy` as wallet address strings have been updated or are being updated in a coordinated release, given this is an undocumented breaking change with no API versioning guard.
- [ ] Check whether a foreign key constraint on the new integer `created_by`/`last_updated_by` columns referencing `users.id` should be added to prevent orphaned attribution records.

## File-by-file

### `migrations/1777033895000-address-book-attribution-to-user-id.ts`

Adds migration `AddressBookAttributionToUserId1777033895000` that converts `space_address_book_items.created_by` and `last_updated_by` from `varchar(42)` wallet addresses to `integer` user IDs, with a pre-flight guard, backfill via `wallets` table JOIN, drop of old columns, rename, and NOT NULL enforcement. The `down` migration reverses the process in a lossy, best-effort manner.

_Enables OIDC-authenticated users (who have no wallet address) to be attributed as authors of address book entries by switching the attribution key to `users.id`._

**Risks:**
- Pre-flight check uses OR instead of AND: a row where only one of created_by/last_updated_by is unresolvable may not be caught correctly — the subquery logic `NOT EXISTS(...) OR NOT EXISTS(...)` is evaluated per-row but the outer EXISTS wraps the whole set, so a row failing only one column will still trigger, but verify the SQL logic is correct under all edge cases
- No foreign key constraint added on the new integer columns referencing `users.id` — orphaned user IDs could accumulate silently
- Backfill and NOT NULL are separate statements with no transaction wrapping; a failure between backfill and NOT NULL could leave nullable columns without the constraint
- Down migration picks an arbitrary wallet (LIMIT 1) for multi-wallet users and leaves NULLs for OIDC entries — downstream code expecting NOT NULL varchar after rollback will break
- Migration timestamp 1777033895000 is far in the future (year 2026+); verify this doesn't conflict with ordering assumptions in the migration runner
- No index added on the new integer `created_by`/`last_updated_by` columns if queries filter by them
- If the table is large, the multi-step ALTER/UPDATE sequence without explicit transaction could leave the table in an inconsistent state on partial failure

### `src/modules/spaces/datasources/entities/address-book-item.entity.db.ts`

In `AddressBookItem` entity, `createdBy` and `lastUpdatedBy` columns changed from `varchar(42)` with `databaseAddressTransformer` (typed as `Address`) to `integer` (typed as `number`), removing the address transformer.

_Stores user IDs instead of wallet addresses for attribution, enabling OIDC users (who have no wallet address) to create/update address book entries._

**Risks:**
- DomainAddressBookItem interface must also be updated to reflect number type or a type mismatch will exist at compile time
- Any existing serialization/deserialization code that expects createdBy/lastUpdatedBy to be an address string will break silently if not updated
- The database migration must guarantee no NULL or unresolvable rows remain before the column type change, otherwise the NOT NULL constraint will fail
- Downstream API consumers receive numeric user ID strings instead of wallet addresses — breaking change with no versioning guard visible here
- Removal of databaseAddressTransformer means no normalization (e.g., checksum casing) is applied; verify no code paths still pass an address to these fields

### `src/modules/spaces/domain/address-books/address-book-items.repository.integration.spec.ts`

Integration tests updated to replace `authPayload.signer_address` (wallet address string) with `Number(authPayload.sub)` (user ID integer) for `createdBy`/`lastUpdatedBy` fields, and two new OIDC-specific test helpers (`createOidcUser`, `createSpaceAsOidcAdmin`) plus two new OIDC test cases (upsert and delete) were added.

_Validates that address book write operations work correctly for OIDC users who have no wallet address, using numeric user IDs for attribution instead._

**Risks:**
- createSpaceAsOidcAdmin uses `space.generatedMaps[0]` for both the member insert and the returned spaceId — verify this map reliably contains `id` for the TypeORM insert result
- `Number(authPayload.sub)` silently returns NaN if `sub` is not a numeric string; no assertion guards against this in the helpers
- createOidcUser uses `user.generatedMaps[0].id` but also returns `user.raw[0]` — confirm `raw[0]` is a valid User entity with an `id` property for the member insert
- invitedBy in createSpaceAsOidcAdmin is set to a random Ethereum address, which may be inconsistent with the OIDC-only intent and could mask bugs if invitedBy is later validated
- The renamed test 'should throw ForbiddenException' → 'should throw NotFoundException' should be verified against actual repository behavior to ensure the exception type is correct

### `src/modules/spaces/domain/address-books/address-book-items.repository.ts`

Removed `isSiwe()` guards and `ForbiddenException` from `updateExistingAddressBookItems` and `createNewAddressBookItems`; replaced `authPayload.signer_address` (wallet address string) with `userId` (number from `getAuthenticatedUserIdOrFail`) for `createdBy`/`lastUpdatedBy` fields. `getAuthenticatedUserIdOrFail` is now called once in `upsertAddressBookItems` and the resolved `userId` is threaded down.

_Allows OIDC-authenticated users (who have no wallet address) to write address book entries by attributing writes to the numeric user ID instead of a wallet address._

**Risks:**
- Verify `createdBy`/`lastUpdatedBy` DB columns are now integer type (migration applied) before this code runs, otherwise type mismatch at insert/update
- Check that `getAuthenticatedUserIdOrFail` throws appropriately for unauthenticated callers — the old guards were redundant but explicit; confirm no regression in auth enforcement
- API consumers reading `createdBy`/`lastUpdatedBy` as wallet address strings will break — confirm client-side changes are coordinated
- Down migration is lossy for OIDC entries; verify rollback strategy is acceptable
- deleteAddressBookItems path (not shown in diff) should also be checked for any remaining `isSiwe()` guards or `signer_address` usage

### `src/modules/spaces/domain/address-books/entities/__tests__/address-book-item.db.builder.ts`

In `addressBookItemBuilder`, `createdBy` and `lastUpdatedBy` fields changed from `getAddress(faker.finance.ethereumAddress())` (wallet address strings) to `faker.number.int({ min: 1, max: DB_MAX_SAFE_INTEGER })` (integer user IDs). A license header was also added.

_Aligns the test builder with the schema migration that changes attribution columns from varchar wallet addresses to integer user IDs, ensuring tests reflect the new data shape._

**Risks:**
- DB_MAX_SAFE_INTEGER import should be verified to match the actual PostgreSQL integer max (2147483647) to avoid generating out-of-range values
- Any existing tests relying on the old builder that assert createdBy/lastUpdatedBy as address strings will silently pass wrong types if not updated
- min: 1 assumes user IDs are never 0 — verify this constraint holds in the actual schema/domain

### `src/modules/spaces/domain/address-books/entities/address-book-item.db.entity.ts`

In `AddressBookDbItemSchema`, `createdBy` and `lastUpdatedBy` fields changed from `AddressSchema as z.ZodType<Address>` (wallet address strings) to `z.number().int()` (integer user IDs), with corresponding TypeScript type changes from `Address` to `number`.

_Aligns the DB entity schema with the migration that replaces varchar wallet address columns with integer user ID columns, enabling OIDC users (who have no wallet address) to be attributed as authors._

**Risks:**
- Any existing code reading createdBy/lastUpdatedBy as Address strings will now receive numbers — check all consumers of AddressBookDbItem for type compatibility
- The SPDX license header addition is unrelated to the functional change — verify it's intentional and consistent with other files
- Ensure the z.number().int() validation correctly rejects non-integer or null values that might exist during/after migration backfill

### `src/modules/spaces/routes/address-books.service.spec.ts`

Unit tests for `AddressBooksService` replace `ForbiddenException` OIDC-blocked tests with `it.each` parameterized tests covering both SIWE and OIDC users for `upsertMany` and `deleteByAddress`, and replace unauthenticated-user `ForbiddenException` assertions with `UnauthorizedException` propagation tests. A new test verifies `createdBy`/`lastUpdatedBy` are serialized as strings.

_Reflects the removal of `isSiwe()` guards that previously blocked OIDC users from write operations, confirming both auth types now succeed and attribution fields are numeric-user-ID-based strings._

**Risks:**
- The unauthenticated-user tests now mock the repository to throw `UnauthorizedException` rather than asserting the repository is never called — verify the service actually calls `getAuthenticatedUserIdOrFail` before the repository, otherwise the guard logic is untested
- The `createdBy` serialization test uses `siweAuthPayloadDtoBuilder` but the feature is meant to use user IDs for both auth types; an OIDC variant of this assertion is missing
- No assertion that `repositoryMock.upsertMany` / `repositoryMock.deleteByAddress` is NOT called when unauthenticated — if the guard was moved into the repository layer, a unit test at the service layer may give false confidence

### `src/modules/spaces/routes/address-books.service.ts`

Removed two `isSiwe()` guard checks (and the `ForbiddenException` import) from `upsertAddressBookItems` and `deleteByAddress` in `AddressBooksService`, and changed `createdBy`/`lastUpdatedBy` mapping to call `.toString()` on what are now numeric user IDs instead of wallet address strings.

_Allows OIDC-authenticated users (who have no wallet address) to perform address book writes, since attribution is now based on numeric user IDs rather than signer addresses._

**Risks:**
- If `item.createdBy` or `item.lastUpdatedBy` can be null/undefined, calling `.toString()` will throw at runtime — null safety should be verified
- Removing the SIWE guard is a security boundary change; confirm that downstream repository methods (`upsertMany`, `deleteByAddress`) still enforce membership/authorization checks for both SIWE and OIDC users
- API breaking change: consumers of `createdBy`/`lastUpdatedBy` fields will now receive a numeric string instead of a wallet address — verify all clients are updated
- No fallback if `getAuthenticatedUserIdOrFail` fails inside the repository — confirm error propagation is handled correctly

### `src/modules/spaces/routes/entities/space-address-book.dto.entity.ts`

In `SpaceAddressBookItemDto`, `createdBy` and `lastUpdatedBy` fields are decoupled from `AddressBookDbItem`'s type (previously typed as `AddressBookDbItem['createdBy']`/`['lastUpdatedBy']`) and explicitly typed as `string`, with Swagger descriptions updated to 'User ID of the creator/last updater'. The `AddressBookDbItem` import is also changed to a type-only import.

_Reflects the migration of `createdBy`/`lastUpdatedBy` from wallet address strings to numeric user ID strings, making the DTO independent of the DB entity's column type so the API contract is explicit._

**Risks:**
- The field type is `string` but user IDs are numeric integers — callers may expect a number; verify serialization converts int to string consistently
- Decoupling from `AddressBookDbItem['createdBy']` means future DB type changes won't propagate to the DTO automatically — could cause silent drift
- No validation decorator (e.g. `@IsString`, `@IsNumberString`) added to enforce the new format at runtime

---
_Generated by prbrief • model: claude-sonnet-4-6 • 11 LLM calls • $0.20 • 99.4s_

</details>

<sub>Generated by <a href="https://popcorn.report">Popcorn Review</a>. · `safe-global/safe-client-gateway@2fa1fbad238d`</sub>

src/modules/spaces/domain/address-books/address-book-items.repository.ts
```diff
@@ -53,6 +48,7 @@ export class AddressBookItemsRepository implements IAddressBookItemsRepository {
     spaceId: Space['id'];
     addressBookItems: UpsertAddressBookItemsDto['items'];
   }): Promise<Array<AddressBookDbItem>> {
+    const userId = getAuthenticatedUserIdOrFail(args.authPayload);
     const space = await this.getSpaceAs({
       ...args,
```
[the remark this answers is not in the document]
0xAryanShah: sub is UserId Value, which is always a integer

migrations/1777033895000-address-book-attribution-to-user-id.ts
```diff
@@ -0,0 +1,118 @@
...
+      FROM wallets w
+      WHERE w.address = sabi.last_updated_by;
+    `);
+
+    // Drop old varchar columns and rename new ones.
+    await queryRunner.query(`
+      ALTER TABLE space_address_book_items
+        DROP COLUMN created_by,
+        DROP COLUMN last_updated_by;
+    `);
```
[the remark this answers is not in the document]
0xAryanShah: Added : Here [1d0cbb3](https://github.com/safe-global/safe-client-gateway/pull/3056/commits/1d0cbb39fa7aa99ee89081b595dc4acdc96c3c01)[1d0cbb3](https://github.com/safe-global/safe-client-gateway/pull/3056/commits/1d0cbb39fa7aa99ee89081b595dc4acdc96c3c01)[1d0cbb3](https://github.com/safe-global/safe-client-gateway/pull/3056/commits/1d0cbb39fa7aa99ee89081b595dc4acdc96c3c01)

src/modules/spaces/domain/address-books/entities/address-book-item.db.entity.ts
```diff
@@ -14,16 +15,16 @@ export const AddressBookDbItemSchema: z.ZodType<
...
   }
 > = RowSchema.extend({
   space: z.lazy(() => SpaceSchema),
   chainIds: z.array(z.string()),
   address: AddressSchema as z.ZodType<Address>,
   name: makeNameSchema({ maxLength: ADDRESS_BOOK_NAME_MAX_LENGTH }),
-  createdBy: AddressSchema as z.ZodType<Address>,
-  lastUpdatedBy: AddressSchema as z.ZodType<Address>,
+  createdBy: z.number().int(),
+  lastUpdatedBy: z.number().int(),
```
LucieFaire: should prob be positive numbers

src/modules/spaces/domain/address-books/address-book-items.repository.integration.spec.ts
```diff
@@ -362,7 +368,7 @@ describe('AddressBookItemsRepository', () => {
       ).rejects.toThrow(new NotFoundException('Space not found.'));
     });
 
-    it('should throw ForbiddenException if the user is not an admin', async () => {
+    it('should throw NotFoundException if the user is not an admin', async () => {
```
LucieFaire, line 375: is the description correct?

src/modules/spaces/routes/entities/space-address-book.dto.entity.ts
```diff
@@ -12,11 +12,11 @@ export class SpaceAddressBookItemDto {
...
   public chainIds!: AddressBookDbItem['chainIds'];
 
-  @ApiProperty({ type: String })
-  public createdBy!: AddressBookDbItem['createdBy'];
+  @ApiProperty({ type: String, description: 'User ID of the creator' })
+  public createdBy!: string;
 
-  @ApiProperty({ type: String })
-  public lastUpdatedBy!: AddressBookDbItem['lastUpdatedBy'];
+  @ApiProperty({ type: String, description: 'User ID of the last updater' })
```
LucieFaire: maybe `User ID of the last editor`

migrations/1777033895000-address-book-attribution-to-user-id.ts
```diff
@@ -0,0 +1,134 @@
...
+    await queryRunner.query(`
+      DO $$
+      BEGIN
+        IF EXISTS (
+          SELECT 1 FROM space_address_book_items sabi
+          WHERE NOT EXISTS (
+            SELECT 1 FROM wallets w WHERE w.address = sabi.created_by
+          )
+          OR NOT EXISTS (
+            SELECT 1 FROM wallets w WHERE w.address = sabi.last_updated_by
```
LucieFaire: same as in the other PR: address case sensitivity

src/modules/spaces/routes/address-books.service.ts
```diff
@@ -77,8 +65,8 @@ export class AddressBooksService {
       name: item.name,
       address: item.address,
       chainIds: item.chainIds,
-      createdBy: item.createdBy,
-      lastUpdatedBy: item.lastUpdatedBy,
+      createdBy: item.createdBy.toString(),
+      lastUpdatedBy: item.lastUpdatedBy.toString(),
```
LucieFaire: I understand why we need to convert them, so we dont break the wallet. but do we use these fields in the wallet? Are they displayed anywhere?

LucieFaire:
@claude review

src/modules/spaces/routes/address-books.service.spec.ts
```diff
@@ -90,36 +112,162 @@ describe('AddressBooksService', () => {
...
+
+      const result = await service.findAllBySpaceId(authPayload, spaceId);
+
+      expect(result.data[0].createdBy).toBe(user1Email);
+      expect(result.data[0].lastUpdatedBy).toBe(user2Wallet.address);
+      expect(result.data[1].createdBy).toBe(user2Wallet.address);
+      expect(result.data[1].lastUpdatedBy).toBe(user1Email);
+    });
+
+    it('should return "Unknown user" when user exists but has no wallet or email', async () => {
```
vseehausen, line 217: Under which circumstances will this happen?

vseehausen (approved):
LGMT. 

But to clarify: 

* Do we want admins with `INVITED` status to be able to write, and members with `INVITED` status to read? (this was preexisting, but could be worth a discussion to remove access for `INVITED` members).

src/modules/spaces/routes/address-books.service.spec.ts
LucieFaire: email is not verified so its not stored. Should not happen but might for whatever reason

src/modules/spaces/routes/entities/space-address-book.dto.entity.ts
```diff
@@ -13,11 +14,31 @@ export class SpaceAddressBookItemDto {
...
 
-  @ApiProperty({ type: String })
-  public createdBy!: AddressBookDbItem['createdBy'];
-
-  @ApiProperty({ type: String })
-  public lastUpdatedBy!: AddressBookDbItem['lastUpdatedBy'];
+  @ApiProperty({
+    type: String,
+    description:
+      'Email or wallet address of the creator, "Unknown user" if the user has no display identity, or "Deleted user"',
```
vseehausen, line 20: Why not use member.name here, too?

vseehausen (approved):

PR #3058, opened 2026-04-27
chore: mark POST /v1/spaces/create-with-user as deprecated

LucieFaire:
## Summary 
[WA-1914](https://linear.app/safe-global/issue/WA-1914/deprecate-create-with-user-endpoint-update-frontend-to-use-spaces-post)

`POST /v1/spaces/create-with-user` is identical to `POST /v1/spaces`, so we are deprecating the first one

## Changes
Add Swagger `deprecateed` marker


0xAryanShah (approved):
👍

PR #3059, opened 2026-04-27
feat: updated space member invitedBy field to be userId

0xAryanShah:
## Summary
- Migrates `members.invited_by` column from `varchar(42)` (wallet address) to `integer` (user ID), enabling OIDC admins to be recorded as inviters
- Adds `invitedByName` field to the `GetSpaceResponse` DTO, resolved at query time from wallet address (preferred) or email (OIDC fallback)
- Fixes `isLimited` space creation check — OIDC-invited memberships no longer incorrectly count toward the limit (previously `invitedBy` was `NULL` for OIDC invites, matching the space-creator condition)
- Adds index on `wallets.user_id` to support the new lookup pattern

## Breaking changes
- `invitedBy` response type changed from `string | null` (address) to `number | null` (user ID) across `GetSpaceResponse`, `MemberDto`, and `Invitation` DTOs
- New optional `invitedByName: string` field added to `GetSpaceResponse` members (populated only for INVITED members whose inviter is still in the space)


0xAryanShah:
@claude review

0xAryanShah:
1-4) are By Design
5) Added here: [3be683e](https://github.com/safe-global/safe-client-gateway/pull/3059/commits/3be683e9ae27ed72df71ee623cf448c45b4931dc)
6) The invitedBy value in those helpers is irrelevant to what the address book tests verify, they test address book CRUD, not invite behavior.

LucieFaire:
@claude review again

migrations/1777274846000-migrate-members-invited-by.ts
```diff
@@ -0,0 +1,80 @@
...
+    await queryRunner.query(`
+      DO $$
+      DECLARE unresolvable INTEGER;
+      BEGIN
+        SELECT count(*) INTO unresolvable
+        FROM "members"
+        WHERE "invited_by" IS NOT NULL
+          AND NOT EXISTS (
+            SELECT 1 FROM "wallets"
+            WHERE "wallets"."address" = "members"."invited_by"
```
LucieFaire: The `wallets.address` column uses `databaseAddressTransformer` which calls `getAddress()`. But `members.invited_by` might not have been stored the same way. So this step is case-sensitive:
```
WHERE "wallets"."address" = "members"."invited_by"
```
Thus the pre-flight check will flag such cases as unresolvable and the migration will abort — even though they're the same address.

migrations/1777274846000-migrate-members-invited-by.ts
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class MigrateMembersInvitedBy1777274846000 implements MigrationInterface {
+  name = 'MigrateMembersInvitedBy1777274846000';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    // 1. Add new nullable integer column
+    await queryRunner.query(
+      `ALTER TABLE "members" ADD COLUMN "invited_by_user_id" INTEGER NULL`,
```
LucieFaire, line 12: qq: why by design we have decided to not FK to users table and set to null in case of deletion?

LucieFaire (commented):
1 remark and 1 qs, otherwise looks good

vseehausen:
@claude review

LucieFaire:
1. that is a tradeoff with the current setup
2. decided againsdt FK as it doesnt serve any purpose
3. tradeoff to compute it twice so we dont overfetch from DB
4. added an extra comment
5. exposure is expected per product reqs
6. known limitation - we only have 1 user - 1 wallet so far
7. added a test

src/modules/spaces/routes/spaces.service.ts
```diff
@@ -38,13 +41,35 @@ export class SpacesService {
...
   }
 
-  public async getActiveOrInvitedSpaces(
+  public getActiveOrInvitedSpaces(
     authPayload: AuthPayload,
+  ): Promise<Array<GetSpaceResponse>> {
+    return this.findSpaces(authPayload);
+  }
+
+  public async getActiveOrInvitedSpace(
```
PooyaRaki, line 50: To be consistent with the naming, we should name this one:
`getActiveOrInvitedSpaceOrFail`

src/modules/users/domain/users.repository.ts
```diff
@@ -347,6 +347,25 @@ export class UsersRepository implements IUsersRepository {
     return user?.email ?? undefined;
   }
 
+  public async findEmailsByIds(
+    userIds: Array<User['id']>,
+  ): Promise<Map<User['id'], string>> {
+    if (userIds.length === 0) return new Map();
```
PooyaRaki: We should not create a new map when `userIds` is null. we can return `null`

src/modules/users/domain/users.repository.ts
```diff
@@ -347,6 +347,25 @@ export class UsersRepository implements IUsersRepository {
...
+    const userRepository =
+      await this.postgresDatabaseService.getRepository(DbUser);
+    const users = await userRepository.find({
+      where: { id: In(userIds) },
```
PooyaRaki, line 366: We should add `email: Not(IsNull()),` here and remove the `flatMap`.

src/modules/spaces/routes/spaces.service.ts
```diff
@@ -67,24 +92,28 @@ export class SpacesService {
...
-    id: number,
-    authPayload: AuthPayload,
-  ): Promise<GetSpaceResponse> {
-    const spaces = await this.getActiveOrInvitedSpaces(authPayload);
-    const space = spaces.find((space) => space.id === id);
-    if (!space) {
-      throw new NotFoundException('Space not found.');
-    }
-    return space;
+    return spaces.map((space) => {
```
PooyaRaki, line 97: Why do we need to do all the mapping in the code? Couldn't we just fetch the structure we want to return directly from the database?

src/modules/spaces/routes/spaces.service.ts
```diff
@@ -98,6 +127,65 @@ export class SpacesService {
...
+    spaces: Array<Space>,
+  ): Promise<Map<number, string>> {
+    const userIds = Array.from(
+      new Set(
+        spaces.flatMap((space) => {
+          const memberUserIds = new Set(
+            space.members.map((member) => member.user.id),
+          );
+
+          return space.members.flatMap((member): Array<number> => {
```
PooyaRaki, line 151: Can't we just query the database and return what we want directly from there instead of traversing spaces one by one here?

src/modules/users/domain/users.repository.integration.spec.ts
```diff
@@ -1153,4 +1153,64 @@ describe('UsersRepository', () => {
...
+      const email2 = faker.internet.email().toLowerCase();
+      const user1 = await dbUserRepository.insert({
+        status: 'ACTIVE',
+        email: email1,
+      });
+      const user2 = await dbUserRepository.insert({
+        status: 'ACTIVE',
+        email: email2,
+      });
+      const userId1 = user1.identifiers[0].id as number;
```
PooyaRaki, line 1132: Why do we need to cast to number? Doesn't TypeOrm return the id with the expected type?

src/modules/spaces/routes/spaces.service.ts
LucieFaire: I thought we only use those naming conventions for repository methods?

src/modules/users/domain/users.repository.ts
LucieFaire: hm, not sure we win much here.

`email` is typed as `EmailAddress | null` at `users.entity.db.ts:58`. TypeORM doesn't narrow result types based on the where clause, so even with `email: Not(IsNull())` the returned user.email is still typed EmailAddress | null. We would need either a non-null assertion or to keep some kind of filter:

```
const users = await userRepository.find({
  where: { id: In(userIds), email: Not(IsNull()) },
  select: { id: true, email: true },
});
return new Map(users.map((user) => [user.id, user.email!]));  // ← needs `!`
```

src/modules/spaces/routes/spaces.service.ts
LucieFaire: we could but that would make SQL more complex for this part. We only really map the `invitedByName` fields, and that data is fetched separately from the main spaces

src/modules/users/domain/users.repository.integration.spec.ts
LucieFaire: It's a TypeORM type ergonomics quirk. `InsertResult.identifiers` is typed as `ObjectLiteral[]` — and in TypeORM it is `{ [key: string]: any }`. So `identifiers[0].id comes back` as `any`, even though PG actually returns the generated number at runtime.

src/modules/spaces/routes/spaces.service.ts
LucieFaire: The flatMap isn't "traversing spaces one by one" against the database — it's filtering data we already have. Adding a DB query to compute the same thing would be a redundant round-trip for inputs already loaded, plus it would require us to write a raw SQL to do a join. Not sure its a better approach here

src/modules/users/domain/users.repository.ts
LucieFaire: addressed [c20db53](https://github.com/safe-global/safe-client-gateway/pull/3059/commits/c20db53f790af3203117104bc386b4155f89ee01)

src/modules/spaces/routes/entities/get-space.dto.entity.ts
```diff
@@ -22,9 +22,12 @@ class SpaceMemberDto {
   @ApiProperty({ type: String })
   public name!: Member['name'];
 
-  @ApiProperty({ type: String })
+  @ApiProperty({ type: Number, nullable: true })
   public invitedBy!: Member['invitedBy'];
 
+  @ApiPropertyOptional({ type: String })
+  public invitedByName?: string;
```
vseehausen, line 29: Is `invitedByIdentifier` more accurate? This avoids confusion with the member name.

LucieFaire: i would prefer to keep it as is at this point, changing it will affect both this PR and the FE and a lot of time was spent here as it is

LucieFaire: we can revisit the field names post release its not urgent

vseehausen: Maybe we could actually return both address and email (and member name) and let the client decide which one to display.

vseehausen (approved):

PR #3060, opened 2026-04-27
Fix/relay fee relayer safetxhash

akshay-ap:
## Summary

Closes: https://linear.app/safe-global/issue/PLA-1268/cgw-bug-relay-fee-relayer-does-not-validate-to-data-value-params

Closes: https://linear.app/safe-global/issue/PLA-1277/cgw-allow-relaying-safe-creation-tx-for-gtf-enabled-chains

The relay-fee relayer was sponsoring transactions based solely on the fee service's `canRelay` response, without verifying that the submitted `to`/`data` actually corresponds to the claimed `safeTxHash`. An attacker could construct calldata that the fee service approves (via a legitimate hash) but execute different transaction parameters. This PR fixes that by verifying the `safeTxHash` on-chain before sponsoring, and consolidates the relay validation logic that was duplicated between `LimitAddressesMapper` and the relay-fee path.

### On-chain `safeTxHash` verification in `relay-fee` relayer

`RelayFeeRelayer.relay()` now dispatches on transaction type before calling the fee service:

- **`execTransaction`**: requires `safeTxHash`. Calls `getTransactionHash` on the Safe contract with the decoded calldata args and the current on-chain nonce, then compares against the submitted hash. Mismatch throws `SafeTxHashMismatchError` (HTTP 422) before the fee service is ever called.
- **`createProxyWithNonce`**: verifies the proxy factory is an official deployment.
- **Anything else**: denied with `RelayTxDeniedError` (HTTP 403).

`SafeTxHashMismatchError` and its exception filter are new; the filter returns HTTP 422 and is registered on `RelayController`.

### `RelayTransactionHelper`

All transaction-inspection logic previously in `LimitAddressesMapper` (`isValidExecTransactionCall`, `isOfficialMastercopy`, `isMultiSend`, `isOfficialMultiSendDeployment`, `getSafeAddressFromMultiSend`, `isOfficialProxyFactoryDeployment`, `isValidCreateProxyWithNonceCall`, `getOwnersFromCreateProxyWithNonce`, `getSafeBeingRecovered`) has been extracted into a dedicated `RelayTransactionHelper` injectable. `isSafeTxHashValid` is added here as the new on-chain verification method.

`LimitAddressesMapper` now delegates entirely to `RelayTransactionHelper`, eliminating the duplicate decoders and RPC dependencies it previously held directly.

### `SafeTransaction*` entities

`SafeTransaction` and `SafeTransactionWithSignature` type added to capture the full `execTransaction` argument tuple (including `signatures`), enabling the `safe-encoder.builder` test helper to be typed against the canonical entity rather than a locally-defined duplicate.

### DIFF
--- a/src/modules/relay/domain/relay-transaction-helper.ts
+++ b/src/modules/relay/domain/relay-transaction-helper.ts
@@ -1,0 +1,540 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { Inject, Injectable } from '@nestjs/common';
+import semverSatisfies from 'semver/functions/satisfies';
+import type { Address, Hex } from 'viem';
+import {
+  encodeAbiParameters,
+  getAddress,
+  keccak256,
+  parseAbiParameters,
+} from 'viem';
+import Safe130 from '@/abis/safe/v1.3.0/GnosisSafe.abi';
+import Safe141 from '@/abis/safe/v1.4.1/Safe.abi';
+import Safe150 from '@/abis/safe/v1.5.0/Safe.abi';
+import { LogType } from '@/domain/common/entities/log-type.entity';
+import {
+  getMultiSendCallOnlyDeployments,
+  getMultiSendDeployments,
+  getProxyFactoryDeployments,
+  getSafeL2SingletonDeployments,
+  getSafeSingletonDeployments,
+  getSignerFactoryDeployments,
+} from '@/domain/common/utils/deployments';
+import { IBlockchainApiManager } from '@/domain/interfaces/blockchain-api.manager.interface';
+import { ILoggingService, LoggingService } from '@/logging/logging.interface';
+import { DelayModifierDecoder } from '@/modules/alerts/domain/contracts/decoders/delay-modifier-decoder.helper';
+import { MultiSendDecoder } from '@/modules/contracts/domain/decoders/multi-send-decoder.helper';
+import { SafeDecoder } from '@/modules/contracts/domain/decoders/safe-decoder.helper';
+import { Erc20Decoder } from '@/modules/relay/domain/contracts/decoders/erc-20-decoder.helper';
+import { ProxyFactoryDecoder } from '@/modules/relay/domain/contracts/decoders/proxy-factory-decoder.helper';
+import { SignerFactoryDecoder } from '@/modules/relay/domain/contracts/decoders/signer-factory-decoder.helper';
+import { InvalidMultiSendError } from '@/modules/relay/domain/errors/invalid-multisend.error';
+import { ISafeRepository } from '@/modules/safe/domain/safe.repository.interface';
+import type { SafeTransaction } from '@/modules/transactions/domain/entities/safe-transaction.entity';
+
+type SafeAbi = typeof Safe130 | typeof Safe141 | typeof Safe150;
+
+@Injectable()
+export class RelayTransactionHelper {
+  constructor(
+    @Inject(ISafeRepository)
+    private readonly safeRepository: ISafeRepository,
+    @Inject(LoggingService)
+    private readonly loggingService: ILoggingService,
+    @Inject(IBlockchainApiManager)
+    private readonly blockchainApiManager: IBlockchainApiManager,
+    private readonly erc20Decoder: Erc20Decoder,
+    private readonly safeDecoder: SafeDecoder,
+    private readonly multiSendDecoder: MultiSendDecoder,
+    private readonly proxyFactoryDecoder: ProxyFactoryDecoder,
+    private readonly delayModifierDecoder: DelayModifierDecoder,
+    private readonly signerFactoryDecoder: SignerFactoryDecoder,
+  ) {}
+
+  private getSafeAbi(version: string): SafeAbi {
+    if (semverSatisfies(version, '>=1.5.0')) return Safe150;
+    if (semverSatisfies(version, '>=1.4.1')) return Safe141;
+    if (semverSatisfies(version, '>=1.0.0')) return Safe130;
+    this.loggingService.warn({
+      type: LogType.TxRelayEligibility,
+      message: `getSafeAbi: unrecognised Safe version ${version}, falling back to 1.3.0 ABI`,
+    });
+    return Safe130;
+  }
+
+  isValidExecTransactionCall(args: { to: Address; data: Hex }): boolean {
+    const decoded = this.decodeExecTransaction(args.data);
+    if (!decoded) return false;
+    return this.isValidDecodedExecTransaction({ to: args.to, decoded });
+  }
+
+  isValidDecodedExecTransaction(args: {
+    to: Address;
+    decoded: SafeTransaction;
+  }): boolean {
+    const { decoded } = args;
+
+    // Only ERC-20 transfer to other party is valid
+    if (this.erc20Decoder.helpers.isTransfer(decoded.data)) {
+      return this.isValidErc20Transfer({ to: args.to, data: decoded.data });
+    }
+
+    // Only ERC-20 transferFrom to other party is valid
+    if (this.erc20Decoder.helpers.isTransferFrom(decoded.data)) {
+      return this.isValidErc20TransferFrom({ to: args.to, data: decoded.data });
+    }
+
+    const toSelf = decoded.to === args.to;
+    if (!toSelf) {
+      return true;
+    }
+
+    return this.isValidSelfTransaction(decoded);
+  }
+
+  private isValidSelfTransaction(decoded: SafeTransaction): boolean {
+    // Only transaction with no value is valid
+    const hasValue = decoded.value > BigInt(0);
+    if (hasValue) {
+      return false;
+    }
+
+    // Cancellations and Safe calls (e.g. owner management) are valid
+    const isCancellation = decoded.data === '0x';
+    return isCancellation || this.safeDecoder.isCall(decoded.data);
+  }
+
+  decodeExecTransaction(data: Hex): SafeTransaction | null {
+    try {
+      const safeDecodedData = this.safeDecoder.decodeFunctionData({
+        data,
+      });
+
+      if (safeDecodedData.functionName !== 'execTransaction') {
+        return null;
+      }
+
+      const [
+        to,
+        value,
+        innerData,
+        operation,
+        safeTxGas,
+        baseGas,
+        gasPrice,
+        gasToken,
+        refundReceiver,
+      ] = safeDecodedData.args;
+
+      return {
+        to,
+        value,
+        data: innerData,
+        operation,
+        safeTxGas,
+        baseGas,
+        gasPrice,
+        gasToken,
+        refundReceiver,
+      };
+    } catch {
+      return null;
+    }
+  }
+
+  private isValidErc20Transfer(args: { to: Address; data: Hex }): boolean {
+    // Can throw but called after this.erc20Decoder.helpers.isTransfer
+    const erc20DecodedData = this.erc20Decoder.decodeFunctionData({
+      data: args.data,
+    });
+
+    if (erc20DecodedData.functionName !== 'transfer') {
+      return false;
+    }
+
+    const [to] = erc20DecodedData.args;
+    // to 'self' (the Safe) is not allowed
+    return to !== args.to;
+  }
+
+  private isValidErc20TransferFrom(args: { to: Address; data: Hex }): boolean {
+    // Can throw but called after this.erc20Decoder.helpers.isTransferFrom
+    const erc20DecodedData = this.erc20Decoder.decodeFunctionData({
+      data: args.data,
+    });
+
+    if (erc20DecodedData.functionName !== 'transferFrom') {
+      return false;
+    }
+
+    const [sender, recipient] = erc20DecodedData.args;
+    // to 'self' (the Safe) or from sender to sender as recipient is not allowed
+    return sender !== recipient && recipient !== args.to;
+  }
+
+  async isOfficialMastercopy(args: {
+    chainId: string;
+    address: Address;
+  }): Promise<boolean> {
+    try {
+      await this.safeRepository.getSafe(args);
+      return true;
+    } catch {
+      return false;
+    }
+  }
+
+  isMultiSend(data: Hex): boolean {
+    return this.multiSendDecoder.helpers.isMultiSend(data);
+  }
+
+  isOfficialMultiSendDeployment(args: {
+    version: string;
+    chainId: string;
+    address: Address;
+  }): boolean {
+    return (
+      getMultiSendCallOnlyDeployments(args).includes(args.address) ||
+      getMultiSendDeployments(args).includes(args.address)
+    );
+  }
+
+  getSafeAddressFromMultiSend(data: Hex): Address {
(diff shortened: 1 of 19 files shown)


akshay-ap:
@claude review

akshay-ap:
> **Claude finished @akshay-ap's task** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25047595285)
> 
> ### Code Review: PR #3060 — Fix relay-fee relayer safeTxHash validation
> Overall this is a well-structured security fix. The refactoring to extract `RelayTransactionHelper` and adding on-chain `safeTxHash` verification addresses a real vulnerability. Below are the items worth attention:
> 
> #### Security
> 1. **`createProxyWithNonce` path skips singleton validation in relay-fee relayer** (`relay-fee.relayer.ts:129-164`): The `relay()` method checks `isValidCreateProxyWithNonceCall` and `isOfficialProxyFactoryDeployment`, but unlike `LimitAddressesMapper`, it does **not** verify that the singleton used in `createProxyWithNonce` is an official deployment. `isValidCreateProxyWithNonceCall` does check the singleton, so this is implicitly covered — but it's worth a comment clarifying that, since the two validation paths look asymmetric.
> 2. **Safe creation without `safeTxHash`** (`relay-fee.relayer.ts:150-164`): When `createProxyWithNonce` is called without a `safeTxHash`, the transaction is relayed with **no fee service check at all**. This is documented as intentional ("fee service decides when hash is available"), but this means the relay-fee path for Safe creation is essentially open if no hash is provided. Consider whether this should at minimum log a warning, or whether `safeTxHash` should be required for this path too.
> 
> #### Potential Bugs / Correctness
> 3. **`data` parameter typed as `Address` instead of `Hex`** (`relay-transaction-helper.ts:56, 96, 134, 179, etc.`): Many parameters that represent calldata (`data`) are typed as `Address` (a 20-byte hex) rather than `Hex`. This is inherited from the old code and won't cause runtime issues (both are `0x`-prefixed strings), but it's semantically misleading and could cause confusion for future contributors. Consider a follow-up to fix these types.
> 4. **`SafeTxHashMismatchError` extends `UnprocessableEntityException`** (`safe-tx-hash-mismatch.error.ts:5`) **but also has a custom exception filter** (`safe-tx-hash-mismatch.exception-filter.ts`): Since it already extends `UnprocessableEntityException`, NestJS's built-in exception handling would already return 422. The custom filter is redundant unless you plan to customize the response shape beyond what `UnprocessableEntityException` provides. Not a bug, but unnecessary complexity.
> 
> #### Code Quality
> 5. **Naming inconsistency**: The class is named `RelayTransactionHelper` but the field referencing it in both `LimitAddressesMapper` and `RelayFeeRelayer` is named `relayTransactionValidator`. The test file is also named `relay-transaction-validator.spec.ts`. Pick one name and align across the codebase.
> 6. **`getExecTransactionArgs` is `private` but needed by `isSafeTxHashValid` and `isValidExecTransactionCall`** — this is fine since they're on the same class, but the method is called from two distinct public-facing code paths (validation vs. hash verification). If the class grows further, consider whether these warrant being separate concerns.
> 
> #### Test Coverage
> 7. **Missing test: `relay()` with `execTransaction` where fee service denies after hash validation passes** — the relay-fee relayer spec tests hash mismatch and fee service denial for proxy creation, but doesn't have a dedicated test for the case where `isSafeTxHashValid` returns `true` but `feeServiceApi.canRelay` returns `false` on the `execTransaction` path.
> 8. **`isSafeTxHashValid` test doesn't verify the ABI selection** — the test always uses version `'1.3.0'`. Consider adding a test with `'1.4.1'` or `'1.5.0'` to exercise the `getSafeAbi` branching logic, especially since this is a security-critical path.
> 
> Good work on closing this vulnerability. The refactoring to `RelayTransactionHelper` is clean and the test coverage is solid overall.
> 
> * [x]  Read repository guidelines (CLAUDE.md)
> * [x]  Read all changed files
> * [x]  Analyze code quality, bugs, security, performance, and test coverage
> * [x]  Post review feedback
>   | Branch: `fix/relay-fee-relayer-safetxhash`

1 - NA (Safe Transaction Service will reject the transaction from being added to the queue.)
2 - NA (SafeTxHash will not be available for Safe creation txs)
3 - FIxed https://github.com/safe-global/safe-client-gateway/pull/3060/commits/bb3224a94cb205cb84bd2c0b1b7aef9100c76613, https://github.com/safe-global/safe-client-gateway/pull/3060/commits/ea25c80082e68c9f8eba32fce030ecbd6faa1e16
4 - NA (Existing coding pattern)
5 - Fixed https://github.com/safe-global/safe-client-gateway/pull/3060/commits/0a8961240f178e66dd2395220594040ee06d6fad
6 - NA
7 - NA (Not needed)

akshay-ap:
@claude review

akshay-ap:
> **Claude finished @akshay-ap's task** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25051120972)
> 
> ### Code Review: PR #3060 — Fix relay-fee relayer safeTxHash validation (2nd review)
> The previous review items 1-7 have been addressed or acknowledged. This is a focused pass on the current state.
> 
> #### Security
> 1. **`isSafeTxHashValid` returns `false` on RPC errors — relay is denied** (`relay-transaction-helper.ts:462-468`): This is the correct fail-closed behavior. Good.
> 2. **Race condition between nonce read and `getTransactionHash`** (`relay-transaction-helper.ts:429-451`): The nonce is read in one call and then used in `getTransactionHash` in a second call. If the Safe's nonce increments between these two reads (e.g. another tx gets mined), the hash will mismatch and the relay will be denied. This is safe (fail-closed) but could cause legitimate transactions to be spuriously rejected. Consider using `multicall` to batch the two reads atomically, or at minimum note this in a comment so future maintainers understand the tradeoff.
> 
> #### Code Quality
> 3. **`relay-fee.relayer.ts:95-100` — decoded `execTransaction` that fails `isValidDecodedExecTransaction` falls through to `createProxyWithNonce` check**: If someone submits valid `execTransaction` calldata that doesn't pass the validity rules (e.g. ERC-20 transfer to self), the code falls through to `isValidCreateProxyWithNonceCall`, which will fail because it's not proxy creation calldata, and then falls to the `else` branch throwing `RelayTxDeniedError`. The end result is correct (denied), but the deny message says "unrecognised tx type" which is misleading — it _was_ recognized as `execTransaction` but invalid. Consider adding an explicit `else if (decoded !== null)` branch to deny with a more descriptive message.
> 4. **`limit-addresses.mapper.spec.ts:88`**: Variable is still named `relayTransactionValidator` (not `relayTransactionHelper`). Minor naming inconsistency surviving from the rename.
> 
> Overall the security fix is solid. The previous review feedback has been largely addressed. LGTM with the minor notes above.
> 
> * [x]  Read repository guidelines (CLAUDE.md)
> * [x]  Read all changed files
> * [x]  Analyze code quality, bugs, security, and test coverage
> * [x]  Post review feedback
> 
> | Branch: [`fix/relay-fee-relayer-safetxhash`](https://github.com/safe-global/safe-client-gateway/tree/fix/relay-fee-relayer-safetxhash) | [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25051120972)

1. No action needed
2. Invalid - Highly unlikely to happen
3. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3060/commits/a86f750d60e2220bb71fab2bd87729a4639be6c9
4. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3060/commits/dd011f593e7be0360c79a8286256d2477195ff0f

akshay-ap:
@claude review

akshay-ap:
> **Claude finished @akshay-ap's task** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25052780181)
> 
> ### Code Review: PR #3060 — Fix relay-fee relayer safeTxHash validation (3rd review)
> Previous review items have been addressed. The code is in good shape. One remaining item:
> 
> #### Test Coverage
> 1. **`isSafeTxHashValid` still only tested with version `'1.3.0'`** (`relay-transaction-helper.spec.ts:274, 305, 333`): The `getSafeAbi` method has three distinct branches (`>=1.5.0`, `>=1.4.0`, `>=1.0.0`) plus a fallback. All `isSafeTxHashValid` tests use `version: '1.3.0'`. Since this is the security-critical on-chain verification path, consider adding at least one test with `'1.4.1'` and one with `'1.5.0'` to confirm the correct ABI is selected for each branch. The `getSafeAbi` fallback (unrecognised version) path is also untested.
> 
> No other issues found. LGTM.
> 
> * [x]  Read repository guidelines (CLAUDE.md)
> * [x]  Read all changed files
> * [x]  Analyze code quality, bugs, security, and test coverage
> * [x]  Post review feedback
> 
> | Branch: [`fix/relay-fee-relayer-safetxhash`](https://github.com/safe-global/safe-client-gateway/tree/fix/relay-fee-relayer-safetxhash) | [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25052780181)

1. Tests are not tied to specific Safe version but choose one of the valid versions. https://github.com/safe-global/safe-client-gateway/pull/3060/commits/58841e6c2d0163535655c4b1e1f10aae3205c7a3

akshay-ap:
@claude review

src/modules/relay/domain/relay-transaction-helper.ts
```diff
@@ -0,0 +1,470 @@
...
+
+  private getSafeAbi(version: string): SafeAbi {
+    if (semverSatisfies(version, '>=1.5.0')) return Safe150;
+    if (semverSatisfies(version, '>=1.4.1')) return Safe141;
+    if (semverSatisfies(version, '>=1.0.0')) return Safe130;
+    this.loggingService.warn({
+      type: LogType.TxRelayEligibility,
+      message: `getSafeAbi: unrecognised Safe version ${version}, falling back to 1.3.0 ABI`,
+    });
+    return Safe130;
```
akshay-ap, line 62: nit: Preferred default 1.3.0 or 1.4.1?
All the ABIs for all versions are same for `execTransaction(...)` function but a differentiation here is just for good coding practise.

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -88,28 +96,146 @@ export class RelayFeeRelayer implements IRelayer {
     gasLimit: bigint | null;
     safeTxHash?: Hex;
```
akshay-ap, line 103: `safeTxHash` is now not needed because the `safeTxHash` is now derived at runtime using RPC calls (`nonce` and `getTransactionHash`).
We can simplify for interfaces and APIs by removing optional `safeTxHash` which has been introduced recently with `relay-fee.relayer.ts` or keep it and make it a convention to provide `safeTxHash` in future for any new relaying method that might come up.

One advantage of having `safeTxHash` in api request is that client (Safe{Wallet}) informs CGW about exact transaction to be relayed rather than ambiguity introduced with only `to` and `data` params.

akshay-ap:
@copilot resolve the merge conflicts in this pull request

src/modules/transactions/domain/entities/safe-transaction.entity.ts
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { z } from 'zod';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { HexSchema } from '@/validation/entities/schemas/hex.schema';
+
+// Zod does not support bigint natively, so we use string and transform
+const BigIntSchema = z.string().transform((val, ctx) => {
```
PooyaRaki: I would extract it to a shared schema:
`src/validation/entities/schemas/`

src/modules/transactions/domain/entities/safe-transaction.entity.ts
```diff
@@ -0,0 +1,32 @@
...
+    ctx.addIssue({ code: 'custom', message: 'Invalid bigint string' });
+    return z.NEVER;
+  }
+});
+
+export const SafeTransactionSchema = z.object({
+  to: AddressSchema,
+  value: BigIntSchema,
+  data: HexSchema,
+  operation: z.number().int(),
```
PooyaRaki: `src/modules/safe/domain/entities/operation.entity.ts`

src/modules/relay/domain/relayers/relay-fee.relayer.ts
```diff
@@ -133,12 +257,18 @@ export class RelayFeeRelayer implements IRelayer {
...
+    safeTxHash?: Hex;
   }): Promise<{ remaining: number; limit: number }> {
     if (!this.relayFeeConfiguration.enabledChainIds.includes(args.chainId)) {
       return { remaining: 0, limit: 0 };
     }
 
+    // Without a safeTxHash we cannot query the fee service; report optimistically
+    // since per-transaction eligibility is enforced in relay().
+    if (!args.safeTxHash) {
+      return { remaining: 1, limit: 1 };
```
PooyaRaki, line 269: Does it mean that we can always relay a `SafeCreationTx` regardless of how many Safe creations we have relayed?

src/modules/relay/domain/relay-transaction-helper.ts
```diff
@@ -0,0 +1,537 @@
...
+
+    // Only transaction with no value is valid
+    const hasValue = decoded.value > BigInt(0);
+    if (hasValue) {
+      return false;
+    }
+
+    // Cancellations and Safe calls (e.g. owner management) are valid
+    const isCancellation = decoded.data === '0x';
+    return isCancellation || this.safeDecoder.isCall(decoded.data);
```
PooyaRaki: The check on line `94` is hard to follow since it's only evaluated if the TX is to self. We could extract all these lines to another method and call it something like `isValidSelfTransfer`

src/modules/relay/domain/relayers/relay-fee.relayer.ts
akshay-ap: yes. I know it a can be costly for mainnet but a malicious actor can always use fresh EOA's if we use daily limit. We can iterate on limiting `SafeCreationTx` relays with a better approach later imo.

src/modules/relay/domain/relay-transaction-helper.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3060/commits/964398cab78857b30e9263b556dae2ceb20f72cf

src/modules/transactions/domain/entities/safe-transaction.entity.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3060/commits/34c5b7450d3d9f95415eb747b9d26a117be9c7c2

src/modules/transactions/domain/entities/safe-transaction.entity.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3060/commits/3178c3cd337dd0fe068ed06921e00ff530f7c24d

PooyaRaki (approved):

PR #3061, opened 2026-04-28
feat: relay sponsorship for SafeWebAuthnSignerFactory.createSigner

compojoom:
## Summary

[WA-2112](https://linear.app/safe-global/issue/WA-2112/cgw-allow-gas-relay-sponsorship-for)

The relay endpoint currently rejects `SafeWebAuthnSignerFactory.createSigner` with `InvalidTransferError`, so passkey signer deployment can't be gas-sponsored — a blocker for the device-bound signing onboarding flow. This PR accepts `createSigner` calls to the official factory and rate-limits them per-passkey.

## Changes

- Add `@safe-global/safe-modules-deployments` as a dependency, pinned to factory **v0.2.1** via `SUPPORTED_SIGNER_FACTORY_VERSION`. Wrap it in `src/domain/common/utils/deployments.ts` (mirrors the existing `safe-deployments` wrapper) and extend the ESLint `no-restricted-imports` rule so only the wrapper may import it.
- New `SignerFactoryDecoder` extending `AbiDecoder`, registered in `RelayDecodersModule`. ABI is loaded from the package at runtime (single source of truth) and projected onto a `parseAbi`-derived type for typed selector helpers.
- `LimitAddressesMapper.getLimitAddresses` gets a new branch that matches the `createSigner` selector, validates `to` is the official v0.2.1 factory for the chain, decodes `(x, y, verifiers)`, and returns `keccak256(abi.encode(x, y, verifiers))` truncated to 20 bytes as the rate-limit key.

### Why a hash-derived key, not the factory address

The factory address is the same constant for every user on a chain. Returning it as the rate-limit key would make every passkey deployment globally share a single daily quota — trivial to grief, hostile to legitimate spikes. Hashing the `createSigner` args gives each unique passkey its own quota.

### Tests

6 new unit tests in `limit-addresses.mapper.spec.ts` cover happy path, determinism, distinctness, anti-regression on factory-keying, unofficial-factory rejection, and malformed-args rejection. Full suite (1144 tests) passes; lint and type-check clean.

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "@nestjs/swagger": "11.2.6",
     "@nestjs/typeorm": "11.0.0",
     "@safe-global/safe-deployments": "1.37.51",
-    "@safe-global/safe-modules-deployments": "^3.0.1",
+    "@safe-global/safe-modules-deployments": "3.0.1",
     "amqp-connection-manager": "5.0.0",
     "amqplib": "0.10.8",
     "bullmq": "^5.56.0",
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
     "@nestjs/swagger": "11.2.6",
     "@nestjs/typeorm": "11.0.0",
     "@safe-global/safe-deployments": "1.37.51",
+    "@safe-global/safe-modules-deployments": "^3.0.1",
     "amqp-connection-manager": "5.0.0",
     "amqplib": "0.10.8",
     "bullmq": "^5.56.0",
--- a/src/domain/common/utils/deployments.ts
+++ b/src/domain/common/utils/deployments.ts
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
 // eslint-disable-next-line no-restricted-imports
 import {
   getMultiSendCallOnlyDeployments as _getMultiSendCallOnlyDeployments,
--- a/src/domain/common/utils/deployments.ts
+++ b/src/domain/common/utils/deployments.ts
@@ -350,10 +350,25 @@ export function getSignerFactoryDeployments(args: {
 }
 
 /**
+ * Function signatures we type-cast the upstream ABI to in {@link SignerFactoryAbi}.
+ * Verified at module load by {@link getSignerFactoryAbi} so a future package
+ * release that drops/renames a function fails fast instead of silently.
+ */
+const REQUIRED_SIGNER_FACTORY_FUNCTIONS = [
+  { name: 'createSigner', inputs: ['uint256', 'uint256', 'uint176'] },
+  { name: 'getSigner', inputs: ['uint256', 'uint256', 'uint176'] },
+] as const;
+
+/**
  * Returns the SafeWebAuthnSignerFactory ABI as published by
  * `@safe-global/safe-modules-deployments`, pinned to
  * {@link SUPPORTED_SIGNER_FACTORY_VERSION}. Throws if the package is missing
- * the deployment (which would indicate a packaging regression upstream).
+ * the deployment or if its ABI no longer contains the function signatures
+ * encoded in {@link SignerFactoryAbi}.
+ *
+ * The runtime check guards against silent drift: the package types
+ * `Deployment.abi` as `any[]`, so the type cast can't catch upstream changes
+ * by itself.
  */
 export function getSignerFactoryAbi(): SignerFactoryAbi {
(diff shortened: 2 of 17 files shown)


compojoom:
@claude review

compojoom:
@claude review

package.json
```diff
@@ -39,29 +39,30 @@
...
-    "@nestjs/core": "11.1.17",
-    "@nestjs/platform-express": "11.1.17",
+    "@nestjs/core": "11.1.14",
+    "@nestjs/platform-express": "11.1.14",
     "@nestjs/schedule": "6.0.0",
     "@nestjs/serve-static": "5.0.4",
     "@nestjs/swagger": "11.2.6",
     "@nestjs/typeorm": "11.0.0",
     "@safe-global/safe-deployments": "1.37.51",
+    "@safe-global/safe-modules-deployments": "^3.0.1",
```
vseehausen: Pin the version to ensure strictness.

src/modules/relay/domain/contracts/__tests__/encoders/signer-factory-encoder.builder.ts
```diff
@@ -0,0 +1,39 @@
...
+  x: bigint;
+  y: bigint;
+  verifiers: bigint;
+};
+
+class CreateSignerEncoder<T extends CreateSignerArgs>
+  extends Builder<T>
+  implements IEncoder
+{
+  encode(): Address {
```
vseehausen: The return value is `Hex`. https://v1.viem.sh/docs/contract/encodeFunctionData.html#return-value

src/modules/relay/domain/limit-addresses.mapper.spec.ts
```diff
@@ -1403,6 +1414,161 @@ describe('LimitAddressesMapper', () => {
...
+              [x, y, verifiers],
+            ),
+          ).slice(-40)}`,
+        );
+
+      const factoryAddresses = getSignerFactoryDeployments({ chainId });
+
+      // The SafeWebAuthnSignerFactory is only deployed on a subset of
+      // supported chains; skip the suite where it isn't.
+      if (factoryAddresses.length === 0) return;
```
vseehausen: Let's ensure we use a supported chainId in the setup and fail loudly if the factory is not deployed.

vseehausen (changes requested):
Small changes request.

And a hint: de1c3175 introduces RelayFeeRelayer. We have to ensure it calls `LimitAddressesMapper`. Ideally add an integration test at this PR to ensure the whole pipeline works.

package.json
compojoom: that was already done

src/domain/common/utils/deployments.ts
```diff
@@ -291,3 +314,92 @@ export const isProxyFactoryDeployed = (
...
+/**
+ * Returns a list of official SafeWebAuthnSignerFactory addresses for a chain.
+ * Uses @safe-global/safe-modules-deployments (separate from core safe-deployments).
+ *
+ * Pinned to {@link SUPPORTED_SIGNER_FACTORY_VERSION}.
+ *
+ * @param {string} args.chainId - the chain ID to filter deployments by
+ * @returns {Array<Address>} - a list of checksummed factory addresses
+ */
+export function getSignerFactoryDeployments(args: {
```
vseehausen, line 339: nit: It returns only one address, right? Could be called `getSignerFactoryDeployment`.

src/modules/relay/domain/relay.manager.ts
```diff
@@ -35,17 +38,29 @@ export class RelayManager implements IRelayManager {
...
+   * @returns The relayer to use
    */
-  public getRelayer(chainId: string): IRelayer {
+  public getRelayer(chainId: string, data?: Address): IRelayer {
+    // Passkey signer deployment via SafeWebAuthnSignerFactory.createSigner is
+    // always sponsored — it must not be subject to noFeeCampaign balance-based
+    // rules or relay-fee charges, so route it straight to the daily-limit
+    // relayer regardless of chain config. The factory address is verified
+    // downstream in LimitAddressesMapper.
+    if (data && this.signerFactoryDecoder.helpers.isCreateSigner(data)) {
```
vseehausen, line 60: I wonder if this could have unintended side effects. I.e. in a NooFee campaign, we throttle add signer signatures. Very limited impact, but worth mentioning.

vseehausen (approved):
2 small comments, feel free to address.

PR #3066, opened 2026-04-29
feat(auth): store and expose verified OIDC email

vseehausen:

This PR implements the email identity layer on top of Auth0 ID token verification. It stores verified OIDC email claims on the CGW user and exposes the stored email on `/v1/auth/me`, so clients can render the authenticated email identity without adding email to the CGW JWT.

## Summary

- add nullable, unique `users.email` storage
- persist verified OIDC email claims during login without overwriting an existing user email
- reject duplicate email ownership during OIDC login and keep the user-facing error generic
- expose stored email on `/v1/auth/me` for OIDC sessions
- add focused unit/integration coverage for persistence, duplicate-email handling, and `/me`

## Notes

- Email is not added to the CGW JWT. `/me` reads the persisted email from the database.
- Unverified email claims are not persisted, but are still checked for duplicate ownership before minting a CGW session.


src/modules/auth/oidc/routes/oidc-auth.controller.ts
```diff
@@ -191,14 +193,27 @@ export class OidcAuthController {
...
+        res.redirect(
+          this.buildErrorRedirectUrl('authentication_failed', state),
+        );
+        return;
+      }
+
+      const error = asError(err);
+      this.loggingService.warn(
+        `Auth callback: authentication failed: ${error.message}`,
       );
```
LucieFaire: I dont think we need a separate catch block here, we can use the existing one. If the `UserEmailAlreadyInUseError` is a proper error with a message (which it is) - the default block will log it as well and the message will be there.

src/modules/auth/oidc/routes/oidc-auth.service.ts
```diff
@@ -75,6 +77,11 @@ export class OidcAuthService {
 
     const userId =
       await this.usersRepository.findOrCreateByExtUserId(extUserId);
+    if (emailVerified && email) {
+      await this.usersRepository.persistVerifiedEmail(userId, email);
+    } else if (email) {
+      await this.usersRepository.assertEmailCanBeUsedByUser(userId, email);
+    }
```
LucieFaire: i would suggest to create a new method in `userRepo` and hide all the logic in there, so that the service has to only deal with OIDC logic and doesnt need to know how user creation and email handling interact at the database level:

```
 const userId =
      await this.usersRepository.findOrCreateByExtUserIdWithEmail(
        extUserId,
        email ? { address: email, verified: emailVerified ?? false } : undefined,
      );
      
```

src/modules/users/domain/users.repository.interface.ts
```diff
@@ -58,6 +58,12 @@ export interface IUsersRepository {
 
   findOrCreateByExtUserId(extUserId: string): Promise<User['id']>;
 
+  assertEmailCanBeUsedByUser(userId: User['id'], email: string): Promise<void>;
+
+  persistVerifiedEmail(userId: User['id'], email: string): Promise<void>;
```
LucieFaire: i would prob make them private and not expose as they are not used by any other services or repo (esp, if we hide it from service layer)

src/modules/users/domain/users.repository.ts
```diff
@@ -251,6 +254,73 @@ export class UsersRepository implements IUsersRepository {
     }
   }
```
LucieFaire, line 284: so as in service we call a new method, we would add it here:
```
 public async findOrCreateByExtUserIdWithEmail(
    extUserId: string,
    email?: { address: string; verified: boolean },
  ): Promise<User['id']> {
    const userRepository =
      await this.postgresDatabaseService.getRepository(DbUser);

    const existingUser = await userRepository.findOne({
      where: { extUserId },
    });

    if (existingUser) {
      await this.handleUserEmail(existingUser.id, email);
      return existingUser.id;
    }

    let userId: User['id'];
    try {
      userId = await this.postgresDatabaseService.transaction((entityManager) =>
        this.create('ACTIVE', entityManager, { extUserId }),
      );
    } catch (error) {
      // Handle race condition: a concurrent call may have created the
      // user between our find and insert, causing a unique constraint
      // violation. Retry the lookup in that case.
      if (
        !(error instanceof Error) ||
        !error.message.includes('idx_users_ext_user_id')
      ) {
        throw error;
      }
      const user = await userRepository.findOneOrFail({
        where: { extUserId },
      });
      userId = user.id;
    }

    await this.handleUserEmail(userId, email);

    return userId;
  }
```


and the helper:
```
  private async handleUserEmail(
    userId: User['id'],
    email?: { address: string; verified: boolean },
  ): Promise<void> {
    if (!email) {
      return;
    }

    if (email.verified) {
      await this.persistVerifiedEmail(userId, email.address);
    } else {
      await this.assertEmailCanBeUsedByUser(userId, email.address);
    }
  }
```

and then update the persistVerifiedEmail (to not do another lookup

```
public async persistVerifiedEmail(
    userId: User['id'],
    email: string,
  ): Promise<void> {
    const userRepository =
      await this.postgresDatabaseService.getRepository(DbUser);
    const normalizedEmail = email.trim().toLowerCase();

    try {
      await userRepository
        .createQueryBuilder()
        .update(DbUser)
        .set({ email: normalizedEmail })
        .where('id = :userId', { userId })
        .andWhere('email IS NULL')
        .execute();
    } catch (error) {
      if (isUniqueConstraintError(error)) {
        throw new UserEmailAlreadyInUseError();
      }
      throw error;
    }
  }
```


In short what changed:
**Before**:

- SELECT user by id (to check if email is already set)
- If email set → return early
- UPDATE with WHERE email IS NULL guard
- If affected === 0 → throw InternalServerErrorException
- If unique constraint → throw UserEmailAlreadyInUseError

**After**:

- UPDATE with WHERE email IS NULL guard
- If unique constraint → throw UserEmailAlreadyInUseError

**What was removed:**

- The SELECT — redundant because the WHERE email IS NULL on the UPDATE already handles the "email already set" case as a no-op
- The InternalServerErrorException — dead code, only reachable if the userId didn't exist, which can't happen in practice
- The existingUser param — no longer needed since there's no SELECT to skip

Same behavior, fewer queries.

LucieFaire (commented):
initial review :)

src/modules/auth/oidc/routes/oidc-auth.controller.ts
vseehausen: 935b26c787744a01ed0dbeb4086e5b7d1e41eed3

src/modules/auth/oidc/routes/oidc-auth.service.ts
vseehausen: 935b26c787744a01ed0dbeb4086e5b7d1e41eed3

src/modules/users/domain/users.repository.interface.ts
vseehausen: 935b26c787744a01ed0dbeb4086e5b7d1e41eed3

src/modules/users/domain/users.repository.ts
vseehausen: 935b26c787744a01ed0dbeb4086e5b7d1e41eed3

vseehausen:
@claude review

src/modules/auth/oidc/routes/oidc-auth.controller.integration.spec.ts
```diff
@@ -567,6 +580,67 @@ describe('OidcAuthController', () => {
         expectErrorRedirect(response, 'authentication_failed');
       });
 
+      it('should hide duplicate verified email details from the user and log a warning', async () => {
```
LucieFaire: The test on line 556 already covers the same scenario (id_token with duplicate email) and just checks expectErrorRedirect without logging — this test is essentially a duplicate with an extra logging assertion bolted on. I would remove it and remove the spying. I would check logging in unit tests, and keep integration test as more of a overall test and user final expectations.

src/modules/users/domain/errors/user-email-already-in-use.error.ts
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { ConflictException, HttpStatus } from '@nestjs/common';
+
+export const USER_EMAIL_ALREADY_IN_USE_ERROR_CODE = 'user_email_already_in_use';
```
LucieFaire: nit: id shorten it to:
```
export const EMAIL_IN_USE_ERROR_CODE = 'email_in_use_error';

File: src/modules/users/domain/users.repository.spec.ts [hunk 7]
```diff
@@ -0,0 +1,211 @@
...
+  return {
+    update: jest.fn().mockReturnThis(),
+    set: jest.fn().mockReturnThis(),
+    where: jest.fn().mockReturnThis(),
+    andWhere: jest.fn().mockReturnThis(),
+    execute: jest.fn(),
+  };
+}
+
+describe('UsersRepository', () => {
```
LucieFaire: The createQueryBuilder mocking pattern is unusual for this repo — other spec files mock at the repository boundary (findOne, getContracts, etc.) rather than chaining internal query builder calls. This makes the test pretty coupled to the implementation.

Also, the transaction mock short-circuits the callback, so the query builder assertions on lines 93–100 aren't really exercising real behavior — they're testing mock wiring.

Since `users.repository.integration.spec.ts` already covers this against a real DB, is the query builder mocking here adding much?
I'd ask Claude to use the existing code patterns here :)

src/modules/users/domain/users.repository.ts
```diff
@@ -251,6 +253,120 @@ export class UsersRepository implements IUsersRepository {
...
+        .execute();
+    } catch (error) {
+      if (isUniqueConstraintError(error)) {
+        throw new UserEmailAlreadyInUseError();
+      }
+      throw error;
+    }
+  }
+
+  private async assertEmailCanBeUsedByUser(
```
LucieFaire: nit: `assertEmailNotTaken` ?

LucieFaire (commented):
all in all good, but left some small comments :)

src/modules/users/domain/users.repository.ts
```diff
@@ -251,6 +253,120 @@ export class UsersRepository implements IUsersRepository {
...
+      await this.postgresDatabaseService.getRepository(DbUser);
+
+    const existingUser = await userRepository.findOne({
+      where: { extUserId },
+    });
+
+    if (existingUser) {
+      await this.handleUserEmail(existingUser.id, email);
+      return existingUser.id;
+    }
```
LucieFaire, line 243: one thing: i would short circuit here is the email is already set.
```
if (existingUser) {
      if (!existingUser.email) {
        await this.handleUserEmail(existingUser.id, email);
      }
      return existingUser.id;
    }
```
that way we dont need to do another DB trip that wont perform any op in `handleUserEmail` as the email is not null anyway

src/modules/users/domain/users.repository.ts
```diff
@@ -251,6 +253,120 @@ export class UsersRepository implements IUsersRepository {
...
+
+    let userId: User['id'];
+    try {
+      userId = await this.postgresDatabaseService.transaction(
+        async (entityManager) => {
+          return await this.create('ACTIVE', entityManager, { extUserId });
+        },
+      );
+    } catch (error) {
```
LucieFaire, line 252: another qs:
Looking at the flow for a new user:
1. User is created and committed in its own transaction (line 277-296)
2. handleUserEmail is called after the transaction (line 298)
3. persistVerifiedEmail tries to set the email → unique constraint violation → throws UserEmailAlreadyInUseError

So the request fails, but the user has already been persisted without an email. We're left with an orphaned user record. Is that a good approach?

LucieFaire: actually this will be addressed/fixed as part [this](https://linear.app/safe-global/issue/WA-2031/connect-ext-user-id-to-already-existing-user) tix, so can leave it as is for now :)

vseehausen: That is true. I still find it better to add a check if the authentication is valid from our side (no other user has this email) before the login attempt. 

But it can be addressed in the linked ticket, yes.

src/modules/users/domain/users.repository.spec.ts
vseehausen: https://github.com/safe-global/safe-client-gateway/pull/3066/changes/a4bbbce42ad7ed88b60aeca176bb7229fae39d0d#diff-9682b5edce98c577f392b56965cf4ed91878de38cc976a797b97b2811e392ae1L13

src/modules/users/domain/users.repository.ts
vseehausen: nice one https://github.com/safe-global/safe-client-gateway/pull/3066/changes/a4bbbce42ad7ed88b60aeca176bb7229fae39d0d#diff-4ddce654ae6970dcfa07af14bc8e3bbb072a22b2bce65cdc96c1cd9118582108L271

vseehausen (commented):
thx

src/modules/users/domain/users.repository.ts
vseehausen: https://github.com/safe-global/safe-client-gateway/pull/3066/changes/a4bbbce42ad7ed88b60aeca176bb7229fae39d0d#diff-4ddce654ae6970dcfa07af14bc8e3bbb072a22b2bce65cdc96c1cd9118582108L314

src/modules/users/domain/errors/user-email-already-in-use.error.ts
vseehausen: Done.

vseehausen:
@claude review again.

src/modules/users/domain/users.repository.ts
```diff
@@ -7,16 +7,30 @@ import {
...
+import { isUniqueConstraintError } from '@/datasources/errors/helpers/is-unique-constraint-error.helper';
 import { User as DbUser } from '@/modules/users/datasources/entities/users.entity.db';
 import { Wallet } from '@/modules/wallets/datasources/entities/wallets.entity.db';
 import { EntityManager } from 'typeorm';
 import type { FindOptionsRelations, FindOptionsWhere } from 'typeorm';
 import { IWalletsRepository } from '@/modules/wallets/domain/wallets.repository.interface';
 import type { Address } from 'viem';
+import { UserEmailAlreadyInUseError } from '@/modules/users/domain/errors/user-email-already-in-use.error';
+
+function isUniqueConstraintViolation(
```
LucieFaire: why not make it a private method as it belongs to only 1 method?

vseehausen:
> **Claude finished @vseehausen's task** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25171434225)
> ### Code Review
> 
> The previous review feedback has been well-addressed. The code is clean and well-structured after the refactoring. A few remaining items:
> 
> **1. No application-layer email validation on the OIDC path** `oidc-auth.service.ts:80` passes the raw email from ID token claims directly to the repository. The `UserSchema` has `z.email().max(255)`, but this schema isn't applied on the OIDC ingest path. A malformed or oversized email string would only be caught by the DB varchar(255) constraint or the lowercase CHECK, yielding a raw DB error rather than a clean domain error.
> 
> **2. `assertEmailNotTaken` is bypassable by timing** `users.repository.ts:320-337`: For unverified emails, `assertEmailNotTaken` does a SELECT-then-decide check without holding any lock. Between the SELECT and the caller minting the CGW session, another concurrent request could persist the same email via `persistVerifiedEmail`. This is a TOCTOU gap — the unverified-email check says "email is available" but by the time the session is minted, another user may own it. Low severity since the unverified path doesn't write, but worth documenting as a known limitation.
> 
> **3. Missing integration test for `findEmailById`** Unit tests exist in `users.repository.spec.ts:127-146`, but there's no corresponding integration test in `users.repository.integration.spec.ts` that exercises `findEmailById` against a real database. Given other methods have integration tests, this is a coverage gap.
> 
> **4. `findOrCreateByExtUserIdWithEmail` — first call for existing user with `email: null` but no email param** At `users.repository.ts:250`, when `existingUser.email` is `null` and no `email` param is passed, `handleUserEmail` is called with `undefined`, which immediately returns. This is correct but results in unnecessary method call overhead on every OIDC login for users who never had email claims. Very minor.
> 
> Overall the PR looks solid. The reviewer feedback was cleanly incorporated. The main actionable items are (1) adding email format/length validation before it hits the DB and (3) adding the integration test for `findEmailById`.
> 
> [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25171434225) |

1: addressed
2: that's fine
3: integration tests cover the expected behavior enough
4: addressed

src/modules/users/domain/users.repository.integration.spec.ts
```diff
@@ -105,6 +106,7 @@ describe('UsersRepository', () => {
   });
 
   afterEach(async () => {
+    jest.restoreAllMocks();
```
LucieFaire: nit: 
`restoreAllMocks` is only needed if you `jest.spyOn()` something in a test, and I don't see that happening here. So it's redundant (yet harmless)

LucieFaire (approved):
2 tiny nits I'd like to be addressed, but otherwise ready to go

src/modules/users/domain/users.repository.ts
vseehausen: https://github.com/safe-global/safe-client-gateway/pull/3066/changes/3e5cbc67c516e4767a2e0e88fdd24a2a39c9a7df

src/modules/users/domain/users.repository.integration.spec.ts
vseehausen: https://github.com/safe-global/safe-client-gateway/pull/3066/changes/3e5cbc67c516e4767a2e0e88fdd24a2a39c9a7df

LucieFaire (approved):

PR #3067, opened 2026-05-04
feat(spaces): support email invites

vseehausen:
Resolves: https://linear.app/safe-global/issue/WA-2022/email-invite-creation-endpoint

This PR adds email-based space invite creation alongside the existing wallet-address invite flow, so admins can invite people before they have a wallet connected.

## Summary

- allow invite requests to contain either a wallet address or an email address
- create or reuse pending users for email invitees inside the invite transaction
- store invite expiration on member records and return it with invitation/member responses
- expose invitee email only to admins and the invited user
- keep pending invite emails hidden from regular member-list responses
- add focused service, repository, and e2e coverage for email invites and invite visibility

## Notes

- Existing wallet-address invites remain supported.
- This PR creates the invite record only; email delivery is handled separately.

### DIFF
--- a/src/modules/spaces/routes/members.controller.e2e-spec.ts
+++ b/src/modules/spaces/routes/members.controller.e2e-spec.ts
@@ -642,7 +642,7 @@ describe('MembersController', () => {
     });
   });
 
-  describe('POST /v1/spaces/:spaceId/members/:userId/resend', () => {
+  describe('POST /v1/spaces/:spaceId/members/resend', () => {
     it('should resend an invitation', async () => {
       const adminAuthPayloadDto = siweAuthPayloadDtoBuilder().build();
       const adminAccessToken = jwtService.sign(adminAuthPayloadDto);
@@ -684,8 +684,9 @@ describe('MembersController', () => {
         .expect(201);
 
       await request(app.getHttpServer())
-        .post(`/v1/spaces/${spaceId}/members/${userId}/resend`)
+        .post(`/v1/spaces/${spaceId}/members/resend`)
         .set('Cookie', [`access_token=${adminAccessToken}`])
+        .send({ address: inviteeAuthPayloadDto.signer_address })
         .expect(201)
         .expect({});
 
--- a/src/modules/spaces/routes/members.controller.e2e-spec.ts
+++ b/src/modules/spaces/routes/members.controller.e2e-spec.ts
@@ -127,6 +127,7 @@ describe('MembersController', () => {
               role: 'ADMIN',
               status: 'INVITED',
               invitedBy: authPayloadDto.signer_address,
+              inviteExpiresAt: expect.any(String),
             },
             {
               userId: expect.any(Number),
@@ -135,6 +136,54 @@ describe('MembersController', () => {
               role: 'MEMBER',
               status: 'INVITED',
               invitedBy: authPayloadDto.signer_address,
+              inviteExpiresAt: expect.any(String),
+            },
+          ]),
+        );
+    });
+
+    it('should invite users by email', async () => {
+      const authPayloadDto = siweAuthPayloadDtoBuilder().build();
+      const accessToken = jwtService.sign(authPayloadDto);
+      const spaceName = nameBuilder();
+      const invitedEmail = faker.internet.email().toLowerCase();
+      const invitedName = faker.person.firstName();
+
+      await request(app.getHttpServer())
+        .post('/v1/users/wallet')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .expect(201);
+
+      const createSpaceResponse = await request(app.getHttpServer())
+        .post('/v1/spaces')
+        .set('Cookie', [`access_token=${accessToken}`])
+        .send({ name: spaceName })
+        .expect(201);
+      const spaceId = createSpaceResponse.body.id;
+
+      await request(app.getHttpServer())
+        .post(`/v1/spaces/${spaceId}/members/invite`)
+        .set('Cookie', [`access_token=${accessToken}`])
+        .send({
+          users: [
+            {
+              role: 'MEMBER',
+              email: invitedEmail,
+              name: invitedName,
+            },
+          ],
+        })
+        .expect(201)
+        .expect(({ body }) =>
+          expect(body).toEqual([
+            {
+              userId: expect.any(Number),
+              spaceId,
+              name: invitedName,
+              role: 'MEMBER',
+              status: 'INVITED',
+              invitedBy: authPayloadDto.signer_address,
+              inviteExpiresAt: expect.any(String),
             },
           ]),
         );
@@ -587,9 +636,74 @@ describe('MembersController', () => {
         })
         .expect(409)
         .expect({
-          message: `${memberAddress} is already in this space or has a pending invite.`,
+          message: 'User is already in this space or has a pending invite.',
           error: 'Conflict',
           statusCode: 409,
+        });
+    });
+  });
+
+  describe('POST /v1/spaces/:spaceId/members/:userId/resend', () => {
+    it('should resend an invitation', async () => {
+      const adminAuthPayloadDto = siweAuthPayloadDtoBuilder().build();
+      const adminAccessToken = jwtService.sign(adminAuthPayloadDto);
+      const inviteeAuthPayloadDto = siweAuthPayloadDtoBuilder().build();
+      const inviteeAccessToken = jwtService.sign(inviteeAuthPayloadDto);
+      const spaceName = nameBuilder();
+      const memberName = nameBuilder();
+
+      await request(app.getHttpServer())
+        .post('/v1/users/wallet')
+        .set('Cookie', [`access_token=${adminAccessToken}`])
+        .expect(201);
+
+      const createSpaceResponse = await request(app.getHttpServer())
+        .post('/v1/spaces')
+        .set('Cookie', [`access_token=${adminAccessToken}`])
+        .send({ name: spaceName })
+        .expect(201);
+      const spaceId = createSpaceResponse.body.id;
+
+      const inviteUsersResponse = await request(app.getHttpServer())
+        .post(`/v1/spaces/${spaceId}/members/invite`)
+        .set('Cookie', [`access_token=${adminAccessToken}`])
+        .send({
+          users: [
+            {
+              role: 'MEMBER',
+              address: inviteeAuthPayloadDto.signer_address,
+              name: memberName,
+            },
+          ],
+        })
+        .expect(201);
+      const userId = inviteUsersResponse.body[0].userId;
+
+      await request(app.getHttpServer())
+        .post(`/v1/spaces/${spaceId}/members/decline`)
+        .set('Cookie', [`access_token=${inviteeAccessToken}`])
+        .expect(201);
+
+      await request(app.getHttpServer())
+        .post(`/v1/spaces/${spaceId}/members/${userId}/resend`)
+        .set('Cookie', [`access_token=${adminAccessToken}`])
+        .expect(201)
+        .expect({});
+
+      await request(app.getHttpServer())
+        .get(`/v1/spaces/${spaceId}/members`)
+        .set('Cookie', [`access_token=${adminAccessToken}`])
+        .expect(200)
+        .expect(({ body }) => {
+          expect(body.members).toEqual(
+            expect.arrayContaining([
+              expect.objectContaining({
+                status: 'INVITED',
+                inviteExpiresAt: expect.any(String),
+                user: expect.objectContaining({ id: userId }),
+              }),
+            ]),
+          );
         });
     });
   });
--- a/src/modules/spaces/routes/members.controller.e2e-spec.ts
+++ b/src/modules/spaces/routes/members.controller.e2e-spec.ts
@@ -177,7 +177,6 @@ describe('MembersController', () => {
         .expect(({ body }) =>
           expect(body).toEqual([
             {
-              userId: expect.any(Number),
               spaceId,
               name: invitedName,
               role: 'MEMBER',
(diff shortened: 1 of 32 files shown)


src/modules/spaces/routes/members.controller.e2e-spec.ts
```diff
@@ -587,13 +636,78 @@ describe('MembersController', () => {
         })
         .expect(409)
         .expect({
-          message: `${memberAddress} is already in this space or has a pending invite.`,
+          message: 'User is already in this space or has a pending invite.',
```
vseehausen, line 638: Removed this to exclude addresses and emails from logs.

src/modules/users/domain/users.repository.ts
```diff
@@ -267,6 +315,55 @@ export class UsersRepository implements IUsersRepository {
     return userId;
   }
 
+  private async linkPendingInviteeByVerifiedEmail(
+    extUserId: string,
+    email: string,
+    userRepository: Repository<DbUser>,
+  ): Promise<User['id'] | null> {
+    const normalizedEmail = email.trim().toLowerCase();
```
vseehausen, line 332: We trim the email twice here, but I like this better than renaming the `email` parameter to `normalizedEmail`.

LucieFaire:
@claude review

LucieFaire:
somehow i cannot comment on files, so here is my review:
1. **src/modules/spaces/routes/entities/invite-users.dto.entity.ts** -> InviteUserSchema
we can avoid using `refine` and have union schema:
```
const SharedInviteFields = {
  role: z.enum(getStringEnumKeys(MemberRole)),
  name: NameSchema,
};

const InviteUserSchema = z.union([
  z.object({ address: AddressSchema, ...SharedInviteFields }).strict(),
  z.object({ email: z.email().max(255), ...SharedInviteFields }).strict(),
]);
```
the only caveat you have to drop `implements z.infer<typeof InviteUsersDtoSchema>` from InviteUsersDto L59

2. **src/modules/spaces/routes/members.controller.ts** -> nit
I think the description lost some staff: `description: 'List of wallet addresses or email addresses to invite to the space'`

3. **src/modules/spaces/routes/members.service.spec.ts**
```
const INVITE_EXPIRY_SECONDS = 7 * 24 * 60 * 60;
const INVITE_CREATED_AT = new Date('2026-04-24T00:00:00.000Z');
const EXPECTED_INVITE_EXPIRES_AT = new Date(
  INVITE_CREATED_AT.getTime() + INVITE_EXPIRY_SECONDS * 1_000,
);
```
lets use faker and then just increment whatever values it provides

4. **src/modules/spaces/routes/members.service.spec.ts**
```
type MembersRepositoryMock = Pick<
  jest.Mocked<IMembersRepositoryInterface>,
  | 'inviteUsers'
  | 'resendInvite'
  | 'findAuthorizedMembersOrFail'
  | 'findSelfMembershipOrFail'
>;

const membersRepositoryMock: MembersRepositoryMock = {
  inviteUsers: jest.fn(),
  resendInvite: jest.fn(),
  findAuthorizedMembersOrFail: jest.fn(),
  findSelfMembershipOrFail: jest.fn(),
};
```
dont know why do we need the type here if we could just simply put:
```
const membersRepositoryMock = {
  inviteUsers: jest.fn(),
  resendInvite: jest.fn(),
  findAuthorizedMembersOrFail: jest.fn(),
  findSelfMembershipOrFail: jest.fn(),
} as jest.MockedObjectDeep<IMembersRepository>;
```
________________________
`describe('inviteUser', () =>` and `describe('resendInvite', () =>` test block we could use before/after instead of try/finally

5. **src/modules/spaces/routes/spaces.service.ts**
-> revert change on L71-78 as it can stay an implicit return

6. **src/modules/users/__tests__/test.users.module.ts**
nit: ` findOrCreateInviteeByEmail: jest.fn().mockResolvedValue(1),` can just be ` findOrCreateInviteeByEmail: jest.fn(),` because that mockReturn will be reset anyway

7. **src/modules/users/domain/members.repository.integration.spec.ts**
nit: lets use faker for `INVITE_EXPIRES_AT`


The rest will come when GH fixes the issue of commenting on files

vseehausen:
Addressed Claude's comments in [81aa055](https://github.com/safe-global/safe-client-gateway/pull/3067/commits/81aa0557174f082f75ea9cf967c0b3bbd87b9669)

And @LucieFaire's in [27bcf9d](https://github.com/safe-global/safe-client-gateway/pull/3067/commits/27bcf9dce001321c83c27e55346ab46019f16b5d)

vseehausen:
@claude review again

vseehausen:
1, 2, 4, 5: addressed in 2431f4312.

3: Ignored. The risk is minimal, as the endpoint is gated to admins of the space and looks for invites of this space. The user exist as a stub for invites. There is no risk of enumerating through registered emails.

@claude review again.

vseehausen:
@claude review again.

1. accepted.
2,3: addressed.

vseehausen:
@claude review again

vseehausen:
@claude review again

vseehausen:
@claude review again

vseehausen:
> **Claude finished @vseehausen's task** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25508545260)
> ### Code Review: feat(spaces): support email invites (seventh review)
> 
> All prior feedback across six rounds has been well addressed. The code is clean and in good shape. One remaining item:
> #### Code Quality
> 
>     1. **`getVisibleMembersForUser` doesn't redact invite emails** — `src/modules/spaces/routes/spaces.service.ts:79-94`: The `getActiveOrInvitedSpaces` method now correctly filters out invited members for non-admins via `getVisibleMembersForUser`. However, for admins, the invited members are returned with their full `user` object (including `email`) without applying the `toMemberDto` email-redaction logic used in `MembersService.get()`. Since `getActiveOrInvitedSpaces` selects `user: { id: true }` (line 64), the email is likely not loaded from the DB in practice. But the `select` doesn't explicitly exclude `email`, so whether it leaks depends on TypeORM's behavior with partial selects — if the `user` relation loads all columns despite the `select`, invite emails would be exposed in the `GET /spaces` response without the controlled visibility applied on the `GET /:spaceId/members` endpoint. Consider explicitly confirming that the `select` on `user` only returns `{ id }`, or applying email redaction for safety.
> 
> 
> LGTM overall. The implementation is solid across all layers — union schemas, discriminated union types, rate limiting, race condition handling, email visibility controls, and comprehensive test coverage are all well done. Ready to merge.
> 
> [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/25508545260)

That is by designs. Admins are allowed to view invitees' emails.

src/modules/users/domain/entities/invitation.entity.ts
```diff
@@ -1,12 +1,30 @@
...
+export type AddressInvitationIdentifier = InvitationIdentifier &
+  Required<Pick<InvitationIdentifier, 'address'>> & {
+    email?: never;
+  };
+
+export type EmailInvitationIdentifier = InvitationIdentifier &
+  Required<Pick<InvitationIdentifier, 'email'>> & {
+    address?: never;
+  };
+
```
LucieFaire: The `InvitationIdentifier` base + `Required<Pick<…>> ` indirection is doing nothing — the `& { email?: never }` clause already overrides the optional field.  We can simplify it:
```
export type InvitationIdentifier =
  | { address: Address; email?: never }
  | { email: string; address?: never };
```

src/modules/users/domain/members.repository.interface.ts
```diff
@@ -48,6 +54,14 @@ export interface IMembersRepository {
     payload: Pick<Member, 'name'>;
   }): Promise<void>;
 
+  resendInvite(
+    args: {
+      authPayload: AuthPayload;
+      spaceId: Space['id'];
+      inviteExpiresAt: Member['inviteExpiresAt'];
+    } & (AddressInvitationIdentifier | EmailInvitationIdentifier),
```
LucieFaire: ` & InvitationIdentifier,`

src/modules/users/domain/members.repository.ts
```diff
@@ -150,24 +157,31 @@ export class MembersRepository implements IMembersRepository {
             role: userToInvite.role,
             status: 'INVITED',
             invitedBy,
+            inviteExpiresAt: args.inviteExpiresAt,
           });
         } catch (err) {
           if (isUniqueConstraintError(err)) {
             throw new UniqueConstraintError(
-              `${userToInvite.address} is already in this space or has a pending invite.`,
+              'User is already in this space or has a pending invite.',
```
LucieFaire, line 165: wonder if we should add some identifier to know which user it is? Does wallet parse this error for the notification?

src/modules/users/domain/members.repository.ts
```diff
@@ -150,24 +157,31 @@ export class MembersRepository implements IMembersRepository {
...
           status: 'INVITED',
           invitedBy,
-        });
+          inviteExpiresAt: args.inviteExpiresAt,
+        };
+        if (userToInvite.email) {
+          // Avoid exposing internal user IDs for email-based invite probes.
+          invitation.userId = undefined;
+        }
+        invitations.push(invitation);
```
LucieFaire, line 180: can be inlined:
```
 invitations.push({
          // Avoid exposing internal user IDs for email-based invite probes.
          userId: userToInvite.email ? undefined : userIdToInvite,
          spaceId: space.id,
          name: userToInvite.name,
          role: userToInvite.role,
          status: 'INVITED',
          invitedBy,
          inviteExpiresAt: args.inviteExpiresAt,
        });
```

src/modules/users/domain/members.repository.ts
```diff
@@ -235,10 +292,52 @@ export class MembersRepository implements IMembersRepository {
     await this.postgresDatabaseService.transaction(async (entityManager) => {
       await entityManager.update(DbMember, member.id, {
         status: 'DECLINED',
+        inviteExpiresAt: null,
```
LucieFaire, line 291: Its inconsistent with the migration. The method explicitly sets `inviteExpiresAt: null` but the migration backfills DECLINED rows (status 2) with created_at + 7 days, violating that invariant for historical data.

The migration's WHERE "status" IN (0, 2) should be WHERE "status" = 0 — only INVITED rows need backfilling. (And as a bonus, that drops the dependency on the DECLINED enum value being 2.)

src/modules/users/domain/members.repository.ts
```diff
@@ -235,10 +292,52 @@ export class MembersRepository implements IMembersRepository {
...
       });
     });
   }
 
+  public async resendInvite(
+    args: Parameters<IMembersRepository['resendInvite']>[0],
```
LucieFaire, line 297: could we extend it to actual parameters as we dont really use this notation

src/modules/users/domain/members.repository.ts
```diff
@@ -235,10 +292,52 @@ export class MembersRepository implements IMembersRepository {
...
+  ): Promise<void> {
+    const actingUserId = getAuthenticatedUserIdOrFail(args.authPayload);
+    const activeAdmins = await this.findActiveAdminsOrFail(args.spaceId);
+
+    this.assertIsActiveAdmin({ members: activeAdmins, userId: actingUserId });
```
LucieFaire, line 302: Fetching every admin and scanning in memory is wasteful — a single targeted `findOne` for the caller would be equivalent and lighter, matching the inviteUsers pattern.

src/modules/users/domain/members.repository.ts
```diff
@@ -235,10 +292,52 @@ export class MembersRepository implements IMembersRepository {
...
+    const queryBuilder = membersRepository
+      .createQueryBuilder()
+      .update(DbMember)
+      .set({
+        status: 'INVITED',
+        inviteExpiresAt: args.inviteExpiresAt,
+      })
+      .where('space_id = :spaceId', { spaceId: args.spaceId })
+      .andWhere('status IN (:...statuses)', {
+        statuses: [MemberStatus.INVITED, MemberStatus.DECLINED],
```
LucieFaire, line 315: why would you want to resend an invitation if it was declined already?

src/modules/users/domain/members.repository.ts
```diff
@@ -235,10 +292,52 @@ export class MembersRepository implements IMembersRepository {
...
+      queryBuilder.andWhere(
+        'user_id IN (SELECT id FROM users WHERE email = :email)',
+        { email: args.email.trim().toLowerCase() },
+      );
+    } else {
+      queryBuilder.andWhere(
+        'user_id IN (SELECT user_id FROM wallets WHERE address = :address)',
+        { address: getAddress(args.address) },
+      );
+    }
```
LucieFaire, line 328: why dont we just fetch the existing member by its id and retrigger? we know that invitation is already present, so we dont need to look by diff criteria. I would propose to update the API to `POST /:spaceId/members/:memberId/resend` and use id to fetch and update, that will simplify this query

src/modules/spaces/routes/members.service.ts
```diff
@@ -69,31 +101,38 @@ export class MembersService {
...
+      authPayload: args.authPayload,
+      spaceId: args.spaceId,
+    });
+    const isAdmin = members.some((member) => {
+      return (
+        member.user.id === userId &&
+        member.status === 'ACTIVE' &&
+        member.role === 'ADMIN'
+      );
+    });
```
LucieFaire, line 115: `findAuthorizedMembersOrFail` already fetches the caller's row inside `findActiveOrInvitedMemberOrFail` for the authz check, then discards it. We then re-scan the bulk list here to recompute `isAdmin`, which is the exact info that row already had.

Could we maybe surface the caller's row from the repository and read isAdmin directly off it instead of `.some()`-ing the list? Or maybe you have other ideas?

LucieFaire (commented):
I could not review it all, because there is a lot going on. In the future i would suggest to split big tickets into subtasks and tackle each topic separately, that way we can review it much quicker and less places to miss something

vseehausen:
Closed in favor of smaller PRs.

PR #3078, opened 2026-05-05
feat(passkeys): WebAuthn coordinate storage API (WA-2113..WA-2119)

compojoom:
## Summary

Implements the [Passkey Coordinate Storage API](docs/plans/2026-05-05-feat-passkey-coordinate-storage-api-plan.md) — a CGW-hosted endpoint that maps a WebAuthn `credentialId` to its P-256 public-key coordinates `(x, y)` plus the `P256.Verifiers` packed `uint176` config, so a synced passkey on a second device can be used as a Safe signer.

Two endpoints, one Postgres table, no auth coupling. Off by default behind `FF_PASSKEYS`.

Resolves: WA-2113, WA-2114, WA-2115, WA-2116, WA-2117, WA-2118, WA-2119

## Changes

One commit per Linear ticket:

| Commit | Ticket | Scope |
|---|---|---|
| `feat(passkeys): add passkey_coordinates migration` | WA-2113 | Reversible migration; CHECK constraints on every fixed-width bytea; FILLFACTOR=100; destructive-rollback warning |
| `feat(passkeys): scaffold PasskeysModule + FF_PASSKEYS` | WA-2114 | Module skeleton, conditional registration, fail-closed bootstrap if any allowlist is empty |
| `feat(passkeys): POST /v1/passkeys + attestation verification` | WA-2115 | `@simplewebauthn/server@13.3.0` attestation verify, deterministic stateless challenge, 24 KiB route body cap, 500 ms verification timeout, idempotent first-write-wins via `ON CONFLICT DO NOTHING` |
| `feat(passkeys): GET /v1/passkeys/{credentialId}` | WA-2116 | Public lookup with `Cache-Control: public, max-age=86400, s-maxage=2592000, immutable` + `ETag`; `no-store` on miss |
| `feat(passkeys): per-IP rate limits` | WA-2117 | Two `RateLimitGuard` subclasses; 429 carries `Retry-After` and `Cache-Control: no-store` |
| `feat(passkeys): OpenAPI/Swagger error envelope` | WA-2118 | `PasskeyErrorResponse` DTO with full errorId enum; documented cache headers |
| `test(passkeys): HTTP-layer integration spec` | WA-2119 | 12 tests through the full Express pipeline with attestation + repository mocked |

### Notable cross-cutting changes
- **Global `CacheControlInterceptor`** now respects an existing `Cache-Control` header set by inner interceptors / `@Header` decorators, instead of unconditionally overwriting to `no-cache`. Without this, the immutable lookup policy was silently downgraded.
- **New dependency**: `@simplewebauthn/server@13.3.0`. Pinned at v13 (current latest); v11 was the API-stability boundary, v12/v13 are functionally compatible.

## Key invariants
- DTO has no `x` / `y` field — coordinates are NEVER trusted from the client; they come from the verified attestation.
- `expectedChallenge` is a deterministic constant-time-comparing function, never `() => true`.
- `verifiers` is `uint176` (22 bytes) — packed `P256.Verifiers`, NOT an Ethereum address. No EIP-55 checksum.
- Rows are immutable; PK conflict → 409 with distinct `PASSKEY_CONFLICT` / `PASSKEY_CROSS_RP_CONFLICT` codes.
- Module bootstrap fails closed if any of `rpId` / `origin` / `verifiers` allowlists is empty when `FF_PASSKEYS=true`.

## How to test it

```bash
# Unit + integration suite (38 tests, all pass)
yarn test src/modules/passkeys/

# TypeScript
yarn build  # or: tsc --noEmit

# Manual smoke (against a local CGW + Postgres):
export FF_PASSKEYS=true
export PASSKEYS_VERIFIERS_ALLOWLIST=0x000000000000000000000000000000000000000000  # 22 bytes hex
# POST a real WebAuthn attestation to /v1/passkeys
# GET /v1/passkeys/{credentialId} returns the canonical record
```

## Out of scope (per RFC author)
- `signer_address` / CREATE2 derivation — dropped from v1; an additive migration path is documented in the plan.
- Real attestation-fixture e2e — requires a live registration ceremony on iOS / Android / hardware key. The integration spec uses mocked attestation; recording fixtures is a follow-up.

## Checklist

- [ ] I've tested the branch on mobile
- [ ] I've documented how it affects the analytics (if at all)
- [x] I've written a unit/e2e test for it (38 tests added)

---
[Safe Engineering](https://github.com/5afe/safe-engineering-plugin) Generated with [Claude Code](https://claude.com/claude-code)


compojoom:
@claude review

compojoom:
@claude review

migrations/1778500000000-create-passkey-coordinates.ts
```diff
@@ -0,0 +1,41 @@
...
+      ) WITH (FILLFACTOR = 100);`,
+    );
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `DO $$
+       DECLARE row_count bigint;
+       BEGIN
+         SELECT count(*) INTO row_count FROM passkey_coordinates;
```
PooyaRaki: We should remove this query or at least change it to `count(credential_id)`

migrations/1778500000000-create-passkey-coordinates.ts
```diff
@@ -0,0 +1,41 @@
...
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE passkey_coordinates (
+        credential_id  bytea       NOT NULL,
+        x              bytea       NOT NULL,
+        y              bytea       NOT NULL,
+        verifiers      bytea       NOT NULL,
+        rp_id          text        NOT NULL,
+        created_at     timestamptz NOT NULL DEFAULT now(),
+        CONSTRAINT PK_passkey_coordinates PRIMARY KEY (credential_id),
```
PooyaRaki: Should we create an integer id as the PK and then make this one unique?

src/config/entities/__tests__/configuration.ts
```diff
@@ -425,6 +426,21 @@ export default (): ReturnType<typeof configuration> => ({
   safeWebApp: {
     baseUri: faker.internet.url({ appendSlash: false }),
   },
+  passkeys: {
+    rpIdAllowlist: ['app.safe.global'],
+    originAllowlist: ['https://app.safe.global'],
+    rateLimit: {
+      registration: {
+        max: faker.number.int({ min: 10, max: 100 }),
+        windowSeconds: 600,
```
PooyaRaki: `windowSeconds` s should also use `faker`

src/config/entities/configuration.ts
```diff
@@ -739,6 +740,46 @@ export default () => ({
...
   },
+  passkeys: {
+    rpIdAllowlist: (
+      process.env.PASSKEYS_RP_ID_ALLOWLIST ?? 'app.safe.global,safe.global'
+    )
+      .split(',')
+      .map((s) => s.trim())
+      .filter(Boolean),
+    originAllowlist: (
+      process.env.PASSKEYS_ORIGIN_ALLOWLIST ?? 'https://app.safe.global'
```
PooyaRaki, line 751: All these env vars should be validated:
`src/config/entities/schemas/configuration.schema.ts`

src/modules/passkeys/domain/entities/passkey-record.entity.ts
```diff
@@ -0,0 +1,27 @@
...
+}
+
+/**
+ * Input shape for write operations. `createdAt` is server-assigned by the DB
+ * (`DEFAULT now()`), so callers don't supply it.
+ */
+export type PasskeyRecordInput = Omit<PasskeyRecord, 'createdAt'>;
+
+export type WriteOutcome =
+  | { status: 'created'; record: PasskeyRecord }
```
PooyaRaki: We should read status from a const e.g.:
https://github.com/safe-global/safe-client-gateway/blob/022284aebe99fd896690709f0615a4fb542fbde9/src/modules/users/domain/entities/member.entity.ts

src/modules/passkeys/domain/passkey-attestation.service.ts
```diff
@@ -0,0 +1,262 @@
...
+  type ILoggingService,
+  LoggingService,
+} from '@/logging/logging.interface';
+
+/**
+ * Errors thrown by the attestation service. Mapped to HTTP status codes by the
+ * service layer; the controller never exposes the error message text — only an
+ * opaque `{ code: errorId }` envelope — so detail strings here are for logs.
+ */
+export class PasskeyAttestationError extends Error {
```
PooyaRaki: Error should be extracted to another file.

src/modules/passkeys/domain/passkey-attestation.service.ts
```diff
@@ -0,0 +1,262 @@
...
+  public constructor(
+    public readonly errorId: PasskeyAttestationErrorId,
+    message?: string,
+  ) {
+    super(message ?? errorId);
+    this.name = 'PasskeyAttestationError';
+  }
+}
+
+export type PasskeyAttestationErrorId =
```
PooyaRaki: Type should be extracted to another file.

src/modules/passkeys/domain/passkey-attestation.service.ts
```diff
@@ -0,0 +1,262 @@
...
+  | 'PASSKEY_RPID_NOT_ALLOWED'
+  | 'PASSKEY_ORIGIN_NOT_ALLOWED'
+  | 'PASSKEY_MALFORMED_ATTESTATION'
+  | 'PASSKEY_UNSUPPORTED_KEY'
+  | 'PASSKEY_RPID_MISMATCH'
+  | 'PASSKEY_ATTESTATION_INVALID'
+  | 'PASSKEY_VERIFICATION_TIMEOUT'
+  | 'PASSKEY_CHALLENGE_INVALID';
+
+export interface VerifyAttestationInput {
```
PooyaRaki: Interface should be extracted to another file.

src/modules/passkeys/domain/passkey-attestation.service.ts
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * Recomputes the expected challenge for this request. The verifier uses a
+   * constant-time comparison against the WebAuthn `challenge` from
+   * clientDataJSON. Returning `null` (or throwing) signals that no
+   * server-recomputable challenge can be derived for this input.
+   */
+  expectedChallenge: () => string;
+}
+
+export interface VerifiedPasskey {
```
PooyaRaki: Extract to another file.

src/modules/passkeys/domain/passkey-attestation.service.ts
```diff
@@ -0,0 +1,262 @@
...
+      });
+      throw err;
+    }
+    // Non-Error throws (strings, etc.) — treat as malformed; nothing useful
+    // to log and we do not want to surface arbitrary thrown values.
+    throw new PasskeyAttestationError('PASSKEY_MALFORMED_ATTESTATION');
+  }
+}
+
+function parseClientDataJSON(b64url: string): { type?: string } {
```
PooyaRaki: These helper methods should be extracted to another file.

src/modules/passkeys/domain/passkey-attestation.service.ts
```diff
@@ -0,0 +1,262 @@
...
+    configurationService: IConfigurationService,
+    @Inject(LoggingService)
+    private readonly loggingService: ILoggingService,
+  ) {
+    this.verificationTimeoutMs = configurationService.getOrThrow<number>(
+      'passkeys.verificationTimeoutMs',
+    );
+  }
+
+  public async verify(
```
PooyaRaki, line 54: This class doesn't have an interface so the methods should have TsDoc here. Interface is not necessary/recommended here but if we had one we could add the comments there.

PooyaRaki, line 54: This method at the time of writing this comment is 79 lines long, we should extract it to smaller methods.

src/modules/passkeys/domain/passkey-attestation.service.ts
```diff
@@ -0,0 +1,262 @@
...
+    throw new PasskeyAttestationError('PASSKEY_MALFORMED_ATTESTATION');
+  }
+}
+
+function parseClientDataJSON(b64url: string): { type?: string } {
+  try {
+    const raw = Buffer.from(b64url, 'base64url').toString('utf8');
+    const parsed: unknown = JSON.parse(raw);
+    if (typeof parsed === 'object' && parsed !== null) {
+      return parsed as { type?: string };
```
PooyaRaki: We should add a type guard instead of using a type cast

src/modules/passkeys/domain/passkeys.repository.ts
```diff
@@ -0,0 +1,108 @@
...
+        y: input.y,
+        verifiers: input.verifiers,
+        rpId: input.rpId,
+      })
+      .orIgnore()
+      .returning('*')
+      .execute();
+
+    if (insertResult.raw.length === 1) {
+      // `insertResult.raw` carries driver-level rows (snake_case columns), not
```
PooyaRaki: We should use `insertResult.generatedMaps` in order to prevent executing another select query.

src/modules/passkeys/domain/passkeys.repository.ts
```diff
@@ -0,0 +1,108 @@
...
+import type { IPasskeysRepository } from '@/modules/passkeys/domain/passkeys.repository.interface';
+
+@Injectable()
+export class PasskeysRepository implements IPasskeysRepository {
+  public constructor(
+    @Inject(PostgresDatabaseService)
+    private readonly postgresDatabaseService: PostgresDatabaseService,
+  ) {}
+
+  public async create(input: PasskeyRecordInput): Promise<WriteOutcome> {
```
PooyaRaki, line 21: This method is too long, can we split it to smaller methods?

src/modules/passkeys/domain/passkeys.repository.ts
```diff
@@ -0,0 +1,108 @@
...
+    const repo =
+      await this.postgresDatabaseService.getRepository(PasskeyCoordinates);
+    const row = await repo.findOneBy({ credentialId });
+    return row ? toRecord(row) : null;
+  }
+}
+
+function toRecord(row: PasskeyCoordinates): PasskeyRecord {
+  return {
+    credentialId: toBuffer(row.credentialId),
```
PooyaRaki: We could use TypeORM transformers so we wouldn’t need to manually transform them here. E.g:
https://github.com/safe-global/safe-client-gateway/blob/022284aebe99fd896690709f0615a4fb542fbde9/src/modules/counterfactual-safes/datasources/entities/counterfactual-safe.entity.db.ts#L38

src/modules/passkeys/routes/entities/register-passkey.dto.entity.ts
```diff
@@ -0,0 +1,81 @@
...
+const ATTESTATION_OBJECT_MAX = 16 * 1024;
+const CLIENT_DATA_JSON_MAX = 2 * 1024;
+
+export const RegisterPasskeySchema = z.object({
+  rpId: z.string().min(1).max(253).regex(RP_ID),
+  attestationObject: z
+    .string()
+    .min(1)
+    .max(ATTESTATION_OBJECT_MAX)
+    .regex(BASE64URL),
```
PooyaRaki: Can we replace this with `z.base64()`

src/modules/passkeys/routes/entities/register-passkey.dto.entity.ts
```diff
@@ -0,0 +1,81 @@
...
+  clientDataJSON: z.string().min(1).max(CLIENT_DATA_JSON_MAX).regex(BASE64URL),
+  // P256.Verifiers (uint176, 22 bytes). Lower 20 bytes = FCL fallback verifier
+  // address; upper 2 bytes = on-chain precompile address (RIP-7212). NOT an
+  // Ethereum address — no EIP-55 checksum applies. Lowercased on the wire.
+  verifiers: z.string().regex(VERIFIERS_HEX),
+  // The origin allowlist in the service is the load-bearing check, but URL
+  // parsing here strips control characters / whitespace / non-URL strings at
+  // the framework boundary as defense-in-depth.
+  origin: z.string().min(1).max(512).url(),
```
PooyaRaki: This is deprecated we should use `z.url()` instead.

src/modules/passkeys/routes/interceptors/passkeys-lookup-cache.interceptor.ts
```diff
@@ -0,0 +1,45 @@
...
+export class PasskeysLookupCacheInterceptor implements NestInterceptor {
+  public intercept(
+    context: ExecutionContext,
+    next: CallHandler,
+  ): Observable<unknown> {
+    const response = context.switchToHttp().getResponse<Response>();
+    return next.handle().pipe(
+      tap(() => {
+        if (!response.headersSent) {
+          response.setHeader('Cache-Control', HIT_CACHE_CONTROL);
```
PooyaRaki: QQ: Why do we need to override the global values here? I’m concerned this could create friction in the future if we want to change the cache duration.

src/modules/passkeys/routes/passkeys.controller.ts
```diff
@@ -0,0 +1,155 @@
...
+      'Attestation verification timed out (slow cert-chain validation).',
+    type: PasskeyErrorResponse,
+  })
+  @Post()
+  @UseGuards(PasskeysRegistrationRateLimitGuard)
+  // Default 201; the service may downgrade to 200 for idempotent re-POST.
+  @HttpCode(HttpStatus.CREATED)
+  public async register(
+    @Body(new ValidationPipe(RegisterPasskeySchema, HttpStatus.BAD_REQUEST))
+    dto: import('@/modules/passkeys/routes/entities/register-passkey.dto.entity').RegisterPasskeyDto,
```
PooyaRaki: Why do we need to import this inline? We could import the type at the top and have:

```
dto: RegisterPasskeyDto
```

src/modules/passkeys/routes/passkeys.service.ts
```diff
@@ -0,0 +1,271 @@
...
+ *     ceremony's clientData / attestation, so an attacker cannot freely vary
+ *     them without regenerating the attestation, and
+ *   - the comparison is constant-time (in PasskeyAttestationService).
+ */
+function deriveStatelessChallenge(dto: RegisterPasskeyDto): string {
+  const material = `passkey:v1\n${dto.rpId}\n${dto.origin}\n${dto.verifiers.toLowerCase()}`;
+  return createHash('sha256').update(material).digest('base64url');
+}
+
+function serialize(record: PasskeyRecord): PasskeyRecordResponse {
```
PooyaRaki: If we used TypeOrm transformers we could remove this function from here.

src/modules/passkeys/routes/passkeys.service.ts
```diff
@@ -0,0 +1,271 @@
...
+ * request inputs the client must already know to assemble a valid attestation.
+ * This is the Cometh-style model called out in the plan — it avoids a stateful
+ * challenge endpoint while keeping replay-resistance because:
+ *
+ *   - the bound inputs (rpId, verifiers, origin) are part of the WebAuthn
+ *     ceremony's clientData / attestation, so an attacker cannot freely vary
+ *     them without regenerating the attestation, and
+ *   - the comparison is constant-time (in PasskeyAttestationService).
+ */
+function deriveStatelessChallenge(dto: RegisterPasskeyDto): string {
```
PooyaRaki: Why is this a function rather than a class method?
Ideally, each file should contain a single type of object (e.g. class, interface, etc.).
Since this method is only used here, it could be implemented as a class method. If we intended to reuse it elsewhere, we should move it into a shared helper within the passkey module.

PooyaRaki (commented):
I reviewed the PR except for the tests, since I expect there will be quite a few changes based on the remarks. I’ll review the tests together with the next iteration.

migrations/1778500000000-create-passkey-coordinates.ts
compojoom: Resolved in a926edfb — removed the diagnostic `RAISE NOTICE` count query entirely. It was purely informational ahead of the `DROP TABLE`, and the destructive nature is already documented in the file header.

migrations/1778500000000-create-passkey-coordinates.ts
compojoom: Done in a926edfb — added a `SERIAL` surrogate `id` as the PK and a `UNIQUE` constraint on `credential_id`, matching the repo convention (`create_users`, `create_wallets`, `create-counterfactual-safes`). The entity was updated to `@PrimaryGeneratedColumn` + `@Unique`, and lookups still go through `credential_id`.

src/config/entities/__tests__/configuration.ts
compojoom: Done in a926edfb — both `windowSeconds` entries now use `faker.number.int({ min: 60, max: 3600 })`, consistent with the surrounding faker-generated fields.

src/config/entities/configuration.ts
compojoom: Done in a926edfb — all `PASSKEYS_*` env vars are now validated in `configuration.schema.ts` (allowlists as CSV piped to arrays, rate-limit maxima and timeout as coerced positive ints). The lookup-cache TTL vars added in this round are validated too.

src/modules/passkeys/domain/entities/passkey-record.entity.ts
compojoom: Done in a926edfb — extracted a `WriteOutcomeStatus` enum mirroring the `MemberStatus` pattern, and migrated all call sites (repository, service, specs).

src/modules/passkeys/domain/passkey-attestation.service.ts
compojoom: Done in a926edfb — `PasskeyAttestationError` moved to `domain/errors/passkey-attestation.error.ts`.

src/modules/passkeys/domain/passkey-attestation.service.ts
compojoom: Done in a926edfb — the error-id union moved to `domain/entities/passkey-attestation-error-id.entity.ts`.

src/modules/passkeys/domain/passkey-attestation.service.ts
compojoom: Done in a926edfb — `VerifyAttestationInput` moved to `domain/entities/verify-attestation-input.entity.ts`.

src/modules/passkeys/domain/passkey-attestation.service.ts
compojoom: Done in a926edfb — `VerifiedPasskey` moved to `domain/entities/verified-passkey.entity.ts`.

src/modules/passkeys/domain/passkey-attestation.service.ts
compojoom: Done in a926edfb — the helpers were moved to `domain/passkey-attestation.helpers.ts` (`parseClientDataJSON`, `timingSafeEqualBase64Url`, `pad32`, plus the new `isClientDataJSON` type guard).

src/modules/passkeys/domain/passkey-attestation.service.ts
compojoom: Done in a926edfb — added TsDoc to every public method of the class.

compojoom: Done in a926edfb — `verify` is split into four focused private methods: `assertAllowlisted`, `assertCreateCeremony`, `runLibraryVerification`, `extractPasskey`, each under 50 lines.

src/modules/passkeys/domain/passkey-attestation.service.ts
compojoom: Done in a926edfb — replaced the cast with an `isClientDataJSON(value): value is ClientDataJSON` type guard that rejects arrays and non-string `type`; failures fall through to `PASSKEY_MALFORMED_ATTESTATION`.

src/modules/passkeys/domain/passkeys.repository.ts
compojoom: Done in a926edfb — the post-INSERT `findOneByOrFail` SELECT is gone; the inserted row is read from `insertResult.generatedMaps[0]` (entity-shaped via the column transformers).

src/modules/passkeys/domain/passkeys.repository.ts
compojoom: Done in a926edfb — `create()` is now an orchestrator delegating to `tryInsert()` and `resolveConflict()`, with `coordinatesMatch()` and `toRecord()` helpers.

src/modules/passkeys/domain/passkeys.repository.ts
compojoom: Done in a926edfb — added a shared `databaseBufferTransformer` (mirroring `databaseAddressTransformer`) and applied it to every `bytea` column, so reads land as `Buffer` and the manual conversions are gone.

src/modules/passkeys/routes/entities/register-passkey.dto.entity.ts
compojoom: Applied in a926edfb, but as `z.base64url()` rather than `z.base64()`. These WebAuthn fields (`attestationObject`, `clientDataJSON`, `challenge`) are base64url-encoded (URL-safe alphabet, no padding), so plain `z.base64()` would reject valid payloads at runtime. `z.base64url()` is the zod v4 built-in matching the actual encoding. Happy to switch if the encoding assumption should change.

src/modules/passkeys/routes/entities/register-passkey.dto.entity.ts
compojoom: Done in a926edfb — `origin` now uses `z.url().min(1).max(512)` instead of the deprecated `z.string().url()`.

src/modules/passkeys/routes/interceptors/passkeys-lookup-cache.interceptor.ts
compojoom: The override is intentional and not avoidable: the global `CacheControlInterceptor` sets `no-cache` for every endpoint (sensible default), but passkey rows are immutable (`credentialId` is the PK and the row is never mutated), so the lookup can be cached aggressively. The global interceptor explicitly defers to inner interceptors via a `getHeader(Cache-Control)` guard, so this is the route-specific policy the framework expects rather than a fight against a configurable global. On the friction concern (fair point): in a926edfb the TTLs are moved out of constants into `passkeys.lookupCache.hitMaxAgeSeconds` / `hitSharedMaxAgeSeconds`, tunable via `PASSKEYS_LOOKUP_CACHE_HIT_MAX_AGE_SECONDS` and `PASSKEYS_LOOKUP_CACHE_HIT_S_MAX_AGE_SECONDS`. Defaults preserve current behaviour, so future tuning is a config change. The `no-store` miss policy stays hardcoded since it is a correctness property (prevents stale-negative caching).

src/modules/passkeys/routes/passkeys.controller.ts
compojoom: Done in a926edfb — added a top-level `import type { RegisterPasskeyDto }` and the parameter now reads `dto: RegisterPasskeyDto`. It is a pure type alias so there is no runtime import or circular-dependency risk.

src/modules/passkeys/routes/passkeys.service.ts
compojoom: Partially — the `databaseBufferTransformer` (a926edfb) removed the manual Buffer conversions, but a serialization step is still needed to shape the `PasskeyRecordResponse` returned to clients (hex/base64url string encoding for the API contract), which is separate from DB storage. So rather than delete it, it is now a `private serialize()` method on the class (see below).

src/modules/passkeys/routes/passkeys.service.ts
compojoom: Done in a926edfb — the standalone helpers in this file (`decodeCredentialId`, `deriveStatelessChallenge`, `serialize`, `stripHex`) are now `private` methods on `PasskeysService`, so the file holds a single class per the convention. Call sites updated to `this.…`.

PooyaRaki:
@compojoom I'll close this PR for now. Please feel free to reopen with an update.

PR #3080, opened 2026-05-07
fix: restore runtime imports for @ApiProperty class types

katspaugh:
## Summary

The Biome 2.0 migration (#3030) auto-converted runtime-needed class imports in entity files to `import type { ... }`. With `emitDecoratorMetadata: true` in `tsconfig.json`, type-only imports are erased before TS emits the `Reflect.metadata("design:type", ...)` call, so every `@ApiProperty()`-decorated field whose type was a type-only-imported class lost its constructor reference. NestJS's `@nestjs/swagger` then fell back to the global `Function` placeholder, corrupting the OpenAPI schema.

Downstream, `safe-wallet-monorepo`'s `@safe-global/store` `AUTO_GENERATED/*.ts` (generated from `https://safe-client.staging.5afe.dev/api-json` via `@rtk-query/codegen-openapi`) emits broken types like:

```ts
export type Function = {}
export type Chain = {
  // ...
  nativeCurrency: Function
  blockExplorerUriTemplate: Function
  balancesProvider: Function
  publicRpcUri: Function
  rpcUri: Function
  safeAppsRpcUri: Function
  theme: Function
  // ...
}
```

That breaks `type-check` in consumer packages (e.g. `@safe-global/test/factories/chains.ts` reading `nativeCurrency.name`).

Affected classes (49 broken `: Function` references in the generated output): `NativeCurrency`, `BlockExplorerUriTemplate`, `BeaconChainExplorerUriTemplate`, `BalancesProvider`, `RpcUri`, `Theme`, `AddressInfo`, `TokenInfo`, `DataDecoded`, `Transaction`, `TransactionData`, `ExecutionInfo`, `SafeAppAccessControl`, `ProtocolMetadata`.

## Changes

Switched the affected imports to the dual-import pattern already in use for `GasPriceOracle`/`GasPriceFixed`/`GasPriceFixedEIP1559` in `chain.entity.ts`:

```ts
// Before
import type {
  NativeCurrency as ApiNativeCurrency,
  NativeCurrency,
} from '@/modules/chains/routes/entities/native-currency.entity';

// After
import {
  NativeCurrency as ApiNativeCurrency,
  type NativeCurrency,
} from '@/modules/chains/routes/entities/native-currency.entity';
```

The class is now value-imported (so `emitDecoratorMetadata` can reference it at runtime) while the type alias remains type-only. `biome.json` already has `"useImportType": "off"`, so these fixes will not auto-revert.

`viem` string-branded aliases (`Address`, `Hash`, `Hex`) and Zod-inferred types are intentionally left as `import type` — they have no runtime representation and were never broken (`emitDecoratorMetadata` correctly emits `String` for them).

31 entity files updated under `src/modules/{chains,messages,positions,safe-apps,transactions}/`.

## Verification

- `yarn lint` — clean (1711 files, no fixes applied)
- `yarn format` — clean (1711 files, no fixes applied)
- `yarn test` — 4814 passed, 18 skipped, 0 failed (293 suites)

To verify the schema fix end-to-end after this PR is deployed to staging, regenerate the consumer types in `safe-wallet-monorepo`:

```bash
yarn workspace @safe-global/store build:dev
grep -rn ": Function" packages/store/src/gateway/AUTO_GENERATED/  # should return nothing
```

### DIFF
--- a/src/modules/chains/routes/entities/chain.entity.ts
+++ b/src/modules/chains/routes/entities/chain.entity.ts
@@ -6,14 +6,14 @@ import {
   getSchemaPath,
 } from '@nestjs/swagger';
 import type { Address } from 'viem';
-import type {
+import {
   BeaconChainExplorerUriTemplate as ApiBeaconChainExplorerUriTemplate,
-  BeaconChainExplorerUriTemplate,
+  type BeaconChainExplorerUriTemplate,
 } from '@/modules/chains/domain/entities/beacon-chain-explorer-uri-template.entity';
-import type { BalancesProvider } from '@/modules/chains/routes/entities/balances-provider.entity';
-import type {
+import { BalancesProvider } from '@/modules/chains/routes/entities/balances-provider.entity';
+import {
   BlockExplorerUriTemplate as ApiBlockExplorerUriTemplate,
-  BlockExplorerUriTemplate,
+  type BlockExplorerUriTemplate,
 } from '@/modules/chains/routes/entities/block-explorer-uri-template.entity';
 import {
   GasPriceFixed as ApiGasPriceFixed,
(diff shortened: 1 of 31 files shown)


vmosharova (approved):
👍🏻

LucieFaire (approved):
thank you!

PR #3084, opened 2026-05-08
feat(csv-export): add gas fee columns to transaction history CSV export

falvaradorodriguez:
- Closes https://linear.app/safe-global/issue/PLA-1419/history-csv-export-should-contain-paid-from-the-safe-info#comment-9b93eec1

- Related with https://github.com/safe-global/safe-transaction-service/pull/2878/changes

  ## Summary                                                                                                                                                                                                                                       
                  
  - Adds `gasToken`, `payment`, `gasTokenSymbol` and `gasTokenDecimals` fields                                                                                                                                                                         
    to `TransactionExportSchema`, parsing the new fields returned by the Tx Service
    `/api/v1/safes/{address}/export/` endpoint (PLA-1419)
      - New fields are parsed as .nullish() in the schema so older API versions that don't return them don't break existing exports
      - Adds a backwards-compatibility test that verifies parsing succeeds when the new fields are absent from the API response                                                                                                                                   
  - Formats the raw `payment` amount using `formatUnits(payment, gasTokenDecimals)`,                                                                                                                                                                       
    matching the existing pattern for `amount`/`assetDecimals`                                                                                                                                                                                     
  - Adds two new columns to the CSV output: **"Amount Gas"** and **"Gas token"**;                                                                                                                                                                  
    both are empty when the fee was paid from the signer wallet (`fee` is null)   
    
    
   `payment` is the amount the Safe paid out to refund the executor for gas costs. When a Safe transaction is executed, the executor can optionally be reimbursed directly from the Safe using a configured gas token. If the executor paid gas from their own wallet, this field is null.

  This is the "paid from the Safe" info referenced in the ticket, the new columns expose that refund amount (payment), the token used (gasToken, gasTokenSymbol, gasTokenDecimals) so it shows up in the CSV export for accounting purposes.

### DIFF
--- a/src/modules/csv-export/v1/csv-export.module.ts
+++ b/src/modules/csv-export/v1/csv-export.module.ts
@@ -12,6 +12,7 @@ import {
   type ILoggingService,
   LoggingService,
 } from '@/logging/logging.interface';
+import { ChainsModule } from '@/modules/chains/chains.module';
 import { CsvExportConsumer } from '@/modules/csv-export/v1/consumers/csv-export.consumer';
 import { CsvExportController } from '@/modules/csv-export/v1/csv-export.controller';
 import { CsvExportService } from '@/modules/csv-export/v1/csv-export.service';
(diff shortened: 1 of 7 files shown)


yagopv (approved):

clovisdasilvaneto (approved):
LGTM

PR #3096, opened 2026-05-15
feat(auth): introduce EmailAddress schema

vseehausen:
## Summary

Introduces a shared `EmailAddressSchema` and branded `EmailAddress` type for email validation across the codebase. Pre-work for [WA-2327 (email invites)](https://linear.app/safe-global/issue/WA-2327/add-email-invite-endpoint).

- Adds `EmailAddressSchema` under `validation/entities/schemas/`, matching the `UuidSchema` pattern: `z.email().toLowerCase().max(255).brand<'EmailAddress'>()`. `z.email()` already rejects strings with surrounding whitespace, so no explicit `.trim()` is needed.
- Replaces inline `z.email().max(255)` declarations across `User`, DB `User`, `Auth0Token`, and `UserSession` with the shared schema. The three env-var emails in `configuration.schema.ts` keep their plain `z.email()` shape — they're identifiers, not user emails, and out of scope here.
- `UsersRepository.persistVerifiedEmail` / `assertEmailNotTaken` drop their hand-rolled `.toLowerCase()` — the schema produces an already-normalized `EmailAddress` at the boundary.
- Adds `fakeEmailAddress()` test helper under `validation/entities/schemas/__tests__/` so fixtures across auth, users, and spaces specs can produce `EmailAddress`-typed values cleanly.
- Spec asserts the brand at compile time via `@ts-expect-error`: removing `.brand<'EmailAddress'>()` from the schema breaks the build because plain `string` would then be assignable to `EmailAddress` and the directive would be flagged as unused.

Three commits: production change, test migration, brand + revert of config-var changes per review feedback.

No HTTP-visible behavior change. Existing DB lowercase check constraint stays as defense-in-depth.

### DIFF
--- a/src/config/entities/schemas/configuration.schema.ts
+++ b/src/config/entities/schemas/configuration.schema.ts
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { z } from 'zod';
+import { EmailAddressSchema } from '@/validation/entities/schemas/email-address.schema';
 
 const relayRulesValidator = z
   .string()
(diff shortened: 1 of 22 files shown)


vseehausen:
@claude review

vseehausen:
2 is fine, the rest is addressed.

@claude review again

vseehausen:
@claude review again

vseehausen:
@claude review again

vseehausen:
Thanks, as we always checked the email before writing it to the DB, we can trust the values in the DB.

LucieFaire (approved):

PR #3097, opened 2026-05-18
Fix: isCancellation returns true when safeTx is empty with non-zero refund params

akshay-ap:
Closes: https://linear.app/safe-global/issue/PLA-1456/on-chain-rejection-tx-in-not-recognized-as-rejection-tx-when-safe-from

**Key changes:**

### Cancellation Transaction Detection Improvements

- Updated the `isCancellation` logic in `custom-transaction.mapper.ts` to correctly identify cancellation transactions not only when all refund parameters are zero or null, but also when all are non-zero and set, covering both native and ERC-20 gas token cases.

### Unit Test Coverage Expansion

- Added comprehensive unit tests in `custom-transaction.mapper.spec.ts` to verify the new cancellation classification logic, including:
  - Empty refund parameters (all zero/null)
  - Non-zero refund parameters with native gas token
  - Non-zero refund parameters with ERC-20 gas token
  - Cases where only some refund parameters are non-zero, ensuring these are not misclassified as cancellations

### DIFF
--- a/src/modules/transactions/routes/mappers/common/custom-transaction.mapper.ts
+++ b/src/modules/transactions/routes/mappers/common/custom-transaction.mapper.ts
@@ -95,23 +95,27 @@ export class CustomTransactionMapper {
         (!value || Number(value) === 0) &&
         operation === Operation.CALL;
 
+      const areAllRefundParamsEmptyOrDefault =
+        (!baseGas || Number(baseGas) === 0) &&
+        (!gasPrice || Number(gasPrice) === 0) &&
+        (!gasToken || gasToken === NULL_ADDRESS) &&
+        (!refundReceiver || refundReceiver === NULL_ADDRESS) &&
+        (!safeTxGas || safeTxGas === 0);
+
+      const areAllRefundParamsSet =
+        baseGas !== null &&
+        Number(baseGas) !== 0 &&
+        gasPrice !== null &&
+        Number(gasPrice) !== 0 &&
+        refundReceiver !== null &&
+        gasToken !== null &&
+        refundReceiver !== NULL_ADDRESS &&
+        safeTxGas !== null &&
+        safeTxGas !== 0;
+
       return (
-        (isEmptySelfCall &&
-          (!baseGas || Number(baseGas) === 0) &&
-          (!gasPrice || Number(gasPrice) === 0) &&
-          (!gasToken || gasToken === NULL_ADDRESS) &&
-          (!refundReceiver || refundReceiver === NULL_ADDRESS) &&
-          (!safeTxGas || safeTxGas === 0)) ||
-        (isEmptySelfCall &&
-          baseGas !== null &&
-          Number(baseGas) !== 0 &&
-          gasPrice !== null &&
-          Number(gasPrice) !== 0 &&
-          refundReceiver !== null &&
-          gasToken !== null &&
-          refundReceiver !== NULL_ADDRESS &&
-          safeTxGas !== null &&
-          safeTxGas !== 0)
+        isEmptySelfCall &&
+        (areAllRefundParamsEmptyOrDefault || areAllRefundParamsSet)
       );
     }
     return false;
--- a/src/modules/transactions/routes/mappers/common/custom-transaction.mapper.ts
+++ b/src/modules/transactions/routes/mappers/common/custom-transaction.mapper.ts
@@ -89,16 +89,28 @@ export class CustomTransactionMapper {
         safeTxGas,
       } = transaction;
 
-      return (
+      const isEmptyTx =
         to === safe &&
         dataSize === 0 &&
         (!value || Number(value) === 0) &&
-        operation === Operation.CALL &&
-        (!baseGas || Number(baseGas) === 0) &&
-        (!gasPrice || Number(gasPrice) === 0) &&
-        (!gasToken || gasToken === NULL_ADDRESS) &&
-        (!refundReceiver || refundReceiver === NULL_ADDRESS) &&
-        (!safeTxGas || safeTxGas === 0)
+        operation === Operation.CALL;
+
+      return (
+        (isEmptyTx &&
+          (!baseGas || Number(baseGas) === 0) &&
+          (!gasPrice || Number(gasPrice) === 0) &&
+          (!gasToken || gasToken === NULL_ADDRESS) &&
+          (!refundReceiver || refundReceiver === NULL_ADDRESS) &&
+          (!safeTxGas || safeTxGas === 0)) ||
+        (isEmptyTx &&
+          baseGas !== null &&
+          Number(baseGas) !== 0 &&
+          gasPrice !== null &&
+          Number(gasPrice) !== 0 &&
+          refundReceiver !== null &&
+          refundReceiver !== NULL_ADDRESS &&
+          safeTxGas !== null &&
+          safeTxGas !== 0)
       );
     }
     return false;
--- a/src/modules/transactions/routes/mappers/common/custom-transaction.mapper.ts
+++ b/src/modules/transactions/routes/mappers/common/custom-transaction.mapper.ts
@@ -108,6 +108,7 @@ export class CustomTransactionMapper {
           gasPrice !== null &&
           Number(gasPrice) !== 0 &&
           refundReceiver !== null &&
+          gasToken !== null &&
           refundReceiver !== NULL_ADDRESS &&
           safeTxGas !== null &&
           safeTxGas !== 0)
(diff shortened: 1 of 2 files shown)


akshay-ap:
@claude review

akshay-ap:
@claude review

akshay-ap:
> **Claude finished @akshay-ap's task in 1m 41s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/26099782443)
> 
> ### Code Review
> Fix is on the right track and the test cases meaningfully expand coverage. A few things worth addressing:
> 
> #### 1. `isCancellation` branches can be simplified and made consistent — `custom-transaction.mapper.ts:98-114`
> The repeated `isEmptySelfCall &&` plus duplicated parameter checks are hard to read. The two branches also use **different null-handling idioms** for the same fields:
> 
> * First branch: `!baseGas || Number(baseGas) === 0` (falsy-safe — handles `undefined`, `null`, `0`, `''`)
> * Second branch: `baseGas !== null && Number(baseGas) !== 0` (treats `undefined` as non-zero, since `Number(undefined) === NaN !== 0`)
> 
> If the API ever returns `undefined` (rather than `null`) for these fields, the second branch would misclassify the tx. Suggest factoring out helpers and combining:
> 
> ```ts
> const isZeroOrEmpty = (v: number | string | null | undefined): boolean =>
>   v == null || Number(v) === 0;
> 
> const allRefundParamsEmpty =
>   isZeroOrEmpty(baseGas) &&
>   isZeroOrEmpty(gasPrice) &&
>   (!gasToken || gasToken === NULL_ADDRESS) &&
>   (!refundReceiver || refundReceiver === NULL_ADDRESS) &&
>   isZeroOrEmpty(safeTxGas);
> 
> const allRefundParamsSet =
>   !isZeroOrEmpty(baseGas) &&
>   !isZeroOrEmpty(gasPrice) &&
>   !!refundReceiver && refundReceiver !== NULL_ADDRESS &&
>   !isZeroOrEmpty(safeTxGas);
> 
> return isEmptySelfCall && (allRefundParamsEmpty || allRefundParamsSet);
> ```
> 
> [Fix this →](https://claude.ai/code?q=In%20src%2Fmodules%2Ftransactions%2Froutes%2Fmappers%2Fcommon%2Fcustom-transaction.mapper.ts%20lines%2098-114%20on%20branch%20fix%2Frejection-tx-pay-with-safe%2C%20refactor%20isCancellation%20to%20extract%20a%20helper%20like%20isZeroOrEmpty%20%28handling%20null%2Fundefined%2F0%29%2C%20factor%20allRefundParamsEmpty%20and%20allRefundParamsSet%20variables%2C%20and%20return%20isEmptySelfCall%20%26%26%20%28allRefundParamsEmpty%20%7C%7C%20allRefundParamsSet%29.%20Make%20null-handling%20idioms%20consistent%20across%20both%20branches.&repo=safe-global/safe-client-gateway)
> 
> #### 2. `gasToken` intentionally omitted from second branch — add a comment
> For the "all refund params set" case `gasToken` can be either `NULL_ADDRESS` (native) or an ERC‑20 address, so it's deliberately not validated. This is non-obvious from the code. A one-line comment would prevent a future reader from "fixing" it.
> 
> #### 3. Test builder state leaks across cases — `custom-transaction.mapper.spec.ts` (the "only some required refund params are non-zero" test)
> `Builder.with()` **mutates** and returns `this` (see `src/__tests__/builder.ts:16`). The test reuses `baseTransaction` across five `mapCustomTransaction` calls, so each `.with(...)` call permanently changes state for subsequent assertions. For example, after the second case sets `baseGas: 0`, the third case has to explicitly reset it to `75608`. The fourth case ("refundReceiver zero") then inherits `gasPrice: '0'` from the third case, so it isn't actually isolating "only refundReceiver zero" — multiple params are zero.
> 
> The assertions all happen to pass (because `isCancellation` is `false` whenever any required param is zero), but the test names mislead. Either build a fresh transaction per case or document the cumulative state explicitly.
> 
> [Fix this →](https://claude.ai/code?q=In%20src%2Fmodules%2Ftransactions%2Froutes%2Fmappers%2Fcommon%2Fcustom-transaction.mapper.spec.ts%20on%20branch%20fix%2Frejection-tx-pay-with-safe%2C%20the%20test%20%27should%20not%20classify%20as%20cancellation%20when%20only%20some%20required%20refund%20params%20are%20non-zero%27%20reuses%20a%20mutable%20builder%20across%20multiple%20cases%20causing%20state%20leakage.%20Refactor%20each%20case%20to%20build%20a%20fresh%20multisigTransactionBuilder%20with%20only%20the%20intended%20field%20zeroed.&repo=safe-global/safe-client-gateway)
> 
> #### 4. Minor — `gasToken` typing
> In the second branch you don't check `gasToken`, but in the empty-params branch you check `!gasToken || gasToken === NULL_ADDRESS`. Worth a quick verification that downstream code (refund-info mapping, etc.) handles a cancellation with a non-null ERC‑20 gasToken correctly — i.e., a UI labelled "Cancellation" but with an ERC‑20 refund displayed shouldn't be surprising.
> 
> Nothing security- or performance-critical. Once items 1–3 are addressed this is good to go. • Branch: `fix/rejection-tx-pay-with-safe`

1. Invalid finding: optional params cannot be `undefined`.
2. Added explicit `gasToken !== null` check
3. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3097/commits/62dee885f3829de5b5774c9a4f34875a1fab280a
4. Invalid

src/modules/transactions/routes/mappers/common/custom-transaction.mapper.ts
```diff
@@ -89,16 +89,29 @@ export class CustomTransactionMapper {
...
-        operation === Operation.CALL &&
-        (!baseGas || Number(baseGas) === 0) &&
-        (!gasPrice || Number(gasPrice) === 0) &&
-        (!gasToken || gasToken === NULL_ADDRESS) &&
-        (!refundReceiver || refundReceiver === NULL_ADDRESS) &&
-        (!safeTxGas || safeTxGas === 0)
+        operation === Operation.CALL;
+
+      return (
+        (isEmptySelfCall &&
```
PooyaRaki: Why do we need to repeat `isEmptySelfCall` twice?
We should extract the conditions into different consts e.g.:

```ts
const isEmptyGasConfig = ...
const isValidGasConfig = ...

return isEmptySelfCall && (isEmptyGasConfig || isValidGasConfig)
```

akshay-ap: we can do that too! 

I kept the logic such that if first condition i.e. `isEmptySelfCall  && isEmptyGasConfig == true`, we skip evaluating the second. Here, as per your suggestion, `isValidGasConfig` is always evaluated.

btw, `isEmptyGasConfig` is also a valid gas config. I prefer the current implementation. We can replace with your suggestion if you have preference.

PooyaRaki: No — it won't be evaluated.
JavaScript's `||` operator short-circuits: as soon as it finds a `truthy` operand, it stops and returns that value without evaluating the rest. So in (`isEmptyGasConfig` || `isValidGasConfig`), if `isEmptyGasConfig` is true, the expression immediately resolves to `true` and `isValidGasConfig` is never evaluated.
The same applies to `&&` in the outer expression — if `isEmptySelfCall` were `false`, the entire parenthesized right side would be skipped altogether.
So with `isEmptySelfCall = true` and `isEmptyGasConfig = true`, only those two are evaluated; `isValidGasConfig` is skipped. This matters if `isValidGasConfig` is a function call or getter with side effects — those side effects won't fire.

PooyaRaki: The naming I suggested is only a placeholder — feel free to change it as needed. The main issue is that, in its current form, the condition is very difficult to read.

akshay-ap: I meant if `isValidGasConfig` is a boolean const outside of boolean expression it is evaluated even though not-needed.

Anyways, I don't have any strong preference here and updated the expression here: https://github.com/safe-global/safe-client-gateway/pull/3097/commits/02b7bdd41fd77fda75b22075e2d9b24c93d3c244

PooyaRaki: Thanks, Yeah that's true we could use functions so that they are loaded lazily but IMO that would be overkill.

PooyaRaki (approved):

PR #3101, opened 2026-05-18
feat(auth): require verified email for OIDC sign-in

vseehausen:
## Summary

- **Require a verified email on OIDC sign-in.** Reject the sign-in early when the ID token's `email` claim is missing or `email_verified` is not `true`. New `UserEmailRequiredError` (401, code `email_required_error`). Auth lives off of either a wallet address (SIWE) or a verified email (OIDC) — there is no other identifier we trust at sign-in.
- **Simplify the repository contract** now that the upstream guard makes the invariant a hard rule:
  - `findOrCreateByExtUserIdWithEmail(extUserId, email)` — `email` is non-optional; verification flag is gone (always verified).
  - The new user row is INSERTed with the email atomically; race recovery distinguishes between an `extUserId` collision (recover via lookup) and an email collision (`UserEmailAlreadyInUseError`).
  - Unverified-email handling (`handleUserEmail`'s else-branch, `assertEmailNotTaken`) is now dead code and removed; `persistVerifiedEmail` becomes `persistEmail`, used only for the "existing user has no email yet" top-up case, and now checks `affected` + re-reads on a race to detect a different-email collision.
- **New `UserEmailMismatchError`** (401, code `email_mismatch_error`). If an existing user (matched by `extUserId`) has a stored email that differs from the freshly authenticated address, we throw rather than silently keep the old value. Email changes at the IdP are rejected loudly.
- Closes the "double user entry" path: today an OIDC user signing in with an unverified email collides with any matching pending email-invite placeholder. Reconciling the two rows would require user-merge code (move memberships/invites, drop the placeholder, work around the unique-email constraint mid-merge, handle partial failures). Requiring verification upstream avoids needing that code at all.
- Symmetry with the invite-accept flow: we already trust verified email when **accepting** invites, so we should equally trust it when **sending / using** them. Both production OIDC providers we rely on (Google, magic-link email) only emit `email_verified: true`, so this is a code-base simplification rather than a UX regression.

## Related

Prerequisite for the WA-2327 email-invite endpoint — together they close both halves of the OIDC ↔ pending-invitee linking gap.


vseehausen:
@claude review

vseehausen:
> **Claude finished @vseehausen's task in 2m 18s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/26088119338)
> ### Claude PR Review
> 
> Nice simplification overall — collapses two-step email persistence into INSERT-time + targeted UPDATE, and the guard fires cleanly before any DB work. A few things worth considering:
> #### 1. Behavior mismatch with PR description: no-email sign-ins are now rejected
> 
> The PR description states _"No-email sign-ins are unaffected"_, but the new guard at `src/modules/auth/oidc/routes/oidc-auth.service.ts:79` is `if (!(email && emailVerified))`, which throws `UserEmailNotVerifiedError` whenever `email` is absent. The spec test at `oidc-auth.service.spec.ts:240-261` ("when the email claim is missing") confirms this. Previously the repository accepted `email?` as optional (`users.repository.interface.ts` pre-PR). This is an intentional tightening, but:
> 
>     * The error message _"Email must be verified to sign in"_ is misleading for the no-email-at-all case — consider either a different error code/message for the missing-claim case, or updating the PR description so reviewers don't expect the old behavior.
> 
> 
> #### 2. Guard condition is slightly redundant
> 
> `Auth0TokenSchema` already guarantees via `superRefine` (`auth0-token.entity.ts:12-20`) that `email` is present whenever `email_verified === true`. So `if (!(email && emailVerified))` can be tightened to `if (!emailVerified)` without behavior change, or kept defensive — your call, but worth noting the redundancy.
> #### 3. `persistEmail` silently no-ops on race
> 
> At `users.repository.ts:288-307`, the `UPDATE ... WHERE email IS NULL` correctly avoids overwriting but doesn't inspect `result.affected`. If a concurrent write set a _different_ email between the find and update (theoretically possible across logins for the same `extUserId`), the call returns silently without surfacing a mismatch on this request. Low-risk (next sign-in would catch via `UserEmailMismatchError`), but consider asserting `affected === 1` or re-reading post-update if you want strict guarantees.
> #### 4. Unit test gap for `findOrCreateByExtUserIdWithEmail`
> 
> `users.repository.spec.ts` (after this PR) only exercises the existing-user matching and mismatching branches. The persist-email-on-existing-user-with-null-email path and the create-new-user path are only covered in the integration spec. The integration coverage is solid, so this is just an observation — adding a mocked test for the `persistEmail` branch would shorten feedback for future regressions.
> #### 5. Minor: `UserEmailMismatchError` not mentioned in PR summary
> 
> The PR description only mentions `UserEmailNotVerifiedError`, but this PR also introduces `UserEmailMismatchError` (used when `extUserId` resolves to a user whose stored email differs from the incoming OIDC email). Worth listing for completeness — it's a user-visible API contract change with its own error code.
> 
> No blocking issues — the refactor reads well, integration coverage is good, and the guard placement is correct. Mostly small clarifications above. • Branch: `feat/oidc-require-verified-email`

2 We keep the defensive check.
4 Coverage with integration tests it enough.

All other addressed.

@claude review again

vseehausen:
@claude review again

vseehausen:
@claude review again. The minor obersactions are acceptable.

src/modules/users/domain/errors/user-email-required.error.ts
```diff
@@ -0,0 +1,14 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { HttpStatus, UnauthorizedException } from '@nestjs/common';
+
+export const EMAIL_REQUIRED_ERROR_CODE = 'email_required_error';
+
+export class UserEmailRequiredError extends UnauthorizedException {
```
PooyaRaki: We could either use `UnauthorizedException` with a different message, or introduce a `UserEmailError` and update the message accordingly. Having separate classes for different email-related cases feels excessive. I’d prefer the former approach.

src/modules/users/domain/users.repository.interface.ts
```diff
@@ -64,7 +64,7 @@ export interface IUsersRepository {
 
   findOrCreateByExtUserIdWithEmail(
     extUserId: string,
-    email?: { address: string; verified: boolean },
+    email: { address: string },
```
PooyaRaki: Two remarks:
1. The name should be `findOrCreateByExtUserIdAndEmail`, or even something more generic.
2. Since we no longer need `verified`, we can change the email type from an `object` to a `string`.

src/modules/users/domain/users.repository.ts
```diff
@@ -82,11 +83,12 @@ export class UsersRepository implements IUsersRepository {
   public async create(
     status: keyof typeof UserStatus,
     entityManager: EntityManager,
-    options?: { extUserId?: string },
+    options?: { extUserId?: string; email?: string },
   ): Promise<User['id']> {
     const userInsertResult = await entityManager.insert(DbUser, {
       status,
       ...(options?.extUserId && { extUserId: options.extUserId }),
+      ...(options?.email && { email: options.email.trim().toLowerCase() }),
```
PooyaRaki: We should move all trimming and lowercasing to the validator layer so we always end up with a ready-to-use email address.

src/modules/users/domain/users.repository.ts
```diff
@@ -230,107 +232,104 @@ export class UsersRepository implements IUsersRepository {
   }
 
   /**
-   * Finds or creates an OIDC user, then validates or persists the optional email claim.
+   * Finds or creates an OIDC user with the given verified email.
    */
   public async findOrCreateByExtUserIdWithEmail(
     extUserId: string,
-    email?: { address: string; verified: boolean },
+    email: { address: string },
```
PooyaRaki: When/If we change the email type, we need to update it everywhere.

src/modules/users/domain/users.repository.ts
```diff
@@ -230,107 +232,104 @@ export class UsersRepository implements IUsersRepository {
...
+          });
         },
       );
     } catch (error) {
-      // Handle race condition: a concurrent call may have created the
-      // user between our find and insert, causing a unique constraint
-      // violation. Retry the lookup in that case.
-      if (!this.isUniqueConstraintViolation(error, 'idx_users_ext_user_id')) {
-        throw error;
+      if (this.isUniqueConstraintViolation(error, 'idx_users_ext_user_id')) {
```
PooyaRaki, line 269: QQ: Why do we need to handle this case and return instead of rejecting?

src/modules/users/domain/users.repository.ts
```diff
@@ -230,107 +232,104 @@ export class UsersRepository implements IUsersRepository {
...
-    const userRepository =
-      await this.postgresDatabaseService.getRepository(DbUser);
-    const normalizedEmail = email.trim().toLowerCase();
-    // The DB unique index is the integrity guard for verified writes; this
-    // check blocks unverified duplicate emails before minting a CGW session.
-    const user = await userRepository.findOne({
-      where: { email: normalizedEmail },
-      select: { id: true },
+    // Race: a concurrent caller set the email between our find and update.
+    const fresh = await userRepository.findOne({
```
PooyaRaki: What is this for? This method persists email so why do we need to execute a find query?

src/modules/users/domain/users.repository.integration.spec.ts
```diff
@@ -966,28 +966,6 @@ describe('UsersRepository', () => {
       expect(user.email).toBe(email.toLowerCase());
     });
 
-    it('should not overwrite an existing email for the same user', async () => {
```
PooyaRaki, line 970: QQ: Why is this removed?

src/modules/users/domain/users.repository.integration.spec.ts
```diff
@@ -1001,37 +979,25 @@ describe('UsersRepository', () => {
       await expect(
         usersRepository.findOrCreateByExtUserIdWithEmail(extUserId, {
           address: email,
-          verified: true,
         }),
       ).rejects.toThrow(UserEmailAlreadyInUseError);
     });
 
-    it('should throw when an unverified email belongs to a different user', async () => {
+    it('should throw UserEmailMismatchError when the stored email differs from the OIDC email', async () => {
```
PooyaRaki: QQ: How is it possible for the OIDC email to differ from the stored email? Could this only happen because of lowercasing or trimming? I assume if the email differs the ID also differs which means that's another user.

src/modules/users/domain/users.repository.ts
vseehausen: Agree, I would add this with https://github.com/safe-global/safe-client-gateway/pull/3096. This moves validation upstream, and adds a type/schema, so downstream consumers can rely on the validated type. Otherwise we would need to manually check if the validation happened. I would leave it in this PR as it is, and implement a proper fix.

src/modules/users/domain/users.repository.integration.spec.ts
vseehausen: As all auth attempts come with verified emails, we can store the email and reject an email, when in use, as done in the next test. https://github.com/safe-global/safe-client-gateway/pull/3101/changes/BASE..ee9d6b8b8d7b2898e7ccd165a355bb40f8609323#diff-70dceb2f39dfa6764e074aa578fa4ef4bc93cd8c3189580adb09afe5ba895913L991

src/modules/users/domain/users.repository.integration.spec.ts
vseehausen: This is defensive. Auth0 should only return matching extId and mail pairs. We trust Auth0 on this, but I added this extra check to have a security guard on our side.

src/modules/users/domain/users.repository.ts
vseehausen: This is to prevent a race condition.

src/modules/users/domain/users.repository.integration.spec.ts
vseehausen: But removed with 77077bf82a44da92e0e44fc9dd6943ac2b05325e

src/modules/users/domain/users.repository.ts
vseehausen: Removed with 77077bf82a44da92e0e44fc9dd6943ac2b05325e

src/modules/users/domain/users.repository.ts
vseehausen: Removed with 77077bf82a44da92e0e44fc9dd6943ac2b05325e

src/modules/users/domain/users.repository.ts
vseehausen: done in 77077bf82a44da92e0e44fc9dd6943ac2b05325e

src/modules/users/domain/users.repository.interface.ts
vseehausen: Good point! Done in 77077bf82a44da92e0e44fc9dd6943ac2b05325e

src/modules/users/domain/errors/user-email-required.error.ts
vseehausen: Ok, I took `UnauthorizedException`: 77077bf82a44da92e0e44fc9dd6943ac2b05325e

LucieFaire (dismissed):

src/modules/users/domain/users.repository.ts
```diff
@@ -231,68 +232,94 @@ export class UsersRepository implements IUsersRepository {
...
+   * Reconciles a verified `email` against an existing user: returns the id when
+   * the email matches, backfills it when none is stored, and rejects a
+   * conflicting one.
+   */
+  private async reconcileEmail(
+    user: Pick<DbUser, 'id' | 'email'>,
+    email: EmailAddress,
+  ): Promise<User['id']> {
+    if (user.email) {
+      if (user.email !== email) {
```
tmjssz, line 309: Should we also normalize emails before comparison here?

LucieFaire: not really, `email` is already normalized through Zod schema `EmailAddress` and the DB record was normalized on `save`

tmjssz (dismissed):

LucieFaire (approved):

tmjssz (approved):

PR #3104, opened 2026-05-19
feat(spaces): support email invites alongside wallet invites

vseehausen:
Resolves: https://linear.app/safe-global/issue/WA-2327/add-email-invite-endpoint

Adds email as an invite identifier on the space invite endpoint, alongside the existing wallet identifier.

## Summary

- New `EmailInviteUserDto` and `WalletInviteUserDto` make the two invite shapes distinct at the API level; the request body is a discriminated array on `type`.
- Backward-compat preprocess defaults a missing `type` to `wallet` so existing wallet-only clients keep working during the wallet-1 migration.
- New `UsersRepository.findOrCreatePendingByEmail`: race-safe `INSERT … ON CONFLICT DO NOTHING` + `SELECT`, accepts an optional outer `EntityManager` so the placeholder lives inside the invite transaction.
- Per-item dispatch in `MembersRepository.inviteUsers`: wallet items keep the existing find-wallet-or-create-user-and-wallet path; email items resolve their user id via the new repo method.
- OIDC sign-in resolution gains a new step (`linkPendingInviteeByEmail`): when a PENDING placeholder matches the verified email of the signing-in user, claim it by setting `extUserId` and flipping the user's status to `ACTIVE`. Race-safe via `WHERE extUserId IS NULL` + affected check + re-read.

## Details

- **Two DTO types** with a discriminated union on `type` — chosen over a single permissive DTO so each item is unambiguously one shape at the API level, and to leave room for future invite identifier types without reshaping the endpoint.
- **Builds on #3101 (require verified email for OIDC sign-in)**. That PR's invariant — emails reaching the repo are verified — is what lets the new claim step here trust the email claim. Without that guard, OIDC sign-in with an unverified email could create a second `User` row alongside the placeholder, leaving them un-reconcilable without merge code; this PR avoids needing to write any such reconciliation.

### DIFF
--- a/src/modules/spaces/routes/entities/__tests__/invite-users.dto.entity.spec.ts
+++ b/src/modules/spaces/routes/entities/__tests__/invite-users.dto.entity.spec.ts
@@ -43,8 +43,7 @@ describe('InviteUsersDtoSchema', () => {
   });
 
   it('should infer type=wallet for legacy clients sending address without type', () => {
-    const legacy = walletInviteItem();
-    delete legacy.type;
+    const { type: _type, ...legacy } = walletInviteItem();
 
     const result = InviteUsersDtoSchema.safeParse({ users: [legacy] });
 
@@ -53,8 +52,7 @@ describe('InviteUsersDtoSchema', () => {
   });
 
   it('should reject an email-shaped item that omits `type`', () => {
-    const legacy = emailInviteItem();
-    delete legacy.type;
+    const { type: _type, ...legacy } = emailInviteItem();
 
     const result = InviteUsersDtoSchema.safeParse({ users: [legacy] });
 
@@ -95,8 +93,7 @@ describe('InviteUsersDtoSchema', () => {
   });
 
   it('should reject an item missing the role field', () => {
-    const wallet = walletInviteItem();
-    delete wallet.role;
+    const { role: _role, ...wallet } = walletInviteItem();
 
     const result = InviteUsersDtoSchema.safeParse({ users: [wallet] });
 
@@ -104,8 +101,7 @@ describe('InviteUsersDtoSchema', () => {
   });
 
   it('should reject a wallet item missing the address', () => {
-    const wallet = walletInviteItem();
-    delete wallet.address;
+    const { address: _address, ...wallet } = walletInviteItem();
 
     const result = InviteUsersDtoSchema.safeParse({ users: [wallet] });
 
@@ -121,8 +117,7 @@ describe('InviteUsersDtoSchema', () => {
   });
 
   it('should reject an email item missing the email', () => {
-    const email = emailInviteItem();
-    delete email.email;
+    const { email: _email, ...email } = emailInviteItem();
 
     const result = InviteUsersDtoSchema.safeParse({ users: [email] });
 
@@ -147,17 +142,40 @@ describe('InviteUsersDtoSchema', () => {
   });
 
   it('should reject an item missing the name field', () => {
-    const wallet = walletInviteItem();
-    delete wallet.name;
+    const { name: _name, ...wallet } = walletInviteItem();
 
     const result = InviteUsersDtoSchema.safeParse({ users: [wallet] });
 
     expect(result.success).toBe(false);
   });
 
-  it('should reject a name exceeding 255 characters', () => {
+  it('should reject a name exceeding 30 characters', () => {
     const result = InviteUsersDtoSchema.safeParse({
-      users: [{ ...walletInviteItem(), name: 'a'.repeat(256) }],
+      users: [{ ...walletInviteItem(), name: 'a'.repeat(31) }],
+    });
+
+    expect(result.success).toBe(false);
+  });
+
+  it('should reject a name shorter than 3 characters', () => {
+    const result = InviteUsersDtoSchema.safeParse({
+      users: [{ ...walletInviteItem(), name: 'ab' }],
+    });
+
+    expect(result.success).toBe(false);
+  });
+
+  it('should reject a whitespace-only name', () => {
+    const result = InviteUsersDtoSchema.safeParse({
+      users: [{ ...walletInviteItem(), name: '   ' }],
+    });
+
+    expect(result.success).toBe(false);
+  });
+
+  it('should reject a name with invalid characters', () => {
+    const result = InviteUsersDtoSchema.safeParse({
+      users: [{ ...walletInviteItem(), name: '<>@!' }],
     });
 
     expect(result.success).toBe(false);
--- a/src/modules/spaces/routes/entities/__tests__/invite-users.dto.entity.spec.ts
+++ b/src/modules/spaces/routes/entities/__tests__/invite-users.dto.entity.spec.ts
@@ -1,26 +1,16 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { faker } from '@faker-js/faker';
 import { getAddress } from 'viem';
+import {
+  emailInviteUserDtoBuilder,
+  walletInviteUserDtoBuilder,
+} from '@/modules/spaces/routes/entities/__tests__/invite-user.dto.builder';
 import { InviteUsersDtoSchema } from '@/modules/spaces/routes/entities/invite-users.dto.entity';
-
-const walletInviteItem = (): Record<string, unknown> => ({
-  type: 'wallet',
-  address: getAddress(faker.finance.ethereumAddress()),
-  role: 'MEMBER',
-  name: faker.person.firstName(),
-});
-
-const emailInviteItem = (): Record<string, unknown> => ({
-  type: 'email',
-  email: faker.internet.email(),
-  role: 'MEMBER',
-  name: faker.person.firstName(),
-});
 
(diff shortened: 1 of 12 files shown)


vseehausen:
@claude review

vseehausen:
@claude review

src/modules/spaces/routes/entities/invite-users.dto.entity.ts
```diff
@@ -9,42 +9,103 @@ import {
...
+
   @ApiProperty({
     type: String,
     minLength: NAME_MIN_LENGTH,
     maxLength: NAME_MAX_LENGTH,
   })
   public readonly name!: string;
+}
+
+export class EmailInviteUserDto {
```
LucieFaire: I would also add `implements z.infer<typeof EmailInviteUserSchema>` here to keep the DTo in sync with the entity type. Also you dont use this class anywhere outside so you can remove `export`

src/modules/spaces/routes/entities/invite-users.dto.entity.ts
```diff
@@ -9,42 +9,103 @@ import {
...
+  return raw;
+}, InviteUserSchema);
 
 export const InviteUsersDtoSchema = z.object({
-  users: InviteUserDtoSchema,
+  users: z.array(InviteUserSchemaWithWalletDefault).min(1),
 });
 
-export class InviteUserDto {
+export class WalletInviteUserDto {
```
LucieFaire: I would also add `implements z.infer<typeof WalletInviteUserSchema>` here to keep the DTo in sync with the entity type. Also you dont use this class anywhere outside so you can remove `export`

src/modules/users/domain/members.repository.integration.spec.ts
```diff
@@ -812,17 +819,93 @@ describe('MembersRepository', () => {
...
+        status: 'ACTIVE',
+      });
+      const spaceId = space.generatedMaps[0].id;
+      await dbMembersRepository.insert({
+        user: owner,
+        space: space.generatedMaps[0],
+        name: nameBuilder(),
+        role: 'ADMIN',
+        status: 'ACTIVE',
+        invitedBy: faker.number.int({ max: DB_MAX_SAFE_INTEGER }),
```
LucieFaire: admin is never invited by anyone, so this should be `null`

src/modules/users/domain/members.repository.integration.spec.ts
```diff
@@ -812,17 +819,93 @@ describe('MembersRepository', () => {
...
```
LucieFaire: same here, `null`

vseehausen:
@claude review again

src/modules/users/domain/members.repository.ts
```diff
@@ -151,9 +168,13 @@ export class MembersRepository implements IMembersRepository {
...
           if (isUniqueConstraintError(err)) {
-            throw new UniqueConstraintError(
-              `${userToInvite.address} is already in this space or has a pending invite.`,
-            );
+            // Wallet addresses are public; emails are PII and kept out of
+            // the message to avoid leaking via server logs.
+            const message =
+              userToInvite.type === InviteType.Wallet
+                ? `${userToInvite.address} is already in this space or has a pending invite.`
+                : 'This invitee is already in this space or has a pending invite.';
```
LucieFaire: hm, for email its a bit hard to tell which user failed without any identification. Maybe we can show masked email or smth, but we can keep it as is for now

src/modules/spaces/routes/entities/__tests__/invite-users.dto.entity.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+  role: 'MEMBER',
+  name: faker.person.firstName(),
+});
+
+const emailInviteItem = (): Record<string, unknown> => ({
+  type: 'email',
+  email: faker.internet.email(),
+  role: 'MEMBER',
+  name: faker.person.firstName(),
+});
```
LucieFaire: hello test builders :)

src/modules/users/domain/members.repository.ts
```diff
@@ -108,37 +107,37 @@ export class MembersRepository implements IMembersRepository {
...
-            });
+        let userIdToInvite: User['id'];
+        switch (userToInvite.type) {
+          case InviteType.Wallet: {
+            userIdToInvite =
+              await this.usersRepository.findOrCreateByWalletAddress(
+                userToInvite.address,
+                'PENDING',
+              );
+            break;
```
LucieFaire, line 155: why did you decide to change it? one query for many wallets was more performant than a query for each one, no?

vseehausen:
@claude review again. We didn't add comments for obvious code, that's fine.

LucieFaire (commented):
Im still looking at the repo methods, but otherwise looks ok. Small remarks.
I also think we need to merge expiration PR before this one so its easier to adjust

Update: repo looks ok

src/modules/users/domain/members.repository.ts
vseehausen: Yes, but the previous implementation had a race condition. Also, the client has the multi-invite flow on the space creation, plus I think the performance improvements are not priority here. However, I can revert back, if there are good arguments for it.

src/modules/spaces/routes/entities/invite-users.dto.entity.ts
vseehausen: addressed in 9a9b5c394

src/modules/users/domain/members.repository.integration.spec.ts
vseehausen: addressed in 9a9b5c394

src/modules/users/domain/members.repository.ts
vseehausen: addressed in 9a9b5c394

src/modules/spaces/routes/entities/invite-users.dto.entity.ts
vseehausen: addressed in 9a9b5c394

src/modules/users/domain/members.repository.integration.spec.ts
vseehausen: addressed in 9a9b5c394

src/modules/spaces/routes/entities/__tests__/invite-users.dto.entity.spec.ts
vseehausen: addressed in 7b2576d98

vseehausen:
@claude review again

src/modules/users/domain/users.repository.ts
```diff
@@ -334,6 +427,8 @@ export class UsersRepository implements IUsersRepository {
         .where('id = :userId', { userId })
         .andWhere('email IS NULL')
         .execute();
+      // Zero affected rows means another request backfilled this user after
+      // reconcileEmail saw email=null; that is idempotent for this path.
```
LucieFaire: nit, not sure we need this comment here

src/modules/users/domain/users.repository.ts
```diff
@@ -284,6 +340,43 @@ export class UsersRepository implements IUsersRepository {
...
+      throw new UserEmailAlreadyInUseError();
+    }
+
+    const result = await userRepository.update(
+      { id: candidate.id, status: 'PENDING', extUserId: IsNull() },
+      { extUserId, status: 'ACTIVE' },
+    );
+
+    // Lost a race to a concurrent claim; let the caller fall through to
+    // INSERT, where idx_users_email will surface UserEmailAlreadyInUseError.
```
LucieFaire: we already mention that in JSDoc comment :)

vseehausen: addressed in c625cf1c4

src/modules/users/domain/users.repository.ts
vseehausen: addressed in c625cf1c4

src/modules/users/domain/users.repository.ts
```diff
@@ -195,53 +195,102 @@ export class UsersRepository implements IUsersRepository {
...
+    entityManager?: EntityManager,
   ): Promise<User['id']> {
-    const existing = await this.findByWalletAddress(address);
-    if (existing) {
-      return existing.id;
-    }
+    const findOrCreate = async (
+      manager: EntityManager,
+    ): Promise<User['id']> => {
+      const existing = await manager.findOne(Wallet, {
```
LucieFaire, line 206: we have the same code on line 229. We can live with this duplication, but its not ideal. We also have a lot of `findByWalletAddress` method variations now instead of reusing 1 canonical. Something to refactor and maybe to keep an eye on  in the future

LucieFaire (approved):

LucieFaire (dismissed):

LucieFaire (approved):

src/modules/users/domain/users.repository.ts
vseehausen: You are right. I created a ticket for that.

PR #3105, opened 2026-05-19
feat: integrate chains repository into relay repository for relayer type resolution

akshay-ap:
Related to: https://linear.app/safe-global/issue/PLA-1267/cgw-refactor-cgw-env-vars-for-relayer-selection-logic
Related PR: https://github.com/safe-global/safe-config-service/pull/1555

This pull request refactors the relayer selection logic CGW. Instead of relying on env vars from configuration, it CGW checks the `relayer` object from config service per-chain.

**Configuration and Validation Refactor:**
- Removed `enabledChainIds` and `dailyLimitRelayerChainsIds` from both runtime and test configurations, as well as the corresponding environment variables and schema validation.

**Chain Entity and Schema Updates:**
- Added a nullable `relay.type` field to the `Chain` entity and its Zod schema, allowing each chain to specify its relayer type and other boolean flags for fine-grain control. Builders and tests updated to support this.

**Fee Service Logic Changes:**
- Updated `FeesService` to determine if relay-fee is available by checking the chain’s `relayer.type` for `RELAY_FEE`, instead of using the removed `isPayWithSafeEnabled` method.
- The `IFeeServiceApi.isPayWithSafeEnabled` method and its implementation/tests were removed as they are no longer needed.

**Test Suite Adjustments:**
- Integration and unit tests for fee previews were updated to mock and assert against the new approach, including dynamic chain building and network responses.

**Dependency Injection and Module Imports:**
- Added `ChainsModule` as a dependency of the `FeesModule` to enable chain lookups.

## Up for discussion
- [ ] If the chain has no relayer defined in config service i.e. `relay.type=null`, should relayManager throw `NoRelayerDefinedError` leading to failed `getRelaysRemaining` api call or return status `200` with `{remaining: 0, limit:0 }`
    https://github.com/safe-global/safe-client-gateway/blob/d4db3c92a1ad14e6c21499c1e58831624889433f/src/modules/relay/routes/relay.controller.ts#L213
    https://github.com/safe-global/safe-client-gateway/blob/d4db3c92a1ad14e6c21499c1e58831624889433f/src/modules/relay/domain/relay.manager.ts#L60-L61


PooyaRaki:
@claude review

akshay-ap:
@claude review

akshay-ap:
> **Claude finished @akshay-ap's task in 1m 21s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/27206455540)
> 
> ### Claude PR Review
> * [x]  Read CLAUDE.md and gather context
> * [x]  Analyze the diff
> * [x]  Review key files
> * [x]  Post review feedback
> 
> Nice refactor — moving relayer selection out of env vars and into per-chain config is a clear improvement. A few things to consider before merging:
> 
> #### 1. Schema silently swallows unknown relayer types
> `src/modules/chains/domain/entities/schemas/chain.schema.ts:75`
> 
> ```ts
> type: z.enum(RelayerType).nullable().catch(null),
> ```
> 
> Any future relayer type added in the config service (e.g. `GTF_V2`) becomes `null` here, which then triggers `NoRelayerDefinedError` → 403 with no log line. This is a silent outage waiting to happen. Use `z.preprocess` / `.transform` and `loggingService.warn` on the fallback so divergence is observable.
> 
> #### 2. `RelayerTypeNotImplementedError` → 403 is the wrong status
> `src/modules/relay/domain/errors/relayer-type-not-implemented.error.ts:4` extends `ForbiddenException`. Semantically this is "server hasn't implemented this," not "you're not allowed." Use `HttpStatus.NOT_IMPLEMENTED` (501) in the exception filter (the filter currently hard-codes 403 for both cases). Also narrow `relayerType: string` → `RelayerType` since callers only pass enum values.
> 
> #### 3. Inconsistent error codes for the same logical state
> * `src/modules/fees/routes/fees.service.ts:25-28` → 400 (`BadRequestException`) when `relayer.type !== RELAY_FEE`.
> * `src/modules/relay/domain/relay.manager.ts:60` → 403 (`NoRelayerDefinedError`).
> 
> Both mean "chain not configured for this relay flow." Pick one contract so clients can branch consistently.
> 
> #### 4. Rollout sequencing risk
> `src/modules/relay/domain/relay.manager.ts:50` — the old behavior fell back to `DailyLimitRelayer` for unconfigured chains. Now any chain whose `relayer.type` is `null` in the config-service response returns 403. Coupled with the chain cache in `config-api.service.ts`, a partial config-service rollout (or a stale cache window) means live 403s on `POST /v1/chains/:id/relay`. Confirm the config-service PR ships first and is fully populated, and consider an admin hook to invalidate the chain cache when relayer config changes.
> 
> #### 5. `relayerBuilder` returns `null` ~20% of the time by default
> `src/modules/chains/domain/entities/__tests__/relayer.builder.ts:13-18`
> 
> ```ts
> .with('type', faker.helpers.arrayElement([...Object.values(RelayerType), null]))
> ```
> 
> Any test that uses `chainBuilder().build()` without overriding `relayer.type` will randomly route to `GTF` (501-ish) or `null` (403). Either default to a valid non-`GTF` value, or always force the field explicitly in tests — otherwise this becomes a flaky-test source.
> 
> #### 6. `getRelaysRemaining` behavior change for RELAY_FEE without `safeTxHash`
> `src/modules/relay/domain/relayers/relay-fee.relayer.ts:332-336` — when no `safeTxHash` is provided on a `RELAY_FEE` chain, it now returns `{ remaining: 1, limit: 1 }` optimistically (was `{ 0, 0 }` previously via the env-var guard). Clients polling without `safeTxHash` will see a different contract. Worth a doc note and/or a test asserting this is intentional.
> 
> #### 7. Exception filter applied per-method
> `src/modules/relay/routes/relay.controller.ts:129` and `:212` register `RelayerNotAvailableExceptionFilter` on each route. Since every route in this controller routes through `RelayManager.getRelayer`, promote it to a class-level `@UseFilters` to remove the chance of forgetting it on a future endpoint.
> 
> #### Minor
> * `src/modules/relay/domain/relay.repository.ts:30,52` — `chain.relayer?.type ?? null` is duplicated; consider a small helper or destructure once.
> * `src/modules/relay/routes/relay.controller.integration.spec.ts:80-88` — `relayerTypeForChainId` reimplements the old env-var partition inside the test; setting `relayerType` directly on the chain builder per case would make the routing logic the SUT actually exercises more explicit.
> 
> Test coverage on the new branching logic looks solid. Top concerns are **#1 (silent schema catch)**, **#2/#3 (status-code semantics)**, and **#4 (rollout sequencing)**. • Branch: `feat/relay-type-from-config-service`

1. Invalid
2. Fixed https://github.com/safe-global/safe-client-gateway/pull/3105/commits/391da53f91a77f82aaf8ebc5f6bd17c9162ab591
3. Invalid
4. Accepted risk. We will make sure this does not happen
5. Fixed https://github.com/safe-global/safe-client-gateway/pull/3105/commits/8f9a4e929c2abed4f712724e98d6af918c693b84
6. Restored existing behavior https://github.com/safe-global/safe-client-gateway/pull/3105/commits/d4db3c92a1ad14e6c21499c1e58831624889433f

yagopv (commented):
@akshay-ap this PR launch a `NoRelayerDefinedError` when no relayer is defined. This is against the:

> Stage 1 — current behaviour (default / kill switch): no fee collected; Gelato relay not mandatory (non-Gelato/EOA flow allowed). Reverting to this stage is the kill switch / security fallback.

Defined in https://linear.app/safe-global/issue/PLA-1560/backend-per-chain-gtf-stage-flag-3-stages-incl-kill-switch

Or am I missing something? I thought that when the relayer is not defined, we default to the current production interface where connected wallets pay

akshay-ap:
> @akshay-ap this PR launch a `NoRelayerDefinedError` when no relayer is defined. This is against the:
> 
> > Stage 1 — current behaviour (default / kill switch): no fee collected; Gelato relay not mandatory (non-Gelato/EOA flow allowed). Reverting to this stage is the kill switch / security fallback.
> 
> Defined in https://linear.app/safe-global/issue/PLA-1560/backend-per-chain-gtf-stage-flag-3-stages-incl-kill-switch
> 
> Or am I missing something? I thought that when the relayer is not defined, we default to the current production interface where connected wallets pay

No relayer enabled means connected connected wallets pay for the transaction. Gelato relayer/Fee service will not be involved. This is the current default behavior on Mainnet.

yagopv (commented):
There are leftovers of the env variables RELAY_DAILY_LIMIT_CHAIN_IDS, RELAY_FEE_CHAIN_IDS, RELAY_SIMULATION_CHAIN_IDS in .env.sample.json and .devcontainer/docker-compose.yml

src/modules/fees/routes/fees.service.ts
```diff
@@ -2,22 +2,27 @@
...
   ) {}
 
   async getFeePreview(args: {
     chainId: string;
     safeAddress: Address;
     feePreviewDto: FeePreviewTransactionDto;
   }): Promise<FeePreviewResponse> {
-    if (!this.feeServiceApi.isPayWithSafeEnabled(args.chainId)) {
+    const chain = await this.chainsRepository.getChain(args.chainId);
+    if (chain.relayer?.type !== RelayerType.RELAY_FEE) {
```
yagopv, line 56: Is this correct instead simply retrieving the configured relayer and that's it? I believe this is not true for Phase 3, as we will have a new specific GTF relayer (I guess)

Additionally, the error seems strange to me: "Pay with Safe not available for this chain." Technically, we will be able to "Pay with Safe" in Phase 3, right?  More accurate would be something like "You haven't configured a relayer for this chain that uses the Fee Service."

akshay-ap: > Is this correct instead simply retrieving the configured relayer and that's it? I believe this is not true for Phase 3, as we will have a new specific GTF relayer (I guess)

Yes, we will have GTF specific relayer and make changes again later. This PR is not supporting GTF relayer.

akshay-ap: > You haven't configured a relayer for this chain that uses the Fee Service.

lets do that.

akshay-ap:
> There are leftovers of the env variables RELAY_DAILY_LIMIT_CHAIN_IDS, RELAY_FEE_CHAIN_IDS, RELAY_SIMULATION_CHAIN_IDS in .env.sample.json and .devcontainer/docker-compose.yml

https://github.com/safe-global/safe-client-gateway/pull/3105/commits/2d413d64da7212017793cc99fc569fd4b0ea1534

src/modules/fees/routes/fees.service.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3105/commits/28bb6a4d4e7dd3e613099fa4070fb918abb3aa45

yagopv (dismissed):

src/modules/chains/domain/entities/schemas/chain.schema.ts
```diff
@@ -70,6 +71,13 @@ export const BalancesProviderSchema = z.object({
   enabled: z.boolean(),
 });
 
+export const RelayerSchema = z.object({
+  type: z.enum(RelayerType).nullable().catch(null),
+  safeCreationSponsored: z.boolean().catch(false).default(false),
+  safeTransactionSponsored: z.boolean().catch(false).default(false),
+  enableTenderlySimulationBeforeRelay: z.boolean().catch(false).default(false),
+});
```
LucieFaire, line 79:
```
const booleanWithFalseFallback = z.boolean().catch(false);

export const RelayerSchema = z.object({
  type: z.enum(RelayerType).nullable().catch(null),
  safeCreationSponsored: booleanWithFalseFallback,
  safeTransactionSponsored: booleanWithFalseFallback,
  enableTenderlySimulationBeforeRelay: booleanWithFalseFallback,
});
```
you can skip `.default(false)` as `.catch(false)` already handles undefined, invalid strings, null, numbers, etc.

src/modules/chains/routes/entities/chain.entity.ts
```diff
@@ -97,6 +98,8 @@ export class Chain {
   theme: ApiTheme;
   @ApiPropertyOptional({ type: String, nullable: true })
   recommendedMasterCopyVersion: string | null;
+  @ApiPropertyOptional({ type: Relayer, nullable: true })
+  relayer: Relayer | null;
```
LucieFaire, line 102: so it can be `some value | null  | undefined` ? According to schema I dont see `undefined` allowed.
I think Swagger decorator should be:
```
@ApiProperty({ type: Relayer, nullable: true })
```

LucieFaire (dismissed):
2 small remarks, otherwise looks good !

src/modules/chains/domain/entities/schemas/chain.schema.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3105/commits/0ab3dd7066978657109e361b2fe79dce75bd37c8

src/modules/chains/routes/entities/chain.entity.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3105/commits/52e08ea60634d2f7d77c2edc41e26f218837380b

LucieFaire (approved):

PR #3107, opened 2026-05-20
feat: per-Space onboarding survey module

tongyaosafe:
## Summary

Resolves [WA-2336](https://linear.app/safe-global/issue/WA-2336/space-onboarding-survey-how-will-you-use-safe).

Adds a new `surveys` module to CGW that backs the Space onboarding survey UI in the Wallet Web app. Captures per-Space usage intent (e.g. `run_payments`, `hold_assets`) for analytics and future per-Space personalisation. Owner-only, idempotent, and structured so multi-page surveys ship later without a wire-format break.

**Branch dependency**: the matching frontend PR in `safe-wallet-monorepo` (`feat/space-onboarding-survey`) consumes this module's endpoints. This PR should merge and deploy to staging **before** the monorepo PR's review cycle so the OpenAPI schema is regenerable.

### Data model

Two new tables (single migration: `1779100947599-create-surveys.ts`).

```sql
surveys (
  id, slug, version, title, subtitle,
  survey_content jsonb NOT NULL,    -- { pages: [{ id, title, subtitle, multiSelect, options }] }
  is_active boolean DEFAULT false,
  created_at,
  UNIQUE (slug, version)
)
-- partial index for the "active" lookup hot path:
idx_surveys_active ON surveys(slug) WHERE is_active = true

survey_responses (
  id, space_id, survey_id, answered_by_user_id (audit, ON DELETE SET NULL),
  selections jsonb NOT NULL,        -- { [pageId]: string[] }
  submitted_at, updated_at,
  UNIQUE (space_id, survey_id),     -- one response per Space per survey
  FK space_id ON DELETE CASCADE
)
-- BEFORE-UPDATE trigger update_updated_at on survey_responses (matches every other
-- timestamped table in this repo).
-- GIN index on selections for future analytics queries.
```

Migration seeds the `onboarding` v1 survey with the six use-case options the design specifies, `is_active = true`.

### API

```
GET  /v1/spaces/:spaceId/surveys/:slug/state
POST /v1/spaces/:spaceId/surveys/:slug/responses
```

Both require `AuthGuard` and reject non-admin members of the target space with `403`. The state endpoint returns the active survey definition **plus** the Space's prior response (or `null`) in a single round-trip. POST is idempotent via Postgres `INSERT ... ON CONFLICT (space_id, survey_id) DO UPDATE` (atomic — no find-then-write race).

Validation runs in two layers:

1. **Zod `ValidationPipe`** on the request body — rejects malformed shapes with `422`.
2. **Service-level `validateSelections`** — every page in the active survey must be present as a key, no unknown page ids, each page's selections must be a subset of that page's option keys, and single-select pages reject `>1` selection. All return `400` with a useful message.

### Module structure

Standard CGW three-tier under `src/modules/surveys/`:

```
domain/        Zod schemas, repository interface
datasources/   TypeORM entities (Survey, SurveyResponse)
routes/        Controller, service, DTOs, e2e tests
```

`SurveysModule` is wired into `AppModule` alongside the existing `users` feature flag group, so it goes live with `users` (no separate flag needed).

### Key design decisions (questions reviewers tend to ask)

- **Per-Space, not per-user**: a Space's onboarding answer is a property of the Space itself; the same human may have different answers for their treasury vs personal Spaces. `answered_by_user_id` is audit only — when the admin leaves, the row stays (`ON DELETE SET NULL`).
- **Admin-only write & read**: matches `PATCH /v1/spaces/:spaceId/members/:userId/role` and the rest of the "modify Space attribute" endpoints. Non-admin members never see the survey UI; the server enforces this anyway.
- **Mandatory (no skip)**: `selections` must be non-empty per page. There is no `status: 'skipped'` column. The frontend disables Finish until ≥1 option is selected.
- **`{ pages: [...] }` JSONB shape now, even though v1 ships one page**: locking the wire format up-front so a future multi-page survey is data-only — no API contract change, no Mixpanel property schema drift if/when the project pipes events to Mixpanel later.
- **Kill switch is `surveys.is_active`**: one `UPDATE surveys SET is_active = false WHERE slug = '…'` disables the survey for **all users globally**, no redeploy. The frontend treats a 404 from `/state` as "no active survey" and silently redirects the user to their Space dashboard.

### What I deliberately did **not** do

- **No Mixpanel server-side push**: CGW has no analytics infrastructure today (no `mixpanel`, `@segment`, `@amplitude`, or `posthog` deps anywhere in source/env). Data lives in `survey_responses` and is queryable via SQL — see "How to test" below. If/when we want event-stream analytics, that's a separate infra PR.
- **No admin UI for editing survey content**: copy changes go through a new migration row + `is_active` flip. Acceptable until volume justifies a CMS.
- **No second survey beyond `onboarding`**: the `slug` column is in place for it; adding one is a seed insert.

## How to test

```bash
# 1. Apply the migration (it runs automatically in dev / CI)
docker exec safe-wallet-dev-env-cgw-1 node -r tsconfig-paths/register \
  ./node_modules/typeorm/cli -d ./dist/src/config/entities/orm.config.js migration:run

# 2. Unit tests (8 cases, mocked repository)
yarn test:unit src/modules/surveys/routes/surveys.service.spec.ts

# 3. E2E tests (real Postgres, 10 cases including upsert + admin/non-admin/auth)
yarn test:e2e src/modules/surveys

# 4. Manual happy path
TOKEN='<admin-jwt>'        # SIWE-signed JWT for an admin of SPACE_ID
SPACE_ID=42
BASE=http://localhost:3030/v1/spaces/$SPACE_ID/surveys/onboarding

curl -H "Cookie: access_token=$TOKEN" $BASE/state                                # 200, spaceResponse: null
curl -H "Cookie: access_token=$TOKEN" -X POST $BASE/responses \
  -H 'Content-Type: application/json' \
  -d '{"selections":{"use_cases":["hold_assets","run_payments"]}}'              # 201
curl -H "Cookie: access_token=$TOKEN" $BASE/state                                # 200, spaceResponse populated

# 5. Manual error paths
# - Missing page id:   {"selections": {}}                                       # 400
# - Unknown page id:   {"selections": {"made_up": ["x"]}}                       # 400
# - Unknown key:       {"selections": {"use_cases": ["totally_made_up"]}}       # 400
# - Non-admin member of space submitting:                                       # 403
# - No access_token cookie:                                                     # 401/403

# 6. Kill switch
psql -c "UPDATE surveys SET is_active = false WHERE slug = 'onboarding';"
curl -H "Cookie: access_token=$TOKEN" $BASE/state                                # 404
# (Frontend treats this as "skip survey", redirecting to dashboard.)
```

## Open follow-ups (intentionally out of scope)

- Wire CGW → Mixpanel (server-side push from the POST handler), if/when product wants event-stream analytics. Adds `mixpanel` dep, a `MixpanelService`, and a project token env var.
- Multi-page UI on the frontend. Schema already supports it — just add more pages to `survey_content.pages`. No backend change required.
- Admin UI to edit survey copy / activate versions. Today: SQL update.
- Re-prompt behaviour when a Space's previous admin already submitted (frontend currently allows admin to re-submit; UX choice — see monorepo PR's "Risks").

## Risks / not verified

- E2E tests pass locally but the CI job uses a different Postgres instance — first CI run will exercise the migration `up()` for real.
- Migration `down()` drops both tables (which cascades the trigger and the seed row). Reversible.
- No load test on the GIN index on `selections` — ships now because future analytics queries are the only reason it exists.
- I have not verified the OpenAPI schema regeneration on the consumer side; that happens when the monorepo PR runs `yarn fetch-schema` against a deployed CGW.


tongyaosafe:
@claude review

PooyaRaki:
@claude review

src/modules/surveys/datasources/entities/survey-response.entity.db.ts
```diff
@@ -0,0 +1,73 @@
...
+  @JoinColumn({
+    name: 'space_id',
+    foreignKeyConstraintName: 'FK_survey_responses_space_id',
+  })
+  space!: Space;
+
+  @ManyToOne(
+    () => Survey,
+    (survey: Survey) => survey.id,
+    { nullable: false },
```
PooyaRaki:
```suggestion
    { nullable: false, onDelete: 'CASCADE' },
```

migrations/1779100947599-create-surveys.ts
```diff
@@ -0,0 +1,95 @@
...
+        "subtitle" character varying(255),
+        "survey_content" jsonb NOT NULL,
+        "is_active" boolean NOT NULL DEFAULT false,
+        "created_at" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        CONSTRAINT "PK_surveys_id" PRIMARY KEY ("id"),
+        CONSTRAINT "UQ_surveys_slug_version" UNIQUE ("slug", "version")
+      )
+    `);
+    await queryRunner.query(
+      `CREATE INDEX "idx_surveys_active" ON "surveys" ("slug") WHERE "is_active" = true`,
```
PooyaRaki: We've got some queries fetching all the surveys regardless of the status i.e. `active` or not.
```suggestion
      `CREATE INDEX "idx_surveys_active" ON "surveys" ("slug")`,
```

tongyaosafe:
> **Claude finished @PooyaRaki's task in 2m 59s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/26237001441)
> 
> ### Claude PR Review
> Solid, well-structured module — three-tier layout, atomic upsert, Zod + service-level validation, good test coverage. A few things worth tightening before merge:
> 
> #### Correctness
> * **`idx_surveys_active` should be `UNIQUE`** — `migrations/1779100947599-create-surveys.ts:23` and `src/modules/surveys/datasources/entities/survey.entity.db.ts:10`. The partial index makes lookup fast but does not prevent two rows from being `is_active = true` for the same `slug`. `findActiveBySlug` calls `findOne` and would silently return whichever row Postgres picks. Since the kill-switch story is "flip `is_active`," an operator who forgets to deactivate v1 before activating v2 gets nondeterministic results. Suggest:
>   ```sql
>   CREATE UNIQUE INDEX "uq_surveys_active" ON "surveys" ("slug") WHERE "is_active" = true
>   ```
> * **`slug` path param is unvalidated** — `src/modules/surveys/routes/surveys.controller.ts:64,96`. `@Param('slug') slug: string` accepts arbitrary input (any length, any chars) and feeds it straight into a DB lookup. The DB column is `varchar(64)`, so an oversized slug becomes a wasted query rather than a crash, but it's still worth a `ValidationPipe` with a small Zod schema (e.g. `z.string().min(1).max(64).regex(/^[a-z0-9-]+$/)`) for symmetry with `spaceId`.
> 
> #### Minor
> * **`upsertResponse` does two round-trips** — `src/modules/surveys/domain/surveys.repository.ts:56-75`. `repo.upsert()` doesn't return rows with relations, so you re-query. For a single-admin write path this is fine, but a single `INSERT ... ON CONFLICT ... RETURNING *` (raw query or QueryBuilder) would halve the latency and is trivial to write. Defer if not a hot path.
> * **Zod schema doesn't enforce min keys on `selections`** — `src/modules/surveys/domain/entities/survey-response.entity.ts:9-12`. `z.record(...)` allows `{}`, so an empty payload reaches the service layer and returns 400 instead of 422. The e2e tests work around this with `expect([400, 422]).toContain(res.status)` — that ambiguity is a code smell. A `.refine((rec) => Object.keys(rec).length > 0, …)` on the schema would let the e2e tests assert a single status.
> * **GIN index on `survey_responses.selections`** — `migrations/1779100947599-create-surveys.ts:45-47`. As the PR description acknowledges, this exists only for hypothetical future analytics. Worth deferring to whichever PR actually adds the analytics query — indexes have write-amplification cost.
> * **`SurveyPageSchema.subtitle` uses `.nullable().optional()`** — `src/modules/surveys/domain/entities/survey.entity.ts:15`. The seeded JSON always provides a subtitle, and the top-level `Survey.subtitle` is just `.nullable()` (not optional). Mismatched modeling for the same logical field; pick one — likely `.nullable()` everywhere — so consumers don't have to handle three states (missing / null / string).
> * **`SurveyResponseResultDto.surveySlug`** — `src/modules/surveys/routes/surveys.service.ts:91`. Returned from the request's path slug rather than the persisted survey row's slug. They'll always match here, but reading from `response.survey.slug` would be slightly more honest. Trivial.
> 
> #### Tests
> Coverage looks good — admin gate, validation, upsert idempotency, unauth/non-admin cases are all there. One gap: no test asserts the `update_updated_at` trigger actually fires on re-submit (i.e. `updatedAt > submittedAt` after the second POST). Worth one assertion in the existing upsert e2e test.
> 
> #### Security
> `AuthGuard` + active-admin gate on both endpoints is correct. JSONB stored is fully validated (page id allowlist, key allowlist), so no injection surface. No PII written beyond `answered_by_user_id`, which is `ON DELETE SET NULL`. Looks good.
> 
> Overall: ship-worthy. The unique-index fix and the slug-param validation are the only items I'd want addressed before merge; the rest are nice-to-haves. • Branch: [`feat/space-onboarding-survey`](https://github.com/safe-global/safe-client-gateway/tree/feat/space-onboarding-survey)

Most Comments are fixed.

The following are not fixed:
- upsertResponse does two round-trips. Reason: Admin click on Finish is one-shot per Space onboarding. The extra round-trip doesn't improve performance.
- GIN index on survey_responses.selections. Reason: This table will be used for analytics data source. Adding GIN index now makes sense.
-  SurveyResponseResultDto.surveySlug. Reason: The result is the same. Making this change does not make the code more readable.

src/modules/surveys/domain/surveys.repository.ts
```diff
@@ -0,0 +1,77 @@
...
+    await repo.upsert(
+      {
+        space: { id: args.spaceId },
+        survey: { id: args.surveyId },
+        answeredBy: { id: args.answeredByUserId },
+        selections: args.selections,
+      },
+      {
+        conflictPaths: ['space', 'survey'],
+        skipUpdateIfNoValuesChanged: false,
```
PooyaRaki: Why are we not skipping update if no values changed?

src/modules/surveys/domain/surveys.repository.ts
```diff
@@ -0,0 +1,77 @@
...
+      },
+    );
+
+    return await repo.findOneOrFail({
```
PooyaRaki: Couldn't we get the result from `generatedMaps` which is returned from the upsert Method(upsert returns `InsertResult` which contains that key). Otherwise we could map the raw result returned and not need another find query.

src/modules/surveys/surveys.module.ts
```diff
@@ -0,0 +1,31 @@
...
+import { ISurveysRepository } from '@/modules/surveys/domain/surveys.repository.interface';
+import { SurveysController } from '@/modules/surveys/routes/surveys.controller';
+import { SurveysService } from '@/modules/surveys/routes/surveys.service';
+import { UsersModule } from '@/modules/users/users.module';
+
+@Module({
+  imports: [
+    PostgresDatabaseModuleV2,
+    TypeOrmModule.forFeature([Survey, SurveyResponse]),
+    forwardRef(() => AuthModule),
```
PooyaRaki: Why do we need to use `forwardRef` here? Is there a circular dependency between this module and auth?

src/modules/surveys/routes/surveys.service.ts
```diff
@@ -0,0 +1,199 @@
...
+    });
+    if (!member) {
+      throw new ForbiddenException(
+        'User is not an active admin of this space.',
+      );
+    }
+    return userId;
+  }
+
+  private async findActiveSurveyOrFail(slug: Survey['slug']): Promise<Survey> {
```
PooyaRaki: This method should stay in the repository as it handles queries.

src/modules/surveys/routes/surveys.service.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    const survey = await this.findActiveSurveyOrFail(args.slug);
+    const response = await this.surveysRepository.findResponse({
+      spaceId: args.spaceId,
+      surveyId: survey.id,
+    });
+
+    return {
+      survey: this.toSurveyDto(survey),
+      spaceResponse: response ? this.toSpaceResponseDto(response) : null,
```
PooyaRaki: Should it be `surveyResponse` ?

src/modules/surveys/routes/surveys.service.ts
```diff
@@ -0,0 +1,199 @@
...
+    const submittedIds = new Set(Object.keys(args.submitted));
+
+    const missing = [...requiredIds].filter((id) => !submittedIds.has(id));
+    if (missing.length > 0) {
+      throw new BadRequestException(
+        `Missing answers for page(s): ${missing.join(', ')}`,
+      );
+    }
+
+    const unknownPages = [...submittedIds].filter((id) => !requiredIds.has(id));
```
PooyaRaki: since we have `requiredIds`, could we also have `optionalIds`? Looks like no considering this condition, therefor I would not name it `requiredIds`.

PooyaRaki:
@tongyaosafe This remark `upsertResponse does two round-trips.` from Claude should be fixed, it's the same as my remark about executing two queries: `upsert` and `find`.

migrations/1779100947599-create-surveys.ts
tongyaosafe: Hi Pooya,
Thanks for the suggestion.
This `WHERE "is_active" = true, ` is intended. Because for the same survey `slug`,  we may have different `version`, but we should only allow one version is active. So this index is a constraint.

src/modules/surveys/domain/surveys.repository.ts
tongyaosafe: You are right. Fixed.

src/modules/surveys/surveys.module.ts
tongyaosafe: We don't need it. fixed.

src/modules/surveys/routes/surveys.service.ts
tongyaosafe: Thanks. fixed

src/modules/surveys/routes/surveys.service.ts
tongyaosafe: Good catch. fixed

src/modules/surveys/routes/surveys.service.ts
tongyaosafe: changed to `allPageIds` Fixed

src/modules/surveys/domain/surveys.repository.ts
tongyaosafe: fixed.

src/modules/surveys/datasources/entities/survey-response.entity.db.ts
tongyaosafe: fixed

src/modules/surveys/domain/surveys.repository.ts
```diff
@@ -0,0 +1,95 @@
...
+        selections: args.selections,
+      })
+      .orUpdate(
+        ['selections', 'answered_by_user_id'],
+        ['space_id', 'survey_id'],
+      )
+      .returning(['id', 'submitted_at', 'updated_at'])
+      .execute();
+
+    const row = result.raw[0] as {
```
PooyaRaki: Why can't we replace it with something like:
```ts
 const { id, submittedAt, updatedAt } = result.generatedMaps[0] as Pick<
    DbSurveyResponse,
    'id' | 'submittedAt' | 'updatedAt'
  >;
  return { id, submittedAt, updatedAt };
```

src/modules/surveys/routes/surveys.service.spec.ts
```diff
@@ -0,0 +1,253 @@
...
+    id: 1,
+    slug: 'onboarding',
+    version: 1,
+    title: 'Space Onboarding Survey',
+    subtitle: null,
+    surveyContent: { pages },
+    isActive: true,
+    createdAt: new Date(),
+  };
+}
```
LucieFaire, line 40: nit, you can also create a builder as we have for other entities and then reuse it everywhere:)

src/modules/surveys/domain/surveys.repository.ts
PooyaRaki: I see you changed the `upsert` to `update` or `insert`? Is there any reason for that? We could just remove:
`skipUpdateIfNoValuesChanged: false,` or change the value to true.

src/modules/surveys/routes/surveys.controller.integration.spec.ts
```diff
@@ -0,0 +1,321 @@
...
+  const res = await request(app.getHttpServer())
+    .post('/v1/spaces')
+    .set('Cookie', [`access_token=${accessToken}`])
+    .send({ name: nameBuilder() })
+    .expect(201);
+
+  return res.body.id as number;
+}
+
+describe('SurveysController', () => {
```
LucieFaire, line 28: fyi, becuase it is e2e test (and not integration) it wil not be run as part of CI but only on demand. Maybe worth to also create integration tests if you want it to be run frequently

src/modules/surveys/surveys.module.ts
```diff
@@ -0,0 +1,31 @@
...
+  ],
+  controllers: [SurveysController],
+  providers: [
+    SurveysService,
+    {
+      provide: ISurveysRepository,
+      useClass: SurveysRepository,
+    },
+  ],
+  exports: [ISurveysRepository],
```
LucieFaire: why do we need to export it? Think we dont need to export anything, its self-sufficient module

migrations/1779100947599-create-surveys.ts
PooyaRaki: Ok, I see. You're only querying the `is_active: true` rows then the remark was invalid and what you did makes sense 👍 .

src/modules/surveys/routes/surveys.service.ts
```diff
@@ -0,0 +1,194 @@
...
+          `Unknown selection key(s) on page "${pageId}": ${unknown.join(', ')}`,
+        );
+      }
+      if (!page.multiSelect && deduped.length > 1) {
+        throw new BadRequestException(
+          `Page "${pageId}" is single-select but received ${deduped.length} selections`,
+        );
+      }
+      result[pageId] = deduped;
+    }
```
clovisdasilvaneto: can we extract it to a separated function? it is quite hard to understand when you do not have the business context

src/modules/surveys/domain/surveys.repository.ts
tongyaosafe: When I first did the refactor I thought `repo.upsert()` wouldn't populate `result.generatedMaps[0]` with `submittedAt` / `updatedAt` on the UPDATE branch, so I dropped to QueryBuilder to get explicit `.returning([...])` control. Re-checked the TypeORM source: `repo.upsert()` is itself a wrapper around `createQueryBuilder().insert().orUpdate().execute()`, and both columns have `default: () => 'CURRENT_TIMESTAMP'` annotations which `getInsertionReturningColumns()` picks up automatically. So `repo.upsert()` does return them in `generatedMaps`, and the QueryBuilder detour was unnecessary. Reverted in the same change.

src/modules/surveys/domain/surveys.repository.ts
tongyaosafe: Done.

src/modules/surveys/routes/surveys.service.ts
tongyaosafe: good suggestion.
Done

src/modules/surveys/routes/surveys.service.spec.ts
tongyaosafe: thanks. fixed

src/modules/surveys/routes/surveys.controller.integration.spec.ts
tongyaosafe: fixed

src/modules/surveys/surveys.module.ts
tongyaosafe: removed. No other modules imports ISurveysRepository

LucieFaire (dismissed):

LucieFaire (approved):

PR #3112, opened 2026-05-26
relay: replace RPC calls with safe-tx-service api call when validating safe-tx-hash

akshay-ap:
- This change is required because CGW has no infura key setup which is causing RPC calls to fail and relay checks to fail.
- The new approach directly compares the decoded transaction payload to the stored proposal.
- Removed the dependency on `IBlockchainApiManager` and all related blockchain RPC calls from `RelayTransactionHelper`
- Added a new private method `checkSafeTx` to `RelayTransactionHelper`
- Changes in tests

### DIFF
--- a/src/modules/relay/domain/__tests__/relay-transaction-helper.spec.ts
+++ b/src/modules/relay/domain/__tests__/relay-transaction-helper.spec.ts
@@ -415,6 +415,40 @@ describe('RelayTransactionHelper', () => {
         }),
       ).resolves.toBe(true);
     });
+
+    it('should match when stored data uses mixed-case hex but decoded data is lowercase', async () => {
+      const chainId = faker.helpers.arrayElement(supportedChainIds);
+      const safeAddress = getAddress(faker.finance.ethereumAddress());
+      const safeTxHash = faker.string.hexadecimal({
+        length: 64,
+        casing: 'lower',
+      }) as Hex;
+      // erc20 transfer payload — non-empty data so case can actually differ.
+      const recipient = getAddress(faker.finance.ethereumAddress());
+      const execData = execTransactionEncoder()
+        .with('data', erc20TransferEncoder().with('to', recipient).encode())
+        .encode();
+      const decoded = helper.decodeExecTransaction(execData)!;
+      // Tx service may return any hex casing; viem-decoded data is lowercase.
+      // Sanity-check the precondition is non-trivial.
+      expect(decoded.data).toBe(decoded.data.toLowerCase());
+      const stored = buildMatchingStored({
+        safeAddress,
+        decodedTo: decoded.to,
+      })
+        .with('data', decoded.data.toUpperCase().replace('0X', '0x') as Hex)
+        .build();
+      mockSafeRepository.getMultiSigTransaction.mockResolvedValue(stored);
+
+      await expect(
+        helper.isSafeTxHashValid({
+          chainId,
+          safeAddress,
+          decoded,
+          safeTxHash,
+        }),
+      ).resolves.toBe(true);
+    });
   });
 
   describe('decodeExecTransaction', () => {
--- a/src/modules/relay/domain/__tests__/relay-transaction-helper.spec.ts
+++ b/src/modules/relay/domain/__tests__/relay-transaction-helper.spec.ts
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { faker } from '@faker-js/faker';
-import type { Hex, PublicClient } from 'viem';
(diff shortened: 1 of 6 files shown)


akshay-ap:
@claude review

akshay-ap:
> **Claude finished @akshay-ap's task in 2m 38s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/26441302893)
> 
> ### Claude PR Review
> Overall the approach looks sound — removing the RPC dependency and trusting the tx service's `safeTxHash` invariant is a reasonable simplification. Tests cover the main cases (happy path, field mismatch, lookup failure, safe mismatch, null defaults). A few things worth tightening:
> 
> #### Suggestions
> 1. **Cache freshness for relay validation** — `src/modules/relay/domain/relay-transaction-helper.ts:491`
>    `safeRepository.getMultiSigTransaction(...)` is the cached variant. For a freshly proposed transaction, the cache may not yet contain it (causing a false rejection of a legitimate relay request), or it may return an outdated record. There is a `getMultiSigTransactionWithNoCache` sibling — consider using it here, since correctness > one extra round trip on relay-fee verification. [Fix this →](https://claude.ai/code?q=In%20src%2Fmodules%2Frelay%2Fdomain%2Frelay-transaction-helper.ts%20around%20line%20491%20on%20branch%20fix%2Fsafe-tx-hash%2C%20switch%20safeRepository.getMultiSigTransaction%20to%20safeRepository.getMultiSigTransactionWithNoCache%20for%20freshness%2C%20and%20update%20the%20mocks%20in%20src%2Fmodules%2Frelay%2Fdomain%2F__tests__%2Frelay-transaction-helper.spec.ts%20accordingly.&repo=safe-global/safe-client-gateway)
> 2. **Case-sensitive `data` comparison can yield false rejections** — `src/modules/relay/domain/relay-transaction-helper.ts:557`
>    ```ts
>    if ((stored.data ?? '0x') !== decoded.data) { mismatches.push('data'); }
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    `decoded.data` from viem is lowercase, but `NullableHexSchema` only validates the `0x...` shape and does not normalize case. If the tx-service-stored payload contains any uppercase hex (mixed-case is legal for hex), this strict-equality compare reports a false mismatch. Normalize both sides (`.toLowerCase()`) before comparing.
> 3. **Defense-in-depth: verify `stored.safeTxHash`** — `src/modules/relay/domain/relay-transaction-helper.ts:491`
>    The whole correctness argument rests on “the tx service computed the stored `safeTxHash` correctly.” Cheap to also assert `stored.safeTxHash.toLowerCase() === args.safeTxHash.toLowerCase()` — protects against any internal cache-key/lookup oddity returning a record under the wrong key.
> 4. **Consider rejecting already-executed proposals** — `src/modules/relay/domain/relay-transaction-helper.ts:510`
>    If `stored.isExecuted === true`, relaying again wastes a relay quota and will revert on-chain. Cheap early rejection with a clearer log line.
> 5. **`version` arg is now unused** — `src/modules/relay/domain/relay-transaction-helper.ts:483`
>    With the ABI/RPC path gone, `args.version` is no longer read in `isSafeTxHashValid`. Removing it (and from the caller in `relay-fee.relayer.ts`) avoids misleading future readers — minor cleanup.
> 
> #### Nits
> * `relay-transaction-helper.spec.ts`: the `buildMatchingStored` return type uses a `infer B extends … : never` trick to dodge naming the builder type — `ReturnType<typeof multisigTransactionBuilder>` directly would be simpler and clearer.
> * Test for `to` mismatch overrides `to` to a random address — also worth a focused test for `value` / `operation` mismatches to make the field-by-field intent explicit, though the current coverage of the mismatch branch is OK.
> 
> #### Security / behavior
> The behavior change shifts the trust anchor from the chain (RPC `getTransactionHash`) to the Safe Transaction Service. This is acceptable for relay quota gating (the worst case is letting through a tx that on-chain validation will then revert), but should be deliberate — worth confirming the team is comfortable that compromise/misbehavior of the tx service no longer fails closed at this layer.
> 
> • Branch: `fix/safe-tx-hash`

1. Invalid
2. Fix: https://github.com/safe-global/safe-client-gateway/pull/3112/commits/c60bb6eb44ecfe196688d223a844689f7e20b0ec
3. Not needed
4. Invalid. Fee service checks if tx is relayed or not. No need to redo the checks
5. Removed: https://github.com/safe-global/safe-client-gateway/pull/3112/commits/4c4f6ef19d78e0cd19372f486a3f3d28531abb22

yagopv (approved):

tmjssz (approved):
LGTM 👍

PR #3114, opened 2026-05-27
fix(hooks): invalidate nested Safe tx cache on parent approveHash execution

compojoom:
## Summary

Resolves [WA-2032](https://linear.app/safe-global/issue/WA-2032). In nested-Safe setups a child transaction is confirmed/executed when a parent Safe executes `approveHash(childSafeTxHash)` on the child Safe. The tx-service emits `EXECUTED_MULTISIG_TRANSACTION` for the parent's `approveHash` tx, but CGW only invalidated the parent caches — leaving the child tx and the child Safe's queue stale until TTL (the "tx disappears / wrong state after signing" bug, reproduced on production).

## Changes

- Decode the executed tx's calldata with `SafeDecoder`; when it's an `approveHash` (directly or wrapped in a `multiSend`), also clear the child tx (`clearMultisigTransaction`) and the child Safe's list (`clearMultisigTransactions`). Child Safe = the `approveHash` call's `to`.
- Wire `SafeDecoder` + `MultiSendDecoder` into `EventCacheHelperModule`.
- Integration tests for direct `approveHash`, `multiSend`-wrapped `approveHash`, and a non-`approveHash` no-op.

### DIFF
--- a/src/modules/hooks/domain/helpers/event-cache.helper.module.ts
+++ b/src/modules/hooks/domain/helpers/event-cache.helper.module.ts
@@ -4,6 +4,8 @@ import { BalancesModule } from '@/modules/balances/balances.module';
 import { BlockchainModule } from '@/modules/blockchain/blockchain.module';
 import { ChainsModule } from '@/modules/chains/chains.module';
 import { CollectiblesModule } from '@/modules/collectibles/collectibles.module';
+import { MultiSendDecoder } from '@/modules/contracts/domain/decoders/multi-send-decoder.helper';
+import { SafeDecoder } from '@/modules/contracts/domain/decoders/safe-decoder.helper';
 import { DelegatesV2RepositoryModule } from '@/modules/delegate/domain/v2/delegates.v2.repository.interface';
 import { EarnModule } from '@/modules/earn/earn.module';
 import { EventCacheHelper } from '@/modules/hooks/domain/helpers/event-cache.helper';
@@ -27,7 +29,7 @@ import { TransactionsModule } from '@/modules/transactions/transactions.module';
     StakingModule,
     TransactionsModule,
   ],
-  providers: [EventCacheHelper],
+  providers: [EventCacheHelper, SafeDecoder, MultiSendDecoder],
   exports: [EventCacheHelper],
 })
 export class EventCacheHelperModule {}
--- a/src/modules/hooks/domain/helpers/event-cache.helper.ts
+++ b/src/modules/hooks/domain/helpers/event-cache.helper.ts
@@ -333,11 +333,10 @@ export class EventCacheHelper {
       }),
     ];
 
-    // Nested Safe: if this executed tx approves a child Safe transaction
-    // (approveHash, optionally wrapped in multiSend), the tx-service only signals
-    // a change to the parent tx. Decode the calldata and also invalidate the
-    // referenced child transaction(s) and their Safe's queue, otherwise the child
-    // keeps serving stale state until its cache expires. See WA-2032.
+    // Nested Safe: cascade invalidation to children referenced by approveHash
+    // calls in the executed calldata. We invalidate regardless of event.failed
+    // (a failed approveHash didn't change child state, but re-fetching is cheap
+    // and always-correct). See WA-2032 and getApprovedNestedTransactions.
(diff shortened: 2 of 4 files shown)


compojoom:
@claude review

LucieFaire:
Hey, regarding the point 1 from Claude's review, maybe [this method](https://github.com/safe-global/safe-client-gateway/blob/feat/oidc-require-verified-email/src/modules/safe-shield/utils/transaction-mapping.utils.ts#L65) can help there? We do recursive check for safe shield

compojoom:
Addressed in `76fcda26`

**1. Nested `multiSend` inside `multiSend`** — refactored detection into a recursive `collectApprovedHashes(data, callee)` walker, following the same recursive shape as @LucieFaire's `mapDecodedTransactions` in `safe-shield/utils/transaction-mapping.utils.ts`. Difference: that helper traverses the already-decoded JSON tree (`dataDecoded`); the event only gives us raw `Hex` calldata, so we walk via `MultiSendDecoder.mapMultiSendTransactions` + `SafeDecoder` instead. New test asserts `multiSend([multiSend([approveHash(...)])])` clears the deep child.

**2. `execTransaction`-on-child** — we tested it and it worked fine

**3. `failed` flag** — intentional. Added a one-liner at the call site: invalidating on a failed `approveHash` is harmless (just causes a re-fetch) and always-correct.

**4. Test coverage gaps** — added three tests:
- `does not clear nested children for events with no data` (the `event.data === undefined` early return)
- `clears only the approveHash children when multiSend mixes approveHash with other calls` — asserts the approveHash child *is* cleared **and** that an unrelated Safe queue keyed off a non-approveHash inner call's `to` is **not** cleared (closes the negative-assertion gap too)
- `recurses into nested multiSend batches` (covers the new recursion)

**5. Selector-only `isApproveHash`** — intentional; added a comment on `decodeApprovedHash` making the `decodeFunctionData` try/catch explicit as the trust boundary for selector collisions / truncated calldata.

**Nits**
- `decodeApprovedHash` name: kept — "Hash" mirrors the Safe contract's `approveHash` method name and `hashToApprove` parameter; the JSDoc says it returns the child `safeTxHash`.
- Redundant comment block: slimmed the inline block to one line; the JSDoc on `getApprovedNestedTransactions` carries the long-form rationale.

PooyaRaki (dismissed):

PooyaRaki:
@claude review

compojoom:
Addressed #1 (recursion depth cap) in `ac7ea423`. Treating #2–#4 as nits — reasoning inline below.

**1. Unbounded recursion** — fixed. Added `MAX_APPROVED_HASH_RECURSION_DEPTH = 16` and a `depth` arg threaded through `collectApprovedHashes`. Legitimate nesting is ≤2 (relayer wrapping a wallet batch), so 16 is generous headroom. New test builds 20 levels of `multiSend` wrapping an `approveHash` and asserts the walker bails (child cache stays cached) rather than recursing through.

**2. `Hex` → `Address` widening** — `MultiSendDecoder.mapMultiSendTransactions` runs `getAddress` internally before returning each inner tx, so the value is always a checksummed Address at runtime; only the static type is wider than ideal. Skipping rather than papering over with a redundant `getAddress` at the boundary.

**3. Duplicate-child dedup** — two identical `approveHash` calls in one batch ⇒ two `unlink`s on the same key. Redundant but idempotent and harmless; not worth the `Set` ceremony.

**4. Malformed multiSend log volume** — `MultiSendDecoder` already swallows + warn-logs internally; we'd see the same warn whether or not we cascade. Happy to revisit if production telemetry shows noise.

**Side note on the recursion bound** — the recursive shape was adapted from @LucieFaire's `mapDecodedTransactions` in `safe-shield/utils/transaction-mapping.utils.ts`, which has the same unbounded recursion (on both `multiSend` and `execTransaction`). Worth a follow-up there since the input vector is structurally the same; cc @LucieFaire.

compojoom:
about "Unbounded recursion" - I actually decided to revert my fix. The transaction can't have an unlimited nested multisend - at the end the limit is the max size of the transaction/block. So I don't think this check is necessary.

PooyaRaki (approved):

PR #3115, opened 2026-05-29
feat: add expiration time to space invitations

LucieFaire:
## Summary [WA-2310](https://linear.app/safe-global/issue/WA-2310/be-add-expiry-to-invitations)
Adds a TTL to space member invitations so pending invites automatically become invalid after a configurable period (default **7 days**). Expired invites can no longer be accepted, declined, or used to gain access to a space, and are filtered out of membership/authorization lookups.

### Changes

**Schema & migration**
- New nullable `invite_expires_at` (`timestamp with time zone`) column on `members` (`migrations/1779000000000-add-member-invite-expires-at.ts`).
- Migration backfills existing outstanding invites (`status = 0` / `INVITED`) with a fresh 7-day TTL from migration time.

**Config**
- New `spaces.invite.ttlMs` setting via `SPACES_INVITE_TTL_MS` env var, defaulting to 7 days (`configuration.ts`, schema, and `.env.sample.json`).

**Domain / entities**
- `inviteExpiresAt: Date | null` added to the `Member` domain entity + Zod schema, the DB entity, and the `MemberDto` API response.

**Invite lifecycle** (`members.repository.ts`)
- `inviteUsers` now stamps each invite with `inviteExpiresAt = now + ttl` (`members.service.ts`).
- Inviting was refactored into `insertOrUpdateInvite`: a new invitee is inserted; an existing `INVITED` member is overwritten (name/role/invitedBy/expiry refreshed); an `ACTIVE` member still throws the duplicate-invite error.
- `acceptInvite` / `declineInvite` now reject expired invites with a **`410 Gone`** (`GoneException`) and clear `inviteExpiresAt` on accept/decline.

**Authorization filtering**
- All "is this user a member?" lookups now treat an `INVITED` row as valid only while `inviteExpiresAt > now`, while `ACTIVE` rows are always valid. Applied in `getAuthorizedMembership`, `assertMember`, `SpacesService.getSpaces`, and address-book access. These move from a single `status IN ('ACTIVE','INVITED')` query to an OR of two where-clauses using `MoreThan(new Date())`.

### Tests
- Repository unit + integration specs cover insert/overwrite of invites, expiry stamping, and `410` on accepting/declining expired invites.
- Service, controller e2e, and assert-util specs updated for the new field and expiry filtering.


LucieFaire:
@claude review

LucieFaire:
@claude review again

src/modules/users/domain/members.repository.integration.spec.ts
```diff
@@ -750,6 +760,7 @@ describe('MembersRepository', () => {
     it('should not create PENDING users for existing ones', async () => {
       const spaceName = nameBuilder();
       const adminName = nameBuilder();
+      const inviteExpiresAt = new Date(Date.now() + 60 * 60 * 1000);
```
vseehausen: nit: could look nicer with a little helper like

```
const inviteExpiresAt = {
  future: () => new Date(Date.now() + 60 * 60 * 1000),
  expired: () => new Date(Date.now() - 60 * 60 * 1000),
};
```

and `const inviteExpiresAt = inviteExpiry.future();`

Not a blocker.

vseehausen (dismissed):

src/modules/users/domain/members.repository.integration.spec.ts
LucieFaire: replaced it with faker, [73d2575](https://github.com/safe-global/safe-client-gateway/pull/3115/commits/73d257576a6c86b4b25d1327d6a14daf4520097f)

vseehausen (approved):

PR #3117, opened 2026-06-02
feat: UUID as spaceId instead of numeric value

vmosharova:
## Summary

Resolves: [WA-1583](https://linear.app/safe-global/issue/WA-1583/replace-sequential-spaceid-with-global-uuid)

Migrates the public Spaces API from sequential numeric primary keys to UUIDs as the externally-visible Space identifier, while keeping a short-lived numeric-ID fallback on all Space routes so the frontend can roll out the change without a hard cutover.

**Why:**

- Numeric, sequential IDs leak space-creation volume and let clients enumerate other tenants' spaces. UUIDs remove both.
- A UUID-based contract is a prerequisite for upcoming sharing/invite-link work, where the identifier ends up in URLs and emails.

**Approach:**

- The internal numeric PK (`spaces.id`) is unchanged — all foreign keys, joins, and indexes keep working as-is. A new `uuid` column is added alongside it and becomes the identifier the API exposes.
- Resolution from the public UUID to the internal `id` happens at the edge, in NestJS pipes, so services and repositories continue to operate on `number` IDs and remain unaware of the encoding.
- During the deprecation window, routes accept **either** a UUID or the legacy numeric ID (`LegacySpaceIdPipe`) so cached FE links and not-yet-migrated FE calls keep working. A strict UUID-only pipe (`SpaceIdPipe`) is implemented and wired into the module, ready to replace the legacy pipe in a follow-up PR once the FE sends UUIDs everywhere. The legacy path is TODO-marked for removal.

## Changes

**Database**

- New migration `migrations/1779200000001-add-space-uuid.ts`: adds `uuid UUID NOT NULL DEFAULT gen_random_uuid()` to `spaces` plus a unique index (`UQ_spaces_uuid`). Existing rows are backfilled server-side via the column default. Reversible (`down` drops the index and column).
- `Space` DB entity (`spaces/datasources/entities/space.entity.db.ts`) exposes the new `uuid` column as immutable (`update: false`, `unique: true`).
- Domain `Space` entity/schema (`spaces/domain/entities/space.entity.ts`) adds `uuid` validated by `UuidSchema`.

**Pipes (new) — `src/modules/spaces/routes/pipes/space-id.pipe.ts`**

- `SpaceIdPipe`: validates input as a UUID (regex), then resolves to the numeric PK via `ISpacesRepository.findIdByUuid`. Wired into `SpacesModule` and ready for the follow-up cutover; not yet attached to any route.
- `LegacySpaceIdPipe`: accepts a UUID **or** a base-10 numeric string bounded by `DB_MAX_SAFE_INTEGER`; delegates to `findIdByIdOrUuid`. Currently used on every Space route. Both pipes return `Space['id']` (`number`), so downstream services are unchanged, and both reject malformed input with a single shared 400 body (`Invalid space identifier`) regardless of which pipe validates.

**Repository / domain**

- `ISpacesRepository` gains `findIdByUuid(uuid)` and the temporary `findIdByIdOrUuid(value)` (TODO-marked for removal). The legacy method branches numeric-vs-UUID and looks up the row, relying on the pipe to have validated input format upstream.
- `SpacesRepository.create` and `update` now return `uuid` alongside `id` (`update` uses a `RETURNING` clause).
- Error wording standardized: `update()` now throws `Workspace not found.` (was `Space not found.`) to match the rest of the repository.

**Controllers** — pipes wired into route params; no behavioural change beyond identifier handling. Every route currently uses `LegacySpaceIdPipe`:

- `spaces` (`spaces.controller.ts`): `GET /:id`, `PATCH /:id`, `DELETE /:id`.
- `members` (`members.controller.ts`): invite / accept / decline / `GET members` / `GET membership` / role / alias / remove / leave.
- `space-safes`, `address-books`, `address-book-requests`, `user-address-book`, `surveys`, and `space-counterfactual-safes`: all Space-scoped routes.

**DTOs** — responses now expose `uuid`:

- `GetSpaceResponse`, `CreateSpaceResponse`, `UpdateSpaceResponse`, and `Invitation` add a `uuid` string field. The numeric `id` is retained but marked `deprecated` ("use uuid; kept for FE fallback").
- Swagger `@ApiParam` examples updated to UUIDs; legacy-accepting routes call out numeric acceptance as deprecated in their descriptions.

**Tests**

- New unit specs for both pipes (`space-id.pipe.spec.ts`): UUID resolution (case-insensitive), numeric resolution and `DB_MAX_SAFE_INTEGER` boundary, malformed/empty/over-max rejection with the shared 400 message, and `NotFoundException` propagation.
- e2e specs (spaces, members, space-safes, address-books, surveys) updated to send/expect UUIDs, with new cases for malformed identifiers returning 400.
- Integration and unit specs for `SpacesRepository`, `SpacesService`, and `SurveysService` aligned with the new shape (and the corrected `Workspace not found.` message).

## Notes (rollout / reviewers)

- The migration is additive and online-safe: the column default fills existing rows server-side, no app-level backfill.
- **Current state vs. final state:** unlike the eventual design, *all* routes — reads and writes — currently accept numeric IDs via `LegacySpaceIdPipe`. This is intentional for the FE-compatibility window; numeric IDs stay enumerable until the follow-up PR.
- **Follow-up PR (FE-coordinated):** once the FE sends UUIDs everywhere, swap `LegacySpaceIdPipe` → `SpaceIdPipe` (starting with mutating routes), then drop `findIdByIdOrUuid`, `LegacySpaceIdPipe`, and the deprecated numeric `id` response fields. All legacy paths are TODO-marked to make the cleanup obvious.


vmosharova:
@claude review

vmosharova:
> acceptInvite / declineInvite are UUID-only ...  if a previously emailed invite link uses a numeric space id, the user clicking it will hit a 400

There is no real invitations, so this review comment is redundant. 

> DROP INDEX IF EXISTS / DROP COLUMN IF EXISTS makes the rollback safer

This will go against the consistent repo convention - the UUID will be the only migration doing such approach.


The rest is addressed in [457a7bb](https://github.com/safe-global/safe-client-gateway/pull/3117/commits/457a7bb17fb38ae92a90f469a3ae6eeb63061a0c), [654830a](https://github.com/safe-global/safe-client-gateway/pull/3117/commits/654830aa6c066d368053f6aad4ad7b24ddf1b712) and [a123c74](https://github.com/safe-global/safe-client-gateway/pull/3117/commits/a123c74ce9b28d34520d90318f04d34bc81a467f).

vmosharova:
@claude review new changes

src/modules/spaces/domain/entities/space.entity.ts
```diff
@@ -17,12 +17,14 @@ export type Space = z.infer<typeof SpaceSchema>;
 // We need explicitly define ZodType due to recursion
 export const SpaceSchema: z.ZodType<
   z.infer<typeof RowSchema> & {
+    uuid: string;
```
PooyaRaki:
```suggestion
import { UUID } from 'crypto';

uuid: UUID;
```

src/modules/spaces/domain/spaces.repository.ts
```diff
@@ -197,13 +197,46 @@ export class SpacesRepository implements ISpacesRepository {
...
+    const result = await spaceRepository
+      .createQueryBuilder()
+      .update(Space)
+      .set(args.updatePayload)
+      .where({ id: args.id })
+      .returning(['id', 'uuid'])
+      .execute();
 
-    return { id: args.id };
+    const row = result.raw[0] as Pick<Space, 'id' | 'uuid'> | undefined;
```
PooyaRaki, line 218:
```suggestion
  const row = result.generatedMaps[0] as Pick<Space, 'id' | 'uuid'> | undefined;
```

vmosharova: Done

src/modules/spaces/domain/entities/space.entity.ts
vmosharova: Updated

src/modules/spaces/domain/spaces.repository.ts
vmosharova: Update: I changed it back to `result.row[0]` [here](https://github.com/safe-global/safe-client-gateway/pull/3117/changes/d87b369653c98a185a86f409d956baeac36af17a). The reason is that there is an `UPDATE` in spaces repository, and TypeOrm doesn't fill `result.generatedMaps[0]` for an `UPDATE` - it was always undefined (hence failing tests).

For the context, the builder:
```
    const result = await spaceRepository
      .createQueryBuilder()
      .update(Space)
      .set(args.updatePayload)
      .where({ id: args.id })
      .returning(['id', 'uuid'])
      .execute();
      ```

**Comment by vmosharova** [comment]
@claude review

**Comment by vmosharova** [comment]
Addressed Claude review comments here: [a598e5d](https://github.com/safe-global/safe-client-gateway/pull/3117/commits/a598e5d1bc4f1a46e5de5155b6a1c556f7c8c291)

File: src/modules/spaces/routes/spaces.service.ts [hunk 3]
```diff
@@ -127,7 +130,8 @@ export class SpacesService {
     const userId = getAuthenticatedUserIdOrFail(args.authPayload);
     await assertAdmin(this.spacesRepository, args.id, userId);
 
-    return await this.spacesRepository.update(args);
+    const result = await this.spacesRepository.update(args);
+    return { id: result.id, uuid: result.uuid };
```
PooyaRaki:
```suggestion
    return await this.spacesRepository.update(args);
```

src/modules/spaces/routes/spaces.service.ts
```diff
@@ -40,7 +40,8 @@ export class SpacesService {
     await this.usersRepository.findOneOrFail({ id: userId });
     await this.usersRepository.activateIfPending(userId);
 
-    return await this.spacesRepository.create({ userId, ...args });
+    const result = await this.spacesRepository.create({ userId, ...args });
+    return { id: result.id, uuid: result.uuid, name: result.name };
```
PooyaRaki:
```suggestion
    return await this.spacesRepository.create({ userId, ...args });
```

src/modules/spaces/routes/spaces.service.spec.ts
```diff
@@ -134,6 +137,24 @@ describe('SpacesService', () => {
       });
     });
 
+    it.each([
+      ['SIWE', siweAuthPayloadDtoBuilder] as const,
+      ['OIDC', oidcAuthPayloadDtoBuilder] as const,
```
PooyaRaki:
```suggestion
      ['SIWE', siweAuthPayloadDtoBuilder],
      ['OIDC', oidcAuthPayloadDtoBuilder],
```

src/modules/spaces/domain/spaces.repository.ts
```diff
@@ -27,7 +28,9 @@ import {
   MemberStatus,
 } from '@/modules/users/domain/entities/member.entity';
 
-// TODO: Add tests
+const UUID_REGEX =
```
PooyaRaki: The same regex is repeated in the pipe, let's extract it and reuse it.

src/modules/spaces/domain/spaces.repository.ts
```diff
@@ -201,13 +205,49 @@ export class SpacesRepository implements ISpacesRepository {
...
+    });
+    return space.id;
+  }
 
-    return { id: args.id };
+  // TODO: remove after FE removes numeric Space ID fallback.
+  // UUID is re-validated here (defence in depth) so a malformed string can't
+  // reach Postgres as a UUID and surface as a raw 500 instead of a 400.
+  public async findIdByIdOrUuid(value: string): Promise<Space['id']> {
+    if (/^\d+$/.test(value)) {
```
PooyaRaki: This is ok since the method is temp but in future we should keep such validations in the controller.

src/modules/spaces/domain/spaces.repository.integration.spec.ts
```diff
@@ -450,10 +458,13 @@ describe('SpacesRepository', () => {
         spacesRepository.findOneOrFail({ where: { id: space.id } }),
       ).resolves.toEqual({
         id: space.id,
+        uuid: space.uuid,
         name,
         status: spaceStatus,
         createdAt: expect.any(Date),
         updatedAt: expect.any(Date),
+        members: undefined,
+        safes: undefined,
```
LucieFaire: how come these got included?

src/modules/spaces/domain/spaces.repository.ts
```diff
@@ -201,13 +203,59 @@ export class SpacesRepository implements ISpacesRepository {
...
+      const space = await this.findOneOrFail({
+        where: { id: Number(value) },
+        select: { id: true },
+      });
+      return space.id;
+    }
+    if (!UUID_REGEX.test(value)) {
```
LucieFaire, line 255: wonder if we can use UuidSchema instead? Not critical, just wonder if we can avoid having this regex

src/modules/spaces/domain/spaces.repository.ts
```diff
@@ -201,13 +203,59 @@ export class SpacesRepository implements ISpacesRepository {
...
+      select: { uuid: true },
+    });
+    return space.uuid;
+  }
+
+  // TODO: remove after FE removes numeric Space ID fallback.
+  // UUID is re-validated here (defence in depth) so a malformed string can't
+  // reach Postgres as a UUID and surface as a raw 500 instead of a 400.
+  public async findIdByIdOrUuid(value: string): Promise<Space['id']> {
+    if (/^\d+$/.test(value)) {
```
LucieFaire: nit: `NUMERIC_REGEX` ?

src/modules/spaces/routes/members.service.spec.ts
```diff
@@ -251,9 +253,11 @@ describe('MembersService', () => {
         .with('role', 'ADMIN')
         .with('status', 'ACTIVE')
         .build();
+      const spaceUuid = faker.string.uuid() as UUID;
```
LucieFaire: nit: maybe its easier to create a const fakeUuid() as we have for fakeEmailAddress so you dont have to cast each time

src/modules/spaces/routes/spaces.service.spec.ts
```diff
@@ -135,8 +138,26 @@ describe('SpacesService', () => {
...
+      ['OIDC', oidcAuthPayloadDtoBuilder],
+    ])('should not scope the membership query to any space for %s user', async (_label, builder) => {
+      const authPayload = new AuthPayload(builder().build());
+
+      membersRepositoryMock.find.mockResolvedValue([]);
+
+      await service.getActiveOrInvitedSpaces(authPayload);
+
+      const options = membersRepositoryMock.find.mock.calls[0][0];
+      const where = options?.where as Array<Record<string, unknown>>;
```
LucieFaire: could we add a comment what exactly is happening here? the test name is very smart, im not sure whats happening :)

LucieFaire (commented):
small remarks, otherwise looks good. My only concern is additional fields in the tests

src/modules/spaces/domain/spaces.repository.integration.spec.ts
vmosharova: You are right, these additions were redundant and will be removed.

src/modules/spaces/domain/spaces.repository.ts
vmosharova: I don't think this update is needed at all - we'll remove this chunk of code in the follow-up cleanup PR, there is also a reminder comment about that.

src/modules/spaces/routes/spaces.service.spec.ts
vmosharova: Done

src/modules/spaces/routes/members.service.spec.ts
vmosharova: Done :)

src/modules/spaces/domain/spaces.repository.ts
vmosharova: Done

LucieFaire (approved):

katspaugh (approved):

PR #3118, opened 2026-06-03
chore: Update http code for create signed msg endpoint

TimDaub:
This PR documents that CGW can return an empty response in `GET /v1/chains/{chainId}/messages/{messageHash}` after `POST /chains/:chainId/safes/:safeAddress/messages`

When using a WalletConnect-connected dapp, this can lead to a race condition in the frontend. The frontend didn't consider that the `200 OK` that CGW returns in `POST /chains/:chainId/safes/:safeAddress/messages` means CGW still has to wait on TxS to find the signed message. `GET /v1/chains/{chainId}/messages/{messageHash}` can return an empty result.

Hence I would suggest to return a `202 Accepted` for `POST /chains/:chainId/safes/:safeAddress/messages` in the future.

Below is a video of the bug on safe-wallet-monorepo (sound on to hear my voice over)

https://github.com/user-attachments/assets/377ca99c-b5be-48ba-ac7b-4ce5b99a3cf2

Anyways, no hard requirement to switch to `202 Accepted`. I will implement polling on the safe-wallet-monorepo independently of whether we change to 202 here or not. Up to you :D

Tracked in ticket [WA-2476](https://linear.app/safe-global/issue/WA-2476/off-chain-message-signing-shows-false-error-confirming-the-message-due)


LucieFaire (commented):
is it only a new test or the endpoint changed as well? See only 1 file changed, not sure where `202` is set

TimDaub:
> is it only a new test or the endpoint changed as well? See only 1 file changed, not sure where `202` is set

So far it's only a test to create awareness because I wanted to get CGW owners' opinion if changing the response status to 202 is a good idea. I'm also happy to close the PR if you decide that 200 OK is sufficient here

PooyaRaki:
👍

TimDaub:
I implemented the behavior and all tests are now passing

LucieFaire (approved):

PR #3130, opened 2026-06-03
feat(space): validation

PooyaRaki:
## Summary

Centralizes validation across the Spaces module by deriving field rules from shared Zod schemas instead of ad-hoc inline validators (`z.string()`, `z.number().int().nullable()`, etc.). This keeps name, member-field, and chain-id validation defined in one place so the rules stay consistent across domain entities and route DTOs.

  ## ⚠️  Behavioral change — invite name validation

  This is **not purely a refactor**. Invite names (`accept-invite`, `invite-users`) now go through `makeNameSchema` (min 3 chars + restricted charset `^[a-zA-Z0-9]+(?:[
  ._-][a-zA-Z0-9]+)*$`), which **reverses** the relaxation from #3128 (`fix(spaces): relax invite name validation to max 255 chars`). Inputs that #3128 specifically allowed are now
  rejected, e.g.:
  - special characters such as `<>@!`
  - 1–2 character names

  Downstream consumers that relied on the relaxed rule will start getting validation errors. Since #3128 merged the same day, this relax-then-tighten round trip should be confirmed
  with its author in case of competing requirements.

  ## Changes

  **Name & member-field validation from shared schemas**
  - Export `AddressBookItemSchema` and source DTO `name` fields from its `.shape`.
  - Apply `makeNameSchema({ maxLength: 255 })` to `accept-invite` and `invite-users` name fields.
  - Derive create/update space `name` from `SpaceSchema.shape.name`.
  - Derive address book request `reviewedBy` from `MemberSchema.shape.invitedBy`.
  - Refactor `SpaceSchema` and `MemberSchema` into `ZodObject`s (exposing `.shape`) using explicitly typed lazy callbacks to break the circular type inference.

  **Address book chainIds validation**
  - Tighten `chainIds` from `z.array(z.string())` to `z.array(ChainIdSchema)` in the address book item entity, user address book item entity, and the upsert DTO.
  - Require the upsert address book item `name` (drop `.optional()`).

  ## Tests
  - Updated `invite-users` spec to expect special-character names to be rejected.
  - Added `upsert-address-book-items.dto.entity.spec.ts` covering required-name, name format, numeric `chainId`, and address validation.

### DIFF
--- a/src/modules/spaces/routes/entities/accept-invite.dto.entity.ts
+++ b/src/modules/spaces/routes/entities/accept-invite.dto.entity.ts
@@ -1,13 +1,16 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { ApiProperty } from '@nestjs/swagger';
 import { z } from 'zod';
-import { makeNameSchema } from '@/domain/common/schemas/name.schema';
+import {
+  makeNameSchema,
+  NAME_MAX_LENGTH,
+} from '@/domain/common/schemas/name.schema';
 
 export const AcceptInviteDtoSchema = z.object({
-  name: makeNameSchema({ maxLength: 255 }),
+  name: makeNameSchema({ maxLength: NAME_MAX_LENGTH }),
 });
 
 export class AcceptInviteDto implements z.infer<typeof AcceptInviteDtoSchema> {
-  @ApiProperty({ type: String, maxLength: 255 })
+  @ApiProperty({ type: String, maxLength: NAME_MAX_LENGTH })
   public readonly name!: string;
 }
--- a/src/modules/spaces/routes/entities/accept-invite.dto.entity.ts
+++ b/src/modules/spaces/routes/entities/accept-invite.dto.entity.ts
@@ -1,16 +1,14 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { ApiProperty } from '@nestjs/swagger';
 import { z } from 'zod';
-import {
-  makeNameSchema,
-  NAME_MAX_LENGTH,
-} from '@/domain/common/schemas/name.schema';
+import { makeNameSchema } from '@/domain/common/schemas/name.schema';
+import { MEMBER_NAME_MAX_LENGTH } from '@/modules/users/domain/entities/member.entity';
 
 export const AcceptInviteDtoSchema = z.object({
-  name: makeNameSchema({ maxLength: NAME_MAX_LENGTH }),
+  name: makeNameSchema({ maxLength: MEMBER_NAME_MAX_LENGTH }),
 });
 
 export class AcceptInviteDto implements z.infer<typeof AcceptInviteDtoSchema> {
-  @ApiProperty({ type: String, maxLength: NAME_MAX_LENGTH })
+  @ApiProperty({ type: String, maxLength: MEMBER_NAME_MAX_LENGTH })
   public readonly name!: string;
 }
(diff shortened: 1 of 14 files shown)


vmosharova:
@claude review

PooyaRaki:
@claude review

🔴 Re-tightens invite name validation that was just relaxed in https://github.com/safe-global/safe-client-gateway/pull/3128 -> Intentional ❌ 
🟠 UpsertAddressBookItemsDto.name silently became optional -> It was a mistake, Fixed ✅ 
🟠 UserAddressBookItemSchema.chainIds missed the ChainIdSchema tightening -> Fixed ✅ 
🟡 reviewedBy now requires positive integers -> Intentional ❌ 
🟡 Minor -> Not Fixed ❌

PooyaRaki:
@claude review

vmosharova (dismissed):
👍🏻

tmjssz (dismissed):

src/modules/spaces/routes/entities/accept-invite.dto.entity.ts
```diff
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { ApiProperty } from '@nestjs/swagger';
 import { z } from 'zod';
+import { makeNameSchema } from '@/domain/common/schemas/name.schema';
 
 export const AcceptInviteDtoSchema = z.object({
-  name: z.string().max(255),
+  name: makeNameSchema({ maxLength: 255 }),
```
vseehausen: Let's define a variable for the maxLength, like in `src/modules/spaces/domain/address-books/entities/address-book-item.entity.ts`.

src/modules/spaces/routes/entities/invite-users.dto.entity.ts
```diff
@@ -17,7 +18,7 @@ export const InviteType = {
 
 const SharedInviteFields = {
   role: z.enum(getStringEnumKeys(MemberRole)),
-  name: z.string().max(255),
+  name: makeNameSchema({ maxLength: 255 }),
```
vseehausen: And use the variable here, too.

vseehausen (changes requested):

src/modules/spaces/routes/entities/accept-invite.dto.entity.ts
PooyaRaki: Done ✅ 
https://github.com/safe-global/safe-client-gateway/pull/3130/commits/5f74ff33dcc2f2ffdfeb6f47a813a0561c1e96f6

src/modules/spaces/routes/entities/invite-users.dto.entity.ts
PooyaRaki: Done ✅ 
https://github.com/safe-global/safe-client-gateway/pull/3130/commits/5f74ff33dcc2f2ffdfeb6f47a813a0561c1e96f6

vseehausen: We have to add a new variable for invite name max length. When we use the 30 charactar limit, it will fail on the onboarding flow, as the address is set as the name.

PooyaRaki: But the member name in the database is already 30 chars.
https://github.com/safe-global/safe-client-gateway/blob/5f74ff33dcc2f2ffdfeb6f47a813a0561c1e96f6/src/modules/users/datasources/entities/member.entity.db.ts#L59

PooyaRaki: I reverted the change:
https://github.com/safe-global/safe-client-gateway/pull/3130/commits/f17710b0d721139cf7f26ca65c2b5c287ae1875b

The limit was set to 30 in the entity code but unfortunately there was not a migration to actually change the limit to 30 chars so I reverted it for now.

vseehausen (approved):

PR #3131, opened 2026-06-03
fix(spaces): expose invited members' email to space admins

vseehausen:
## Summary

- `GET /v1/spaces/:spaceId/members` now returns the stored email of `INVITED` members when the caller is an active admin of the space; previously emails were only exposed once a member accepted the invite.
- Non-admin callers (including invited admins) keep seeing `email: null` for invited members; active members' emails are unchanged.
- Admin status is checked via the existing `MembersRepository.findActiveAdmin`, so the role/status predicate is not duplicated in the service.
- Unit tests cover both caller roles; e2e tests use email-type invites (the case where an invited user actually has a stored email) for the admin and non-admin caller.

### DIFF
--- a/src/modules/spaces/routes/members.service.ts
+++ b/src/modules/spaces/routes/members.service.ts
@@ -79,18 +79,27 @@ export class MembersService {
     authPayload: AuthPayload;
     spaceId: Space['id'];
   }): Promise<MembersDto> {
+    const members = await this.membersRepository.findAuthorizedMembersOrFail({
+      authPayload: args.authPayload,
+      spaceId: args.spaceId,
+    });
+    const isActiveAdmin = Boolean(
+      await this.membersRepository.findActiveAdmin({
+        userId: getAuthenticatedUserIdOrFail(args.authPayload),
+        spaceId: args.spaceId,
+      }),
+    );
     return {
-      members: (
-        await this.membersRepository.findAuthorizedMembersOrFail({
-          authPayload: args.authPayload,
-          spaceId: args.spaceId,
-        })
-      ).map((member) => ({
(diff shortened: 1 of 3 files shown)


src/modules/spaces/routes/members.service.ts
```diff
@@ -79,18 +79,27 @@ export class MembersService {
...
+    const members = await this.membersRepository.findAuthorizedMembersOrFail({
+      authPayload: args.authPayload,
+      spaceId: args.spaceId,
+    });
+    const isActiveAdmin = Boolean(
+      await this.membersRepository.findActiveAdmin({
+        userId: getAuthenticatedUserIdOrFail(args.authPayload),
+        spaceId: args.spaceId,
+      }),
+    );
```
LucieFaire: nit: we can parallelize it:
```
   const [members, activeAdmin] = await Promise.all([
      this.membersRepository.findAuthorizedMembersOrFail({
        spaceId: args.spaceId,
      }),
      this.membersRepository.findActiveAdmin({
        userId: getAuthenticatedUserIdOrFail(args.authPayload),
        spaceId: args.spaceId,
      }),
    ]);
    const isActiveAdmin = Boolean(activeAdmin);
```

LucieFaire (commented):
one remark otherwise good to go

src/modules/spaces/routes/members.service.ts
vseehausen: Always optimizing... :D

vseehausen: addressed in c816f03a2

LucieFaire (approved):

PR #3134, opened 2026-06-04
feat: renew invites endpoint

LucieFaire:
## WA-2313 — BE: Invitation can be resent

Linear: https://linear.app/safe-global/issue/WA-2313/be-invitation-can-be-resent

### What & why
Adds an admin-only endpoint to renew a pending space-member invitation, refreshing its expiry. There is no "invitation id" in the model (invitations are member rows keyed by `(user, space)`), and reusing the invite endpoint would force the frontend to reconstruct a wallet payload from data it can't reliably reconstruct (the members list doesn't expose wallet addresses, and a user can have multiple wallets). So renewal is done server-side by `userId`: the backend resets the existing invite itself — no payload, no wallet-address exposure.

### Endpoint
`POST /v1/spaces/:spaceId/members/:userId/invite/renew`
- Admin-only (reuses the existing active-admin check).
- `INVITED` (expired or not) → refreshes `inviteExpiresAt`; status stays `INVITED`; `invitedBy`/`name`/`role` preserved.
- `DECLINED` / `ACTIVE` → 409 (not renewable).
- member not found → 404; non-admin → 403.

### Changes
- src/modules/spaces/routes/members.controller.ts — new `renewInvite` route (guards + Swagger), mirrors accept/decline/role routes.
- src/modules/spaces/routes/members.service.ts — `renewInvite`: active-admin check + computes `now + inviteTtlMs`, delegates to repo.
- src/modules/users/domain/members.repository.ts — `renewInvite`: loads member by `(userId, spaceId)`, status guard, updates only `inviteExpiresAt`, returns the `Invitation`.
- src/modules/users/domain/members.repository.interface.ts — `renewInvite` signature + doc; clarified `inviteUsers` doc (renew-on-reinvite; active/declined can't be re-invited).
- Tests: repo unit + integration (renew expired-pending preserves metadata; not-found/active/declined rejected) and controller e2e (happy / 403 / 404 / 409).

Also includes the earlier commit hardening the existing invite-renew path with tests (declined not resendable; active rejected).

### Scope notes
- Renewal currently only refreshes the invite's expiry. Re-sending the invite **email** is out of scope — invite-email delivery isn't implemented yet; this endpoint is where that hook will live (follow-up).
- The frontend need here is renew-only (it does not display the invited wallet address), so no wallet address is exposed.

### Testing
- `yarn test src/modules/users/domain/members.repository.spec.ts`
- `yarn test:integration src/modules/users/domain/members.repository.integration.spec.ts`
- `yarn test:e2e src/modules/spaces/routes/members.controller.e2e-spec.ts`
- (integration/e2e require the local Postgres)

### DIFF
--- a/src/modules/users/domain/members.repository.ts
+++ b/src/modules/users/domain/members.repository.ts
@@ -251,6 +251,36 @@ export class MembersRepository implements IMembersRepository {
     throw duplicateInviteError();
   }
 
+  public async renewInvite(args: {
+    spaceId: Space['id'];
+    userId: User['id'];
+    inviteExpiresAt: Date;
+  }): Promise<Invitation> {
+    const member = await this.findOneOrFail({
+      user: { id: args.userId },
+      space: { id: args.spaceId },
+    });
+
+    if (member.status !== 'INVITED') {
+      throw new ConflictException('Only a pending invitation can be renewed.');
+    }
+
+    await this.postgresDatabaseService.transaction(async (entityManager) => {
+      await entityManager.update(DbMember, member.id, {
+        inviteExpiresAt: args.inviteExpiresAt,
+      });
+    });
+
+    return {
+      userId: args.userId,
+      spaceId: args.spaceId,
+      name: member.name,
+      role: member.role,
+      status: member.status,
+      invitedBy: member.invitedBy,
+    };
+  }
+
   public async acceptInvite(args: {
     authPayload: AuthPayload;
     spaceId: Space['id'];
--- a/src/modules/users/domain/members.repository.ts
+++ b/src/modules/users/domain/members.repository.ts
@@ -252,33 +252,14 @@ export class MembersRepository implements IMembersRepository {
   }
 
   public async renewInvite(args: {
-    spaceId: Space['id'];
-    userId: User['id'];
+    memberId: Member['id'];
     inviteExpiresAt: Date;
-  }): Promise<Invitation> {
-    const member = await this.findOneOrFail({
-      user: { id: args.userId },
-      space: { id: args.spaceId },
+  }): Promise<void> {
+    const membersRepository =
+      await this.postgresDatabaseService.getRepository(DbMember);
+    await membersRepository.update(args.memberId, {
+      inviteExpiresAt: args.inviteExpiresAt,
     });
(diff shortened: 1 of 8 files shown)


src/modules/users/domain/members.repository.ts
```diff
@@ -251,6 +251,36 @@ export class MembersRepository implements IMembersRepository {
...
+    const member = await this.findOneOrFail({
+      user: { id: args.userId },
+      space: { id: args.spaceId },
+    });
+
+    if (member.status !== 'INVITED') {
+      throw new ConflictException('Only a pending invitation can be renewed.');
+    }
+
+    await this.postgresDatabaseService.transaction(async (entityManager) => {
```
PooyaRaki: We don't need a database transaction here.

src/modules/users/domain/members.repository.ts
```diff
@@ -251,6 +251,36 @@ export class MembersRepository implements IMembersRepository {
...
+    spaceId: Space['id'];
+    userId: User['id'];
+    inviteExpiresAt: Date;
+  }): Promise<Invitation> {
+    const member = await this.findOneOrFail({
+      user: { id: args.userId },
+      space: { id: args.spaceId },
+    });
+
+    if (member.status !== 'INVITED') {
```
PooyaRaki: Usually this logic checks should stay in the service layer.

LucieFaire: yes, agree, but this service/repo are weird. One does everything while another doesnt. Its on my todo to refactor these post-launch, but i can change this one already :)

LucieFaire: [be5f74c](https://github.com/safe-global/safe-client-gateway/pull/3134/commits/be5f74c26d342bcf82e7514a6e30451fedda49de)

src/modules/users/domain/members.repository.ts
LucieFaire: [be5f74c](https://github.com/safe-global/safe-client-gateway/pull/3134/commits/be5f74c26d342bcf82e7514a6e30451fedda49de)

PooyaRaki (approved):

PR #3136, opened 2026-06-04
feat: replace Jest with Vitest

PooyaRaki:
## Summary

  Replaces the test runner — **Jest 30 + ts-jest → Vitest 4** — across the whole suite. All 5,236 unit tests pass; scripts, config, and CI hooks keep the same names so the workflow is unchanged. NestJS DI is preserved by transforming tests with SWC (decorator metadata), and coverage
  still emits lcov for Coveralls.
  
  ## Changes

  **Tooling / deps**
  - Add `vitest@4.1.7`, `@vitest/coverage-v8@4.1.7`, `unplugin-swc@1.5.9`, `@swc/core@1.15.40`, `vite-tsconfig-paths@6.1.1`.
  - Remove `jest`, `ts-jest`, `@types/jest`.
  
  **Config**
  - New `vitest.config.ts` with three projects — `unit`, `integration`, `e2e` — replacing the four `test/jest-*.json` files and the inline `package.json` jest config.
  - `unplugin-swc` (`legacyDecorator` + `decoratorMetadata`) so NestJS `emitDecoratorMetadata` survives — required for DI in integration/e2e bootstrap.
  - Path aliases (`@/*`, `@/abis/*`) via `vite-tsconfig-paths`; `globals: true`; `TZ=UTC` moved from `global-setup.ts` to `test.env`.
  - Coverage via `@vitest/coverage-v8` → `coverage/` as `lcov` (Coveralls), `text`, `html`. Delete `test/jest-*.json` and `test/global-setup.ts`.
  - `biome.json`: `globals` `["jest"]` → Vitest globals; disable `useArrowFunction` for test files (Vitest needs constructable, non-arrow mocks).
  - `tsconfig.json`: add `vitest/globals` to `types`.
  
  **Code**
  - Codemod across 238 files: `jest.*` → `vi.*`; `jest.MockedObjectDeep`/`Mocked`/`Mock`/`MockedFunction`/`MockedClass`/`SpyInstance` → native Vitest types imported from `vitest`.
  - `package.json` test scripts repointed to `vitest run --project …` (names unchanged: `test`, `test:unit`, `test:integration`, `test:e2e`, `test:all`, + `:cov`/`:watch`/`:debug`).
  - Update the custom `anyStringOrNull` matcher and `TESTING.md`.
  
  **Test adaptations for Vitest's stricter semantics** (please review)
  - Class-mock implementations used with `new` switched from arrow to `function` (Vitest 4 rejects arrows as constructors): `aws-cloud-storage`, `aws-ses`.
  - `viem.getAddress` spy → partial `vi.mock` (ESM named exports aren't spyable): `safe-signature`.
  - Error matchers: Vitest deep-compares `toThrow(errorInstance)` where Jest only checked `.message`. This surfaced **two latent issues Jest masked** — flagging for a maintainer:
    - `transaction-verifier.helper.spec.ts`: tests asserted `422` but `verifyConfirmation` actually throws `502` (via `verifyApiTransaction`). Kept the original `422` intent by asserting the message only; **the 422-vs-502 discrepancy is a real code question, not changed here.**
    - `database-migrator.service.integration.spec.ts`: a test rejected with a *function* instead of an `Error` (Jest passed by stringifying it); fixed to reject with the `Error`.


PooyaRaki:
@claude review

PooyaRaki:
@claude review

PooyaRaki:
@claude review

 Thanks for the thorough re-review. I went through all six against the `main` (Jest) baseline and verified each by running the code. Summary: none are regressions introduced by
  the migration — the suite is green (unit: 304 files, 5266 passed, 0 failed under Vitest 4.1.7).
  
  **1. `resetAllMocks()` on real-fn spies** — Not an issue, and the premise differs for Vitest 4. Unlike Jest, `vi.resetAllMocks()` *restores* the original implementation for
  `vi.spyOn` spies rather than neutering them (verified: spy → `resetAllMocks()` → original value returned, not `undefined`). On top of that, all three cited spots already restore
  explicitly — relay uses `vi.restoreAllMocks()` at the end of each `Date.now` test, blockchain-api uses `fetchSpy.mockRestore()` per test, and safe-signature uses `mockClear()`
  (no `resetAllMocks`, so the wrapped `getAddress` impl is preserved). All were 1:1 `jest.`→`vi.` conversions. No change needed.
  
  **2. Coverage exclude / node_modules** — Equivalent to the prior Jest config. `include` is `src/**` + `scripts/**`, so `node_modules` can't be included. The old 
  `coveragePathIgnorePatterns` scoped builder/factory exclusions to `__tests__/` exactly like the new config, so the same set of files is counted. (Happy to add 
  `...configDefaults.coverage.exclude` as belt-and-suspenders, but it changes nothing measurable.)

  **3. 422 → 502 in transaction-verifier** — Good catch; this is a pre-existing app discrepancy that the migration surfaced (Jest's `toThrow(error)` only compared `.message`, so 
  the 422-vs-502 mismatch was masked). Opening a follow-up issue to align the helper's status code with the test/expectation.
  
  **4. Stricter `toThrow(error)` semantics** — Confirmed real behavior change (Vitest 4 deep-compares enumerable error props, not the class). Already handled across the PR: the 
  only affected assertions are the 3 documented transaction-verifier relaxations plus the `expectedError`→`NetworkResponseError`/`InternalServerErrorException` switches to 
  constructor matching (still asserts the correct error type). Since the full suite passes under the stricter matcher, no further masked failures remain.
  
  **5. `globals: true` × ESM hoisting** — No issue. Only 5 `vi.mock` factories and none reference surrounding-scope vars (`vi.hoisted` not needed); `jest.mock` enforced the same 
  restriction, and all relevant specs pass.
  
  **6. Empty `beforeEach(() => {})`** — This pre-dates the migration (present in `main`), not a codemod artifact.

PooyaRaki:
@claude review

PooyaRaki:
Addressed in `94ce039e`.

  **1. `.vscode/launch.json` still invoked Jest** — Fixed. All three debug configs now run Vitest (`vitest run --project unit` / `--project e2e`) instead of Jest, dropped `--coverage false` and the deleted
  `test/jest-e2e.json` reference. `--runInBand` is replaced by Vitest's equivalent `--no-file-parallelism` (matching the existing `test:debug` script) so breakpoints stay reliable under `--inspect-brk`. You were
  right that it hadn't landed — the previous diff was empty.

  **2. `TESTING.md` test counts** — Fixed by removing the exact counts rather than swapping in "thousands"/"dozens". The unit suite is ~5.2k tests, but the integration suite is ~1,184 test cases across 81 files,
  so "dozens" would also have been inaccurate. Dropped the `Count` bullets and the counts in the CI job descriptions so the doc no longer carries numbers that drift.

  **3. `quickstart.md:189` `jest.Mocked`** — Fixed → `MockedObject<IChainsRepository>`, matching the rest of the migration.
  
  **4. `anyStringOrNull` return type** — Fixed → `anyStringOrNull(): void`, consistent with the `AsymmetricMatchersContaining` declaration. Confirmed no caller uses the return value.
  
  **5. Stale CI comment** — Fixed. Rephrased to reference Vitest transpiling via SWC (strips types without type-checking) instead of ts-jest.
  
  **6. `vi.mock` factory capture** — Fixed → `const { mockSend } = vi.hoisted(() => ({ mockSend: vi.fn() }))`. (Minor clarification: it worked before because Vitest allows `mock`-prefixed variables inside
  factories, not purely by lazy ordering — but `vi.hoisted` is the explicit, robust form.)

  **7. Coverage `exclude` dropping Vitest defaults** — Not applied, and I think the finding doesn't hold on v4. This repo is on Vitest 4.1.7, where both `configDefaults.coverage.exclude` and
  `coverageConfigDefaults.exclude` are `[]` — the default exclude list that existed in v3 was removed. So `...configDefaults.coverage.exclude` would spread an empty array: a no-op that misleadingly reads as if
  it pulls in defaults. The current narrow `include: ['src/**/*.{ts,js}']` already scopes coverage correctly. If we want explicit insurance against a future widened `include`, the meaningful change would be
  literal `'**/node_modules/**'` / `'**/dist/**'` globs rather than the default spread — can add those if you'd prefer.

  **8. 422-vs-502 follow-up** — Acknowledged and flagging here so it isn't lost: `transaction-verifier.helper` throws `502` on its synchronous paths and `422` on the equivalent async paths for identical
  messages, and the spec deliberately uses message-only assertions (~lines 1935/2002/2313) to stay green across both. Not opening a separate issue for now; recording it in this thread instead.

PooyaRaki:
@claude review

compojoom (approved):

PR #3139, opened 2026-06-05
chore(tests): reduce CI flakiness (clearMocks + seeded faker)

vseehausen:
## Summary

Reduces CGW CI flakiness (WA-1753). The flakes don't reproduce locally, so fixes were verified entirely on remote CI: re-running the **same commit SHA** via `gh run rerun`, stopping on the first failure, fixing it, and restarting on a new commit until the final commit survived a clean streak.

**Verification: 50/50 consecutive green CI runs** on the final commit (`2e3d7543`, PR run [27129491048](https://github.com/safe-global/safe-client-gateway/actions/runs/27129491048)) across every job (unit, integration, biome, license-headers, env-validation). Rule of three ⇒ per-run flake rate **< 6% at 95% confidence** (baseline was ~7%).

### Fixes

- **`clearMocks` + seeded faker (all jest configs)** — recorded mock calls no longer leak between tests; faker is seeded per file and the seed is logged (`FAKER_SEED`-overridable) so a failing run's data is reproducible.
- **`job-queue.service.integration`** — BullMQ `completed`/`failed` events fire asynchronously; `afterEach` cleared the consumer state before they arrived, leaking a prior job into the next test. Now drains unstarted jobs, then waits until every handled job has emitted its terminal event before clearing.
- **`transaction-data.mapper` (production bug)** — `getAddress(value)` throws *synchronously* as an argument, before the `.catch` attaches, so a decoded `address`-typed parameter carrying a non-address value surfaced as an uncaught **HTTP 500** in transaction preview. Guarded with `isAddress(value, { strict: false })`.
- **`redis.cache.service` (production bug)** — `deviateRandomlyByPercentage` could return a fractional TTL, which Redis `EX`/`EXPIRE` reject (`ERR value is not an integer`). Floored in `enforceMaxRedisTTL`, covering `hSet` / `increment` / `setCounter`.
- **`rate-limit.guard`** — the test used `faker.string.sample(10)` as the "invalid" IP, which occasionally parses as a valid IPv4/IPv6 so the guard accepted it ("resolved instead of rejected"). Now uses a deterministically-invalid literal.
- **`faker-setup.ts`** — added the SPDX license header (the new file failed the `license-headers` job).

### Findings

- The rerun loop surfaced **two real production bugs**, not just test issues: the unguarded `getAddress` 500 and the fractional Redis TTL.
- The rate-limit flake only appeared after **38 consecutive green runs** (~2.6% rate) — a long-tail flake no single CI run would catch, which is why the high rerun count matters.

Left as **draft** for review.

## Related

- WA-1753
- Flaky-test tracking data: `reports/flaky-tests-baseline` branch


src/routes/common/guards/rate-limit.guard.spec.ts
```diff
@@ -86,7 +86,10 @@ describe('RateLimitGuard', () => {
   });
 
   it('should log a warning for invalid client IP', async () => {
-    const invalidIp = faker.string.sample(10);
+    // A literal, never a random string: faker.string.sample can occasionally
+    // produce a value that parses as a valid IPv4/IPv6, which made the guard
+    // accept it and the rejection assertion flake.
+    const invalidIp = 'invalid-ip-address';
```
vseehausen, line 93: Instead of a literal, we could write a small function/loop to generate an invalid IP, or use a library.

PooyaRaki (dismissed):

PooyaRaki (approved):

PR #3140, opened 2026-06-05
ci: reduce pipeline runtime

PooyaRaki:
## Summary
  
Reduce CI pipeline runtime by eliminating redundant work across all jobs.

## Changes

  - Scope push trigger to main only — PRs were firing both push and pull_request events on every commit, running the full pipeline twice. CI still runs on every PR commit via
  pull_request: synchronize
  - Add concurrency group to cancel superseded runs on rapid PR pushes (main is exempt so in-flight staging deploys are never killed)
  - Replace nest build with yarn generate-abis in unit-tests and integration-tests — tests transform src/ directly via the test runner and never use dist/, so the full TypeScript
  compile was wasted work
  - Cache node_modules in all four Node jobs, keyed on yarn.lock + Node major version — skips the link step entirely on a warm cache (the existing setup-node yarn cache only
  covered downloads)
  - Drop QEMU-emulated arm64 Docker build — production runs on amd64; both publish jobs now build linux/amd64 natively on ubuntu-latest

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,17 +77,12 @@ jobs:
         with:
           node-version: '24.11.0' # Krypton
           cache: 'yarn'
-      # Cache the linked node_modules so jobs skip the install/link step on a
-      # warm cache. Keyed on the lockfile + Node major; any dependency change
-      # busts it and the conditional install below repopulates it.
-      - name: Cache node_modules
-        id: node-modules
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: node_modules
-          key: node-modules-${{ runner.os }}-node24-${{ hashFiles('yarn.lock') }}
-      - if: steps.node-modules.outputs.cache-hit != 'true'
-        run: yarn install --immutable
+      # Install unconditionally rather than caching the linked node_modules:
+      # `enableHardenedMode` (.yarnrc.yml) verifies the supply chain only during
+      # `yarn install`, so skipping install on a warm cache would bypass that
+      # check. The `cache: 'yarn'` download cache above keeps this fast — tarballs
+      # are already fetched, so install only re-links and re-verifies.
+      - run: yarn install --immutable
       - run: yarn run env:validate:silent
 
   biome:
@@ -98,16 +93,16 @@ jobs:
         with:
           node-version: '24.11.0' # Krypton
           cache: 'yarn'
-      - name: Cache node_modules
-        id: node-modules
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: node_modules
-          key: node-modules-${{ runner.os }}-node24-${{ hashFiles('yarn.lock') }}
-      - if: steps.node-modules.outputs.cache-hit != 'true'
-        run: yarn install --immutable
+      - run: yarn install --immutable
       - run: yarn run format-check
       - run: yarn run lint-check
+      # `nest build` (a full tsc compile) was dropped from the test jobs, which
+      # removed whole-program type-checking on PRs — ts-jest only type-checks
+      # files a test imports. Run an explicit no-emit type-check here so type
+      # errors fail the PR instead of only surfacing on the main-only Docker
+      # build. Needs the generated ABIs that the `@/abis` alias resolves to.
+      - run: yarn generate-abis
+      - run: yarn tsc --noEmit -p tsconfig.build.json
 
   unit-tests:
     runs-on: ubuntu-latest
@@ -154,14 +149,7 @@ jobs:
         with:
           node-version: '22.11.0' # jod
           cache: 'yarn'
-      - name: Cache node_modules
-        id: node-modules
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
-        with:
-          path: node_modules
-          key: node-modules-${{ runner.os }}-node22-${{ hashFiles('yarn.lock') }}
-      - if: steps.node-modules.outputs.cache-hit != 'true'
-        run: yarn install --immutable
+      - run: yarn install --immutable
       # Tests are transformed straight from `src` by the test runner, so the
       # full `nest build` (compile to dist) isn't needed here — only the
       # generated ABIs that the `@/abis` import alias resolves to.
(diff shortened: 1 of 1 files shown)


clovisdasilvaneto (commented):
I see now we are caching the node_modules could you explain me how the cache will be invalidated when we update or add new dependencies? Which strategy is it going to use?

PooyaRaki:
@clovisdasilvaneto 
``` bash
key: node-modules-${{ runner.os }}-node24-${{ hashFiles('yarn.lock') }}
```
The last part - hashing the lock file - is how we invalidate the cache. So if you add a new dep the hash would be different as a result the cache would be invalidated.

LucieFaire:
@claude review

.github/workflows/ci.yml
```diff
@@ -64,7 +77,17 @@ jobs:
...
-      - run: yarn install --immutable
+      # Cache the linked node_modules so jobs skip the install/link step on a
+      # warm cache. Keyed on the lockfile + Node major; any dependency change
+      # busts it and the conditional install below repopulates it.
+      - name: Cache node_modules
+        id: node-modules
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: node_modules
+          key: node-modules-${{ runner.os }}-node24-${{ hashFiles('yarn.lock') }}
```
LucieFaire: in pair with what Clovis mentioned:
The cache is keyed on `hashFiles('yarn.lock')` and `yarn install --immutable` is
skipped on a hit. Two concerns:

1. **Skipped integrity check.** On a cache hit we never run `yarn install
   --immutable`, so the lockfile-vs-contents verification never happens. Whatever
   is in the cached `node_modules` is trusted as-is and then executed by
   `yarn generate-abis` / the test run — in jobs that hold `secrets.GITHUB_TOKEN`
   and `secrets.SAFE_CONFIG_BASE_URI`.

2. **Cache poisoning.** Anyone who can push a branch (collaborator / compromised
   account) can write a cache entry. Tampering with `node_modules` while keeping
   `yarn.lock` byte-identical produces the same key, which then gets restored on
   other branches and on `main`.

`setup-node` already does `cache: 'yarn'` (the download cache). Suggest dropping
the `node_modules` cache and keeping an unconditional `yarn install --immutable` —
it's fast on a warm download cache and preserves the integrity check. Trades a bit
of speed for removing the poisoning + integrity-skip vector.

LucieFaire:
hm I wonder if because of these improvements - the tests are failing? DB seems to not have the latest migrations

PooyaRaki:
→ AI review thread:

  Validated each point against the repo — outcomes below. Changes in cda62e3b.

  **1. Type coverage — valid, fixed.** Confirmed `nest build` was a real type-check: `nest-cli.json` sets no builder, so it defaults to tsc and compiled the whole program. ts-jest only checks files a test imports, so dropping it did lose PR-wide type checking. Added an explicit
  no-emit check to the `biome` job: `yarn generate-abis` + `yarn tsc --noEmit -p tsconfig.build.json` (same scope `nest build` covered — prod `src`, specs stay on ts-jest). Verified locally: passes, ~4s, no DB/services needed.

  **2. Redundant `if` — keeping it intentionally.** Correct that `on.push.branches: [main]` already makes it redundant today. But this job pushes the `:staging` image and chains into `autodeploy`, so I'd rather it stay self-guarding if the push trigger is ever broadened. Left the ref
  check in with a comment explaining why.

  **3. `restore-keys` — moot.** We went the other way on caching (see the security thread): the `node_modules` cache is removed entirely, so there's nothing to add restore-keys to.

  **4. postinstall/abis — confirmed.** `abis/` is generated, not committed; `biome` (static lint) and `env:validate` never resolve `@/abis`, so they don't need it. Note `enableScripts: false` means the `postinstall` hook doesn't auto-run on install anyway, so the explicit
  `generate-abis` in the test jobs is required, not just warm-cache insurance.

  **5. Release concurrency — no change.** The group is keyed on `github.ref`, so two different release tags land in different concurrency groups and can't cancel each other. The only shared-group case is re-running the same tag. Current config is correct.

  → @clovisdasilvaneto  & @LucieFaire (cache security) thread:

  The integrity-skip concern is valid — and sharper than it first looks. `enableHardenedMode` only runs during `yarn install`, so caching the linked `node_modules` and skipping install on a hit silently bypasses the supply-chain verification we explicitly opted into. The `cache:
  'yarn'` download cache is different — yarn checksum-verifies it against the lockfile on install — so it's not trusted as-is the way `node_modules` is.

  One correction on the poisoning path: GitHub Actions cache is branch-scoped. `main` (default branch) restores only caches written on `main`, and writing those needs merge access; a feature-branch cache doesn't propagate to `main` or to sibling branches. So the "restored on other
  branches and on main" route doesn't actually hold. Doesn't change the conclusion though.

  Took the recommendation (cda62e3b): removed the `node_modules` cache from all four jobs, kept `setup-node` `cache: 'yarn'` + an unconditional `yarn install --immutable`. Tarball downloads stay cached (the expensive part), and hardened-mode verification now runs on every job. I'll
  check the warm-cache install time in the first run to confirm the speed cost is small.

PooyaRaki:
@claude review again

PooyaRaki:
Addressed the round-2 nits — keeping all three as-is, reasoning below:
  
  1. Splitting tsc --noEmit into its own type-check job

  Holding off on this one. This branch's goal is to cut pipeline runtime, and a separate job re-pays the expensive setup (checkout + setup-node + yarn install --immutable) just to move one tsc step. The current ordering is also intentionally fail-fast: the near-instant
  format-check/lint-check gate the slower generate-abis + tsc, so most failing PRs never burn runner minutes on the type-check. The masking case (format error hides a type error) is low-probability since format issues are auto-fixable and caught by pre-commit locally. Net runner cost
  outweighs the DX gain here — happy to revisit if format-masking actually bites in practice.

  2. incremental: true without a .tsbuildinfo cache

  incremental is there for the production nest build, not CI. On a fresh CI checkout the build-info file lands in dist/ and is discarded — negligible, as noted. Caching it across runs adds a source-keyed cache for marginal gain (cold cache on every first PR push regardless), and
  dropping incremental would penalize the real build. Leaving it.

  3. yarn generate-abis across three jobs

  It's a fast synchronous node ./scripts/generate-abis.js — reads JSON from node_modules/@safe-global/safe-deployments and writes abis/safe. Won't show up in CI timing, so skipping the abis/ cache to avoid the key-management overhead.

  All three are non-blocking; no functional changes this round.

PooyaRaki:
@claude review again

LucieFaire (approved):

PR #3141, opened 2026-06-08
Replace hardcoded Zerion routing with config service values

Zalens:
Drives Zerion routing from the config service's per-chain `balancesProvider.enabled` instead of a global env flag. Closes PLA-1116.

### DIFF
--- a/src/modules/balances/datasources/balances-api.manager.ts
+++ b/src/modules/balances/datasources/balances-api.manager.ts
@@ -28,6 +28,7 @@ import { type Raw, rawify } from '@/validation/entities/raw.entity';
 export class BalancesApiManager implements IBalancesApiManager {
   private safeBalancesApiMap: Record<string, SafeBalancesApi> = {};
   private readonly isCounterFactualBalancesEnabled: boolean;
+  private readonly zerionBalancesEnabled: boolean;
   private readonly zerionBalancesApi: IBalancesApi;
   private readonly useVpcUrl: boolean;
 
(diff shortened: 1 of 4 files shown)


src/modules/safe/routes/v2/safes.v2.service.ts
```diff
@@ -140,6 +140,7 @@ export class SafesV2Service {
     const { chain, safeAddress, currency, trusted } = args;
 
     if (
+      chain.balancesProvider?.enabled &&
```
vseehausen, line 143: Remove. We don't need a fourth gate on this.

src/modules/balances/datasources/balances-api.manager.ts
```diff
@@ -61,13 +65,22 @@ export class BalancesApiManager implements IBalancesApiManager {
...
     if (isSafe) {
       return this._getSafeBalancesApi(chainId);
     }
-    return this.zerionBalancesApi;
+
+    if (this.zerionBalancesEnabled) {
+      const chain = await this.configApi
+        .getChain(chainId)
+        .then(ChainSchema.parse);
+      if (chain.balancesProvider.enabled) {
```
vseehausen: Let's not add balancesProvider enabled here. It's unused currently, and the name is misleading. We should rather consolidate and simplify how we manage balances, than adding a new feature flag.

Please also take into account, that this is basically dead code, as wallet does not query the balances endpoint for counterfactual Safes.

vseehausen (changes requested):

src/modules/safe/routes/v2/safes.v2.service.ts
Zalens: this is the main point of PLA-1116 - making `balancesProvider.enabled` actually drive the routing. it's not really a redundant gate: `getZerionChainName` only looks at `chainName` and `zerionBalancesEnabled` is the global kill-switch, neither of them checks enabled. so right now a chain with enabled: false but a `chainName` set still goes to Zerion.

src/modules/balances/datasources/balances-api.manager.ts
Zalens: yeah agree, reverting this one. it's a dead path anyway since wallet doesn't hit balances for counterfactual safes

PR #3142, opened 2026-06-08
fix(spaces): wire SES invite email queue

vseehausen:
## Summary

Fixes SES invite email delivery by wiring the SES queue module into the Spaces module, where `SpaceInviteEmailService` is instantiated.

## Root cause

`SesEmailModule` was imported as an AppModule sibling. That starts the SES queue module, but it does not make `SesEmailQueueService` visible to providers declared inside `SpacesModule`. Because the queue dependency in `SpaceInviteEmailService` is optional, invite creation succeeded while email queueing became a no-op.

## Changes

- Import `SesEmailModule` conditionally from `SpacesModule`.
- Remove the top-level `SesEmailModule` import from `AppModule` to avoid duplicate queue workers.

## Checks

- `yarn test:e2e src/app.module.e2e-spec.ts --no-watchman`
- `yarn run format-check`
- `yarn run lint-check`

### DIFF
--- a/src/app.module.ts
+++ b/src/app.module.ts
@@ -44,7 +44,6 @@ import { CounterfactualSafesModule } from '@/modules/counterfactual-safes/counte
 import { CsvExportModule } from '@/modules/csv-export/csv-export.module';
 import { DataDecoderModule } from '@/modules/data-decoder/data-decoder.module';
 import { DelegateModule } from '@/modules/delegate/delegate.module';
-import { SesEmailModule } from '@/modules/email/ses/ses-email.module';
 import { EstimationsModule } from '@/modules/estimations/estimations.module';
 import { FeesModule } from '@/modules/fees/fees.module';
 import { HealthModule } from '@/modules/health/health.module';
@@ -80,7 +79,6 @@ export class AppModule implements NestModule {
       oidc_auth: isOidcAuthFeatureEnabled,
       users: isUsersFeatureEnabled,
       email: isEmailFeatureEnabled,
-      sesEmail: isSesEmailFeatureEnabled,
       zerionPositions: isZerionPositionsFeatureEnabled,
     } = configFactory().features;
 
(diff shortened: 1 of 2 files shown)


src/config/entities/schemas/configuration.schema.ts
LucieFaire, line 229: I think something wasnt copied correctly last time:
```
  .superRefine((config, ctx) => {
    // These fields are only required in deployed (production/staging) environments.
    const isDeployedEnv =
      !!config.CGW_ENV && ['production', 'staging'].includes(config.CGW_ENV);
    if (!isDeployedEnv) {
      return;
    }
    const isSesEnabled = config.FF_SES_EMAIL?.toLowerCase() === 'true';

    for (const {
      field,
      requiredWhen = true,
      message = 'is required in production and staging environments',
    } of [
      { field: 'AWS_ACCESS_KEY_ID' },
      { field: 'AWS_KMS_ENCRYPTION_KEY_ID' },
      { field: 'AWS_SECRET_ACCESS_KEY' },
      { field: 'AWS_REGION' },
      { field: 'CSV_AWS_ACCESS_KEY_ID' },
      { field: 'CSV_AWS_SECRET_ACCESS_KEY' },
      { field: 'BLOCKAID_CLIENT_API_KEY' },
      // SES permissions are set via IRSA in deployed environments, not static AWS keys.
      {
        field: 'AWS_WEB_IDENTITY_TOKEN_FILE',
        requiredWhen: isSesEnabled,
        message:
          'is required in production and staging environments when SES email is enabled',
      },
    ]) {
      if (requiredWhen && !(config as Record<string, unknown>)[field]) {
        ctx.addIssue({ code: 'custom', message, path: [field] });
      }
    }
  });
```
This refine only exists for staging/prod. Adding checks for local env defeats the purpose. I would suggest to remove the checks for local AWS* config

src/config/configuration.validator.spec.ts
```diff
@@ -332,6 +334,51 @@ describe('Configuration validator', () => {
...
+        AWS_WEB_IDENTITY_TOKEN_FILE: faker.system.filePath(),
+        FF_SES_EMAIL: 'true',
+      };
+
+      expect(configurationValidator(config, RootConfigurationSchema)).toBe(
+        config,
+      );
+    });
+  });
+
```
LucieFaire: how are these possible if we explicitly state: on production, if SES is enabled => token file is a MUST ? Prod and Staging should never accept separate AWS* keys

LucieFaire (approved):

PR #3153, opened 2026-06-09
feat: send email on renew invite flow

LucieFaire:
## Summary [WA-2512](https://linear.app/safe-global/issue/WA-2512/wire-email-delivery-into-resend-endpoint)

Sends an invite email when an email-based space invite is **renewed**, reusing the same template/subject as the original invite. 

## Changes

**Renewal email flow**
- `src/modules/spaces/routes/members.service.ts`: `renewInvite` now loads the member with its `user` relation (`{ user: true }`), and after a successful renewal enqueues a renewal email when the member has a stored email. Email presence is the signal for an email invite — email invites create a user via `findOrCreateByEmail` (sets `user.email`); wallet invites don't.
- `src/modules/spaces/routes/space-invite-email.service.ts`: extracted a shared private `enqueueEmails({ recipients, spaceId })` core; `enqueueInviteEmails` maps email invites to recipients and delegates; added a public `enqueueRenewalEmail({ name, email, spaceId })` for the single-recipient renewal case. Same subject, template, and metadata as the original invite.
- `src/modules/spaces/routes/entities/invite-users.dto.entity.ts`: exported `EmailInviteUserInput` (inferred from the existing email invite schema) to type the invite filter cleanly.

**Config validation refactor**
- `src/config/entities/schemas/configuration.schema.ts`: simplified the production/staging `superRefine` — early-return for non-deployed environments and hoisted the `isSesEnabled` check, removing the repeated per-field env guard.

## Tests
- `src/modules/spaces/routes/members.service.spec.ts`: renew loads the member with its user relation; enqueues a renewal email when the member has an email; does not enqueue (but still renews) when there's no email.
- `src/modules/spaces/routes/space-invite-email.service.spec.ts`: `enqueueRenewalEmail` enqueues a correct job, is a no-op when the SES queue is unavailable, and swallows/logs enqueue errors.

### DIFF
--- a/src/config/entities/schemas/configuration.schema.ts
+++ b/src/config/entities/schemas/configuration.schema.ts
@@ -195,7 +195,14 @@ export const RootConfigurationSchema = z
     CAPTCHA_SECRET_KEY: z.string().optional(),
   })
   .superRefine((config, ctx) => {
-    // Check for AWS_* and Blockaid fields in production and staging environments
+    // These fields are only required in deployed (production/staging) environments.
+    const isDeployedEnv =
+      !!config.CGW_ENV && ['production', 'staging'].includes(config.CGW_ENV);
+    if (!isDeployedEnv) {
+      return;
+    }
+    const isSesEnabled = config.FF_SES_EMAIL?.toLowerCase() === 'true';
+
     for (const {
       field,
       requiredWhen = true,
@@ -211,18 +218,12 @@ export const RootConfigurationSchema = z
       // SES permissions are set via IRSA in deployed environments, not static AWS keys.
       {
         field: 'AWS_WEB_IDENTITY_TOKEN_FILE',
-        requiredWhen: config.FF_SES_EMAIL?.toLowerCase() === 'true',
+        requiredWhen: isSesEnabled,
         message:
           'is required in production and staging environments when SES email is enabled',
       },
     ]) {
-      if (
-        config.CGW_ENV &&
-        config instanceof Object &&
-        ['production', 'staging'].includes(config.CGW_ENV) &&
-        requiredWhen &&
-        !(config as Record<string, unknown>)[field]
-      ) {
+      if (requiredWhen && !(config as Record<string, unknown>)[field]) {
         ctx.addIssue({ code: 'custom', message, path: [field] });
       }
     }
--- a/src/modules/spaces/routes/members.service.ts
+++ b/src/modules/spaces/routes/members.service.ts
@@ -106,10 +106,10 @@ export class MembersService {
     return {
       userId: args.userId,
       spaceId: args.spaceId,
-      name: name,
-      role: role,
-      status: status,
-      invitedBy: invitedBy,
+      name,
+      role,
+      status,
+      invitedBy,
     };
   }
 
--- a/src/modules/spaces/routes/members.service.ts
+++ b/src/modules/spaces/routes/members.service.ts
@@ -79,25 +79,37 @@ export class MembersService {
       authPayload: args.authPayload,
       spaceId: args.spaceId,
     });
-    const member = await this.membersRepository.findOneOrFail({
-      user: { id: args.userId },
-      space: { id: args.spaceId },
-    });
-    if (member.status !== 'INVITED') {
+    const { id, user, name, status, role, invitedBy } =
+      await this.membersRepository.findOneOrFail(
+        {
+          user: { id: args.userId },
+          space: { id: args.spaceId },
+        },
+        { user: true },
+      );
+    if (status !== 'INVITED') {
       throw new ConflictException('Only a pending invitation can be renewed.');
     }
     await this.membersRepository.renewInvite({
-      memberId: member.id,
+      memberId: id,
       inviteExpiresAt: new Date(Date.now() + this.inviteTtlMs),
     });
+
+    if (user.email) {
+      await this.spaceInviteEmailService.enqueueRenewalEmail({
+        name: name,
+        email: user.email,
+        spaceId: args.spaceId,
+      });
+    }
 
     return {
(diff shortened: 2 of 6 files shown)


src/modules/spaces/routes/members.service.ts
```diff
@@ -79,25 +79,37 @@ export class MembersService {
       authPayload: args.authPayload,
       spaceId: args.spaceId,
     });
-    const member = await this.membersRepository.findOneOrFail({
-      user: { id: args.userId },
-      space: { id: args.spaceId },
-    });
-    if (member.status !== 'INVITED') {
+    const { id, user, name, status, role, invitedBy } =
```
vseehausen, line 82: What's your rationale for destructuring the member object?

src/modules/spaces/routes/members.service.ts
```diff
@@ -79,25 +79,37 @@ export class MembersService {
...
       userId: args.userId,
       spaceId: args.spaceId,
-      name: member.name,
-      role: member.role,
-      status: member.status,
-      invitedBy: member.invitedBy,
+      name: name,
+      role: role,
+      status: status,
+      invitedBy: invitedBy,
```
vseehausen: And then not using the short form in the return object?

src/modules/spaces/routes/space-invite-email.service.ts
```diff
@@ -51,41 +58,69 @@ export class SpaceInviteEmailService {
...
 
-    if (emailInvites.length === 0) {
+    if (!recipients.length) {
       return;
     }
 
     try {
-      const space = await this.spacesRepository.findOneOrFail({
-        where: { id: args.spaceId },
+      const { name } = await this.spacesRepository.findOneOrFail({
```
vseehausen: Again, not the biggest fan of this destructuring. I prefer having objects, as they signal provenance. In this function, we have space.name, recipient.name, so we should keep provenance clear. Not a blocker, though.

src/config/entities/schemas/configuration.schema.ts
```diff
@@ -195,7 +195,14 @@ export const RootConfigurationSchema = z
...
   })
   .superRefine((config, ctx) => {
-    // Check for AWS_* and Blockaid fields in production and staging environments
+    // These fields are only required in deployed (production/staging) environments.
+    const isDeployedEnv =
+      !!config.CGW_ENV && ['production', 'staging'].includes(config.CGW_ENV);
+    if (!isDeployedEnv) {
+      return;
+    }
+    const isSesEnabled = config.FF_SES_EMAIL?.toLowerCase() === 'true';
```
vseehausen, line 204: I am always a fan of the Girl Scout rule. Happy to approve these non-related changes.

vseehausen (changes requested):
Code convention is not to desctrucure return values, but to keep the named reciever (`const member = ...` instead of `const { name } = ...`). I would suggest we keep and establish this rule.

src/modules/spaces/routes/members.service.ts
LucieFaire: I prefer to extract only what i need and use directly wihtout constant `args.*` especially is its repeated multiple times

src/modules/spaces/routes/members.service.ts
LucieFaire: thats an overlook :) will update

src/modules/spaces/routes/space-invite-email.service.ts
LucieFaire: I will override the name to make it clearer

src/config/entities/schemas/configuration.schema.ts
LucieFaire: yeah, i didnt want to add yet another comment on your PR, so I thought i just tweak it in my own

LucieFaire:
> Code convention is not to desctrucure return values, but to keep the named reciever (`const member = ...` instead of `const { name } = ...`). I would suggest we keep and establish this rule.

we have such convention? could you point me there please

src/modules/spaces/routes/space-invite-email.service.ts
LucieFaire: [e4e665b](https://github.com/safe-global/safe-client-gateway/pull/3153/commits/e4e665b71885e543efbfb60fe1b64c7e20d22c28)

src/modules/spaces/routes/members.service.ts
LucieFaire: [e4e665b](https://github.com/safe-global/safe-client-gateway/pull/3153/commits/e4e665b71885e543efbfb60fe1b64c7e20d22c28)

LucieFaire:
fyi:
destructuring is widespread in production code. The full list is **79 non-test files** spanning services, datasources, mappers, repositories, controllers, and filters. A few highlights showing it's an established pattern across the codebase:

- **Services:** `members.service.ts`, `space-invite-email.service.ts`, `balances.service.ts`, `positions.service.ts`, `csv-export.service.ts`, the whole `safe-shield/` module
- **Datasources:** `transaction-api.service.ts`, `kiln-api.service.ts`, `gelato-api.service.ts`, `fetch.network.service.ts`
- **Mappers:** most files under `transactions/routes/mappers/`
- **Repositories:** `safe.repository.ts`, `members.repository.ts`, `bridge.repository.ts`
- **Controllers / filters / guards:** `auth.controller.ts`, `global-error.filter.ts`, `rate-limit.guard.ts`

(A handful in `src/__tests__/` and `*.builder.ts` are test-adjacent helpers, but the large majority above are genuine production code.) So it's clearly an accepted, common pattern — not something the repo avoids.

vseehausen:
If you insist.

vseehausen (approved):

PR #3155, opened 2026-06-10
feat: rate limit space invite and invite renewal endpoints

vseehausen:
## Summary

- Replaces the per-IP, request-counting rate limit on space invites with volume-based limits: the unit is now invited users (not HTTP requests), so the limit can no longer be multiplied by batching up to `maxInvites` users per request.
- Per-Space invitee budget (default 200 per 10 min, `429` when exceeded), enforced in `MembersService` for both initial invites (counting `users.length`) and renewals (counting 1). Keying on the Space instead of the IP avoids corporate-NAT collisions and makes minting extra admins inside a Space pointless, since all of its admins share one budget.
- Per-recipient email cap (default 5 per 24 h) in `SpaceInviteEmailService`: invite and renewal emails to one address are suppressed beyond the cap (invitation stays valid and visible in-app; no error is surfaced so callers cannot probe how often an address has been invited). Recipient addresses are sha256-hashed before use in cache keys/logs.
- Global invite email counter (default 1,000 per hour) that only logs a warning above the threshold — alerting plus the existing kill switch are the enforcement, so a legitimate spike cannot block invite emails system-wide.
- Adds `ICacheService.incrementBy` (Redis `INCRBY`) to support counting more than one unit per operation.
- Removes `SpacesInvitationRateLimitGuard` and its e2e bypass; the per-IP guard pattern remains on the space-creation surface where it fits.
- Config under `spaces.rateLimit.invitation` via `SPACES_INVITATION_RATE_LIMIT_MAX/_WINDOW_SECONDS`, `SPACES_INVITATION_RECIPIENT_RATE_LIMIT_MAX/_WINDOW_SECONDS` and `SPACES_INVITATION_EMAILS_ALERT_THRESHOLD/_WINDOW_SECONDS`, documented in `.env.sample.json`.

### DIFF
--- a/.env.sample.json
+++ b/.env.sample.json
@@ -1434,6 +1434,42 @@
     "required": false
   },
   {
+    "name": "SPACES_INVITATION_RATE_LIMIT_MAX",
+    "description": "Maximum invited users (initial invites plus renewals) per Space per time window",
+    "defaultValue": 200,
+    "required": false
+  },
+  {
+    "name": "SPACES_INVITATION_RATE_LIMIT_WINDOW_SECONDS",
+    "description": "Time window in seconds for the per-Space invitation rate limit",
+    "defaultValue": 600,
+    "required": false
+  },
+  {
+    "name": "SPACES_INVITATION_RECIPIENT_RATE_LIMIT_MAX",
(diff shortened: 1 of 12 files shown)


vseehausen:
Closed for now.

PR #3157, opened 2026-06-10
chore: rename zerionBalancesEnabled flag to zerion

vseehausen:
## Summary

- Renames the `features.zerionBalancesEnabled` feature flag to `features.zerion` — it gates Zerion as a whole, not just balances. The `FF_ZERION_ENABLED` env var is unchanged.
- Drops the legacy `FF_ZERION_BALANCES_CHAIN_IDS` env var fallback, plus its `.env.sample.json` and devcontainer entries. The flag is now driven solely by `FF_ZERION_ENABLED`.
- Parses `FF_ZERION_ENABLED` as a strict boolean like the other `FF_*` flags: only the literal `true` (case-insensitive) enables it. Previously any non-empty value — including `false` or a carried-over chain-id list — enabled the flag.
- Updates all consumers (`safes.v2.service`, portfolio v1 route guard) and affected specs, and adds parsing tests for the flag.

⚠️ **Deploy note:** the flag now defaults to **off** when unset — Zerion is only enabled where `FF_ZERION_ENABLED=true` is set explicitly. Confirmed with InfraSec that `FF_ZERION_ENABLED=true` is already set on both staging and production, so the rollout ordering (env var first, then deploy) is satisfied and this can go out safely.

## Out of scope (follow-up)

Consolidating all Zerion gating behind `features.zerion` — replacing `features.zerionPositions` / `FF_ZERION_POSITIONS_DISABLED` and the per-chain `PORTFOLIO_ENDPOINT` / `balancesProvider.chainName` gates with the Zerion supported-chain list — is intentionally left for a follow-up PR.

## Related

- [WA-2750](https://linear.app/safe-global/issue/WA-2750/rename-zerionbalancesenabled-flag-to-zerionenabled) (part of [WA-2749](https://linear.app/safe-global/issue/WA-2749/refactor-balances-and-portfolio-endpoint-at-cgw))
- Rebased on top of #3199 (per-chain Zerion eligibility), which now reads this flag via `resolveZerionEligibility` — the rename is applied there too.


vseehausen:
@claude review

vseehausen:
@claude review again

vseehausen:
Re: the review above —

- **Rollout ordering**: confirmed with InfraSec that `FF_ZERION_ENABLED=true` is already set on both staging and production, so both silent-disable modes are contained. Deploy note updated accordingly.
- **`defaultValue` type nit**: aligned to the bare boolean `false` in d9c10856c (bare booleans are the dominant style in the file, 27:1).
- **`vi.stubEnv` nit**: left as is — `stubEnv` isn't used anywhere in the repo; the save/restore pattern matches the adjacent `trustProxy` suite in the same file.

vseehausen:
@claude review again.

The deploy risk is contained, InfraSec confirmed the flag is set on staging and on production.

vseehausen:
@claude review again. Deploy risk is confirmed and contained, no need to mention it again. The tests obviously work, so it's fine to set the flag at afterEach.

src/config/entities/__tests__/configuration.ts
```diff
@@ -204,7 +204,7 @@ export default (): ReturnType<typeof configuration> => ({
   features: {
     email: false,
     sesEmail: false,
-    zerionBalancesEnabled: false,
+    zerionEnabled: false,
```
LucieFaire: nothing against it, but it kind of doesnt go very well with other FF defined here. We have `zerionPositions` already there, so just calling it `Zerion` would be confusing, what about `zerionBalances` or smth with a bit more context (yet i know it was partialy that before)

LucieFaire: or if you say its for whole Zerion, should we lose `zerionPositions` ?

LucieFaire: but would still then call it `zerion` then w/o `enabled`

vseehausen: > or if you say its for whole Zerion, should we lose zerionPositions ?

Yes, I think that is the way to go. 👍 

> but would still then call it zerion then w/o enabled

Right.

vseehausen: Let's merge this rename and consolidate the feature flags in another PR.

src/config/entities/configuration.spec.ts
```diff
@@ -32,3 +32,34 @@ describe('configuration - express.trustProxy', () => {
...
+
+    expect(configuration().features.zerion).toBe(true);
+  });
+
+  it.each(['false', '0', '1,10,137'])('is disabled when set to %s', (value) => {
+    process.env[ENV_KEY] = value;
+
+    expect(configuration().features.zerion).toBe(false);
+  });
+});
```
LucieFaire, line 65: dont hurt to have but also not sure we really need these tests :) thye dont bring much value

LucieFaire (approved):

src/config/entities/configuration.spec.ts
vseehausen: Yep, it's been one of the multiple Claude reviews that made me add them. I will leave them for now.

LucieFaire: dont trust everything claude says :)

PR #3160, opened 2026-06-10
feat: filter spam tokens notifications

LucieFaire:
### Problem
Safe{Mobile} gets a push from CGW for every incoming ERC20 transfer, including tokens the app later hides as spam. Users get a notification, open the app, and see nothing — confusing and erodes trust in notifications.

### Solution
Before queueing an `INCOMING_TOKEN` push, check whether the token appears in the Safe's spam-filtered asset list — the same source of truth the app shows. Push only if present; otherwise drop the notification silently. Native (`INCOMING_ETHER`) and all other event types are unaffected.

The check runs once per event in `PushNotificationService.processEvent`, before subscriber resolution, and emits a dedicated `NOTIFICATION_SPAM_TOKEN_DROPPED` log when it suppresses (distinct from a normal zero-delivery event, so spam-filter rate is observable).

### Source of truth (hybrid, mirroring the app)
The mobile asset list is provider-dependent, so the gate mirrors it:
- **Zerion-covered chains** → the **Zerion portfolio** (`getPortfolio({ trusted: true, excludeDust: false })`, i.e. `filter[trash]=only_non_trash`) is authoritative. This is the app's default "Default tokens" view and the literal "Zerion list, spam filter on."
- **Chains Zerion doesn't cover** → fall back to the **balances list** (`getBalances({ trusted: false, excludeSpam: true })`) — the app's own fallback. Coverage is determined via `ZerionChainMappingService.getNetworkFromChainId`.

### Fail-safe behavior
- Portfolio lookup unavailable on a covered chain, or chain lookup fails → **suppress** the push (don't risk notifying on possible spam).
- Chain-mapping (coverage) lookup fails → **degrade gracefully** to the balances list rather than suppress.

### Files changed
- src/modules/notifications/domain/push/push-notification.service.ts — `INCOMING_TOKEN` gate + `isTokenOnAssetList` / `isTokenInPortfolio` / `isTokenInBalances`; injects portfolio, balances, chains repos and the Zerion chain-mapping service.
- src/modules/notifications/domain/push/push-notification.module.ts — imports `PortfolioModule`, `ZerionModule`, `BalancesModule`, `ChainsModule`.
- src/domain/common/entities/log-type.entity.ts — adds `NotificationSpamTokenDropped`.
- src/modules/notifications/domain/push/push-notification.service.spec.ts — unit tests for both paths (portfolio/balances), coverage routing, checksum-insensitive match, native/non-asset bypass, and fail-safe cases.
- src/modules/notifications/domain/push/push-notification.integration.spec.ts — incoming-token cases mock the asset-list lookup; added a spam-suppression case.

### Known trade-off
On a Zerion-covered chain the portfolio is authoritative, so a token Zerion flags as trash is dropped even if the tx-service `exclude_spam` filter would have kept it. Push parity is therefore with the app's default "Default tokens" view; a user in "All tokens" mode could occasionally see an in-app token that didn't push. This is inherent to the asset list being user-configurable.

A newly received legitimate token may also be briefly suppressed until the provider indexes it — consistent with the app, which wouldn't show it yet either. The hook clears the balances cache on the event, so the lookup is fresh.


compojoom (approved):

tmjssz (approved):

src/modules/notifications/domain/push/push-notification.service.ts
```diff
@@ -427,6 +455,83 @@ export class PushNotificationService implements IPushNotificationService {
...
+      return false;
+    }
+  }
+
+  /**
+   * Checks the token against the Zerion portfolio with the spam filter on
+   * (trusted=true → only non-trash). Dust is not excluded: a small legitimate
+   * transfer is not spam.
+   */
+  private async isTokenInPortfolio(
```
vseehausen, line 493: I see a potential race condition. When a legitimate token transfer did not settle on Zerion, when this even triggers, it won't be included in the Zerion response, and thus tagged as spam.

It would be better to use one of the endpoints for fungibles of Zerion.

vseehausen (changes requested):

PR #3163, opened 2026-06-11
Feat/space audit log

compojoom:
## Summary

Adds an append-only audit log for Spaces: every space mutation (space create/update/delete, member lifecycle, Safe add/remove, address book changes, account-deletion departures) is recorded in the same database transaction as the mutation, and ACTIVE space members can read it through a new paginated endpoint. Powers the Activity log in the web app (separate PR).

Related to: https://linear.app/safe-global/issue/WA-2550

## Changes

- New `space_audit_log` table; DB triggers reject `UPDATE`/`DELETE`/`TRUNCATE` and `created_at` is always set server-side
- `SpaceAuditRepository.record()` joins the caller's transaction, so the event and the mutation commit or roll back together; writes are gated by `FF_SPACE_AUDIT_LOG` (default off)
- All space mutation paths instrumented (15 event types), including member departures caused by user-account deletion
- Address book request approval now runs the state transition and the upsert in one shared transaction (replaces the manual revert-to-pending)
- New endpoints: `GET /v1/spaces/:spaceId/audit-log` (event-type/actor/date filters, sort direction, pagination) and `GET /v1/spaces/:spaceId/audit-log/actors`; ACTIVE members only, display names resolved server-side with the same email-visibility rules as the members endpoint
- Unit + integration coverage: transactional atomicity, trigger enforcement, stable pagination ordering, filters, display-name masking

## How to test

1. Run migrations and start the service with `FF_SPACE_AUDIT_LOG=true`
2. Exercise space mutations (create a space, invite/accept a member, add a Safe, upsert/delete an address book contact)
3. As an ACTIVE member, `GET /v1/spaces/:spaceId/audit-log` returns the events newest-first; query params narrow the feed; `/audit-log/actors` lists distinct actors
4. `yarn test:unit --testPathPatterns space-audit` and `yarn test:integration --testPathPatterns space-audit`

### DIFF
--- a/migrations/1781200000000-create-space-audit-log.ts
+++ b/migrations/1781200000000-create-space-audit-log.ts
@@ -1,0 +1,64 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+export class CreateSpaceAuditLog1781200000000 implements MigrationInterface {
+  name = 'CreateSpaceAuditLog1781200000000';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      CREATE TABLE "space_audit_log" (
+        "id" BIGINT GENERATED ALWAYS AS IDENTITY,
+        -- no FK: the log must survive space deletion
+        "space_id" integer NOT NULL,
+        "space_uuid" uuid NOT NULL,
+        "event_type" character varying(64) NOT NULL,
+        "actor_user_id" integer NOT NULL,
+        "payload" jsonb NOT NULL DEFAULT '{}',
+        "created_at" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        CONSTRAINT "PK_SAL_id" PRIMARY KEY ("id")
+      )
+    `);
+    await queryRunner.query(
+      `CREATE INDEX "IDX_SAL_space_created_id" ON "space_audit_log" ("space_id", "created_at", "id")`,
+    );
+
+    await queryRunner.query(`
+      CREATE FUNCTION space_audit_log_immutable()
+      RETURNS TRIGGER AS $$
+      BEGIN
+          RAISE EXCEPTION 'space_audit_log is append-only';
+      END;
+      $$ language 'plpgsql';`);
+    await queryRunner.query(`
+      CREATE TRIGGER space_audit_log_no_update_delete
+        BEFORE UPDATE OR DELETE ON space_audit_log
+        FOR EACH ROW EXECUTE PROCEDURE space_audit_log_immutable();`);
+    await queryRunner.query(`
+      CREATE TRIGGER space_audit_log_no_truncate
+        BEFORE TRUNCATE ON space_audit_log
+        FOR EACH STATEMENT EXECUTE PROCEDURE space_audit_log_immutable();`);
+
+    await queryRunner.query(`
+      CREATE FUNCTION space_audit_log_force_created_at()
+      RETURNS TRIGGER AS $$
+      BEGIN
+          NEW.created_at = CURRENT_TIMESTAMP;
+          RETURN NEW;
+      END;
+      $$ language 'plpgsql';`);
+    await queryRunner.query(`
+      CREATE TRIGGER space_audit_log_force_created_at
+        BEFORE INSERT ON space_audit_log
+        FOR EACH ROW EXECUTE PROCEDURE space_audit_log_force_created_at();`);
+  }
+
+  public async down(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`DROP TABLE "space_audit_log"`);
+    await queryRunner.query(
+      `DROP FUNCTION IF EXISTS space_audit_log_immutable`,
+    );
+    await queryRunner.query(
+      `DROP FUNCTION IF EXISTS space_audit_log_force_created_at`,
+    );
+  }
+}
--- a/src/modules/spaces/datasources/entities/space-audit-log.entity.db.ts
+++ b/src/modules/spaces/datasources/entities/space-audit-log.entity.db.ts
@@ -1,0 +1,51 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { UUID } from 'node:crypto';
+import { Column, Entity, Index, PrimaryGeneratedColumn } from 'typeorm';
+import type {
+  SpaceAuditEventPayload,
+  SpaceAuditEventType,
+} from '@/modules/spaces/domain/audit/entities/space-audit-event.entity';
+
+/**
+ * Append-only audit log of space mutations. Rows are immutable — the table
+ * carries triggers that reject UPDATE/DELETE/TRUNCATE.
+ */
+@Entity('space_audit_log')
+@Index('IDX_SAL_space_created_id', ['spaceId', 'createdAt', 'id'])
+export class SpaceAuditLog {
+  @PrimaryGeneratedColumn('identity', {
+    type: 'bigint',
+    generatedIdentity: 'ALWAYS',
+    primaryKeyConstraintName: 'PK_SAL_id',
+  })
+  id!: string;
+
+  @Column({ name: 'space_id', type: 'integer', update: false })
+  spaceId!: number;
+
+  @Column({ name: 'space_uuid', type: 'uuid', update: false })
+  spaceUuid!: UUID;
+
+  @Column({
+    name: 'event_type',
+    type: 'varchar',
+    length: 64,
+    update: false,
+  })
+  eventType!: keyof typeof SpaceAuditEventType;
+
+  @Column({ name: 'actor_user_id', type: 'integer', update: false })
+  actorUserId!: number;
+
+  @Column({ type: 'jsonb', default: {}, update: false })
+  payload!: SpaceAuditEventPayload;
+
+  @Column({
+    name: 'created_at',
+    type: 'timestamp with time zone',
+    default: () => 'CURRENT_TIMESTAMP',
+    update: false,
+    insert: false,
+  })
+  createdAt!: Date;
+}
--- a/src/modules/spaces/domain/address-books/address-book-items.repository.ts
+++ b/src/modules/spaces/domain/address-books/address-book-items.repository.ts
@@ -11,6 +11,8 @@ import { AddressBookItem as DbAddressBookItem } from '@/modules/spaces/datasourc
 import { IAddressBookItemsRepository } from '@/modules/spaces/domain/address-books/address-book-items.repository.interface';
 import type { AddressBookDbItem } from '@/modules/spaces/domain/address-books/entities/address-book-item.db.entity';
 import { AddressBookItem } from '@/modules/spaces/domain/address-books/entities/address-book-item.entity';
+import { SpaceAuditEventType } from '@/modules/spaces/domain/audit/entities/space-audit-event.entity';
+import { ISpaceAuditRepository } from '@/modules/spaces/domain/audit/space-audit.repository.interface';
 import { Space } from '@/modules/spaces/domain/entities/space.entity';
 import { ISpacesRepository } from '@/modules/spaces/domain/spaces.repository.interface';
 import { UpsertAddressBookItemsDto } from '@/modules/spaces/routes/entities/upsert-address-book-items.dto.entity';
@@ -28,6 +30,8 @@ export class AddressBookItemsRepository implements IAddressBookItemsRepository {
     private readonly spacesRepository: ISpacesRepository,
     @Inject(IConfigurationService)
     private readonly configurationService: IConfigurationService,
+    @Inject(ISpaceAuditRepository)
+    private readonly spaceAuditRepository: ISpaceAuditRepository,
   ) {
     this.maxItems = this.configurationService.getOrThrow<number>(
       'spaces.addressBooks.maxItems',
@@ -51,15 +55,18 @@ export class AddressBookItemsRepository implements IAddressBookItemsRepository {
     spaceId: Space['id'];
     addressBookItems: UpsertAddressBookItemsDto['items'];
     createdByOverride?: number;
+    entityManager?: EntityManager;
   }): Promise<Array<AddressBookDbItem>> {
     const userId = getAuthenticatedUserIdOrFail(args.authPayload);
     const space = await this.getSpaceAs({
       ...args,
       memberRoleIn: ['ADMIN'],
     });
-    const repository = await this.db.getRepository(DbAddressBookItem);
-    await this.db.transaction(async (entityManager) => {
-      const existingAddresses = await this.updateExistingAddressBookItems({
+
+    const run = async (
+      entityManager: EntityManager,
+    ): Promise<Array<AddressBookDbItem>> => {
+      const updated = await this.updateExistingAddressBookItems({
         entityManager,
         addressBookItems: args.addressBookItems,
         space,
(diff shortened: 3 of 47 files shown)


compojoom:
@claude review

src/modules/spaces/routes/space-audit.service.ts
```diff
@@ -0,0 +1,194 @@
...
+        return targetUserId === null
+          ? [row.actorUserId]
+          : [row.actorUserId, targetUserId];
+      }),
+    });
+
+    return {
+      count,
+      next: buildNextPageURL(args.routeUrl, count)?.toString() ?? null,
+      previous: buildPreviousPageURL(args.routeUrl)?.toString() ?? null,
```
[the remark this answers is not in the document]
compojoom: Fixed in 7b7bd6ed — page links are now built from the clamped `limit`/`offset` (the route URL's cursor is normalized before `buildNextPageURL`/`buildPreviousPageURL`), so an oversized or negative cursor can no longer truncate the feed silently. Added a unit test for the 500-limit/150-rows case.

compojoom:
Addressed the Claude review:

- **1 pagination links vs clamped values** — fixed in 7b7bd6ed (links are built from the normalized limit/offset; test added for the oversized-cursor case)
- **2 MEMBER_LEFT for INVITED members** — fixed in f69e8bdc: only ACTIVE memberships record a departure on account deletion; integration test now covers an accepted + a pending invite side by side
- **4 over-fetch in findActiveMemberUserIds** — fixed in f69e8bdc (`select: { id, user: { id } }`)
- **5 empty `event_type=`** — fixed in f69e8bdc (behaves like an omitted param; schema test added)
- **3 uuid lookup roundtrip** — deferred: it's a PK-indexed single-row select inside an already-open transaction; threading the uuid through every caller signature isn't worth the churn at current write volumes
- **6 `record()` strictness** — keeping as-is deliberately: a malformed payload should fail the mutation rather than silently produce an incomplete log

migrations/1781200000000-create-space-audit-log.ts
```diff
@@ -0,0 +1,64 @@
...
+  name = 'CreateSpaceAuditLog1781200000000';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+      CREATE TABLE "space_audit_log" (
+        "id" BIGINT GENERATED ALWAYS AS IDENTITY,
+        -- no FK: the log must survive space deletion
+        "space_id" integer NOT NULL,
+        "space_uuid" uuid NOT NULL,
+        "event_type" character varying(64) NOT NULL,
```
LucieFaire, line 14: what if we create an enum and store numbers instead correlating to it?

src/modules/spaces/datasources/entities/space-audit-log.entity.db.ts
```diff
@@ -0,0 +1,51 @@
...
+  @PrimaryGeneratedColumn('identity', {
+    type: 'bigint',
+    generatedIdentity: 'ALWAYS',
+    primaryKeyConstraintName: 'PK_SAL_id',
+  })
+  id!: string;
+
+  @Column({ name: 'space_id', type: 'integer', update: false })
+  spaceId!: number;
+
```
LucieFaire, line 25: when we clean these up, we should be able to remove this column, right?

src/modules/spaces/domain/audit/__tests__/space-audit.repository.mock.ts
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { ISpaceAuditRepository } from '@/modules/spaces/domain/audit/space-audit.repository.interface';
+
+export function createMockSpaceAuditRepository(): jest.MockedObjectDeep<ISpaceAuditRepository> {
```
LucieFaire, line 4: nit: nothing against it this way, but it break the existing pattern of mocks creation

src/modules/spaces/domain/address-books/address-book-items.repository.ts
```diff
@@ -119,17 +166,19 @@ export class AddressBookItemsRepository implements IAddressBookItemsRepository {
...
   }
 
+  /** @returns the updated items as `{ address, name }` pairs (new name). */
   private async updateExistingAddressBookItems(args: {
     userId: number;
     addressBookItems: Array<AddressBookItem>;
     space: Space;
     entityManager: EntityManager;
-  }): Promise<Array<DbAddressBookItem['address']>> {
+  }): Promise<Array<Pick<DbAddressBookItem, 'address' | 'name'>>> {
```
LucieFaire, line 175: why did you decide to change this method?

src/modules/spaces/domain/audit/entities/space-audit-event.entity.ts
```diff
@@ -0,0 +1,176 @@
...
+  ADDRESS_BOOK_UPSERTED = 'ADDRESS_BOOK_UPSERTED',
+  ADDRESS_BOOK_DELETED = 'ADDRESS_BOOK_DELETED',
+}
+
+export const SpaceAuditEventTypeSchema = z.enum(
+  getStringEnumKeys(SpaceAuditEventType),
+);
+
+const LowercaseAddressSchema = AddressSchema.transform((value) =>
+  value.toLowerCase(),
```
LucieFaire: Im not sure we should lowercase it. `AddressSchema` already runs `getAddress()` from viem, which both validates and returns the EIP-55 checksummed form.

src/modules/spaces/domain/audit/space-audit.repository.ts
```diff
@@ -0,0 +1,105 @@
...
+    });
+  }
+
+  public async findBySpaceId(
+    args: SpaceAuditFindArgs,
+  ): Promise<[Array<SpaceAuditLog>, number]> {
+    const repository =
+      await this.postgresDatabaseService.getRepository(SpaceAuditLog);
+
+    const where: FindOptionsWhere<SpaceAuditLog> = { spaceId: args.spaceId };
```
LucieFaire: nit, think it is a bit easier to read:
```
    const createdAtBounds = [
      ...(args.createdAtGte ? [MoreThanOrEqual(args.createdAtGte)] : []),
      ...(args.createdAtLte ? [LessThanOrEqual(args.createdAtLte)] : []),
    ];
    const where: FindOptionsWhere<SpaceAuditLog> = {
      spaceId: args.spaceId,
      ...(args.eventTypes?.length && { eventType: In(args.eventTypes) }),
      ...(args.actorUserId !== undefined && {
        actorUserId: args.actorUserId,
      }),
      ...(createdAtBounds.length && { createdAt: And(...createdAtBounds) }),
    };
```

src/modules/spaces/domain/space-safes.repository.ts
```diff
@@ -102,14 +138,12 @@ export class SpaceSafesRepository implements ISpaceSafesRepository {
 
   public async delete(args: {
     spaceId: Space['id'];
```
LucieFaire, line 140: not exactly related, but I thought we switched to using and searching by UUID and not id, wasnt this repo missed? CAuse right now you are searching to space to get UUID but it should have been avoided 👀

src/modules/spaces/routes/entities/space-audit-log.dto.entity.ts
```diff
@@ -0,0 +1,100 @@
...
+  @ApiProperty({ type: Number })
+  actorUserId!: number;
+
+  @ApiProperty({
+    type: String,
+    description: 'Resolved (and masked) display string of the acting user.',
+  })
+  actor!: string;
+
+  @ApiPropertyOptional({
```
LucieFaire: is it optional? if so should be `targetUser?:`

src/modules/spaces/domain/spaces.repository.ts
```diff
@@ -203,24 +216,72 @@ export class SpacesRepository implements ISpacesRepository {
...
+    );
+  }
 
-    const row = result.raw[0] as Pick<Space, 'id' | 'uuid'> | undefined;
-    if (!row) {
-      throw new NotFoundException('Workspace not found.');
+  /** Changed `name`/`status` fields of a space update, or `null` for a no-op. */
+  private diffSpaceUpdate(
+    current: Pick<Space, 'name' | 'status'>,
+    updatePayload: QueryDeepPartialEntity<Space>,
```
LucieFaire: if we type it instead as `Partial<Pick<Space, 'name' | 'status'>>,`

then we can simplify the method:
```
  private diffSpaceUpdate(
    current: Pick<Space, 'name' | 'status'>,
    updatePayload: Partial<Pick<Space, 'name' | 'status'>>,
  ): SpaceUpdatedPayload | null {
    const { name, status } = updatePayload;
    const oldFields: SpaceUpdatedPayload['old'] = {};
    const newFields: SpaceUpdatedPayload['new'] = {};

    if (name !== undefined && name !== current.name) {
      oldFields.name = current.name;
      newFields.name = name;
    }
    if (status !== undefined && status !== current.status) {
      oldFields.status = current.status;
      newFields.status = status;
    }

    return Object.keys(newFields).length > 0
      ? { old: oldFields, new: newFields }
      : null;
  }
```

src/modules/spaces/domain/spaces.repository.ts
```diff
@@ -203,24 +216,72 @@ export class SpacesRepository implements ISpacesRepository {
...
+  ): {
+    old: { name?: string; status?: keyof typeof SpaceStatus };
+    new: { name?: string; status?: keyof typeof SpaceStatus };
+  } | null {
```
LucieFaire: can we maybe create a type alias for it? or we can use:
```
type SpaceUpdatedPayload = z.infer<typeof SpaceUpdatedEventSchema>['payload'];
```

src/modules/spaces/routes/members.service.ts
```diff
@@ -93,6 +93,10 @@ export class MembersService {
     await this.membersRepository.renewInvite({
       memberId: id,
       inviteExpiresAt: new Date(Date.now() + this.inviteTtlMs),
+      spaceId: args.spaceId,
+      spaceUuid: space.uuid,
+      targetUserId: args.userId,
+      actorUserId: getAuthenticatedUserIdOrFail(args.authPayload),
```
LucieFaire, line 99: side note: we kinda of do the same in 
```
 await this.assertActiveAdmin({
      spaceId: args.spaceId,
    });
```
its prob a bigger change to switch the signature, so we can keep it as is for now, but something to refactor in the future.

src/modules/spaces/routes/space-audit.service.ts
```diff
@@ -0,0 +1,198 @@
...
+    paginationData: PaginationData;
+    filters: SpaceAuditLogFilters;
+  }): Promise<SpaceAuditLogPage> {
+    const viewerIsActiveAdmin = await this.assertViewer(args);
+
+    const limit = Math.min(args.paginationData.limit, MAX_LIMIT);
+    const offset = Math.max(args.paginationData.offset, 0);
+    // Page links must be derived from the clamped values, not the raw cursor.
+    const normalizedUrl = new URL(args.routeUrl);
+    normalizedUrl.searchParams.set('cursor', `limit=${limit}&offset=${offset}`);
```
LucieFaire: i think we can use `setCursor` here from `PaginationData`
```
  const normalizedUrl = setCursor(
      args.routeUrl,
      new PaginationData(limit, offset),
    );
```

src/modules/spaces/datasources/entities/space-audit-log.entity.db.ts
compojoom: yes

migrations/1781200000000-create-space-audit-log.ts
compojoom: I would keep the strings: the log is append-only and long-lived, so rows should stay self-describing — an int↔enum mapping only lives in app code, and renumbering would silently corrupt history. The storage diff is negligible and `event_type` is not indexed. A native PG enum would also force a DDL migration for every new event type, while the taxonomy is designed to grow additively.

src/modules/spaces/domain/audit/__tests__/space-audit.repository.mock.ts
compojoom: Kept the factory deliberately — it returns a fresh mock per test, so no shared mock state between tests (the shared-const pattern relies on `clearAllMocks`). Happy to conform to the const pattern if you feel strongly.

src/modules/spaces/domain/address-books/address-book-items.repository.ts
compojoom: The `ADDRESS_BOOK_UPSERTED` payload needs `{ address, name }` pairs of the items actually updated (with the new name). The old version returned addresses only, and included all matched rows even when no patch was applied.

src/modules/spaces/domain/audit/entities/space-audit-event.entity.ts
compojoom: Agreed — dropped the lowercase transform; payloads now store the checksummed form straight from `AddressSchema`. Done in dc066d76.

src/modules/spaces/domain/audit/space-audit.repository.ts
compojoom: Adopted in dc066d76 (with boolean conditions so TS accepts the spreads).

src/modules/spaces/domain/space-safes.repository.ts
compojoom: Internal tables still FK on the numeric `space_id` — routes resolve uuid→id at the boundary (the `findIdByIdOrUuid` fallback). The audit row stores both because it must survive space deletion, and the in-transaction lookup doubles as an existence check. Skipping it would mean `assertAdmin` returning the space and threading it through the services — same bucket as your members.service comment; I would take both in one follow-up refactor.

src/modules/spaces/routes/entities/space-audit-log.dto.entity.ts
compojoom: It is always present in the response (explicitly `null` when the event has no target), so `?:` would mis-describe the wire shape. The decorator was the actual bug though — switched to `@ApiProperty({ nullable: true })` (required-but-nullable) in dc066d76.

src/modules/spaces/domain/spaces.repository.ts
compojoom: Adopted as suggested in dc066d76 — and good catch: the interface already declared the narrow type, only the impl was widening it to `QueryDeepPartialEntity`.

src/modules/spaces/domain/spaces.repository.ts
compojoom: Done — added `SpaceUpdatedPayload` derived from `SpaceUpdatedEventSchema` and used it in the diff helper (dc066d76).

src/modules/spaces/routes/members.service.ts
compojoom: Agreed — I would fold this into the same follow-up refactor as the space-safes lookup (assert utils returning the loaded entity).

src/modules/spaces/routes/space-audit.service.ts
compojoom: Done in dc066d76 — `setCursor` was module-private, so I exported it and used it here.

src/modules/spaces/domain/audit/__tests__/space-audit.repository.mock.ts
LucieFaire: is ok :)

LucieFaire (approved):
LGTM 🚀

PR #3165, opened 2026-06-12
feat: accept the proposed contact inline on address book requests, drop private items

compojoom:
## Summary

Spaces members propose address book contacts by sending the full contact (`address`, `name`, `chainIds`) in the create-request body. Previously the endpoint only accepted an `address` and resolved the rest from the member's **private address book item**, which forced clients to upload a private copy first.

Private contacts never shipped to production (staging-only) and are removed entirely: they were invisible to the space audit log, and the upload-then-request dance deleted the client's local copy — a rejected request made the contact vanish.

The web app counterpart (local-storage contacts + single-call request flow) follows in `safe-global/safe-wallet-monorepo`.

```mermaid
flowchart TB
  subgraph Before
    L1[Client contact] -->|1. PUT private item| P1[(user_address_book_items)]
    C1["POST request { address }"] -->|lookup name+chainIds, 404 if missing| P1
    P1 --> R1[(address_book_requests)]
  end
  subgraph After
    L2[Client contact] -->|"POST request { address, name, chainIds }"| R2[(address_book_requests)]
    R2 -->|admin approves| S2[(shared address book + ADDRESS_BOOK_UPSERTED audit event)]
  end
```

## Changes

**Commit 1 — request endpoint accepts the contact inline**

- `CreateAddressBookRequestDto` → `{ address, name, chainIds }`, validated with the shared-book schema (`makeNameSchema` ≤50 chars, `AddressSchema`, `ChainIdSchema[]` with `.min(1)` + dedupe transform) + schema spec
- `createRequest` writes the body directly to `address_book_requests` (the row already stores name/address/chainIds); the private-item lookup and its 404 are gone
- Authorization tightened: `assertActiveMember` — INVITED (not yet accepted) members can no longer create requests
- Anti-abuse (the 500-item private-items cap was the only indirect bound before):
  - max pending requests per (user, space): `spaces.addressBookRequests.maxPending`, default 100 (`SPACES_MAX_PENDING_ADDRESS_BOOK_REQUESTS_PER_USER`)
  - rate limit on POST: `SpacesAddressBookRequestsRateLimitGuard`, `spaces.rateLimit.addressBookRequestCreation`, default 100/600s
- Migration: unique constraint `(space, requester, address)` becomes a **partial unique index on PENDING only** — a rejected address can be requested again; duplicate pending requests still 409
- `approve`/`reject` untouched (they already read from the request row only)

**Commit 2 — remove the private address book module**

- Deletes the `user_address_book_items` entity, domain entity, repository + interface, controller, service, DTOs, e2e spec, and module wiring
- Migration drops the `user_address_book_items` table (pre-prod; staging data abandoned)
- Requests e2e spec rewritten for the new flow: submitted-contact creation, 422 validation, 403 for unauthenticated/INVITED, 409 duplicate pending, re-request after rejection, pending cap, approve→shared-book copy, reject

## Testing

- New DTO schema spec (8 tests), spaces unit suites green (176)
- Requests e2e spec green (17/17) against dockerised Postgres, including the INVITED-403 and re-request-after-rejection paths
- `tsc --noEmit` and biome clean

## Deploy note

The currently deployed staging web app still uses the old flow; between this deploy and the wallet-monorepo PR, "Request to add" / private contacts on staging are non-functional (422/404). Staging-only and expected — land the web PR right after.


src/modules/spaces/routes/entities/address-book-request.dto.entity.ts
```diff
@@ -70,16 +74,34 @@ export class AddressBookRequestsDto {
   public data!: Array<AddressBookRequestItemDto>;
 }
 
-export const CreateAddressBookRequestSchema = z.object({
-  address: AddressSchema,
+export const CreateAddressBookRequestSchema = AddressBookItemSchema.extend({
+  chainIds: z
+    .array(ChainIdSchema)
+    .min(1)
+    .transform((chainIds) => Array.from(new Set(chainIds))),
```
LucieFaire: `overwrite(...`
https://zod.dev/v4?id=overwrite#overwrite

LucieFaire (dismissed):
1 remark, otherwise 👍

src/modules/spaces/datasources/entities/address-book-request.entity.db.ts
```diff
@@ -21,7 +20,11 @@ import {
 import { User } from '@/modules/users/datasources/entities/users.entity.db';
 
 @Entity('address_book_requests')
-@Unique('UQ_ABR_space_requester_address', ['space', 'requestedBy', 'address'])
+@Index(
+  'UQ_ABR_space_requester_address_pending',
+  ['space', 'requestedBy', 'address'],
+  { unique: true, where: 'status = 0' },
```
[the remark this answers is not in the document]
compojoom: you have to draw the line somewhere. The described behavior makes no sense.

src/modules/spaces/routes/entities/address-book-request.dto.entity.ts
compojoom: Good point — switched to `.overwrite(...)` in 2de5086c.

TimDaub (approved):

LucieFaire (approved):

PR #3166, opened 2026-06-15
feat: remove client-facing legacy numeric spaceId; keep UUID only

vmosharova:
## Summary

Resolves: [WA-2570](https://linear.app/safe-global/issue/WA-2570/cleanup-remove-client-facing-legacy-numeric-spaceid-from-cgw)

Removes the client-facing legacy numeric Space identifier so the Space `uuid` is the **only** identifier exposed over the API. The numeric `id` is retained internally as the DB primary key but no longer appears in any response or is accepted in any request path param.

**Why:** A sequential numeric id in responses/params allows sequential scraping / enumeration of spaces (IDOR-style reconnaissance). A random UUID does not. This is a security hardening. The frontend already uses the UUID `spaceId` everywhere, so this is a clean cut with no deprecation window (existing rows already have UUIDs via the `gen_random_uuid()` default).

## Changes

**Responses — numeric id removed (UUID retained):**
- `GetSpaceResponse`, `CreateSpaceResponse`, `UpdateSpaceResponse` — dropped `id`
- `Invitation`, `SpaceAddressBookDto`, `AddressBookRequestsDto`, `SurveyResponseResultDto` — dropped numeric `spaceId`

**Mappers/services — stopped emitting the numeric id:**
- `SpacesService.findSpaces`, `AddressBooksService`, `AddressBookRequestsService`, `SurveysService`, `MembersService`

**Requests — UUID-only path params:**
- Swapped `LegacySpaceIdPipe` → UUID-only `SpaceIdPipe` across all 8 controllers (Spaces, Members, AddressBooks, AddressBookRequests, SpaceAudit, SpaceSafes, Surveys, SpaceCounterfactualSafes)
- A numeric `:spaceId` now returns `400 Invalid space identifier`

**Cleanup:**
- Deleted `LegacySpaceIdPipe` and its module registration
- Removed `SpacesRepository.findIdByIdOrUuid` (interface + impl)
- Dropped `id` from `create`/`update` repository return shapes
- Kept `findUuidById` (now load-bearing: derives the UUID from the internal numeric id for response mapping)

**Tests:**
- Updated unit, integration, and e2e specs to assert UUID-only responses and pass UUIDs as path params
- Replaced the legacy "accepts numeric space id (FE fallback)" test with one asserting a numeric id is rejected with `400`

## ⚠️ 

`spaces.*` API no longer returns the numeric Space `id` nor accepts it in URLs. Clients must use the Space `uuid`. ⚠️ Legacy bookmarked URLs containing a numeric `spaceId` will now 400 — some changes on the FE are following to enable graceful redirect.

Depending: [PR in the frontend](https://github.com/safe-global/safe-wallet-monorepo/pull/8132) that should be release after this one.

### DIFF
--- a/src/modules/spaces/domain/spaces.repository.integration.spec.ts
+++ b/src/modules/spaces/domain/spaces.repository.integration.spec.ts
@@ -1,6 +1,5 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { faker } from '@faker-js/faker';
-import { BadRequestException } from '@nestjs/common';
 import type { ConfigService } from '@nestjs/config';
 import { DataSource, EntityNotFoundError, In } from 'typeorm';
 import type { IConfigurationService } from '@/config/configuration.service.interface';
@@ -220,7 +219,6 @@ describe('SpacesRepository', () => {
       });
 
       expect(space).toEqual({
-        id: expect.any(Number),
         uuid: expect.any(String),
         name,
       });
@@ -289,7 +287,6 @@ describe('SpacesRepository', () => {
           status: spaceStatus,
         }),
       ).resolves.toEqual({
-        id: expect.any(Number),
         uuid: expect.any(String),
         name,
       });
@@ -346,7 +343,6 @@ describe('SpacesRepository', () => {
           status: 'ACTIVE',
         }),
       ).resolves.toEqual({
-        id: expect.any(Number),
         uuid: expect.any(String),
         name: expect.any(String),
       });
@@ -368,7 +364,6 @@ describe('SpacesRepository', () => {
       });
 
       expect(space).toEqual({
-        id: expect.any(Number),
         uuid: expect.any(String),
         name: spaceName,
       });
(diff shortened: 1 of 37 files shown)


vmosharova:
@claude review

src/modules/spaces/domain/spaces.repository.integration.spec.ts
```diff
@@ -463,11 +458,12 @@ describe('SpacesRepository', () => {
         name: name,
         status: spaceStatus,
       });
+      const spaceId = await spacesRepository.findIdByUuid(space.uuid);
```
LucieFaire, line 461: im confused why do we need to do that?

LucieFaire (dismissed):
Hey, so we still use numeric id for DB queries as its indexed we just dont expose it?

vmosharova:
> Hey, so we still use numeric id for DB queries as its indexed we just dont expose it?

@LucieFaire Exactly

src/modules/spaces/domain/spaces.repository.integration.spec.ts
vmosharova: We use it to obtain the internal (numeric) ID based on UUID that we get after space creation, as we get `uuid` only and not an internal `id`.

LucieFaire (approved):

PR #3170, opened 2026-06-16
fix: key rate limits on real client IP via trust proxy

compojoom:
## Summary

Resolves https://linear.app/safe-global/issue/WA-2661/add-trust-proxy-to-cgw

`req.ip` was resolving to the **nginx ingress pod IP**, not the client, because Express `trust proxy` was never set. The shared `RateLimitGuard` (spaces create/update, OIDC auth, address-book ×2), the captcha guard, and the blocklist guard all key on `req.ip`, so every client behind a given ingress pod shared **one** rate-limit bucket. In production the spaces-creation counter ran past 220 against a limit of 10, throttling space creation platform-wide (≈2–3 distinct "client" IPs ever seen — the ingress pods).

This configures `trust proxy` so `req.ip` is the real client.

```mermaid
flowchart LR
  C[Client 203.0.113.7] --> N[nginx ingress 10.0.x.x]
  N --> A[CGW pod]
  subgraph Before["Before — trust proxy unset"]
    A --> B1["req.ip = 10.0.x.x (ingress)<br/>→ one shared bucket"]
  end
  subgraph After["After — trust proxy"]
    A --> B2["req.ip = 203.0.113.7 (XFF)<br/>→ per-client bucket"]
  end
```

## Changes

- **app.provider.ts** — new `configureTrustProxy` (added to `DEFAULT_CONFIGURATION`); accepts a subnet/preset list or a numeric hop count.
- **configuration.ts** — new `express.trustProxy` (env `EXPRESS_TRUST_PROXY`, default `loopback, uniquelocal`; uses `||` so an empty value falls back rather than disabling trust).
- **.env.sample.json** — documents `EXPRESS_TRUST_PROXY` (local default `0` = no proxy).
- **captcha.guard.ts / blocklist.guard.ts** — use `req.ip` instead of the spoofable `getClientIp`; helper + its spec deleted.
- **Tests** — new `app.provider.spec.ts` (resolution, forged-leftmost ignored, hop-count, disabled) and `configuration.spec.ts` (unset/empty/custom); `captcha.guard.spec.ts` updated.


compojoom:
@claude review

LucieFaire (approved):

PR #3156, opened 2026-06-10
fix(spaces): require ACTIVE membership to access space contents

vseehausen:
Resolves [WA-2542](https://linear.app/safe-global/issue/WA-2542/block-workspace-access-for-pending-invite-users)

## Summary
Pending (`INVITED`) members could access a space's contents **before accepting** the invite. Two authorization paths allowed active-or-pending members; both are now restricted to `ACTIVE`:

- **`getSpaceAs`** (shared space address book — `AddressBooksController`): an invited member could read the space address book, so it surfaced in their wallet before joining. **This is the main fix.**
- **`assertMember`** (user/private address book, space safes, counterfactual safes, address-book requests): invited members could read/write these too.

Invited users must now accept first; the invitation itself stays visible (listing still uses the active-or-pending rule).

## Tests
- Repository integration: new test that a **non-expired invited** member is denied the shared address book (the previously-untested hole).
- `space-assert` unit spec updated to ACTIVE-only.
- `user-address-book` e2e: accept the invite in the member helper; add a **403 for an unaccepted invite** and a **403 for a non-member**.
- `address-books` e2e: accept the invite in the 'as member' read tests.

## Notes
- The other `assertMember`-gated e2e specs (`space-safes`, `address-book-requests`, `counterfactual-safes`) still set members up via invite-without-accept and may need the same accept step — relying on CI to confirm.
- Some `user-address-book` / shared-address-book e2e assertions are already red on `main` (the new `spaceUuid` field from #3117, exact-match `.expect()`); not touched here.

PR #3171, opened 2026-06-18
fix: show account/member count to pending members

LucieFaire:
## Summary [WA-2626](https://linear.app/safe-global/issue/WA-2626/invitation-shows-0-count-for-membersaccounts)
Pending (INVITED) members can now see the **member count** and **account (Safe) count** for spaces they've been invited to, while the individual members' data stays hidden.

Commit `1c9323a6` (#3156) hid space contents from pending invitees. As a side effect, the spaces list and single-space endpoints returned a filtered `members` array (own row only) **and** a forced `safeCount: 0` for invitees — so the invitation UI showed `0` members / `0` accounts. Product wants the counts visible on the invitation, just not the member roster itself.

## Changes

- **`get-space.dto.entity.ts`** — added a new required `memberCount` field to `GetSpaceResponse` (active members in the space), alongside the existing `safeCount`. Documented that `members` only contains the caller's own row for pending members.
- **`spaces.service.ts`** (`findSpaces`) —
  - `safeCount` is now always the real count (no longer forced to `0` for invitees).
  - `memberCount: visibleMembers.length` — always the real count of ACTIVE members.
  - `members` is still filtered to the caller's own row for pending invitees (member data stays hidden); active members get the full roster.

## Affected endpoints

Both return `GetSpaceResponse` via `findSpaces`:
- `GET /spaces` — list of the caller's spaces (member of or invited to)
- `GET /spaces/:id` — single space by UUID/legacy id

## Notes

- **Not a breaking change.** `memberCount` is an additive response field, always populated — existing clients ignore it; typed clients pick it up on regeneration. Kept required (not optional) since the server always returns it, same as `safeCount`.
- Out of scope / unchanged: the rest of #3156 stands — pending invitees are still denied space contents (address book, safes, audit log, alias updates), and `GET /spaces/:id/members` still returns only the caller's own row for pending callers.


src/modules/spaces/routes/spaces.service.ts
```diff
@@ -125,7 +126,10 @@ export class SpacesService {
               invitedByName: invitedByNames.get(member.invitedBy),
             }),
         })),
-        safeCount: callerIsActive ? (space.safes?.length ?? 0) : 0,
+        memberCount: space.members.filter(
+          (member) => member.status === 'ACTIVE',
+        ).length,
+        safeCount: space.safes?.length ?? 0,
```
compojoom, line 132: why did we remove the callerIsActive check here?

LucieFaire: Because we want the pending member to see how many safes are in the space, instead of showing 0

LucieFaire: do you have some concern about exposing the count here @compojoom ?

compojoom (approved):

PR #3172, opened 2026-06-18
chore: remove deprecated /create-with-user endpoint

LucieFaire:
## Summary (remainder of [WA-1914](https://linear.app/safe-global/issue/WA-1914/deprecate-create-with-user-endpoint-update-frontend-to-use-spaces-post))
Removes the deprecated `POST /v1/spaces/create-with-user` endpoint.

This endpoint was fully redundant with `POST /v1/spaces` — both handlers called the same `SpacesService.create()` with identical arguments, guards (`AuthGuard` + `SpacesCreationRateLimitGuard`), request DTO (`CreateSpaceDto`), and response type (`CreateSpaceResponse`). The endpoint had been marked `deprecated` in its OpenAPI docs with a note to use `POST /v1/spaces` instead, and the frontend has been updated to use it.

## Changes

- **src/modules/spaces/routes/spaces.controller.ts** — removed the `createWithUser` handler and its associated decorators.
- **src/modules/spaces/routes/spaces.controller.e2e-spec.ts** — removed the `POST /v1/spaces/create-with-user` describe block (5 test cases). Equivalent coverage already exists in the `POST /v1/spaces` suite.


vmosharova (approved):
👍🏻

PR #3173, opened 2026-06-19
fix: tolerate client/server clock skew when validating SiWe message t…

PooyaRaki:
## Summary

  SiWe login intermittently failed with `401 "Invalid message"` immediately after signing, then succeeded after waiting ~1 minute. Root cause: the SiWe message time bounds (`issuedAt`, `expirationTime`, `notBefore`) were compared against the server clock with **zero tolerance**. When
  the client clock runs ahead of the server clock (clock skew), a freshly-signed message's `issuedAt` is in the future relative to the server and is rejected as "Message yet issued" until the server clock catches up.

  This adds a configurable clock-skew allowance, applied symmetrically to the three time bounds, mirroring the leeway used by JWT/OIDC validators. Replay protection and freshness remain enforced independently by the single-use, TTL-bound nonce, so the tolerance does not weaken those
  guarantees.

  ## Changes

  - Convert `SiweMessageSchema` into a factory `buildSiweMessageSchema(clockSkewSeconds?)` that applies the skew to the `issuedAt` / `expirationTime` / `notBefore` checks (follows the existing `page.schema.factory.ts` schema-factory pattern).
  - `SiweRepository` reads `auth.clockSkewSeconds` via `IConfigurationService` and passes it into the schema.
  - Add `auth.clockSkewSeconds` config, sourced from `AUTH_CLOCK_SKEW_SECONDS` (default **30s**).
  - Document `AUTH_CLOCK_SKEW_SECONDS` in `.env.sample.json` and `src/modules/auth/AUTH.md`.
  - Add a `buildSiweMessageSchema` unit spec (the schema had none) covering within/beyond-tolerance boundaries for all three bounds and asserting the configured skew is honored.
  - Update the message-expiry integration test to expire beyond the tolerance window.

### DIFF
--- a/src/modules/auth/AUTH.md
+++ b/src/modules/auth/AUTH.md
@@ -232,6 +232,7 @@ Use `assertAuthenticated(payload)` from `utils/assert-authenticated.utils.ts` to
 - Default max lifetime: **24 hours** (`AUTH_VALIDITY_PERIOD_SECONDS`, default `86400`)
 - SiWe messages may include `expirationTime`; gateway enforces whichever is shorter
 - SiWe messages may include `notBefore`; if present, the JWT `nbf` claim is set and the token is not valid before that time
+- SiWe message time bounds (`issuedAt`, `expirationTime`, `notBefore`) are validated with a tolerated clock skew between client and server (`AUTH_CLOCK_SKEW_SECONDS`, default `30`) to avoid rejecting valid messages when clocks are slightly out of sync
 - Auth0 tokens inherit their `exp` from Auth0; the cookie `maxAge` is derived from the JWT `exp` claim
 - Logout redirect checks `auth_method` from the current token (without re-verifying it) to decide whether to route through Auth0's logout endpoint
 
@@ -253,6 +254,7 @@ Post-login redirects are validated against `AUTH_POST_LOGIN_REDIRECT_URI`:
 | -------------------------------- | -------- | -------------------------------------------------- |
 | `AUTH_NONCE_TTL_SECONDS`         | `300`    | How long a SiWe nonce is valid                     |
 | `AUTH_VALIDITY_PERIOD_SECONDS`   | `86400`  | Max token lifetime                                 |
+| `AUTH_CLOCK_SKEW_SECONDS`        | `30`     | Tolerated client/server clock skew for SiWe time bounds |
 | `AUTH_STATE_TTL_MILLISECONDS`    | `300000` | OIDC state cookie TTL                              |
 | `AUTH_POST_LOGIN_REDIRECT_URI`   | —        | Required. Default redirect after login             |
 | `AUTH_ALLOWED_REDIRECT_DOMAIN`   | —        | Optional. Extra allowed redirect domain (non-prod) |
--- a/src/modules/auth/AUTH.md
+++ b/src/modules/auth/AUTH.md
@@ -250,16 +250,16 @@ Post-login redirects are validated against `AUTH_POST_LOGIN_REDIRECT_URI`:
 
 ## Other Auth Config
 
-| Env var                          | Default  | Description                                        |
-| -------------------------------- | -------- | -------------------------------------------------- |
-| `AUTH_NONCE_TTL_SECONDS`         | `300`    | How long a SiWe nonce is valid                     |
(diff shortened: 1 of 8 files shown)


src/modules/auth/AUTH.md
```diff
@@ -253,6 +254,7 @@ Post-login redirects are validated against `AUTH_POST_LOGIN_REDIRECT_URI`:
 | -------------------------------- | -------- | -------------------------------------------------- |
 | `AUTH_NONCE_TTL_SECONDS`         | `300`    | How long a SiWe nonce is valid                     |
 | `AUTH_VALIDITY_PERIOD_SECONDS`   | `86400`  | Max token lifetime                                 |
+| `AUTH_CLOCK_SKEW_SECONDS`        | `30`     | Tolerated client/server clock skew for SiWe time bounds |
```
nlordell, line 257: nit: table should probably be re-formatted.

src/modules/siwe/domain/entities/siwe-message.entity.ts
```diff
@@ -16,56 +16,72 @@ import { AddressSchema } from '@/validation/entities/schemas/address.schema';
...
-    if (message.notBefore && message.notBefore > now) {
-      ctx.addIssue({
-        code: 'custom',
-        message: 'Message yet valid',
-      });
-    }
+      if (message.notBefore && message.notBefore.getTime() > now + skewMs) {
+        ctx.addIssue({
+          code: 'custom',
+          message: 'Message yet valid',
```
nlordell:
```suggestion
          message: 'Message not yet valid',
```

src/modules/siwe/domain/entities/siwe-message.entity.ts
```diff
@@ -16,56 +16,72 @@ import { AddressSchema } from '@/validation/entities/schemas/address.schema';
...
-    if (!message.issuedAt || message.issuedAt > now) {
-      ctx.addIssue({
-        code: 'custom',
-        message: 'Message yet issued',
-      });
-    }
+      if (!message.issuedAt || message.issuedAt.getTime() > now + skewMs) {
+        ctx.addIssue({
+          code: 'custom',
+          message: 'Message yet issued',
```
compojoom: Shouldn't that be "message not issued yet"?

src/modules/siwe/domain/siwe.repository.ts
```diff
@@ -33,7 +42,8 @@ export class SiweRepository implements ISiweRepository {
     message: string;
     signature: Hex;
   }): Promise<SiweMessage> {
-    const result = SiweMessageSchema.safeParse(args.message);
+    const siweMessageSchema = buildSiweMessageSchema(this.clockSkewSeconds);
```
tongyaosafe: A small suggestion / discussion:
Seems `buildSiweMessageSchema` will run for every login? but `this.clockSkewSeconds` never change, so `siweMessageSchema` will always be the same? Maybe it is worth to move it to constructor?

tongyaosafe (dismissed):

src/modules/siwe/domain/siwe.repository.ts
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/3173/commits/6dcb6b6be764f85c36b83e4a8eb39ea07d8100a1

src/modules/siwe/domain/entities/siwe-message.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3173/commits/6dcb6b6be764f85c36b83e4a8eb39ea07d8100a1

src/modules/siwe/domain/entities/siwe-message.entity.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3173/commits/6dcb6b6be764f85c36b83e4a8eb39ea07d8100a1

src/modules/auth/AUTH.md
PooyaRaki: Done
https://github.com/safe-global/safe-client-gateway/pull/3173/commits/2c7a44508723ff3a62fe683355953e2ca20a38d9

tongyaosafe (approved):

PR #3178, opened 2026-06-22
feat: enable utf-8 in names

vmosharova:
## Summary

Resolves: [WA-2668](https://linear.app/safe-global/issue/WA-2668/allow-unicode-characters-in-spaces-member-and-address-book-names-cgw)

Enables full UTF-8 support in names (address book, spaces, members) so users can save names like `José`, `山田太郎`, `Анна`, `Contact #1`, and `maria@web.com`, while keeping validation safe via a character allowlist and hardening CSV export against formula injection.

Replaces the previous ASCII-only regex with a Unicode-aware policy: names are NFC-normalized, invisible control/format characters are stripped, and only letters, marks, numbers, spaces, and a small punctuation set (`. _ - # @ & ' , ( )`) are permitted — any other visible character (e.g. `= + * / < > " $ { }`, emoji) is rejected with a clear 422. This restores the "no dangerous characters" invariant the old allowlist provided, without limiting names to ASCII.

## Changes

- **Name validation (`name.schema.ts`)**
  - NFC-normalize input and strip invisible `Cc`/`Cf` characters (bidi overrides, zero-width spaces, joiners — Trojan Source / CVE-2021-42574 defense).
  - Allow letters/marks/numbers from any script + ` . _ - # @ & ' , ( )`.
  - Reject any other visible character with `422`:
    `Names can only contain letters, numbers, spaces and the characters . _ - # @ & ' , ( )`.
  - Count length by code point (not UTF-16 units); per-rule error messages for too-short / too-long / disallowed-character.
- **CSV export hardening**
  - Add `escapeCsvFormula` and wire it into `CSV_OPTIONS.cast.string` to neutralize spreadsheet formula-injection triggers (`= + - @`, leading whitespace, full-width variants); numeric columns (`amount`, `payment`) are exempt.
- **DB alignment (`address-book-item.entity.db.ts`)**
  - Use `ADDRESS_BOOK_NAME_MAX_LENGTH` (50) for the `name` column to match the existing migration and the validation schema (previously the entity used 30).
- **Tests**
  - New/updated unit specs for `name.schema`, `sanitizeName`, `escapeCsvFormula`, and `nameBuilder`.
  - E2e: UTF-8 round-trip (multi-script + allowed punctuation, bidi stripping) and a disallowed-character `422` case.

### Notes for reviewers / FE
- Emoji are intentionally **not** supported (dropped ZWJ/ZWNJ exemption) to remove the invisible-joiner spoofing vector.
- The same policy now applies to space and member names (shared `makeNameSchema`).
- The CSV escaper is the load-bearing defense for `= + @` in exported names — any new export format must apply equivalent escaping.


vmosharova:
@claude review

vmosharova:
> Findings

Fixed 1, 4, 5 and 7 in [5c4fd0b](https://github.com/safe-global/safe-client-gateway/pull/3178/commits/5c4fd0b39ecc3b9c518d1ad31c11609939a2c393).

vmosharova:
@claude review

vmosharova:
> A few remaining things worth considering — none blockers.

Fixed 1 and 2 in [1fd2688](https://github.com/safe-global/safe-client-gateway/pull/3178/changes/1fd2688c4a2ec517706f3d7b589995893c09d8c9)

src/domain/common/schemas/name.schema.ts
```diff
@@ -1,25 +1,65 @@
...
-    })
-    .max(args?.maxLength ?? NAME_MAX_LENGTH, {
-      message: `Names must be at most ${args?.maxLength ?? NAME_MAX_LENGTH} characters long`,
-    })
-    .regex(/^[a-zA-Z0-9]+(?:[ ._-][a-zA-Z0-9]+)*$/, {
-      message:
-        'Names must start with a letter or number and can contain alphanumeric characters, spaces, periods, underscores, or hyphens',
-    });
+}): z.ZodType<string> => {
+  const min = args?.minLength ?? NAME_MIN_LENGTH;
```
LucieFaire, line 46: qs: if I do:
```
makeNameSchema({ minLegth: 0 }) 
```
this would allows for empty string - that is expected?

src/domain/common/schemas/name.schema.ts
```diff
@@ -1,25 +1,65 @@
...
+      })
+      // Skip empty (already reported above) so a blank name yields one message.
+      .refine((s) => s.length === 0 || [...s].length >= min, {
+        message: `Names must be at least ${min} character(s) long`,
+      })
+      .refine((s) => [...s].length <= max, {
+        message: `Names must be at most ${max} characters long`,
+      })
+  );
 };
```
LucieFaire, line 91: we could use `superRefine` instead of chain of `refine`s to keep validation in 1 place and avoid duplicated computations.

```
const getNameLength = (name: string): number => [...name].length;

export const makeNameSchema = (args?: {
  minLength?: number;
  maxLength?: number;
}): z.ZodType<string> => {
  const min = args?.minLength ?? NAME_MIN_LENGTH;
  const max = args?.maxLength ?? NAME_MAX_LENGTH;

  return z
    .string()
    .transform(sanitizeName)
    .superRefine((name, ctx) => {
      const isEmpty = name.length === 0;

      if (isEmpty) {
        if (min > 0) {
          ctx.addIssue({
            code: z.ZodIssueCode.custom,
            message: EMPTY_NAME_MESSAGE,
          });
        }

        return;
      }

      if (!ALLOWED_NAME_REGEX.test(name)) {
        ctx.addIssue({
          code: z.ZodIssueCode.custom,
          message: DISALLOWED_CHARACTER_MESSAGE,
        });
      }

      const length = getNameLength(name);

      if (length < min) {
        ctx.addIssue({
          code: z.ZodIssueCode.custom,
          message: `Names must be at least ${min} character(s) long`,
        });
      }

      if (length > max) {
        ctx.addIssue({
          code: z.ZodIssueCode.custom,
          message: `Names must be at most ${max} characters long`,
        });
      }
    });
};

**Review by LucieFaire — COMMENTED** [review]
~~I dont understand why CSV export is part of this ticket? We dont have any spaces-related data there and no user input~~

It ok to have CSV fix here too

File: src/modules/csv-export/csv-utils/escape-csv-formula.ts [hunk 3]
```diff
@@ -0,0 +1,31 @@
...
+  '@',
+  ' ',
+  '\t',
+  '\r',
+  '\n',
+  '＝',
+  '＋',
+  '－',
+  '＠',
+];
```
LucieFaire:
```
const FORMULA_TRIGGERS: ReadonlySet<string> = new Set([
  '=',
  '+',
  '-',
  '@',
  ' ',
  '\t',
  '\r',
  '\n',
  '＝',
  '＋',
  '－',
  '＠',
]);
```
We can make it a `ReadonlySet<string>` so that the lookup uses `.has()` (L27, **need update too**) and (O(1)) — matching how NUMERIC_COLUMNS is already declared (instead of O(n) where n is the length of array 😄)

src/domain/common/schemas/name.schema.ts
vmosharova: Yes, it allows an empty string in this case - it is a flexibility added for the future optional name fields. Right now there is a default length 3 for all names.

src/modules/csv-export/csv-utils/escape-csv-formula.ts
vmosharova: Nice idea, thanks!

vmosharova: Addressed: [40510d4](https://github.com/safe-global/safe-client-gateway/pull/3178/changes/40510d494f6791f13f421d1f3a4ead8844636ac2)

src/domain/common/schemas/name.schema.ts
vmosharova: Addressed: [cb5250e](https://github.com/safe-global/safe-client-gateway/pull/3178/commits/cb5250eedfc7a1d8f111321de8b86cd9569626f8)

LucieFaire (approved):

TimDaub:
If we trust csv-stringify, we should consider using their [formula escape function](https://csv.js.org/stringify/options/escape_formulas/#example). I hope that is applicable advice here. The benefits would be:

- We already trust the library (I guess we chose it wisely)
- Less code to maintain
- Potentially this library (if it is good) does a better job than what we cook in-house

Drawbacks:

- I don't know if their implementation is good or better than ours
- npm Supply chain risk

PR #3189, opened 2026-06-23
refactor: extract UsersRepositoryModule

LucieFaire:
## Summary [WA-1819](https://linear.app/safe-global/issue/WA-1819/tech-debt-extract-usersrepositorymodule-to-decouple-oidcauthmodule)
`OidcAuthModule` imported `UsersModule` just to get `IUsersRepository`. But `UsersModule` also registers `UsersController` and forward-imports `AuthModule` and `SpacesModule`, so enabling `FF_OIDC_AUTH=true` pulled in the entire users/auth/spaces dependency tree regardless of those features' own flags. This breaks the principle of independent feature-flag rollout.

This mirrors the existing `AuthRepositoryModule` extraction done for the same reason (shared `IAuthRepository` between SiWe and OIDC auth).

## Changes

- **New `src/modules/users/domain/users-repository.module.ts`** — standalone module providing and exporting `IUsersRepository`, importing only what `UsersRepository` actually needs (`PostgresDatabaseModuleV2`, `WalletsModule`, `SpaceAuditModule`, TypeORM entities).
- **`src/modules/auth/oidc/oidc-auth.module.ts`** — imports `UsersRepositoryModule` instead of `forwardRef(() => UsersModule)`. The `forwardRef` is no longer needed since `UsersRepositoryModule` has no back-edge to `OidcAuthModule`.
- **`src/modules/users/users.module.ts`** — imports and re-exports `UsersRepositoryModule` so existing consumers resolve `IUsersRepository` unchanged. Drops the now-redundant local `IUsersRepository` provider and the unused `WalletsModule` import.
- **`src/modules/auth/oidc/routes/oidc-auth.controller.integration.spec.ts`** — adds a `UsersRepositoryModule → TestUsersModule` override (with explanatory comment) so `OidcAuthService` receives the mock repository.

## Notes

`IMembersRepository` was deliberately left in `UsersModule`, not moved into `UsersRepositoryModule`. Its implementation depends on `ISpacesRepository`, so moving it would reintroduce the `SpacesModule` coupling this change removes. `OidcAuthService` only needs `IUsersRepository`, which has no such dependency.

### DIFF
--- a/src/modules/auth/oidc/oidc-auth.module.ts
+++ b/src/modules/auth/oidc/oidc-auth.module.ts
@@ -1,13 +1,13 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
-import { forwardRef, Module } from '@nestjs/common';
+import { Module } from '@nestjs/common';
 import { AuthRepositoryModule } from '@/modules/auth/domain/auth-repository.module';
 import { Auth0Module } from '@/modules/auth/oidc/auth0/auth0.module';
 import { OidcAuthController } from '@/modules/auth/oidc/routes/oidc-auth.controller';
 import { OidcAuthService } from '@/modules/auth/oidc/routes/oidc-auth.service';
-import { UsersModule } from '@/modules/users/users.module';
+import { UsersRepositoryModule } from '@/modules/users/domain/users-repository.module';
 
 @Module({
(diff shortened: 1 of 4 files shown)


vmosharova (approved):
👍🏻

PR #3190, opened 2026-06-23
feat: migrate HTTP platform from Express to Fastify

PooyaRaki:
## What

Migrates the gateway's HTTP platform from **Express to Fastify**: swaps `@nestjs/platform-express` → `@nestjs/platform-fastify`, replaces `cookie-parser` with `@fastify/cookie`, and adds `@fastify/static` for static assets.

## Why

Fastify offers lower per-request overhead and a more modern plugin model. This is an internal infrastructure change intended to be behaviour-preserving for clients.

## Key changes

- **Adapter built from config at bootstrap.** Fastify needs `trustProxy` and the JSON `bodyLimit` at construction time — *before* the DI container exists — so `main.ts` reads the raw configuration and builds the adapter directly (`createFastifyAdapterFromConfiguration`). This replaces the old Express `app.set('trust proxy')` / `app.use(json(...))` post-construction setup. - **Request access abstracted** behind `src/routes/common/http/http-request.utils.ts` (`getHeader` / `getRoutePath` / `getRouteParam` / `getRouteUrl`), so guards, interceptors, filters and decorators no longer depend on Express types.
- **Response API migrated:** `res.json()` → `res.send()`, `res.cookie()` → `res.setCookie()`, `res.set()`/`.header()` → `reply.header()`, `res.headersSent` → `reply.sent`, `res.redirect(code, url)` → `res.redirect(url, code)`.
- **Cookie `maxAge` is now in seconds** (Fastify/`@fastify/cookie`) rather than milliseconds (Express) — auth and OIDC flows updated accordingly (`getSecondsUntil`, ms→s conversions).
- **Test harness:** `createTestApplication` / `initTestApplication` call Fastify's `ready()` after `init()` so `supertest` can reach the server; all boot specs (integration + e2e) use them.
- **Removed the `AppProvider` / `DefaultAppProvider` abstraction** — the production bootstrap is inlined into `main.ts`, and `TestAppProvider` is now a standalone test helper (its public `provide()` API is unchanged, so consumer specs were untouched).

  ## Testing

- ✅ Unit suite: **5363 passing**, 0 failing.
- ✅ Production build typecheck (`tsconfig.build.json`): clean.
- ✅ Biome: clean.
- ⏳ **Integration & e2e** require live Redis/Postgres/AMQP and must be verified in CI / locally.

## Reviewer notes

- Please confirm `yarn test:integration` and `yarn test:e2e` are green in CI.
- A quick `yarn start` smoke test is worthwhile to confirm Fastify boots with Swagger (`/api`), static assets, and cookie auth.
- The config namespace is still named `express.*` (`jsonLimit`, `trustProxy`) — left as-is to avoid touching deployed env vars; can be renamed in a follow-up.


PooyaRaki:
@claude review

PooyaRaki:
@claude review

PooyaRaki:
@claude review

PooyaRaki:
All prior review feedback has been addressed. Latest commit (49ef8cb9) covers the remaining items:
  
  - AUTH_STATE_TTL_MILLISECONDS now rejects sub-second values at startup (.min(1_000)) — misconfiguration fails fast instead of silently collapsing the OIDC state cookie TTL.
  - targeted-messaging 204 branch now calls .send() with no body, consistent with HTTP spec and Fastify's own behaviour.
  - Deferred express.* config rename marked TODO(fastify-rename) for greppability.

  Items not actioned (with rationale):
  - getRouteUrl localhost fallback — Express did not throw on a missing Host; it produced http://undefined/…. No regression.
  - parseBodyLimit('1.5b')→1 — deliberate bytes-library parity, documented and tested.
  - Tenderly rawBody HMAC — pre-existing, behaviour-affecting; out of scope for this PR.
  - ServeStaticModule exclude — unchanged from main, not a migration artifact.
  - Test coverage items — all three scenarios are already exercised by existing integration specs.

PooyaRaki:
@claude review

PooyaRaki:
Reduced maxParamLength from `1024` → `256` based on review feedback.
 
The longest real composite id in the API is `multisig_<address>_<safeTxHash>`, which caps out around ~`118` chars. `256` is ~`2`× that, giving a comfortable safety margin for any legitimate value while closing the gap that previously let clients send kilobyte-long path segments — those would flow through ValidationPipe/zod on every matched request, adding gratuitous overhead and creating a mild DoS surface. `1024` was set conservatively during the initial migration; `256` is the right long-term value.

PooyaRaki:
@claude do a final review, all the previous remarks are fixed.

PooyaRaki:
@claude review all the remarks are fixed.

LucieFaire (dismissed):
Lets deploy to staging and see how it behaves

LucieFaire (approved):

PR #3194, opened 2026-06-24
feat(spaces): per-space encryption of audit log payload names

PooyaRaki:
## Summary
  Encrypts audit-log payload names under the owning space's data key, replacing the previous app-wide encryption path. Depends on the spaces+members PR.

  ## Changes
  - `space-audit-event.encryption.ts`: add `collectAuditPayloadNames`/`applyAuditPayloadNames` — collect names, encrypt/decrypt them as one batch, map back in traversal order (per-space crypto is async, so it can't run inside the synchronous name mapper).
  - `SpaceAuditRepository`: on record, encrypt payload names under the space DEK (one unwrap per operation; mint-and-persist if absent); on read, decrypt all of the space's rows in one batch. Now depends on `PerEntityFieldCrypto` instead of the app-wide service.

  ⚠️  `space-audit.repository.integration.spec.ts` still asserts the prior app-wide `enc:v1` format and needs a per-space rewrite before the integration CI job passes.

### DIFF
--- a/src/modules/spaces/domain/audit/entities/space-audit-event.encryption.spec.ts
+++ b/src/modules/spaces/domain/audit/entities/space-audit-event.encryption.spec.ts
@@ -1,0 +1,109 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+
+import type { IFieldEncryptionService } from '@/datasources/encryption/field-encryption.service.interface';
+import {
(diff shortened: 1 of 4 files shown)


PooyaRaki:
I’ll close this PR in favor of this one:
https://github.com/safe-global/safe-client-gateway/pull/3196

PR #3195, opened 2026-06-24
feat(users): per-user email encryption with blind index

PooyaRaki:
## Summary
  Encrypts `users.email` under a per-user data key and adds an HMAC blind index so uniqueness and equality lookups keep working. Independent of the per-space work (different key, different table); depends only on the foundation PR.

  ## Changes
  - Migration: widen `email` to `text`, add `users.encrypted_data_key` and a unique `users.email_index`, drop the old unique index on `email`.
  - `User` entity: per-user key + `email_index` columns; email transformer removed.
  - `UsersRepository`: encrypt the email value under a per-user DEK on create/persist (user-id-bound context); compute the blind index for storage; route every `where:{email}` lookup through the blind index; decrypt email on read. Unique-constraint handling retargeted to
  `idx_users_email_index`.
  - Unit spec updated.
  
  ⚠️  Rollout: while encryption is disabled, `email_index` is null and email uniqueness isn't enforced — enable encryption + run the backfill promptly after this lands (or keep the old index during the transition).
  
### DIFF
--- a/migrations/1781600000000-user-email-field-encryption.ts
+++ b/migrations/1781600000000-user-email-field-encryption.ts
@@ -1,0 +1,41 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { MigrationInterface, QueryRunner } from 'typeorm';
+
+/**
+ * Prepares users.email for deterministic field-level encryption.
+ *
+ * - Widens `email` to `text` to hold ciphertext.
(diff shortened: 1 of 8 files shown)


PooyaRaki:
I’ll close this PR in favor of this one:
https://github.com/safe-global/safe-client-gateway/pull/3196

PR #3197, opened 2026-06-25
feat: make Zerion wallet portfolio cache TTL configurable

vseehausen:
## Summary

- The `/v2/safes` overview caches the Zerion wallet-portfolio response (`ZerionWalletPortfolioApi`) with a previously **hardcoded 10s TTL**.
- Exposes it as **`ZERION_WALLET_PORTFOLIO_CACHE_TTL_SECONDS`** (config `balances.providers.zerion.walletPortfolioCacheTtlSeconds`, default `10`) so it can be tuned on prod without a redeploy.
- Motivation: relieve Zerion rate-limit (429) pressure from the overview path. Raising the TTL reduces cross-poll refresh frequency.

## Note

This is a **tuning knob, not the root-cause fix**. The bigger driver of duplicate Zerion calls is an intra-request race in the overview: a multichain Safe fans out to N concurrent identical portfolio fetches (same address, different chains) that all miss the cache before any write lands. That's addressed separately by a per-address dedupe in `getSafeOverview`.

## Related

- [WA-2752](https://linear.app/safe-global/issue/WA-2752/make-zerion-wallet-portfolio-cache-ttl-configurable) (part of [WA-2749](https://linear.app/safe-global/issue/WA-2749/refactor-balances-and-portfolio-endpoint-at-cgw))


src/config/entities/configuration.ts
```diff
@@ -151,6 +151,10 @@ export default () => ({
           process.env.ZERION_RATE_LIMIT_PER_ADDRESS_CALLS_BY_PERIOD ?? `${0}`,
           10,
         ),
+        walletPortfolioCacheTtlSeconds: Number.parseInt(
```
LucieFaire: nit: could we please rename it to `walletPortfolioTtlSeconds` to be consistent with other TTLs

LucieFaire: and lets default to 60s

src/modules/balances/datasources/zerion-wallet-portfolio-api.service.ts
```diff
@@ -68,6 +68,9 @@ export class ZerionWalletPortfolioApi implements IZerionWalletPortfolioApi {
     this.baseUri = this.configurationService.getOrThrow<string>(
       'balances.providers.zerion.baseUri',
     );
+    this.cacheTtlSeconds = this.configurationService.getOrThrow<number>(
```
LucieFaire, line 71: as we discussed, we want to increase the cache to 1h, but for that we should also add invalidation on tx event. Could you add it as a follow-up please? for now we just increate TTL to a minute w/o invalidation

vseehausen: First we have to make it configurable, right?

LucieFaire: Not exclusively :) one doesnt stop the other 
And it can be tackled together, but for the time saving lets split in 2 Prs and try to roll this one today

src/config/entities/configuration.ts
vseehausen: Renamed in https://github.com/safe-global/safe-client-gateway/pull/3197/changes/d313bbb1759273615ea1bce00fd032535ddd0d1c

I left the default to 10s as before. Now we can configure it.

LucieFaire (approved):

PR #3199, opened 2026-06-25
fix: Zerion /portfolio 429 storm — fetch-once-per-wallet + shared rate limit

compojoom:
## Summary

`GET /v2/safes` (with the per-chain `PORTFOLIO_ENDPOINT` flag on) called Zerion
`GET /v1/wallets/{address}/portfolio` **once per `(chainId, address)` pair**, even though a
portfolio response is **per-wallet and already contains every chain** (`positions_distribution_by_chain`)
and the cache key is address-scoped. A Safe on N chains therefore triggered N identical, unthrottled
calls. With no rate limit, this exceeded Zerion's ~10 RPS account ceiling and produced an ~88% `429`
rate, so multi-chain Safes intermittently dropped out of the overview.

This PR makes two structural fixes plus the honest degradation and observability they require:

- **W1 — fetch once per wallet-identity.** `getSafeOverview` now fetches the portfolio once per
  `(address, isTestnet, currency, trusted)` identity and slices the per-chain value, instead of once
  per chain. `getChain` is deduped per `chainId`. Call count per request drops from O(chains) to
  O(unique wallet-identities).
- **W2 — shared cross-pod rate limit.** A `ZerionRateLimiter` enforces the account-wide budget
  (global + optional per-address sub-budget) via a Redis fixed-window counter, so cluster traffic
  stays under the ceiling and calls **succeed** (warming the existing 10s cache) instead of storming
  `429`s. Every Zerion account consumer — wallet portfolio, balances, positions, and the portfolio
  route — now shares this one budget.
- **Honest degradation.** A failed/over-budget portfolio degrades the wallet to the balances
  repository (a real computed total) — it is **never dropped on a 429** and **never shown a
  fabricated `$0`**. Only a double failure (portfolio *and* balances-repo) drops the Safe.
- **Redis-6-safe counter.** The existing `increment` relied on `EXPIRE … NX` (Redis 7.0+), so on
  Redis 6 the window never resets. A new version-agnostic Lua `INCR`+`EXPIRE` counter (`incrWithTtl`)
  works on both; the existing balances counter is migrated onto it.

## Changes

- `cache.service.interface.ts` / `redis.cache.service.ts` / `fake.cache.service.ts`: add
  `incrWithTtl` (atomic Lua `INCR`+`EXPIRE`, Redis 6/7 safe) + tests.
- `modules/zerion`: new `ZerionRateLimiter` (global + per-address tiers, fail-open on Redis errors)
  exported from `ZerionModule`; `BalancesModule` now imports `ZerionModule`.
- `ZerionBalancesApi` migrated off `increment`; `ZerionPositionsApi`, `ZerionPortfolioApi`, and
  `ZerionWalletPortfolioApi` wired through the shared budget.
- `SafesV2Service.getSafeOverview` restructured (fetch-once-per-identity + dedupe `getChain` +
  honest degradation); new `LogType.PortfolioDegradedServe` for observability.
- New config: `ZERION_RATE_LIMIT_PER_ADDRESS_CALLS_BY_PERIOD` / `_PERIOD_SECONDS` (per-address
  sub-budget, disabled by default).
- New specs: `zerion-rate-limiter.service.spec.ts`, `zerion-wallet-portfolio-api.service.spec.ts`,
  `safes.v2.service.spec.ts`; extended balances, fake-cache, redis-integration, and overview
  integration specs.

> **Config (no code, required for the real ceiling):** code defaults are unchanged and
> safe-by-default. To enforce a hard `<10 RPS` ceiling, set `ZERION_RATE_LIMIT_CALLS_BY_PERIOD` /
> `ZERION_RATE_LIMIT_PERIOD_SECONDS` to a short window (e.g. `period=1, calls≈9`) per environment.

## Affected flows

- `/v2/safes` overview (primary): single-chain unchanged; multi-chain now one portfolio call.
- `/balances`, `/positions`, portfolio route: now share the Zerion budget (can `429` under load;
  previously unbounded).
- Non-Zerion / flag-off chains and empty Safes: unchanged (real `$0` still renders).

## Risks / not checked

- Production Redis 6-vs-7 behaviour: the Lua counter is correct on both; not run against prod Redis.
- Production rate-limit tuning is an ops/env step (defaults are safe placeholders).
- Integration/e2e specs require a live Postgres + Redis and are not in CI; verified here via the unit
  suite (full run green) + analysis of the affected integration assertions.

## Visual summary

```mermaid
flowchart TD
  R["GET /v2/safes — Safe on N chains, P pods"] --> M["getSafeOverview"]
  subgraph Before
    M -.->|"getPortfolio per (chainId,address) — N identical, unthrottled"| Zb["Zerion /portfolio"]
    Zb -.->|"burst > 10 RPS → 429 (88%)"| Drop["throws → Safe dropped"]
  end
  subgraph After
    M -->|"fetch once per (address,isTestnet,currency,trusted)"| RL["ZerionRateLimiter\nglobal + per-address budget"]
    RL -->|"within budget"| Za["Zerion /portfolio (1 call)"]
    Za --> Slice["slice positions_distribution_by_chain → per chain"]
    RL -->|"over budget / failure"| Deg["degrade → balances repo\n(never \$0, never drop)"]
  end
```

## Checklist

- [x] I've added unit tests for the new logic (rate limiter, wallet portfolio datasource, overview service)
- [ ] I've tested the branch on mobile (n/a — backend gateway change)
- [x] I've documented how it affects analytics (new `PortfolioDegradedServe` + `RateLimit{datasource,scope}` logs)

### DIFF
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -80,10 +80,12 @@ export class FakeCacheService implements ICacheService, ICacheReadiness {
   }
 
   incrWithTtl(cacheKey: string, _ttlSeconds: number): Promise<number> {
-    let currentValue: number = this.cache[cacheKey] as number;
-    currentValue = currentValue ? currentValue + 1 : 1;
-    this.cache[cacheKey] = currentValue;
-    return Promise.resolve(currentValue);
+    const currentValue = (this.cache[cacheKey] as number | undefined) ?? 0;
+    const nextValue = currentValue + 1;
+
+    this.cache[cacheKey] = nextValue;
+
+    return Promise.resolve(nextValue);
   }
 
   setCounter(
--- a/src/datasources/cache/__tests__/fake.cache.service.ts
+++ b/src/datasources/cache/__tests__/fake.cache.service.ts
@@ -79,6 +79,13 @@ export class FakeCacheService implements ICacheService, ICacheReadiness {
     return Promise.resolve(currentValue);
   }
 
+  incrWithTtl(cacheKey: string, _ttlSeconds: number): Promise<number> {
+    let currentValue: number = this.cache[cacheKey] as number;
+    currentValue = currentValue ? currentValue + 1 : 1;
+    this.cache[cacheKey] = currentValue;
+    return Promise.resolve(currentValue);
+  }
+
   setCounter(
     key: string,
     value: number,
--- a/src/datasources/cache/cache.service.interface.ts
+++ b/src/datasources/cache/cache.service.interface.ts
@@ -24,15 +24,11 @@ export interface ICacheService {
   ): Promise<number>;
 
   /**
-   * Atomically increments a counter and sets its TTL on first creation.
-   *
-   * Unlike {@link increment} (which relies on `EXPIRE … NX`, a Redis 7.0+
-   * feature), this uses a Lua `INCR`+`EXPIRE` script that works on Redis 6 and
-   * 7 alike. Use this for fixed-window rate limiting so the window reliably
-   * resets regardless of the deployed Redis version.
+   * Atomically increments a counter and sets its TTL on first creation, for
+   * fixed-window rate limiting.
    *
    * @param cacheKey - Counter key (not prefixed, matching {@link increment}).
-   * @param ttlSeconds - Window length in seconds, applied on first increment.
+   * @param ttlSeconds - Window length in seconds, applied on the first increment.
    * @returns The counter value after incrementing.
    */
   incrWithTtl(cacheKey: string, ttlSeconds: number): Promise<number>;
--- a/src/datasources/cache/cache.service.interface.ts
+++ b/src/datasources/cache/cache.service.interface.ts
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
 import type { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 
 export const CacheService = Symbol('ICacheService');
--- a/src/datasources/cache/cache.service.interface.ts
+++ b/src/datasources/cache/cache.service.interface.ts
@@ -22,6 +22,20 @@ export interface ICacheService {
     expireDeviatePercent?: number,
   ): Promise<number>;
 
+  /**
+   * Atomically increments a counter and sets its TTL on first creation.
+   *
+   * Unlike {@link increment} (which relies on `EXPIRE … NX`, a Redis 7.0+
+   * feature), this uses a Lua `INCR`+`EXPIRE` script that works on Redis 6 and
+   * 7 alike. Use this for fixed-window rate limiting so the window reliably
+   * resets regardless of the deployed Redis version.
+   *
+   * @param cacheKey - Counter key (not prefixed, matching {@link increment}).
+   * @param ttlSeconds - Window length in seconds, applied on first increment.
+   * @returns The counter value after incrementing.
+   */
+  incrWithTtl(cacheKey: string, ttlSeconds: number): Promise<number>;
+
   setCounter(
     key: string,
     value: number,
--- a/src/modules/balances/datasources/zerion-balances-api.service.ts
+++ b/src/modules/balances/datasources/zerion-balances-api.service.ts
@@ -45,6 +45,7 @@ import type {
 import type { Chain } from '@/modules/chains/domain/entities/chain.entity';
 import type { Collectible } from '@/modules/collectibles/domain/entities/collectible.entity';
 import { ZerionChainsSchema } from '@/modules/portfolio/datasources/entities/zerion-chain.entity';
+import { ZerionRateLimiter } from '@/modules/zerion/datasources/zerion-rate-limiter.service';
 import { type Raw, rawify } from '@/validation/entities/raw.entity';
 
 export const IZerionBalancesApi = Symbol('IZerionBalancesApi');
@@ -52,15 +53,10 @@ export const IZerionBalancesApi = Symbol('IZerionBalancesApi');
 @Injectable()
 export class ZerionBalancesApi implements IBalancesApi {
   private static readonly COLLECTIBLES_SORTING = '-floor_price';
-  private static readonly RATE_LIMIT_CACHE_KEY_PREFIX = 'zerion';
   private readonly apiKey: string | undefined;
   private readonly baseUri: string;
   private readonly defaultExpirationTimeInSeconds: number;
   private readonly fiatCodes: Array<string>;
-  // Number of seconds for each rate-limit cycle
-  private readonly limitPeriodSeconds: number;
-  // Number of allowed calls on each rate-limit cycle
-  private readonly limitCalls: number;
   // In-memory chain mappings: chainId -> zerion chain name
   private chainMappings: {
     mainnet: Record<string, string>;
@@ -81,6 +77,7 @@ export class ZerionBalancesApi implements IBalancesApi {
     @Inject(IConfigurationService)
     private readonly configurationService: IConfigurationService,
     private readonly httpErrorFactory: HttpErrorFactory,
+    private readonly zerionRateLimiter: ZerionRateLimiter,
   ) {
     this.apiKey = this.configurationService.get<string>(
       'balances.providers.zerion.apiKey',
@@ -94,12 +91,6 @@ export class ZerionBalancesApi implements IBalancesApi {
       );
     this.fiatCodes = this.configurationService.getOrThrow<Array<string>>(
       'balances.providers.zerion.currencies',
-    );
-    this.limitPeriodSeconds = configurationService.getOrThrow(
-      'balances.providers.zerion.limitPeriodSeconds',
-    );
-    this.limitCalls = configurationService.getOrThrow(
-      'balances.providers.zerion.limitCalls',
     );
   }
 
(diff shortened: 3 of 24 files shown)


src/modules/safe/routes/v2/safes.v2.service.ts
```diff
@@ -54,26 +67,38 @@ export class SafesV2Service {
   }): Promise<Array<SafeOverview>> {
     const limitedSafes = args.addresses.slice(0, this.maxOverviews);
 
+    // The Zerion portfolio is per-wallet and contains every chain, so fetch it
+    // once per wallet-identity instead of once per (chainId, address). The
+    // per-chain work below (chain, Safe, queue) stays per entry.
+    const chainsById = await this.resolveChains(limitedSafes);
+    const zerionEnabledById = await this.resolveZerionEligibility(chainsById);
+    const portfoliosByIdentity = await this.fetchPortfoliosOncePerIdentity({
```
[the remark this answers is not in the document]
compojoom: Good catch — fixed in e6bfd68a. `getSafeOverview` now resolves and validates every Safe up front (`resolveSafes`) and only builds wallet-identities (and fetches portfolios) from validated entries. A non-Safe/stale address is dropped before any Zerion call, so it no longer spends a call or consumes the shared budget. The fetch-once-per-identity dedup is preserved. Added a unit test covering 'non-Safe address ⇒ no `getPortfolio` + dropped'.

src/modules/balances/datasources/zerion-balances-api.service.ts
```diff
@@ -202,7 +195,9 @@ export class ZerionBalancesApi implements IBalancesApi {
       return this._buildCollectiblesPage(data.links.next, data.data);
     }
     try {
-      await this._checkRateLimit();
+      await this.zerionRateLimiter.assertWithinBudget({
+        datasource: 'collectibles',
+      });
```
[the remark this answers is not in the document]
compojoom: Fixed in e6bfd68a — `getCollectibles` now rethrows `LimitReachedError` (alongside `ZodError`) so an over-budget request surfaces as 429 instead of being wrapped into a 5xx, matching `getBalances`. Added a unit test asserting the error is rethrown and not passed through `httpErrorFactory.from`.

src/datasources/cache/__tests__/fake.cache.service.ts
```diff
@@ -79,6 +79,13 @@ export class FakeCacheService implements ICacheService, ICacheReadiness {
     return Promise.resolve(currentValue);
   }
 
+  incrWithTtl(cacheKey: string, _ttlSeconds: number): Promise<number> {
+    let currentValue: number = this.cache[cacheKey] as number;
+    currentValue = currentValue ? currentValue + 1 : 1;
+    this.cache[cacheKey] = currentValue;
+    return Promise.resolve(currentValue);
+  }
```
LucieFaire, line 89: hah can make it 1 liner:
```
incrWithTtl(cacheKey: string, _ttlSeconds: number): Promise<number> {
  return Promise.resolve(
    (this.cache[cacheKey] = ((this.cache[cacheKey] as number | undefined) ?? 0) + 1)
  );
}
```
but maybe its a bit too cryptic, there is another version:
```
incrWithTtl(cacheKey: string, _ttlSeconds: number): Promise<number> {
  const currentValue = (this.cache[cacheKey] as number | undefined) ?? 0;
  const nextValue = currentValue + 1;

  this.cache[cacheKey] = nextValue;

  return Promise.resolve(nextValue);
}
```

src/datasources/cache/cache.service.interface.ts
```diff
@@ -22,6 +23,20 @@ export interface ICacheService {
...
 
+  /**
+   * Atomically increments a counter and sets its TTL on first creation.
+   *
+   * Unlike {@link increment} (which relies on `EXPIRE … NX`, a Redis 7.0+
+   * feature), this uses a Lua `INCR`+`EXPIRE` script that works on Redis 6 and
+   * 7 alike. Use this for fixed-window rate limiting so the window reliably
+   * resets regardless of the deployed Redis version.
+   *
+   * @param cacheKey - Counter key (not prefixed, matching {@link increment}).
```
LucieFaire: maybe we can hide the Redis6/7 details and just specify what the method does?

src/modules/safe/routes/v2/safes.v2.service.ts
```diff
@@ -110,6 +145,188 @@ export class SafesV2Service {
...
+          return [chainId, await this.chainsRepository.getChain(chainId)];
+        } catch (error) {
+          this.loggingService.warn(
+            `Error while getting chain ${chainId}: ${asError(error)} `,
+          );
+          return [chainId, null];
+        }
+      }),
+    );
+    return new Map(entries);
```
LucieFaire: nit: we can return a Map right away:
```
    return new Map(
      await Promise.all(
        uniqueChainIds.map(async (chainId): Promise<[string, Chain | null]> => {
          try {
            return [chainId, await this.chainsRepository.getChain(chainId)];
          } catch (error) {
            this.loggingService.warn(
              `Error while getting chain ${chainId}: ${asError(error)} `,
            );
            return [chainId, null];
          }
        }),
      ),
    );
```

src/modules/safe/routes/v2/safes.v2.service.ts
```diff
@@ -110,6 +145,188 @@ export class SafesV2Service {
...
+              this.getZerionChainName(chain)
+            )
+          ) {
+            return [chainId, false];
+          }
+          return [chainId, await this.isPortfolioEndpointFeatureEnabled(chain)];
+        },
+      ),
+    );
+    return new Map(entries);
```
LucieFaire: we can short-cIrcuit here in case Zerion FF is off:
```
    if (!this.zerionBalancesEnabled) {
      return new Map();
    }
    return new Map(
      await Promise.all(
        [...chainsById.entries()].map(
          async ([chainId, chain]): Promise<[string, boolean]> => {
            if (!(chain && this.getZerionChainName(chain))) {
              return [chainId, false];
            }
            return [
              chainId,
              await this.isPortfolioEndpointFeatureEnabled(chain),
            ];
          },
        ),
      ),
    );
```

src/modules/safe/routes/v2/safes.v2.service.ts
```diff
@@ -110,6 +145,188 @@ export class SafesV2Service {
...
+          const portfolio = await this.zerionWalletPortfolioApi.getPortfolio({
+            address,
+            currency,
+            isTestnet,
+            trusted,
+          });
+          results.set(key, { status: 'ok', portfolio });
+        } catch (error) {
+          this.logPortfolioFetchError({ address, error });
+          results.set(key, { status: 'degraded' });
```
LucieFaire: `PortfolioResult` / the `degraded` variant looks droppable — can simplify to a plain portfolio map

The `degraded` status doesn't carry any information that changes behavior. The only reader is `resolveFiatBalance`.

A degraded result and a missing key both fall through to `serveDegradedBalance` — same outcome. And the two things the marker might seem to be "for" are already handled elsewhere:

* Error logging happens once per failed identity in the catch via `logPortfolioFetchError`.
* The degraded-serve metric is emitted per-Safe inside `serveDegradedBalance`.
So neither depends on the stored marker. We could drop the `PortfolioResult` union entirely and make the map `Map<string, ZerionWalletPortfolio>`: the catch just doesn't set anything, and the reader becomes:

```
const portfolio = args.portfoliosByIdentity.get(key);
if (portfolio) {
  const value = this.extractChainFiatBalance(portfolio, chain);
  if (value !== null) return value;
}
return await this.serveDegradedBalance({ ... });
```
That deletes the union, both status literals, and the results.set(key, { status: 'degraded' }) line.

Caveat for the record: this collapses "fetch failed" and "never fetched" into the same undefined. Today that's moot — every Zerion-enabled identity always gets a result, and the early !zerionEnabled return means we never look up the map for disabled chains, so there's no observable difference. The only thing we'd give up is cheaply distinguishing the two off the map later (e.g. a "fetched-but-failed" metric read from the map rather than at the failure site) — and since the failure is already logged + metered where it happens, that seems fine to lose.

src/datasources/cache/__tests__/fake.cache.service.ts
compojoom: Done in fd7c2b42 — went with the explicit `?? 0` / `nextValue` version (kept it over the one-liner for readability).

src/datasources/cache/cache.service.interface.ts
compojoom: Done in fd7c2b42 — trimmed the doc to what the method does (atomic incr + TTL on first creation, for fixed-window rate limiting) and dropped the Redis 6/7 implementation detail.

src/modules/safe/routes/v2/safes.v2.service.ts
compojoom: Done in fd7c2b42 — inlined the `await Promise.all(...)` into `new Map(...)`, dropping the intermediate `entries`.

src/modules/safe/routes/v2/safes.v2.service.ts
compojoom: Done in fd7c2b42 — added an early `return new Map()` when `zerionBalancesEnabled` is off, and dropped the now-redundant flag check from the inner condition.

src/modules/safe/routes/v2/safes.v2.service.ts
compojoom: Done in fd7c2b42 — dropped the `PortfolioResult` union. The map now stores `ZerionWalletPortfolio` directly; a failed fetch is simply a missing key and the reader degrades via `serveDegradedBalance`. The failure is already logged + metered at the failure site, so nothing is lost.

LucieFaire (approved):
LGTM

PR #3200, opened 2026-06-25
[CGW] GTF fee flow: call fee-service /fees + extend /preview with unified fee breakdown

moisses89:
## Summary
Extends GET /v1/chains/{chainId}/safes/{safeAddress}/fees/preview to support the GTF fee flow, alongside the existing relay-fee flow. Which flow applies is decided internally, based on the chain's relayer.type, and is transparent to the caller — the endpoint always returns a fee preview, just shaped differently depending on the fee model.

## Changes

- Add getGtfFees to IFeeServiceApi / FeeServiceApi, calling the fee service's gtf/fees endpoint and caching the response via a new getGtfFeePreviewCacheDir (keyed on chain, safe, tx params, and nonce).
- Add GtfFeesRequest/GtfFeesResponse domain entities and schemas (tx data, fee breakdown with per-token valuation details, pricing context snapshot).
- FeesService#getFeePreview now branches on chain.relayer?.type:
- RELAY_FEE → existing getRelayFees flow.
- GTF → new getGtfFees flow.
- anything else (DAILY_LIMIT, NO_FEE_CAMPAIGN, no relayer) → 400 Bad Request.
- FeePreviewResponse gains fromRelayFees / fromGtfFees factory methods; relayCost and feeBreakdown (plus maxFeeCapUsd) are now optional fields, populated depending on which flow ran.
- Drop the stale pricingContextSnapshot/phase fields from the relay-fees response schema and entity — the non-legacy /relay/fees endpoint doesn't return them (only the deprecated /relay-fees response did).
- Migrate the relay fee request/response contract to the real fee-service shape (URL, request DTO parsing to strip fields outside the endpoint contract).
- Update Swagger docs on FeesController to reflect the dual fee-model behavior.
- Add unit/integration test coverage: FeeServiceApi.getGtfFees, FeesService.getFeePreview branching per relayer type, and controller integration tests for both flows.


moisses89:
@claude review

moisses89:
> **Claude finished @moisses89's task in 3m 17s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/28596897919)
> 
> ### Claude Code Review
> Focused on GTF flow correctness, contract consistency, and coverage.
> 
> #### Findings
> 1. **Inconsistent request stripping between the two flows** — `src/datasources/fee-service-api/fee-service-api.service.ts:100-104` vs `:137-140`
>    `getGtfFees` runs `GtfFeesRequestSchema.parse(...)` to strip fields outside the endpoint contract, but `getRelayFees` still forwards `{...args.request, origin}` unfiltered. If the DTO ever grows a field the relay endpoint doesn't understand (e.g. anything added to `FeePreviewTransactionDto`), it will silently leak through and, worse, become part of the cache key indirectly via `data`. The PR description explicitly calls out DTO stripping — do the same on the relay side by parsing through `TxFeesRequestSchema`.
> 2. **`Origin` default duplicated as a bare string in cache keys** — `src/datasources/cache/cache.router.ts:1124` and `:1170`
>    Both cache-dir builders use `args.origin ?? 'NATIVE'` while the datasource uses `args.request.origin ?? Origin.NATIVE`. If `Origin.NATIVE` ever changes value the cache key silently diverges from what's sent to the service (stale cache hits with wrong `origin`). Import and use `Origin.NATIVE` here too.
> 3. **`FeePreviewTxData` silently drops extra fields via structural typing** — `src/modules/fees/routes/entities/fee-preview-response.entity.ts:66-71`
>    `FeePreviewTxData.fromGtfFees` spreads the full `GtfTxData` (which includes `to`, `value`, `data`, `operation`, `nonce`) into the constructor, and the constructor happens to only pick the intended fields. It works, but the "why does `body.txData.to` come back undefined" behavior is only enforced by the constructor body — not by the type — and the integration test asserts on that side-effect (`body.txData.to` undefined, `body.txData.nonce` undefined). Prefer an explicit `pick`/destructure so the intent is visible and refactor-safe.
> 4. **Fee value representation diverges between flows**
>    `FeePreviewRelayCost.fiatValue` is a string (relay path), while every `*Usd` field on `FeePreviewFeeBreakdown` / `FeePreviewValuationDetail` is a `number` (GTF path, per `GtfFeesResponseSchema`). Clients now have to handle two different numeric encodings on the same endpoint, and JS `number` loses precision above 2^53. Consider stringifying the USD amounts at the CGW boundary, or at minimum documenting the split in the Swagger description on `FeePreviewResponse`.
> 5. **Response chainId not cross-checked against request chainId** — `src/modules/fees/domain/entities/schemas/gtf-fees-response.schema.ts:10`
>    `GtfTxDataSchema.chainId` only enforces "positive integer string". The service could return a `txData.chainId` that differs from the requested chain and it would be cached and returned as-is. Small blast radius, but worth a guard given fee-service is a separate service.
> 6. **BadRequest message leaks internal enum names** — `src/modules/fees/routes/fees.service.ts:75-77`
>    `"only available for chains with RELAY_FEE or GTF relayer"` exposes internal `RelayerType` values in a user-facing 400. Minor, but the previous relay-only error probably didn't do this. Consider a caller-facing phrasing (e.g. "fee preview is not available for this chain").
> 7. **Test coverage gap — cache hit path**
>    `FeeServiceApi.getGtfFees` spec covers the mocked `dataSource.post` call and error forwarding, but not the cache-dir composition. Given both flows have subtly different key inputs (relay includes `fiatCode`, GTF does not), one spec asserting that different `nonce`/`origin`/`to` values produce different `CacheDir` keys would prevent silent cache collisions.
> 8. **Nit — `GtfValuationDetailSchema.tokenAddress` optional, but `symbol`/`amount` required for native transfers**
>    `src/modules/fees/domain/entities/schemas/gtf-fees-response.schema.ts:24-30` — the test `"should validate without optional fields (native token transfer)"` asserts `priceUsd`/`valueUsd` are optional, but if the service ever returns `null` (vs missing) for those, the schema will reject. If the fee-service payload uses `null`, use `.nullable()`.
> 
> Overall the branching is clean and the tests exercise both happy paths and the failure modes on relayer type. Main asks are the parity fix in (1), the cache-key `Origin` fix in (2), and clarifying the mixed number/string USD contract in (4). · [Branch](https://github.com/safe-global/safe-client-gateway/tree/moisesf/pla-1558-cgw-gtf-fee-flow-call-fee-service-fees-extend-preview-with)

1. Inconsistent request stripping — Fixed: parse via schema
2. Origin default duplicated — Fixed: use Origin.NATIVE enum
3. FeePreviewTxData structural typing — Fixed: explicit field picking
4. Fee value representation diverges — Not changed, no action
5. chainId not cross-checked — Skipped, not necessary
6. BadRequest leaks enum names — Fixed: generic error message
7. Test coverage gap (cache) — Fixed: added hash tests

src/modules/fees/domain/entities/schemas/gtf-fees-response.schema.ts
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { z } from 'zod';
+import { Origin } from '@/modules/fees/domain/entities/origin.entity';
+import { PriceSource } from '@/modules/fees/domain/entities/price-source.entity';
+import { Operation } from '@/modules/safe/domain/entities/operation.entity';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { HexSchema } from '@/validation/entities/schemas/hex.schema';
+
+export const GtfTxDataSchema = z.object({
+  chainId: z.coerce.string().regex(/^[1-9]\d*$/),
```
LucieFaire: `NonNegativeNumericStringSchema` wont work here?

src/modules/fees/domain/entities/gtf-fees-request.entity.ts
```diff
@@ -0,0 +1,5 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { z } from 'zod';
+import type { GtfFeesRequestSchema } from '@/modules/fees/domain/entities/schemas/gtf-fees-request.schema';
+
+export type GtfFeesRequest = z.infer<typeof GtfFeesRequestSchema>;
```
LucieFaire, line 10: we usually keep schemas and type inferred in the same file `*.entity.ts`, could we change it in this folder please

src/modules/fees/routes/entities/fee-preview-response.entity.ts
```diff
@@ -97,21 +93,129 @@ export class FeePreviewRelayCost {
...
+  }
+}
+
 export class FeePreviewResponse {
   @ApiProperty({ type: FeePreviewTxData })
   txData: FeePreviewTxData;
 
-  @ApiProperty({ type: FeePreviewRelayCost })
-  relayCost: FeePreviewRelayCost;
+  @ApiProperty({
```
LucieFaire: lets mark optional fields as `@ApiPropertyOptional`

src/modules/fees/domain/entities/schemas/gtf-fees-response.schema.ts
moisses89: NonNegativeNumericStringSchema (^\d+$) allows leading zeros ("01", "007") and "0".

LucieFaire (approved):

falvaradorodriguez (approved):

moisses89:
@LucieFaire 
Updated `NonNegativeNumericStringSchema` to reject leading zeros, with tests added. it's used for `nonce` in the fee request schemas, but also for chainId in the response fee schemas via .pipe() plus a refine to reject 0, so I don't think that could break any other part of CGW. 
I found a [ChainIdSchema](https://github.com/safe-global/safe-client-gateway/blob/main/src/modules/chains/domain/entities/schemas/chain-id.schema.ts) but it's built on `NumericStringSchema`, which allows negatives, decimals, and leading zeros, it also doesn't coerce numbers to strings like our fee response needs. 
There's duplication between our two chainId blocks, and `ChainIdSchema` itself could probably be tightened to cover this case too, but since it's used elsewhere (counterfactual-safes, address-books, portfolio), I'd rather tackle that consolidation in a separate PR than widen the scope of this one.

PR #3214, opened 2026-06-30
feat: strip Blockaid-malicious Safes from the owners endpoints

compojoom:
## Summary

The owners endpoints return every Safe that lists an address as an owner. Because that index is permissionless, address-poisoning campaigns can make look-alike Safes surface in a user's owned-Safes list. This adds a Blockaid bulk address-scan layer that strips Safes classified as `Malicious`, behind the `FF_OWNERS_MALICIOUS_FILTER` flag.

Resolves: https://linear.app/safe-global/issue/WA-2722/block-known-decoys

## Changes

- New `MaliciousAddressScanner`: runs owned Safe addresses through Blockaid's bulk address-scan, caches each verdict per `(chainId, address)`, chunks requests at 100, and fails open (keeps all) on any Blockaid or cache error. Uses a per-call 1.5s timeout to override the SDK's 60s default on the hot path.
- Wired into the V1/V2/V3 owners routes behind `FF_OWNERS_MALICIOUS_FILTER`. V2 strips the union with the existing similarity/creation-date heuristic; V1 and V3 are Blockaid-only.
- Response shape is preserved: `null` chains (upstream fetch failed) and empty arrays pass through untouched, and surviving addresses are unchanged. Flag off ⇒ zero Blockaid calls and prior behaviour.

## How to test

1. Set `FF_OWNERS_MALICIOUS_FILTER=true` with a valid `BLOCKAID_CLIENT_API_KEY`.
2. Call `GET /v2/owners/{ownerAddress}/safes` for an owner holding a Safe that Blockaid classifies as `Malicious` → it is omitted from the response; other Safes remain.
3. Set the flag off and repeat → the unfiltered list is returned (no Blockaid calls).
4. Unit tests: `yarn test src/modules/safe-shield/malicious-address-scan src/modules/owners`.


src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+const hGet = vi.fn();
+const hSet = vi.fn();
+const mockCacheService = { hGet, hSet } as unknown as ICacheService;
+
+const mockLoggingService = {
+  warn: vi.fn(),
+  debug: vi.fn(),
+  error: vi.fn(),
+  info: vi.fn(),
+} as unknown as ILoggingService;
```
LucieFaire: maybe nit, but we usually use MockedObject instead of unknown casting
```
const loggingService = {
  debug: vi.fn(),
  info: vi.fn(),
} as MockedObject<ILoggingService>;
```
I would like us to stay consistent among test setups

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.spec.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  it('chunks batches larger than the max (bot-sized owner)', async () => {
+    const many = Array.from({ length: 150 }, lowerAddress);
+    scan.mockResolvedValue({});
+
+    await scanner.getMaliciousAddresses('1', many);
+
+    expect(scan).toHaveBeenCalledTimes(2);
+    expect(scan.mock.calls[0][0].addresses).toHaveLength(100);
+    expect(scan.mock.calls[1][0].addresses).toHaveLength(50);
```
LucieFaire: another nit: instead of `.mock.calls` please try to use `.haveBeenCalledWith()` its much easier to read

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.ts
```diff
@@ -0,0 +1,125 @@
...
+  LoggingService,
+} from '@/logging/logging.interface';
+import { asError } from '@/logging/utils';
+import { getBlockaidChainName } from '@/modules/safe-shield/threat-analysis/blockaid/blockaid-chain-mapping';
+import { BlockaidClient } from '@/modules/safe-shield/threat-analysis/blockaid/blockaid-client.provider';
+
+const SCAN_TIMEOUT_MS = 1500;
+const MAX_BATCH = 100;
+const CACHE_TTL_SECONDS = 300;
+const SCAN_DOMAIN = 'safe.global';
```
LucieFaire: config service :)

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.ts
```diff
@@ -0,0 +1,125 @@
...
+    @Inject(CacheService) private readonly cacheService: ICacheService,
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
+  ) {}
+
+  async getMaliciousAddresses(
+    chainId: string,
+    addresses: Array<string>,
+  ): Promise<Set<string>> {
+    const chain = getBlockaidChainName(chainId);
+    if (!chain || addresses.length === 0) return new Set();
```
LucieFaire: maybe im too particular, but why not just say `!addresses.length` ? So much leaner

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  private async scan(
+    chain: TransactionScanSupportedChain,
+    addresses: Array<string>,
+  ): Promise<Map<string, boolean>> {
+    try {
+      const response = await this.client.evm.addressBulk.scan(
+        { addresses, chain, metadata: { domain: SCAN_DOMAIN } },
+        { timeout: SCAN_TIMEOUT_MS, maxRetries: 0 },
+      );
```
LucieFaire: why not add this to `BlockaidApi` to keep all blockaid-related logic in one place?

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.spec.ts
compojoom: Done in 1a285314 — all mocks now use `MockedObject`.

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.spec.ts
compojoom: Done in 1a285314 — switched to `toHaveBeenNthCalledWith` with the exact batch slices.

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.ts
compojoom: Done in 1a285314 — timeout, batch size and cache TTL now come from the config service (`safeShield.maliciousAddressScan.*`).

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.ts
compojoom: Done in 1a285314 — biome then simplified it further to `!(chain && addresses.length)`.

src/modules/safe-shield/malicious-address-scan/malicious-address-scanner.service.ts
compojoom: Done in 1a285314 — the bulk scan moved to `BlockaidApi.scanAddressBulk` (owning timeout/retries/domain), and the separate Blockaid client provider is gone.

LucieFaire (approved):

src/modules/safe-shield/threat-analysis/blockaid/blockaid-api.service.ts
```diff
@@ -20,12 +26,18 @@ import {
...
+  private readonly addressScanTimeoutMs: number;
 
   constructor(
     @Inject(LoggingService)
     private readonly loggingService: ILoggingService,
+    @Inject(IConfigurationService)
+    private readonly configurationService: IConfigurationService,
   ) {
     this.blockaidClient = new Blockaid();
+    this.addressScanTimeoutMs = this.configurationService.getOrThrow<number>(
```
LucieFaire, line 38: no need to change it now, but ideally this should also live in scanner-service and be passed in. BlockaidApi is a generic client and should not know about any biz specifics :)

PR #3217, opened 2026-06-30
feat: implement billing service webhook receiver authentication service

LucieFaire:
### Context
The CGW needs to receive webhooks from the billing service and verify their origin. This implements the **"receiver issues the credential it later checks"** model: the CGW mints a long-lived **ES256** (ECDSA P-256) token, provisions it to the billing service, and verifies each incoming `Authorization: Bearer <token>` **statelessly/offline** against its own public key — no JWKS, no callback, no shared secret over the body.

### Verification (runtime)
- **`BillingWebhookAuthGuard`** — thin guard that extracts the bearer token and delegates to `BillingAuthService`.
- **`BillingAuthService.verify`** — two layers: (1) ES256 signature + `iss`/`aud`/`exp` via `IJwtService`; (2) service-token authorization (Zod schema: `roles` contains `SERVICE_ACCESS`, or `data` has `permission_type=SERVICE_ACCESS` + `user_type=SERVICE_USER` + non-empty `service_name`). Failures log at `warn` and return `401`.
- **Endpoint** — placeholder `POST /v1/billing/webhooks` returning `202` (payload handling is a follow-up).
- **Feature-flag gated** — the module is only wired in when `FF_BILLING_WEBHOOK=true`; the app reads its public key lazily. Enabling the flag without a key fails fast at boot. `iss`/`aud` are a single configured identifier.
- The running app **never signs and never calls KMS** — it only verifies, offline, with the public key.

### Provisioning (`yarn generate-token` / `scripts/generate-token.ts`)
Mints the long-lived token. Dual signing mode, same claim shape either way:
- **KMS (production)** — signs via an asymmetric KMS key (`ECC_NIST_P256`); the private key never leaves KMS. Also prints the public-key PEM to configure the verifier.
- **Local PEM (dev/CI)** — signs with `BILLING_WEBHOOK_JWT_PRIVATE_KEY`.
- **Enforced:** when `CGW_ENV=production`, local-PEM signing is refused (KMS required).
- Strict `parseArgs` (unknown flags rejected); `--sub` is charset-validated.

KMS specifics:
- `AwsKmsSignerService` (`src/datasources/kms/`) — `Sign` (ECDSA_SHA_256) + `GetPublicKey`; credentials via IRSA or the default AWS chain; region from `AWS_REGION`.
- `BillingAuthService.mintViaSigner` hand-assembles the JWS around the KMS signature; DER→JOSE conversion via `ecdsa-sig-formatter` (promoted from a transitive to a direct dependency).

### Configuration
| Variable | Used by | Default |
| -- | -- | -- |
| `FF_BILLING_WEBHOOK` | app (enables endpoint) | `false` |
| `BILLING_WEBHOOK_JWT_PUBLIC_KEY` | app (verify) | — |
| `BILLING_WEBHOOK_JWT_ISSUER` | app + CLI (`iss`/`aud`) | `safe-client-gateway` |
| `BILLING_WEBHOOK_JWT_KMS_KEY_ID` | CLI (KMS signing) | — |
| `BILLING_WEBHOOK_JWT_PRIVATE_KEY` | CLI (local signing) | — |

The app needs **no** KMS permissions; `kms:Sign`/`kms:GetPublicKey` go to a dedicated mint principal (CI/bastion), not the app's runtime role. See `src/modules/billing/README.md`.

### Notes
- Webhook **payload** processing is intentionally out of scope (placeholder endpoint).
- The mint CLI is a dev/ops tool; in the deployed image invoke `node dist/scripts/generate-token.js` (not `yarn`) — see the module README for the KMS/IAM provisioning guide.
- Related: PLA-1679 is the opposite direction (CGW → billing outbound credential); keys/tokens are kept separate.

### DIFF
--- a/scripts/generate-token.ts
+++ b/scripts/generate-token.ts
@@ -1,0 +1,99 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+// biome-ignore-all lint/suspicious/noConsole: CLI script — console is the intended output.
+
+/**
+ * Mints the long-lived ES256 service-to-service token that the billing service
+ * presents on each webhook call to the CGW (`Authorization: Bearer <token>`).
+ *
+ * The CGW is the receiver that issues the credential it later verifies: this
+ * script signs with the CGW's private key, and the running app verifies
+ * incoming tokens with the matching public key (BILLING_WEBHOOK_JWT_PUBLIC_KEY).
+ * The private key lives only here, never in the running app's config.
+ *
+ * Usage:
+ *   BILLING_WEBHOOK_JWT_PRIVATE_KEY="$(cat ec-priv.pem)" \
+ *     yarn generate-token --sub billing-service --expires-in 1825
+ *
+ * Environment:
+ *   BILLING_WEBHOOK_JWT_PRIVATE_KEY  ES256 (EC P-256) private key in PEM (required)
+ *   BILLING_WEBHOOK_JWT_ISSUER       the CGW identifier, used as iss and aud
+ *                                    (default: safe-client-gateway)
+ *
+ * The issuer is resolved from the app config, so it matches what the webhook
+ * guard verifies against.
+ *
+ * See src/modules/billing/README.md for the full usage and provisioning guide.
+ */
+import configuration from '@/config/entities/configuration';
+import {
+  BillingAuthService,
+  DEFAULT_BILLING_SERVICE_TOKEN_SUBJECT,
+} from '@/modules/billing/domain/billing-auth.service';
+
+const DEFAULT_EXPIRES_IN_DAYS = 5 * 365; // ~5 years, long-lived service credential
+
+/**
+ * Reads a `--flag value` style argument from process.argv.
+ */
+function getArg(name: string): string | undefined {
+  const index = process.argv.indexOf(`--${name}`);
+  if (index === -1 || index === process.argv.length - 1) {
+    return undefined;
+  }
+  return process.argv[index + 1];
+}
+
+function fail(message: string): never {
+  console.error(`ERROR: ${message}`);
+  process.exit(1);
+}
+
+function main(): void {
+  const privateKey = process.env.BILLING_WEBHOOK_JWT_PRIVATE_KEY?.replace(
+    /\\n/g,
+    '\n',
+  );
+  if (!privateKey) {
+    fail(
+      'BILLING_WEBHOOK_JWT_PRIVATE_KEY environment variable is not defined (ES256 EC P-256 private key, PEM).',
+    );
+  }
+
+  // Resolve the issuer the same way the verifying app does (env or default),
+  // so a minted token always matches what the guard verifies.
+  const { issuer } = configuration().billing.webhook;
+  const subject = getArg('sub') ?? DEFAULT_BILLING_SERVICE_TOKEN_SUBJECT;
+
+  const expiresInArg = getArg('expires-in');
+  const expiresInDays = expiresInArg
+    ? Number.parseInt(expiresInArg, 10)
+    : DEFAULT_EXPIRES_IN_DAYS;
+  if (!Number.isFinite(expiresInDays) || expiresInDays <= 0) {
+    fail(
+      `Invalid --expires-in value: ${expiresInArg}. Must be a positive number of days.`,
+    );
+  }
+
+  let token: string;
+  try {
+    token = BillingAuthService.mint({
+      privateKey,
+      issuer,
+      subject,
+      expiresInDays,
+    });
+  } catch (error) {
+    fail(
+      `Failed to mint token: ${error instanceof Error ? error.message : 'unknown error'}`,
+    );
+  }
+
+  console.log('\n✓ Generated billing-service webhook token:');
+  console.log(`\n${token}`);
+  console.log(`\n• Subject (sub): ${subject}`);
+  console.log(`• Issuer & Audience (iss/aud): ${issuer}`);
+  console.log(`• Algorithm: ES256`);
+  console.log(`• Expires in: ${expiresInDays} days`);
+}
+
+main();
--- a/scripts/generate-token.ts
+++ b/scripts/generate-token.ts
@@ -15,7 +15,8 @@
  *     key). The private key never leaves KMS. Also prints the public key PEM to
  *     configure the verifier. Requires AWS_REGION + credentials (AWS_WEB_IDENTITY_TOKEN_FILE).
  *   - **Local PEM** — set BILLING_WEBHOOK_JWT_PRIVATE_KEY (ES256 EC P-256 PEM).
- *     For dev/CI only; rejected when CGW_ENV=production (KMS is required there).
+ *     For dev/CI only; rejected when CGW_ENV is production or staging (KMS is
+ *     required there).
  *
  * Usage:
  *   # local key
@@ -31,16 +32,26 @@ import { createPublicKey } from 'node:crypto';
 import { parseArgs } from 'node:util';
 import configuration from '@/config/entities/configuration';
 import { AwsKmsSignerService } from '@/datasources/kms/aws-kms-signer.service';
-import {
-  BillingAuthService,
-  DEFAULT_BILLING_SERVICE_TOKEN_SUBJECT,
-} from '@/modules/billing/domain/billing-auth.service';
+import { DEFAULT_BILLING_SERVICE_TOKEN_SUBJECT } from '@/modules/billing/domain/billing-auth.constants';
+import { BillingAuthService } from '@/modules/billing/domain/billing-auth.service';
 
 const DEFAULT_EXPIRES_IN_DAYS = 5 * 365; // ~5 years, long-lived service credential
 
 // Service identifiers only — keeps the `sub`/`service_name` claims clean and
 // avoids control/escape characters reaching the terminal or the token.
 const SUBJECT_PATTERN = /^[a-zA-Z0-9._-]{1,128}$/;
+
+type ClaimsArgs = {
+  issuer: string;
+  subject: string;
+  expiresInDays: number;
+};
+
+type MintResult = {
+  token: string;
+  /** Only set for KMS mode, where the caller can't derive it from local env. */
+  publicKeyPem?: string;
+};
 
 function fail(message: string): never {
   console.error(`ERROR: ${message}`);
@@ -65,19 +76,17 @@ function parseCliArgs(): { sub?: string; expiresIn?: string } {
   }
 }
 
-async function main(): Promise<void> {
-  const { sub, expiresIn } = parseCliArgs();
-
-  const config = configuration();
-  const { issuer, kms } = config.billing.webhook;
-
+function resolveSubject(sub?: string): string {
   const subject = sub ?? DEFAULT_BILLING_SERVICE_TOKEN_SUBJECT;
   if (!SUBJECT_PATTERN.test(subject)) {
     fail(
       `Invalid --sub value: ${JSON.stringify(subject)}. Allowed: letters, digits, '.', '_', '-' (max 128 chars).`,
     );
   }
+  return subject;
+}
 
+function resolveExpiresInDays(expiresIn?: string): number {
   const expiresInDays = expiresIn
     ? Number.parseInt(expiresIn, 10)
     : DEFAULT_EXPIRES_IN_DAYS;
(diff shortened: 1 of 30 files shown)


src/modules/billing/routes/guards/billing-webhook-auth.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Protects the billing-service webhook endpoint with a service-to-service JWT.
+ *
+ * Extracts the `Authorization: Bearer <token>` credential (an HTTP concern) and
+ * delegates verification to {@link BillingAuthService}. Only instantiated
+ * when the `billingWebhook` feature is enabled.
+ */
+@Injectable()
+export class BillingWebhookAuthGuard implements CanActivate {
+  constructor(private readonly tokenService: BillingAuthService) {}
```
moisses89, line 21: `BillingWebhookAuthGuard` injects the concrete `BillingAuthService` class, can we consider following the same pattern as `AuthGuard`, which injects an interface `IAuthRepository`?.

src/modules/billing/routes/guards/billing-webhook-auth.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+    return true;
+  }
+
+  private getBearerToken(request: Request): string | null {
+    const header = request.headers.authorization;
+    if (!header?.startsWith('Bearer ')) {
+      return null;
+    }
+    const token = header.slice('Bearer '.length).trim();
+    return token.length > 0 ? token : null;
```
moisses89, line 44: Just to add coverage, can you add a test for empty key after Bearer?

LucieFaire: [a3de9e6](https://github.com/safe-global/safe-client-gateway/pull/3217/commits/a3de9e611429f3115331c1433ef9740427bf2c91)

src/modules/billing/routes/guards/billing-webhook-auth.guard.ts
LucieFaire: hey, thanks for pointing out. I feel like calling that class `AuthRepository` was maybe not exactly correct. (an there is an interface primarily cause the convention is: repo + interface).
It only calls JWT methods and doesnt have any data access. If we look at `CapthaGuard`, for instance, it also contains a concrete service. I dont expect us to have diff implementation for billing auth, so I would argue service is a correct choice

src/modules/billing/routes/guards/billing-webhook-auth.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+      );
+    }
+
+    // Throws UnauthorizedException if the token is invalid or unauthorized.
+    this.tokenService.verify(token);
+    return true;
+  }
+
+  private getBearerToken(request: Request): string | null {
+    const header = request.headers.authorization;
```
PooyaRaki, line 39: Nit: We could make the header name configurable.
```suggestion
   const AUTH_HEADER_NAME = 'Authorization';
    const header = request.headers[AUTH_HEADER_NAME];
```

src/modules/billing/routes/guards/billing-webhook-auth.guard.ts
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Protects the billing-service webhook endpoint with a service-to-service JWT.
+ *
+ * Extracts the `Authorization: Bearer <token>` credential (an HTTP concern) and
+ * delegates verification to {@link BillingAuthService}. Only instantiated
+ * when the `billingWebhook` feature is enabled.
+ */
+@Injectable()
+export class BillingWebhookAuthGuard implements CanActivate {
```
PooyaRaki, line 20: We could use or extend the current `AuthGuard` and support passing the auth property as a header value. The code here is mostly duplicated.

src/modules/billing/domain/billing-auth.service.ts
```diff
@@ -0,0 +1,136 @@
...
+  constructor(
+    @Inject(IJwtService) private readonly jwtService: IJwtService,
+    @Inject(IConfigurationService)
+    configurationService: IConfigurationService,
+    @Inject(LoggingService) private readonly loggingService: ILoggingService,
+  ) {
+    this.publicKey = configurationService
+      .getOrThrow<string>('billing.webhook.publicKey')
+      // PEM keys provided via env often arrive with escaped newlines.
+      .replace(/\\n/g, '\n');
```
PooyaRaki: Nit: We can handle it in our validator. We should validate this env variable and transform it if it's valid.

src/modules/billing/domain/billing-auth.service.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      );
+    });
+
+    it('honours a custom subject and reflects it in service_name', () => {
+      const { privateKey } = generateKeyPair();
+
+      const token = BillingAuthService.mint({
+        privateKey,
+        issuer: ISSUER,
+        expiresInDays: 365,
```
PooyaRaki: We should use a random number

src/modules/billing/domain/billing-auth.service.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+          sub: 'billing-service-staging',
+          data: { service_name: 'billing-service-staging' },
+        },
+      );
+    });
+
+    it('derives iat/exp from the issued-at time and expiresInDays', () => {
+      const { privateKey } = generateKeyPair();
+      // Whole seconds so the second-based NumericDate comparison is exact.
+      const now = new Date(1_700_000_000_000);
```
PooyaRaki: We should use `faker`.

src/modules/billing/domain/billing-auth.service.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+import { UnauthorizedException } from '@nestjs/common';
+import type { MockedObject } from 'vitest';
+import { FakeConfigurationService } from '@/config/__tests__/fake.configuration.service';
+import { JWT_ES_ALGORITHM } from '@/datasources/jwt/jwt.constants';
+import { jwtClientFactory } from '@/datasources/jwt/jwt.module';
+import { JwtService } from '@/datasources/jwt/jwt.service';
+import type { ILoggingService } from '@/logging/logging.interface';
+import { BillingAuthService } from '@/modules/billing/domain/billing-auth.service';
+
+const ISSUER = 'safe-client-gateway';
```
PooyaRaki: Should be a random string.

src/modules/billing/domain/billing-auth.service.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      const service = createService(publicKey);
+
+      const token = BillingAuthService.mint({
+        privateKey,
+        issuer: ISSUER,
+        expiresInDays: 365,
+      });
+
+      expect(service.verify(token)).toMatchObject({
+        roles: ['SERVICE_ACCESS'],
```
PooyaRaki: It's a bit unclear in this test where the `SERVICE_ACCESS` value is coming from and why, I know that's the default. Can we randomly pass a role when we mint a token and make sure that the role is returned when we decode the token?

src/modules/billing/domain/billing-auth.service.spec.ts
```diff
@@ -0,0 +1,191 @@
...
+      // billing.webhook.publicKey intentionally not set.
+
+      expect(
+        () =>
+          new BillingAuthService(
+            new JwtService(jwtClientFactory(), configurationService),
+            configurationService,
+            loggingService,
+          ),
+      ).toThrow();
```
PooyaRaki: Nit: We should check the exact error here.

PooyaRaki (dismissed):

src/modules/billing/domain/billing-auth.service.ts
LucieFaire: i moved it to `configuration.ts` directly becuase it wont be applied in schema or validator as they are used strictly as a pass/fail gate.

src/modules/billing/routes/guards/billing-webhook-auth.guard.ts
LucieFaire: From my analysis, the similarity is mostly superficial, and merging would couple two security paths. I would argue against coupling these two together

src/modules/billing/domain/billing-auth.service.spec.ts
LucieFaire: The role can't be randomized as it's not a mint input. `buildClaims` specifically sets roles: [SERVICE_ACCESS_ROLE] in order to mint exactly one credential type (a service token) as per contract. Adding a role parameter would contradict that.

LucieFaire: but i changed the test to check for a constant value and not a literal one :)

moisses89 (dismissed):

src/modules/billing/routes/guards/billing-webhook-auth.guard.integration.spec.ts
```diff
@@ -0,0 +1,201 @@
...
+import configuration from '@/config/entities/__tests__/configuration';
+import { JWT_ES_ALGORITHM } from '@/datasources/jwt/jwt.constants';
+import { jwtClientFactory } from '@/datasources/jwt/jwt.module';
+import { TestLoggingModule } from '@/logging/__tests__/test.logging.module';
+import { BillingModule } from '@/modules/billing/billing.module';
+import { BillingAuthService } from '@/modules/billing/domain/billing-auth.service';
+import { BillingController } from '@/modules/billing/routes/billing.controller';
+import { BillingWebhookAuthGuard } from '@/modules/billing/routes/guards/billing-webhook-auth.guard';
+
+const ISSUER = 'safe-client-gateway';
```
PooyaRaki: Nit: We should use `faker` here as well

LucieFaire: [01c4018](https://github.com/safe-global/safe-client-gateway/pull/3217/commits/01c40180571914632cee93438c2392ff7ee20d75)

scripts/generate-token.ts
```diff
@@ -0,0 +1,160 @@
...
+        'expires-in': { type: 'string' },
+      },
+    });
+    return { sub: values.sub, expiresIn: values['expires-in'] };
+  } catch (error) {
+    fail(error instanceof Error ? error.message : 'Invalid arguments');
+  }
+}
+
+async function main(): Promise<void> {
```
PooyaRaki: This method is quite large and very hard to follow, especially with the local private key logic which is handled in an `if` condition. We should extract it to several smaller methods for maintainability.

scripts/generate-token.ts
```diff
@@ -0,0 +1,160 @@
...
+        .export({ format: 'pem', type: 'spki' })
+        .toString();
+    } else {
+      // Local PEM signing mode — dev/CI only; production must sign via KMS.
+      if (config.application.isProduction) {
+        fail(
+          'Production requires KMS signing: set BILLING_WEBHOOK_JWT_KMS_KEY_ID. Minting with a local BILLING_WEBHOOK_JWT_PRIVATE_KEY is not allowed in production.',
+        );
+      }
+      const privateKey = process.env.BILLING_WEBHOOK_JWT_PRIVATE_KEY?.replace(
```
PooyaRaki: Nit: We could also check this in the env variable validator and ensure it doesn't exist in production.

src/datasources/kms/aws-kms-signer.service.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+} from '@aws-sdk/client-kms';
+import { faker } from '@faker-js/faker';
+import { mockClient } from 'aws-sdk-client-mock';
+import { AwsKmsSignerService } from '@/datasources/kms/aws-kms-signer.service';
+
+const kmsMock = mockClient(KMSClient);
+
+describe('AwsKmsSignerService', () => {
+  const keyId = faker.string.uuid();
+  const config = { keyId };
```
PooyaRaki: It's used only once on line 23, we could remove the `const` and pass the object directly to the constructor.

src/datasources/kms/aws-kms-signer.service.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import {
+  GetPublicKeyCommand,
+  KMSClient,
+  SignCommand,
+} from '@aws-sdk/client-kms';
+import { fromTokenFile } from '@aws-sdk/credential-provider-web-identity';
+import type { IKmsSigner } from '@/datasources/kms/kms-signer.interface';
+
+export interface KmsSignerConfig {
```
PooyaRaki: Interface should be extracted to another file.

src/datasources/kms/aws-kms-signer.service.ts
```diff
@@ -0,0 +1,63 @@
...
+  keyId: string;
+  webIdentityTokenFile?: string;
+}
+
+function resolveCredentials(
```
PooyaRaki: Helpers should also be extracted to another file.

src/datasources/kms/aws-kms-signer.service.ts
```diff
@@ -0,0 +1,63 @@
...
+  config: KmsSignerConfig,
+): ReturnType<typeof fromTokenFile> | undefined {
+  // EKS provides pod credentials via IRSA (web identity token). Otherwise fall
+  // back to the default AWS provider chain (env keys).
```
PooyaRaki: QQ: Should we use dedicated credentials for this instance instead of the default ones?

src/modules/billing/domain/billing-auth.service.ts
```diff
@@ -0,0 +1,201 @@
...
+} from '@/logging/logging.interface';
+import {
+  type BillingServiceToken,
+  BillingServiceTokenSchema,
+  SERVICE_ACCESS_PERMISSION_TYPE,
+  SERVICE_ACCESS_ROLE,
+  SERVICE_USER_TYPE,
+} from '@/modules/billing/domain/entities/billing-service-token.entity';
+
+export const DEFAULT_BILLING_SERVICE_TOKEN_SUBJECT = 'billing-service';
```
PooyaRaki: const and `type` should be extracted to a separate file.

src/modules/billing/domain/billing-auth.service.ts
```diff
@@ -0,0 +1,201 @@
...
+   *   2. Authorization — confirm it is a service token.
+   */
+  verify(token: string): BillingServiceToken {
+    try {
+      // Layer 1: signature + standard claims.
+      const payload = this.jwtService.verify<BillingServiceToken>(token, {
+        secretOrPrivateKey: this.publicKey,
+        algorithms: [JWT_ES_ALGORITHM],
+        issuer: this.issuer,
+        audience: this.issuer,
```
PooyaRaki, line 72: Shouldn't the issuer be different from the audience? In this case I expect the issuer to be the billing service and the audience to be `CGW`

PooyaRaki:
@claude review

src/datasources/kms/aws-kms-signer.service.ts
LucieFaire: in practise we should not need static params as everything would be in token file. This is just a fallback, that would fail if we invoke the methods. I think we can leave it as is

src/modules/billing/domain/billing-auth.service.ts
LucieFaire: CGW issues the token, provides it to billing-service, so that billing-service can use it to auth in CGW. Thus issuer and audience are both CGW. We have the same pattern for auth0

LucieFaire:
@claude review again

LucieFaire:
Claude review:
Addressed (2,5,6,8), (1,3,4,7) is irrelevant, added extra tests (9-11) and resolved the nits.

PooyaRaki:
@claude review
As @LucieFaire said:

> Addressed (2,5,6,8), (1,3,4,7) is irrelevant, added extra tests (9-11) and resolved the nits.

Check all the previous remarks from yourself and other reviewers.

PooyaRaki (approved):

PR #3218, opened 2026-07-01
feat(safe-apps): resolve SafeApp via origin URL

PooyaRaki:
## Summary 
  Resolve a message/transaction's SafeApp by its **origin URL** instead of a stored `safeAppId`, and centralize `SafeAppInfoMapper` in the `safe-apps` module. Always-on (no feature flag). Part 1/6 of the `feat/migrateToQueueService` split — base for the rest of the stack, mergeable on its own.
  
  ## Changes 
  - Move `SafeAppInfoMapper` from `transactions/routes/mappers/common/` to `safe-apps/mappers/`; provide and export it from `SafeAppsModule`, drop the old provider from `TransactionsModule`.
  - Change `mapSafeAppInfo` to `(chainId, origin, safeTxHash)` and resolve the SafeApp by origin URL.
  - Add `id` to `SafeAppInfo`.
  - Expose `safeAppInfo` on the messages route entity; keep `logoUri` / `name` / `safeAppId` as deprecated mirror fields.
  - Update `message-mapper` and the `human-description` / `multisig-transaction` / `multisig-transaction-details` mappers for the new path and signature.
  - Add `queue/helpers/origin.helper` (standalone origin parser the resolution depends on).

### DIFF
--- a/src/modules/messages/routes/entities/message.entity.ts
+++ b/src/modules/messages/routes/entities/message.entity.ts
@@ -8,7 +8,8 @@ import {
 import type { Hash, Hex } from 'viem';
 import { MessageConfirmation } from '@/modules/messages/routes/entities/message-confirmation.entity';
 import { TypedData } from '@/modules/messages/routes/entities/typed-data.entity';
-import { AddressInfo } from '@/routes/common/entities/address-info.entity';
+import { SafeAppInfo } from '@/modules/transactions/routes/entities/safe-app-info.entity';
+import type { AddressInfo } from '@/routes/common/entities/address-info.entity';
 
 export enum MessageStatus {
   NeedsConfirmation = 'NEEDS_CONFIRMATION',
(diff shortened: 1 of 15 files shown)


src/modules/messages/routes/entities/message.entity.ts
```diff
@@ -45,6 +48,11 @@ export class Message {
   preparedSignature: Hex | null;
   @ApiPropertyOptional({ type: String, nullable: true })
   origin: string | null;
+  @ApiPropertyOptional({ type: SafeAppInfo, nullable: true })
+  safeAppInfo: SafeAppInfo | null;
+  /** @deprecated Read `safeAppInfo.id` instead. */
+  @ApiPropertyOptional({ type: Number, nullable: true, deprecated: true })
+  safeAppId: number | null;
```
LucieFaire, line 55: why do you add a deprecated field?

src/modules/queue/helpers/origin.helper.ts
```diff
@@ -0,0 +1,76 @@
...
+} {
+  const parsedOrigin: {
+    originName?: string;
+    originUrl?: string;
+    note?: string;
+  } = {
+    originName: undefined,
+    originUrl: undefined,
+    note: undefined,
+  };
```
LucieFaire, line 27: can it be simplified to:
```
  const parsedOrigin: {
    originName?: string;
    originUrl?: string;
    note?: string;
  } = {};
```
as optional fields will be `undefined` by default already

PooyaRaki:

PR #3220, opened 2026-07-01
feat(queue): add Queue Service client module

PooyaRaki:
## Summary
  Add the Queue Service client module. Self-contained and not yet wired into any repository (zero runtime effect). Part 3/6 of the `feat/migrateToQueueService` split.

  ## Changes
  - Add `src/modules/queue/*`: `IQueueService` interface, `QueueService`, schemas, entities, mappers (transaction + message), helpers, `QueueServiceModule`, and unit tests/mocks.
  - Add the queue auth-header helper.
  - Wire the queue test module into the shared testing module.
  - Remove the vestigial `safeAppId` from the domain `Message` schema (required by the queue message mapper).

### DIFF
--- a/src/datasources/network/auth/queue-auth-headers.helper.spec.ts
+++ b/src/datasources/network/auth/queue-auth-headers.helper.spec.ts
@@ -1,0 +1,87 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { Mocked } from 'vitest';
+import type { IConfigurationService } from '@/config/configuration.service.interface';
+import { getQueueAuthHeaders } from '@/datasources/network/auth/queue-auth-headers.helper';
+
+describe('getQueueAuthHeaders', () => {
+  let mockConfigService: Mocked<IConfigurationService>;
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mockConfigService = {
+      getOrThrow: vi.fn(),
+      get: vi.fn(),
+    } as Mocked<IConfigurationService>;
+  });
+
+  it('should return Authorization header when Queue auth is enabled', () => {
+    const apiKey = 'test-api-key-123';
+    mockConfigService.getOrThrow.mockImplementation((key: string) => {
+      if (key === 'application.isDevelopment') return true;
+      if (key === 'queueService.useVpcUrl') return false;
+      throw new Error(`Unexpected key: ${key}`);
+    });
+    mockConfigService.get.mockReturnValue(apiKey);
+
+    const result = getQueueAuthHeaders(mockConfigService);
+
+    expect(result).toEqual({
+      Authorization: `Bearer ${apiKey}`,
+    });
+    expect(mockConfigService.getOrThrow).toHaveBeenCalledWith(
+      'application.isDevelopment',
+    );
+    expect(mockConfigService.getOrThrow).toHaveBeenCalledWith(
+      'queueService.useVpcUrl',
+    );
+    expect(mockConfigService.get).toHaveBeenCalledWith('queueService.apiKey');
+  });
+
+  it.each([
+    {
+      description: 'not in development mode',
+      isDevelopment: false,
+      useVpcUrl: false,
+      apiKey: 'test-key',
+    },
+    {
+      description: 'useVpcUrl is true',
+      isDevelopment: true,
+      useVpcUrl: true,
+      apiKey: 'test-key',
+    },
+    {
+      description: 'API key is undefined',
+      isDevelopment: true,
+      useVpcUrl: false,
+      apiKey: undefined,
+    },
+    {
+      description: 'API key is empty string',
+      isDevelopment: true,
+      useVpcUrl: false,
+      apiKey: '',
+    },
+    {
+      description: 'in production with VPC URL',
+      isDevelopment: false,
+      useVpcUrl: true,
+      apiKey: 'test-key',
+    },
+  ])('should return undefined when $description', ({
+    isDevelopment,
+    useVpcUrl,
+    apiKey,
+  }) => {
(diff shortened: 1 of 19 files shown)


src/datasources/network/auth/queue-auth-headers.helper.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+describe('getQueueAuthHeaders', () => {
+  let mockConfigService: Mocked<IConfigurationService>;
+
+  beforeEach(() => {
+    vi.resetAllMocks();
+    mockConfigService = {
+      getOrThrow: vi.fn(),
+      get: vi.fn(),
+    } as Mocked<IConfigurationService>;
```
LucieFaire, line 14: nit: what if we use fakeConfigService instead of mocked one?

src/datasources/network/auth/queue-auth-headers.helper.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Returns Queue Service auth headers when running in development
+ * against the public Queue Service.
+ *
+ * @param configurationService - Configuration service used to read settings
+ * @returns An object containing the `Authorization` header (`{ Authorization: `Bearer ${apiKey}` }`)
+ *          when Queue auth is enabled and an API key is configured; otherwise `undefined`.
+ */
+export function getQueueAuthHeaders(
```
LucieFaire, line 12: hm, i wonder if we can have a base function that then is implemented/extended by both queue-service and tx-service helpers as the logic is exactly the same?

src/modules/queue/__tests__/queue.mock.ts
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+
+import type { MockedObject } from 'vitest';
+import type { IQueueService } from '@/modules/queue/queue.interface';
+
+export function createMockQueueService(): MockedObject<IQueueService> {
```
LucieFaire, line 6: could we please rename the file to `queue-service.mock` or something like this

src/modules/queue/__tests__/test.queue.module.ts
```diff
@@ -0,0 +1,35 @@
...
+import { QueueService } from '@/modules/queue/queue.service';
+
+/**
+ * Test module that overrides {@link QueueServiceModule} with mocked dependencies.
+ *
+ * Key points:
+ * - Reuses the same NetworkService mock instance from {@link TestNetworkModule}
+ * - Provides real CacheFirstDataSource & HttpErrorFactory (with mocked NetworkService injected)
+ */
+@Module({
```
LucieFaire, line 21: would also like it to be called `test.queue-service.module.ts` as i still feel like over time it can be confused with an actual queue

src/modules/queue/entities/delegate.entity.ts
```diff
@@ -0,0 +1,24 @@
...
+import { z } from 'zod';
+import { buildPageSchema } from '@/domain/entities/schemas/page.schema.factory';
+import { ChainIdSchema } from '@/modules/chains/domain/entities/schemas/chain-id.schema';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import {
+  NullableAddressSchema,
+  NullableStringSchema,
+} from '@/validation/entities/schemas/nullable.schema';
+
+export type QueueDelegate = z.infer<typeof QueueDelegateSchema>;
```
LucieFaire, line 12: `QueueServiceDelegate` ?

src/modules/queue/mappers/message.mapper.ts
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+
+import type { Message } from '@/modules/messages/domain/entities/message.entity';
+import type { QueueMessage } from '@/modules/queue/entities/message.entity';
+import { buildOrigin } from '@/modules/queue/helpers/origin.helper';
+
+export function mapQueueToMessage(msg: QueueMessage): Message {
```
LucieFaire, line 7: a bit of a weird name

src/modules/queue/queue.service.ts
```diff
@@ -0,0 +1,555 @@
...
+    chainId: string;
+    safeAddress: Address;
+    proposeTransactionDto: ProposeTransactionDto;
+  }): Promise<unknown> {
+    try {
+      const dto = args.proposeTransactionDto;
+      const { originName, originUrl, note } = parseOrigin(dto.origin);
+
+      const url = `${this.baseUri}/api/v1/multisig-transactions`;
+      const { data } = await this.networkService.post<unknown>({
```
LucieFaire, line 77: any way we could avoid `unknown` type?

src/modules/queue/queue.service.ts
```diff
@@ -0,0 +1,555 @@
...
+  }
+
+  async postConfirmation(args: {
+    chainId: string;
+    safeTxHash: string;
+    signature: string;
+  }): Promise<unknown> {
+    try {
+      const url = `${this.baseUri}/api/v1/multisig-transactions/${encodeURIComponent(args.safeTxHash)}/signatures`;
+      const { data } = await this.networkService.post<unknown>({
```
LucieFaire, line 221: also here, could be nice to not have `unknown` if possible

LucieFaire (commented):
I am still thinking about the whole module name. I know calling the main bizz file `queue-service.service.ts` and the module `queue-service.module.ts` is slightly hideous, but leaving it as just `queue.smth` might also confuse anyone who looks at the codebase in the future. Can we maybe call it something else completely? Like queued-txs or smth

PooyaRaki:

PR #3221, opened 2026-07-01
feat(safe): route multisig transactions through Queue Service

PooyaRaki:
## Summary
  Route multisig transactions through the Queue Service when `FF_QUEUE_SERVICE` is enabled (default off → behavior unchanged). Part 4/6 of the `feat/migrateToQueueService` split.

  ## Changes
  - `SafeRepository`: when the flag is on, serve multisig transaction reads/writes and queue listings from the Queue Service (queue-first), with origin overlay and cache invalidation; otherwise fall back to the Transaction Service.
  - Import `QueueServiceModule` into `SafeModule`.
  - Cap `propose-transaction` origin length (`MAX_ORIGIN_DTO_LENGTH = 2048`).

### DIFF
--- a/src/modules/safe/domain/safe.repository.interface.ts
+++ b/src/modules/safe/domain/safe.repository.interface.ts
@@ -1,22 +1,24 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
+
 import { Module } from '@nestjs/common';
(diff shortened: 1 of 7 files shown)


PooyaRaki:

PR #3222, opened 2026-07-01
feat(delegate): add delegates v3 module with Queue Service routing

PooyaRaki:
## Summary
  Add a delegates **v3** module (update + delete) that routes to the Queue Service when `FF_QUEUE_SERVICE` is enabled, falling back to the Transaction Service v2 endpoints when off. Part 5/6 of the `feat/migrateToQueueService` split.

  ## Changes
  - Add the v3 delegates controller, service, repository, DTOs, schemas, and tests.
  - Route delegate operations to the Queue Service when the flag is on; otherwise use Transaction Service v2.
  - Add `updateDelegateV2` to the Transaction API interface and implementation.
  - Register the v3 module/repository in `DelegateModule`.

### DIFF
--- a/src/domain/interfaces/transaction-api.interface.ts
+++ b/src/domain/interfaces/transaction-api.interface.ts
@@ -77,6 +77,14 @@ export interface ITransactionApi {
     label: string;
   }): Promise<void>;
 
+  updateDelegateV2(args: {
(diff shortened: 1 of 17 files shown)


PooyaRaki:

PR #3223, opened 2026-07-01
feat(queue-service): Integrate queue service

PooyaRaki:
## Summary
  Route messages through the Queue Service when `FF_QUEUE_SERVICE` is enabled (default off → behavior unchanged), and drop the vestigial `safeAppId` from message creation. Part 6/6 of the `feat/migrateToQueueService` split; top of the stack, merge last.

  ## Changes 
  - `MessagesRepository`: when the flag is on, serve message reads/writes from the Queue Service with cross-chain filtering; otherwise use the Transaction Service.
  - Import `QueueServiceModule` into `MessagesModule`.
  - Remove `safeAppId` from the create-message path (interface, service, repository); keep the `CreateMessageDto` field as a deprecated no-op.


PooyaRaki:
@claude review

PooyaRaki:
Re: Claude's review

  Went through each point:

  1. getMultisigTransactionsBatch swallowing chunk failures — fixed. The comment claimed failed chunks were "detected and logged" by the caller, but nothing actually logged them — a rejected chunk was silently mapped to [] and merged in. Added LoggingService injection to QueueService and
  a LogType.QueueServiceBatchChunkError warn log (with chainId, the affected safeTxHashes, and the underlying error) for every rejected chunk, so partial-batch failures are now observable instead of silently dropped. Added a spec covering the mixed fulfilled/rejected case.

  2. parseOrigin — fixed. To be precise: it doesn't return a bare string (it returns {originName?, originUrl?, note?}), but the underlying issue was real — name/url/note were destructured straight out of JSON.parse with no check that they were actually strings, so e.g. {"name": 123}
  would silently produce a non-string originName typed as string. Now each field is only kept if typeof x === 'string', otherwise it's dropped to undefined per the declared type.

  3. Deprecated safeAppId still enforcing validation — fixed, in a compat-preserving way. Since the value is fully discarded downstream (confirmed by tracing messages.service.ts → messages.repository.ts — it's never read, never forwarded to the queue/transaction service), rejecting a
  request over a malformed safeAppId serves no purpose except breaking legacy clients. Changed z.number().int().gte(0).nullish().default(null) to add .catch(null), so out-of-range/non-integer values fall back to null instead of failing the whole request — the field stays accepted (compat
  preserved) without us pretending we still validate a value we never look at. Updated the two specs that asserted the old strict-rejection behavior to assert the new fallback-to-null behavior instead.

  4. updateMessageSignature extra network hop on cache miss — not a bug, dismissing. Traced the call path: on a queue-cache miss, dataSource.get fetches from the Queue Service itself (${baseUri}/api/v1/messages/{messageHash}), not the Transaction Service. The Transaction Service is only
  ever called when queueServiceEnabled is false — a separate top-level branch, not a fallback triggered by a cache miss within the queue path. So there's no extra hop being introduced here; it's normal cache-aside latency on the single path that's actually taken.

  All changes covered by existing + new unit tests (queue.service.spec.ts, create-message.dto.schema.spec.ts), typecheck clean on the touched files.

PooyaRaki:
@claude review

PooyaRaki:
Fixed

  - 1 — mapper type leak. Correct. ...msg leaked chainId/originName/originUrl onto an object typed Message; TypeScript misses it because object spread bypasses excess-property checks. Replaced with explicit field picking. (It never reached responses since the route MessageMapper
  reconstructs via new Message(...), but it's a latent footgun worth removing.)
  - 4 — missing unit coverage. Added messages.repository.spec.ts covering the cross-chain path directly: row-dropping, count adjustment (decrement, zero-floor, null passthrough, all-matching no-op) and the getMessageByHash chain-id guard. These were only exercised implicitly through
  integration specs before.
  
  Pushing back 

  - 2 — validate pagination URLs at the schema. next/previous come from the shared BasePageSchema used by every paginated endpoint, and the legacy TX-service path does the same new URL(). The source is a trusted internal service. Adding .url() only for queue messages is an inconsistent
  one-off, and changing it globally is out of scope here — net effect is a 500 becoming a 502 for a case that shouldn't occur. Prefer to leave it.
  - 3 — aggregate metric on filteredOut > 0. There's no metrics/counter infrastructure in the codebase today, so this would mean introducing a whole pattern for a should-never-happen case (the queue is already queried by chainId). Per-row warn logs already exist. Happy to file a separate
  ticket if we want cross-cutting observability, but it doesn't belong in this PR.
  - 5 — safeAppId: null in the TX branch. This is intentional and documented (61e3e9c12 "drop vestigial safeAppId" + a575dc8cb) — the DTO field is kept as a deprecated no-op. Not an accidental regression, so keeping as-is.
  - 6 — silent catch in origin.helper.ts. Parse failure there is expected control flow, not an error — origin is frequently a plain non-JSON string, so logging would be pure noise. It mirrors MultisigTransactionNoteMapper, and the helper is a pure function with no logger injected; adding
  a logging dependency costs more than it's worth.

PooyaRaki:
@claude review

src/modules/safe-queue/safe-queue.service.ts
```diff
@@ -0,0 +1,586 @@
...
+      throw this.httpErrorFactory.from(error);
+    }
+  }
+
+  async getDelegates(args: {
+    chainId: string;
+    safeAddress?: Address;
+    delegate?: Address;
+    delegator?: Address;
+    label?: string;
```
LucieFaire, line 311: what happens to the `label`? It is not used in the network call

src/config/entities/configuration.ts
```diff
@@ -786,6 +787,12 @@ export default () => ({
       ),
     },
   },
+  queueService: {
```
LucieFaire: I am still convinced we should call it `txQueueService` to not mix with a real queue

src/datasources/circuit-breaker/circuit-breaker.keys.ts
```diff
@@ -10,6 +10,7 @@ export class CircuitBreakerKeys {
   private static readonly SERVICE_PREFIX = {
     TRANSACTION_SERVICE: 'txs-service',
     DATA_DECODER_SERVICE: 'data-decoder-service',
+    QUEUE_SERVICE: 'queue-service',
```
LucieFaire: `tx-queue-service` ?

src/modules/messages/domain/messages.repository.ts
```diff
@@ -1,62 +1,119 @@
...
+    // pages and earlier pages also dropped rows, it can still over-report. The
+    // queue is queried with the chain id, so cross-chain results should be rare
+    // to begin with.
+    const filteredOut = parsed.results.length - results.length;
+    const count =
+      parsed.count === null ? null : Math.max(0, parsed.count - filteredOut);
+    return { ...parsed, count, results };
   }
 
   async createMessage(args: {
```
LucieFaire, line 88: `createMessage` and `updateMessageSignature` post through the queue service but never clear the `getQueueMessagesBySafe / getQueueMessageByHash` caches. The analogous transaction path does exactly this — `safe.repository.ts:645-667` clears the queue caches inline with the comment "in case event is not received." Message-cache invalidation currently relies entirely on tx-service webhooks `event-cache.helper.ts:594-612`; if the Queue Service doesn't emit MessageCreated/confirmation events for pending messages (which live only in the queue until execution), a just-signed message or added confirmation won't show up in reads for up to the default cache TTL — and the web app re-fetches immediately after signing. Worth either adding inline clears mirroring the safe repository, or confirming the queue service does emit these webhooks.

src/modules/messages/domain/messages.repository.ts
```diff
@@ -100,28 +160,43 @@ export class MessagesRepository implements IMessagesRepository {
...
+      );
+      return transactionService.postMessageSignature({
+        messageHash: args.messageHash,
+        signature: args.signature,
+      });
+    }
+    return this.queueService.postMessageSignature(args);
   }
 
   async clearMessagesBySafe(args: {
```
LucieFaire, line 168: `clearMessagesBySafe/ByHash` clear only one cache layer, unlike the transaction equivalent.
`messages.repository.ts:175-201` picks the queue or the tx-service cache based on the feature flag. But safe.repository.ts's `clearMultisigTransactions` deliberately clears both layers ("independent layers, so a failure in one must not skip invalidating the other"). With the flag-conditional version, a rollback of FF_QUEUE_SERVICE starts serving tx-service message caches that missed every invalidation while the flag was on (and vice versa during rollout). Suggest mirroring the both-layers pattern.

src/modules/delegate/domain/v3/delegates.v3.repository.interface.ts
```diff
@@ -0,0 +1,66 @@
...
+  imports: [TransactionApiManagerModule, QueueServiceModule],
+  providers: [
+    {
+      provide: IDelegatesV3Repository,
+      useClass: DelegatesV3Repository,
+    },
+  ],
+  exports: [IDelegatesV3Repository],
+})
+export class DelegatesV3RepositoryModule {}
```
LucieFaire, line 66: should we extract module to its own file?

src/modules/delegate/routes/v3/entities/delete-delegate.v3.dto.entity.ts
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import type { Address } from 'viem';
+import type { z } from 'zod';
+import type { DeleteDelegateV3DtoSchema } from '@/modules/delegate/routes/v3/entities/schemas/delete-delegate.v3.dto.schema';
+
+export class DeleteDelegateV3Dto
+  implements z.infer<typeof DeleteDelegateV3DtoSchema>
+{
+  @ApiPropertyOptional({ type: String, nullable: true })
```
LucieFaire, line 10: so its optional and nullable?  something doesnt add up here

src/modules/delegate/routes/v3/entities/delete-delegate.v3.dto.entity.ts
```diff
@@ -0,0 +1,16 @@
...
+  delegator!: Address | null;
+  @ApiPropertyOptional({ type: String, nullable: true })
```
LucieFaire, line 12: same here, if its optional it should be `safe?:...`

src/modules/delegate/routes/v3/entities/update-delegate.v3.dto.entity.ts
```diff
@@ -0,0 +1,20 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { ApiProperty, ApiPropertyOptional } from '@nestjs/swagger';
+import type { Address } from 'viem';
+import type { z } from 'zod';
+import type { UpdateDelegateV3DtoSchema } from '@/modules/delegate/routes/v3/entities/schemas/update-delegate.v3.dto.schema';
+
+export class UpdateDelegateV3Dto
+  implements z.infer<typeof UpdateDelegateV3DtoSchema>
+{
+  @ApiPropertyOptional({ type: String, nullable: true })
```
LucieFaire, line 10: +1

src/modules/delegate/routes/v3/delegates.v3.controller.ts
```diff
@@ -0,0 +1,213 @@
...
+  @ApiBadRequestResponse({
+    description: 'Invalid signature or unauthorized deletion attempt',
+  })
+  @Delete('chains/:chainId/delegates/:delegateAddress')
+  deleteDelegate(
+    @Param('chainId') chainId: string,
+    @Param('delegateAddress') delegateAddress: Address,
+    @Body(new ValidationPipe(DeleteDelegateV3DtoSchema))
+    deleteDelegateV3Dto: DeleteDelegateV3Dto,
+  ): Promise<unknown> {
```
LucieFaire: why is it unknown?

src/modules/delegate/routes/v3/delegates.v3.service.ts
```diff
@@ -0,0 +1,138 @@
...
+    chainId: string;
+    createDelegateDto: CreateDelegateDto;
+  }): Promise<void> {
+    await this.repository.postDelegate({
+      chainId: args.chainId,
+      safeAddress: args.createDelegateDto.safe,
+      delegate: args.createDelegateDto.delegate,
+      delegator: args.createDelegateDto.delegator,
+      signature: args.createDelegateDto.signature,
+      label: args.createDelegateDto.label,
```
LucieFaire: maybe we can deconstruct it above and pass values, cause this looks a bit heavy

src/modules/messages/routes/entities/message.entity.ts
```diff
@@ -45,6 +48,11 @@ export class Message {
   preparedSignature: Hex | null;
   @ApiPropertyOptional({ type: String, nullable: true })
   origin: string | null;
+  @ApiPropertyOptional({ type: SafeAppInfo, nullable: true })
+  safeAppInfo: SafeAppInfo | null;
+  /** @deprecated Read `safeAppInfo.id` instead. */
+  @ApiPropertyOptional({ type: Number, nullable: true, deprecated: true })
+  safeAppId: number | null;
```
LucieFaire: could we please check the swagger annotations, because nullable is not optional, and i dont see `undefined` allowed on these values anywhere

src/modules/safe-queue/entities/schemas/origin.schema.ts
```diff
@@ -0,0 +1,33 @@
...
+// z.url() in zod 4 hands invalid inputs to chained refines rather than rejecting
+// upfront, so the refine has to guard `new URL()` itself; otherwise the throw
+// escapes `.catch(null)` and a single bad row poisons the whole batch.
+export const OriginUrlSchema = z
+  .url()
+  .max(MAX_ORIGIN_URL_LENGTH)
+  .refine(
+    (value) => {
+      try {
+        return ALLOWED_ORIGIN_URL_PROTOCOLS.has(new URL(value).protocol);
```
LucieFaire, line 24: will this fail on localhost or origin comes from somewhere else?

src/modules/queue/queue.interface.ts
```diff
@@ -0,0 +1,139 @@
...
+import type { Page } from '@/domain/entities/page.entity';
+import type { Delegate } from '@/modules/delegate/domain/entities/delegate.entity';
+import type { QueueMessage } from '@/modules/queue/entities/message.entity';
+import type { QueueMultisigTransactionEntity } from '@/modules/queue/entities/multisig-transaction.entity';
+import type { ProposeTransactionDto } from '@/modules/transactions/domain/entities/propose-transaction.dto.entity';
+import type { Raw } from '@/validation/entities/raw.entity';
+
+export const IQueueService = Symbol('IQueueService');
+
+export interface IQueueService {
```
LucieFaire: ITxQueueService?

src/modules/queue/queue.module.ts
```diff
@@ -0,0 +1,40 @@
...
+      },
+      inject: [FetchClientToken, LoggingService, IConfigurationService],
+    },
+    CacheFirstDataSource,
+    HttpErrorFactory,
+    { provide: IQueueService, useClass: QueueService },
+  ],
+  exports: [IQueueService],
+})
+export class QueueServiceModule {}
```
LucieFaire: TxQueueServiceModule maybe? :)

src/modules/safe-queue/safe-queue.service.ts
```diff
@@ -0,0 +1,586 @@
...
+        chainId: args.chainId,
+        messageHash: args.messageHash,
+      });
+      const url = `${this.baseUri}/api/v1/messages/${encodeURIComponent(args.messageHash)}`;
+      const data = await this.dataSource.get({
+        cacheDir,
+        url,
+        notFoundExpireTimeSeconds: this.defaultNotFoundExpirationTimeSeconds,
+        expireTimeSeconds: this.defaultExpirationTimeInSeconds,
+        networkRequest: {
```
LucieFaire, line 449: just to make sure - there is no `chainId` param expected?

LucieFaire (commented):
I have also run code reviewer skill and here are the findings (some are dupes of what i marked directly in code):
```
[
  {
    "file": "src/modules/safe/domain/safe.repository.ts",
    "line": 634,
    "summary": "deleteTransaction resolves the tx via the tx-service even in queue mode, so queue-proposed pending transactions cannot be deleted",
    "failure_scenario": "FF_QUEUE_SERVICE on: proposeTransaction writes only to the queue service (lines 913-919), so during AMQP propagation the tx exists only there. DELETE /v1/chains/:chainId/transactions/:safeTxHash calls transactionService.getMultisigTransaction (lines 631-637) unconditionally, which 404s (and caches the not-found) before the queueService.deleteTransaction branch at line 639 is reached — the delete always fails for the exact transactions the queue path owns. Use the queue-first resolveMultisigTransaction (line 561) like every other read."
  },
  {
    "file": "src/modules/queue/mappers/transaction.mapper.ts",
    "line": 37,
    "summary": "isSuccessful: !tx.failed coerces an unknown (null) outcome to true, so a failed-but-unreconciled tx renders as Success",
    "failure_scenario": "Queue schema allows failed: z.boolean().nullable() and txHash non-null (multisig-transaction.entity.ts:52,56); _getTransactionQueue (safe.repository.ts:319-327) maps all rows with no txHash guard. A row with txHash set and failed still null maps to isExecuted=true, isSuccessful=true, and MultisigTransactionStatusMapper returns Success for a transaction whose outcome is unknown — or actually failed. Domain contract is isSuccessful=null until known; should be tx.failed === null ? null : !tx.failed."
  },
  {
    "file": "src/modules/safe/domain/safe.repository.ts",
    "line": 208,
    "summary": "addConfirmation in queue mode never clears the queue tx cache its own verification read just primed, so the POST response omits the new signature",
    "failure_scenario": "addConfirmation calls getMultiSigTransaction (line 190), which cache-first-primes {chainId}_queue_multisig_transaction_{hash} with the pre-confirmation entity; queueService.postConfirmation (queue.service.ts:242-262) neither clears nor overwrites that key and its returned updated entity is discarded. TransactionsService.addConfirmation then synchronously re-reads (transactions.service.ts:282-285), hits the just-primed stale entry, and returns details missing the signature; GETs stay stale until TTL or the AMQP-lagged webhook. deleteTransaction clears these keys explicitly (lines 650-660); addConfirmation does not."
  },
  {
    "file": "src/modules/safe/domain/safe.repository.ts",
    "line": 312,
    "summary": "Queue branch of _getTransactionQueue drops the executed:false / nonceGte:safe.nonce / trusted filters the tx-service branch applies",
    "failure_scenario": "The queue request sends only safes/nonce_order/limit/offset (queue.service.ts:226-231) and no client-side filtering follows (lines 319-327); trusted:true is hardcoded (transaction.mapper.ts:48). The schema explicitly permits executed rows (txHash nullable, isExecuted derived from it), so a superseded proposal below the Safe's current nonce — or an executed-but-unreconciled tx — reappears in the queued list, shown as trusted and potentially as 'Next'. Depends on whether the queue-service endpoint filters server-side, which this repo cannot show — verify or filter client-side."
  },
  {
    "file": "src/modules/queue/queue.service.ts",
    "line": 303,
    "summary": "getDelegates accepts and cache-keys `label` but never sends it upstream, so the v3 label filter is a silent no-op in queue mode",
    "failure_scenario": "GET /v3/chains/1/delegates?label=foo with the flag on: the controller/service/repository pass label through to queueService.getDelegates, the cache field includes it (cache.router.ts:286), but the HTTP params (lines 303-310) omit it — the queue service returns ALL delegates, and each distinct label value caches its own copy of the same unfiltered page. The flag-off path (getDelegatesV2) does forward label, so results silently diverge per flag."
  },
  {
    "file": "src/modules/messages/domain/messages.repository.ts",
    "line": 142,
    "summary": "createMessage and updateMessageSignature on the queue path never invalidate the cache-first queue message caches",
    "failure_scenario": "queueService.postMessage/postMessageSignature are bare POSTs with no cache clearing, while getMessagesBySafe/getMessageByHash are cache-first with default TTL. Worse, updateMessageSignature's verification read (line 150) primes {chainId}_queue_message_{hash} milliseconds before the write, guaranteeing an immediate re-GET returns the message without the new confirmation until TTL or the AMQP-lagged webhook. DelegatesV3Repository in this same PR invalidates after every write (delegates.v3.repository.ts:96-97, 122-123, 147-148) — messages should match."
  },
  {
    "file": "src/modules/safe/domain/safe.repository.ts",
    "line": 307,
    "summary": "'-modified' ordering degrades to nonce-desc on the queue path, freezing the queuedTransactionTag clients poll to detect queue changes",
    "failure_scenario": "getTransactionQueueByModified's queue branch sorts by nonce desc (comment at lines 305-310 acknowledges the degradation, no mitigation). getQueuedTransactionTag (safes.service.ts:259-267) reads limit=1 and returns results[0].modified as the SafeState tag; a confirmation added to a lower-nonce queued tx changes another row's modified but not the first row, so the tag never changes and polling clients never refetch — the new confirmation stays invisible until the highest-nonce row happens to change."
  },
  {
    "file": "src/modules/hooks/domain/helpers/event-cache.helper.ts",
    "line": 682,
    "summary": "Delegate webhook events clear only the tx-service delegates cache; nothing event-driven ever clears the new queue_delegates cache",
    "failure_scenario": "NEW/UPDATED/DELETED_DELEGATE events call IDelegatesV2Repository.clearDelegates, which deletes only {chainId}_delegates_{safe}; the {chainId}_queue_delegates_{safe} key (cache.router.ts:272) is cleared only by DelegatesV3Repository's own writes. With the flag on, a delegate created via the still-registered v2 endpoints (delegates.v2.controller.ts:128,164) or any tx-service-side change leaves GET /v3 delegates serving the stale cached page for the full TTL."
  },
  {
    "file": "src/modules/messages/domain/messages.repository.ts",
    "line": 179,
    "summary": "clearMessagesBySafe/ByHash clear only the flag-selected cache layer, so caches primed under the other flag state go stale across a flag flip",
    "failure_scenario": "The strict if/else (lines 175-201, early return at 184/198) clears queue keys when on, tx-service keys when off — while the flag is ON, tx-service message caches are never invalidated by webhooks. Roll the flag back OFF and GET /v1 messages serves the pre-flip tx-service entry, missing every message/confirmation written during the flag-on window, until TTL. safe.repository.ts deliberately clears both layers for exactly this reason (lines 483-486, 516-519, 681-684)."
  },
]
```

And some cleanup:
1. Uncached hot-path batch call — safe.repository.ts:393: bindQueueOrigins uses getMultisigTransactionsBatch, which is a direct network call (no CacheFirstDataSource, unlike its siblings), adding an uncached queue-service round trip to every executed-transactions history page render.

2. Double zod validation — queue.service.ts:111: QueueService parses QueueMultisigTransactionSchema then rawifies, and SafeRepository re-parses the identical data (safe.repository.ts:397, 575), against the codebase's Raw<T> parse-once convention.

3. Dead code — safe.repository.ts:606: the repository method getMultiSigTransactionWithNoCache has zero production callers (only its own spec exercises it) and internally double-fetches the Safe.

4. Cache-key fragmentation — queue.service.ts:217: the cacheDir is built from un-normalized args, so omitted nonceOrder yields a queue_undefined_… field while explicit 'asc' yields queue_asc_… — two cache entries for the same upstream request.

5. Queue-field leak via spread — transaction.mapper.ts:31: ...tx puts queue-only keys (string chainId, notes, originName, …) on the domain MultisigTransaction; the sibling message.mapper explicitly picks fields to avoid exactly this.

src/modules/messages/domain/messages.repository.ts
PooyaRaki: But this is for off-chain messaging, which means the messages are only stored in the service. You would lose any messages stored in the queue service. We use different cache keys for the TX service and the queue service, so changing the feature flag status won’t be an issue.

src/modules/messages/domain/messages.repository.ts
PooyaRaki: It is confirmed and works as expected, We should refactor the code though and take the same approach everywhere but I'd do it in another PR.

src/modules/safe-queue/safe-queue.service.ts
PooyaRaki: True, Please check to docs [here](https://api.safe.global/queue#/SafeMessages/get_safe_message_api_v1_messages__message_hash__get)

PooyaRaki:
@LucieFaire 

## Addressed review feedback

  Pushed a follow-up commit addressing @LucieFaire's review. Summary below.

  ### Fixed
  - **`deleteTransaction` resolved via tx-service unconditionally** — now resolves queue-first via `resolveMultisigTransaction`, so a queue-only pending tx no longer 404s before the delete branch is reached.
  - **`isSuccessful: !tx.failed` coerced `null` → `true`** — now `tx.failed === null ? null : !tx.failed`.
  - **`addConfirmation` never cleared the cache it primes** — now clears the multisig-transaction cache (both layers) after posting.
  - **Delegate webhooks only cleared the v2/tx-service cache** — `EventCacheHelper` now depends on `IDelegatesV3Repository`, whose `clearDelegates` clears both cache layers unconditionally.
  - **`createMessage` / `updateMessageSignature` didn't invalidate queue message caches** — both now invalidate the relevant caches after writing.
  - **`clearMessagesBySafe` / `clearMessagesByHash` cleared only the flag-selected layer** — both now clear both layers, independent of the flag (mirrors the existing pattern in `safe.repository.ts`).
  - **`origin.schema.ts` origin URL protocol** — `http:` is now allowed alongside `https:`; in development the protocol check is skipped entirely.
  - **`message.entity.ts` Swagger annotations** — fields that are always present but nullable (`logoUri`, `name`, `preparedSignature`, `origin`, `safeAppInfo`, `safeAppId`) switched from `@ApiPropertyOptional` to `@ApiProperty({ nullable: true })`.
  - **`delegates.v3.controller.ts`'s `deleteDelegate` returned `unknown`** — traced the chain end-to-end (nothing reads the response body); now `void` across repository/service/controller, matching `postDelegate`/`updateDelegate`.
  - Minor: deconstructed `deleteDelegateV3Dto` fields in `delegates.v3.service.ts`; extracted a shared `clearBothCacheLayers` helper used by both messages and delegates instead of duplicating the "clear both layers, catch-and-log per layer" logic.
  - **Renamed `QueueService`/`IQueueService` (and everything derived from it) to `SafeQueueService`** — including the `queue/` → `safe-queue/` directory, config keys, and env vars (`FF_QUEUE_SERVICE` → `FF_SAFE_QUEUE_SERVICE`, etc.). Went with `SafeQueueService` rather than
  `TransactionQueue`/`TxQueue`, since `transactionQueue` is already an established term in this codebase for something unrelated (`SafeRepository.getTransactionQueue()`, `QueuedItem` — the pending/not-yet-executed-transactions list feature) and reusing it here would've made two unrelated
  concepts look identical. `SafeQueueService` matches the upstream service's own name (its Swagger doc is titled "Safe Queue Service").
 
  ### Investigated, left as-is (with reasoning)
  - **`getDelegates`: `label` param not forwarded upstream.** Confirmed via the queue service's own OpenAPI spec — `GET /api/v1/delegates` has no `label` query param at all, it's simply not supported server-side. Tried client-side filtering as a workaround; reverted because filtering
  after the server already applies `limit`/`offset` corrupts pagination (a filtered page can silently under-fill, and matches outside the current raw page never surface). Not fixable without a queue-service API change.
  - **`_getTransactionQueue` (queue branch) drops `executed`/`nonceGte`/`trusted` filters.** Same root cause and same story — the `/multisig-transactions/queue` endpoint has no such filters server-side. Tried and reverted client-side filtering for the same pagination-correctness reason.
  - **`trusted` / "isTrusted from the downstream service".** Checked the queue service's OpenAPI spec directly — there is no `trusted`/`isTrusted` concept anywhere in its API (no field on the transaction response, no query param). `trusted: true` stays hardcoded in the mapper; there's 
  nothing upstream to read.
  - **`-modified` ordering degrades to nonce-desc.** The queue service can only sort by nonce; already documented in the code as an accepted best-effort limitation. No further fix possible without a queue-service API change.
  - **"Extract module to its own file?" (`delegates.v3.repository.interface.ts`)** — checked first: colocating the `@Module` block in `*.repository.interface.ts` is the established convention across this codebase (same pattern in `delegates.v2`, `safe.repository.interface.ts`). Splitting
  v3 out would be the inconsistent choice, so left as-is.
  - Naming nits (`ITxQueueService`, `TxQueueServiceModule`, `tx-queue-service`) — superseded by the `SafeQueueService` rename above, which addresses the same underlying concern.

PooyaRaki:
## Follow-up: addressed remaining cleanup items

  Pushed a second follow-up commit closing out the last 5 items from the review (the code-reviewer-skill "cleanup" dump that hadn't been looked at yet).

  - **Uncached `getMultisigTransactionsBatch`** — this was the only method in `SafeQueueService` bypassing `CacheFirstDataSource`, adding an uncached round-trip to every executed-transactions history page render. Now cached per chunk, keyed by a sha256 hash of that chunk's `safeTxHashes`
  (`CacheRouter.getSafeQueueMultisigTransactionsBatchCacheDir`).
  - **Double zod validation** — `getMultisigTransaction` pre-validated with `SafeQueueMultisigTransactionSchema.parse` then `rawify`'d the result, and `SafeRepository` parsed it *again*. Removed the internal parse there, so it now returns genuinely unvalidated data — consistent with
  `getMessageByHash`/the tx-service convention (one parse point, at the repository). `getMultisigTransactionsBatch` is different: it validates per-chunk via `safeParse` on purpose, so a single malformed row doesn't take down the whole batch — that has to stay. Instead of hiding that
  under a misleading `Raw<T>`, its return type is now the plain validated array, and the repository trusts it instead of re-parsing.
  - **Dead code** — `getMultiSigTransactionWithNoCache` had zero callers outside its own spec file. Removed it, along with the `noCache` option in `resolveMultisigTransaction`/`fetchTxServiceMultisig` that only existed to support it (now simplified back to no-param calls). Removed 3 spec
  tests that were duplicating existing `getMultiSigTransaction` coverage of the same code paths.
  - **Cache-key fragmentation** — `getTransactionQueue` built the cache key from the raw, un-normalized `nonceOrder` (`undefined` if omitted) but defaulted to `'asc'` for the actual network request afterward, so two functionally identical calls produced two different cache entries. Now
  normalizes before building the cache dir.
  - **Queue-field leak via spread** — `mapSafeQueueToMultisigTransaction`/`mapConfirmation` spread the queue entity onto objects typed as the domain `MultisigTransaction`/`Confirmation`, leaking queue-only fields (`chainId`, `notes`, `originName`, `originUrl`, `txHash` on the
  transaction; `created`/`modified` on confirmations) that TypeScript's excess-property check doesn't catch through a spread. Switched to explicit field-picking, mirroring the same fix already applied to the message mapper.
- Added/updated tests for all five: cache-dir determinism tests for the new batch cache key, a real end-to-end sanitization test moved to where validation now actually lives (`SafeQueueService`, not the repository), a nonceOrder cache-key regression test, and a field-leak regression test
  on the transaction mapper.

src/modules/delegate/domain/v3/delegates.v3.repository.ts
```diff
@@ -0,0 +1,182 @@
...
+      args.chainId,
+    );
+    const page = await transactionService.getDelegatesV2({
+      safeAddress: args.safeAddress,
+      delegate: args.delegate,
+      delegator: args.delegator,
+      label: args.label,
+      limit: args.limit,
+      offset: args.offset,
+    });
```
tmjssz: Can we simpify?
```suggestion
    const page = await transactionService.getDelegatesV2(args);
```
Same for all the other transactionService function calls here

src/modules/safe-queue/helpers/origin.helper.ts
```diff
@@ -0,0 +1,75 @@
...
+ * field, so we lift it out here.
+ *
+ * @param origin - The JSON string to parse (typically from the Safe Transaction Service's "origin").
+ *
+ * @returns An object with the extracted originName, originUrl and note fields if present.
+ */
+export function parseOrigin(origin: string | null): {
+  originName?: string;
+  originUrl?: string;
+  note?: string;
```
tmjssz: I would extract this as a type to re-use it below

src/modules/safe/domain/safe.repository.ts
```diff
@@ -381,19 +613,42 @@ export class SafeRepository implements ISafeRepository {
...
+          chainId: args.chainId,
+          safeTxHash: args.safeTxHash,
+        }),
+        this.safeQueueService.clearAllTransactions({
+          chainId: args.chainId,
+          safeAddress: safe,
+        }),
+      );
+    }
+    Promise.all(cacheClears).catch((error) => {
```
tmjssz: Should we also use Promise.allSettled here? To make sure all caches are being cleared.

tmjssz: And why don't we re-use `this.clearMultisigTransactions` here? We could extend it by an optionl safeTxHash param and inlude the `clearMultisigTransaction` call if it's present

src/modules/safe-queue/safe-queue.service.ts
LucieFaire: I saw the reasoning - but should we check in queue-service if lack of it changes the response?

src/modules/safe/domain/safe.repository.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3223/commits/4cf353f7620666cd92ab32ae54ff8cc3f6bfc5f2

src/modules/safe-queue/helpers/origin.helper.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3223/commits/4cf353f7620666cd92ab32ae54ff8cc3f6bfc5f2

src/modules/delegate/domain/v3/delegates.v3.repository.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3223/commits/4cf353f7620666cd92ab32ae54ff8cc3f6bfc5f2

PooyaRaki:
1. safe-queue.service.ts — label thread:
 
  Checked the queue-service OpenAPI spec (https://api.safe.global/queue/openapi.json): `GET /api/v1/delegates` only accepts `chainId`, `safe`, `delegator`, `delegate`, `offset` and `limit` — there is no `label` filter. So yes, the response does change: with the queue enabled, a
  `?label=...` request returns the unfiltered delegate list, while the TX-service v2 path filters server-side. Fixing this needs either label support added to the queue-service endpoint or dropping the `label` filter from the v3 route
 
  2. delete-delegate.v3.dto.entity.ts:10 — delegator:
  
  It's both by design: `NullableAddressSchema` is `AddressSchema.nullish().default(null)`, so the client may omit the field or send `null` (hence `@ApiPropertyOptional({ nullable: true })` on the Swagger side). The class implements `z.infer<...>`, which is the schema's *output* type — after parsing, the default fills in `undefined`, so the property is always present as `Address | null`. Writing `delegator?:` would actually fail the `implements` check. Same pattern as the existing `delete-delegate.v2.dto.entity.ts`.

  3. delete-delegate.v3.dto.entity.ts:12 — safe:

  Same as above — optional on input (`.nullish()`), but the parsed output is always `Address | null` because of `.default(null)`, and the class implements the output type.

  4. update-delegate.v3.dto.entity.ts:10 — "+1":

  Same reasoning as on the delete DTO — `.nullish().default(null)` makes it optional/nullable on input but non-optional `Address | null` after parsing, which is what the class implements. Mirrors the v2 DTOs.

  5. delegates.v3.service.ts:69 — destructuring:

  Done in bb024368 — `postDelegate` and `updateDelegate` now destructure the DTO once and pass the values.

LucieFaire (dismissed):

tmjssz (dismissed):

TimDaub (dismissed):

tmjssz (dismissed):

LucieFaire (dismissed):

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude arch-review

Uxio0:
@PooyaRaki as changes are still not commited and no one is using the queue service we dediced to modify a little the API for the nested transactions feature:

It's documented on `API change` on:

https://github.com/safe-global/safe-queue-service/pull/264

src/modules/safe/domain/safe.repository.ts
```diff
@@ -617,6 +895,23 @@ export class SafeRepository implements ISafeRepository {
       transaction,
     });
 
+    if (this.safeQueueEnabled) {
+      return await this.safeQueueService.proposeTransaction({
+        chainId: args.chainId,
+        safeAddress: args.safeAddress,
+        proposeTransactionDto: args.proposeTransactionDto,
+      });
```
LucieFaire, line 957: Queue-mode proposals exist only in the queue, but `getNonces()` still reads the highest nonce exclusively from the Transaction Service. After proposing nonce N, it can recommend N again, creating conflicting transactions.

src/modules/delegate/domain/v3/delegates.v3.repository.ts
```diff
@@ -0,0 +1,158 @@
...
+    label: string;
+  }): Promise<void> {
+    if (this.safeQueueEnabled) {
+      await this.safeQueueService.postDelegate(args);
+      this.clearDelegates({
+        chainId: args.chainId,
+        safeAddress: args.safeAddress ?? undefined,
+      });
+      return;
+    }
```
LucieFaire, line 97: New delegates are stored in the queue, but `TransactionVerifierHelper.verifyProposalSignature()` still queries `DelegatesV2Repository` against the Transaction Service. When proposal signature verification is enabled, these delegates receive “Invalid signature” when proposing transactions.

src/modules/safe/domain/safe.repository.ts
```diff
@@ -268,18 +294,42 @@ export class SafeRepository implements ISafeRepository {
...
+    const nonceOrder = args.ordering.startsWith('-') ? 'desc' : 'asc';
+    const page = await this.safeQueueService.getTransactionQueue({
+      chainId: args.chainId,
       safeAddress: args.safe.address,
-      executed: false,
-      nonceGte: args.safe.nonce,
+      nonceOrder,
+      limit: args.limit,
+      offset: args.offset,
     });
```
LucieFaire, line 348: `getQueuedTransactionTag()` requests the most recently modified transaction with `limit: 1`, but queue mode substitutes descending nonce order. Confirming a lower-nonce transaction therefore leaves `txQueuedTag` unchanged, causing clients to miss updates.

src/modules/delegate/routes/v3/delegates.v3.controller.ts
```diff
@@ -0,0 +1,213 @@
...
+  @ApiNoContentResponse({
+    description: 'Delegate deleted successfully',
+  })
+  @ApiBadRequestResponse({
+    description: 'Invalid signature or unauthorized deletion attempt',
+  })
+  @Delete('chains/:chainId/delegates/:delegateAddress')
+  deleteDelegate(
+    @Param('chainId') chainId: string,
+    @Param('delegateAddress') delegateAddress: Address,
```
LucieFaire: `chainId` and `delegateAddress` bypass runtime validation.

src/modules/safe/domain/safe.repository.ts
PooyaRaki: Confirmed, thanks — `getLastTransactionSortedByNonce` only checked the tx-service page. Fixed in 6f81982e696374ee99847041cb11d11d4e28f534: when queue mode is enabled it now also fetches the highest-nonce queued transaction and returns whichever of the two has the higher nonce, so `getNonces` accounts for queue-only proposals.

src/modules/delegate/domain/v3/delegates.v3.repository.ts
PooyaRaki: Confirmed, thanks — the proposal-signature check went through `IDelegatesV2Repository`, which only reads the tx-service and never sees a delegate created through the queue. Fixed in 6f81982e696374ee99847041cb11d11d4e28f534: `TransactionVerifierHelper` now checks delegates through `IDelegatesV3Repository`, which routes to the queue when enabled.

src/modules/safe/domain/safe.repository.ts
PooyaRaki: Confirmed, thanks. Fixed in 6f81982e696374ee99847041cb11d11d4e28f534: the `-modified` branch of `_getTransactionQueue` now fetches the full queue and sorts by `modified` locally in queue mode, rather than degrading to nonce-desc, so a confirmation on a lower-nonce transaction updates `txQueuedTag` correctly.

src/modules/delegate/routes/v3/delegates.v3.controller.ts
PooyaRaki: Fixed in 93fb36d55becadf39c7598861a6937856b674790 — `chainId` now goes through `ValidationPipe(ChainIdSchema)` and `delegateAddress` through `ValidationPipe(AddressSchema)`, matching the pattern already used for the other params in this controller and elsewhere (e.g. `safes.controller.ts`, `relay.controller.ts`).

LucieFaire (approved):

PR #3227, opened 2026-07-06
feat: Filter spam token push notifications

LucieFaire:
## Summary

Resolves [WA-1677](https://linear.app/safe-global/issue/WA-1677/filter-spam-token-push-notifications-on-mobile)

Safe{Mobile} sends push notifications for incoming ERC20 transfers that the app itself then hides as spam — the user gets a push, opens the app, and sees nothing.

Since [PLA-1661](https://linear.app/safe-global/issue/PLA-1661) ([safe-transaction-service#2906](https://github.com/safe-global/safe-transaction-service/pull/2906)), `INCOMING_TOKEN` / `OUTGOING_TOKEN` webhook events carry a `trusted` boolean resolved by the Transaction Service from its curated trusted-token set. This PR parses that flag and drops incoming-token push notifications when the token is not trusted, so pushes match what the app actually shows.

This supersedes the earlier Zerion/balances-lookup approach (#3160) — no extra lookups are needed anymore; the decision ships on the event itself.

## Changes

- **Event schemas** (`src/modules/hooks/routes/entities/schemas/incoming-token.schema.ts`, `outgoing-token.schema.ts`): added `trusted: z.boolean()`. The field is **required**: the Transaction Service emits it unconditionally for both directions and both ERC20/ERC721 (verified against the upstream PR — unknown tokens come through as `trusted: false`, never omitted), and no deployments predate the flag. `OUTGOING_TOKEN` never triggers pushes; its schema change is wire-format parity only.
- **Push notification filtering** (`src/modules/notifications/domain/push/push-notification.service.ts`): `processEvent` drops `INCOMING_TOKEN` events with `trusted: false` right after the notifiable-type gate, before any Safe/subscriber lookups. Each drop is logged with the new `NOTIFICATION_SPAM_TOKEN_DROPPED` log type (`src/domain/common/entities/log-type.entity.ts`) for observability.
- **FCM payload unchanged**: the `trusted` flag is stripped before the delivery job is queued (`mapIncomingAssetEventNotification`), and `IncomingTokenNotification` is now `Omit<IncomingTokenEvent, 'trusted'>` (`src/modules/notifications/domain/v2/entities/notification.entity.ts`). `FirebaseNotification['data']` only accepts string values, so passing the boolean through would have made FCM reject every delivered token push — and the flag carries no information for the client anyway (only trusted events survive filtering). The push payload mobile receives is byte-identical to today's.
- **Tests**: event builders default to `trusted: true`; inline token-event payloads in the hooks integration/e2e specs gained the now-required field; new cases cover schema validation (accepts booleans, rejects missing/non-boolean), the drop path (no delivery jobs, no subscriber fetch, spam log emitted), the send path (`trusted: true` flows as before, end-to-end through the queue), and that the delivery payload contains no `trusted` key.

## Behavior

| Event | Before | After |
|---|---|---|
| `INCOMING_TOKEN`, `trusted: true` | push sent | push sent (unchanged payload) |
| `INCOMING_TOKEN`, `trusted: false` (spam/unknown token) | push sent | **dropped**, logged as `NOTIFICATION_SPAM_TOKEN_DROPPED` |
| `INCOMING_ETHER`, executed/pending/message/module events | push sent | unchanged |

Note: `INCOMING_TOKEN` covers ERC721 as well, so incoming-NFT pushes follow the same trusted-set semantics decided upstream in PLA-1661. No new notification settings are introduced.

### DIFF
--- a/src/domain/common/entities/log-type.entity.ts
+++ b/src/domain/common/entities/log-type.entity.ts
@@ -30,6 +30,7 @@ export enum LogType {
   NotificationSent = 'NOTIFICATION_SENT',
   NotificationEventProcessed = 'NOTIFICATION_EVENT_PROCESSED',
   NotificationDeliveryQueued = 'NOTIFICATION_DELIVERY_QUEUED',
+  NotificationSpamTokenDropped = 'NOTIFICATION_SPAM_TOKEN_DROPPED',
   PortfolioDegradedServe = 'PORTFOLIO_DEGRADED_SERVE',
   PortfolioRequestError = 'PORTFOLIO_REQUEST_ERROR',
   RateLimit = 'RATE_LIMIT',
--- a/src/modules/hooks/routes/__tests__/event-hooks-queue.e2e-spec.ts
+++ b/src/modules/hooks/routes/__tests__/event-hooks-queue.e2e-spec.ts
@@ -64,6 +64,7 @@ describe('Events queue processing e2e tests', () => {
       type: 'INCOMING_TOKEN',
       tokenAddress: faker.finance.ethereumAddress(),
       txHash: faker.string.hexadecimal({ length: 32 }),
+      trusted: true,
     },
     {
       type: 'OUTGOING_ETHER',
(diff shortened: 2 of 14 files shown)


LucieFaire:
@claude review

tmjssz (approved):

PR #3228, opened 2026-07-06
Add safe-billing-service API client module

moisses89:
# Add safe-billing-service API client module
## Short description

Introduces a new datasource module, SafeBillingServiceApiModule, that provides a typed HTTP client (ISafeBillingServiceApi) for the safe-billing-service, covering plans, customers, subscriptions, payment links, and checkout sessions. Wires it in behind the (renamed) FF_BILLING_SERVICE flag alongside the existing billing webhook, and adds config, caching, and full unit test coverage.

## Highlights

- **New API client** (safe-billing-service-api.service.ts) implementing `listPlans, getPlan, getCustomer, getSubscriptionsByCustomerId, listPaymentLinks, createCheckoutSession, and getCheckoutSession`.
- **Caching**: read endpoints are cached via a new cache.router.ts entry with a short-lived TTL (EXPIRATION_TIME_BILLING_SECONDS, default 30s) for resources not yet invalidated by webhooks (customer, subscriptions); **createCheckoutSession/getCheckoutSession are explicitly not cached since they must always be fresh.**
- **Feature flag rename**: FF_BILLING_WEBHOOK → FF_BILLING_SERVICE, now gating both the webhook receiver and the new API client module.
- **Auth**: uses a static, out-of-band-minted SAFE_BILLING_SERVICE_API_TOKEN (HS256 bearer) — CGW never mints or rotates it. Required in prod/staging when the flag is enabled, enforced via the config schema.
- **Config additions**: SAFE_BILLING_SERVICE_BASE_URI, SAFE_BILLING_SERVICE_API_TOKEN, SAFE_BILLING_SERVICE_REQUEST_TIMEOUT_MILLISECONDS, plus schema validation and .env.sample.json entries.


moisses89:
@claude review

moisses89:
> #### 1. Cache-key collision between “no filter” and `status: 'all'`
> `src/datasources/safe-billing-service-api/safe-billing-service-api.service.ts:81-93` and `src/datasources/cache/cache.router.ts:1200-1207`:
> 
> ```ts
> params: args.status ? { status: args.status } : undefined,
> // cacheDir field: args.status ?? 'all'
> ```
> 
> Two different requests hit the **same** cache key:
> 
> * `getSubscriptionsByCustomerId({ customerId })` → no `status` param → cache field `'all'`
> * `getSubscriptionsByCustomerId({ customerId, status: 'all' })` → `?status=all` → cache field `'all'`
> 
> If the backend treats an omitted param and `status=all` differently (or if only one form is supported), the second caller will silently receive a stale/wrong response. Either normalise both to the same wire request, or use distinct cache-field values.
> 
> #### 2. `SubscriptionStatusFilter` value `'all'` — is the backend contract confirmed?
> `src/datasources/safe-billing-service-api/entities/subscription.entity.ts:20-23` adds `'all'` to the filter enum and the service forwards it verbatim as `?status=all`. Worth confirming the upstream service accepts that literal — most REST filter conventions represent "no filter" as an omitted param, not a magic `all`. If the backend only knows the 8 real statuses, this call will 400. If confirmed, a short JSDoc note on `SubscriptionStatusFilter` would help.
> 
> #### 3. `customerId` naming is ambiguous
> `src/domain/interfaces/safe-billing-service-api.interface.ts:32,41`: JSDoc says _"Upstream customer ID"_ but the parameter is named `customerId`, and `CustomerSchema` exposes both `id` and `upstreamCustomerId`. Callers can't tell which one to pass. Consider renaming the argument to `upstreamCustomerId` (or clarifying explicitly) — the current setup invites bugs where the internal id is passed and 404s result.
> 
> #### 4. Single TTL for all resources
> `safe-billing-service-api.service.ts:44-46` — plans, customers, and subscriptions all share `expirationTimeInSeconds.default`. Plans change rarely; subscriptions change on user action. In practice you'll want dedicated `safeBillingService.plansTtlSeconds` / `subscriptionsTtlSeconds` (or at least an invalidation hook once the write endpoints land), otherwise cancellations will appear delayed by the default TTL.
> 
> #### 5. `customerGroup: z.string()` is very loose
> `entities/customer.entity.ts:13` — if the set of groups is known/finite upstream, prefer `z.enum([...])` so an unexpected value fails at the boundary rather than propagating into consumers.
> 
> #### 6. Test gap: `status: 'all'` path
> `safe-billing-service-api.service.spec.ts:229-297` covers "no status" and `status: 'active'`, but not `status: 'all'` — which given point (1) above is exactly the case worth pinning down.
> 
> #### 7. Minor
> * `configuration.ts:751-757` — the fallback `` `${5_000}` `` is a bit awkward; a plain `'5000'` (or a numeric constant) reads better.
> * `.env.sample.json:1220` — `dummy-token` default with `required: false` means a misconfigured deploy silently sends a bogus Bearer to the real service. Consider `required: true` or removing the default.
>   · Branch: `moisesf/pla-1721-cgw-billing-service-api-client-module-plans-customers`

1. Cache-key collision — Ignored: no caller uses it yet
2. status: 'all' contract — Ignored: confirmed valid in OpenAPI
3. customerId naming — Ignored: matches backend's own naming
4. Single TTL — Ignored: no invalidation logic yet
5. customerGroup: z.string() — Ignored: backend has no enum
6. Test gap for status: 'all' — Applied: added missing test
7. ${5_000} fallback style — Ignored: matches existing convention
8. dummy-token required: false — Applied: matches secret convention

moisses89:
@claude review

moisses89:
> **Claude finished @moisses89's task in 2m 19s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/29007618115)
> 
> ### Claude PR Review
> * [x]  Read repo CLAUDE.md and gather PR context
> * [x]  Review new API client service and interface
> * [x]  Review entities and validation schemas
> * [x]  Review config, module wiring, and cache router
> * [x]  Review test coverage
> * [x]  Post final review
> 
> ### Findings
> **1. Cache key naming is inconsistent** — `src/datasources/cache/cache.router.ts:45-50`
> 
> Two prefixes are used side-by-side (`SAFE_BILLING_*` vs `BILLING_*`), and the ID is embedded into the _key prefix_ for customer/subscription/plan (`${planId}_billing_plan`) instead of passed as the CacheDir _field_ (second arg), which is the pattern used elsewhere in this file. This will make future bulk invalidation (`prefix:*`) awkward — e.g. "invalidate all customers" can't be done via a shared prefix.
> 
> Suggest normalizing on one prefix (`safe_billing_*`) and using `new CacheDir(SAFE_BILLING_CUSTOMER_KEY, upstreamCustomerId)` so all billing keys share a scannable prefix.
> 
> **2. `listPaymentLinks` uses `customerId`, everything else uses `upstreamCustomerId`** — `safe-billing-service-api.service.ts:129-142`, `safe-billing-service-api.interface.ts:30`
> 
> Given commit `b84e97d` explicitly renamed `customerId → upstreamCustomerId` "to align with other endpoints", `listPaymentLinks(args?: { customerId?: string })` still forwards `?customerId=...` — either the alignment is incomplete or the upstream contract genuinely differs; if the latter, the interface docstring should call it out explicitly to avoid future confusion.
> 
> **3. Loose schemas will let malformed upstream payloads through**
> 
> * `checkout-session.entity.ts:10-27` — `mode`, `payment_status`, `status`, `currency` are `z.string()` but are known Stripe enums. Consider `z.enum([...])` (or at minimum a comment justifying the passthrough).
> * `checkout-session.entity.ts:20` — `metadata: z.record(z.string(), z.unknown())` is required; if upstream ever omits it (Stripe often does), parsing throws. Should be `.optional()` or `.default({})`.
> * `payment-link.entity.ts:17` — `unitAmount: z.unknown().optional()` for what should be a `number | null`. This is essentially "accept anything" and defeats the purpose of a schema.
> 
> **4. `status: 'all'` is forwarded as a query param** — `safe-billing-service-api.service.ts:120-126`
> 
> When the caller passes `status: 'all'`, the client sends `?status=all` upstream _and_ keys the cache under `'all'`. If the upstream service treats `'all'` as an unknown enum (only `SubscriptionStatusSchema` values are real statuses), this will 400 in production. If "all" is CGW-side sugar for "no filter", omit the param when `status === 'all'` and cache under `''`. Worth verifying against the upstream OpenAPI.
> 
> **5. Hardcoded staging URL as production default** — `configuration.ts:783-785`
> 
> ```ts
> baseUri: process.env.SAFE_BILLING_SERVICE_BASE_URI || 'https://safe-billing-service.staging.5afe.dev',
> ```
> 
> If `SAFE_BILLING_SERVICE_BASE_URI` is misconfigured in prod, the app will silently talk to staging with a prod token. `SAFE_BILLING_SERVICE_API_TOKEN` is already gated in `configuration.schema.ts:245-249` for deployed envs — do the same for the base URI (or drop the default entirely).
> 
> **6. Minor: constructor field-init pattern**
> 
> Reading six config values in the constructor and stashing them is fine, but calls like `getOrThrow('safeBillingService.baseUri')` will throw synchronously at DI-time if the config is missing. In practice this is caught by the schema, but if you want a defense-in-depth check, `expireTimeSeconds` isn't validated to be `> 0` anywhere.
> 
> ### Nits
> * `safe-billing-service-api.service.ts:74-77` — the `.transform(body => body.plans)` unwrap is repeated 3× (plans, customer, subscriptions). A small helper `unwrap(key, schema)` would DRY it and make the pattern explicit.
> * Test file `safe-billing-service-api.service.spec.ts:28` — `mockDataSource` declares a `post` method that `CacheFirstDataSource` doesn't actually expose; the `expect(mockDataSource.post).not.toHaveBeenCalled()` assertions test a non-existent surface. Harmless but misleading — safe to drop.
> 
> Overall the module is well-structured, cleanly separates cached-vs-uncached reads, and test coverage across the happy path, malformed payloads, and network errors is solid. The findings above are mostly consistency/robustness rather than blockers. • Branch: `moisesf/pla-1721-cgw-billing-service-api-client-module-plans-customers`

1. Renamed billing cache keys to drop the _safe_ substring collision with the Safe-metadata debug-log branch.
2. Consolidated all 5 billing cache-key constants under one consistent BILLING_* prefix.
3. Fixed .env.sample.json's SAFE_BILLING_SERVICE_API_TOKEN to "required": false (matches its actual conditional requirement).
4. Renamed listPaymentLinks's customerId → upstreamCustomerId for consistency (confirmed via the real OpenAPI spec that it's the same id, just inconsistently named upstream).
5. Added SAFE_BILLING_SERVICE_BASE_URI to the requiredWhen-gated config check, preventing a silent staging fallback in production.

src/config/entities/schemas/configuration.schema.ts
```diff
@@ -230,6 +241,18 @@ export const RootConfigurationSchema = z
         message:
           'is required in production and staging environments when SES email is enabled',
       },
+      {
```
LucieFaire, line 293: oh could you please also add a check for `BILLING_WEBHOOK_JWT_PUBLIC_KEY`, I forgot about it

src/config/entities/configuration.ts
```diff
@@ -775,6 +779,16 @@ export default () => ({
       ),
     },
   },
+  safeBillingService: {
```
LucieFaire: there is already a section called - `billing`, could you please consolidate them? I think the convention we use is to use the topic name, and lose `service` and so on part (like we have to `transactions`)

src/datasources/safe-billing-service-api/entities/plan.entity.ts
```diff
@@ -0,0 +1,17 @@
...
+  id: z.string(),
+  name: z.string().nullable(),
+  description: z.string().nullable(),
+  currentPrice: z.number(),
+  originalPrice: z.number(),
+  paymentMethod: z.literal('fiat'),
+  currency: z.enum(['usd', 'eur']),
+  features: z.array(z.string()),
+  billingCycle: z.enum(['monthly', 'yearly']).nullable(),
+  type: z.enum(['standard', 'premium', 'enterprise']),
```
LucieFaire: maybe we could set them as `const` to avoid using literals?

src/datasources/billing-api/entities/subscription.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export const SubscriptionStatusSchema = z.enum([
+  'active',
+  'canceled',
+  'incomplete',
+  'incomplete_expired',
+  'past_due',
+  'paused',
+  'trialing',
+  'unpaid',
```
LucieFaire, line 15:
```
export const SubscriptionStatus = [
`active`,
...
] as const;
```

and then it can be used as `z.enum([...SubscriptionStatus])` and you can then also pass it directly to `faker` in test builder

src/datasources/billing-api/billing-api.service.ts
```diff
@@ -0,0 +1,220 @@
...
+      url: `${this.baseUri}/api/v1/customers/${args.upstreamCustomerId}/subscriptions`,
+      params: args.status ? { status: args.status } : undefined,
+      schema: z
+        .object({ subscriptions: z.array(SubscriptionSchema) })
+        .transform((body) => body.subscriptions),
+      expireTimeSeconds: this.billingExpireTimeSeconds,
+    });
+  }
+
+  listPaymentLinks(args?: {
```
LucieFaire: nit: maybe we can lose `args` here is they are optional anyway. Less checks

src/datasources/billing-api/billing-api.service.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  /** Not cached: this creates a new resource on every call. */
+  createCheckoutSession(args: {
+    paymentLinkId: string;
+    upstreamCustomerId: string;
+    returnUrl: string;
+  }): Promise<CheckoutSession> {
+    return this.parse(
+      this.networkService
+        .post<unknown>({
```
LucieFaire: anyway to avoid `unknown` type?

src/datasources/billing-api/billing-api.service.ts
```diff
@@ -0,0 +1,220 @@
...
+
+  private request<T>(args: {
+    cacheDir: CacheDir;
+    url: string;
+    params?: NetworkRequest['params'];
+    schema: z.ZodType<T>;
+    expireTimeSeconds?: number;
+  }): Promise<T> {
+    return this.parse(
+      this.dataSource.get<unknown>({
```
LucieFaire: same here, could we have generic instead of unknown?

src/app.module.ts
```diff
@@ -93,7 +94,9 @@ export class AppModule implements NestModule {
         ...(isAuthFeatureEnabled ? [AuthModule] : []),
         ...(isOidcAuthFeatureEnabled ? [OidcAuthModule] : []),
         BalancesModule,
-        ...(isBillingWebhookFeatureEnabled ? [BillingModule] : []),
+        ...(isBillingServiceFeatureEnabled
+          ? [BillingModule, SafeBillingServiceApiModule]
```
LucieFaire: actually maybe we can merge these 2 modules in one? At lest that was my idea when i created it. They have to be wired together anyway

LucieFaire (commented):
one note: we now have `SafeBilling*` and just `Billing` namespaces. Could we consolidate it into 1 so there is no confusion as they are one and the same service

src/config/entities/schemas/configuration.schema.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3228/changes/12ebfa46a27efb0e0066549fd6167fc40242f216

src/datasources/safe-billing-service-api/entities/plan.entity.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3228/changes/8ef116e217fbf3c5395d959553647509de615e3e

src/datasources/billing-api/entities/subscription.entity.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3228/changes/8ef116e217fbf3c5395d959553647509de615e3e

src/app.module.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3228/changes/0b12e6003ec3d9d9a7bd25b59d6cc45e9167509d

src/datasources/billing-api/billing-api.service.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3228/changes/9a0b6b94218d6be48443ad5dd0244596af5e4ed0

src/datasources/billing-api/billing-api.service.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3228/changes/9a0b6b94218d6be48443ad5dd0244596af5e4ed0

src/datasources/billing-api/billing-api.service.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3228/changes/34613504707e54b9139d450cfbc2ccfbb70140f1

LucieFaire (dismissed):
🚀

LucieFaire (approved):

PR #3241, opened 2026-07-07
feat: invalidate Zerion caches on Safe transaction events and manual refresh

vseehausen:
## Summary

- On-chain events (`INCOMING/OUTGOING_ETHER/TOKEN`, `EXECUTED_MULTISIG_TRANSACTION`, `MODULE_TRANSACTION`, `SAFE_CREATED`) now clear the three Zerion-backed caches for the affected Safe together: wallet-portfolio (`/v2/safes` overview), portfolio route, and positions. The caches are cross-chain aggregates, so they are cleared by address regardless of the event's chain. Off-chain events (confirmations, messages, delegates) are untouched.
- The manual refresh (`DELETE /portfolio/:address`) routes through the same invalidation. Previously it cleared only the portfolio-route cache, leaving the overview and positions serving stale data until their TTLs expired.
- Both triggers call a single `ZerionCacheService.invalidate(address, source)`, so the "clear all three together" guarantee is structural. Each invalidation emits a `ZERION_CACHE_INVALIDATED` log with its trigger (`source` = event type or `refresh`).
- Scope change during review: an earlier revision paired invalidation with a sync flag driving Zerion's `sync=true` re-aggregation. Zerion has deprecated the `sync` parameter, so the flag mechanism was removed — freshness after invalidation now relies on Zerion's own aggregation. The client-facing `?sync=true` query param on the portfolio/positions routes is untouched.
- Unblocks raising the overview cache TTL (300s → 1h) in a follow-up config change: with event-driven invalidation, the TTL becomes a cost knob rather than a freshness bound.

## Related

- Resolves [WA-2779](https://linear.app/safe-global/issue/WA-2779/cgw-invalidate-zerion-portfolio-caches-on-transaction-events)
- Improves [WA-2788](https://linear.app/safe-global/issue/WA-2788) (stale DeFi positions after withdrawal): the triggering event now clears the positions cache, so staleness is bounded by Zerion's aggregation lag instead of our cache TTL

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -161,6 +161,12 @@ export default () => ({
           process.env.ZERION_WALLET_PORTFOLIO_TTL_SECONDS ?? `${10}`,
           10,
         ),
+        // Minimum lifetime of the sync flag set after an on-chain event; must
+        // outlive Zerion's re-aggregation lag.
+        syncFlagMinSeconds: Number.parseInt(
+          process.env.ZERION_SYNC_FLAG_MIN_SECONDS ?? `${1800}`,
+          10,
+        ),
       },
     },
   },
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -161,12 +161,6 @@ export default () => ({
           process.env.ZERION_WALLET_PORTFOLIO_TTL_SECONDS ?? `${10}`,
           10,
         ),
-        // Minimum lifetime of the sync flag set after an on-chain event; must
-        // outlive Zerion's re-aggregation lag.
-        syncFlagMinSeconds: Number.parseInt(
-          process.env.ZERION_SYNC_FLAG_MIN_SECONDS ?? `${1800}`,
-          10,
-        ),
       },
     },
   },
--- a/src/modules/hooks/domain/helpers/event-cache.helper.ts
+++ b/src/modules/hooks/domain/helpers/event-cache.helper.ts
@@ -11,10 +11,12 @@ import {
   type ICacheService,
 } from '@/datasources/cache/cache.service.interface';
 import { MAX_TTL } from '@/datasources/cache/constants';
+import { LogType } from '@/domain/common/entities/log-type.entity';
 import {
   type ILoggingService,
   LoggingService,
 } from '@/logging/logging.interface';
+import { IZerionWalletPortfolioApi } from '@/modules/balances/datasources/zerion-wallet-portfolio-api.service';
 import { IBalancesRepository } from '@/modules/balances/domain/balances.repository.interface';
 import { IBlockchainRepository } from '@/modules/blockchain/domain/blockchain.repository.interface';
 import { IChainsRepository } from '@/modules/chains/domain/chains.repository.interface';
@@ -29,6 +31,8 @@ import {
   TransactionEventType,
 } from '@/modules/hooks/routes/entities/event-type.entity';
 import { IMessagesRepository } from '@/modules/messages/domain/messages.repository.interface';
+import { IPortfolioRepository } from '@/modules/portfolio/domain/portfolio.repository.interface';
+import { IPositionsRepository } from '@/modules/positions/domain/positions.repository.interface';
 import { ISafeRepository } from '@/modules/safe/domain/safe.repository.interface';
 import { ISafeAppsRepository } from '@/modules/safe-apps/domain/safe-apps.repository.interface';
 import { IStakingRepositoryWithRewardsFee } from '@/modules/staking/domain/staking.repository.interface';
@@ -55,6 +59,12 @@ export class EventCacheHelper {
     private readonly delegatesRepository: IDelegatesV2Repository,
     @Inject(IMessagesRepository)
     private readonly messagesRepository: IMessagesRepository,
+    @Inject(IPortfolioRepository)
+    private readonly portfolioRepository: IPortfolioRepository,
+    @Inject(IPositionsRepository)
+    private readonly positionsRepository: IPositionsRepository,
+    @Inject(IZerionWalletPortfolioApi)
+    private readonly zerionWalletPortfolioApi: IZerionWalletPortfolioApi,
     @Inject(ISafeAppsRepository)
     private readonly safeAppsRepository: ISafeAppsRepository,
     @Inject(ISafeRepository)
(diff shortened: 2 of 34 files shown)


vseehausen:
**Verified locally end-to-end via the real AMQP path** (local RabbitMQ + wallet dev UI against this branch):

1. Published an `INCOMING_ETHER` event for a Sepolia safe to the `safe-transaction-service-events` exchange (RabbitMQ management API) — CGW's consumer picked it up without restart.
2. The handler logged the new structured event and did **only Redis work** (no outbound calls — the Zerion fetch below happened a full minute later, driven by a client request):

```
{ type: 'PORTFOLIO_CACHE_INVALIDATED', event_type: 'INCOMING_ETHER',
  chain_id: '11155111', address: '0xfd6002…733f4' }        # 16:05:04
```

3. Observed effects on subsequent wallet requests:
   - balances re-fetched upstream at 16:05:17 (51ms, cache_write) showing the **new** balance — visible immediately after the event instead of waiting out a TTL;
   - positions took **297ms** at 16:06:05 — cache bypassed and `sync=true` sent to Zerion via the flag; the successful fetch consumed the flag (next positions call back to single-digit ms cache hits);
   - queue/multisig caches re-populated as with the pre-existing clears.

Confirms: event → invalidate (Redis-only, call-count neutral) → next fetch per Zerion path is fresh + synced → flag consumed by first success.

src/modules/hooks/domain/helpers/event-cache.helper.ts
```diff
@@ -252,6 +262,34 @@ export class EventCacheHelper {
...
+   * Clears the Zerion-backed portfolio caches after an on-chain event. The
+   * wallet-portfolio and portfolio-route caches are cross-chain aggregates,
+   * so they are cleared by address regardless of the event's chain.
+   */
+  private clearZerionCaches(event: {
+    type: string;
+    chainId: string;
+    address: Address;
+  }): Array<Promise<void>> {
+    this.loggingService.info({
```
gjeanmart: Would rather make a `debug` log, it might print way too many logs in DD given the number of events we are treating

src/config/entities/configuration.ts
```diff
@@ -161,6 +161,12 @@ export default () => ({
           process.env.ZERION_WALLET_PORTFOLIO_TTL_SECONDS ?? `${10}`,
           10,
         ),
+        // Minimum lifetime of the sync flag set after an on-chain event; must
+        // outlive Zerion's re-aggregation lag.
+        syncFlagMinSeconds: Number.parseInt(
+          process.env.ZERION_SYNC_FLAG_MIN_SECONDS ?? `${1800}`,
+          10,
+        ),
```
gjeanmart: Not sure I fully get this flag and why we just don't invalidate the actual cache of the the Zerion's response on received RabbitMQ events?

Something like this, maybe I'm oversimplifying...

Image

vseehausen: Good question, and this is a debatable design decision.

The reason is the s[ync parameter of Zerion](https://developers.zerion.io/changelog#portfolio-api-%E2%80%94-sync-parameter). Zerion serves cached values regularly, which are highly accurate in our experience. We can, however, request a fresh indexing, to ensure everything is indexed. From our measurements this add less then a second to the response time. 

We don't require a synced response on regular calls. But when we know for sure, that a transaction settled, we want to pay the extra time cost, as we required accurate and fresh data. 

So basically we had three obvious options:
1. Never use the sync parameter.
2. Only use the sync parameter on calls (potentially directly) after a transaction.
3. Always use the sync parameter.

We decided for 2., even though this adds a little bit of extra logic and rountrip to Redis.

I was drawn between 2. and 3. and would be open to move to option 3.

src/modules/hooks/domain/helpers/event-cache.helper.ts
vseehausen: Good one, fixed in 0a19d85ad

gjeanmart (dismissed):
LGTM 🔥

PooyaRaki:
@claude review

vseehausen:
@claude review again. We just learned the sync parameter gets deprecated and adjusted the scope of this PR

vseehausen:
Thanks — all points addressed in 09bd939 and 43673ec:

1. **Vestigial `sync` plumbing** — acknowledged, but out of scope for this PR. The client-facing `?sync=true` param predates it; removing it is an API change we'll handle separately once the deprecation timeline is clear.
2. **`Promise.all` partial-failure** — reworked to best-effort: `Promise.allSettled` with a warn log per failed key, and `invalidate` never throws. Rationale: neither caller can act on a failure — the tx listener won't replay the event for a cache miss, and the wallet shouldn't get a 500 (or retry) for a failed invalidation whose entry expires with its TTL anyway. Note `Promise.all` never cancelled the sibling deletes (all three always executed); the change is purely error semantics plus the per-key operator signal.
3. **`source: string` too loose** — typed as `ZerionCacheInvalidationSource = TransactionEventType | 'refresh'`.
4. **`clearPortfolio` blast radius** — renamed the internal chain (v1 service → domain service → repository) to `clearZerionCaches`; the endpoint and the controller method mirroring it keep their names. Swagger description updated to spell out all three caches.
5. **Integration coverage** — the Zerion-cache test is now an `it.each` over all seven invalidating event types.

src/config/entities/configuration.ts
vseehausen: New information: Zerion will deprecate the sync parameter, so I removed this logic from the PR.

src/modules/hooks/domain/hooks.repository.spec.ts
```diff
@@ -58,6 +59,10 @@ const mockSafeAppsRepository = vi.mocked({
   clearSafeApps: vi.fn(),
 } as MockedObject<SafeAppsRepository>);
 
+const mockZerionCache = vi.mocked({
+  invalidate: vi.fn(),
+} as unknown as MockedObject<ZerionCacheService>);
```
LucieFaire: nit: no need to cast to `unknown`

src/modules/hooks/routes/hooks-cache.integration.spec.ts
```diff
@@ -85,6 +86,102 @@ describe('Hook Events for Cache', () => {
...
+      txHash: faker.string.hexadecimal({ length: 32 }),
+      value: faker.string.numeric(),
+    },
+    {
+      type: 'MODULE_TRANSACTION',
+      module: faker.finance.ethereumAddress(),
+      txHash: faker.string.hexadecimal({ length: 32 }),
+    },
+    {
+      type: 'EXECUTED_MULTISIG_TRANSACTION',
```
LucieFaire, line 118: should these events also invalidate Zerion? I see below for balances we only check for incoming/outgoing tokens

src/modules/portfolio/domain/portfolio.repository.spec.ts
```diff
@@ -48,10 +50,15 @@ describe('PortfolioRepository', () => {
       }),
     } as unknown as MockedObject<IConfigurationService>;
 
+    mockZerionCache = {
+      invalidate: vi.fn(),
+    } as unknown as MockedObject<ZerionCacheService>;
```
LucieFaire: same here, please tell your claude to not cast to unknown its redundant

src/modules/hooks/routes/hooks-cache.integration.spec.ts
vseehausen: Yes, they could trigger a protocol transaction that doesn't necessarily trigger a ERC20 transaction to the Safe. Potentially every non-admin transaction could trigger a change in portfolio, and they are cheap to listen to, so it's best to listen to all transactions that could change the portfolio composition.

src/modules/portfolio/domain/portfolio.repository.spec.ts
vseehausen: I did :) https://github.com/safe-global/safe-client-gateway/pull/3241/changes/df816a058015b76a7e492e501a0e8d7196e56692

src/modules/hooks/domain/hooks.repository.spec.ts
vseehausen: I did :) https://github.com/safe-global/safe-client-gateway/pull/3241/changes/df816a058015b76a7e492e501a0e8d7196e56692

LucieFaire (approved):

PR #3242, opened 2026-07-08
fix(tests): resolve remaining flaky tests after Vitest migration

vseehausen:
## Summary

Fixes three tests that kept flaking **after** #3139. All three were visible in Datadog CI Pipeline Health but missed by our own flaky-test report, whose log parser is Jest-specific and went blind once the suite moved to Vitest (#3136).

- **`name.builder`** — `nameBuilder()` sanitized (trimmed) the faker name and *then* truncated to `NAME_MAX_LENGTH`; the cut could land right after a space and re-introduce trailing whitespace, so `sanitizeName(name) !== name`. Now re-`sanitizeName()` after truncation so the result is a fixed point.
- **`bridge-name.entity.spec`** — the "unknown bridge name" case used `faker.word.noun()`, which can itself be a bridge name (`omni`, `polygon`, `across`, `gnosis`, …), making `safeParse` succeed and the assertion flip to `expected false to deeply equal …`. Now resamples until the word is genuinely not a bridge name.
- **Integration specs (61 files)** — `afterEach`/`afterAll` called `await app.close()` unguarded, so any `beforeEach` failure left `app` undefined and every test in the file reported `Cannot read properties of undefined (reading 'close')` instead of the real error (the "27 tests fail at teardown" cascade). Guarded with `await app?.close()`.

## Testing

- `yarn test:unit` for `name.builder` + `bridge-name.entity` → pass
- `biome check` clean; `tsc` introduces no new errors

## Related

- WA-1753
- Follows #3139 (clearMocks + seeded faker)

### DIFF
--- a/src/datasources/job-queue/__tests__/job-queue.service.integration.spec.ts
+++ b/src/datasources/job-queue/__tests__/job-queue.service.integration.spec.ts
@@ -68,7 +68,7 @@ describe('JobQueueService & TestJobConsumer integration', () => {
 
   afterAll(async () => {
     await queue.close();
-    await app.close();
+    await app?.close();
   });
 
   it('should process a job added to the queue', async () => {
--- a/src/datasources/network/network.module.integration.spec.ts
+++ b/src/datasources/network/network.module.integration.spec.ts
@@ -97,7 +97,7 @@ describe('NetworkModule', () => {
     });
 
     afterAll(async () => {
-      await app.close();
+      await app?.close();
     });
 
     it(`fetch client is created with timeout and is kept alive`, async () => {
@@ -188,7 +188,7 @@ describe('NetworkModule', () => {
     });
 
     afterAll(async () => {
-      await app.close();
+      await app?.close();
     });
 
     it('caches GET requests based on URL and options', async () => {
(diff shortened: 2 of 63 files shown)


vseehausen:
@claude review

vseehausen:
Thanks for the review! Addressed #1 and the matching nit on `messages.controller.integration.spec.ts` in ce81716b2 — dropped the redundant `if (app)` guards so both match the `await app?.close()` pattern used everywhere else.

On #2 (bridge-name resample loop): keeping the unbounded `while` as-is. With a bounded loop + hard-coded fallback we'd risk handing the test a deterministic, attacker-chosen-looking name in the pathological case, which defeats the point of drawing a random non-bridge noun. Given `faker.word.noun()` draws from a very large vocab vs. only ~20 bridge names, we'd rather accept the astronomically-unlikely non-termination than a potentially deterministic name.

Leaving #3 (name.builder note) as-is — agreed it's not reachable in practice.

vseehausen:
@claude review again

PooyaRaki:
@claude review

PooyaRaki (approved):

PR #3243, opened 2026-07-09
feat(auth): add loopback for localhost redirect

LucieFaire:
## Summary
resolves [WA-2807](https://linear.app/safe-global/issue/WA-2807/adjust-cgw-to-allow-auth-from-local-wallet-app)

Allows locally running frontends (e.g. `http://localhost:3000`) to authenticate against deployed non-production CGW instances (staging/dev). Previously the post-login/logout redirect validation rejected localhost targets on three counts — non-HTTPS scheme, explicit port, and hostname outside `AUTH_ALLOWED_REDIRECT_DOMAIN` — making it impossible to develop a web app locally against staging auth.

## Changes

- **Loopback redirect exception** in `src/modules/auth/utils/auth-redirect.helper.ts`: in non-production environments with `AUTH_ALLOWED_REDIRECT_DOMAIN` set, redirect URLs targeting loopback hosts (`localhost`, `127.0.0.1`, `[::1]`) are now allowed over `http:` or `https:` with any port. This follows the loopback exception of [RFC 8252 §7.3](https://datatracker.ietf.org/doc/html/rfc8252#section-7.3).
- **Refactored `isAllowedRedirectUrl`** from nested conditionals into flat guards: production same-origin check first, then a shared userinfo rejection, then one policy branch per hostname class (loopback vs. allowed domain).
- **Tests**: new `loopback exception` block covering accepted hosts (with/without port, http/https), lookalike-hostname rejections, userinfo rejection, non-HTTP scheme rejection, and production still rejecting loopback.

## Security considerations

- **Production behavior is unchanged** — strict same-origin validation against `AUTH_POST_LOGIN_REDIRECT_URI`; the loopback branch is unreachable when `application.isProduction` is true.
- **No suffix matching on loopback hosts** — hostnames are compared exactly against a fixed set, so lookalikes such as `localhost.attacker.com` or `127.0.0.1.evil.dev` are still rejected.
- **URLs carrying userinfo** (`http://user:pass@localhost`) are still rejected.
- **Loopback redirects are not remotely exploitable**: the redirect lands on the requesting user's own machine, and no token travels in the URL (the access token is set as an `httpOnly` cookie on the CGW domain before the redirect).

### DIFF
--- a/src/modules/auth/utils/auth-redirect.helper.spec.ts
+++ b/src/modules/auth/utils/auth-redirect.helper.spec.ts
@@ -202,6 +202,70 @@ describe('Auth-redirect helper functions', () => {
           resolveAndValidateRedirectUrl(config, faker.internet.url()),
         ).toThrow(BadRequestException);
       });
+
+      describe('loopback exception', () => {
+        it.each([
+          { host: 'localhost' },
+          { host: '127.0.0.1' },
+          { host: '[::1]' },
+        ])('should accept http://$host with any port', ({ host }) => {
+          const url = `http://${host}:${faker.internet.port()}/${faker.word.noun()}`;
+
+          expect(resolveAndValidateRedirectUrl(config, url)).toBe(url);
+        });
+
+        it('should accept loopback without a port', () => {
+          const url = `http://localhost/${faker.word.noun()}`;
+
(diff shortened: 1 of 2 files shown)


PooyaRaki:
@claude review

PooyaRaki (approved):

PR #3244, opened 2026-07-10
feat: replace Gelato with Rhinestone as the relay provider

yagopv:
## Summary

Replaces Gelato with Rhinestone as the relay provider. Relay submission and task-status
polling move from Gelato's JSON-RPC endpoint to Rhinestone's REST API, and the twelve
per-chain API keys collapse into a single project-level key.

## Changes

**Relay provider migration and configuration simplification:**

* Updated `.env.sample.json` and configuration files to use a single `RELAY_PROVIDER_API_KEY` (project-level key) instead of per-chain keys, and changed the default relay provider base URI to Rhinestone (`https://v1.orchestrator.rhinestone.dev`). [[1]](diffhunk://#diff-ab7dd5bc4f296032b95e861acb01d599f30e41cd04a5e8a1378145286e96858fL1324-R1330) [[2]](diffhunk://#diff-a13ec7a4d04b743754a74ef5f390ded5f6b6aeb41e68de735adba4f15b9df7edL693-R701)
* Updated the configuration schema and validator tests to expect only `RELAY_PROVIDER_API_KEY` instead of per-chain keys. [[1]](diffhunk://#diff-6bedff3adfc4c76d39fae16033ea8bf7e94e073d2a5334c075041cac5e255a63L131-R131) [[2]](diffhunk://#diff-885cfe573d195b536beaa5c07edd3c5d5d4aabc0fcac19cb08d580700ff5bd6fL43-R43) [[3]](diffhunk://#diff-885cfe573d195b536beaa5c07edd3c5d5d4aabc0fcac19cb08d580700ff5bd6fL118-R107) [[4]](diffhunk://#diff-885cfe573d195b536beaa5c07edd3c5d5d4aabc0fcac19cb08d580700ff5bd6fL184-R162) [[5]](diffhunk://#diff-885cfe573d195b536beaa5c07edd3c5d5d4aabc0fcac19cb08d580700ff5bd6fL251-R218)
* Simplified the `relay.apiKey` configuration from an object mapping chain IDs to keys, to a single string value. [[1]](diffhunk://#diff-8e9fd4781f9f2ef0d5ba9d9c955b9d74d993588ca787837d4da9e103dd3373d0L354-R354) [[2]](diffhunk://#diff-a13ec7a4d04b743754a74ef5f390ded5f6b6aeb41e68de735adba4f15b9df7edL693-R701)

**Code and test updates:**
* Added new `rhinestone-api.service.ts` using the Rhinestone API integration
* Updated the `IRelayApi` interface to allow an optional `safeTxHash` parameter in the `relay` method.
* Removed the now-obsolete `gelato-api.service.spec.ts` test file, which was specific to the old Gelato per-chain key setup.
* Added a new `RELAY_SUPPORTED_CHAIN_IDS` constant for use in relay tests, reflecting supported chains with the new project-level key approach.


src/modules/relay/datasources/rhinestone-api.service.ts
```diff
@@ -0,0 +1,185 @@
...
+  private headers(): Record<string, string> {
+    return { 'x-api-key': this.apiKey };
+  }
+
+  /**
+   * Builds a reproducible cURL for the Rhinestone submit request, so the exact
+   * failing call can be shared for debugging. Only logged on error. The API key
+   * is redacted — replace `$RHINESTONE_API_KEY` with a real key to run it.
+   */
+  private buildCurl(url: string, body: object): string {
```
akshay-ap: why do we need this in error? Developer can build curl command by looking at the error.

src/modules/relay/datasources/rhinestone-api.service.ts
```diff
@@ -0,0 +1,185 @@
...
+          url,
+          networkRequest: {
+            headers: this.headers(),
+          },
+        });
+
+      const response = RhinestoneTaskStatusResponseSchema.parse(data);
+      return {
+        chainId: args.chainId,
+        id: response.taskId,
```
akshay-ap, line 141: no `createdAt`? 

https://github.com/safe-global/safe-client-gateway/blob/b915eb7e1b5a98560ab770ff72f4446daa5329f8/src/modules/relay/domain/entities/relay-task-status.entity.ts#L15

akshay-ap (changes requested):

src/modules/relay/datasources/rhinestone-api.service.ts
yagopv: Is only for testing hehe. Will remove before merge. Is useful for me to report them

src/modules/relay/datasources/rhinestone-api.service.ts
yagopv: nope only taskId

src/modules/relay/datasources/rhinestone-api.service.ts
falvaradorodriguez: Agree with Akshay, this shouldn’t be here

src/modules/relay/datasources/rhinestone-api.service.ts
```diff
@@ -0,0 +1,185 @@
...
+      this.loggingService.error(
+        `Error relaying transaction for chain ${args.chainId}: ${this.formatError(error)}\n${this.buildCurl(url, body)}`,
+      );
+      throw this.httpErrorFactory.from(error);
+    }
+  }
+
+  /**
+   * Retrieves the status of a relay task from Rhinestone. Status codes are
+   * Gelato-compatible (100/110/200/400/500), so they map 1:1 onto
```
falvaradorodriguez: We shouldn’t mention Gelato here; after this PR, there won’t be anything related to it. Can you check elsewhere?

src/config/entities/schemas/configuration.schema.ts
```diff
@@ -128,18 +128,7 @@ export const RootConfigurationSchema = z
...
-    RELAY_PROVIDER_API_KEY_POLYGON: z.string(),
-    RELAY_PROVIDER_API_KEY_POLYGON_ZKEVM: z.string(),
-    RELAY_PROVIDER_API_KEY_BASE: z.string(),
-    RELAY_PROVIDER_API_KEY_ARBITRUM_ONE: z.string(),
-    RELAY_PROVIDER_API_KEY_AVALANCHE: z.string(),
-    RELAY_PROVIDER_API_KEY_LINEA: z.string(),
-    RELAY_PROVIDER_API_KEY_BLAST: z.string(),
-    RELAY_PROVIDER_API_KEY_UNICHAIN: z.string(),
-    RELAY_PROVIDER_API_KEY_SEPOLIA: z.string(),
+    RELAY_PROVIDER_API_KEY: z.string(),
```
falvaradorodriguez, line 186: We should update this in `docker-compose.yml`

src/modules/relay/datasources/rhinestone-api.service.ts
yagopv: done

src/modules/relay/datasources/rhinestone-api.service.ts
yagopv: done

src/config/entities/schemas/configuration.schema.ts
yagopv: done

falvaradorodriguez (dismissed):
LGTM

LucieFaire (dismissed):
👍

akshay-ap (dismissed):

PooyaRaki:
@claude arch-review

dasanra:
@claude arch-review

src/__tests__/deployments.helper.ts
```diff
@@ -137,3 +137,26 @@ export function getDeploymentVersionsByChainIds(
...
 }
+
+/**
+ * Chain IDs used across relay tests to build deployment fixtures. Previously
+ * derived from `relay.apiKey` (one key per chain); that config is now a single
+ * project-level key, so relay tests reference this explicit list instead. The
+ * runtime relay chain gate is the chain's `relayer.type` (from ChainsRepository),
+ * not this list.
+ */
+export const RELAY_SUPPORTED_CHAIN_IDS: Array<string> = [
```
PooyaRaki, line 148: Can we read these values from ENV/ config service so that in case we want to add a new network we don't need a deployment?

src/modules/relay/routes/relay.controller.ts
```diff
@@ -44,6 +45,19 @@ import { HexSchema } from '@/validation/entities/schemas/hex.schema';
 import { NumericStringSchema } from '@/validation/entities/schemas/numeric-string.schema';
 import { ValidationPipe } from '@/validation/pipes/validation.pipe';
 
+/**
+ * The relay provider identifies tasks with an opaque id that we interpolate
+ * into its URL path, so constrain it to characters that cannot alter the URL
+ * we build. Mirrors the equivalent pipe in `BillingController`.
+ */
+const taskIdPipe = new ValidationPipe(
```
PooyaRaki: Nit: We should extract it to a separate file.

src/modules/relay/datasources/rhinestone-api.service.ts
```diff
@@ -0,0 +1,181 @@
...
+   *   what is actually executed is determined by `to` and `data`.
+   */
+  async relay(args: {
+    chainId: string;
+    to: Address;
+    data: string;
+    safeTxHash?: Hex;
+  }): Promise<Relay> {
+    const url = `${this.baseUri}/safe-transactions`;
+    const body = {
```
PooyaRaki: We could inline this object and remove the `const`

src/modules/relay/datasources/rhinestone-api.service.ts
```diff
@@ -0,0 +1,181 @@
...
+        `Error getting task status ${args.taskId} for chain ${args.chainId}: ${this.formatError(error)}`,
+      );
+      throw this.httpErrorFactory.from(error);
+    }
+  }
+
+  async getRelayCount(args: {
+    chainId: string;
+    address: Address;
+    // TODO: Change to Raw when cache service is migrated
```
PooyaRaki, line 158: Why should we change to `Raw` here?

src/modules/relay/routes/relay.controller.ts
dasanra: Done https://github.com/safe-global/safe-client-gateway/pull/3244/commits/76d96bb1822b2cd0ab94e28cc43434b2e3e8f0d4

src/modules/relay/datasources/rhinestone-api.service.ts
dasanra: Done https://github.com/safe-global/safe-client-gateway/pull/3244/commits/02ece294ef4fd5a30c0a9bb7769db442b745f063

src/modules/relay/datasources/rhinestone-api.service.ts
dasanra: This is pre-existing, it was like that in the `gelato-api.service.ts`. From my perspective this should not be part of the migration.

src/__tests__/deployments.helper.ts
dasanra: This is a test helper, I would not point it to ENV/config service for that. This way they are atomic and isolated from external config.

PooyaRaki:
@claude arch-check

src/modules/relay/datasources/rhinestone-api.service.ts
PooyaRaki: We should remove the comment then.

PooyaRaki:
@claude arch-review

src/__tests__/deployments.helper.ts
PooyaRaki: I’ll leave it up to your discretion but I'd highly recommend we replace it with something to read from the deployment repo:
```ts
// deployments.helper.ts
  export function getDeploymentChainIds(): Array<string> {
    return Object.keys(versionsByDeploymentByChainId.Safe);
  }

  // each relay spec
  const supportedChainIds = faker.helpers.arrayElements(getDeploymentChainIds(), 3);
```

PooyaRaki (approved):
Approved to unblock but we need to ensure Claude remarks are fixed either here or in a follow up PR.

PR #3245, opened 2026-07-10
Feat/space encryption add fields

PooyaRaki:
## What

  Expands KMS field-level encryption from `users.email` (already shipped) to every remaining sensitive field across the wallets, users/members, and spaces modules, plus the space audit-log payload. Values are encrypted directly by KMS and bound to their owner via the KMS encryption
  context; searchable/unique columns get a sibling blind-index column so equality and uniqueness keep working over ciphertext.

  Encrypted in this PR:

  | Table | Field(s) | Context | Blind index |
  |---|---|---|---|
  | `wallets` | `address` | `{ userId }` | `address_index` |
  | `spaces` | `name` | `{ spaceId }` | — |
  | `space_safes` | `address` | `{ spaceId }` | `address_index` |
  | `space_address_book_items` | `address`, `name` | `{ spaceId }` | `address_index` |
  | `address_book_requests` | `address`, `name` | `{ spaceId }` | `address_index` |
  | `members` | `name`, `alias` | `{ spaceId }` | — |
  | `space_audit_log` | `payload` (whole blob) | `{ spaceId }` | — |
  
  ## How
  
  **One stateless crypto core.** All mechanics live in `KmsEncryptionService` (`kms:v1:<base64url>` format), which knows nothing about scopes: `encrypt`/`decrypt` take a caller-supplied encryption context (bound into the ciphertext as AAD), and `blindIndex` computes a keyed HMAC over the
  normalized value. The old email-only `EmailEncryptionService` is removed and folded into this.
  
  **Thin per-scope wrappers.** `WalletEncryptionService`, `UserEncryptionService`, `MemberEncryptionService`, and `SpaceEncryptionService` each own only their context shape (`{ userId }` / `{ spaceId }`) and batch helpers, so repositories never handle the context directly.
  
  **Blind-index uniqueness.** Each unique address column's constraint is split into two partial unique indexes, following the `users.email` precedent:
  - `*_plain` — `WHERE address_index IS NULL` (plaintext rows, encryption disabled)
  - `*_index` — `WHERE address_index IS NOT NULL` (encrypted rows, uniqueness enforced on the blind index)
  
  **Audit payload as a single encrypted blob.** The audit payload is serialized to JSON and encrypted as one value under the row's `{ spaceId }` scope, then decrypted and re-parsed against the event taxonomy on read. There is **no per-address schema** — addresses inside the payload are
  not individually encrypted; `AddressSchema` stays plaintext because the enclosing blob is what's encrypted.
  
  **Domain type widening.** `wallets.address` (and the address columns above) widen to `text` — KMS ciphertext exceeds the old `varchar` caps. Plaintext length limits are enforced in the DTO schemas instead. The domain wallet address type widens to `string` accordingly.
  
  ## Migrations

  - `1781700000000-space-field-encryption` — purely structural: widens encrypted columns to `text`, adds nullable `address_index` columns, splits the uniqueness constraints, drops the now-meaningless `idx_members_name` btree. No data rewrite.
  - `1781710000000-encrypt-space-audit-payload` — widens the audit payload column for ciphertext.
  
  ## Backfill
  
  `scripts/backfill-field-encryption.ts` encrypts existing rows for every field above (users → wallets → spaces → space_safes → address book → members → audit payload). It forces encryption on for itself regardless of `ENCRYPTION_ENABLED`, and is idempotent:

  ```
  ts-node scripts/backfill-field-encryption.ts [--verify] [--dry-run]
  ```
  - `--dry-run` — report per-table plaintext row counts; write nothing.
  - `--verify` — assert no plaintext remains; non-zero exit if any does.
 
  ## Rollout order (must be followed)
  
  There is **no plaintext-read passthrough while enabled** — that dual-read safety net was removed. `KmsEncryptionService.decrypt` now hard-errors on a plaintext value read while encryption is enabled, so **the backfill must complete before the flag flips**:

  1. Deploy this code with `ENCRYPTION_ENABLED=false`.
  2. Run both column migrations (structural, no rewrite).
  3. Run the backfill script until `--verify` passes (encrypts all existing rows while the fleet still writes plaintext).
  4. Flip `ENCRYPTION_ENABLED=true` (new writes encrypt; all reads expect ciphertext).
 
  Enabling the flag requires `ENCRYPTION_INDEX_KEY` and `AWS_KMS_ENCRYPTION_KEY_ID` (validated at boot). KMS credentials come from IRSA or the dedicated `KMS_AWS_*` static pair.

  ## Rollback safety
  
  Blind indexes are computed whenever the index key is configured, independent of `ENCRYPTION_ENABLED`, and `kms:` ciphertext is always decryptable regardless of the flag — so disabling the flag after a rollout keeps existing encrypted rows readable and their lookups matching.
 
  ## Testing
  - Unit specs for the crypto core, all four wrapper services, and every touched repository/service.
  - Integration specs assert `address_index` presence and blind-index matching on encrypted rows.
  - Audit reader covers the degrade paths: a decrypt failure (e.g. an un-backfilled plaintext row read while enabled) and a schema-validation miss both degrade to an empty payload **and emit a `warn`** — no silent swallowing.

  ## Notes / follow-ups (from review)
  
  - Generic `find`/`findOne` on the members and spaces repositories don't blanket-decrypt `name`/`alias`; every current consumer decrypts at egress. Making the boundary uniform is tracked as a follow-up (a generic `find` can span multiple spaces, so a per-space context isn't always
  available).
  - `SpaceSafesRepository.find`/`findOrFail` (and `decryptLoadedSpaceSafes`) currently have no production callers — candidate for removal as YAGNI in a follow-up.
  - Collapsing the identical `safeAddressIndex`/`itemAddressIndex`/`requestAddressIndex` accessors into one `addressIndex()` is deferred to avoid API churn here.


PooyaRaki:
@claude review

PooyaRaki:
Thanks for the thorough review. Pushed a fix for item 4; rundown on the rest below.
  
  ### Fixed
  **Item 4 — audit `decodePayload` swallowed every error into `{}`.** It now emits a `warn` on both the decrypt/parse `catch` and the schema-validation miss (which returned `{}` without ever throwing), carrying `spaceId`/`auditLogId`/`eventType` but never the decrypted payload. Added
  tests for the decrypt-throw and schema-miss degrade paths.

  ### Valid, but not a live bug — follow-up
  **Items 1 / 2 — generic finders don't decrypt `name`/`alias`.** Accurate, but not a leak today: every consumer of `find`/`findOne` either doesn't read `name`/`alias` (auth checks use `role`/`status`/`user.id`) or decrypts at egress (`spaces.service.ts:132`, `members.service.ts:98`,
  `space-invite-email.service.ts:113`). It isn't uniform for a reason — `find` takes an arbitrary `where` that can span multiple spaces, and `name`/`alias` need a per-space KMS context whose `space` relation isn't always loaded, so a blanket boundary-decrypt isn't safe for members. Agree
  it's a foot-gun; tracking as a follow-up rather than reworking the boundary here.
 
  **Item 3 — `decryptLoadedSpaceSafes` throws on missing relation.** The throw is real, but `SpaceSafesRepository.find`/`findOrFail` have **no production callers** (only `findBySpaceId`/`create`/`delete` are used) — so it's unreachable today, and it's the same method cited as the "good
  example" in item 2. Leaning toward deleting `find`/`findOrFail`/`decryptLoadedSpaceSafes` as YAGNI rather than hardening dead code — open to keeping it if there's a planned caller.

  ### Deferred 
  - **Item 5** — `safeAddressIndex`/`itemAddressIndex`/`requestAddressIndex` are identical delegations to `blindIndex`; happy to collapse to one `addressIndex()`, holding it out to avoid churning the `SpaceEncryptionService` API alongside a correctness fix.
  - **Item 6** — dedup of the four wrapper services is a cross-module stylistic refactor; follow-up.
  - **Item 7** — `findOrCreateByWalletAddress` orphan cleanup is pre-existing, not introduced here.
  - **Item 8** — per-row KMS fan-out is fine at current traffic; a semaphore mirroring the backfill's `ENCRYPT_CONCURRENCY` can come with load headroom.

  ### PR description (items 9 / 10) — both correct, will update
  - No `AuditAddressSchema` exists — the payload is a single encrypted blob.
  - The plaintext-read passthrough was removed: `KmsEncryptionService.decrypt` now hard-errors on plaintext-while-enabled (`kms-encryption.service.ts:114`), so **backfill must complete before the flag flips**, or rows hit the (now-logged) empty-payload path from item 4. Also fixing the
  stale comment at `space-encryption.service.ts:150` that implies plaintext always passes through — it only does while encryption is disabled.

PooyaRaki:
@claude review

PooyaRaki:
A + B — stale docstring & unsafe rollout order ✅ Fixed in addc4f580
  Confirmed. Since 0ec08692f, KmsEncryptionService.decrypt hard-errors on plaintext when enabled and the repos match on the blind index alone — the "temporary passthrough", "dual-read arms", and "throw-on-plaintext guard" the docstring cited no longer exist, and the prescribed order
  (flip flag → then backfill) would turn every un-backfilled row into a hard read error. Rewrote the choreography to deploy → migrate → backfill until --verify passes → flip, documented the write race between --verify and the flip (with mitigation: freeze writes across the flip + a final
  --verify immediately before flipping, roll back if non-zero), and corrected the now-false --verify error message.

  D — O(existing × payload) blind-index recompute ✅ Fixed in addc4f580
  Confirmed. Replaced the per-candidate itemAddressIndex() recompute inside the loop with a Map<addressIndex, item> built once (one HMAC per input item), preserving the prior first-match semantics.
 
  C — shared AAD per scope — Won't do in this PR
  Accurate as described, but it's non-critical defense-in-depth (presupposes DB write access), and adding a field disambiguator to the AAD means re-encrypting every existing ciphertext plus touching all wrapper services. Not worth the added complexity here; can be revisited as its own
  hardening change if we decide the transplant vector is in scope.
 
  E — SpaceSafesRepository.create limit check outside the transaction — Won't do in this PR
  Real TOCTOU, but pre-existing (as noted, not introduced here) and orthogonal to the encryption work. Deferring to its own change rather than widening this PR's scope.

  Test coverage — the decrypt-throws-on-plaintext contract is already partly covered in kms-encryption.service.spec.ts; the --verify non-zero-exit path would need a real DB (integration), so leaving it out of this PR.

PooyaRaki:
@claude review

PooyaRaki:
Round 3 — addressed F + M, pushing back on F's severity framing.

  F — fixed the guard, but downgrading the "before flag flip" framing. Short-circuited '' in KmsEncryptionService.decrypt (06dbc4b85) as defense-in-depth against a stray blank landing in the race window, plus a covering unit test. It's a cheap guard, worth having — but it's not a rollout
  blocker, and the scenario as written doesn't hold:
  
  - members.alias === '' can't reach the throw. decryptMembers already guards it: member.alias ? decrypt(...) : Promise.resolve(null) (member-encryption.service.ts:33-35). A falsy alias short-circuits to null, no KMS call.
  - The "--verify passes with a blank-in-DB warning outstanding" invariant is false. countTablePlaintext counts '' as plaintext-remaining ("col" IS NOT NULL AND "col" NOT LIKE 'kms:v1:%' — true for ''), so --verify exits non-zero (backfill:437-447, 508-513). The blank-warning block cited
  as evidence (:452-468) says it itself: "…before --verify can pass." A blank row blocks the flip; it doesn't slip past it.
 
  So the invariant chain (verify passes → flip → 500s) requires flipping despite a failing --verify, which the whole choreography exists to prevent. Guard kept anyway.

  M — fixed. Docstring now references the ENCRYPTED_PREFIX constant alongside the literal so they can't drift (06dbc4b85).

  G, H, J, K, I, L — acknowledged, deferring. All correctly labeled minor/cosmetic/rare. I/L (startup KMS log + boot smoke-encrypt) are worthwhile ops guardrails but out of scope for this PR; will track separately.

PooyaRaki:
@claude review

scripts/login-as-seeded-user.ts
```diff
@@ -0,0 +1,151 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
```
PooyaRaki: This file will be removed before merging.

scripts/seed-fake-users.ts
```diff
@@ -0,0 +1,234 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
```
PooyaRaki: This will be removed before merging.

src/config/entities/configuration.ts
```diff
@@ -916,6 +884,20 @@ export default () => ({
       },
     },
   },
+  // Field-level encryption via AWS KMS. Values are encrypted directly by
+  // KMS, bound to their owner via the KMS encryption context; a KMS-wrapped
+  // HMAC key computes blind indexes for lookups/uniqueness. Domain-free:
+  // callers supply the context + field-id label (see KmsEncryptionService).
+  encryption: {
+    enabled: process.env.ENCRYPTION_ENABLED?.toLowerCase() === 'true',
```
LucieFaire, line 889: shouldnt this be a FF like the rest?

LucieFaire (dismissed):
I tried to look through, but honestly from code I dont see anything out of the ordinary, and the logic we can only properly test when we migrate :)
So approve from me

src/config/entities/configuration.ts
PooyaRaki: Well, we have some other keys that are like this:
https://github.com/safe-global/safe-client-gateway/blob/7d72ab9f212a11dcf20ddea5cf20ea7a0177279e/src/config/entities/configuration.ts#L1069
I would leave it as is for now.

LucieFaire: ok, but i would say that also breaks the convention

PooyaRaki: This PR is already huge. Let's open another PR to rename all such keys.

TimDaub (dismissed):

TimDaub (approved):

PR #3247, opened 2026-07-13
fix: CF safes: filter deployed Safes from counterfactual space endpoint

vmosharova:
## Summary

Resolves: [WA-2847](https://linear.app/safe-global/issue/WA-2847/cgw-fix-stop-returning-depolyed-safes-as-cf)
First-aid fix for `GET /v1/spaces/:spaceId/counterfactual-safes` returning already-deployed Safes as counterfactual.

A `counterfactual_safes` row records a *predicted* Safe but is never cleared when that Safe is later deployed on-chain, so a stale row makes a deployed Safe surface on this endpoint as counterfactual. The read path never checked deployment status.

This patch cross-checks each candidate against the transaction service (`isSafe`) and drops the ones that are actually deployed. Fail-open: if the check errors, the row is kept so a transaction-service outage can't hide a genuinely counterfactual Safe.

This is a read-side fix only — it hides deployed Safes from the response but does not stop the bad rows from being written. **Write-side hardening (reject deployed addresses on `POST /counterfactual-safes`, FE guard) and cleanup of existing stale rows will follow in a child PR.**

## Changes

- Filter deployed Safes out of the space counterfactual-safes response (`SpaceCounterfactualSafesService`), via `ISafeRepository.isSafe`, run in parallel, fail-open on error.
- Inject `SafeRepositoryModule` into `CounterfactualSafesModule`.
- Add unit tests for `SpaceCounterfactualSafesService` (first unit spec for this service): deployed filtering, undeployed pass-through, fail-open, empty space, non-member, and multi-chain per-chain filtering.

### DIFF
--- a/src/modules/counterfactual-safes/routes/space-counterfactual-safes.service.spec.ts
+++ b/src/modules/counterfactual-safes/routes/space-counterfactual-safes.service.spec.ts
@@ -1,0 +1,170 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+
+import { faker } from '@faker-js/faker';
+import { getAddress } from 'viem';
+import type { MockedObject } from 'vitest';
+import { siweAuthPayloadDtoBuilder } from '@/modules/auth/domain/entities/__tests__/auth-payload-dto.entity.builder';
+import { AuthPayload } from '@/modules/auth/domain/entities/auth-payload.entity';
+import { counterfactualSafeBuilder } from '@/modules/counterfactual-safes/datasources/entities/__tests__/counterfactual-safe.entity.db.builder';
+import type { ICounterfactualSafesRepository } from '@/modules/counterfactual-safes/domain/counterfactual-safes.repository.interface';
+import { SpaceCounterfactualSafesService } from '@/modules/counterfactual-safes/routes/space-counterfactual-safes.service';
+import type { ISafeRepository } from '@/modules/safe/domain/safe.repository.interface';
+import type { ISpaceSafesRepository } from '@/modules/spaces/domain/safes/space-safes.repository.interface';
+import type { IMembersRepository } from '@/modules/users/domain/members/members.repository.interface';
+
+const mockMembersRepository = vi.mocked({
+  findOne: vi.fn(),
+} as unknown as MockedObject<IMembersRepository>);
+
+const mockSpaceSafesRepository = vi.mocked({
+  findBySpaceId: vi.fn(),
+} as unknown as MockedObject<ISpaceSafesRepository>);
+
+const mockCounterfactualSafesRepository = vi.mocked({
+  find: vi.fn(),
+} as unknown as MockedObject<ICounterfactualSafesRepository>);
+
+const mockSafeRepository = vi.mocked({
+  isSafe: vi.fn(),
+} as unknown as MockedObject<ISafeRepository>);
+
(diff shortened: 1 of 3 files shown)


src/modules/counterfactual-safes/routes/space-counterfactual-safes.service.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+import { counterfactualSafeBuilder } from '@/modules/counterfactual-safes/datasources/entities/__tests__/counterfactual-safe.entity.db.builder';
+import type { ICounterfactualSafesRepository } from '@/modules/counterfactual-safes/domain/counterfactual-safes.repository.interface';
+import { SpaceCounterfactualSafesService } from '@/modules/counterfactual-safes/routes/space-counterfactual-safes.service';
+import type { ISafeRepository } from '@/modules/safe/domain/safe.repository.interface';
+import type { ISpaceSafesRepository } from '@/modules/spaces/domain/safes/space-safes.repository.interface';
+import type { IMembersRepository } from '@/modules/users/domain/members/members.repository.interface';
+
+const mockMembersRepository = vi.mocked({
+  findOne: vi.fn(),
+} as unknown as MockedObject<IMembersRepository>);
```
LucieFaire: nit: please ask your Claude to memo that there is no need to cast to unknown :)

src/modules/counterfactual-safes/routes/space-counterfactual-safes.service.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+      .with('address', getAddress(faker.finance.ethereumAddress()))
+      .build();
+    const undeployed = counterfactualSafeBuilder()
+      .with('chainId', chainId)
+      .with('address', getAddress(faker.finance.ethereumAddress()))
+      .build();
+
+    mockMembersRepository.findOne.mockResolvedValue({
+      id: userId,
+    } as never);
```
LucieFaire: a bit weird - why do we cast to `never` ?

src/modules/counterfactual-safes/routes/space-counterfactual-safes.service.spec.ts
```diff
@@ -0,0 +1,172 @@
...
+    mockSpaceSafesRepository.findBySpaceId.mockResolvedValue([
+      { chainId, address: undeployed.address },
+    ]);
+    mockCounterfactualSafesRepository.find.mockResolvedValue([undeployed]);
+    mockSafeRepository.isSafe.mockResolvedValue(false);
+
+    const result = await target.get(spaceId as never, authPayload);
```
LucieFaire: looks like there are a bunch of places where it uses `never`, maybe we can avoid it

LucieFaire (commented):
looks good, just small remarks in tests

src/modules/counterfactual-safes/routes/space-counterfactual-safes.service.spec.ts
vmosharova: Thanks for pointing it out! Dropped mock casts here: [3f9d775](https://github.com/safe-global/safe-client-gateway/pull/3247/commits/3f9d775f7fc97f17157d7f440481a1a558195984)

LucieFaire (approved):

PR #3249, opened 2026-07-13
feat: add host configuration for application

PooyaRaki:
## Summary
Fixes a Fastify issue where the app was not bound to the server.

## Changes
- Adds the host address to the listen call so the application binds correctly.

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -37,6 +37,7 @@ export default () => ({
     // Defaults to true.
     runMigrations: process.env.RUN_MIGRATIONS?.toLowerCase() !== 'false',
     port: process.env.APPLICATION_PORT || '3000',
+    host: process.env.APPLICATION_HOST || '0.0.0.0',
     allowCors: process.env.ALLOW_CORS?.toLowerCase() === 'true',
   },
   auth: {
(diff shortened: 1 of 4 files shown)


src/config/entities/configuration.ts
```diff
@@ -37,6 +37,7 @@ export default () => ({
     // Defaults to true.
     runMigrations: process.env.RUN_MIGRATIONS?.toLowerCase() !== 'false',
     port: process.env.APPLICATION_PORT || '3000',
+    host: process.env.APPLICATION_HOST || '0.0.0.0',
```
LucieFaire, line 40: do you know why it suddenly required?

LucieFaire (dismissed):

src/config/entities/configuration.ts
PooyaRaki: The server module behaves differently in Fastify:
> By default, the server will listen on the address(es) resolved by localhost when no specific host is provided. If listening on any available interface is desired, then specifying 0.0.0.0 for the address will listen on all IPv4 addresses. The address argument provided above will then return the first such IPv4 address. The following table details the possible values for host when targeting localhost, and what the result of those values for host will be.

https://fastify.dev/docs/latest/Reference/Server/

LucieFaire (approved):

PR #3250, opened 2026-07-13
fix: reject deployed safes on POST for counterfactual safes

vmosharova:
## Summary

Resolves: [WA-2856](https://linear.app/safe-global/issue/WA-2856/cgw-guard-for-post-counterfactual-safes-reject-writes-for-already)

Layer 2 of the counterfactual-safes hardening: reject writes for already-deployed Safes (layer 1 is [#3247](https://github.com/safe-global/safe-client-gateway/pull/3247)).

`POST /v1/users/counterfactual-safes` previously accepted creation params for any address, with no check that the Safe was still undeployed. A row is only meant to hold a *predicted* Safe, so persisting a deployed address creates a stale record that later surfaces the Safe as counterfactual.

The create service now cross-checks each submitted Safe against the transaction service (`isSafe`) before persisting. If any Safe in the batch is already deployed, the whole request is rejected with `409 Conflict` and nothing is written (atomic, consistent with the endpoint's existing 409 on param collisions). Fail-open: if the deployment check errors, the write is allowed, matching the read-side filter.

Follows the read-side filter (previous PR). A frontend guard in `persistCounterfactualSafe` and cleanup of existing stale rows will follow.

## Changes

- Reject `POST /counterfactual-safes` with 409 when any submitted Safe is already deployed on-chain (`CounterfactualSafesService.create` → `assertNoneDeployed`), via `ISafeRepository.isSafe`, run in parallel, fail-open on error.
- Inject `SafeRepositoryModule` into `CounterfactualSafesModule`.
- Add unit tests for `CounterfactualSafesService.create`: all-undeployed persists, any-deployed rejects with 409 and persists nothing, fail-open on check error.


PooyaRaki:
@claude review

src/modules/counterfactual-safes/routes/counterfactual-safes.service.ts
```diff
@@ -69,4 +81,37 @@ export class CounterfactualSafesService {
       payload: args.payload,
     });
   }
+
+  // Reject the whole request if any submitted Safe is already deployed —
+  // counterfactual rows must only hold undeployed Safes. Fail open on error.
+  private async assertNoneDeployed(
+    payload: Array<CreateCounterfactualSafeDto>,
+  ): Promise<void> {
+    const deploymentChecks = await Promise.all(
```
TimDaub, line 90: One nice optimization that is sadly not possible (unless we even change `isSafe`) is if `isSafe` would accept an `signal: AbortSignal` then as soon as we would have found one Promise returning that a Safe is deployed, we could have aborted all other, already launched Promises launched in the map iteration. 

In fact, what would be nice is if there was a `Promise.some` we could just use here. But from what I know TS/JS doesn't have this :/

TimDaub (approved):
Looks good to me!

Just left an optional comment for further improving, maybe even in a follow up ticket

src/modules/counterfactual-safes/routes/counterfactual-safes.service.ts
vmosharova: Thanks for the review and an optimization offer! 🙏🏻

PR #3264, opened 2026-07-14
chore: refactor Manager API classes

LucieFaire:
## Summary

Extracts the per-chain API caching boilerplate — duplicated across 8 datasource managers/factories — into a single generic base class, and collapses the near-verbatim `EarnApiManager`/`StakingApiManager` clone pair into one shared implementation.

### New: `ChainApiManager<T>`

`src/datasources/common/chain-api.manager.ts` owns the lazy per-chain cache that every manager previously reimplemented:

- `protected abstract createApi(chainId)` — subclasses define only how an instance is built
- `protected getOrCreateApi(chainId)` — returns the cached instance or creates and caches it (`??=`, preserving the original lazy/memoized semantics)
- `destroyApi(chainId)` — evicts, so the next retrieval rebuilds (e.g. after `CHAIN_UPDATED`)
- `protected hasApi(chainId)` — for subclasses that need to know whether anything was cached

The base deliberately exposes no public `getApi`: each subclass declares its own one-line `getApi` (keeping interface conformance explicit), which also lets `BalancesApiManager` reuse the cache mechanics despite a different public API.

### Refactored to extend it

- `staking-api.manager.ts`, `earn-api.manager.ts`, `swaps-api.factory.ts`, `bridge-api.factory.ts`, `transaction-api.manager.ts`, `balances-api.manager.ts`, `csv-export/v1/datasources/export-api.manager.ts`, `blockchain-api.manager.ts`
- Special cases preserved exactly:
  - `BlockchainApiManager` overrides `destroyApi` to keep its best-effort RPC cache eviction, including the early return when no client was cached
  - `BalancesApiManager` keeps its own `getApi` and inherits the cache for `SafeBalancesApi` instances

### Earn/Staking collapse

`EarnApiManager` was an author-acknowledged copy of `StakingApiManager` (only the config key prefix and the `KilnApi` widget arg differed). The shared logic now lives in abstract `KilnApiManager` (`src/modules/staking/datasources/kiln-api.manager.ts`), parameterized by a `widget: 'staking' | 'earn'` field; both concrete classes are now 8 lines and keep their names, so `staking.module.ts` / `earn.module.ts` providers are untouched.

### DIFF
--- a/src/datasources/common/chain-api.manager.spec.ts
+++ b/src/datasources/common/chain-api.manager.spec.ts
@@ -1,0 +1,55 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { faker } from '@faker-js/faker';
+import { ChainApiManager } from '@/datasources/common/chain-api.manager';
+
+class TestApiManager extends ChainApiManager<{ chainId: string }> {
+  createApi = vi.fn((chainId: string) => Promise.resolve({ chainId }));
+
+  getApi(chainId: string): Promise<{ chainId: string }> {
+    return this.getOrCreateApi(chainId);
+  }
+}
+
+describe('ChainApiManager', () => {
+  let target: TestApiManager;
+
+  beforeEach(() => {
+    target = new TestApiManager();
+  });
+
+  it('should create an API instance on first retrieval and cache it', async () => {
+    const chainId = faker.string.numeric();
+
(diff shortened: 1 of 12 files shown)


PooyaRaki:
@claude review

PooyaRaki (approved):

PR #3266, opened 2026-07-14
fix(fastify): allow empty JSON bodies

LucieFaire:
## Summary

Fixes requests with `Content-Type: application/json` and an empty body (e.g. `DELETE /v1/spaces/:id`) failing with `400 Body cannot be empty when content-type is set to 'application/json'` after the Fastify migration (#3190).

Unlike Express' body-parser, which special-cases an empty JSON body as `{}`, Fastify rejects it with `FST_ERR_CTP_EMPTY_JSON_BODY`. Many HTTP clients set the JSON content type header on every request, including bodyless DELETEs, so these started failing after the migration.

## Changes

- **`src/app.provider.ts`** — new `configureFastifyBodyParsers(adapter)` helper, applied in `createFastifyAdapterFromConfiguration` so production and tests share it:
  - Registers an `application/json` parser that returns `{}` for an empty body (Express parity) and delegates non-empty bodies to Fastify's default parser, preserving its prototype-poisoning protection.
  - Uses Nest's `adapter.useBodyParser` rather than raw `addContentTypeParser`: Nest registers its own JSON parser during `app.init()`, which would otherwise collide with ours (`FST_ERR_CTP_ALREADY_PRESENT`).
  - Re-registers the `application/x-www-form-urlencoded` parser: calling `useBodyParser` sets the adapter's `_isParserRegistered` flag, which suppresses *all* of Nest's default parser registration, not just JSON. Uses `node:querystring` (same flat parsing as the `fast-querystring` Nest uses internally).
- **`src/__tests__/test-app.provider.ts`** — wraps the bare adapter in `createTestApplication` with the same helper so tests match production behaviour.
- **`src/app.provider.spec.ts`** — regression tests: empty-body DELETE with a JSON content type succeeds, an empty JSON body parses as `{}`, malformed JSON still returns 400, and prototype-poisoning payloads are still rejected.

### DIFF
--- a/src/__tests__/test-app.provider.ts
+++ b/src/__tests__/test-app.provider.ts
@@ -6,6 +6,7 @@ import {
 } from '@nestjs/platform-fastify';
 import { TestingModule } from '@nestjs/testing';
 import {
+  configureFastifyBodyParsers,
   configureShutdownHooks,
   configureVersioning,
   createFastifyAdapter,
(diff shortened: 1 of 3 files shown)


vseehausen (dismissed):

PooyaRaki (dismissed):

PooyaRaki:
@claude review

LucieFaire:
@claude review based on previous review

PooyaRaki (approved):

PR #3267, opened 2026-07-14
ci: add daily flaky test report and fix automation

vseehausen:
## Summary

- Adds a scheduled workflow (daily, 05:15 UTC, plus `workflow_dispatch`) with two sequential jobs.
- **update-data**: checks out `reports/flaky-tests-baseline`, runs `scripts/analyze-flaky-tests.ts` (incremental — resumes from `lastCheckedAt`), and commits the refreshed `reports/flaky-tests/*` data + `REPORT.md` back to that branch. Deterministic, no Claude involved; the analyzer only needs `gh` and Node builtins, so no `yarn install`.
- **fix-flaky-test**: runs `anthropics/claude-code-action@v1` (Claude Fable 5) to pick the highest-failure-count open flaky test from the refreshed data, diagnose the root cause (mock state pollution, faker collisions, unmocked calls, teardown races — never masking with retries/timeouts), verify the fix with ≥10 consecutive local runs against the same Postgres/Redis/RabbitMQ services CI uses, and open a draft PR from a `flaky-fix/*` branch as `claude[bot]`.
- Multiple fix PRs may be open in parallel — one PR per test, never duplicates: the workflow collects the open and recently merged `flaky-fix/*` PRs with their changed files and the fix stage excludes tests already covered.
- Each fix PR then goes through an in-job review loop (max 3 rounds, early exit on approve): an independent fresh Claude session (Opus, same review instructions as `claude-code-review.yml`) posts its findings as a PR comment ending in `VERDICT: approve|request-changes`; on request-changes another fresh Fable session addresses the findings, re-verifies, pushes, and replies. The shared round prompts live in `.github/prompts/flaky-fix-{review,address}.md`. A missing verdict counts as request-changes so a failed review can't silently approve.
- After the loop, the job mechanically re-runs the fixed test file 10× (via a `verify-cmd` file the fix session writes), so a skipped or hallucinated verification fails the job instead of leaving a broken PR open.
- Both jobs share a non-cancelling concurrency group so scheduled runs never overlap.

## Related

- Continues the WA-1753 flaky test work; consumes the analyzer and data branch introduced there.
- Uses the existing `ANTHROPIC_API_KEY` secret and the already-installed Claude GitHub App (the `id-token: write` OIDC exchange is what attributes commits/PRs to `claude[bot]` and lets CI trigger on the fix PRs).

### DIFF
--- a/.github/actions/flaky-review-round/action.yml
+++ b/.github/actions/flaky-review-round/action.yml
@@ -36,10 +36,10 @@ runs:
     - name: Review
       if: inputs.prev-done != 'true'
       continue-on-error: true
-      uses: anthropics/claude-code-action@v1
+      uses: anthropics/claude-code-action@e90deca47693f9457b72f2b53c17d7c445a87342 # v1.0.171
       with:
         anthropic_api_key: ${{ inputs.anthropic-api-key }}
-        claude_args: '--model opus --max-turns 60 --allowedTools "Bash,Read,Glob,Grep"'
+        claude_args: '--model opus --max-turns 60 --allowedTools "Read,Glob,Grep,Bash(gh pr view:*),Bash(gh pr diff:*),Bash(gh pr comment:*),Bash(gh pr checkout:*),Bash(yarn:*),Bash(cat:*),Bash(jq:*),Bash(ls:*),Bash(echo:*),Bash(tr:*)"'
         prompt: |
           Review round ${{ inputs.round }} of ${{ inputs.total-rounds }}.
           Follow the instructions in .github/prompts/flaky-fix-review.md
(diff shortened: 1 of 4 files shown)


vseehausen:
@claude review

vseehausen:
@claude review again

vseehausen:
@claude review again

vseehausen:
@claude review

Notes on findings from the previous round that were intentionally not applied:

- **Verdict runs unconditionally after an approve** (finding 5): deliberate. The verdict step is the carrier of the latched `done` flag, so it must execute in every round to propagate `prev-done`; guarding it would reintroduce the multi-flag fallback chains in the `if:` conditions. The leftover file read/`rm -f` on an already-approved round is harmless. (The rounds now live in a composite action, `.github/actions/flaky-review-round/action.yml`, per finding 4.)
- **No `yarn install` after `gh pr checkout` in the mechanical verify** (finding 6): fix PRs must never touch dependencies by design. If one ever does, the verify step fails loudly on the stale install, which is the desired signal — guarding against it would cost minutes on every daily run for a case that shouldn't occur.

vseehausen:
@claude review

Applied from the previous round: `continue-on-error` on the Review/Address steps so a transient action failure degrades to a request-changes verdict instead of aborting the loop and the mechanical verify (finding 1), argument-array execution instead of `eval` in the mechanical verify (finding 2), and CR/LF-stripped verdict read (finding 3).

Intentionally not applied: mechanically appending the round-3 "unreviewed" PR comment (finding 4) — fix PRs are drafts gated on human review, and the prompt-level disclosure was deemed sufficient in the previous round.

vseehausen:
@claude review

Applied from the previous round: the empty-array guard after `read -ra` so a whitespace-only `verify-cmd` fails the mechanical verification instead of no-opping ten green runs, and the verdict read now strips all whitespace (`tr -d '[:space:]'`).

Intentionally not applied (third time, same rationale): `yarn install --immutable && yarn generate-abis` after `gh pr checkout` in the mechanical verify. Fix PRs must never touch dependencies or ABI sources by design; if one ever does, the verify step failing loudly on the stale install is the desired signal, and the guard would cost minutes on every daily run for a case that shouldn't occur. Please don't re-raise this one.

vseehausen:
Will be reopened.

PR #3268, opened 2026-07-14
chore: follow up APIs cleanup

LucieFaire:
## Summary

Follow-up to the [`ChainApiManager` extraction:](https://github.com/safe-global/safe-client-gateway/pull/3264) makes the per-chain cache single-flight by storing the creation promise instead of the resolved instance, plus small cleanups from review.

### `ChainApiManager<T>` — single-flight creation (`src/datasources/common/chain-api.manager.ts`)

The cache map is now `Record<string, Promise<T>>`:

- **Concurrent `getApi` calls for the same chain share one creation.** Previously both callers observed an empty slot and each invoked `createApi`, with the loser's instance discarded.
- **Failed creations are not cached.** The creation promise evicts its own map entry on rejection, so the next retrieval retries — preserving pre-existing behavior that a naive promise cache would break. The eviction is identity-guarded (`this.apis[chainId] === created`) so a late rejection can't evict a newer entry created after an interleaved `destroyApi`.
- **Synchronous `createApi` throws become rejections** (creation is deferred via `Promise.resolve().then(...)`), so e.g. `SwapsApiFactory`'s sync `getOrThrow` failure surfaces consistently as a rejected promise.
- `destroyApi` drops its redundant existence guard (`delete` on a missing key is a no-op).

### DIFF
--- a/src/datasources/common/chain-api.manager.spec.ts
+++ b/src/datasources/common/chain-api.manager.spec.ts
@@ -81,6 +81,26 @@ describe('ChainApiManager', () => {
     expect(target.createApi).toHaveBeenCalledTimes(2);
   });
 
+  it('should not evict a replacement entry when a destroyed creation fails', async () => {
+    const chainId = faker.string.numeric();
+    const error = new Error('Creation failed');
+    let rejectCreation!: (error: Error) => void;
+    target.createApi.mockReturnValueOnce(
+      new Promise((_, reject) => {
+        rejectCreation = reject;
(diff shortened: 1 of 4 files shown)


LucieFaire:
@claude review based on the previous review https://github.com/safe-global/safe-client-gateway/pull/3264#issuecomment-4970650785

yagopv (approved):

tmjssz (approved):

PR #3269, opened 2026-07-15
feat: implement billing endpoints in CGW

falvaradorodriguez:
- Closes [PLA-1678](https://linear.app/safe-global/issue/PLA-1678/cgw-implement-billing-service-webhook-receiver-authentication-service)

## Summary

This PR implements the billing and Stripe proxy endpoints in Client Gateway as part of **PLA-1755**, replacing `safe-auth-service` as the proxy between Wallet Web and `safe-billing-service`.

Most of the outbound HTTP client infrastructure (`IBillingApi`, `BillingApi`, and its configuration) was already introduced during the webhook work in **PLA-1678**. This PR builds on that work by:

* Adding the missing client-facing billing endpoints.
* Fixing several issues in the existing billing client.
* Wiring the new functionality into the routes layer.

## Endpoints added

* `GET /v1/billing/spaces/:spaceId (uuid)/subscriptions`
* `GET /v1/billing/plans/:planId`
* `GET /v1/billing/spaces/:spaceId/session-url`
* `GET /v1/billing/spaces/:spaceId/payment-links`
* `GET /v1/billing/spaces/:spaceId/payment-links/:paymentLinkId/checkout-url`
* `GET /v1/billing/sessions/:sessionId`

All endpoints are protected by `AuthGuard`.

Space-scoped endpoints also require the user to be an active member of the corresponding space.

### Use `Space.id` as `upstreamCustomerId`

`safe-auth-service` previously used its own `user_id` as the customer identifier sent to `safe-billing-service`.

Client Gateway does not have an equivalent per-user billing concept, so the numeric `Space.id` is used instead. This mirrors the workspace-style abstraction previously provided by `safe-auth-service`.

Authorization for space-scoped endpoints reuses the existing `Member` and `SpaceIdPipe` helpers from the `spaces` module. No new authorization mechanism was introduced.

### Merge payment-link endpoints

Instead of exposing separate endpoints for space-specific and general-catalog payment links, the following endpoint now retrieves both:

```
GET /v1/billing/spaces/:spaceId/payment-links
```

Both requests to `safe-billing-service` are executed concurrently. Their results are then combined and de-duplicated by payment-link ID.

### DIFF
--- a/src/datasources/billing-api/billing-api.service.ts
+++ b/src/datasources/billing-api/billing-api.service.ts
@@ -26,7 +26,6 @@ import { CacheFirstDataSource } from '@/datasources/cache/cache.first.data.sourc
 import { CacheRouter } from '@/datasources/cache/cache.router';
 import type { CacheDir } from '@/datasources/cache/entities/cache-dir.entity';
 import { HttpErrorFactory } from '@/datasources/errors/http-error-factory';
-import { NetworkResponseError } from '@/datasources/network/entities/network.error.entity';
 import type { NetworkRequest } from '@/datasources/network/entities/network.request.entity';
 import {
   type INetworkService,
@@ -108,8 +107,8 @@ export class BillingApi implements IBillingApi {
     });
   }
 
-  /** Not cached. Bypasses {@link NetworkService}, whose response.json() parsing can't handle this endpoint's plain-text body. */
-  async getCustomerSessionUrl(args: {
+  /** Not cached: this endpoint returns a fresh, single-use portal session URL on every call. */
+  getCustomerSessionUrl(args: {
     upstreamCustomerId: string;
     returnUrl: string;
   }): Promise<string> {
@@ -118,24 +117,19 @@ export class BillingApi implements IBillingApi {
     );
     url.searchParams.set('returnUrl', args.returnUrl);
 
-    let response: Response;
-    try {
-      response = await fetch(url, {
-        headers: this.authHeaders,
-        signal: AbortSignal.timeout(this.requestTimeout),
-      });
-    } catch (error) {
-      throw this.httpErrorFactory.from(error);
-    }
-
-    if (!response.ok) {
-      const data = await response.text().catch(() => undefined);
-      throw this.httpErrorFactory.from(
-        new NetworkResponseError(url, response, data),
-      );
-    }
-
-    return response.text();
+    return this.parse(
+      this.networkService
+        .get<string>({
+          url: url.toString(),
+          networkRequest: {
+            headers: this.authHeaders,
+            timeout: this.requestTimeout,
+            responseType: 'text',
+          },
+        })
+        .then(({ data }) => data),
+      z.string(),
+    );
   }
 
   /**
--- a/src/datasources/billing-api/billing-api.service.ts
+++ b/src/datasources/billing-api/billing-api.service.ts
@@ -2,8 +2,14 @@
 import { Inject, Injectable } from '@nestjs/common';
 import { ZodError, z } from 'zod';
 import { IConfigurationService } from '@/config/configuration.service.interface';
-import type { CheckoutSession } from '@/datasources/billing-api/entities/checkout-session.entity';
-import { CheckoutSessionSchema } from '@/datasources/billing-api/entities/checkout-session.entity';
+import type {
+  CheckoutSession,
+  CheckoutSessionResult,
+} from '@/datasources/billing-api/entities/checkout-session.entity';
+import {
+  CheckoutSessionResultSchema,
+  CheckoutSessionSchema,
+} from '@/datasources/billing-api/entities/checkout-session.entity';
 import type { Customer } from '@/datasources/billing-api/entities/customer.entity';
 import { CustomerSchema } from '@/datasources/billing-api/entities/customer.entity';
 import type { PaymentLink } from '@/datasources/billing-api/entities/payment-link.entity';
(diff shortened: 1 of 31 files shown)


yagopv:
I believe PR Description is outdated. Space.id should be workspace uuid right?

falvaradorodriguez:
> I believe PR Description is outdated. Space.id should be workspace uuid right?

That's not entirely true. This service doesn't use the concept of a “workspace,” but it does use “space.” The IDs are UUIDs, in fact, it already has a format validator.

I've updated the Swagger, which had an error related to this.

yagopv:
@claude review

falvaradorodriguez:
> 6\. **No controller/e2e coverage for the new endpoints** — only `billing.service.spec.ts` was added. Guard application (`AuthGuard` on each new route), 400 on invalid UUID via `SpaceIdPipe`, 401/403 paths, and `returnUrl` validation aren't exercised at HTTP level. The pattern from `billing-webhook-auth.guard.integration.spec.ts` would fit. Adding at least one integration test per new route would catch decorator/guard regressions the unit tests can't.

Done [here](https://github.com/safe-global/safe-client-gateway/pull/3269/commits/f4a251a8d6f024d167324fd1e4851963682515e3) 

## Changed

**Security: open redirect via `returnUrl`**
Reused the existing OIDC redirect-validation helper (`resolveAndValidateRedirectUrl`) instead of introducing new billing-specific config, so `returnUrl` on `session-url` and `checkout-url` is now restricted to the same allowed origin as the post-login redirect.

**Correctness: de-dup precedence in `getSpacePaymentLinks`**
The merge order was inverted. A duplicate id kept the general link and discarded the space-specific one. Swapped the order so the space-specific link wins.

**Code quality: no format validation on `planId`/`paymentLinkId`/`sessionId`**
Added a regex-based `ValidationPipe` (alphanumeric + `_`/`-`) on all three.

**Tests**
- Added `billing.controller.e2e-spec.ts`: guard checks on every endpoint, 401/403/400/422 paths, returnUrl allow/deny, and the de-dup fix, all exercised end-to-end over real HTTP.
- De-dup unit test now uses two distinct objects (same id, different `active`) instead of the same object twice, so it actually locks the precedence behavior in.

           
## Not changed                                                                                                                                                                                                                                     
- **Security: session ownership check on `GET /sessions/:sessionId`**                                                                                                                                                                               Investigated in depth, including reading billing-system's source. There's no reliable way to map a session `client_reference_id` is only set by billing-system when the Stripe customer is *newly created*, once a space has an existing customer (the common case), the session carries `customer` (Stripe's own id) instead, and billing-system exposes no reverse lookup from that id to our `upstreamCustomerId`. Left the endpoint as-is (any authenticated caller can fetch it). Same accepted gap as elsewhere in the auth service.                                                                                                                                                                                                                                     

- **Code quality — double `@Param('spaceId', ...)` per endpoint**
Intentional, not redundant: one resolves the numeric id via `SpaceIdPipe` for the membership check, the other reads the raw UUID for the upstream API call. The pipe only runs once (only one of the two `@Param`s references it), so there's no extra DB round-trip. Matches the existing `space-safes.controller.ts` pattern.

yagopv (dismissed):

src/datasources/billing-api/entities/plan.entity.ts
```diff
@@ -1,23 +1,53 @@
...
+
+export type Product = z.infer<typeof ProductSchema>;
+
+export const ProductSchema = z.object({
+  id: z.string(),
+  active: z.boolean(),
+  description: z.string(),
+  marketingFeatures: z.array(MarketingFeatureSchema),
+  // Generic, like Stripe metadata itself: see PaymentLinkMetadataSchema.
+  metadata: z.record(z.string(), z.string()),
```
tmjssz: Might make sense to re-use `PaymentLinkMetadataSchema` (with a more generic name) here and other places?

src/modules/billing/routes/billing.service.ts
```diff
@@ -0,0 +1,127 @@
...
+    await this.assertSpaceMember(args.spaceId, args.authPayload);
+
+    return await this.billingApi.createCheckoutSession({
+      paymentLinkId: args.paymentLinkId,
+      upstreamCustomerId: args.spaceUuid,
+      returnUrl: this.validateReturnUrl(args.returnUrl),
+    });
+  }
+
+  public async getCheckoutSession(sessionId: string): Promise<CheckoutSession> {
```
tmjssz, line 110: Shouldn't this function also have a `assertSpaceMember` assertion, like the other functions in this file?

src/datasources/billing-api/billing-api.service.ts
```diff
@@ -92,14 +100,44 @@ export class BillingApi implements IBillingApi {
...
-      url: `${this.baseUri}/api/v1/customers/${args.upstreamCustomerId}`,
+      url: `${this.baseUri}/api/v1/customers/${stripDashes(args.upstreamCustomerId)}`,
       schema: z
         .object({ customer: CustomerSchema })
         .transform((body) => body.customer),
       expireTimeSeconds: this.billingExpireTimeSeconds,
     });
   }
 
+  /** Not cached. Bypasses {@link NetworkService}, whose response.json() parsing can't handle this endpoint's plain-text body. */
```
tmjssz: If we bypass NetworkService I think we also skip things like logging and circuit breaker. Maybe consider extending NetworkService by an option to avoid response.json()

src/modules/billing/routes/billing.service.ts
falvaradorodriguez: Good catch! This was intentional, as explained in the PR description.

This endpoint only receives a sessionId, but assertSpaceMember requires a spaceId. The Checkout Session does not always provide a reliable way to identify the related space. client_reference_id is only set when a new Stripe customer is created. In the usual case, the session only contains Stripe’s customer ID, and billing-system does not provide a way to map it back to our upstreamCustomerId.

Adding a spaceId to the URL would not fully solve the problem either, because a caller could provide any valid spaceId they belong to, even if it is unrelated to the requested session.

src/datasources/billing-api/entities/plan.entity.ts
falvaradorodriguez: Yep, totally agree, fixed https://github.com/safe-global/safe-client-gateway/pull/3269/changes/4498ea4b161425c980291e01ac0b4c10806187f2

src/datasources/billing-api/billing-api.service.ts
falvaradorodriguez: Done here https://github.com/safe-global/safe-client-gateway/pull/3269/changes/1ff34495ebe277858085e6696e87e7fff8da32cd

tmjssz (approved):

tmjssz (approved):

PR #3290, opened 2026-07-22
chore(deps): upgrade typeorm to 1.1.0

PooyaRaki:
## Summary

  - Upgrades typeorm from 0.3.28 to 1.1.0, the current stable release line.
  - Primary driver: an unpatched advisory affecting UpdateQueryBuilder/SoftDeleteQueryBuilder — the order argument passed to orderBy/addOrderBy on those builders is concatenated into generated SQL without validation, permitting blind SQL injection via the sort direction
  (SelectQueryBuilder.orderBy was already unaffected). Fixed upstream in 0.3.29/1.0.0.
    - Verified this was not currently exploitable in this codebase: this project only connects to Postgres (no MySQL/MariaDB driver, and the advisory is scoped to those), and the only orderBy/addOrderBy call in the codebase runs on a SelectQueryBuilder with a hardcoded literal, not user
  input. Upgrading anyway as defense-in-depth since it's a direct dependency of our ORM.
  - Went with 1.1.0 (latest) rather than the minimal 0.3.29 patch per request, picking up additional upstream fixes (aggregate-column resolution, an eager-loading bug that could overwrite manually-set relations, a migration query-runner leak, returning option for update/upsert).

##  Breaking changes migrated

  - typeorm/driver/postgres/PostgresConnectionOptions was renamed to typeorm/driver/postgres/PostgresDataSourceOptions — updated the one import/type usage in src/config/entities/postgres.config.ts.
  - FindOptionsRelations no longer accepts the string-array shorthand (relations: ['x']); TypeORM 1.x requires the object form (relations: { x: true }). Updated all 4 affected call sites in notifications/spaces repositories and services.
  - Updated the corresponding assertion in spaces.service.spec.ts to match the new relations shape.

##  Test plan

  - [x] nest build — passes with no TypeScript errors.
  - [x] yarn test:unit — full suite green: 336 files / 5727 passed / 18 skipped (pre-existing skips, unrelated).
  - [ ] yarn test:integration / yarn test:e2e — not run in this environment (require live Postgres/Redis); please run in CI before merge.
  - [ ] Sanity-check any code relying on eager-loading + manually-set relations together, given the upstream fix changes that interaction — no such pattern found during review, but worth a second look in CI.

### DIFF
--- a/src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
+++ b/src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
@@ -157,19 +157,13 @@ describe('CounterfactualSafesRepository', () => {
   afterEach(async () => {
     vi.resetAllMocks();
 
-    // Delete in dependency order
-    await dbCounterfactualSafeUserRepo
-      .createQueryBuilder()
-      .delete()
-      .where('1=1')
-      .execute();
-    await dbCounterfactualSafeRepo
-      .createQueryBuilder()
-      .delete()
-      .where('1=1')
-      .execute();
-    await dbWalletRepo.createQueryBuilder().delete().where('1=1').execute();
-    await dbUserRepo.createQueryBuilder().delete().where('1=1').execute();
+    // Delete in dependency order. No .where() is intentional: these clear
+    // every row to reset state between tests, and TypeORM rejects a
+    // WHERE clause that renders as an unfiltered '1=1' as a likely mistake.
+    await dbCounterfactualSafeUserRepo.createQueryBuilder().delete().execute();
+    await dbCounterfactualSafeRepo.createQueryBuilder().delete().execute();
+    await dbWalletRepo.createQueryBuilder().delete().execute();
+    await dbUserRepo.createQueryBuilder().delete().execute();
   });
 
   afterAll(async () => {
--- a/src/modules/spaces/domain/safes/space-safes.repository.integration.spec.ts
+++ b/src/modules/spaces/domain/safes/space-safes.repository.integration.spec.ts
@@ -127,20 +127,17 @@ describe('SpaceSafesRepository', () => {
     await dbMembersRepository
       .createQueryBuilder()
       .delete()
-      .where('1=1')
       .execute();
     await dbSpaceSafesRepository
       .createQueryBuilder()
       .delete()
-      .where('1=1')
       .execute();
     await dbSpaceRepository
       .createQueryBuilder()
       .delete()
-      .where('1=1')
       .execute();
-    await dbWalletRepo.createQueryBuilder().delete().where('1=1').execute();
-    await dbUserRepo.createQueryBuilder().delete().where('1=1').execute();
+    await dbWalletRepo.createQueryBuilder().delete().execute();
+    await dbUserRepo.createQueryBuilder().delete().execute();
   });
 
   afterAll(async () => {
--- a/src/modules/spaces/domain/safes/space-safes.repository.integration.spec.ts
+++ b/src/modules/spaces/domain/safes/space-safes.repository.integration.spec.ts
@@ -123,7 +123,10 @@ describe('SpaceSafesRepository', () => {
   afterEach(async () => {
     vi.resetAllMocks();
 
-    // Delete in dependency order to avoid deadlocks
+    // Delete in dependency order to avoid deadlocks. No .where(): TypeORM
+    // 1.1.0 rejects a WHERE clause that renders as an unfiltered '1=1' as
+    // a likely mistake — see counterfactual-safes.repository.integration.spec.ts
+    // for the full rationale.
     await dbMembersRepository.createQueryBuilder().delete().execute();
     await dbSpaceSafesRepository.createQueryBuilder().delete().execute();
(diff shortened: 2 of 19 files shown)


PooyaRaki:
@claude review

PooyaRaki:
Verified each point against the code before making changes — all checked out.

  - enum.ts: confirmed from() returns keyof T | null but still throws via getEnumKey for non-null unrecognized values. Kept the current behavior (data corruption should stay loud) and added a comment on databaseEnumTransformer.from() clarifying that only null is treated as absence — a
  non-null value that doesn't match a known enum member is corrupt/legacy data, not absence.
  - Truncate comment duplication: confirmed only counterfactual-safes.repository.integration.spec.ts had the rationale — the other five specs (spaces, members, users, wallets, notifications) had no comment at all, not even an unexplained repeat. Added a one-line pointer comment at each
  site back to the full rationale.
  - whereInIds vs manager.delete inconsistency: confirmed users.repository.ts:294 still uses manager.delete(DbUser, userId), unaffected since it's a single id. Left as-is — checked TypeORM 1.1.0's source and found no @deprecated marker on whereInIds, so I didn't treat that as an
  established fact to act on.
  - orderBy CVE / relations migration / PostgresDataSourceOptions rename: re-confirmed, no changes needed.
  
  Commit: c6c61fdb

PooyaRaki:
@claude review addressed the remarks

PooyaRaki:
Confirmed — space-safes.repository.integration.spec.ts:126-131 was genuinely missed in the earlier sweep. Added the pointer comment there, and also added it to space-audit.repository.integration.spec.ts:149-160, since its existing comment covers a different concern (why audit rows
  aren't deleted) and didn't explain the WHERE-1=1 pattern either.

  Commit: 71f967f2

src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
```diff
@@ -157,19 +157,13 @@ describe('CounterfactualSafesRepository', () => {
...
-    await dbCounterfactualSafeRepo
-      .createQueryBuilder()
-      .delete()
-      .where('1=1')
-      .execute();
-    await dbWalletRepo.createQueryBuilder().delete().where('1=1').execute();
-    await dbUserRepo.createQueryBuilder().delete().where('1=1').execute();
+    // Delete in dependency order. No .where() is intentional: these clear
+    // every row to reset state between tests, and TypeORM rejects a
+    // WHERE clause that renders as an unfiltered '1=1' as a likely mistake.
```
tongyaosafe, line 162: I think this comment is bit redundent

src/modules/spaces/domain/safes/space-safes.repository.integration.spec.ts
```diff
@@ -123,24 +123,15 @@ describe('SpaceSafesRepository', () => {
...
-      .execute();
-    await dbSpaceRepository
-      .createQueryBuilder()
-      .delete()
-      .where('1=1')
-      .execute();
-    await dbWalletRepo.createQueryBuilder().delete().where('1=1').execute();
-    await dbUserRepo.createQueryBuilder().delete().where('1=1').execute();
+    // Delete in dependency order to avoid deadlocks. No .where(): TypeORM
+    // 1.1.0 rejects a WHERE clause that renders as an unfiltered '1=1' as
```
tongyaosafe: Here also. I think we don't need to explain explicitly

src/modules/spaces/domain/spaces.repository.integration.spec.ts
```diff
@@ -119,30 +119,29 @@ describe('SpacesRepository', () => {
   afterEach(async () => {
     vi.resetAllMocks();
 
-    // Delete in dependency order to avoid deadlocks
+    // Delete in dependency order to avoid deadlocks. No .where(): TypeORM
+    // 1.1.0 rejects a WHERE clause that renders as an unfiltered '1=1' as
```
tongyaosafe: here the same

compojoom (dismissed):

tongyaosafe (dismissed):
Approved. I think the comments are bit redundant. Maybe we can simplify them?

src/modules/spaces/domain/spaces.repository.integration.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3290/commits/23d8b06974be2826be7342393ffdcccb79112ea1

src/modules/spaces/domain/safes/space-safes.repository.integration.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3290/commits/23d8b06974be2826be7342393ffdcccb79112ea1

src/modules/counterfactual-safes/domain/counterfactual-safes.repository.integration.spec.ts
PooyaRaki: https://github.com/safe-global/safe-client-gateway/pull/3290/commits/23d8b06974be2826be7342393ffdcccb79112ea1

tongyaosafe (approved):
🚀

PR #3294, opened 2026-07-22
build(deps): bump tmp to 0.2.7

PooyaRaki:
## Summary
  
  Bumps the transitive `tmp` dependency from `0.0.33` to `0.2.7`.

  `tmp` is pulled in only via dev tooling (`@nestjs/cli` / `@angular-devkit/schematics-cli` → `@inquirer/prompts` → `@inquirer/editor` → `external-editor`), which requests the outdated `^0.0.33` range. Since that range can't resolve forward on its own, a `resolutions` override is added
  to `package.json` to pull the current release. This dedupes the tree onto a single `tmp@0.2.7` and drops the now-unneeded `os-tmpdir` sub-dependency.
  
  The bump is API-compatible: `external-editor` uses `tmp.tmpNameSync()`, which is unchanged in `0.2.7`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,5 +116,8 @@
   },
   "main": "main.ts",
   "repository": "https://github.com/5afe/safe-client-gateway-nest.git",
-  "packageManager": "yarn@4.14.1"
+  "packageManager": "yarn@4.14.1",
+  "resolutions": {
+    "tmp": "0.2.7"
+  }
 }
--- a/yarn.lock
+++ b/yarn.lock
@@ -6332,13 +6332,6 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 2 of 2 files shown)


compojoom (approved):

package.json
```diff
@@ -116,5 +116,8 @@
   },
   "main": "main.ts",
   "repository": "https://github.com/5afe/safe-client-gateway-nest.git",
-  "packageManager": "yarn@4.14.1"
+  "packageManager": "yarn@4.14.1",
```
compojoom, line 119: @PooyaRaki it is better to pin against a hash.

PooyaRaki: Agree. I'll open another PR for that 👍

PR #3312, opened 2026-07-28
fix: accept isFailed on EXECUTED_MULTISIG_TRANSACTION events (IR-46)

gfendres:
## Summary

Fixes [IR-46](https://app.datadoghq.eu/incidents/46).

|  |  |
|---|---|
| **Symptom** | Executed transactions missing from the wallet for ~10 min. Users can't create the next tx. |
| **Cause** | tx-service dropped the `failed` field ([#2964](https://github.com/safe-global/safe-transaction-service/pull/2964), merged 09:13 UTC today). Our schema still required it. |
| **Effect** | Every `EXECUTED_MULTISIG_TRANSACTION` event fails `EventSchema.parse` and gets thrown away. |
| **Blast radius** | Every executed multisig tx, all chains. Push notifications too. |
| **Fix** | Read `isFailed`, which tx-service has been sending since [#2907](https://github.com/safe-global/safe-transaction-service/pull/2907) (6 weeks ago). |

### What's happening

```mermaid
flowchart LR
  A["tx-service<br/><b>isFailed: false</b>"] --> B{"EventSchema.parse<br/>requires 'failed'"}
  B -->|"ZodError"| C["catch → log<br/><b>message dropped</b>"]
  C --> D["nothing clears cache"]
  D --> E["stale queue, history<br/>and <b>nonce</b> for ~10 min"]
  style C fill:#ffdddd,stroke:#cc0000
  style E fill:#ffdddd,stroke:#cc0000
```

The event is the only thing that triggers invalidation. Drop it and the cache just sits there until its TTL runs out. That's the 10 minutes people were waiting.

Real payload from production, no `failed` anywhere:

```json
{"type":"EXECUTED_MULTISIG_TRANSACTION","isFailed":false,"chainId":"137",
 "address":"0x1286bb...","safeTxHash":"0x1a033c...","txHash":"0xa4e8fb..."}
```

### Why it looked like a multisend problem

| Executed tx | Transfer event fires? | Queue + history cleared | Nonce cleared |
|---|---|---|---|
| Send ETH / tokens | Yes | ✅ by the transfer event | ❌ |
| Zero-value multiSend | No | ❌ | ❌ |
| addOwner / changeThreshold | No | ❌ | ❌ |

`OUTGOING_ETHER` / `OUTGOING_TOKEN` clear the transaction lists themselves, so transfers papered over the bug. Multisends emit no transfer, so nothing cleared them and they got noticed first.

But `clearSafe` (the nonce) only ever runs from this event and `MODULE_TRANSACTION`. So the stale nonce, which is what actually blocked people from creating a tx, hit **all three rows**.

## Changes

| File | What |
|---|---|
| `executed-transaction.schema.ts` | `failed` → `isFailed`, optional. |
| `notification.entity.ts` | Notification type can't alias the event schema anymore, the two contracts differ. Same shape as `IncomingTokenNotification`. |
| `push-notification.service.ts` | Reads `isFailed`, still sends `failed` as a string. |
| `executed-transaction.schema.spec.ts` | `isFailed` true / false / absent / non-boolean. |
| `hooks-cache.integration.spec.ts` | Safe (nonce) + queue caches actually clear, for every payload shape. |
| `push-notification.service.spec.ts` | Stringified flag per case. |
| 20 test payloads across 2 specs | `failed` → `isFailed`. Zod was stripping the old key, so they passed while sending nothing. |

Two deliberate calls:

- **`isFailed` is optional.** Nothing in cache invalidation reads it, it only feeds the notification string. Missing flag costs a wrong `failed: 'false'` in one push. Requiring it costs the dropped event and stale cache that caused this incident. Easy to flip if reviewers prefer strictness.
- **Push payload unchanged.** Firebase's `data` is `Record<string, string>` and the apps already read `data.failed`, so it goes out byte for byte the same. No client work.

## Verification

| Check | Result |
|---|---|
| Test reproduces the bug | ✅ Put `failed` back as required and the new test fails on exactly the payloads production sends |
| Production payload parses | ✅ Ran the payload above through `EventSchema` |
| `src/modules/hooks` + notification entities | ✅ 337 passing |
| Lint + format | ✅ Clean |
| `tsc` | ✅ No new errors (110 pre-existing, none in touched files) |

## Not covered

- `event-hooks-queue.e2e-spec.ts` and `notifications.repository.integration.spec.ts` need Redis / RabbitMQ / Postgres, so they skip here. Payloads in the e2e one are updated to `isFailed`.
- The 10 minutes comes from `EXPIRATION_TIME_DEFAULT_SECONDS`. I didn't read the deployed value, that's just what people observed.
- **Unrelated:** safe-decoder-service chokes on some multiSend calldata with `InsufficientDataBytes`. ~10/day for at least two weeks, so not from today. Not a CGW problem, our viem decoder handles the same payload fine. Worth its own issue.

### DIFF
--- a/src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
+++ b/src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
@@ -10,7 +10,14 @@ export const ExecutedTransactionEventSchema = HookEventBaseSchema.extend({
   to: AddressSchema,
   safeTxHash: HexSchema,
   txHash: HexSchema,
-  failed: z.enum(['true', 'false']),
+  // The Transaction Service moved the execution status from the stringified
+  // `failed` to the boolean `isFailed`. Both are optional so that an event from
+  // either version parses: the queue consumer discards any event that fails
+  // validation, which would leave every cache for this Safe — including its
+  // nonce — stale until it expires on its own TTL. Read via
+  // `isExecutedTransactionFailed`.
+  isFailed: z.boolean().optional(),
+  failed: z.enum(['true', 'false']).optional(),
   // FirebaseNotification['data'] does not accept null values
   data: z.preprocess((val) => val ?? undefined, HexSchema.optional()),
 });
@@ -18,3 +25,19 @@ export const ExecutedTransactionEventSchema = HookEventBaseSchema.extend({
 export type ExecutedTransactionEvent = z.infer<
   typeof ExecutedTransactionEventSchema
 >;
+
+/**
+ * Resolves the execution status of an executed transaction event across both
+ * Transaction Service payload versions.
+ *
+ * Falls back to `false` when neither field is present: an executed transaction
+ * that reports no status is a successful one.
+ *
+ * @param event - the executed transaction event to read the status from
+ * @returns whether the on-chain execution reverted
+ */
+export function isExecutedTransactionFailed(
+  event: Pick<ExecutedTransactionEvent, 'isFailed' | 'failed'>,
+): boolean {
+  return event.isFailed ?? event.failed === 'true';
+}
--- a/src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
+++ b/src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
@@ -10,11 +10,7 @@ export const ExecutedTransactionEventSchema = HookEventBaseSchema.extend({
   to: AddressSchema,
   safeTxHash: HexSchema,
   txHash: HexSchema,
-  // Optional on purpose: nothing in cache invalidation reads the execution
-  // status, so a missing flag must never reject the event. The queue consumer
-  // discards whatever fails validation, which leaves every cache for this Safe,
-  // its nonce included, stale until it expires on its own TTL.
-  isFailed: z.boolean().optional(),
+  isFailed: z.boolean(),
   // FirebaseNotification['data'] does not accept null values
   data: z.preprocess((val) => val ?? undefined, HexSchema.optional()),
 });
--- a/src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
+++ b/src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
@@ -10,14 +10,11 @@ export const ExecutedTransactionEventSchema = HookEventBaseSchema.extend({
   to: AddressSchema,
   safeTxHash: HexSchema,
   txHash: HexSchema,
-  // The Transaction Service moved the execution status from the stringified
-  // `failed` to the boolean `isFailed`. Both are optional so that an event from
-  // either version parses: the queue consumer discards any event that fails
-  // validation, which would leave every cache for this Safe — including its
-  // nonce — stale until it expires on its own TTL. Read via
-  // `isExecutedTransactionFailed`.
+  // Optional on purpose: nothing in cache invalidation reads the execution
+  // status, so a missing flag must never reject the event. The queue consumer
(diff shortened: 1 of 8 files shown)


src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
```diff
@@ -10,11 +10,34 @@ export const ExecutedTransactionEventSchema = HookEventBaseSchema.extend({
...
   txHash: HexSchema,
-  failed: z.enum(['true', 'false']),
+  // The Transaction Service moved the execution status from the stringified
+  // `failed` to the boolean `isFailed`. Both are optional so that an event from
+  // either version parses: the queue consumer discards any event that fails
+  // validation, which would leave every cache for this Safe — including its
+  // nonce — stale until it expires on its own TTL. Read via
+  // `isExecutedTransactionFailed`.
+  isFailed: z.boolean().optional(),
+  failed: z.enum(['true', 'false']).optional(),
```
Uxio0: `isFailed` was already deployed months ago. No need to keep `failed`

Uxio0 (commented):
199 lines look like a lot for a simple field renaming and type change

src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
gfendres: You're right, dropped it in b496d01. Checked the history: `isFailed` landed in #2907 on 2026-06-16, so nothing has been sending only `failed` for six weeks.

Two knock-on things worth a look:

**The notification type can't alias the event schema anymore.** `ExecutedMultisigTransactionNotification` was `= ExecutedTransactionEvent`, and removing `failed` from the schema made it a type error. The two contracts genuinely differ now: the event has a boolean `isFailed`, the push payload still sends a stringified `failed` because Firebase's `data` is `Record<string, string>` and the apps already read that key. Modelled it the same way as `IncomingTokenNotification`:

```ts
export type ExecutedMultisigTransactionNotification = Omit<
  ExecutedTransactionEvent,
  'isFailed'
> & { failed: 'true' | 'false' };
```

So the outgoing notification is unchanged, no client work.

**Test payloads were passing for the wrong reason.** Once `failed` left the schema, Zod stripped it as an unknown key, so ~20 specs still building `failed: 'true'` kept passing while sending nothing. Switched them all to `isFailed`.

I did keep `isFailed` optional rather than required. Nothing in cache invalidation reads the flag, it only feeds the notification string, so a missing one costs us a wrong `failed: 'false'` in a push. A required one costs us the dropped event and stale cache that caused this incident. Happy to make it required if you'd rather have the strictness, it's a one word change.

---

LucieFaire (dismissed):

src/modules/hooks/routes/entities/schemas/executed-transaction.schema.ts
```diff
@@ -10,7 +10,11 @@ export const ExecutedTransactionEventSchema = HookEventBaseSchema.extend({
   to: AddressSchema,
   safeTxHash: HexSchema,
   txHash: HexSchema,
-  failed: z.enum(['true', 'false']),
+  // Optional on purpose: nothing in cache invalidation reads the execution
+  // status, so a missing flag must never reject the event. The queue consumer
+  // discards whatever fails validation, which leaves every cache for this Safe,
+  // its nonce included, stale until it expires on its own TTL.
+  isFailed: z.boolean().optional(),
```
PooyaRaki: it's a required field, It shouldn't be optional, if it's not sent it means that the event received is malformed.

LucieFaire (approved):

Uxio0 (approved):

PR #3315, opened 2026-07-31
feat: add step-up authentication for sensitive workspace actions

TimDaub:
Merging: Do not turn on feature flag before changes have landed in wallet-monorepo

## Summary

Milestone 2 of [Workspace 2FA](https://app.notion.com/p/safe-global/Workspace-2FA-3978180fe5738087a0c8ffedbd17e925) (WA-2725): sensitive Workspace actions now require a second factor presented within a short elevation window, enforced server-side.

A session carries a single new claim, `mfa_verified_at`, stamped from the gateway's own clock at the OIDC callback whenever the Auth0 ID token's `amr` proves a challenge was passed. Auth0 injects `amr` **only for the transaction in which the user actually presented a factor** — it is absent when an existing session is reused — which makes it the freshness proof step-up needs. Because MFA is mandatory on every login, a plain login stamps the claim too, so the minutes right after signing in are already elevated and users are not challenged twice.

Auth0's `auth_time` is deliberately **not** used: it records when the SSO session was established, and a step-up inside that session does not advance it, so it cannot express step-up freshness. The tech spike carried `amr`, `acr`, `auth_time` and `mfa_verified_at`; only the last is persisted, and `amr` is consumed transiently at the callback.

## Changes

- `mfa_verified_at` on the OIDC auth payload, plus `AuthPayload.hasFreshMfa()` (fails closed on an absent claim, an expired window, and future stamps from clock skew)
- `ElevationGuard`, returning `403 elevation_required` so clients can distinguish "redo MFA" from a plain authorization failure
- `?elevate=true` on `GET /v1/auth/oidc/authorize` requests the multi-factor `acr_values`; the callback independently rejects any elevation whose token lacks `amr: ["mfa"]`
- `AUTH_ELEVATION_WINDOW_SECONDS` (default 30 min)
- `elevation.e2e-spec.ts` pins the gated **and** ungated lists below, so changing either is always a deliberate edit rather than a side effect of touching a controller

### Routes now requiring a fresh second factor

| Method | Route | Action |
| --- | --- | --- |
| `PATCH` | `/v1/spaces/:id` | Rename space |
| `DELETE` | `/v1/spaces/:id` | Delete space |
| `POST` | `/v1/spaces/:spaceId/members/invite` | Invite member |
| `PATCH` | `/v1/spaces/:spaceId/members/:userId/role` | Change member role |
| `DELETE` | `/v1/spaces/:spaceId/members/:userId` | Remove member |
| `POST` | `/v1/spaces/:spaceId/safes` | Add Safe to space |
| `DELETE` | `/v1/spaces/:spaceId/safes` | Remove Safe from space |
| `PUT` | `/v1/spaces/:spaceId/address-book` | Add/edit shared contact |
| `DELETE` | `/v1/spaces/:spaceId/address-book/:address` | Remove shared contact |
| `PUT` | `/v1/spaces/:spaceId/address-book/requests/:requestId/approve` | Approve contact request |

### Mutating routes deliberately left ungated

| Method | Route | Why not |
| --- | --- | --- |
| `POST` | `/v1/spaces` | Creating a space has no existing state to tamper with; gating it would put a challenge in the middle of onboarding |
| `POST` | `/v1/spaces/:spaceId/address-book/requests` | A proposal only — the admin approval that acts on it is gated |
| `PUT` | `/v1/spaces/:spaceId/address-book/requests/:requestId/reject` | Denial-only; an attacker gains nothing but griefing |
| `POST` | `/v1/spaces/:spaceId/members/:userId/invite/renew` | Re-sends an invite that a gated call already created |
| `PATCH` | `/v1/spaces/:spaceId/members/alias` | The caller's own display name |
| `DELETE` | `/v1/spaces/:spaceId/members` | Self-removal; recoverable by re-invite |
| POST | /v1/spaces/:spaceId/members/accept | Responding to an invite only moves the caller in; the invite itself is gated |
| POST | /v1/spaces/:spaceId/members/decline | Declining only affects the caller's own pending invite |

All `GET` routes are unaffected.

`ElevationGuard` always runs after `AuthGuard` (which attaches the session payload) and before the per-route rate-limit guards.

Sign-In-with-Ethereum sessions are exempt — they never pass through the OIDC provider and carry no MFA proof, so gating them would lock SIWE admins out of Workspace administration. Extending step-up to them is Milestone 3 (WA-2852).

## Verified against a real Auth0 tenant

The design rests on one vendor claim — that `amr` is transaction-scoped rather than session-scoped. That was verified empirically rather than taken from the docs, by probing the raw claim on a dev tenant with `mfa_policy = "all-applications"`:

| Scenario | Challenge shown | `amr` observed | `mfa_verified_at` |
| --- | --- | --- | --- |
| Login, OTP entered | yes | `["mfa"]` | stamped |
| Remembered device, plain `/authorize` | **none** | **`undefined`** | absent |
| Remembered device, `?elevate=true`, no challenge | none | `undefined` | **rejected** (`authentication_failed`) |
| Remembered device, `?elevate=true`, challenge forced | OTP | — | stamped |
| `?elevate=true` → recovery code | recovery code | — | stamped |

Row 2 is the dangerous case: a remembered browser skipping the challenge seconds after a successful MFA. Auth0 did **not** carry `amr` forward, so no session is elevated without a code. Row 3 is the attack — a client that requests elevation but arranges for no challenge to happen — and it fails closed. Row 5 confirms recovery-code users can elevate, so a lost authenticator is not a lockout.

Also verified: exactly one challenge per round-trip (no double prompt), and the `403 elevation_required` → step-up → retry cycle end to end.

## Note on the Auth0 side

`acr_values` is accompanied by a post-login action in [terraform-auth0-module#24](https://github.com/safe-global/terraform-auth0-module/pull/24). That action is **defence in depth, not a prerequisite** for this PR: while `mfa_policy = "all-applications"` is in force, Auth0 was measured honouring the multi-factor `acr_values` on its own, overriding a remembered browser, with no action deployed. The action keeps step-up working if that policy is ever relaxed to `never`/`confidence-score`. Either way this PR fails closed, because the callback requires `amr` regardless of how the challenge was produced.

## How to test it

1. Set `AUTH_ELEVATION_WINDOW_SECONDS=60` and sign in via OIDC — the session is elevated for 60s.
2. Call a gated route (e.g. `POST /v1/spaces/:spaceId/members/invite`) → succeeds.
3. Wait out the window, call it again → `403 elevation_required`.
4. `GET /v1/auth/oidc/authorize?elevate=true&redirect_url=...`, complete the challenge, retry → succeeds.
5. Repeat step 3 with a SIWE session → succeeds without any challenge (M3 scope).

Automated: `yarn test:unit`, plus `yarn vitest run --project e2e src/modules/spaces/routes/elevation.e2e-spec.ts` (70 cases across the gated/ungated matrix).

## Visual summary

```mermaid
sequenceDiagram
    participant C as Client
    participant G as CGW
    participant A as Auth0

    C->>G: POST /spaces/:id/members/invite
    G->>G: AuthGuard, then ElevationGuard
    G--xC: 403 elevation_required

    C->>G: GET /auth/oidc/authorize?elevate=true
    G->>C: 302 (acr_values=multi-factor, elevate in signed state)
    C->>A: /authorize
    A->>C: OTP challenge (remember-browser suppressed)
    C->>A: code
    A->>G: /callback?code&state
    G->>A: exchange code
    A->>G: id_token (amr=["mfa"])
    Note over G: amr proves a challenge happened now, so<br/>stamp mfa_verified_at from the server clock.<br/>No amr on an elevate round-trip means reject.
    G->>C: Set-Cookie access_token (mfa_verified_at)

    C->>G: POST /spaces/:id/members/invite (retry)
    G->>C: 200
```

## Feature flag
- Step up authentication can be turned on and off with `features.mfaStepUp`


PooyaRaki:
@claude arch-review

PooyaRaki:
@claude review

TimDaub:
Thanks @PooyaRaki — review findings addressed in `bfdb584c` (rebased onto the `main` merge). Disclosure: I wrote most of this PR, so this is the author responding, not an independent second opinion.

### Fixed

**Env vars missing from `RootConfigurationSchema`** — correct, and the described failure is exact. `AUTH_ELEVATION_WINDOW_SECONDS` and `FF_MFA_STEP_UP` are now declared. ~~I also added `AUTH_CLOCK_SKEW_SECONDS`, which was already unvalidated but which this PR newly puts on the same NaN-sensitive path (see below). Covered by tests in `configuration.validator.spec.ts` that reject `not-a-number`, `0` and `-1` at boot — verified non-vacuous by removing the schema line and confirming they fail.~~

**Feature flag owner/removal condition** — added, naming WA-2725 as owner and WA-2726 shipping as the removal condition. Per `configuration-and-flags.md` no mapping carries the full comment yet, so this is the first.

**Split `@UseGuards` on the address-book PUT** — consolidated to `@UseGuards(AuthGuard, ElevationGuard, SpacesAddressBookRateLimitGuard)`. Same resulting order, one decorator to read.

**R-002 `as never` / R-005 `.mock.calls`** — both gone; the token builder is typed `Auth0Token` and the assertions go through `toHaveBeenCalledWith`.

### ~~Also found while answering this~~

~~`hasFreshMfa` rejected *any* future stamp. The stamp is written by whichever instance handles the OIDC callback and read by whichever handles the next request, so a slightly negative age is ordinary clock skew, not a forged stamp — the gateway would have 403'd a session seconds after it passed a real challenge, and the retry could land on the same skewed pair again. It now tolerates `auth.clockSkewSeconds`, the bound SIWE already applies to message timestamps, which keeps the future window bounded rather than removing the check.~~

### Not changed

**`elevation_required` only in `message`** — agreed in principle, but this is the client-facing error contract and the consumer (WA-2726, wallet-monorepo) isn't written yet. Worth deciding deliberately rather than as review cleanup, since changing it later is the expensive version. @TimDaub — your call; happy to move it to a `code` field now if you'd prefer.

**Guards imported from `auth/routes/`** — real per non-negotiable #4, but `AuthGuard` is imported exactly this way by these same controllers today. Following the local convention keeps this PR consistent; re-exporting guards through `auth/domain/` is a repo-wide refactor that shouldn't ride along here.

**Duplicated `ELEVATION_WINDOW_SECONDS`** — deliberate. Both specs use 5 minutes precisely because it differs from the 30-minute production default: a shared constant equal to the default would let a guard that hardcoded the value instead of reading config still pass. Each file now says so.

**PR title and body sections** — @TimDaub's to make.

Full unit (348 files / 5892) and integration (81 files / 1529, including the 70-case elevation matrix) pass on the rebased head.

---

src/modules/auth/oidc/routes/oidc-auth.service.ts
```diff
@@ -65,8 +67,24 @@ export class OidcAuthService {
...
+  }
+
+  /**
+   * @param elevate - Whether this callback completes a step-up round-trip. When
+   *   true the provider must have performed a multi-factor challenge, and the
+   *   request is rejected if it did not.
+   */
   public async authenticateWithOidc(
     code: string,
+    { elevate = false }: { elevate?: boolean } = {},
```
PooyaRaki: Nit: This is very hard to read, we should either extract the type to a proper type or change it to something like:
```suggestion
   elevate: boolean = false,
```

src/modules/auth/oidc/auth0/datasources/auth0-api.service.ts
```diff
@@ -30,6 +33,12 @@ const ManagementApiTokenResponseSchema = z.object({
 export class Auth0Api implements IAuth0Api {
   private static readonly AUTHORIZATION_CODE_GRANT_TYPE = 'authorization_code';
   private static readonly CLIENT_CREDENTIALS_GRANT_TYPE = 'client_credentials';
+  /**
+   * OpenID Provider Authentication Policy Extension policy asserting that the
+   * end user authenticated with more than one factor.
+   */
+  private static readonly MULTI_FACTOR_ACR_VALUE =
```
PooyaRaki, line 40: Shall we move it to the config (env) with a default value or do you think it will always stay the same?

PooyaRaki (dismissed):

PooyaRaki:
@claude arch-review

src/modules/auth/oidc/routes/oidc-auth.service.ts
TimDaub: Addressed here [d990f94](https://github.com/safe-global/safe-client-gateway/pull/3315/commits/d990f94207493dbfddba8cf39ea237061dc85129)

src/modules/auth/oidc/auth0/datasources/auth0-api.service.ts
TimDaub: I have no strong opinion here. I don't really expect this value to change any time though

TimDaub:
> ### Also found while answering this
> `hasFreshMfa` rejected _any_ future stamp. The stamp is written by whichever instance handles the OIDC callback and read by whichever handles the next request, so a slightly negative age is ordinary clock skew, not a forged stamp — the gateway would have 403'd a session seconds after it passed a real challenge, and the retry could land on the same skewed pair again. It now tolerates `auth.clockSkewSeconds`, the bound SIWE already applies to message timestamps, which keeps the future window bounded rather than removing the check.

Removed this again in [4a906b3](https://github.com/safe-global/safe-client-gateway/pull/3315/commits/4a906b377a219adf4a8fe9c47907d75ae3d9d5d1) because it didn't make sense to me since all JWTs are issued by CGW itself, so there can't be multiple clock states.

PooyaRaki:
@claude arch-review

PooyaRaki (dismissed):

TimDaub:
@claude arch-review

TimDaub:
@claude arch-review

TimDaub:
@claude arch-review

PooyaRaki (approved):

PR #3327, opened 2026-08-06
feat: add entitlements data model and billing webhook sync

moisses89:
## Summary

Adds the entitlements data model (feature catalog, subscriptions, subscription entitlements, space feature usage, space seat selection) and keeps it in sync with Stripe via a billing webhook.

An event whose payload is a complete snapshot is materialized straight from it; anything short of that is a trigger only, and the authoritative state is re-fetched before being written. Events are ordered by their own `created` stamp rather than by delivery: `subscriptions.last_event_at` records the stamp behind each row, and a payload that does not order after the newest stamp for the space re-fetches instead of writing — so a `customer.subscription.updated` delivered after a `customer.subscription.deleted` cannot resurrect the subscription. The check runs inside `materialize()`'s transaction behind a `pg_advisory_xact_lock` on the space; re-fetched state is exempt and stamped with the later of its trigger and the stored mark, so the mark never goes down.

That mark is read twice per webhook, so it is a `MAX` served by `IDX_subscriptions_space_id_last_event_at`: `EXPLAIN ANALYZE` over 50k rows gives 0.03 ms whether a space has 10 subscriptions or 20,000, against 29 ms at 20,000 for reading the newest row by ordering. `space_id` stays the leading column, so the index still serves the FK cascade check and any space-scoped read.

Epoch bounds stay deferred: an out-of-range upstream epoch can no longer reach the database (`fromSecondsTimestamp` maps it to null, which triggers the re-fetch), but the shared schema for it also belongs on `jwt-claims.entity.ts`, where it changes auth parsing — so it needs its own security-reviewed PR.

First of three PRs. This one ships the schema — all of it, in one migration, so the follow-ups never alter these tables — and the webhook sync, with nothing reading it over HTTP yet, which is why `EntitlementsModule` is reachable only through `BillingModule` for now. PR 2 adds the HTTP layer (the two endpoints, their DTOs, the quota-exceeded filter, `EntitlementsController` under the existing `routes/`, and the `app.module.ts` entry). PR 3 adds enforcement (`checkQuotaOrFail` in the space-safes and members repositories) and the seat-selection and usage-tracking write paths — the readers for `space_seat_selection` and `space_feature_usage`.

https://linear.app/safe-global/issue/PLA-1643/cgw-entitlement-data-model-features-catalog-subscriptions-billing

## Changes

- `migrations/` — the five tables with their constraints and indexes, including `subscriptions.last_event_at` and the `(space_id, last_event_at)` index
- `src/modules/entitlements/datasources/` — TypeORM entities for the five tables
- `src/modules/entitlements/domain/` — per-table repositories (including the space advisory lock and the ordering-mark read), Zod schemas, the feature-package parser, and the active-status constants
- `src/modules/entitlements/routes/` — `EntitlementsService.materialize()` with the ordering guard, and `SubscriptionSyncService`, plus their test builders
- `src/modules/billing/` — `WebhookEventSchema` and `BillingController.postWebhook` behind `BillingWebhookAuthGuard`
- `src/datasources/billing-api/` — `validUntil` replaced by `currentPeriodStart`/`currentPeriodEnd`, matching what billing sends today


moisses89:
@claude review

moisses89:
@claude arch-review

moisses89:
@claude arch-review

moisses89:
> ## Architecture check
> Applicable guides: ARCHITECTURE.md, best-practices.md, nestjs-patterns.md, api-dtos-and-validation.md, security.md, module-structure.md, caching-and-performance.md, database-and-migrations.md, testing.md, commits-and-prs.md, remarks.md (configuration-and-flags.md and agent-tooling.md: not applicable).
> 
> * `src/modules/entitlements/entitlements.module.ts:7` — Canonical skeleton: the module declares no controller and no HTTP endpoint (the `/v1/spaces/:spaceId/entitlements` routes are deferred), yet hosts `EntitlementsService`/`SubscriptionSyncService` under `routes/`, which the guide admits "only when the module exposes HTTP endpoints".
> * `src/modules/entitlements/domain/entities/space-seat-selection.entity.ts:7` — No speculative abstractions (YAGNI): `space_seat_selection` and `space_feature_usage` ship as tables, TypeORM entities and domain schemas (here an empty `RowSchema.extend({})`) with no repository, reader or writer, built for follow-up PRs the description defers.
> * `src/modules/entitlements/domain/subscription-sync.service.interface.ts:4` — New Symbol+interface pairs need a reason: `ISubscriptionSyncService` has a single implementation, no `test.*.module.ts` swap, and is not the repository Symbol-DI seam the rule exempts.
> * `src/modules/entitlements/domain/entities/space-feature-usage.entity.ts:7` — The 4-representations checklist: `SpaceFeatureUsage` and `SpaceSeatSelection` get a schema, an entity and a migration but no builder under `domain/entities/__tests__/` (testing.md, Builders and faker only).
> * `src/modules/entitlements/routes/subscription-sync.service.spec.ts:178` — R-005: the idempotency assertion destructures `materialize.mock.calls` instead of using the call matchers.
> * PR description — Linking: the body carries no Linear identifier or issue URL.
> * `docs/agents/commits-and-prs.md:80` — No stale guides: the PR-title anti-example still cites this PR's former title (`Feat add subscriptions model`, open PR [feat: add entitlements data model and billing webhook sync #3327](https://github.com/safe-global/safe-client-gateway/pull/3327)), which the retitle invalidated, and the guide is not updated here.
> 
> **Verdict:** ❌ 7 finding(s)

**1. entitlements.module.ts:7 — routes/ without controller**
Deferred by design: the controller lands in the follow-up PR. Moving these files now would only churn them back.
**2. space-seat-selection.entity.ts:7 — YAGNI**
Intentional: the whole layer's schema ships in one migration. Readers land in the follow-up PRs.
**3. subscription-sync.service.interface.ts:4 — Symbol needs a reason**
Required by module-structure.md:73 — billing may only reach this through `domain/`. Removing the Symbol would violate that rule instead.
**4. space-feature-usage.entity.ts:7 — missing builders**
Omitted deliberately: a builder with no reader is dead code. They land with their readers.
**5. subscription-sync.service.spec.ts:178 — R-005**
Fixed.
**6. PR description — Linear link**
Added.
**7. commits-and-prs.md:80 — stale guide**
Out of scope — separate docs PR.

moisses89:
@claude review

moisses89:
> **Claude finished @moisses89's task in 3m 20s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/31490282155)
> 
> ### Claude Code Review
> Focused on things to improve. Prior reviews covered most concerns and they've been addressed (deterministic sort tie‑break, `eager: true` removed, `.mock.calls` destructure gone, FK‑violation race handled, builders in place, etc.). Remaining items are lower-severity.
> 
> #### Correctness
> * **Stale entitlements after cancellation.** `src/modules/entitlements/routes/entitlements.service.ts:108`. `materialize()` only deletes and rewrites entitlements when the batch contains a subscription with `entitlements !== null` (i.e. the currently‑active one). When a subscription transitions from active → canceled and there is no new active row, `toMaterializedSubscriptions` sets `entitlements: null` for every row, so the `subscription_entitlements` rows attached to the just‑canceled subscription are left in the DB. Cascade only cleans them up on space deletion. If PR 2's readers filter strictly by active-status subscription this is invisible, but any join that walks `subscription_entitlements` directly will see phantom rows. Consider deleting entitlements for any subscription being demoted out of `ACTIVE_SUBSCRIPTION_STATUSES` in the same transaction.
> * **Warning log at `entitlements.service.ts:127` drops the event context.** When a purchased entitlement is silently discarded because a feature was renamed/removed between the parse and the re‑fetch, the log line names the space id and the feature key, but not the upstream subscription id or the webhook event id (they aren't threaded down here). This is exactly the "rare and hard to reproduce" case where the identifiers matter — worth passing them into `materialize()` so the trace is actionable.
> 
> #### Code quality
> * **`subscriptions.repository.ts:42-48` — the "TypeORM upsert doesn't return the id on the update path" workaround is real, but the follow‑up `findOneOrFail` selects only `id: true` on an entity whose `INSERT ... ON CONFLICT DO UPDATE` just ran on the same key. `RETURNING id` on the upsert (or `qb.insert().orUpdate().returning('id')`) would fold this into a single round trip, and remove the extra select. Minor: single round trip on the write path.
> * **`subscription-sync.service.ts:120-127` cache invalidation is still coupled to `status: 'all'`.** The prior review already called this out; the coincidence between the caller reading with `'all'` and `deleteByKey` unlinking the whole hash is fragile. A one‑line `CacheRouter.getBillingSubscriptionsHashKey(upstreamCustomerId)` that returns only the parent key would make the "wipe every status" intent explicit and outlive future readers that ask for a different status filter.
> 
> #### Tests
> * **`subscription-sync.service.spec.ts:46-75` `webhookEvent()` helper.** The helper now wraps the shared builders (good), but it still exists as a per‑file convenience wrapper that flattens a few overrides. Once the builders themselves cover the same overrides (they largely do), inlining `webhookEventBuilder().with('type', …).with('data', { … customer: webhookEventCustomerBuilder().with('upstreamCustomerId', …).build() }).build()` at the callsites would remove the middle layer and one more place to keep in sync.
> * **No spec for the "active → canceled with no new active" transition.** The integration spec covers create, idempotency and unknown‑space, but not the transition where the only active subscription becomes canceled. That's the code path that reveals the stale‑entitlement question above; worth a dedicated case in `subscription-sync.integration.spec.ts` regardless of which way the semantics are pinned.
> * [x]  Load review tools
> * [x]  Read key changed files
> * [x]  Analyze correctness, quality, tests
> * [x]  Post review
>   • `feat-add-subscriptions-model`

> **Stale entitlements after cancellation.** Consider deleting entitlements for any subscription being demoted out of ACTIVE_SUBSCRIPTION_STATUSES in the same transaction.

By design: the RFC wants the outgoing package kept.
Now pinned by a test — this exact change fails it.

---

> **Warning log at entitlements.service.ts:127 drops the event context.** Worth passing them into materialize() so the trace is actionable.

Added the upstream subscription id.
Event id would need a signature change — skipped.

---

> **subscriptions.repository.ts:42-48.** RETURNING id on the upsert (or qb.insert().orUpdate().returning('id')) would fold this into a single round trip.

Deliberate workaround, precedented in notifications.repository.
One SELECT on a cold write path. Left as is.

---

> **subscription-sync.service.ts:120-127 cache invalidation is still coupled to status: 'all'.** A one-line CacheRouter.getBillingSubscriptionsHashKey(upstreamCustomerId) would make the intent explicit.

Out of scope: cache.router.ts isn't in this diff.

---

> **subscription-sync.service.spec.ts:46-75 webhookEvent() helper.** Inlining at the callsites would remove the middle layer and one more place to keep in sync.

One helper is one place; fourteen callsites are fourteen.
Keeping it.

---

> **No spec for the "active → canceled with no new active" transition.** Worth a dedicated case in subscription-sync.integration.spec.ts.

Added.

src/datasources/billing-api/entities/subscription.entity.ts
```diff
@@ -40,6 +40,7 @@ export const SubscriptionSchema = z.object({
   startAt: z.number(),
   cancelledAt: z.number().nullable(),
   cancelAt: z.number().nullable(),
-  validUntil: z.number().nullish(),
```
moisses89, line 43: This field didn't exist from the billing service, modified to period Start and End that we need to fill the database.

moisses89:
@claude review

src/modules/entitlements/domain/subscription.mapper.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  return {
+    upstreamSubscriptionId,
+    status,
+    planId,
+    planName: data.metadata?.[PLAN_NAME_METADATA_KEY] ?? null,
+    currentPeriodStart,
+    currentPeriodEnd,
+    entitlements: isActiveSubscriptionStatus(status)
+      ? mapFeaturePackage({
```
falvaradorodriguez, line 83: Here, if the metadata is null or empty, it will delete the entries currently associated with the subscription. This might be expected (IMO). I'm not sure if it's necessary to handle this case.

src/modules/entitlements/domain/feature-package.mapper.ts
```diff
@@ -0,0 +1,108 @@
...
+export function mapFeaturePackage(args: {
+  metadata: StripeMetadata | null | undefined;
+  featureTypeByKey: Map<FeatureKey, FeatureType>;
+  onWarning: (message: string) => void;
+}): Array<ParsedEntitlement> {
+  const packageByKey = new Map<FeatureKey, ParsedEntitlement>();
+
+  for (const [metadataKey, rawValue] of Object.entries(args.metadata ?? {})) {
+    if (!metadataKey.startsWith(FEATURE_METADATA_PREFIX)) {
+      args.onWarning(`Unrecognized metadata key: ${metadataKey}`);
```
falvaradorodriguez: In this case, `planName` will always generate a false warning, right?

migrations/1785406836453-create-entitlements.ts
```diff
@@ -0,0 +1,131 @@
...
+ * `*.entity.db.ts` entities (README convention), then pruned: the generator
+ * also emitted normalization churn for pre-existing tables (constraint/index
+ * renames unrelated to this feature), which was dropped.
+ */
+export class CreateEntitlements1785406836453 implements MigrationInterface {
+  name = 'CreateEntitlements1785406836453';
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(
+      `CREATE TABLE "features" ("id" SERIAL NOT NULL, "key" character varying(64) NOT NULL, "type" character varying(16) NOT NULL, "description" text NOT NULL DEFAULT '', "free_enabled" boolean NOT NULL DEFAULT false, "free_quota" integer, "free_value" character varying(255), "free_period" integer, "created_at" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(), "updated_at" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now(), CONSTRAINT "UQ_features_key" UNIQUE ("key"), CONSTRAINT "CHK_features_type" CHECK ("type" IN ('binary','metered','value')), CONSTRAINT "PK_features_id" PRIMARY KEY ("id"))`,
```
falvaradorodriguez, line 19: IMO, it would be a good idea to create a task to add the necessary features so we don't lose track of that pending task.

src/modules/entitlements/domain/subscription.mapper.ts
```diff
@@ -0,0 +1,126 @@
...
+  const status = parsedStatus.data;
+
+  const currentPeriodStart = fromSecondsTimestamp(data.currentPeriodStart);
+  const currentPeriodEnd = fromSecondsTimestamp(data.currentPeriodEnd);
+  if (
+    currentPeriodStart === null ||
+    (data.currentPeriodEnd != null && currentPeriodEnd === null)
+  ) {
+    args.onWarning(
+      `Billing webhook event ${args.event.id} carries an unrepresentable billing period: ${data.currentPeriodStart}–${data.currentPeriodEnd}`,
```
falvaradorodriguez, line 70: Should we return null here to force a re-fetch?

src/modules/entitlements/routes/subscription-sync.service.ts
```diff
@@ -0,0 +1,186 @@
...
+      }).key,
+    );
+    const features = await this.featuresRepository.getFeatures();
+    const featureTypeByKey = new Map(
+      features.map((feature) => [feature.key, feature.type]),
+    );
+
+    const onWarning = (message: string): void =>
+      this.loggingService.warn(message);
+    const eventSubscription = mapEventToSubscription({
```
falvaradorodriguez, line 177: This only re-fetches when the payload isn't a complete snapshot, so is it possible that a late-delivered subscription.updated (active, old plan) applied after subscription.deleted flips the row back to active and rewrites the old entitlements? If so, with no reconciliation job here, the space would stay wrongly entitled until some later event arrives. WDYT?

moisses89: Fixed https://github.com/safe-global/safe-client-gateway/pull/3327/changes/29cae3f63333e7890d44bd36da65b1c476e79528

src/modules/entitlements/domain/subscription.mapper.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3327/changes/0b9a080d2842e3173e582fc9ef8d2b7889d1e5a3

migrations/1785406836453-create-entitlements.ts
moisses89: I´d say that the feature can be added in a migration in the same PR where we add the enforcement logic.

src/modules/entitlements/domain/subscription.mapper.ts
moisses89: https://github.com/safe-global/safe-client-gateway/pull/3327/changes/0b9a080d2842e3173e582fc9ef8d2b7889d1e5a3

src/modules/entitlements/routes/subscription-sync.service.ts
falvaradorodriguez: The payload path is covered now, but the re-fetch path still looks racy to me: storedEventAt is read before the getSubscriptionsByCustomerId() call and outside the lock, and materialize() gates its ordering check on !authoritative. So a slow re-fetch can land after a newer deleted was already materialized and write the space back to active, leaving the mark at the newer stamp so nothing detects it. Could we pass the pre-fetch storedEventAt into materialize() and bail inside the lock if it moved? WDYT?

src/modules/entitlements/routes/entitlements.service.ts
```diff
@@ -0,0 +1,204 @@
...
+      // is what the write is actually ordered against.
+      await this.subscriptionsRepository.lockSpaceForSync(
+        args.spaceId,
+        entityManager,
+      );
+      const storedEventAt = await this.subscriptionsRepository.getLastEventAt(
+        args.spaceId,
+        entityManager,
+      );
+      if (
```
falvaradorodriguez: getLastEventAt() is a MAX over the whole space, but the payloads it gates are per-subscription, so on a plan change, where upstream emits events for two different upstream_subscription_ids, the older-stamped one gets dropped here even though it's the only event that carries the new subscription. E.g. created (new sub, T1) losing the race to canceled (old sub, T2) leaves the space with no active subscription at all. And since the skip just warns and returns, the webhook is acked and that state is lost until some later event arrives.

Should we either scope the comparison to the subscription's own last_event_at, or signal the skip to the caller so it re-fetches?

src/modules/entitlements/routes/subscription-sync.service.ts
```diff
@@ -0,0 +1,209 @@
...
+      features.map((feature) => [feature.key, feature.type]),
+    );
+
+    const onWarning = (message: string): void =>
+      this.loggingService.warn(message);
+    const eventAt = fromSecondsTimestamp(event.created);
+    const storedEventAt =
+      await this.subscriptionsRepository.getLastEventAt(spaceId);
+    const isStale =
+      eventAt === null || (storedEventAt !== null && eventAt < storedEventAt);
```
falvaradorodriguez: eventAt < storedEventAt is strict and Stripe's created has second granularity, an updated and a deleted emitted in the same second and delivered out of order count as "orders after" and get applied. <= would send ties down the re-fetch path instead. No test covers equal stamps.

src/modules/entitlements/domain/subscriptions.repository.ts
```diff
@@ -0,0 +1,106 @@
...
+      .where('space_id = :spaceId', { spaceId })
+      .getRawOne<{ lastEventAt: Date | null }>();
+    return newest?.lastEventAt ?? null;
+  }
+
+  public async lockSpaceForSync(
+    spaceId: Space['id'],
+    entityManager: EntityManager,
+  ): Promise<void> {
+    await entityManager.query('SELECT pg_advisory_xact_lock($1)', [spaceId]);
```
falvaradorodriguez: Nit: single-argument pg_advisory_xact_lock($1) shares the global key space with PostgresDatabaseMigrationHook.LOCK_MAGIC_NUMBER = 132, so the webhook for space id = 132 contends with the migration lock. The two-int form would isolate it.

src/modules/entitlements/domain/entities/space-subscription.entity.ts
```diff
@@ -0,0 +1,17 @@
...
+export type SpaceSubscription = z.infer<typeof SpaceSubscriptionSchema>;
+
+export const SpaceSubscriptionSchema = RowSchema.extend({
+  upstreamSubscriptionId: z.string(),
+  status: SubscriptionStatusSchema,
+  planId: z.string(),
+  planName: z.string().nullable(),
+  currentPeriodStart: z.date().nullable(),
+  currentPeriodEnd: z.date().nullable(),
+  entitlements: z.array(SubscriptionEntitlementSchema).optional(),
```
falvaradorodriguez, line 17: missing lastEventAt??

moisses89: Added https://github.com/safe-global/safe-client-gateway/pull/3327/changes#diff-1a4a21518ec8342bcfe88b376ed269f41650a81d4fe9e656450d2ea7c158dacfR16

src/modules/entitlements/routes/subscription-sync.service.ts
moisses89: the order guard was moved to materialize and the handler just fallbacks to pre-fetch when materialize guard detect an out of order event/

falvaradorodriguez (dismissed):
LGTM!

falvaradorodriguez (dismissed):

PooyaRaki:
@claude arch-review

moisses89:
> ## Architecture check
> Applicable guides: ARCHITECTURE, module-structure, api-dtos-and-validation, database-and-migrations, nestjs-patterns, security, caching-and-performance, best-practices, testing, commits-and-prs, reviewing + remarks.
> 
> src/modules/billing/routes/entities/subscription.entity.ts:28 — URI versioning: the shipped v1 `Subscription` response has `validUntil` removed/renamed in place instead of a new controller version alongside it. src/modules/entitlements/datasources/entities/space-feature-usage.entity.db.ts:23 — No speculative abstractions (YAGNI): `space_feature_usage` and `space_seat_selection` ship entity, schema and table with no reader or writer anywhere in this PR (deferred to PR 3). src/modules/entitlements/routes/entitlements.service.ts:24 — Canonical skeleton: the new module ships a `routes/` directory although it exposes no HTTP endpoints (guide: `routes/` only when the module exposes endpoints). src/modules/entitlements/entitlements.module.ts:38 — New-module checklist: `EntitlementsModule` has no entry in `src/app.module.ts`, deferred to PR 2 (cf. remarks R-028). src/modules/entitlements/routes/subscription-sync.service.ts:210 — R-023 (`info` level for a high-volume event path): every processed webhook logs at `info`; a per-event path logs at `debug`.
> 
> **Verdict:** ❌ 5 finding(s)

1. URI versioning: Accurate by the rule, but its premise does not hold: the billing service stopped sending validUntil before this client shipped, so no response ever carried it, the DTO documented a field upstream had already dropped. The endpoint is also behind `FF_BILLING`. Not versioning; recorded in the description.

2. YAGNI:  PLA-1643 scopes the entitlements data model itself, so these tables are the deliverable rather than speculation, omitting two of the five would be the deviation. The whole schema lands in one migration.

3. Canonical skeleton:  Accurate: the directory lands one PR before its controller. The alternative is worse by the same guide, route-shaped services under domain/ is the portfolio anti-example it names explicitly. [PR](https://github.com/safe-global/safe-client-gateway/pull/3346)  adds EntitlementsController here; ISubscriptionSyncService in domain/ keeps billing out of routes/ meanwhile.

4. New-module checklist:  The module is wired: BillingModule imports it, and AppModule gates that behind `FF_BILLING` intentional, since entitlements cannot be enabled without billing. The AppModule entry belongs with the HTTP surface it exposes, so it arrives with the controller in PR 2, not as an R-028 deferral.

5. R-023: Half accepted, and the flood is elsewhere: the billing service forwards Stripe's whole catalog, so the ignore paths at :91 and :101 fire for every invoice.*, charge.* and payment_intent.*. Both move to debug. :210 fires once per materialization, the per-request fact R-023 keeps at info.

src/modules/billing/routes/billing.controller.ts
```diff
@@ -55,9 +58,14 @@ export class BillingController {
...
-  postWebhook(): void {
-    // Origin is authenticated by BillingWebhookAuthGuard.
-    // TODO(PLA-1678 follow-up): validate and process the webhook payload.
+  public async postWebhook(
+    @Body(new ValidationPipe(WebhookEventSchema)) payload: WebhookEvent,
+  ): Promise<void> {
+    // Origin is authenticated by BillingWebhookAuthGuard. A malformed body
+    // 422s here like any other piped input; a well-formed-but-unprocessable
+    // event (unknown space, `api` customer group) is acked further down in
+    // SubscriptionSyncService.handleWebhook, since retrying that cannot help.
```
LucieFaire: nit: dont think we need this comment

src/modules/entitlements/domain/subscriptions.repository.interface.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export type SubscriptionValues = Pick<
+  SpaceSubscription,
+  | 'status'
+  | 'planId'
+  | 'planName'
+  | 'currentPeriodStart'
+  | 'currentPeriodEnd'
+  | 'lastEventAt'
+>;
```
LucieFaire: a bit unexpected place for it :)

src/modules/entitlements/routes/subscription-sync.service.ts
```diff
@@ -0,0 +1,266 @@
...
+    private readonly featuresRepository: IFeaturesRepository,
+    @Inject(ISubscriptionsRepository)
+    private readonly subscriptionsRepository: ISubscriptionsRepository,
+    @Inject(CacheService)
+    private readonly cacheService: ICacheService,
+    @Inject(LoggingService)
+    private readonly loggingService: ILoggingService,
+  ) {}
+
+  public async handleWebhook(event: WebhookEvent): Promise<void> {
```
LucieFaire, line 83: do you think its possible to split this method into smaller chunks or private ones? Because its quite big

LucieFaire (dismissed):
nothing blocking, can be merged as is, but good to haves

src/modules/billing/routes/billing.controller.ts
moisses89: Done https://github.com/safe-global/safe-client-gateway/pull/3327/changes/a761ad06b1b9f6bca16ffda954ba0c22c245d594

src/modules/entitlements/domain/subscriptions.repository.interface.ts
moisses89: True! repository is not the place Done https://github.com/safe-global/safe-client-gateway/pull/3327/changes/a761ad06b1b9f6bca16ffda954ba0c22c245d594

src/modules/entitlements/routes/subscription-sync.service.ts
moisses89: Divided in two, getting the Space and syncyng the subscription https://github.com/safe-global/safe-client-gateway/pull/3327/changes/b1d5cf28a3806012cebf4719712e9d31a60a75b7

LucieFaire (approved):

PR #3328, opened 2026-08-06
docs: Add AI guidelines for agents and repository standards

PooyaRaki:
## Summary

  Establish comprehensive AI guidelines and agent documentation infrastructure to provide clear, centralized guidance for development work in this repository. These guidelines define architectural patterns, security standards, testing conventions, and best practices for contributors and AI agents working on the codebase.
 
  ## Changes

  - **docs/agents/**: Add 10 specialized guide documents:
    - `ARCHITECTURE.md` — Request lifecycle, layers, validation boundaries, error funnels
    - `best-practices.md` — TypeScript conventions, patterns, and cross-cutting dos/don'ts
    - `api-dtos-and-validation.md` — Endpoint design, DTO schemas, validation pipelines
    - `security.md` — Auth guards, signature recovery, secrets management, PII handling
    - `module-structure.md` — Module layout, naming, versioning, import rules
    - `caching-and-performance.md` — Cache patterns, datasources, upstream calls, perf
    - `database-and-migrations.md` — TypeORM, migrations, entities, naming 
    - `testing.md` — Test conventions, builders, fixture patterns, Vitest setup
    - `configuration-and-flags.md` — Env vars, feature flags, RootConfigurationSchema
    - `reviewing.md` — PR review checklist, verification standards, approval criteria
    
  - **AGENTS.md** — Routing table with read-first mapping by task type (10 categories: anything non-trivial, TypeScript changes, endpoints, auth, modules, datasources, migrations, tests, config, PRs)

  - **CLAUDE.md** — Project-level documentation index pointing to AGENTS.md and specialized guides
 
  - **TESTING.md** — Condensed to integration-test specifics; full testing guide moved to docs/agents/testing.md

  - **CONTRIBUTING.md** — Point contributors at `docs/agents/` and the routing table; add the SPDX license header the file was missing

  - **.gitignore** — Track docs/agents/ content while excluding other local docs; ignore `.claude/cache/`, which was previously untracked and unignored

  - **Subsystem pointers** — Add `AGENTS.md` + `CLAUDE.md` stubs in the directories that carry their own constraints, so the governing guide loads when work happens in that subtree:

    | Directory | Points at |
    |---|---|
    | `migrations/` | `database-and-migrations.md` — the FK/WHERE-column index rule |
    | `src/modules/` | `module-structure.md` — canonical skeleton, naming, versioning |
    | `src/routes/`, `src/domain/` | `module-structure.md` (frozen trees — new feature code belongs in `src/modules/`) plus the layering and `Schema.parse()` rules |
    | `src/datasources/` | `caching-and-performance.md` — cross-cutting infrastructure only, `CacheRouter`/TTL rules, error funnels |
    | `src/config/` | `configuration-and-flags.md` — the `configuration.ts` + `RootConfigurationSchema` declaration rule |
    | `src/modules/auth/` | `security.md` — SIWE/JWT identity, guard placement, `HttpExceptionNoLog` |


PooyaRaki:
@claude review

gfendres (dismissed):

PooyaRaki:
@claude review

------
Thanks — went through each point and verified it against the tree. Two fixed, three don't hold up, the rest I'm leaving.
 
  Fixed

  .env.test is gitignored. Correct, and the guide was asserting something false: .env.test is in .gitignore, every test:* script loads it with --env-file-if-exists (silent no-op when absent), so a fresh clone has no such file. Rewrote docs/agents/testing.md to point at the real source of
  the defaults — src/config/entities/__tests__/configuration.ts already falls back to POSTGRES_TEST_PORT || '5433' / POSTGRES_TEST_DB || 'test-db', matching the Compose services — and to describe .env.test as an optional override you create yourself. No sample file needed; integration
  tests already run without it.

  safe-shield is off-shape and not called out. Correct. src/modules/safe-shield/ has no domain/ or routes/ at all — a controller at the module root plus one sub-module per analysis kind. Added it to the skeleton anti-example next to portfolio. (Small correction: module-structure.md:37
  flags only portfolio; csv-export/v1 is at line 111 under the versioning rule, a different rule.)
 
  Not doing

  !/docs/agents/** for explicitness. The remedy is a no-op — re-including the directory already covers descendants. Verified: with the current pattern, docs/agents/nested/x.md is not ignored. The observation that a future docs/adr/ would be ignored is right, but that's the intent, per
  the comment on the line above.

  PR-number citations need a one-line "what it broke". They already have one — every citation sits inside a **Why:** line that says exactly that. security.md:34 reads: "PR #290 — cache-hook endpoints shipped with no auth guard at all, so anything that could reach their path could trigger
  them."

  OptionalAuthGuard justification missing from reviewing.md. By design. reviewing.md has no guard-specific items at all: Part 1 is "every rule of each applicable guide is verified against the diff." Copying individual guide rules into the checklist is the duplication that structure
  exists to avoid.
 
  Leaving as-is

  - AGENTS.md #10 should also link security.md — premise is right (security.md:166-178 does cover HttpExceptionNoLog; the cited security.md:8-11 doesn't), but the routing table already sends auth and error-handling work to security.md.
  - No automated check on the "zero instances" baselines — fair, and the guide is upfront that it's a baseline to hold. Enforcement is its own PR, not a docs one.
  - Prose weight / TOCs — noted as non-blocking; keeping the current shape for now.
  - 8×2 AGENTS.md/CLAUDE.md stubs — deliberate, tool-agnostic. Each CLAUDE.md is a five-line pointer with nothing to drift, and symlinks break on Windows checkouts.
  - Non-negotiables #1 vs #10 overlap — they don't overlap: #1 is input validation, #10 is which exception type each layer throws

PooyaRaki:
@claude review

PooyaRaki:
@claude review

PooyaRaki:
Addressed in 9c062831c:
 
  **CodeQL: untrusted checkout + TOCTOU** — fixed structurally. The workflow no longer checks out the PR head at all: the checkout is now the default branch (only there to supply AGENTS.md and the guides), and the agent reads the PR exclusively through `gh pr view` / `gh pr diff`, which
  were already in the allowlist. Untrusted code never reaches the runner's disk and a mid-review force-push can't swap the analyzed content, so both alerts are cleared at the root rather than by pinning a SHA. `fetch-depth: 0` is gone too, since no local diff is computed anymore.

  **`gh pr comment` scope** — all three `gh pr` commands in the allowlist are now pinned to the triggering PR's number via `${{ github.event.issue.number || github.event.pull_request.number }}`.

  **Grep fragility in testing.md / database-and-migrations.md** — added `-F` to the constructor-sweep command in both guides so class names match literally.

  **agent-tooling.md** — updated the CI-gate section to describe the no-PR-checkout design, per the stale-guides check.
  
  **Concurrency key** — verified against the webhook payload shapes, no change: `issue_comment` events carry `issue.number` and `pull_request_review_comment` events carry `pull_request.number`, so the `||` fallback always resolves to a real number and the group name can't collapse to
  `architecture-check-`.

  Two items deliberately not taken:

  - **CI guard for the "4 double-casts" baseline** — ratchet enforcement is earmarked for a separate enforcement PR (see `docs/superpowers/plans/2026-07-23-ai-guidelines-content.md`); a one-off grep-count check in `ci.yml` here would pre-empt that PR's design.
  - **Runtime JSON validation in the PreToolUse hook** — embedding `python3 -c 'json.load(...)'` in the hook command contradicts the hook's own portability constraint in agent-tooling.md (only `git`, `test`, `printf` — the same reason `jq` is banned). The payload is a static literal, and
  the guide already mandates re-testing both hook paths after any edit, which covers the edit-time risk.

PooyaRaki:
@claude review

PooyaRaki:
@claude review

PooyaRaki:
Both findings addressed in 48ac1aa53:

1. Stale eight-item references — cgw-review/SKILL.md now enumerates the seven current Part 2 items (with ratchet-baseline growth noted as a sub-item of "No stale guides") and says "Seven independent checks"; architecture-check.yml now says "all seven checks" and its standalone ratchet adaptation is folded under a "No stale guides" adaptation, matching where the item lives in reviewing.md.
2. JSON validator portability — agent-tooling.md now recommends node -e 'JSON.parse(require("fs").readFileSync(0, "utf8"))' instead of python3, since Node is the one runtime this repo already guarantees via package.json.

PooyaRaki:
@claude review

LucieFaire (dismissed):

PooyaRaki:
Confirmed and fixed in f9d9250ae.

  Verified each point against the tree: email-encryption.constants.ts / email-encryption.service.ts are gone, the mechanics live in src/datasources/kms/kms-encryption.service.ts with the frozen on-disk contract in src/datasources/kms/encryption.constants.ts, and the per-module wrappers
  are the new consumption shape — so the guide's "write a new per-field service, don't parameterize" rule had inverted against the code.

  The rewritten section now:
  - points the canonical example at encryption.constants.ts + kms-encryption.service.ts for the frozen contract/mechanics, and at user-encryption.service.ts (with its { userId: String(userId) } context binding) for how a module consumes it;
  - replaces "implemented once today" with the actual coverage: users.email, wallet addresses, member names/aliases, and the spaces module's names, Safe addresses, address-book entries, and audit payloads — one addition to your list: SpaceEncryptionService.encryptSafeAddress also covers
  space Safe addresses;
  - rewrites the new-PII-column rule to the current shape: extend the owning module's wrapper (or add one) to bind that entity's encryptionContext and delegate to KmsEncryptionService; never reimplement the envelope/blind-index mechanics per field, never alter the frozen constants.
  
  Remaining EmailEncryptionService mentions exist only in docs/superpowers/ specs/plans, which are point-in-time historical documents, not living guides, so they were left untouched.

PooyaRaki:
@claude review

LucieFaire (approved):

PR #3329, opened 2026-08-07
feat: recognise Safe 1.5.0 as a known, trusted version

tongyaosafe:
## Summary

Closes WA-2921

Teaches the CGW paths that reason about Safe versions to recognise 1.5.0,
ahead of the per-chain rollout (WA-2937). Without this, every 1.5.0 Safe
bridging to its own address gets a spurious CRITICAL `INCOMPATIBLE_SAFE`
from Safe Shield, and the new privileged `setModuleGuard` method renders as
a raw contract interaction.

All changes rely only on deployment data already present in the pinned
1.37.58; the chain-coverage bump ships separately in
`tongyao/wa-2936-bump-safe-deployments` and neither PR depends on the
other. Two items from the original ticket needed no code (see the
scope-audit comment on WA-2921): the safe mapper's version state is driven
by Config Service + Transaction Service data, and the
official-fallback-handler check already derives its version list
dynamically from `safe-deployments`.

API impact is additive only: `setModuleGuard` transactions move from
`txInfo.type: "Custom"` to `"SettingsChange"` with two new
`settingsInfo.type` values; existing clients degrade gracefully.

## Changes

- `src/modules/safe-shield/recipient-analysis/` — add `1.5.0` to
  `SAFE_VERSIONS` and `MULTICHAIN_SUPPORTED_VERSIONS`; tests cover both a
  compatible target chain (Sepolia) and a randomly selected 1.4.1-only
  target chain
- `src/modules/transactions/routes/` — decode `setModuleGuard` into
  `SET_MODULE_GUARD` / `DELETE_MODULE_GUARD` settings-change entities via a
  shared guard-like handler (zero address deletes, mirroring `setGuard`);
  OpenAPI registration included


tongyaosafe:
@Claude review

tongyaosafe:
> **Claude finished @tongyaosafe's task in 2m 20s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/31163432035)
> 
> ### Claude PR Review
> Overall this is a clean, focused change that mirrors existing patterns well. Only nits to flag — nothing blocking.
> 
> #### Findings
> **1. `handleSetModuleGuard` is a near-verbatim copy of `handleSetGuard`** — `src/modules/transactions/routes/mappers/common/settings-change.mapper.ts:187-227`. Same shape, same param index, same NULL_ADDRESS branch — only the class names differ. Not worth blocking on (matches how the other guard handler was written), but if you're touching this again, extracting a shared `handleGuardLike(chainId, dataDecoded, SetCls, DeleteCls)` would remove ~20 lines and one future drift risk. [Fix this →](https://claude.ai/code?q=In%20src%2Fmodules%2Ftransactions%2Froutes%2Fmappers%2Fcommon%2Fsettings-change.mapper.ts%20on%20branch%20tongyao%2Fwa-2921-cgw-recognise-150-as-a-known-trusted-version%2C%20refactor%20handleSetGuard%20and%20handleSetModuleGuard%20into%20a%20shared%20private%20helper%20that%20takes%20the%20Set-%20and%20Delete-%20entity%20constructors%2C%20since%20the%20two%20methods%20are%20otherwise%20identical.&repo=safe-global/safe-client-gateway)
> 
> **2. Missing negative-path test for the `MULTICHAIN_SUPPORTED_VERSIONS` change** — `src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts:1695`. The added test confirms a 1.5.0 Safe → chain-with-1.5.0-deployments yields `MISSING_OWNERSHIP`. There's no complementary test asserting that a 1.5.0 Safe → chain **without** 1.5.0 deployments still returns `INCOMPATIBLE_SAFE`. Since the whole reason for adding `'1.5.0'` to that list is that `checkVersionCompatibility` gates it per-chain, a regression there (e.g., a future refactor short-circuiting to `true`) would silently mark 1.5.0 as compatible everywhere. One extra test on a chain with only 1.4.1 deployments would lock this in.
> 
> **3. `.yarnrc.yml` — `npmPreapprovedPackages: ['@safe-global/*']` bypasses the 7-day age gate** — worth being explicit: any package published under `@safe-global/*` now skips the cooling-off window that hardened-mode adds. If an `@safe-global` npm publish token were ever compromised, a malicious version would land in CI/prod immediately instead of after a week. The trade-off is fine for in-house packages (and matches safe-wallet-monorepo), but the scope is broad — narrowing to specific packages you actually consume, or at least documenting the threat model, would be safer than a wildcard.
> 
> **4. Nit — `NULL_ADDRESS` string comparison** — `handleSetModuleGuard` compares a decoded string to `NULL_ADDRESS` (all-lowercase) with strict `!==`. The zero address has no checksum bits, so this is safe in practice, but it inherits the same fragility as `handleSetGuard`: if the decoder ever normalized to checksummed form on non-zero addresses, this exact comparison would still be fine, but a lowercase-vs-uppercase mismatch elsewhere would break. Non-actionable for this PR — flagging only because the pattern is now duplicated.
> 
> #### Not concerns
> * Entity shape mirrors `SetGuard`/`DeleteGuard` exactly; OpenAPI registration + `SETTINGS_CHANGE_METHODS` list are all in sync.
> * `MULTICHAIN_SUPPORTED_VERSIONS` ordering (1.5.0 first) is fine — it's a `.some()` short-circuit, not a priority list.
> * Deployment-fixture updates in `deployments.helper.spec.ts` are pure data changes tracking the `safe-deployments` bump.
> * [x]  Read diff vs origin/main
> * [x]  Review recipient-analysis changes
> * [x]  Review settings-change mapper + entities
> * [x]  Review tests
> * [x]  Post review
>   · Branch: `tongyao/wa-2921-cgw-recognise-150-as-a-known-trusted-version`

1, 2, 4 are fixed.
3 I did not fix. keep the same pattern as monorepo

PooyaRaki:
@claude arch-review

tongyaosafe:
> ## Architecture check
> Guides applied: `commits-and-prs.md`, `api-dtos-and-validation.md`, `best-practices.md`, `module-structure.md`, `testing.md`, `remarks.md`.
> 
> * `package.json:65` — Splitting a change: one PR carries a dependency bump, a supply-chain-gate policy change, and feature work in two unrelated modules (`safe-shield` recipient analysis, `transactions` settings mapper), which the guide lists as too large on two counts ("mixes a dependency bump with feature work", "changes more than one module for reasons that are not a single mechanical consequence").
> * `.yarnrc.yml:18` — No scope creep: `npmPreapprovedPackages: ['@safe-global/*']` is a standing, repo-wide exemption from the 7-day `npmMinimalAgeGate` for every in-house package, broader than the single `safe-deployments` bump requires and not a consequence of recognising Safe 1.5.0.
> * PR title — Linking: the title carries `(WA-2921)`, and squash-merge makes it a permanent commit subject with an issue reference the guide reserves for the body (0 of the last 400 subjects on `main` do this).
> * `src/modules/transactions/routes/mappers/common/settings-change.mapper.spec.ts:308` — R-008 (Tests that assert the framework, not the behavior): asserts `SETTINGS_CHANGE_METHODS` contains the literal the same diff just added to it, already implied by the two mapper tests above it.
> 
> **Verdict:** ❌ 4 finding(s)

fixed

PooyaRaki:
@claude arch-review

src/modules/transactions/routes/mappers/common/settings-change.mapper.spec.ts
```diff
@@ -266,6 +268,43 @@ describe('Multisig Settings Change Transaction mapper (Unit)', () => {
...
+      faker.string.numeric(),
+      dataDecoded,
+    );
+
+    const expected = new SetModuleGuard(new AddressInfo(moduleGuardAddress));
+    expect(actual).toEqual(expected);
+  });
+
+  it('should build a DeleteModuleGuard setting', async () => {
+    const moduleGuardValue = '0x0000000000000000000000000000000000000000';
```
PooyaRaki: Nit: We should use a random value generated by `faker`

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
```diff
@@ -1692,6 +1692,118 @@ describe('RecipientAnalysisService', () => {
...
+        txInfo: mockTxInfo,
+      });
+
+      expect(result[mockSafeAddress]?.BRIDGE?.[0]?.type).toBe(
+        'MISSING_OWNERSHIP',
+      );
+    });
+
+    it('should return INCOMPATIBLE_SAFE for a 1.5.0 Safe when the target chain only has 1.4.1 deployments', async () => {
+      const targetChainId = '1101'; // Polygon zkEVM has 1.4.1 deployments but no 1.5.0 ones
```
PooyaRaki: Since we’re mocking the responses, we should be able to make the `chainId` random.

PooyaRaki (dismissed):

src/modules/transactions/routes/mappers/common/settings-change.mapper.spec.ts
tongyaosafe: fixed. Thanks
c7aae6b13d68491d274d6508100ae53a2e3bf942

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
tongyaosafe: fixed. Thanks
c7aae6b13d68491d274d6508100ae53a2e3bf942

PooyaRaki (approved):

PR #3334, opened 2026-08-10
feat: recognise the 1.5.0 ExtensibleFallbackHandler as official

tongyaosafe:
## Summary

WA-2921 / WA-2862

Safe 1.5.0 ships **two** official fallback handlers, but our trust checks only consulted the `CompatibilityFallbackHandler` deployment tables — so the official `ExtensibleFallbackHandler` (canonical `0x85a8ca358D388530ad0fB95D0cb89Dd44Fc242c3` on 120 chains, v1.5.0) was treated as untrusted everywhere:

- **Safe Shield contract analysis**: a `setFallbackHandler(0x85a8…42c3)` transaction produced an `UNOFFICIAL_FALLBACK_HANDLER` WARN on every chain. Reproducible on staging today against any Sepolia Safe — the same result that flags the handler as unofficial also names it via the Decoder Service:

  ```json
  {
    "severity": "WARN",
    "type": "UNOFFICIAL_FALLBACK_HANDLER",
    "title": "Unofficial fallback handler",
    "fallbackHandler": {
      "address": "0x85a8ca358D388530ad0fB95D0cb89Dd44Fc242c3",
      "name": "Safe: ExtensibleFallbackHandler 1.5.0"
    }
  }
  ```

- **Safe Shield recipient analysis**: a Safe *created* with the extensible handler failed the `fallbackHandlerExists` leg of `checkVersionCompatibility`, so bridge analysis reported `INCOMPATIBLE_SAFE` on every target chain.
- Same code path: the `SafeToL2Setup` existence check was pinned to `'1.4.1'`, so chains that only register the 1.5.0 setup contract would also be reported incompatible.

The fix derives everything from `@safe-global/safe-deployments` (no hardcoded addresses): new `ExtensibleFallbackHandler` wrappers in the deployments util, OR'd into both checks, and the `SafeToL2Setup` versions derived from the package instead of pinned.

Product context: creation/upgrade keeps setting the `CompatibilityFallbackHandler` (WA-2928 decides any change to that) — this PR is recognition-only, so users whose handler is the extensible one (CoW composable orders, module/policy setups) stop getting false warnings.

## Changes

- `src/domain/common/utils/deployments.ts` — add `getExtensibleFallbackHandlerDeployments` / `getExtensibleFallbackHandlerVersions` / `isExtensibleFallbackHandlerDeployed` (mirroring the existing compat wrappers) and `getSafeToL2SetupVersions`
- `src/modules/safe-shield/contract-analysis/contract-analysis.service.ts` — `isOfficialFallbackHandler` now also matches extensible deployments
- `src/modules/safe-shield/recipient-analysis/recipient-analysis.service.ts` — `fallbackHandlerExists` also matches extensible deployments; `SafeToL2Setup` check iterates the package's versions (1.4.1 + 1.5.0) instead of pinning `'1.4.1'`
- `src/domain/common/utils/__tests__/deployments.spec.ts` — new spec for the added wrappers, incl. the negative case (compat tables do **not** match the extensible address)
- `contract-analysis.service.spec.ts` / `recipient-analysis.service.spec.ts` — cases for `setFallbackHandler(extensible)` producing no warning, and an extensible-created 1.5.0 Safe no longer reported `INCOMPATIBLE_SAFE`

### DIFF
--- a/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
+++ b/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
@@ -1279,6 +1279,57 @@ describe('ContractAnalysisService', () => {
         expect(mockDataDecoderApi.getContracts).toHaveBeenCalledTimes(1);
       });
 
+      it('should omit FALLBACK_HANDLER when handler is the official ExtensibleFallbackHandler', async () => {
+        const extensibleHandler = getAddress(
+          '0x85a8ca358D388530ad0fB95D0cb89Dd44Fc242c3',
+        );
+        const mainnetChainId = '1';
+
+        const mockContractPage = pageBuilder()
+          .with('count', 1)
+          .with('results', [
+            contractBuilder()
+              .with('address', contractAddress)
+              .with('abi', {
+                abiJson: [{ type: 'function', name: 'test' }],
+                abiHash: faker.string.hexadecimal() as Hex,
+                modified: faker.date.recent(),
+              })
+              .with('displayName', name)
+              .with('logoUrl', logoUrl)
+              .build(),
+          ])
+          .build();
+
+        mockDataDecoderApi.getContracts.mockResolvedValue(
+          rawify(mockContractPage),
+        );
+
+        const result = await service.verifyContract({
+          chainId: mainnetChainId,
+          contract: {
+            address: contractAddress,
+            isDelegateCall: false,
+            fallbackHandler: extensibleHandler,
+          },
+        });
+
+        expect(result).toEqual({
+          logoUrl,
+          name,
+          CONTRACT_VERIFICATION: [
+            {
+              severity: SEVERITY_MAPPING.VERIFIED,
+              type: 'VERIFIED',
+              title: TITLE_MAPPING.VERIFIED,
+              description: DESCRIPTION_MAPPING.VERIFIED({ name }),
+            },
+          ],
+        });
+
+        expect(mockDataDecoderApi.getContracts).toHaveBeenCalledTimes(1);
+      });
+
       it('should omit FALLBACK_HANDLER when handler address is undefined', async () => {
         const mockContractPage = pageBuilder()
           .with('count', 1)
--- a/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
+++ b/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
@@ -8,6 +8,7 @@ import type { IConfigurationService } from '@/config/configuration.service.inter
 import { FakeCacheService } from '@/datasources/cache/__tests__/fake.cache.service';
 import { CacheRouter } from '@/datasources/cache/cache.router';
 import { LogType } from '@/domain/common/entities/log-type.entity';
+import { getExtensibleFallbackHandlerDeployments } from '@/domain/common/utils/deployments';
 import { pageBuilder } from '@/domain/entities/__tests__/page.builder';
 import type { IDataDecoderApi } from '@/domain/interfaces/data-decoder-api.interface';
 import type { ITransactionApi } from '@/domain/interfaces/transaction-api.interface';
@@ -1280,10 +1281,13 @@ describe('ContractAnalysisService', () => {
       });
 
       it('should omit FALLBACK_HANDLER when handler is the official ExtensibleFallbackHandler', async () => {
-        const extensibleHandler = getAddress(
-          '0x85a8ca358D388530ad0fB95D0cb89Dd44Fc242c3',
-        );
         const mainnetChainId = '1';
+        const extensibleHandler = faker.helpers.arrayElement(
+          getExtensibleFallbackHandlerDeployments({
+            chainId: mainnetChainId,
+            version: '1.5.0',
+          }),
+        );
 
         const mockContractPage = pageBuilder()
           .with('count', 1)
--- a/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
+++ b/src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
@@ -1280,7 +1280,7 @@ describe('ContractAnalysisService', () => {
         expect(mockDataDecoderApi.getContracts).toHaveBeenCalledTimes(1);
       });
 
-      it('should omit FALLBACK_HANDLER when handler is the official ExtensibleFallbackHandler', async () => {
+      it('should return VERIFIED when the handler is the official ExtensibleFallbackHandler', async () => {
         const mainnetChainId = '1';
         const extensibleHandler = faker.helpers.arrayElement(
           getExtensibleFallbackHandlerDeployments({
--- a/src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
+++ b/src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
@@ -1749,6 +1749,62 @@ describe('RecipientAnalysisService', () => {
       );
     });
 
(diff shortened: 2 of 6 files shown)


tongyaosafe:
@claude arch-review

tongyaosafe:
@claude arch-review

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
```diff
@@ -1749,6 +1749,62 @@ describe('RecipientAnalysisService', () => {
       );
     });
 
+    it('should not return INCOMPATIBLE_SAFE for a 1.5.0 Safe created with the ExtensibleFallbackHandler', async () => {
+      const targetChainId = '11155111'; // Sepolia has canonical 1.5.0 deployments
```
PooyaRaki: Nit: We should read this from deployments.

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
```diff
@@ -1749,6 +1749,62 @@ describe('RecipientAnalysisService', () => {
...
+        return Promise.resolve({
+          ...mockTransactionApi,
+          getSafe: vi.fn().mockResolvedValue(null),
+        });
+      });
+
+      const v150CreationTransaction = {
+        ...createMockCreationTransaction([mockSafeAddress], 1),
+        masterCopy: '0xFf51A5898e281Db6DfC7855790607438dF2ca44b' as Address, // Safe 1.5.0
+        factoryAddress: '0x14F2982D601c9458F93bd70B218933A6f8165e7b' as Address, // SafeProxyFactory 1.5.0
```
PooyaRaki: Nit: Can we read them from the deployments library?

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
```diff
@@ -1749,6 +1749,62 @@ describe('RecipientAnalysisService', () => {
...
+      );
+
+      const result = await service.analyzeBridge({
+        chainId: mockChainId,
+        safeAddress: mockSafeAddress,
+        txInfo: mockTxInfo,
+      });
+
+      expect(result[mockSafeAddress]?.BRIDGE?.[0]?.type).toBe(
+        'MISSING_OWNERSHIP',
```
PooyaRaki: The test title expects it to not return `INCOMPATIBLE_SAFE` but we're checking `MISSING_OWNERSHIP` here. This is a bit confusing.

src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
```diff
@@ -1279,6 +1279,57 @@ describe('ContractAnalysisService', () => {
         expect(mockDataDecoderApi.getContracts).toHaveBeenCalledTimes(1);
       });
 
+      it('should omit FALLBACK_HANDLER when handler is the official ExtensibleFallbackHandler', async () => {
```
PooyaRaki: The title should be something like:
```
Should return `VERIFIED` when the handler is the official ExtensibleFallbackHandler.
```

PooyaRaki (dismissed):

PooyaRaki:
@claude review

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
tongyaosafe: 19a166391098ea0fc8f73c922b1976dc0403cf71
fixed. Thanks.

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
tongyaosafe: 19a166391098ea0fc8f73c922b1976dc0403cf71
fixed. Thanks.

src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts
tongyaosafe: fixed. And a comment to explain
19a166391098ea0fc8f73c922b1976dc0403cf71

src/modules/safe-shield/contract-analysis/contract-analysis.service.spec.ts
tongyaosafe: fixed
c028a3194104fdd7dfdbfab6436bf07cd9ff5cea
Thanks

tongyaosafe:
> **Claude finished @PooyaRaki's task in 1m 36s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/31589303212)
> 
> ### Code Review
> * [x]  Read changed files
> * [x]  Assess code quality, bugs, performance, security, test coverage
> * [x]  Post review feedback
> 
> The reviewer has already flagged the main test-quality items. A few more worth calling out:
> 
> **Test doesn't actually verify what its title claims** `src/modules/safe-shield/recipient-analysis/recipient-analysis.service.spec.ts:1747` and `:1803` — Both tests are titled around _absence_ of `INCOMPATIBLE_SAFE` but positively assert `MISSING_OWNERSHIP` on `BRIDGE[0]`. If a regression made both errors coexist (in either order), the assertion still passes while the intent silently degrades. A negative assertion over the full `BRIDGE` array is more faithful:
> 
> ```ts
> expect(
>   result[mockSafeAddress]?.BRIDGE?.map((b) => b.type) ?? [],
> ).not.toContain('INCOMPATIBLE_SAFE');
> ```
> 
> Alternatively — since the setup is contrived to hit the ownership branch on purpose — split into two assertions: the negative one for the titled invariant, and a separate positive one documenting _why_ the array is non-empty.
> 
> **Literal deployment addresses in service specs (R-002)** The extensible + compat + singleton + factory addresses are hardcoded across `recipient-analysis.service.spec.ts:1771,1772,1780,1848` and `contract-analysis.service.spec.ts:1281`. As the reviewer noted, these should be pulled from `@safe-global/safe-deployments` (via the existing helpers in `src/domain/common/utils/deployments.ts`, e.g. `getExtensibleFallbackHandlerDeployments({ chainId, version: '1.5.0' })[0]`). Then when the package bumps a canonical address, the tests don't silently pin to a stale value.
> 
> **Test title reflects internals, not behaviour** `contract-analysis.service.spec.ts:1282` — the title reads "should omit `FALLBACK_HANDLER`" but the assertion is on the presence of `CONTRACT_VERIFICATION: VERIFIED`. Aligning the title to the observable outcome (as the reviewer suggested) makes the intent obvious without reading the body.
> 
> **Micro-duplication in `fallbackHandlerExists`** `src/modules/safe-shield/recipient-analysis/recipient-analysis.service.ts:678-687` — the two `is*Deployed` calls take an identical `{ version, chainId, address }` object. Extracting `const args = { version, chainId: chain.chainId, address: safeCreationData.safeAccountConfig.fallbackHandler }` and reusing it (or mirroring the `some(version => ...)` shape already used in `contract-analysis.service.ts`) removes the repeated field list. Purely cosmetic.
> 
> **Nice touches**
> 
> * Negative test in `deployments.spec.ts` asserting the compat table does NOT match the extensible address — good regression guard against the wrong wrapper being wired later.
> * Deriving `SafeToL2Setup` versions from the package (`getSafeToL2SetupVersions()`) rather than pinning `'1.4.1'` is exactly the right shape and generalises cleanly to future versions.
> * Private `getExtensibleFallbackHandlerDeployments` — good response to the earlier `arch-review` finding.
>   · branch `tongyao/wa-2921-whitelist-extensible-fallback-handler`

fixed.

PooyaRaki (approved):
LGTM 🚀

PR #3345, opened 2026-08-13
fix(balances): serve only currencies the positions provider supports

vseehausen:
## Summary

WA-3157

`/v1/balances/supported-fiat-codes` returned CoinGecko's `simple/supported_vs_currencies` verbatim — 63 codes today, of which only 16 work end to end. Clients populate their currency selector from this endpoint, so the extra 47 are all selectable and each breaks something:

- `LINK`, `BITS` and `SATS` are not well-formed ISO 4217 codes, so `Intl.NumberFormat({ style: 'currency' })` throws `RangeError: Invalid currency code`. On web that throws while rendering every fiat value, and the selection is persisted — so a reload reproduces it.
- The remaining ~44 (`BRL`, `MXN`, `SEK`, `XAU`, …) are accepted by the balances path, since prices come from CoinGecko, but `ZerionPositionsApi` rejects any code outside `balances.providers.zerion.currencies` with a 400. Positions silently fail while token balances still render.

Zerion is the narrowest constraint in the stack, and all 16 of its codes are in CoinGecko's list, so the intersection is exactly the Zerion list. The endpoint now returns that config value directly rather than intersecting the two: the constraint is Zerion's, and `CoingeckoApi.getFiatCodes()` swallows errors and returns `[]`, so intersecting would empty the currency selector during a CoinGecko outage. Dropping the dependency removes that failure mode — the new integration test asserts the endpoint still answers when the prices provider is entirely unreachable.

Narrowing the advertised list is a deliberate regression for anyone who has `BRL`, `MXN` or another CoinGecko-only fiat selected today: balances stop being denominated in it, and their selector shows no match until they pick again. That is accepted — those users' positions are already broken, and serving positions in USD alongside balances in BRL would be a mixed-currency UI.

This leaves the CoinGecko fiat-codes path with no callers. It is not deleted here to keep this fix to one concern; WA-3158 tracks removing it.

## Changes

- `src/modules/balances/datasources/balances-api.manager.ts` — `getFiatCodes()` returns `balances.providers.zerion.currencies`, read via `getOrThrow` in the constructor, instead of delegating to `CoingeckoApi`
- `src/config/entities/configuration.ts` — note that the list is also the source of truth for the endpoint, so it is not widened past what Zerion accepts
- `src/modules/balances/datasources/balances-api.manager.spec.ts` — assert against the configured codes and that CoinGecko is not queried; the `vpcUrl` case gains the new config key, as its mock throws on unexpected keys
- `src/modules/balances/routes/balances.controller.integration.spec.ts` — replace the three CoinGecko-passthrough cases (lowercase, uppercase, empty-on-failure) with the configured-codes response and the prices-provider-outage case

### DIFF
--- a/src/config/entities/configuration.ts
+++ b/src/config/entities/configuration.ts
@@ -126,6 +126,7 @@ export default () => ({
         assetsApiKey:
           process.env.ZERION_ASSETS_API_KEY ?? process.env.ZERION_API_KEY,
         baseUri: process.env.ZERION_BASE_URI || 'https://api.zerion.io',
+        // Also the source of truth for /v1/balances/supported-fiat-codes.
         currencies: [
           'USD',
           'EUR',
--- a/src/modules/balances/datasources/balances-api.manager.spec.ts
+++ b/src/modules/balances/datasources/balances-api.manager.spec.ts
@@ -53,10 +53,16 @@ const networkService = {
 
 const networkServiceMock = vi.mocked(networkService);
 
+const fiatCodes = Array.from(
+  { length: faker.number.int({ min: 2, max: 5 }) },
+  () => faker.finance.currencyCode(),
+);
+
 beforeEach(() => {
   vi.resetAllMocks();
   configurationServiceMock.getOrThrow.mockImplementation((key) => {
     if (key === 'safeTransaction.useVpcUrl') return false;
+    if (key === 'balances.providers.zerion.currencies') return fiatCodes;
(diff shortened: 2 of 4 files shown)


vseehausen:
@claude review

tmjssz (dismissed):

vseehausen:
1. That's fine.
2. addressed.
3. addressed.
4. That's fine.

tmjssz (approved):

PR #3357, opened 2026-08-18
ci: sync releases to the Linear release pipeline

gfendres:
## Summary

Extends [WA-2846](https://linear.app/safe-global/issue/WA-2846/integrate-github-with-linear-releases-web-mobile) to the Client Gateway. That one wired the web and mobile pipelines in the wallet monorepo ([safe-wallet-monorepo#8291](https://github.com/safe-global/safe-wallet-monorepo/pull/8291)) and deliberately left CGW out of scope; this is that follow-up.

Publishing a GitHub release now records it in the [CGW pipeline](https://linear.app/safe-global/pipeline/cgw-7ua/releases) in Linear, so you can tell which issues are actually live instead of just merged. Same action as the wallet repo, pinned to the same commit, adapted to how we release. Two differences, both because our flow is simpler: we have no release branch or code freeze, so everything hangs off the published GitHub release rather than a release-start workflow; and the CGW pipeline is **continuous**, so a release auto-completes the moment it syncs. No separate `complete` call, and its single `Released` stage means there's nothing to move between.

It lives in its own workflow rather than in `ci.yml` so Linear never rides on the docker build. The sync fails loudly instead of failing open: if Linear is unreachable or rejects the key, the run goes red and you re-run it from the Actions UI. That's safe here because this workflow publishes nothing itself, so a red run can't block or revert a release that already shipped.

**Before this works**, someone needs to add `LINEAR_ACCESS_KEY` as a repo secret — the CGW pipeline's access key from its Linear settings, not a personal API key.

The risk worth flagging: our commit subjects don't carry Linear issue IDs. They look like `feat: recognise Safe 1.5.0 as a known, trusted version (#3329)`, while the wallet's look like `feat(web): support 1.5.0 in multichain creation (WA-2930)`. The action falls back to PR references, which may resolve through the Linear/GitHub integration, but I couldn't verify that. With no manual trigger there's no way to dry-run first, so the next published tag is also the first real test — if the Linear release comes back with no issues attached, that's the missing issue keys, not a broken workflow, and the fix is putting them in PR titles. Nothing has run yet; YAML and the inline shell are syntax-checked.

## Changes

**`.github/workflows/linear-release.yml`** (new)

- Triggers only on `release: [released]`. Resolves the version from the tag (`v1.122.0` → `1.122.0`) and runs `sync` with the GitHub release attached as a link.
- Checks out the release tag rather than a branch, so the commit scan covers that release's history.
- Full history (`fetch-depth: 0`) so the action can scan commits. No `include_paths` — we're not a monorepo, so any path filter would silently drop commits.
- Reads the tag and release URL straight off the event, so it makes no API calls of its own.
- `timeout-minutes: 10`, matching how `claude-code-review.yml` and `architecture-check.yml` bound their jobs.

### DIFF
--- a/.github/workflows/linear-release.yml
+++ b/.github/workflows/linear-release.yml
@@ -33,6 +33,7 @@ concurrency:
 jobs:
   linear-release:
     runs-on: ubuntu-latest
+    timeout-minutes: 10 # A hung sync should fail today, not in six hours
     permissions:
       contents: read
 
@@ -50,11 +51,6 @@ jobs:
         env:
           RELEASE_TAG: ${{ github.event.release.tag_name }}
         run: |
-          if [ -z "$RELEASE_TAG" ]; then
-            echo "::error::Release tag is empty; refusing to sync an untagged release."
-            exit 1
-          fi
-
           # Linear versions its releases without the tag's `v` prefix.
           echo "version=${RELEASE_TAG#v}" >> "$GITHUB_OUTPUT"
           echo "Linear release ${RELEASE_TAG#v} (tag $RELEASE_TAG)"
--- a/.github/workflows/linear-release.yml
+++ b/.github/workflows/linear-release.yml
@@ -4,14 +4,16 @@ name: Linear Release
 
 # Syncs Client Gateway releases to the continuous "CGW" pipeline in Linear.
 #
-# Automated path: publishing a GitHub release (tag `vX.Y.Z`) runs `sync`, which
-# scans the commits since the last release for Linear issue IDs, attaches them,
-# links the GitHub release, and completes it in one step - continuous pipelines
+# Publishing a GitHub release (tag `vX.Y.Z`) runs `sync`, which scans the
+# commits since the last release for Linear issue IDs, attaches them, links the
+# GitHub release, and completes it in one step - continuous pipelines
 # auto-complete, so there is no separate `complete` call and no stage to move
-# between. The step is guarded so a Linear outage never fails a shipped release.
+# between.
 #
-# Manual path: `workflow_dispatch` is the backfill / repair tool - re-sync a
-# release, or preview the result with dry_run.
+# The sync fails loudly rather than fail-open: if Linear is unreachable or the
+# key is rejected, this run goes red and is re-run from the Actions UI. It
+# publishes nothing itself, so a red run here never blocks or reverts a release
+# that has already shipped.
 #
 # Setup: add the pipeline access key as the `LINEAR_ACCESS_KEY` repository
 # secret (Settings -> Secrets and variables -> Actions). This must be a pipeline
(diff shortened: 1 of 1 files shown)


PooyaRaki:
@claude review
@claude arch-review

PooyaRaki:
@claude arch-review

.github/workflows/linear-release.yml
```diff
@@ -0,0 +1,96 @@
...
+#
+# Setup: add the pipeline access key as the `LINEAR_ACCESS_KEY` repository
+# secret (Settings -> Secrets and variables -> Actions). This must be a pipeline
+# access key generated from the pipeline settings in Linear, not a personal API
+# key.
+
+on:
+  release:
+    types: [released]
+  workflow_dispatch:
```
PooyaRaki: Why do we need it?

gfendres: It's the recovery hatch for the fail-open sync.

The release-triggered sync is `continue-on-error`, so if Linear is down or the key is bad when we publish, it skips silently and that release never lands in the pipeline. Without a manual trigger the only way to fix that is cutting a fake release. With it you just re-run the sync for `1.122.0`.

Two other uses: backfilling releases that shipped before this workflow existed, and `dry_run` against a real version. That second one matters right now — our commit subjects don't carry Linear issue IDs (`feat: ... (#3329)`, not `feat: ... (WA-2930)`), so nobody knows yet whether the action actually picks anything up. The dry run answers that without writing to Linear.

Happy to drop it if you'd rather the automated path just fail loudly instead, but then a Linear hiccup starts turning shipped releases red.

---

PooyaRaki: Let's drop it.

gfendres: Done in `7a4d9b5`. Only trigger now is a published release.

I also flipped the sync to fail loudly, since that was the other half of the trade — without a manual re-run, a fail-open sync would drop the release silently with no way to recover. Now a Linear outage turns the run red and you re-run it from the Actions UI. Safe to be red here: this workflow publishes nothing, so it can't block or revert a release that already shipped.

Dropped everything that only existed for the manual path too — the `version`/`dry_run` inputs, their validation, and the release lookups. Tag and URL come off the event now, so no API calls. 96 → 68 lines.

One thing this costs us: we can no longer dry-run before the first real release. So the first published tag after merge is also the first test of whether the action finds any issues in our commits. If it comes back empty, that's the missing-issue-keys problem, not a broken workflow.

---

PooyaRaki: Approve the PR 👍

PooyaRaki:
Approve 👍

gfendres:
@PooyaRaki thanks — but GitHub is still recording these as comments, not approvals. All reviews on the PR show `COMMENTED`, so it's sitting at `mergeable_state: blocked`.

Needs the actual button: **Files changed → Review changes → Approve → Submit**. Can't do it from this side, it's my own PR.

---

PooyaRaki:
Please check all the findings and fix them.

gfendres:
All addressed. Both reviews ran against older commits, so their line numbers no longer exist — here's the mapping against `7a4d9b5`:

| # | Finding | Status |
|---|---|---|
| 1 | Manual dispatch checked out the launch branch, not the tag | **Fixed** in `5506983`. Now `ref: refs/tags/${{ github.event.release.tag_name }}` (L45) — the `\|\| github.ref` fallback is gone |
| 2 | Backfill runs lose the release link | **Gone** in `7a4d9b5`. `links` comes off the event (L68), always populated on a release trigger |
| 3 | `dry_run` expands to an empty string | **Gone** in `7a4d9b5`. The input was removed; the action's own default applies |
| 4 | No format validation on the version input | **Gone** in `7a4d9b5`. There's no typed input left. L53-56 guards an empty tag, same as `ci.yml`'s "Validate release tag" |
| 5 | Concurrency group can collide | **Gone** in `7a4d9b5`. Group is now just the tag (L30). The collision needed the `github.ref` fallback on the dispatch path |
| arch | PR body carried sections beyond Summary/Changes | **Fixed** — body rewritten to the template's two sections, risk moved into Summary prose |

Only #1 needed a real code change. The other four all pointed at the manual-trigger path, which you asked me to drop — so they dissolved with it. #5 I'd pushed back on originally, but it's moot now either way.

Worth re-running `@claude arch-review` on the current head if you want a fresh verdict rather than my word for it. The last one graded `058f9db`, three commits back.

---

PooyaRaki:
@claude review

gfendres:
Both fixed in `de4376d`.

**`timeout-minutes: 10`** — agreed, six hours to notice a hung sync is useless. Also matches what `claude-code-review.yml` (5) and `architecture-check.yml` (15) already do, so it's the house style here rather than a one-off.

**Dropped the empty-tag guard.** You're right that it's unreachable, and there's a second reason: checkout resolves `refs/tags/<tag>` on the step *before* it, so an empty tag already blows up there. The guard couldn't fire even if the event somehow shipped without one.

For anyone comparing — the near-identical guard in `ci.yml`'s `docker-publish-release` is *not* redundant. That job checks out the default ref, so nothing fails ahead of it and `RELEASE_TAG` feeds straight into the image tag. Different situation, worth keeping there.

---

PooyaRaki:
@claude review

gfendres:
Not taking this one.

It guards a trigger that doesn't exist. One review ago the argument was to delete the empty-tag guard because it "invites future readers to reason about a state the runtime doesn't produce" — this is that same thing, just pre-emptive instead of leftover. Can't apply the principle in one direction only.

There's also a substantive problem: `github.run_id` is unique per run, so the fallback means *no* concurrency control on the hypothetical path. That's a guess about what a future trigger would want, baked in now. Whoever adds that trigger is editing these exact four lines and will see the key.

Two triggers is where this becomes real. Happy to add it in the PR that adds one.

Everything else in the review is a non-issue and the earlier findings are all resolved, so from my side this is done — it just needs a formal approval to unblock.

---

PooyaRaki:
@claude arch-review

gfendres:
Fair catch — body now links [WA-2846](https://linear.app/safe-global/issue/WA-2846/integrate-github-with-linear-releases-web-mobile), the ticket that wired the same integration for web and mobile and explicitly left CGW out of scope. This is its follow-up.

Worth saying plainly: WA-2846 is `Done`, so it's the decision record rather than a live ticket. If you'd rather this had its own CGW issue, say the word and I'll open one — I didn't want to create workspace state unprompted.

---

PooyaRaki (approved):

PR #3368, opened 2026-08-19
fix: recognise v1.5.0 migrations for safes with unknown mastercopies

tongyaosafe:
## Summary

WA-2933

For Safes whose mastercopy the Transaction Service does not recognise (`version = null`), CGW cannot compute `safeTxHash` from the Safe itself, so it infers the version from the transaction: a delegate call to an official `SafeMigration` contract identifies the target version (`detectSafeMigration`, added in #2746). That fix hardcoded the `SafeMigration` version to `1.4.1` and the candidate target versions to `['1.3.0', '1.4.1']`.

Now that chains are being flipped to recommend v1.5.0, migration transactions for these Safes target the `SafeMigration` **1.5.0** contract, which the hardcoded check does not recognise. `safeTxHash` computation then fails for the exact population the inference exists for — rejecting proposals/confirmations with 422 and failing queue/history rendering with `MalformedHash`. This is the same defect #2746 fixed, reintroduced by the next contract release. (The web app routes some unknown-mastercopy Safes to the CLI, but that does not shield CGW: CLI-proposed migrations still land in the Transaction Service and must pass CGW's hash verification to render, and other clients propose through CGW directly.)

Instead of appending `'1.5.0'` to the literals — and re-opening this bug at the next release — the versions are now derived from `@safe-global/safe-deployments`:

- `SafeMigration` recognition checks the transaction target against **all** versions the package ships.
- Target-version inference builds its candidates from the L1/L2 singleton versions in the package, filtered to `>=1.3.0` (the range `SafeMigration` can target). Any version in that range produces the same EIP-712 domain, so which candidate the loop picks does not affect the computed hash — the derivation changes recognition, not hashing.

Future `SafeMigration`/singleton releases are now covered by a `safe-deployments` dependency bump, with no code change.

## Changes

- `src/domain/common/utils/deployments.ts` — add version-list getters derived from the `safe-deployments` package constants: `getSafeSingletonVersions`, `getSafeL2SingletonVersions`, `getSafeMigrationVersions`
- `src/domain/common/utils/safe.ts` — `detectSafeMigration` matches the transaction target against all known `SafeMigration` versions; target-version candidates come from `getSafeMigrationTargetVersions()` (singleton versions `>=1.3.0`, ascending) instead of a hardcoded list
- `src/domain/common/utils/__tests__/deployments.spec.ts` — cases for the three new getters
- `src/domain/common/utils/__tests__/safe.spec.ts` — migration-detection cases rewritten against package-derived versions (covers the 1.5.0 contract)

### DIFF
--- a/src/domain/common/utils/__tests__/deployments.spec.ts
+++ b/src/domain/common/utils/__tests__/deployments.spec.ts
@@ -7,7 +7,6 @@ import {
   getSafeL2SingletonVersions,
   getSafeMigrationVersions,
   getSafeSingletonVersions,
-  getSafeToL2MigrationVersions,
   getSafeToL2SetupVersions,
   isExtensibleFallbackHandlerDeployed,
   isFallbackHandlerDeployed,
@@ -67,14 +66,6 @@ describe('deployments', () => {
     it('should include all SafeMigration versions', () => {
       expect(getSafeMigrationVersions()).toEqual(
         expect.arrayContaining(['1.4.1', '1.5.0']),
-      );
-    });
-  });
-
-  describe('getSafeToL2MigrationVersions', () => {
-    it('should include all SafeToL2Migration versions', () => {
-      expect(getSafeToL2MigrationVersions()).toEqual(
-        expect.arrayContaining(['1.4.1']),
       );
     });
   });
--- a/src/domain/common/utils/__tests__/deployments.spec.ts
+++ b/src/domain/common/utils/__tests__/deployments.spec.ts
@@ -4,6 +4,10 @@ import { getAddress } from 'viem';
 import {
   getExtensibleFallbackHandlerVersions,
   getFallbackHandlerVersions,
+  getSafeL2SingletonVersions,
+  getSafeMigrationVersions,
+  getSafeSingletonVersions,
+  getSafeToL2MigrationVersions,
   getSafeToL2SetupVersions,
   isExtensibleFallbackHandlerDeployed,
   isFallbackHandlerDeployed,
@@ -39,6 +43,38 @@ describe('deployments', () => {
     it('should include all SafeToL2Setup versions', () => {
       expect(getSafeToL2SetupVersions()).toEqual(
         expect.arrayContaining(['1.4.1', '1.5.0']),
+      );
+    });
+  });
+
+  describe('getSafeSingletonVersions', () => {
+    it('should include all L1 singleton versions', () => {
+      expect(getSafeSingletonVersions()).toEqual(
+        expect.arrayContaining(['1.3.0', '1.4.1', '1.5.0']),
+      );
+    });
+  });
+
+  describe('getSafeL2SingletonVersions', () => {
+    it('should include all L2 singleton versions', () => {
+      expect(getSafeL2SingletonVersions()).toEqual(
+        expect.arrayContaining(['1.3.0', '1.4.1', '1.5.0']),
+      );
+    });
+  });
+
+  describe('getSafeMigrationVersions', () => {
+    it('should include all SafeMigration versions', () => {
+      expect(getSafeMigrationVersions()).toEqual(
+        expect.arrayContaining(['1.4.1', '1.5.0']),
+      );
+    });
+  });
+
+  describe('getSafeToL2MigrationVersions', () => {
+    it('should include all SafeToL2Migration versions', () => {
+      expect(getSafeToL2MigrationVersions()).toEqual(
+        expect.arrayContaining(['1.4.1']),
       );
     });
   });
--- a/src/domain/common/utils/__tests__/safe.spec.ts
+++ b/src/domain/common/utils/__tests__/safe.spec.ts
@@ -1,12 +1,15 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 import { faker } from '@faker-js/faker';
+import semverSatisfies from 'semver/functions/satisfies';
 import { getAddress, type Hex } from 'viem';
 import type { IBuilder } from '@/__tests__/builder';
 import { Builder } from '@/__tests__/builder';
 import {
   getSafeL2SingletonDeployments,
+  getSafeL2SingletonVersions,
   getSafeMigrationDeployments,
   getSafeSingletonDeployments,
+  getSafeSingletonVersions,
 } from '@/domain/common/utils/deployments';
 import type { BaseMultisigTransaction } from '@/domain/common/utils/safe';
 import {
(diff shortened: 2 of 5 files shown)


tongyaosafe:
@claude review

tongyaosafe:
@claude arch review

tongyaosafe:
@claude arch-review

tongyaosafe:
> **Claude finished Tong Yao (@tongyaosafe)'s task in 2m 30s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/32236697706)
> 
> ### Claude PR Review
> * [x]  Read changed files
> * [x]  Analyze code quality, bugs, security, tests
> * [x]  Post review
> 
> The derive-from-`safe-deployments` refactor is on point — the hash-equivalence argument for candidate ordering (`getSafeMigrationTargetVersions()` sort asc within `>=1.3.0`) is sound because all `>=1.3.0` domains and types are identical, so which candidate the loop picks does not change the resulting `safeTxHash`. A few things to consider:
> 
> ### Suggestions
> **1. Hardcoded singleton lists in `recipient-analysis.service.ts` are still there** — `src/modules/safe-shield/recipient-analysis/recipient-analysis.service.ts:65-72` (`SAFE_VERSIONS`) and `:82-86` (`MULTICHAIN_SUPPORTED_VERSIONS`).
> 
> The PR body says "future SafeMigration/singleton releases are now covered by a safe-deployments dependency bump, with no code change." That is true for `SafeMigration` and `SafeToL2Migration`, but `SAFE_VERSIONS` is consumed by `determineMasterCopyVersion` (`:596-613`) — a v1.6.0 mastercopy will not be recognized here without editing this list. Same holds for `MULTICHAIN_SUPPORTED_VERSIONS` used in `isNetworkCompatible` (`:642-646`). Consider deriving these from `getSafeSingletonVersions()` / `getSafeL2SingletonVersions()` in the same PR, or note it as a follow-up so the next release does not re-open a similar gap. [Fix this →](https://claude.ai/code?q=In%20src%2Fmodules%2Fsafe-shield%2Frecipient-analysis%2Frecipient-analysis.service.ts%20on%20branch%20tongyao%2Fwa-2933-cgw-recognise-the-150-migration-as-an-official-mastercopy%2C%20replace%20the%20hardcoded%20SAFE_VERSIONS%20(lines%2065-72)%20and%20MULTICHAIN_SUPPORTED_VERSIONS%20(lines%2082-86)%20with%20values%20derived%20from%20getSafeSingletonVersions%20and%20getSafeL2SingletonVersions%20so%20future%20singleton%20releases%20are%20recognised%20by%20determineMasterCopyVersion%20and%20isNetworkCompatible%20without%20a%20code%20change.&repo=safe-global/safe-client-gateway)
> 
> **2. Test still uses an inline literal version list** — `src/domain/common/utils/__tests__/safe.spec.ts:913`
> 
> ```ts
> const hasAnySingletons = ['1.3.0', '1.4.1', '1.5.0'].some((version) => { ... });
> ```
> 
> This defeats the purpose of deriving in production and will silently miss a future version. Use `getSafeSingletonVersions()` / `getSafeL2SingletonVersions()` (or the same `>=1.3.0` filter as production) so the guard tracks the package. [Fix this →](https://claude.ai/code?q=In%20src%2Fdomain%2Fcommon%2Futils%2F__tests__%2Fsafe.spec.ts%20line%20913%20on%20branch%20tongyao%2Fwa-2933-cgw-recognise-the-150-migration-as-an-official-mastercopy%2C%20replace%20the%20hardcoded%20%5B%271.3.0%27%2C%20%271.4.1%27%2C%20%271.5.0%27%5D%20list%20with%20a%20union%20of%20getSafeSingletonVersions%20and%20getSafeL2SingletonVersions%20filtered%20to%20%3E%3D1.3.0%20so%20the%20test%20tracks%20the%20safe-deployments%20package%20instead%20of%20drifting%20on%20new%20versions.&repo=safe-global/safe-client-gateway)
> 
> **3. Silent catch in `detectSafeMigration`** — `src/domain/common/utils/safe.ts:225`
> 
> Not new to this PR (pre-existing outer `try { ... } catch { return null; }`), but since the function is now the sole path for a real-world 422 failure mode, a debug/warn log on the caught error would help future incident triage instead of silently returning `null`.
> 
> **4. Minor: version-getter naming vs. semantics** — `getSafeSingletonVersions()` returns _all_ versions the package ships (including pre-1.3.0 like `1.0.0`, `1.1.1`, `1.2.0`). That is fine for the current consumer because `getSafeMigrationTargetVersions()` filters by range, but any future caller that assumes "these are singleton versions I can safely feed into a SafeTx hash flow" could be tripped by the older entries. A short JSDoc note (or a helper that already filters `>=1.3.0`) would make the contract explicit.
> 
> ### Nothing blocking
> Changes look well-scoped and behaviorally correct. Tests exercise both `1.4.1` and `1.5.0` `SafeMigration` deployments and verify hash equivalence against an explicit `1.3.0` version — that is the right invariant to lock in. • Branch: `tongyao/wa-2933-cgw-recognise-the-150-migration-as-an-official-mastercopy`

Fixed other comments. 1,3 are not fixed. Reasons:

1. Both lists already include 1.5.0, so there's no current gap. Deriving them changes the behaviour of determineMasterCopyVersion and isNetworkCompatible are out of scope for current task.

3. The caller already logs the full context via logMalformedSafeTxHash whenever hash computation fails (chainId/safe/tx/source), which is where triage actually starts. Adding a logger to a pure util in domain/common/utils would mean threading DI through the signature for a redundant debug line.

tongyaosafe:
> ## Architecture check
> Applicable guides: best-practices.md, testing.md, commits-and-prs.md, reviewing.md (+ remarks.md)
> 
> src/modules/safe-shield/recipient-analysis/recipient-analysis.service.ts:706 — No scope creep: the SafeToL2Migration rewrite and its supporting `getSafeToL2MigrationVersions` getter are behaviour-neutral by the PR description's own account, pulling a second module into a v1.5.0 SafeMigration fix (R-029). PR title — Allowed types: a behavior correction (unrecognised v1.5.0 `SafeMigration` causing 422s and `MalformedHash`) is typed `chore`, which the table reserves as a last resort and excludes in favour of `fix`.
> 
> **Verdict:** ❌ 2 finding(s)

fixed.

tongyaosafe:
I tested locally. The migration runs well for 1.3.0 / 1.4.1 -> 1.5.0

PooyaRaki:
@claude arch-review

tongyaosafe:
@claude arch-review

PooyaRaki (approved):

PR #3369, opened 2026-08-19
docs: codify staged delivery and transitive module registration

PooyaRaki:
## Summary

  Adjudicating the five Architecture Check findings on #3327 against the author's rebuttals showed the code needed only the log-level fix (which already landed in-PR); the other four traced back to gaps in the guides rather than defects in the PR. This encodes the accepted rebuttals so
  the next staged rollout isn't re-flagged:

  - The new-module checklist demanded a direct `app.module.ts` entry, but transitive wiring is normal here — 14 module directories have no directly-registered class at all (`backbone`, `blockchain`, `bridge`, `earn`, `email`, `entitlements`, `human-description`, `queues`, `siwe`,
  `staking`, `swaps`, `tokens`, `wallets`, `zerion`), and controller-bearing sub-modules like `delegate/routes/` and `csv-export/v1/` are reached through their parent too. The checklist now asks for a path from `app.module.ts`, direct or transitive, since reachability is the actual
  invariant.
  - Neither the YAGNI check in reviewing.md nor the canonical skeleton in module-structure.md had room for a declared multi-PR series, so a schema-first PR (tables whose readers land in a linked follow-up) and a `routes/` directory landing one PR ahead of its controller both produced
  findings. Both now exempt staged delivery — conditional on the PR description itself declaring the series and identifying the follow-up, so an undeclared "we'll use it later" is still a finding and the arch-check workflow can verify the condition from the description it already
  fetches.
  - The URI-versioning rule already keyed breaking changes on an *already-shipped client's* parsing, but didn't say how to treat a field no released client could ever have parsed (an endpoint flag-gated since introduction, or a field upstream never populated). Such a field may now change
  in place, but only when the PR description states the justification — runtime flag state and upstream behavior aren't visible in a diff.
 
  Follow-up commits address arch-check findings against this PR's own earlier commits, most of them the same mistake: a guide sentence that quantifies over the whole repo (a universal, or a closed count) is falsifiable by one counterexample, which is what the no-stale-guides check looks
  for. The first pass replaced the wrong `app.module.ts` condition with "every module that declares a controller registers directly" and enumerated "four" description-text requirements; declaring a controller turns out not to discriminate at all, and there are at least six such
  requirements. Both sentences now state what is invariant and leave the requirement set explicitly open. The duplicated-convention sites in other guides are reconciled in the same pass.

  No change to the arch-check workflow itself — it reads these guides at run time. The `cgw-*` skills are meant to be loaders, but `cgw-commits-and-prs` had restated the PR-body rule instead of pointing at it, so rewording that rule left the skill contradicting it; it now defers the
  enumeration to the doc, which is also what stops the next addition from needing two edits. A sweep over every skill confirms none of the others restates a rule this branch touches.

  ## Changes
  
  - `docs/agents/module-structure.md` — new-module checklist now requires a path from `src/app.module.ts`, direct or transitive, rather than a direct entry; staged-rollout paragraph in the canonical skeleton rule, whose follow-up bar is identification rather than a link so the first PR
  of a series can satisfy it
  - `docs/agents/reviewing.md` — staged-delivery sub-bullet under the YAGNI check
  - `docs/agents/api-dtos-and-validation.md` — never-shipped-field paragraph under URI versioning
  - `docs/agents/commits-and-prs.md` — PR-body rule now names every requirement other guides place on the description text, attributed to the owning guide and marked as an open set
  - `docs/agents/ARCHITECTURE.md` — pointer sentence after the duplicated module skeleton noting the staged-delivery qualification, keeping the conditions single-sourced in `module-structure.md`
  - `.claude/skills/cgw-commits-and-prs/SKILL.md` — the loader stops restating which justifications belong in the PR description and defers to the guide


PooyaRaki:
@claude arch-review

vseehausen (dismissed):

moisses89 (dismissed):

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude review

vseehausen (approved):

PR #3370, opened 2026-08-20
feat: handle HTTP 451 banned-Safe responses with a custom message

clovisdasilvaneto:
## Summary

The Transaction Service now answers every Safe-scoped endpoint with HTTP 451 Unavailable For Legal Reasons when a Safe is banned from the indexer (safe-global/safe-transaction-service#2966), reporting the reason under `detail`.

CGW already forwarded the 451 status correctly — `HttpErrorFactory` produces a `DataSourceError` carrying the upstream status, and `DataSourceErrorFilter` turns it into the response. The gap was the message: the factory only reads the `message` key from an upstream error body, and DRF sends `detail`, so a banned Safe reached clients as `451 {"code": 451, "message": "An error occurred"}`.

This adds `mapBannedSafeError`, which rewrites a 451 payload into the shape the factory reads, and applies it in the three datasources that call the Transaction Service. Clients now get `451 {"code": 451, "message": "Unavailable for legal reasons"}`. The upstream `detail` text is deliberately not forwarded — it is not client-facing copy and its shape is not part of a contract CGW relies on.

The mapping sits in those datasources rather than inside `HttpErrorFactory` because 451 only carries this meaning for the Transaction Service. Putting it in the shared factory was the first attempt, and it broke ~40 existing specs across `staking`, `community`, `recovery` and `data-decoder` that encode "any 4xx/5xx forwards `data.message`" and draw a random error status that can land on 451 — a global branch would have changed the error contract for every upstream and pulled four unrelated modules into this diff. Scoping it keeps the production change to five files and leaves those specs untouched.

The message is status-semantic rather than Safe-specific for the same reason: the helper feeds a shared factory, and the 451 status is the machine-readable signal clients switch on.

Within `TransactionApi.mapError`, the 451 check runs before the existing `nonFieldErrors` branch: the status identifies a banned Safe, so it decides regardless of payload shape, even one the Transaction Service does not currently pair with 451.

Two related behaviours are left as-is and worth a reviewer's attention:

- `GET /v1/safes` (multi-Safe overview) collects results with `Promise.allSettled` and drops rejected entries with a `warn` log, so a banned Safe is omitted from the list rather than surfacing 451. That is the existing behaviour for every upstream failure on that endpoint, and arguably right for a batch read; changing it would also change how 404s and 500s behave there, so it is not in this PR.
- 451 is not negatively cached. `CacheFirstDataSource` caches only 404s, so a banned Safe reaches the Transaction Service on every poll. The Transaction Service made its ban check a deliberately uncached indexed lookup so a ban applies immediately; caching it in CGW would delay un-banning by a TTL.

Verification: `yarn format`, `yarn lint` and `yarn build` clean; `yarn test` green (353 files, 6262 passed, 18 skipped). Every new test was checked to be load-bearing by reverting the production change it covers — including the `mapError` ordering fix, which fails against the previous branch order, and each of the three `SafeBalancesApi` funnels, which fail with their catch site reverted. Specs that draw a random error status were run under multiple `FAKER_SEED` values (with `staking` as an untouched control) to confirm no seed sensitivity.

Refs WA-2969.

## Changes

**Production**

- `src/datasources/errors/constants.ts` — new: `UNAVAILABLE_FOR_LEGAL_REASONS_STATUS` (451 is absent from Nest's `HttpStatus`) and the paired client-facing message
- `src/datasources/errors/helpers/banned-safe.helper.ts` — new: `isBannedSafeError` (status-only predicate) and `mapBannedSafeError`, which rewrites a 451 payload before `HttpErrorFactory` reads it and passes anything else through untouched
- `src/modules/transactions/datasources/transaction-api.service.ts` — apply it in the existing `mapError` seam, ahead of the `nonFieldErrors` branch, covering all 37 catch sites
- `src/modules/balances/datasources/safe-balances-api.service.ts` — apply it at the three catch sites (`getBalances`, `getBalance`, `getCollectibles`)
- `src/modules/csv-export/v1/datasources/export-api.service.ts` — apply it at the `/export/` catch site

**Tests**

- `src/datasources/errors/helpers/banned-safe.helper.spec.ts` — new: the 451 rewrite, the predicate (including a payload shape the Transaction Service does not send), the three pass-through cases, and the resulting `DataSourceError` once funneled through `HttpErrorFactory`
- `src/__tests__/faker.ts` — new shared `errorStatusCodeExcluding(...excluded)`; 451 now has dedicated handling, so a spec asserting the generic forward path must not draw it
- `src/modules/transactions/datasources/transaction-api.service.spec.ts` — the 31 `should forward a %s error` cases route through that helper instead of repeating the loop; `getSafe` gains a banned-Safe case and one pinning the branch order
- `src/modules/csv-export/v1/datasources/export-api.service.spec.ts` — same helper for its one random-status case
- `src/modules/balances/datasources/safe-balances-api.service.spec.ts` — new: parameterised over all three read methods, asserting a 451 yields the dedicated message and that any other status still forwards the upstream one. This is the only coverage of `getBalance`, which no route reaches (it bypasses the cache and its only caller is the no-fee relayer validation path); for its two siblings it complements the route-level cases below with fast, isolated coverage of the same wiring
- `src/modules/safe/routes/safes.controller.integration.spec.ts` — route-level 451 through `TransactionApi`
- `src/modules/balances/routes/balances.controller.integration.spec.ts` — route-level 451 through `SafeBalancesApi.getBalances`
- `src/modules/collectibles/routes/collectibles.controller.integration.spec.ts` — route-level 451 through `SafeBalancesApi.getCollectibles`

**Docs**

- `docs/agents/ARCHITECTURE.md` — document the `message`-key limitation and this one explicit exception in the Error handling section
- `docs/agents/api-dtos-and-validation.md` — refresh the "One error funnel per layer" canonical example, which quoted the `SafeBalancesApi` catch blocks this PR rewrote


PooyaRaki:
@claude review

PooyaRaki:
@claude arch-review

clovisdasilvaneto:
Addressed both reviews in 42f16be. Five findings acted on, one declined.

**Code review #1 — `mapError` ordering.** Fixed. `mapError` now guards with a new `isBannedSafeError` up front, so the status decides before any payload shape is read. I used an exported type guard rather than an identity comparison on the mapper's return value, so the intent reads directly and the status literal stays in one place. Added a test that a 451 carrying `nonFieldErrors` still gets the banned-Safe message — it fails against the previous ordering, so the fix is pinned rather than just asserted.

**Code review #2 — duplicated status loop.** Fixed. Extracted to `errorStatusCodeExcluding(...excluded)` in `src/__tests__/faker.ts`, as suggested. All three sites now call it; the local `errorStatusCodeExcludingBannedSafe` in the Transaction Service spec is gone.

**Code review #3 — `/collectibles` coverage.** Fixed, and the gap was bigger than the finding says: `src/modules/balances/datasources/safe-balances-api.service.spec.ts` **does not exist**, so "the datasource-level spec covers both" doesn't hold — `getCollectibles` and `getBalance` had no banned-Safe coverage at any level. Added a route-level `451 for a banned Safe` case to `collectibles.controller.integration.spec.ts`, mirroring the `/balances` one. (`getBalance` is reachable only from the no-fee relayer validation path and has no route of its own; it stays covered by the helper unit spec.)

**Code review #4 — naming.** Declining, with reasoning. `remarks.md`'s R-017 is "a name that describes the mechanism instead of the thing" — it asks for the domain thing, not the mechanism. `map451ResponsePayload` is the mechanism-flavoured option ("451", "payload"); `mapBannedSafeError` names what the caller is handling. The hypothetical also argues for keeping it: if another upstream returned 451 for an unrelated reason, we would want a *separate* mapper, not this one silently applied — that upstream is not banned-Safe, and reusing this helper would put the wrong message on its response. That the helper is applied only to the three Transaction Service datasources is the deliberate scoping decision this PR is built on, and the name records it.

**Arch check #1 — stale guide.** Fixed. `api-dtos-and-validation.md`'s "One error funnel per layer" canonical example quoted the `SafeBalancesApi` catch blocks this diff rewrote. It now leads with `data-decoder-api.service.ts` for the plain funnel (verified: three catch sites, all the bare form) and keeps `SafeBalancesApi` as the variant that threads an upstream payload fix-up through the same funnel.

**Arch check #2 — R-004 literal test data.** Fixed. The upstream text is now `faker.word.words()` in all five places. The `detail` **key** stays literal on purpose: that the reason arrives under `detail` and not `message` is exactly what the mapper exists for, so it is the assertion boundary R-004 exempts. Comments updated to say so.

Verification on the new head: `yarn format`, `yarn lint`, `yarn build` clean; `yarn test` green (352 files, 6256 passed, 18 skipped). The specs that draw a random error status were re-run under five `FAKER_SEED` values, including `staking` as an untouched control, to confirm the shared helper did not introduce seed sensitivity.

---

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude review

clovisdasilvaneto:
Second review: one finding acted on in 2982eff, one declined.

**Finding 2 — `getBalance` banned-Safe coverage.** Fixed, and the diagnosis was exactly right. Added `src/modules/balances/datasources/safe-balances-api.service.spec.ts`: a 451 from `networkService.get` now asserts `DataSourceError(UNAVAILABLE_FOR_LEGAL_REASONS_MESSAGE, 451)`, plus a companion case that any other error status still forwards the upstream `message` from that same catch block. I used the real `HttpErrorFactory` rather than a mock so the assertion lands on the `DataSourceError` a caller actually sees, and I verified the test is load-bearing by reverting just that one catch site — it fails without the mapper and passes with it. Scoped to the error funnel on purpose: `getBalance`'s success and token-not-found paths are pre-existing behaviour this PR does not touch, so covering them would be scope creep.

**Finding 1 — the duplicated predicate in `mapError`.** Declining, and flagging why this one is worth not churning on.

The suggested `const mapped = mapBannedSafeError(error); if (mapped !== error) return mapped;` form is what this PR had before 42f16be. I removed it deliberately: it is a needless intermediate (`remarks.md` R-026) whose correctness rests on an invariant the call site cannot see — that the helper returns the *same reference* when it does not match. `isBannedSafeError(error)` states the condition in the reader's terms; the identity comparison encodes it as a side effect of the helper's return contract.

More to the point, this finding reverses the previous review's request on the same three lines. That review asked for the banned-Safe case to be an early return so the status decides before any payload shape is read; 42f16be implemented that with the explicit guard *specifically* to avoid the identity trick. Swapping back would re-introduce what I removed one commit ago and re-open the finding the earlier review raised.

The cost being avoided is one `instanceof` plus one integer compare, on an error path. The review notes the current form "reads more literally" and calls it "a taste call, not a correctness one" — I agree, and on that basis I'd rather leave the three lines stable than trade them back and forth between two reviews. Happy to change it if a human reviewer prefers the other shape.

Verification on the new head: `yarn format`, `yarn lint`, `yarn build` clean; `yarn test` green (353 files, 6258 passed, 18 skipped). The new spec was also run under three `FAKER_SEED` values, since it draws a random non-451 status through the shared `errorStatusCodeExcluding` helper.

---

PooyaRaki:
@claude review

src/modules/safe/routes/safes.controller.integration.spec.ts
```diff
@@ -2395,4 +2400,43 @@ describe('Safes Controller', () => {
...
+            data: rawify([singletonBuilder().build()]),
+            status: 200,
+          });
+        case safeUrl:
+          return Promise.reject(
+            new NetworkResponseError(
+              new URL(safeUrl),
+              {
+                status: UNAVAILABLE_FOR_LEGAL_REASONS_STATUS,
+              } as Response,
```
PooyaRaki: Why are we casting? Can we create a Response object?

src/modules/balances/datasources/safe-balances-api.service.ts
```diff
@@ -105,7 +106,7 @@ export class SafeBalancesApi implements IBalancesApi {
       if (error instanceof ZodError) {
         throw error;
       }
-      throw this.httpErrorFactory.from(error);
+      throw this.httpErrorFactory.from(mapBannedSafeError(error));
```
PooyaRaki, line 109: Can we create a proper response mapper so that we can map other response codes as well if we ever decide to do so in the future?

src/modules/safe/routes/safes.controller.integration.spec.ts
clovisdasilvaneto: Good catch — no reason to cast here. `new Response(null, { status })` is valid for any 4xx/5xx and carries everything `HttpErrorFactory` reads, so I switched all 13 `NetworkResponseError` sites this PR adds to a real `Response` in e310c2c. Pre-existing `as Response` casts elsewhere in the suite (169 of them) are left alone to keep the diff to this PR's own lines.

Unit suite green (353 files, 6262 passed) and the three touched integration specs pass locally.

---

src/modules/balances/datasources/safe-balances-api.service.ts
clovisdasilvaneto: I'd rather you decide this one, because it runs into the repo's own review standard: `docs/agents/reviewing.md` lists "no speculative abstractions" as a Part 2 finding, and spells out that "a helper generalized (parameterized, exported, or made reusable) beyond what its one call site actually needs, or an abstraction built for an anticipated future need, is a finding on the same basis as literal dead code". A status→message table with one entry is exactly that shape, so I didn't want to add it unilaterally — the `@claude arch-review` job on this PR applies that same checklist.

That said, the extensible form is small and I'm happy to push it if you want it. It would replace `isBannedSafeError`/`mapBannedSafeError` with:

```ts
const MESSAGE_BY_STATUS: Readonly<Record<number, string>> = {
  [UNAVAILABLE_FOR_LEGAL_REASONS_STATUS]: UNAVAILABLE_FOR_LEGAL_REASONS_MESSAGE,
};

export function hasMappedMessage(error: unknown): error is NetworkResponseError {
  return (
    error instanceof NetworkResponseError &&
    error.response.status in MESSAGE_BY_STATUS
  );
}

export function mapUpstreamErrorMessage(error: unknown): unknown {
  if (!hasMappedMessage(error)) {
    return error;
  }
  return new NetworkResponseError(error.url, error.response, {
    message: MESSAGE_BY_STATUS[error.response.status],
  });
}
```

Adding a future status then costs one line in the table, and `TransactionApi.mapError`'s "status wins over payload shape" guard becomes "a status with a dedicated message wins" — same behaviour today, five call sites unchanged apart from the rename.

The one behavioural caveat worth naming: as soon as the table holds more than 451, the helper stops being Transaction-Service-specific in meaning, and it is applied only in the three TS-facing datasources. Any second entry would need to be a status that means the same thing for every upstream, or the mapping would have to move.

Say the word and I'll push it; otherwise I'll leave the single-purpose helper as is.

---

PooyaRaki: How are we mapping the response codes in other parts of the code?

PooyaRaki (approved):

src/modules/balances/datasources/safe-balances-api.service.ts
clovisdasilvaneto: Short answer: nowhere — the codebase forwards status codes rather than mapping them, and there is no status→message table anywhere in `src`. Concretely, three distinct patterns exist today:

**1. The generic funnel forwards the status verbatim.** `HttpErrorFactory.from()` (`src/datasources/errors/http-error-factory.ts:19`) takes `source.response.status` as the `DataSourceError` code untouched and reads only `data.message` for the text, falling back to `'An error occurred'`. `DataSourceErrorFilter` (`src/routes/common/filters/data-source-error.filter.ts:23`) then emits `{ code, message }`. Every upstream status reaches the client through that one path with no per-status logic.

**2. Where a datasource needs different text, it reshapes the payload — keyed on payload shape, not on status.** Three precedents:
- `TransactionApi.mapError` (`src/modules/transactions/datasources/transaction-api.service.ts:1201`) — lifts `nonFieldErrors[0]` into `message`
- `FirebaseCloudMessagingApiService.mapError` (`src/datasources/push-notifications-api/firebase-cloud-messaging-api.service.ts:223`) — first available error path
- `CoingeckoApi` (`src/modules/balances/datasources/coingecko-api.service.ts:282`) — reads `data.error_code`

**3. Where a status itself needs handling, it is an explicit named branch at the layer that cares — never a lookup.** Grepping `response.status ===` over non-spec `src` returns exactly three sites before this PR:
- `CacheFirstDataSource:146` and `:167` — 404 is negatively cached and rethrown from cache
- `TransactionApi:249` — 404 on `/safes/{address}` means "not a Safe", so it returns `false` rather than throwing
- `network.module.ts:150` — the circuit breaker counts only `>= 500` as a failure

There are zero `Record<number, …>` status maps in the repo.

So `isBannedSafeError` + `mapBannedSafeError` is pattern 2 and 3 combined, which is the existing convention: a named predicate for the one status that carries meaning, and a payload rewrite feeding the same shared funnel. A `MESSAGE_BY_STATUS` table would be the first of its kind here — which is why I'd rather add it only if you want it, and it's a one-commit change either way.

Worth noting the reason the funnel can't own this: 451 means "banned Safe" for the Transaction Service specifically. Putting the branch in `HttpErrorFactory` was my first attempt and it broke ~40 specs across `staking`, `community`, `recovery` and `data-decoder` that assert "any 4xx/5xx forwards `data.message`" — a shared table inherits that same problem the moment it holds a status another upstream uses differently.

---

PR #3371, opened 2026-08-21
docs: make arch-check findings numbered, linked and actionable

TimDaub:
## Summary

Commenting `@claude arch-review` on a PR runs the Architecture Check workflow: Claude reads this repo's `AGENTS.md` and `docs/agents/` guides and posts one comment listing where the diff breaks them. Those comments are hard to read and hard to act on — [this one on #3315][ex] is typical: five findings run together as a single paragraph, nothing linked, rules named with no way to look them up.

What this changes:

- Findings render as a numbered list. The prompt asked for "one line each", which a bare newline satisfies, so GitHub collapsed them into one block on some runs and rendered a list on others.
- Numbering means a reader can reply "point 3" and be understood.
- Every rule name and guide is a link. "No stale guides" tells you nothing on its own.
- No bare `#<number>`: GitHub turned "non-negotiable #9" into a link to an unrelated 2022 PR.
- Each finding ends with the change to make — the compliant PR title, the missing line — instead of restating the rule it broke.
- A footer says what ran and links the guides, so a reader who has never seen this check knows what it is.

Verified before/after by running the workflow's own prompt (extracted from the YAML, not retyped) against #3315 frozen at head `840996e`: two runs of the current prompt scored 1/10 on a set of mechanical format checks and reproduced the run-together paragraph; two runs of this prompt scored 10/10. Findings quality was unchanged — every finding in all four runs checked out against the tree, so this is a presentation change, not a behaviour change.

Takes effect once merged: `issue_comment` workflows always run from the default branch, so the comment format on an open PR does not change until this is on `main`.

[ex]: https://github.com/safe-global/safe-client-gateway/pull/3315#issuecomment-5366179172

## Changes

- `.github/workflows/architecture-check.yml` — the comment spec: numbered list, linked rule names, permalinks pinned to the reviewed commit, a `Fix:` per finding, no bare `#N`, explainer footer. Adds `headRefOid` to the `gh pr view --json` fetch so the permalinks can be pinned.
- `docs/agents/reviewing.md` — the canonical Reporting format rewritten to match, since the workflow prompt defers to it.
- `.claude/skills/cgw-review/SKILL.md`, `docs/agents/agent-tooling.md` — the two places that restate the reporting format, kept in sync.


PooyaRaki:
The only thing I’d like to discuss is the PR description.
I intentionally wanted Claude to check the PR descriptions and flag missing or extra sections as findings so that we can have a unified structure across all PRs.
The issue is that sometimes a PR description becomes a book that no one really reads, which is even worse than having just a Summary and Changes section.
If you think we should add more sections and make them optional, that’s fine. But I don’t want to give developers the possibility to define their own sections, because then Claude might start generating random text and we’ll lose the consistency we’re trying to achieve.

TimDaub:
OK, removed all changes related to PR description sections

PooyaRaki (approved):
LGTM 🚀

PR #3372, opened 2026-08-21
refactor: apply the Rhinestone migration review follow-ups

dasanra:
### Summary

Follow-ups agreed with reviewers on #3244 (Gelato → Rhinestone relay provider), deferred rather than pushed onto the approved branch. Tracked in PLA-1891.

Restores the `Raw<T>` validation boundary in the relay module: the datasource returns `Raw<T>` and `RelayRepository` is now the single runtime parse point for `RelaySchema`/`RelayTaskStatusSchema` (previously enforced only at compile time, and `RelaySchema` on just one of the three relayer paths). Also tightens the relay `chainId` params with `ChainIdSchema`, which adds the `CHAIN_ID_MAXLENGTH` bound on top of the existing numeric check — over-length ids now 422 at the pipe instead of reaching the service.

The ticket's item 5 (deriving relay test chain IDs from the deployments package) was implemented, evaluated, and declined — the sampling traded determinism for no fault-detection gain; rationale recorded on PLA-1891.

Now also carries the review follow-ups from #3373 (relay provider error logging), moved here by agreement with the reviewer: they are code-correctness cleanups rather than functionality, and #3373 needed to reach staging for QA without waiting on them. `formatError`, `describeUpstreamError` and `truncateForLog` held no instance state, so `datasources/helpers/rhinestone-error.helper.ts` now owns them — per remarks R-013, which puts helpers in a `helpers/` module rather than inline in a service. They are asserted directly now: 12 cases with no network mock and no mocked logger, where previously every formatting case had to be driven through `relay()` with a rejected mock. The message and error caps are exported so the tests derive their boundary inputs from them instead of restating the numbers.

One behavior change rides along, from the same review: blank `errors[]` entries are discarded *before* the count cap is applied rather than after, so a body whose leading entries carry no message still reports the ones that do. Previously three blank leading entries suppressed the diagnosis entirely.

## Changes

- `src/modules/relay/datasources/rhinestone-api.service.ts` — `relay`/`getTaskStatus` return `Raw<T>` via `rawify()`; provider-shape parses stay; stale `Raw` TODO removed from `getRelayCount`; the three error-formatting methods and their two constants move out, leaving two calls to `formatRhinestoneError`
- `src/domain/interfaces/relay-api.interface.ts`, `src/modules/relay/domain/interfaces/relayer.interface.ts`, the three relayers — signatures carry `Raw<Relay>`/`Raw<RelayTaskStatus>`; relay-fee drops its local `RelaySchema.parse`
- `src/modules/relay/domain/relay.repository.ts` — the single parse point for both domain schemas; new `relay.repository.spec.ts` covering valid/malformed payloads on both paths
- `src/modules/relay/routes/relay.controller.ts` — the three `chainId` params pipe `ChainIdSchema`; new 422 integration test for an over-length chainId
- `src/modules/relay/datasources/helpers/rhinestone-error.helper.ts` — new; `formatRhinestoneError` is the only export besides the two caps, the other two functions stay module-private
- `src/modules/relay/datasources/helpers/__tests__/rhinestone-error.helper.spec.ts` — new; 12 direct cases covering the whitelist, `context` omission, newline collapsing, both caps, and the blank-entry ordering
- `src/modules/relay/datasources/rhinestone-api.service.spec.ts` — the three formatting cases move to the helper spec; the remaining hard-coded `status`/`statusText` replaced with faker
- `src/modules/relay/datasources/schemas/rhinestone-error.schema.ts` — unused `RhinestoneErrorResponse` export removed
- `docs/agents/security.md` — `OpaqueIdSchema` added to the permitted upstream-URL schema list, with the constraint that the shared regex may only widen if every consumer (relay, billing) accepts it


PooyaRaki:
@claude arch-review

PooyaRaki (dismissed):

PooyaRaki:
@claude review

PR #3373, opened 2026-08-21
fix(relay): log the provider's rejection reason on relay failures

dasanra:
## Summary
A relay rejection from Rhinestone reached the logs as nothing but a status code. Rhinestone nests its reason under `errors[].message`, while `HttpErrorFactory` reads only `data.message`, so the reason hit the literal
`'An error occurred'` fallback and was thrown away.

Because the body is third-party-controlled text, only two fields are copied out of it: `errors[].message` and `traceId`. The `errors[].context` object is deliberately dropped — it echoes back the chain ID and addresses from the
request, which the structured-logging rule in `docs/agents/security.md` keeps out of log storage; the schema simply does not declare it, so Zod strips it.
Messages are whitespace-collapsed (newlines would let an upstream string forge log lines) and capped at 3 entries x 200 chars.

The client-facing response is unchanged and deliberately so: the throw still goes through `httpErrorFactory.from()`, so callers keep the generic `{code, message}` payload and the provider's text stays server-side. A test asserts that, so a later change wiring the detail into the throw fails rather than silently leaking it.

`getTaskStatus` shares `formatError` and gains the same detail.

## Changes
- `src/modules/relay/datasources/schemas/rhinestone-error.schema.ts` — new; the error envelope, all fields optional so parsing an error path cannot throw and mask the original failure
- `src/modules/relay/datasources/rhinestone-api.service.ts` — `formatError` appends the whitelisted fields; `describeUpstreamError` + `truncateForLog` helpers
- `src/modules/relay/datasources/rhinestone-api.service.spec.ts` — 5 cases: message + traceId, `context` absent, truncation/newline collapsing, the error-count cap, and the response staying generic

### DIFF
--- a/src/modules/relay/datasources/rhinestone-api.service.ts
+++ b/src/modules/relay/datasources/rhinestone-api.service.ts
@@ -22,6 +22,7 @@ import {
   LoggingService,
 } from '@/logging/logging.interface';
 import { asError } from '@/logging/utils';
+import { RhinestoneErrorResponseSchema } from '@/modules/relay/datasources/schemas/rhinestone-error.schema';
 import {
   type Relay,
   type RhinestoneRelayResponse,
@@ -32,6 +33,19 @@ import {
   type RhinestoneTaskStatusResponse,
   RhinestoneTaskStatusResponseSchema,
 } from '@/modules/relay/domain/entities/relay-task-status.entity';
+
+/**
+ * Cap on how many of a Rhinestone error body's `errors[]` entries are copied
+ * into a log line. Rhinestone can report several validation failures at once;
+ * the first few carry the diagnosis, the rest would only bloat the entry.
+ */
+const MAX_LOGGED_UPSTREAM_ERRORS = 3;
+
+/**
+ * Per-message character cap for upstream-controlled strings copied into a log
+ * line, so a long or hostile body cannot flood log storage.
+ */
+const MAX_LOGGED_UPSTREAM_MESSAGE_LENGTH = 200;
 
 @Injectable()
 export class RhinestoneApi implements IRelayApi {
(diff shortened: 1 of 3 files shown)


dasanra:
@claude arch-review

PooyaRaki:
@claude review

PooyaRaki (approved):

src/modules/relay/datasources/rhinestone-api.service.ts
```diff
@@ -61,20 +75,71 @@ export class RhinestoneApi implements IRelayApi {
...
+   * Collapses whitespace (including newlines, which would otherwise let an
+   * upstream string forge log lines) and caps length.
+   */
+  private truncateForLog(value: string): string {
+    const collapsed = value.replace(/\s+/g, ' ').trim();
+    return collapsed.length > MAX_LOGGED_UPSTREAM_MESSAGE_LENGTH
+      ? `${collapsed.slice(0, MAX_LOGGED_UPSTREAM_MESSAGE_LENGTH)}…`
+      : collapsed;
+  }
+
```
LucieFaire, line 142: I dont want to block the PR and testing of rhinestone, but these 3 func are pure functions with no instance state. Ideally they should live outside of the service,so that we can test them in isolation and not through relay() - as every formatting case has to drag along mockRejectedValue + rejects.toThrow + a mocked logger and so on.

LucieFaire (approved):

src/modules/relay/datasources/rhinestone-api.service.ts
dasanra: I will tackle this under #3372 instead, as functionality of this PR is accepted and we need it as soon as possible for easier debug.

PR #3374, opened 2026-08-24
fix(auth): keep the prior session expiry on step-up

TimDaub:
## Summary

A step-up round-trip minted a brand-new session, so elevating at hour 20 of a
24h session handed back a fresh lifetime instead of the 4h the user had left.
The session lifetime is an absolute timeout — its job is to bound how long any
one session lives, and a step-up proves only the second factor — so it must
not reset that clock. The elevated token now carries over the prior cookie's
`exp`, bounded by the max-validity constant.

With no live session to elevate — it expired while the user was on the
challenge page — the step-up now fails (`401`, surfaced to the client as the
usual `authentication_failed` redirect) instead of minting a fresh session:
elevation would otherwise double as a login whose first factor is only Auth0's
longer-lived SSO session. The user signs in again through the normal flow.
WA-3268.

## Changes

- `src/modules/auth/oidc/routes/oidc-auth.service.ts` — on elevation, require a
  valid prior session and carry its verified `exp` into the new token, bounded by
  `maxExpirationTime`; reject with `UnauthorizedException` when there is none
  (decode failures are debug-logged, message only, never the token)
- `src/modules/auth/oidc/routes/oidc-auth.controller.ts` — pass the session cookie
  into authentication on the callback
- `src/modules/auth/oidc/routes/oidc-auth.service.spec.ts` — carry-over, bound,
  rejected-when-undecodable, rejected-when-absent, and plain-login cases
- `src/modules/auth/oidc/routes/oidc-auth.controller.integration.spec.ts` — the
  elevate round-trip carries a live session cookie; asserts the carried-over `exp`
  and the refusal when no session is present

### DIFF
--- a/src/modules/auth/oidc/routes/oidc-auth.controller.integration.spec.ts
+++ b/src/modules/auth/oidc/routes/oidc-auth.controller.integration.spec.ts
@@ -692,9 +692,12 @@ describe('OidcAuthController', () => {
       async function roundTrip({
         elevate,
         amr,
+        priorSessionExp,
       }: {
         elevate: boolean;
         amr?: Array<string>;
+        /** When set, the callback arrives with a live session cookie expiring then. */
+        priorSessionExp?: Date;
       }): Promise<Record<string, unknown> | undefined> {
         vi.setSystemTime(NOW_MS);
 
@@ -729,9 +732,19 @@ describe('OidcAuthController', () => {
           .find((cookie) => cookie.startsWith('auth_state='))
           ?.split(';')[0];
 
+        const cookies = [stateCookie!];
+        if (priorSessionExp) {
+          const priorToken = jwtService.sign({
+            auth_method: 'oidc',
+            sub: faker.string.numeric({ exclude: ['0'] }),
+            exp: priorSessionExp,
+          });
+          cookies.push(`access_token=${priorToken}`);
+        }
+
         const callbackResponse = await request(app.getHttpServer())
           .get('/v1/auth/oidc/callback')
-          .set('Cookie', stateCookie!)
+          .set('Cookie', cookies)
           .query({ code: 'auth-code', state })
           .expect(302);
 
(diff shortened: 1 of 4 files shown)


PooyaRaki:
@claude review

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude review

PooyaRaki (dismissed):

TimDaub:
> 2\. **[PR title](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/commits-and-prs.md#L66)** — PR title — `fix: Session duration step up` has a capitalised, non-imperative noun-phrase description, and squash-merge makes it the permanent commit subject on `main`. **Fix:** retitle to `fix(auth): keep the prior session expiry on step-up`.

done

> 1. assertExpirationTime still bounds Auth0's exp on the elevate path

done

TimDaub:
> 1. **[Builders and faker only](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/testing.md#L68)** — [src/modules/auth/oidc/routes/oidc-auth.controller.integration.spec.ts:737](https://github.com/safe-global/safe-client-gateway/blob/0acc2918e5bdfb78f412a7840ab9b95a80d0debd/src/modules/auth/oidc/routes/oidc-auth.controller.integration.spec.ts#L737) — the prior session cookie is minted from a hand-written literal payload that re-types what `oidcAuthPayloadDtoBuilder()` already produces (same `auth_method`, same `faker.string.numeric({ exclude: ['0'] })` for `sub`), which `non-negotiable 6` and R-004 forbid; the sibling unit spec in this same PR uses the builder. **Fix:** add `import { oidcAuthPayloadDtoBuilder } from '@/modules/auth/domain/entities/__tests__/auth-payload-dto.entity.builder';` and replace the literal with `const priorToken = jwtService.sign({ ...oidcAuthPayloadDtoBuilder().build(), exp: priorSessionExp });`.

done

PooyaRaki (approved):

PR #3375, opened 2026-08-24
fix(users): reject account deletion by a space's last admin

tmjssz:
## Summary

`DELETE /v1/users` skipped the "Cannot remove last admin" guard that every other member-removal flow enforces. When a space's only active admin deleted their account, the `ON DELETE CASCADE` on `members.user_id` removed the admin row and left the space behind: nobody could be promoted, the space could not be renamed or deleted, and the remaining members could not even leave it. Twenty spaces are in this state in production today.

Account deletion now returns **409** while the caller is the last active admin of any space. They can promote another admin or delete the space, then retry — no new endpoint needed.

The guard is only sound if it cannot decide against a membership list another request is about to change, so this also closes three concurrency holes, each found by review and each reproduced against a real database first:

- The four paths that can drop an admin membership (account deletion, plus `updateRole`, `removeUser` and `removeSelf`) now lock the space row before reading its admin list. The three member paths previously read that list before opening a transaction at all.
- Account deletion locks the caller's `users` row first. A space created concurrently would otherwise never be in the lock set, and its only membership got cascaded away — leaving a space with zero members.
- `updateRole` locks the promoted member's `users` row too. A role change touches no foreign-key column, so it took no lock on that account by itself.

Locking order is documented for future paths: user rows first, then space rows in ascending id order.

A database-level "every space keeps an active admin" trigger was tried first and abandoned — it is incompatible with the fixtures of 7 integration specs, which deliberately construct member-only and admin-less spaces to exercise the paths that handle them.

Two things are deliberately **out of scope**, and need a follow-up: the twenty spaces already in this state, and the 404 that stops their remaining members from leaving (`removeSelf` requires an admin list to exist before it will let anyone out).

Refs [WA-2980](https://linear.app/safe-global/issue/WA-2980/spaces-deleting-the-last-admins-user-account-leaves-the-space-in-an-un)

## Changes

- `src/modules/users/domain/users.repository.ts` — reject the deletion with 409; lock the caller's account row and each administered space inside the deletion transaction
- `src/modules/users/domain/members/utils/members.utils.ts` — `isActiveAdmin` / `isLastActiveAdminOfSpace`, shared with `MembersRepository` so the rule has one definition, plus the two lock helpers and the ordering rule
- `src/modules/users/domain/members/members.repository.ts` — move the admin read inside the transaction behind the space lock on all three member paths; lock the promoted member's account row in `updateRole`
- `src/modules/users/routes/users.controller.ts` — document the 409
- Tests — 11 unit cases for the shared predicates; 6 integration cases for the guard; a separate `*.concurrency.integration.spec.ts` for the three race regressions, which needs two connections held at once and so cannot share a spec file's connection pool

### DIFF
--- a/src/modules/users/domain/members/members.repository.ts
+++ b/src/modules/users/domain/members/members.repository.ts
@@ -41,6 +41,7 @@ import {
   isActiveAdmin,
   isLastActiveAdminOfSpace,
   lockSpaceForAdminChange,
+  lockUserForAdminChange,
 } from '@/modules/users/domain/members/utils/members.utils';
 import { UserEncryptionService } from '@/modules/users/domain/user-encryption.service';
 import { IUsersRepository } from '@/modules/users/domain/users.repository.interface';
@@ -525,6 +526,11 @@ export class MembersRepository implements IMembersRepository {
     const actingUserId = getAuthenticatedUserIdOrFail(args.authPayload);
 
     await this.postgresDatabaseService.transaction(async (entityManager) => {
+      // A role change touches no FK column, so on its own it takes no lock on
+      // the target's account row - while their in-flight account deletion picks
+      // which spaces to lock from the memberships it can see. Lock that row so
+      // the two serialize; user rows before space rows, per the helper's doc.
+      await lockUserForAdminChange(entityManager, args.userId);
       await lockSpaceForAdminChange(entityManager, args.spaceId);
       const activeAdmins = await this.findActiveAdminsForUpdateOrFail(
         entityManager,
--- a/src/modules/users/domain/members/members.repository.ts
+++ b/src/modules/users/domain/members/members.repository.ts
@@ -40,6 +40,7 @@ import {
   activeOrPendingMemberWhere,
   isActiveAdmin,
   isLastActiveAdminOfSpace,
+  lockSpaceForAdminChange,
 } from '@/modules/users/domain/members/utils/members.utils';
 import { UserEncryptionService } from '@/modules/users/domain/user-encryption.service';
 import { IUsersRepository } from '@/modules/users/domain/users.repository.interface';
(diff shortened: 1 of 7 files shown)


tmjssz:
@claude review

tmjssz:
@claude review again

PooyaRaki:
@claude review

PooyaRaki:
@claude arch-review

tmjssz:
@PooyaRaki 

Regarding findings from Claude: keeping as-is.
1. Lock loop: correct, but it inlines lock semantics outside `lockSpaceForAdminChange`, whose docstring holds the ordering rule; not worth bypassing the helper.
2. findActiveAdmin: not a duplicate: a one-row `(user, space)` null-check vs. all admins of a space under lock.
3. Param comment: would be inaccurate: `space` also builds the lock set, and the docstring at `:202-211` already covers what the guard trusts.
4. Hand-wired specs: real footgun, but the existing baseline across 24 specs and already in the database guide; moving to the DI graph is a repo-wide refactor.

Architecture check:
1. Rejecting: There's nothing in `spaces/domain` to import instead, TypeORM needs a real class, and the one there is a type that disappears at runtime. The users module [already imports the entity](https://github.com/safe-global/safe-client-gateway/blob/fix-zombie-safe/src/modules/users/datasources/entities/member.entity.db.ts#L13) anyway, for the `space_id` foreign key, so this adds no new coupling. The rule is aimed at `routes/*` imports; entities shared across a foreign key aren't covered. Better fixed in `module-structure.md` than here.
2. Fixed - [00e14e6](https://github.com/safe-global/safe-client-gateway/pull/3375/commits/00e14e60ef01b1f823637d9e76d19ae82f0f7a03)

LucieFaire (approved):

PR #3376, opened 2026-08-24
ci: consolidate claude reviews into one two-pass workflow

PooyaRaki:
## Summary
  Replaces the two separate Claude CI workflows (`architecture-check.yml`, triggered by `@claude arch-review`, and `claude-code-review.yml`, triggered by `@claude review`) with a single `claude-review.yml` that runs both reviews in one pass. An orchestrator dispatches two independent
  subagents — a guideline pass applying `reviewing.md` against the AGENTS.md routing table, and a general quality pass for bugs, security, performance and test coverage — and posts one combined comment with an `## Architecture check` section (with verdict) and a `## Code review` section.
  Neither pass sees the other's findings, so the checklist pass cannot anchor the open-ended one.
 
  Triggering: on demand when a collaborator (`COLLABORATOR`/`MEMBER`/`OWNER`) comments `@claude review`, and automatically when a same-repo PR is marked ready for review — the automatic path first checks for an existing `claude[bot]` comment (by hidden marker or by the visible `##
  Architecture check` heading, so a dropped marker can't cause a repeat post) and skips if one exists, so it never duplicates a review. The security posture of the stricter deleted workflow is preserved and tightened: the PR head is never checked out (guides come from the default branch,
  PR content only via `gh pr view`/`gh pr diff`), all allowed `gh` commands are pinned to the triggering PR number (the pin gates subagent Bash calls too, since permission rules are session-wide), the `pull-requests: write` grant to `GITHUB_TOKEN` is gone (commenting happens via the
  OIDC-exchanged Claude App token), and the checkout sets `persist-credentials: false` so `GITHUB_TOKEN` never lands in `.git/config` on the runner. Both passes run on Claude Fable 5 at max effort.

  ## Changes
  - `.github/workflows/claude-review.yml` — new combined workflow: trigger gates, prior-review dedup step, `--agents` stub env var and its validation step, pinned `--allowedTools`
  - `.github/claude-review/` — new `orchestrator.md`, `guideline-pass.md` and `quality-pass.md` prompts, read at run time from the default-branch checkout, plus `agents.json` holding the `--agents` stub the workflow loads and validates
  - `.github/workflows/architecture-check.yml` — deleted, superseded
  - `.github/workflows/claude-code-review.yml` — deleted, superseded
  - `AGENTS.md` — delivery paragraph updated to describe the combined workflow
  - `docs/agents/agent-tooling.md` — "The CI gate" section rewritten for the two-pass design, its injection-containment reasoning, the `agents.json` mechanism, and the accepted residual risk that either subagent's `Bash` access can reach `gh pr comment` directly


PooyaRaki:
@claude review

PooyaRaki:
@claude review

PooyaRaki:
Evaluated all eight; four fixed, one addressed structurally, three declined with reasons.

  1. Subagent effort — declined, the claim is false. Verified empirically on the CLI (2.1.241) by capturing API request bodies against a mock endpoint: subagent requests inherit the session `--effort max` (`output_config: {"effort":"max"}`), and no per-agent `effort` key exists in
  `--agents` definitions — the suggested fix would be silently ignored. The docs claim stands.
  2. Trigger word boundary — declined after trying it. The only exposure is a collaborator starting a comment with a fused prefix (`@claude reviewer …`), and the consequence is one advisory review; the sibling-workflow collision is hypothetical and the naming constraint is documented in
  agent-tooling.md. A word-boundary gate was implemented and then removed as not worth its lines (GitHub expressions can't match a newline, so the check can't live in the job `if:` without breaking multi-line trigger comments).
  3. --agents validation — addressed structurally, and the premise was confirmed: the CLI silently accepts malformed `--agents` JSON and runs without agents. The 4.7k-char interpolated payload no longer exists — prompts moved to `.github/claude-review/*.md` and `--agents` is now a short
  static stub with no interpolation, so there is nothing left to break at run time.
  4. Duplicated `if:` — fixed. One gate step computes a single `run` output; checkout and review both reference it.
  5. cancel-in-progress — declined. A superseding event should review the newer state: cancelling the stale run costs less than queueing it (which would burn both runs and post two comments), and the superseding run always posts, so the PR is never left unreviewed.
  6. Dropping --paginate — declined, the suggestion introduces a bug. The issue-comments API returns oldest-first (not "bot comments at the top"), so a first-page-only check misses the marker on PRs with >100 comments and would cause duplicate reviews. Pagination costs a few extra
  requests in a rare case.
  7. Concurrency-group comment — fixed as suggested.
  8. Minors — `set -euo pipefail` added to the gate step; a comment above `claude_args` now points at the `--allowedTools` dependency for the subagents' Bash access; the long agent-tooling.md sentence is split.

  Note the workflow was also restructured since this review (prompts extracted to files, inline review-thread trigger dropped), so some fixes landed in that shape rather than as line edits to the reviewed diff.

PooyaRaki:
@claude review

PooyaRaki:
Point-by-point:
 
  1. **Confirmed, root-caused.** The fixes existed in a local commit that was never pushed, so this review ran against a stale remote — the prompt files, trimmed triggers, `set -euo pipefail` and concurrency comment were all already in `9015b92`. Both commits are on the branch now.
  2. **Fixed.** The agents JSON is defined once as a job-level `AGENTS_JSON` env var, referenced from `claude_args`, and a step before the review fails the job unless it parses and contains both agent keys. It uses `node -e` rather than `jq`: agent-tooling.md's own convention is that
  Node is the one runtime this repo guarantees. Verified by running the exact step command against the real value (passes), truncated JSON (fails) and a missing agent key (fails).
  3. **Fixed, in the docs.** The exact-string pin↔prompt coupling (including that only `--body:*` is a prefix match) is documented in agent-tooling.md's CI-gate section; this repo keeps workflow files comment-free and puts rationale in the guides.
  4. **Declined.** A deterministic post-edit via `PATCH /issues/comments/{id}` needs `issues: write` — the job deliberately grants `issues: read`, and the comment is posted through the action's app token — and a mechanical HTML strip would mangle code spans that merely look like tags
  (`Raw<T>` appears in real findings). Worst-case miss is formatting or a stray mention inside one advisory comment; recorded as an accepted residual risk in agent-tooling.md.
  5. Minor: 
     - Gate logging: **fixed** — the explicit-trigger branch now logs that it skips the dedup check.
     - Dedup marker: **declined** — the marker already sits before the `<sub>`, and the dedup check greps the raw body returned by the REST API, which stores the posted source verbatim; GitHub's HTML sanitizer affects rendering only, so it can never remove the marker from what the gate
  reads.
     - Effort-claim coupling: **fixed** — the claude-code-action v1.0.193 verification note and the re-verify-on-upgrade instruction are in agent-tooling.md next to the inheritance claim.

PooyaRaki:
@claude review

PooyaRaki:
**1 — Confirmed, root cause of half the review.** Two local commits (9015b92, 9ad8c93) containing the previously described fixes were never pushed, so this review ran against the stale 7ccbdc3. They are on the branch now — findings 2, 3, and two of the minors were already fixed in them.

  **2 — Fixed in 9015b92.** The gate step's `run:` block now opens with `set -euo pipefail`, so a transient `gh api` failure fails the job instead of silently degrading the dedup guard to `found=0`.
  
  **3 — Fixed in 9ad8c93.** The `--agents` value is now a job-level `AGENTS_JSON` env var, checked before the review by a `node -e` step that `JSON.parse`s it and asserts both agent keys exist. The three prompts were extracted to `.github/claude-review/*.md`, so the stub is short and static.

  **4 — Fixed, tighter than suggested.** The comment match is now exact: `github.event.comment.body == '@claude review'`. No prefix form at all — `@claude review please` no longer triggers (accepted trade-off), and trailing text can neither fire a sibling workflow nor reach the agent (the workflow prompt is fixed; the comment body is never interpolated anywhere).

  **5 — Accepted, with the justification now verified and documented.** Per the claude-code-action security docs, the app token that posts the comment is scoped exclusively to the repository the workflow runs in, so a smuggled second `--repo` fails auth; the PR number is positional and already supplied, so it cannot be overridden by appended flags. The escape leads nowhere. A `--body-file` exact-string pin was prototyped and rejected (it requires granting the agent a Write tool), as was a post-run assertion step (overengineering for an advisory workflow) and a separate posting step (requires `issues: write` on the job token). Rationale recorded in `agent-tooling.md`.

**6 — Minors:** dedup-skip log line was already in 9015b92; the v1.0.193 re-verify note was already in `agent-tooling.md` ("re-verify it when bumping the action pin"); the long CI-gate paragraph is now split at "Neither pass sees the other's findings". The cancel-in-progress YAML comment is declined — the design was already defended, and this thread now records the intent.

PooyaRaki:
@claude review

PooyaRaki:
@claude arch-review

PooyaRaki:
Findings 1–3 and the stale-guides item fixed:

  1. Normalized `AGENTS_JSON` tools strings to `"Read,Grep,Glob,Bash"` (no spaces), matching `--allowedTools`.
  2. Added `test -f` checks for all three prompt files to the JSON validator step, moved after checkout so the files actually exist when checked.
  3. Added a single-quote guard to the same validator — `AGENTS_JSON` containing `'` now fails the step instead of silently corrupting `--agents` interpolation.
  - Stale guides: fixed `agent-tooling.md:7` to mention the automatic `ready_for_review` trigger.
  
  Finding 4 (gh pr diff prefix-match risk) is refuted: `Bash(cmd)` without a `:*` suffix is an exact-string match in Claude Code, not a prefix match — this is already stated at `agent-tooling.md:65`. Only `--body:*` is a genuine prefix rule, so appended flags like `--patch`/`--web` don't
  match the pinned `gh pr diff` commands.

  Finding 5 minor nits (grep drift-guard, paragraph split, orchestrator defensive line) left as-is, per your own "not a blocker" framing.

PooyaRaki:
@claude review

PooyaRaki:
@claude arch-review

PooyaRaki:
@claude review

PR #3393, opened 2026-08-26
feat(policies): serve active spending-limit policies for a space

akshay-ap:
## Summary

Related to: https://linear.app/safe-global/issue/WA-3064/implement-get-policiesactive

Adds a REST API for reading the policies currently in effect on Safes in a Spaces.

`GET /v1/spaces/{spaceId}/policies/active` returns the active policies of every
Safe in a Space. Spending limits are the only policy type reported for now. Policies arrive
through an assembler registry, so reporting another mechanism is a new assembler
plus one registry entry; no change to the route service.

~#3394, next in the stack, is that second assembler: it registers `GuardPolicyAssembler`
through the same `POLICY_ASSEMBLERS` token, with no change to the route service.~

---

  - [x] Read the guides the routing table in AGENTS.md maps this diff to
  - [x] One concern per PR; nothing unrelated refactored or reformatted
  - [x] Env var? Declared in configuration.ts AND RootConfigurationSchema, added to
        .env.sample.json, mirrored in __tests__/configuration.ts, `yarn env:validate` clean
  - [x] Migration? Every FK and WHERE-target column indexed in the same migration; `down` reverts
  - [x] New/changed endpoint? Every input through `new ValidationPipe(Schema)`; DTO implements
        z.infer; guard on state-changing or caller-scoped routes
  - [x] Cached read added or TTL raised? Invalidation lands in this PR, or the trade-off is
        stated in this description
  - [x] SPDX license header on every file this PR touched, not only the new ones
  - [x] No debug scripts, seed helpers, or transient logging left in the branch

---

## Changes

**Route**

- `routes/space-policies.controller.ts` - the new endpoint, behind `AuthGuard`. Every input goes through a `ValidationPipe`.
- `routes/policies.service.ts` - asserts the caller is a member of the Space, resolves the Space's Safes, and fans out one indexer read plus the per-Safe module reads. Narrowing filters the Space's own Safes rather than mapping the request, so a Safe asked for twice - repeated outright, or in another casing - is still read once.
- `routes/mappers/spending-limit.mapper.ts` - turns the allowance module's rows into one policy per `(safe, module)`, with every spender and token nested inside. A limit on a module the Safe has not enabled is reported as `enabled: false`. `resetsAtMinute` walks forward from `lastResetMin` in whole periods to the first boundary after now, because the module only rewrites that value lazily, on a transfer.
- `routes/entities/policy.dto.entity.ts` - response DTOs for the OpenAPI schema. The shapes are the domain types, so there is no mapping layer to keep in sync.

**Domain**

- `domain/entities/active-policy.entity.ts` - an `ActivePolicy` now carries the `safe` it applies to. Allowance windows are expressed in minutes (`resetPeriodMinutes`, `resetsAtMinute`), matching the module's own unit, and `token_address` is now `tokenAddress` for consistent casing.
- `domain/entities/indexer/policy-indexer-state.entity.ts` - doc comment only; deriving the next boundary is the mapper's job now.

**Shared validation**

- `validation/entities/schemas/caip-10-addresses.schema.ts` - the CAIP-10 schema moved out of the safe module and made shareable, so both the Safes overview routes and this one parse `{chainId}:{address}` the same way. The chain id now goes through `ChainIdSchema`, and invalid input reports which half of the identifier is wrong.
- `modules/safe/routes/**` - updated to the moved schema; no behaviour change.

**Wiring**

- `policies.module.ts` - registers the controller, the service and the mapper.
- `app.module.ts` - mounts `PoliciesModule`.

**Tests**

- Unit tests for the route service, the spending-limit mapper and the CAIP-10 schema; an e2e spec covering the endpoint end to end.

## How a request for active policies is served

`GET /v1/spaces/:spaceId/policies/active?safes={chainId}:{address},...`

```mermaid

%%{init: {'theme':'dark', 'themeVariables': {
  'primaryTextColor':'#ffffff','textColor':'#ffffff',
  'actorBkg':'#22272e','actorTextColor':'#ffffff','actorBorder':'#8b949e','actorLineColor':'#8b949e',
  'signalColor':'#8b949e','signalTextColor':'#ffffff',
  'labelBoxBkgColor':'#22272e','labelBoxBorderColor':'#8b949e','labelTextColor':'#ffffff',
  'loopTextColor':'#ffffff',
  'noteBkgColor':'#2d333b','noteTextColor':'#ffffff','noteBorderColor':'#8b949e',
  'activationBkgColor':'#2d333b','activationBorderColor':'#8b949e',
  'sequenceNumberColor':'#ffffff'
}}}%%

sequenceDiagram
    autonumber
    actor Client
    participant Guard as AuthGuard
    participant Pipes as SpaceIdPipe + ValidationPipe
    participant Ctrl as SpacePoliciesController
    participant Svc as PoliciesService
    participant Members as MembersRepository
    participant Spaces as SpaceSafesRepository
    participant Repo as PolicyIndexerRepository
    participant Api as PolicyIndexerApi
    participant Cache as CacheService - Redis
    participant Net as NetworkService
    participant Indexer as Policy Indexer - GraphQL
    participant SafeRepo as SafeRepository
    participant Mapper as SpendingLimitMapper

    Client->>Guard: GET /v1/spaces/:spaceId/policies/active?safes=11155111:0xAAA
    Guard->>Guard: verify access_token JWT, build AuthPayload
    Note right of Guard: 403 if the cookie is missing or invalid

    Guard->>Pipes: admitted
    Pipes->>Pipes: SpaceIdPipe, UUID to numeric space id
    Note right of Pipes: 400 on a non-UUID
    Pipes->>Pipes: ValidationPipe with Caip10AddressesSchema.optional
    Note right of Pipes: 422 on a malformed chainId:address pair.<br/>The address is checksummed here.

    Pipes->>Ctrl: spaceId, authPayload, safes
    Ctrl->>Svc: getSpaceActivePolicies

    rect rgba(128, 128, 128, 0.15)
        Note over Svc,Spaces: 1. Authorize, then resolve WHICH Safes
        Svc->>Svc: getAuthenticatedUserIdOrFail
        Svc->>Members: assertMember of the Space
        Note right of Members: 403 if the caller is not a member
        Svc->>Spaces: findBySpaceId
        Spaces-->>Svc: every Safe in the Space
        Svc->>Svc: narrow to the safes query param, via isAddressEqual
        Note right of Svc: 422 if a requested Safe is not in the Space.<br/>Filtering the Space's own Safes means the query<br/>can only narrow, and never duplicates a Safe.
    end

    rect rgba(128, 128, 128, 0.15)
        Note over Svc,SafeRepo: 2. Two concurrent reads, Promise.all
        par policy state, one read covering every Safe
            Svc->>Repo: getState
            Repo->>Repo: checksum and de-duplicate the pairs
            Note right of Repo: the indexer stores addresses checksummed and<br/>a lower-cased filter matches nothing, silently
            Repo->>Api: getState

            Api->>Cache: hGet per Safe
            Cache-->>Api: hits and misses
            alt every Safe hit the cache
                Note right of Api: no upstream call at all
            else at least one miss
                Api->>Net: POST /v1/graphql, ONE request for all misses
                Note right of Net: circuit breaker key: policy-indexer
                Net->>Indexer: PolicyIndexerState query
                Indexer-->>Net: _meta, SafeAllowance, SafeDelegate
                Net-->>Api: body
                Api->>Api: GraphQlResponseSchema.parse
                Note right of Api: GraphQL reports failure as errors inside a 200,<br/>so it is logged, then raised through HttpErrorFactory
                Api->>Api: filterPoliciesStateBySafe, cut the answer per Safe
                Api->>Cache: hSet per Safe, TTL from config
            end
            Api-->>Repo: Raw payload, merged and still unvalidated

            Repo->>Repo: PolicyIndexerResponseSchema.parse on the envelope
            Repo->>Repo: parseRows on each of the three root fields
            Note right of Repo: an unreadable row is dropped and warn-logged,<br/>it never fails the whole read
            Repo->>Repo: withAllowanceDelegateStatus
            Note right of Repo: folds SafeDelegate.active onto the allowance<br/>as isDelegateActive, absent means false
            Repo-->>Svc: PolicyIndexerState
        and enabled modules, per Safe
            Svc->>SafeRepo: getSafe
            Note right of SafeRepo: cache-first, via the Transaction Service
            SafeRepo-->>Svc: safe.modules
        end
    end

    rect rgba(128, 128, 128, 0.15)
        Note over Svc,Mapper: 3. Map, one Safe at a time
        loop for each Safe
            Svc->>Svc: getAllowancesBySafe
            Note right of Svc: scopes the shared read to this Safe, so the<br/>mapper can never see another Safe's rows
            Svc->>Mapper: map with safe, allowances, enabledModules
            Note right of Mapper: drops zero-amount rows, groups by module then<br/>by delegate, and marks a policy enabled only<br/>while the Safe has that module enabled
            Mapper-->>Svc: ActivePolicy list, each carrying its safe
        end
    end

    Svc-->>Ctrl: ActivePolicy list
    Ctrl-->>Client: 200, array of ActivePolicyDto
```

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


PooyaRaki:
@claude arch-review

PooyaRaki:
@claude review

akshay-ap:
> **Claude finished Pooya Raki (@PooyaRaki)'s task in 2m 29s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/35227552431)
> 
> ### Claude PR Review
> Focused on new issues not already covered by the earlier `@claude arch-review`.
> 
> #### 🐛 Correctness / API contract
> **1. Swagger 401 vs actual 403; missing 503** — [`space-policies.controller.ts:75`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/routes/space-policies.controller.ts#L75) `@ApiUnauthorizedResponse` documents 401, but `AuthGuard` throws `ForbiddenException` — the e2e in this PR asserts `.expect(403)` (line 478). Also, `should fail when the indexer is unavailable` asserts `.expect(503)` (line 461), but that status is not documented. Replace `@ApiUnauthorizedResponse` and add `@ApiServiceUnavailableResponse({ description: 'The policy indexer is unavailable' })`.
> 
> **2. `nextResetMinute` breaks on `now < lastResetMin`** — [`spending-limit.assembler.ts:96-102`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/domain/assemblers/spending-limit.assembler.ts#L96) If the indexer serves a `lastResetMin` in the future (clock skew, or a same-block config the indexer indexed ahead of local clock), `elapsed` is negative and `Math.floor(elapsed / period) + 1` can yield 0 or a past boundary. Either guard with `Math.max(0, elapsed)` or document that `lastResetMin` is contractually `<= now`.
> 
> #### 🧪 Tests
> **3. R-002: `as never` in `mockResolvedValue`** — [`policies.service.spec.ts:61,63,98,306,317`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/routes/__tests__/policies.service.spec.ts#L61) Five sites use `mockResolvedValue({...} as never)`. The type `never` inhabits no value, so the doubles are no longer type-checked against `IMembersRepository`/`ISpaceSafesRepository`. Use `memberBuilder().build()` for the member and drop the `as never` on the space-safes doubles (they are already the shape `findBySpaceId` returns).
> 
> #### 🧹 Cleanup / YAGNI
> **4. Premature abstraction: `POLICY_ASSEMBLERS` token** — [`policies.module.ts:39-47`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/policies.module.ts#L39), [`policy-assembler.interface.ts`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/domain/assemblers/policy-assembler.interface.ts) The Symbol + interface + `useFactory` array back a single `SpendingLimitAssembler`, with no test-module swap. Inject `SpendingLimitAssembler` directly and reintroduce the abstraction when the second assembler lands (or name the follow-up PR in the description).
> 
> **5. Unused export `policyIndexerSafeDelegateBuilder`** — [`safe-allowance.builder.ts:109`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/domain/entities/indexer/__tests__/safe-allowance.builder.ts#L109) Not imported anywhere in the branch. Delete it (its raw sibling on line 65 is what the e2e uses).
> 
> **6. Misleading variable name** — [`policies.service.ts:93`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/routes/policies.service.ts#L93) `safesNotInSpace` (plural) is set from `.find()` and is a single `SafeRef | undefined`. Rename to `safeNotInSpace`.
> 
> **7. Unused `state.delegates` scoped through** — [`policy-state.utils.ts:29`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/modules/policies/domain/utils/policy-state.utils.ts#L29) `SpendingLimitAssembler` never reads `state.delegates` — `isDelegateActive` is already folded onto each allowance. Filtering delegates here is dead work. Either drop the field from `PolicyAssemblerContext` for now, or leave a comment naming the assembler that will consume it.
> 
> #### 📋 Convention
> **8. Chain-id bound re-implemented** — [`caip-10-addresses.schema.ts:8`](https://github.com/safe-global/safe-client-gateway/blob/feat/policies-03-active-allowance/src/validation/entities/schemas/caip-10-addresses.schema.ts#L8) `NumericStringSchema.max(CHAIN_ID_MAXLENGTH)` should be `ChainIdSchema` (from `@/modules/chains/domain/entities/schemas/chain-id.schema`), which layers the same bound and is what the rest of the codebase uses. Also matches R-011.
> 
> **Verdict:** 🟡 Mergeable after the correctness items (1, 2) and the mock `as never` cleanup (3). The rest are quality nits.
> 
> * [x]  Read changed files
> * [x]  Review for correctness, security, best-practices
> * [x]  Check tests
> * [x]  Post final review
>   • Branch: [`feat/policies-03-active-allowance`](https://github.com/safe-global/safe-client-gateway/tree/feat/policies-03-active-allowance)

1. Fixed. https://github.com/safe-global/safe-client-gateway/pull/3393/commits/c076e31f56c3c5312a92db9ea8104c78d0ac31ec

2. `lastResetMin` will be always `<=currentMin`. see: https://github.com/safe-fndn/safe-modules/blob/9e5f3d0701c4bd5d5fa4ce8ff7d5206f2ccc4acc/modules/allowances/contracts/AllowanceModule.sol#L148. https://github.com/safe-global/safe-client-gateway/pull/3393/commits/5f78d6fc12a07583f1f7000b3f9630090a08e8db

3. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/b8814563fe6a6a5c6add3bb3deec979eaf415867

4. Invalid. In #3394 we will have more assemblers. It is a planned change.
5. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/d60d820486501abd5f240acc5679a1c5de4216ef
6. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/79867eb4de05ced97e718946c25b8e12b2e9929f
7. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/a0f2a9e652e3ea88562c72cdefa0543d768ad72a
8. Fixed: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/a2a697eb7ea49b3dbc64a52bc1dfa168684cd97a

src/modules/policies/domain/assemblers/spending-limit.assembler.ts
```diff
@@ -0,0 +1,133 @@
...
+  SpendingLimitAllowance,
+  SpendingLimitPolicyData,
+} from '@/modules/policies/domain/entities/active-policy.entity';
+import type { PolicyIndexerSafeAllowance } from '@/modules/policies/domain/entities/indexer/policy-indexer-state.entity';
+import {
+  PolicyEnforcementKind,
+  PolicyType,
+} from '@/modules/policies/domain/entities/policy-type.entity';
+
+const MILLISECONDS_IN_MINUTE = 60_000;
```
PooyaRaki: It should be moved to another file.

src/validation/entities/schemas/caip-10-addresses.schema.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * An address scoped to the chain it lives on, CAIP-10 shaped:
+ * `{chainId}:{address}`.
+ *
+ * Note: a lower-cased address is accepted and checksummed, mirroring every
+ * other address in a CGW route. Anything else that is not a valid address is
+ * rejected with a 422 by the `ValidationPipe`.
+ */
+export const Caip10AddressSchema = z.string().transform((value, ctx) => {
```
PooyaRaki: Can we use this:
src/modules/safe/routes/entities/caip-10-addresses.entity.ts
We could extract the helper it to a shared folder and import it in different modules.

src/modules/policies/routes/__tests__/policies.e2e-spec.ts
```diff
@@ -0,0 +1,494 @@
...
+import { safeBuilder } from '@/modules/safe/domain/entities/__tests__/safe.builder';
+import { SpacesCreationRateLimitGuard } from '@/modules/spaces/routes/guards/spaces-creation-rate-limit.guard';
+import { rawify } from '@/validation/entities/raw.entity';
+
+const SEPOLIA_CHAIN_ID = '11155111';
+const DAY_IN_MINUTES = 1440;
+const MILLISECONDS_IN_MINUTE = 60_000;
+const POLYGON_CHAIN_ID = '137';
+
+describe('Policies routes (e2e)', () => {
```
LucieFaire, line 46: I think this test is more of an integration than e2e as you are mocking network calls. I would rename this test suite and the file to `.integration.spec.ts` so it runs in CI

src/modules/policies/routes/mappers/spending-limit.mapper.ts
```diff
@@ -0,0 +1,167 @@
...
+   */
+  private toAllowance(
+    allowance: PolicyIndexerSafeAllowance,
+  ): SpendingLimitAllowance {
+    const resetsPeriodically = allowance.resetTimeMinutes > 0;
+
+    return {
+      tokenAddress: allowance.token,
+      amount: allowance.amount,
+      spent: allowance.spent,
```
LucieFaire, line 96: `spent` is copied from the last indexed window while `resetsAtMinute` advances to the next future boundary. A fully spent allowance therefore still appears exhausted after its reset, potentially indefinitely without another transfer. Return `spent: '0'` when a periodic allowance has crossed its reset boundary.  
And maybe cover that case with a nonzero-spent test.

src/validation/entities/schemas/caip-10-addresses.schema.ts
```diff
@@ -0,0 +1,60 @@
...
+        code: 'custom',
+        path: issue.path,
+        message: issue.message,
+      });
+    }
+    return z.NEVER;
+  }
+
+  return parsed.data;
+});
```
LucieFaire, line 47: we can simplify, so them you dont need to do an internal parse:
```
z
  .string()
  .transform((value, ctx) => {
    const parts = value.split(':');

    if (parts.length !== 2) {
      ctx.addIssue({
        code: 'custom',
        message: 'Invalid CAIP-10 address, expected {chainId}:{address}',
      });
      return z.NEVER;
    }

    const [chainId, address] = parts;
    return { chainId, address };
  })
  .pipe(Caip10AddressPartsSchema);
```

PooyaRaki (dismissed):

src/modules/policies/routes/mappers/spending-limit.mapper.ts
akshay-ap: may be we should also expose `lastResetMin`. Will go over this again.

src/validation/entities/schemas/caip-10-addresses.schema.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/e450629ce070909ad9457951895a29ec2be70430

src/modules/policies/routes/mappers/spending-limit.mapper.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/95d962c75fe9955db45eb52cc9689b007721c143

src/modules/policies/routes/__tests__/policies.e2e-spec.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/9cbb662c4845aca2869e0b79ab1f1095fab595c6

LucieFaire (dismissed):

TimDaub:
The allowances carry only `tokenAddress`, so the wallet cannot render them without a second lookup. In safe-global/safe-wallet-monorepo#8768 the policies table needs each token's symbol, decimals and logo for three things: the token icons, the rule summary ("5,000 USDC / month"), and search by token.

  To get them, #8768 calls `/v1/chains/{chainId}/tokens` once per chain and per 20 tokens (`apps/web/src/store/api/gateway/policyTokenInfos.ts`). That is one extra request per chain on every page load, and it adds to the 429s we already see from staging.

  Could each allowance carry the token metadata instead, the way other CGW responses do? For example:

  `token: { address, symbol, decimals, logoUri }` in place of `tokenAddress`

  With that, #8768 can drop the lookup entirely. Sending `remaining` (amount minus spent) would also save the wallet computing it.

src/modules/policies/policies.module.ts
```diff
@@ -1,18 +1,36 @@
...
  */
 @Module({
+  imports: [
+    SafeRepositoryModule,
+    // Space membership and the Safe-in-space check
+    forwardRef(() => SpacesModule),
+    forwardRef(() => UsersModule),
+    forwardRef(() => AuthModule),
+  ],
+  controllers: [SpacePoliciesController],
```
LucieFaire: `HooksModule → EventCacheHelperModule` already imports `PoliciesModule` unconditionally. Adding `SpacePoliciesController` here therefore registers the new endpoint even when `FF_USERS=false`; the conditional import in `AppModule` cannot prevent it. These imports also pull in the Workspace controllers. Split the indexer repository providers into a module that hooks can import, and keep the route module behind the feature flag.

akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3393/commits/592054b62f160ab67a2cd2a846585c5d1a5fe015

akshay-ap:
> The allowances carry only `tokenAddress`, so the wallet cannot render them without a second lookup. In [safe-global/safe-wallet-monorepo#8768](https://github.com/safe-global/safe-wallet-monorepo/pull/8768) the policies table needs each token's symbol, decimals and logo for three things: the token icons, the rule summary ("5,000 USDC / month"), and search by token.
> 
> To get them, #8768 calls `/v1/chains/{chainId}/tokens` once per chain and per 20 tokens (`apps/web/src/store/api/gateway/policyTokenInfos.ts`). That is one extra request per chain on every page load, and it adds to the 429s we already see from staging.
> 
> Could each allowance carry the token metadata instead, the way other CGW responses do? For example:
> 
> `token: { address, symbol, decimals, logoUri }` in place of `tokenAddress`
> 
> With that, #8768 can drop the lookup entirely. Sending `remaining` (amount minus spent) would also save the wallet computing it.

I will do it in another PR.

PR #3397, opened 2026-08-27
fix: reject an empty chainId instead of returning 502

vseehausen:
Resolves: https://linear.app/safe-global/issue/WA-3510/fix-5xx-and-redirect-on-the-message-by-hash-endpoint

## Summary

`GET /v1/chains/:chainId/messages/:messageHash` returned 502 on **4.5%** of requests (1,168 5xx of 25,906 over 30 days). Every one arrived with an **empty `chainId`** — `/v1/chains//messages/0x…` — and no request carrying a real chain has ever 502'd on this route.

`@Param('chainId')` had no validation pipe, so `''` reached `ConfigApi.getChain`, which built `/api/v1/chains/` — the chain *collection* endpoint. That returns **200** with a page payload, `ChainSchema.parse` throws, and `ZodErrorFilter` maps the domain-level `ZodError` to `BAD_GATEWAY`. The page was then cached under the key `_chain`, so later requests failed identically from cache until the TTL expired. The transaction service is never called on this path: traces show a single outbound span (config service, 200) and no `*-safe-transaction-web` span — so this is neither an upstream fault nor a timeout.

## Fix

Validate at the boundary. `chainId` goes through `ChainIdSchema` and `messageHash` through `HexSchema` on all four routes of the messages controller, so a malformed value is rejected with **422** before any datasource is touched. Nothing changes below the controller: lower layers trust their callers.

This removes 99.3% of the endpoint's 5xx (1,160 of 1,168), taking its error rate from **4.51% to ~0.03%**. The remainder are eight genuine circuit-breaker 503s over 30 days, which clears the route to rejoin the CGW journey SLOs.

Verified locally against staging with real message hashes taken from `env:staging` traces: an empty chainId goes 502 → 422, real chains still return 200, wrong-chain lookups still return 404.

Independent of #3398 — either can merge first.

## Out of scope

38 other routes still take `@Param('chainId')` without a pipe, so an empty chainId on those still reaches `getChain`. Tracked in WA-3543.

## Changes

- `src/modules/messages/routes/messages.controller.ts` — `ChainIdSchema` / `HexSchema` pipes on all four routes.
- `src/modules/messages/routes/messages.controller.integration.spec.ts` — parametrized empty-chainId case across all four routes, asserting 422 and no upstream call.
- `src/modules/spaces/routes/safes/space-safes.controller.e2e-spec.ts` — length assertions read `CHAIN_ID_MAXLENGTH` instead of repeating the literals.
- `src/modules/chains/domain/entities/schemas/chain-id.schema.ts` — typo in the validation message.

### DIFF
--- a/src/datasources/config-api/config-api.service.spec.ts
+++ b/src/datasources/config-api/config-api.service.spec.ts
@@ -117,6 +117,15 @@ describe('ConfigApi', () => {
     expect(mockHttpErrorFactory.from).toHaveBeenCalledTimes(0);
   });
 
+  it('should reject an empty chainId without addressing the chain collection', async () => {
+    await expect(service.getChain('')).rejects.toThrow(
+      new DataSourceError('Chain not found', 404),
+    );
+
+    expect(mockDataSource.get).not.toHaveBeenCalled();
+    expect(mockHttpErrorFactory.from).toHaveBeenCalledTimes(0);
+  });
+
   it('should return the gas tokens retrieved by chainId', async () => {
     const chainId = faker.string.numeric();
     const data = [gasTokenBuilder().build(), gasTokenBuilder().build()];
--- a/src/datasources/config-api/config-api.service.spec.ts
+++ b/src/datasources/config-api/config-api.service.spec.ts
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
 
 import { faker } from '@faker-js/faker';
+import { HttpStatus } from '@nestjs/common';
 import type { MockedObject } from 'vitest';
 import { FakeConfigurationService } from '@/config/__tests__/fake.configuration.service';
 import type { CacheFirstDataSource } from '@/datasources/cache/cache.first.data.source';
@@ -119,7 +120,7 @@ describe('ConfigApi', () => {
 
   it('should reject an empty chainId without addressing the chain collection', async () => {
     await expect(service.getChain('')).rejects.toThrow(
-      new DataSourceError('Chain not found', 404),
+      new DataSourceError('Chain not found', HttpStatus.NOT_FOUND),
     );
 
     expect(mockDataSource.get).not.toHaveBeenCalled();
--- a/src/datasources/config-api/config-api.service.spec.ts
+++ b/src/datasources/config-api/config-api.service.spec.ts
@@ -14,6 +14,7 @@ import type { ILoggingService } from '@/logging/logging.interface';
 import { chainBuilder } from '@/modules/chains/domain/entities/__tests__/chain.builder';
 import { gasTokenBuilder } from '@/modules/fees/domain/entities/__tests__/gas-token.builder';
 import { safeAppBuilder } from '@/modules/safe-apps/domain/entities/__tests__/safe-app.builder';
+import { CHAIN_ID_MAXLENGTH } from '@/routes/common/constants';
 import { rawify } from '@/validation/entities/raw.entity';
 
 const dataSource = {
@@ -118,7 +119,7 @@ describe('ConfigApi', () => {
     expect(mockHttpErrorFactory.from).toHaveBeenCalledTimes(0);
   });
 
-  it.each(['', faker.string.alpha()])(
+  it.each(['', faker.string.alpha(), '1'.repeat(CHAIN_ID_MAXLENGTH + 1)])(
     'should reject chainId %j without addressing the chain collection',
     async (chainId) => {
       await expect(service.getChain(chainId)).rejects.toThrow(
@@ -313,7 +314,7 @@ describe('ConfigApi', () => {
       });
     });
 
-    it.each(['', faker.string.alpha()])(
+    it.each(['', faker.string.alpha(), '1'.repeat(CHAIN_ID_MAXLENGTH + 1)])(
       'should reject chainId %j without addressing the v2 chain collection',
       async (chainId) => {
         await expect(
--- a/src/datasources/config-api/config-api.service.spec.ts
+++ b/src/datasources/config-api/config-api.service.spec.ts
@@ -118,14 +118,17 @@ describe('ConfigApi', () => {
     expect(mockHttpErrorFactory.from).toHaveBeenCalledTimes(0);
   });
 
-  it('should reject an empty chainId without addressing the chain collection', async () => {
-    await expect(service.getChain('')).rejects.toThrow(
-      new DataSourceError('Chain not found', HttpStatus.NOT_FOUND),
-    );
+  it.each(['', faker.string.alpha()])(
+    'should reject chainId %j without addressing the chain collection',
+    async (chainId) => {
+      await expect(service.getChain(chainId)).rejects.toThrow(
+        new DataSourceError('Chain not found', HttpStatus.NOT_FOUND),
+      );
 
-    expect(mockDataSource.get).not.toHaveBeenCalled();
-    expect(mockHttpErrorFactory.from).toHaveBeenCalledTimes(0);
-  });
+      expect(mockDataSource.get).not.toHaveBeenCalled();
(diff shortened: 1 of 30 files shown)


vseehausen:
@claude review

PooyaRaki:
@vseehausen The remarks from Claude are worth reviewing.

PooyaRaki:
@claude arch-review

vseehausen:
Thanks both — pushed `4424a1bc`. Taking each point:

**arch-review 1 — `ChainIdSchema`.** Fixed. All four params now use `ChainIdSchema`. The guide is explicit that bare `NumericStringSchema` on route params "are also not precedent to extend", so this was the right call.

**arch-review 2 — PR body.** Rewritten to `## Summary` + `## Changes` only.

**arch-review 3 — scope.** Retitled to `fix: reject an empty chainId instead of returning 502`. The change does touch `src/datasources/config-api/`, so the scope shouldn't have been there.

**Review 3 — comment.** Trimmed to two lines naming the cache-poisoning, matching the file's style.

**Review 5 — `HttpStatus.NOT_FOUND`.** The spec now asserts `HttpStatus.NOT_FOUND` instead of the literal.

**Review 1 and 2 — the other 41 sites, and 422-vs-404 drift.** Both fair, and the body overclaimed. Reworded: the guard stops *this* cache-poisoning bug for every caller, but it is not a substitute for piping the remaining `@Param('chainId')` sites, and until those land an empty chainId answers 422 here and 404 elsewhere. Happy to file the fan-out as a follow-up.

**Review 4 — tests.** Half accepted, half not.

*Accepted:* only one of the four routes was exercised. There is now a parametrized case across all four. Worth flagging what that turned up — the naive version was **vacuous on the two POST routes**: with an empty body the `@Body` pipe fails first, so they returned 422 regardless of the chainId pipe and would have passed with the change reverted. The POST cases now send valid bodies from the existing builders, so `chainId` is genuinely the failing pipe.

*Not accepted — the `toThrow` claim.* The premise is that `DataSourceError`'s `super()` leaves `.message` empty. It doesn't: `readonly message: string` is a TypeScript parameter property, so the constructor assigns `this.message = message` after `super()`. And vitest's `toThrow(errorInstance)` compares more than the message here. Measured, all three against the real class:

| assertion vs thrown `('Chain not found', 404)` | result |
|---|---|
| `toThrow(new DataSourceError('Chain not found', 404))` | passes |
| `toThrow(new DataSourceError('Chain not found', 999))` | **fails** |
| `toThrow(new DataSourceError('Something else', 404))` | **fails** |

The failure reads `expected a thrown error to be Error: Chain not found { code: 999 }` — `code` is part of the comparison. So the existing assertion does pin the 404, and swapping to `toMatchObject` would not add coverage. Left as-is; happy to change it if you'd still prefer the more explicit form.

`yarn test` green (359 files / 6,350). Messages integration spec green (44).

PooyaRaki:
@claude arch-review

PooyaRaki (dismissed):

vseehausen:
Fixed in `c9e92558`.

The finding was right, and it applied more widely than the one line it flagged: **none** of the four routes declared 422, and all four now carry param pipes. Added `@ApiUnprocessableEntityResponse` to each, with a description naming that route's own validated params:

| Route | Description |
|---|---|
| `GET chains/:chainId/messages/:messageHash` | Invalid chainId or messageHash |
| `GET chains/:chainId/safes/:safeAddress/messages` | Invalid chainId or safeAddress |
| `POST chains/:chainId/safes/:safeAddress/messages` | Invalid chainId, safeAddress or request body |
| `POST chains/:chainId/messages/:messageHash/signatures` | Invalid chainId, messageHash or request body |

Worth noting the two POST routes could already return 422 before this PR — they've had `@Body(new ValidationPipe(...))` all along and declared only `@ApiBadRequestResponse`. So this closes a pre-existing gap as well as the one I introduced.

`yarn test` green (359 files / 6,350).

vseehausen:
@claude review (final)

vseehausen:
@claude review again

vseehausen:
@claude review

src/datasources/config-api/config-api.service.spec.ts
```diff
@@ -117,6 +119,18 @@ describe('ConfigApi', () => {
     expect(mockHttpErrorFactory.from).toHaveBeenCalledTimes(0);
   });
 
+  it.each(['', faker.string.alpha(), '1'.repeat(CHAIN_ID_MAXLENGTH + 1)])(
```
PooyaRaki, line 122: Nit: instead of repeating the string we could just randomly pick a number between 1 and the biggest possible chain id.

src/datasources/config-api/config-api.service.ts
```diff
@@ -71,6 +73,7 @@ export class ConfigApi implements IConfigApi {
   }
 
   async getChain(chainId: string): Promise<Raw<Chain>> {
+    this.assertChainId(chainId);
```
PooyaRaki, line 76: Instead of validating it in the service layer, why don’t we validate the `chainId` at the controller level?

src/datasources/config-api/config-api.service.ts
```diff
@@ -209,4 +213,12 @@ export class ConfigApi implements IConfigApi {
     const key = CacheRouter.getSafeAppsKey(chainId);
     await this.cacheService.deleteByKey(key);
   }
+
+  // A non-numeric chainId such as '' resolves to the chain collection endpoint, whose
```
PooyaRaki, line 217: Should have proper TsDoc comments or no comments.

src/modules/spaces/routes/safes/space-safes.controller.e2e-spec.ts
```diff
@@ -571,7 +571,7 @@ describe('SpaceSafesController', () => {
         .expect({
           statusCode: 422,
           code: 'custom',
-          message: 'Value must be less than or euqal to 78',
+          message: 'Value must be less than or equal to 78',
```
PooyaRaki, line 574: Instead of repeating 78 multiple times we should create a maxLength const here in the test file and fetch the number from that.

src/datasources/config-api/config-api.service.spec.ts
vseehausen: Not sure if this applies, as a valid chain id would be accepted, and this test would fail.

src/datasources/config-api/config-api.service.ts
vseehausen: Good question. We should do it. I created a follow up ticket, as there are other services that don't validate the chain id, neither. Ideally this can be done together. https://linear.app/safe-global/issue/WA-3543/add-chainid-validation-to-controller-level-and-add-type-check-to

src/datasources/config-api/config-api.service.ts
vseehausen: fixed in 089c4dbe

src/modules/spaces/routes/safes/space-safes.controller.e2e-spec.ts
vseehausen: fixed in 089c4dbe

vseehausen:
@claude review again

Fixes applied in 089c4dbe:

- `config-api.service.ts`: `assertChainId` now has a TSDoc comment instead of a `//` block.
- `space-safes.controller.e2e-spec.ts`: the two `Value must be less than or equal to 78` assertions read the bound from `CHAIN_ID_MAXLENGTH`, and the duplicated `chainIdMinLength = 79` is a single top-level `CHAIN_ID_MAXLENGTH + 1`.

src/datasources/config-api/config-api.service.ts
vseehausen: Agreed. The datasource guard is removed; validation stays at the controller only, and lower layers trust their callers. The other bare `@Param('chainId')` routes are tracked in WA-3543.

LucieFaire (approved):

PR #3398, opened 2026-08-27
fix(transactions): drop the redirect on every message-by-hash lookup

vseehausen:
Resolves: https://linear.app/safe-global/issue/WA-3510/fix-5xx-and-redirect-on-the-message-by-hash-endpoint

## Problem

`getMessageByHash` (`transaction-api.service.ts:1056`) builds `/api/v1/messages/{hash}` without a trailing slash. Django's `APPEND_SLASH` answers with a **301** to the slashed path, so every single-message lookup costs two round trips instead of one.

## Evidence

Production APM spans, 15 days, `service:safe-client-nest`:

| Upstream path | Status | Spans |
|---|---|---|
| `/api/v1/messages/?` (no slash) | **301** | 3,501 |
| `/api/v1/messages/?/` (followed) | 404 | 4,252 |
| `/api/v1/messages/?/` (followed) | 200 | 1,402 |

The slashless path returned **301 and nothing else**, across 25 chain hosts. The neighbouring URLs in the same file — `safes/{address}/messages/` and `messages/{hash}/signatures/` — already end in a slash; this one was the outlier.

## This fixes no errors

The redirect is followed and the correct data comes back, so this removes **zero** 5xx. It is a latency and request-count change only: one round trip instead of two on the busiest message path, on every chain. The 5xx on this endpoint are addressed by #3397.

## Notes

- **No cache invalidation.** The key comes from `CacheRouter.getMessageByHashCacheDir({ chainId, messageHash })`, not the URL, so existing entries stay valid.
- Test mocks for this endpoint updated to match; the `/signatures/` mocks are a different endpoint and are untouched.

## Verification

Run locally against staging with the message cache evicted first, so the new URL was genuinely exercised rather than served from cache: the message fetched **200**, and 404s and all other cases were unchanged from `main`. `yarn test` green — 359 files, 6,349 tests.

The 301 itself cannot be reproduced locally: both `api.5afe.dev` and `api.safe.global` answer 200/404 for either form. It occurs only on the in-cluster `*-safe-transaction-web…svc.cluster.local` host that deployed CGW uses, so the saved round trip is confirmed by the production span data above and will be visible on staging after merge.

Independent of #3397 — branched separately from `main`, either can merge first.

### DIFF
--- a/src/modules/messages/routes/messages.controller.integration.spec.ts
+++ b/src/modules/messages/routes/messages.controller.integration.spec.ts
@@ -108,7 +108,7 @@ describe('Messages controller', () => {
         switch (url) {
           case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
             return Promise.resolve({ data: rawify(chain), status: 200 });
-          case `${chain.transactionService}/api/v1/messages/${message.messageHash}`:
+          case `${chain.transactionService}/api/v1/messages/${message.messageHash}/`:
             return Promise.resolve({
               data: rawify(messageToJson(message)),
               status: 200,
@@ -176,7 +176,7 @@ describe('Messages controller', () => {
         switch (url) {
           case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
             return Promise.resolve({ data: rawify(chain), status: 200 });
-          case `${chain.transactionService}/api/v1/messages/${message.messageHash}`:
+          case `${chain.transactionService}/api/v1/messages/${message.messageHash}/`:
             return Promise.resolve({
               data: rawify(messageToJson(message)),
               status: 200,
@@ -241,7 +241,7 @@ describe('Messages controller', () => {
         switch (url) {
           case `${safeConfigUrl}/api/v1/chains/${chain.chainId}`:
             return Promise.resolve({ data: rawify(chain), status: 200 });
-          case `${chain.transactionService}/api/v1/messages/${message.messageHash}`:
+          case `${chain.transactionService}/api/v1/messages/${message.messageHash}/`:
             return Promise.resolve({
               data: rawify(messageToJson(message)),
               status: 200,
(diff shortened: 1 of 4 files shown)


PooyaRaki:
@vseehausen  It doesn't redirect without a slash it responds with a 200.

vseehausen:
@PooyaRaki You're right about the public API, and that's exactly what makes this confusing — but CGW doesn't call the public API.

I checked the same thing you did and got the same result. Both edges answer **200 for either form**:

```
api.safe.global/tx-service/eth/api/v1/messages/{hash}    -> 200
api.safe.global/tx-service/eth/api/v1/messages/{hash}/   -> 200
api.5afe.dev/tx-service/sep/api/v1/messages/{hash}       -> 200
api.5afe.dev/tx-service/sep/api/v1/messages/{hash}/       -> 200
```

Deployed CGW resolves `transactionService` to the **in-cluster** host, not the public edge, and that one does redirect. A real span from production this morning:

```
GET http://sepolia-safe-transaction-web.safe-transaction-sepolia.svc.cluster.local
      /api/v1/messages/0xc01cd000fb0981846ab34b75c8dd3187ead3d729d4bf58faa955918db40f876d
-> 301   (undici.request, 2026-08-28T09:31:56Z, peer.service: sepolia-safe-transaction-web)
```

That's Django's `APPEND_SLASH` on the service itself; the public edge sits in front of it and normalises the path, which is why curl from a laptop never sees it.

Aggregated over 15 days of production spans:

| Upstream path | Status | Spans |
|---|---|---|
| `/api/v1/messages/?` (no slash) | **301** | 3,501 |
| `/api/v1/messages/?/` (followed) | 404 | 4,252 |
| `/api/v1/messages/?/` (followed) | 200 | 1,402 |

The slashless path returned **301 and nothing else**, across 25 chain hosts — and 867 more in the last 3 days alone. The neighbouring URLs in the same file (`safes/{address}/messages/`, `messages/{hash}/signatures/`) already end in a slash; this one line was the outlier.

To be clear about what this does and doesn't do: the redirect is followed and the correct data comes back, so this fixes **no errors** — it removes one round trip per lookup. The 5xx on this endpoint are #3397's.

Caveat I've put in the description too: this isn't reproducible from a laptop for the reason above, so the saved round trip will only be observable once it's on staging.

vseehausen:
@claude review

vseehausen:
@claude review

LucieFaire (approved):

PR #3399, opened 2026-08-27
feat: add pre enforcement filtered payment links

moisses89:
## Summary

`GET /v1/billing/spaces/:spaceId/payment-links` returned the full upstream catalog, so a workspace saw offers it could not take. It now returns only what the workspace is eligible for, and `.../checkout-url` rejects anything outside that set with a 403 — otherwise the filtered list would only be a hint.

The rule: a workspace that has never subscribed is offered exactly one trial, the one matching its enforcement side (the legacy grace for workspaces created before `ENTITLEMENTS_ENFORCEMENT_STARTS_AT`, the standard trial otherwise). Once it has subscribed in any status, trials are off and every paid link is offered except the plan it is already on. Links negotiated for a single customer are always offered — they are not subject to the general rule. An absent or unmatched `planName` never excludes, so under-tagging cannot hide a purchase option.

Two rollout notes for reviewers. `ENTITLEMENTS_ENFORCEMENT_STARTS_AT` defaults to `2099-01-01`, so until it is set every workspace reads as pre-enforcement and is offered the legacy grace — the deliberate direction, since over-granting grace beats under-granting it. And a trial link whose `gracePeriod` metadata is missing or unrecognized is offered to nobody, which makes an untagged catalog look like an empty one; that case now logs a warning naming the dropped link ids rather than silently serving fewer offers.

Also here: `entitlements.enforcementStartsAt` is parsed to a `Date` once at config load and throws on an unparseable value, replacing the same parse-or-throw block that had been copied into two service constructors. `RootConfigurationSchema` does not cover this on its own — `configuration.validator` skips validation entirely under `NODE_ENV=test`, and an `Invalid Date` would read as "enforcement off, grace for everyone". No new env var, no migration.

One deliberate perf choice: reading the active plan name uses a new single-column repository query instead of `getActiveSubscriptionBySpaceId`, which hydrates the entitlements/feature relation tree to serve one string.

PLA-1870

## Changes

- `src/modules/billing/domain/` — `payment-link-offer.rules.ts`: the offer predicate and its `gracePeriod`/`planName` metadata readers, plus specs
- `src/modules/billing/routes/` — `billing.service.ts` filters the general catalog, merges space-specific links over it, gates checkout on the same rule, and warns on unclassified trial links
- `src/modules/entitlements/domain/` — `hasAnySubscription` and `getActivePlanName` on the subscriptions repository; `predatesEnforcement` rule; `GRACE_PERIOD_METADATA_KEY`
- `src/modules/spaces/domain/` — `findCreatedAtById` accessor, replacing two copies of the same hand-composed query
- `src/config/entities/` — parse `enforcementStartsAt` to a `Date` at load time, with specs for the default, a valid value, and malformed input
- `src/datasources/billing-api/entities/` — additive optional `trialPeriodDays` on the payment-link schema and DTO, plus a `trialPaymentLinkBuilder`

### DIFF
--- a/src/modules/billing/domain/payment-link-offer.rules.ts
+++ b/src/modules/billing/domain/payment-link-offer.rules.ts
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import type { PaymentLink } from '@/datasources/billing-api/entities/payment-link.entity';
+import {
+  GRACE_PERIOD_METADATA_KEY,
+  PLAN_NAME_METADATA_KEY,
+} from '@/modules/entitlements/domain/entitlements.constants';
+
+/** What the offer filter needs to know about the workspace. */
+export type SpaceOfferEligibility = {
+  createdBeforeEnforcement: boolean;
+  hasEverSubscribed: boolean;
+  activePlanName: string | null;
+};
+
+/**
+ * Which trial offer a link belongs to, read from its `gracePeriod` metadata:
+ * `true` for the legacy grace offered to pre-enforcement workspaces, `false`
+ * for the standard trial. `null` when the link carries no (or an
+ * unrecognized) value, which offers it to neither.
+ */
+export function gracePeriodOf(link: PaymentLink): boolean | null {
+  const value = link.metadata[GRACE_PERIOD_METADATA_KEY];
+  if (value === 'true') {
+    return true;
+  }
+  if (value === 'false') {
+    return false;
+  }
+  return null;
+}
+
+export function planNameOf(link: PaymentLink): string | null {
+  return link.metadata[PLAN_NAME_METADATA_KEY] ?? null;
+}
+
+/** Whether the link offers a free period; a paid link bills immediately. */
+function isTrialLink(link: PaymentLink): boolean {
+  return link.trialPeriodDays != null;
+}
+
+/** A trial link no workspace is offered, because its tag is missing or unknown. */
+export function isUnclassifiedTrialLink(link: PaymentLink): boolean {
+  return isTrialLink(link) && gracePeriodOf(link) === null;
+}
+
+/** Whether the workspace is offered this link. */
+export function isOfferedToSpace(
+  link: PaymentLink,
+  args: SpaceOfferEligibility,
+): boolean {
+  const isTrial = isTrialLink(link);
+
+  switch (true) {
+    // Already subscribed once: trials are off the table.
+    case isTrial && args.hasEverSubscribed:
+      return false;
+    // Never subscribed: offer only the trial matching this enforcement side.
+    case isTrial:
+      return gracePeriodOf(link) === args.createdBeforeEnforcement;
+    // Never subscribed: no paid link until a plan is picked first.
+    case !args.hasEverSubscribed:
+      return false;
+    // Subscribed: every paid plan is offered except the current one.
+    default:
+      return (
+        args.activePlanName === null || planNameOf(link) !== args.activePlanName
+      );
+  }
+}
--- a/src/modules/billing/routes/__tests__/billing.controller.e2e-spec.ts
+++ b/src/modules/billing/routes/__tests__/billing.controller.e2e-spec.ts
@@ -115,8 +115,6 @@ describe('BillingController', () => {
   afterEach(() => {
     networkService.get.mockReset();
     networkService.post.mockReset();
-    // The general payment-link catalog is cached under a customer-independent
-    // key, so without this a previous test's catalog is served to the next one.
     fakeCacheService.clear();
   });
 
--- a/src/modules/billing/routes/__tests__/billing.controller.e2e-spec.ts
+++ b/src/modules/billing/routes/__tests__/billing.controller.e2e-spec.ts
@@ -17,9 +17,15 @@ import {
   checkoutSessionBuilder,
   checkoutSessionResultBuilder,
 } from '@/datasources/billing-api/entities/__tests__/checkout-session.builder';
-import { paymentLinkBuilder } from '@/datasources/billing-api/entities/__tests__/payment-link.builder';
+import {
+  paymentLinkBuilder,
+  trialPaymentLinkBuilder,
+} from '@/datasources/billing-api/entities/__tests__/payment-link.builder';
 import { planBuilder } from '@/datasources/billing-api/entities/__tests__/plan.builder';
 import { subscriptionBuilder } from '@/datasources/billing-api/entities/__tests__/subscription.builder';
+import type { PaymentLink } from '@/datasources/billing-api/entities/payment-link.entity';
+import type { FakeCacheService } from '@/datasources/cache/__tests__/fake.cache.service';
+import { CacheService } from '@/datasources/cache/cache.service.interface';
 import { IJwtService } from '@/datasources/jwt/jwt.service.interface';
 import {
   type INetworkService,
@@ -41,6 +47,7 @@ describe('BillingController', () => {
   let networkService: MockedObject<INetworkService>;
   let billingBaseUri: string;
   let postLoginRedirectUri: string;
+  let fakeCacheService: FakeCacheService;
 
   beforeAll(async () => {
     vi.resetAllMocks();
@@ -53,6 +60,14 @@ describe('BillingController', () => {
         auth: true,
         users: true,
         billingService: true,
+      },
+      entitlements: {
+        ...defaultConfiguration.entitlements,
+        // In the past, so a workspace a spec creates now reads as
+        // post-enforcement and is offered the standard trial rather than the
+        // legacy grace. The shared default is a future date instead (safe for
+        // suites unrelated to billing: enforcement stays inactive).
+        enforcementStartsAt: new Date('2020-01-01T00:00:00Z'),
       },
       billing: {
         ...defaultConfiguration.billing,
@@ -84,6 +99,7 @@ describe('BillingController', () => {
 
     jwtService = moduleFixture.get<IJwtService>(IJwtService);
     networkService = moduleFixture.get(NetworkService);
+    fakeCacheService = moduleFixture.get<FakeCacheService>(CacheService);
     const configurationService = moduleFixture.get<IConfigurationService>(
       IConfigurationService,
     );
@@ -99,6 +115,9 @@ describe('BillingController', () => {
   afterEach(() => {
     networkService.get.mockReset();
     networkService.post.mockReset();
+    // The general payment-link catalog is cached under a customer-independent
+    // key, so without this a previous test's catalog is served to the next one.
+    fakeCacheService.clear();
   });
 
   afterAll(async () => {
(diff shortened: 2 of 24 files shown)


moisses89:
@claude review

moisses89:
https://github.com/safe-global/safe-client-gateway/pull/3399#issuecomment-5441694162
1. Done: folded both into a single getSubscriptionSummary aggregate.
2. Acknowledged: keeping it: billing service exposes no fetch-one.
3. Fixed: now warns once per link id per process.
4. Keeping it: the switch shape was explicitly requested in review.
5. Done: same change as 1; hasAnySubscription removed.
6. Keeping the throw: matches main's behavior; empty means misconfiguration.

moisses89:
@claude arch-review

moisses89:
https://github.com/safe-global/safe-client-gateway/pull/3399#issuecomment-5450416500
**Bounded in-process memos**:  Memo removed entirely; now logs an error on every occurrence.
**R-006**: Asserting that an error was thrown, not which — Fixed; now pins the message, matching the billing sibling.

src/modules/billing/routes/billing.service.ts
```diff
@@ -121,6 +142,60 @@ export class BillingService {
...
+      this.billingApi.listPaymentLinks({
+        upstreamCustomerId: args.spaceUuid,
+      }),
+      this.billingApi.listPaymentLinks(),
+      this.getOfferEligibility(args.spaceId),
+    ]);
+
+    const unclassified = generalLinks.filter(isUnclassifiedTrialLink);
+    if (unclassified.length > 0) {
+      this.loggingService.error(
```
falvaradorodriguez, line 165: This fires on every call, and `listOfferedPaymentLinks` is hit by both endpoints (`getSpacePaymentLinks` and `createCheckoutUrl`'s 403 gate), so one mistagged link produces an error log per `GET /payment-links` and per `POST /checkout-url`, indefinitely. What it reports is a static property of the general catalog, a single shared cached key (`billing_payment_links`, `billing-api.service.ts:162`), so we're logging at request cadence a fact that changes once per TTL, at the level that trips error rate alerting. The request itself succeeds; it's an operational misconfiguration to fix in Stripe, which is what `warn` is for and what the repo uses elsewhere for upstream data inconsistencies.

src/modules/billing/domain/payment-link-offer.rules.ts
```diff
@@ -0,0 +1,69 @@
...
+
+  switch (true) {
+    // Already subscribed once: trials are off the table.
+    case isTrial && args.hasEverSubscribed:
+      return false;
+    // Never subscribed: offer only the trial matching this enforcement side.
+    case isTrial:
+      return gracePeriodOf(link) === args.createdBeforeEnforcement;
+    // Never subscribed: no paid link until a plan is picked first.
+    case !args.hasEverSubscribed:
```
falvaradorodriguez, line 61: A never-subscribed workspace can only be offered a trial link, and a link only counts as a trial if it carries `trialPeriodDays`. If upstream isn't emitting it yet, every link reads as paid and every new workspace gets an empty catalog plus a 403 on checkout. The unclassified-trial log doesn't cover it either: `isUnclassifiedTrialLink` short-circuits on `isTrialLink`, so a missing `gracePeriod` is logged but a missing `trialPeriodDays` is silent. Worth logging when the general catalog yields zero offers for a space that has never subscribed, and confirming the deploy order against the billing service.

src/modules/billing/domain/payment-link-offer.rules.ts
```diff
@@ -0,0 +1,69 @@
...
+      return false;
+    // Subscribed: every paid plan is offered except the current one.
+    default:
+      return (
```
falvaradorodriguez, line 65: `activePlanName` reads `ACTIVE_SUBSCRIPTION_STATUSES` (`subscriptions.repository.ts:61`), which is `['active', 'trialing']`, so a workspace trialing "Pro" has the paid Pro link excluded here and 403s on checkout for it. Grace periods carry no card, so they never auto convert: every pre-enforcement workspace that wants to stay on its plan when the 2 month grace ends has to go through checkout, and this is the link that hides it.

src/modules/entitlements/domain/subscriptions.repository.ts
```diff
@@ -41,6 +45,30 @@ export class SubscriptionsRepository implements ISubscriptionsRepository {
...
+  ): Promise<SpaceSubscriptionSummary> {
+    const repository = await getScopedRepository(
+      this.postgresDatabaseService,
+      SpaceSubscription,
+    );
+    // `MAX` collapses the active rows, of which the partial unique index
+    // allows at most one.
+    const summary = await repository
+      .createQueryBuilder('subscription')
+      .select('COUNT(*) > 0', 'hasEverSubscribed')
```
falvaradorodriguez, line 59: **Nit:** `hasEverSubscribed` counts every status, so an `incomplete_expired` row (initial payment never succeeded, no money moved) permanently turns trials off for the workspace, with no way back since those rows stay as history. Reachable via a negotiated space specific link whose card bounces: a pre-enforcement workspace silently loses its 2 month grace. `canceled`/`past_due`/`unpaid` are right to count; `incomplete_expired` looks like the wrong one.

falvaradorodriguez (approved):

src/modules/billing/routes/billing.service.ts
moisses89: Correct, staying at error; this needs the Stripe config fixed ASAP.

src/modules/billing/domain/payment-link-offer.rules.ts
moisses89: Stripe config issue, reading it as paid is correct. No code change.

src/modules/billing/domain/payment-link-offer.rules.ts
moisses89: Repurchase is handled in the frontend; it reappears in payment-links only if the trial ends unpaid.

src/modules/entitlements/domain/subscriptions.repository.ts
moisses89: No change needed, free plan needs no payment; unpaid ones auto-cancel in Stripe, freeing another plan choice.

src/modules/billing/domain/payment-link-offer.rules.ts
falvaradorodriguez: Perhaps we could set up an alert for this. WDYT @safe-global/safe-client-gateway-maintainers?

falvaradorodriguez (dismissed):

src/modules/billing/domain/payment-link-offer.rules.ts
```diff
@@ -0,0 +1,69 @@
...
+export function gracePeriodOf(link: PaymentLink): boolean | null {
+  const value = link.metadata[GRACE_PERIOD_METADATA_KEY];
+  if (value === 'true') {
+    return true;
+  }
+  if (value === 'false') {
+    return false;
+  }
+  return null;
+}
```
LucieFaire, line 30: nit, can be shorter (but maybe its harder to read):
```
export function gracePeriodOf(link: PaymentLink): boolean | null {
  const value = link.metadata[GRACE_PERIOD_METADATA_KEY];
  return value === 'true' ? true : value === 'false' ? false : null;
}
```

src/modules/billing/routes/__tests__/billing.controller.e2e-spec.ts
```diff
@@ -99,6 +115,9 @@ describe('BillingController', () => {
   afterEach(() => {
     networkService.get.mockReset();
     networkService.post.mockReset();
+    // The general payment-link catalog is cached under a customer-independent
+    // key, so without this a previous test's catalog is served to the next one.
```
LucieFaire: nit: i think we dont need a comment here, many tests clear the cache, we have it in a lot of places

src/modules/billing/routes/billing.service.ts
```diff
@@ -121,6 +142,60 @@ export class BillingService {
...
+    }
+
+    const offeredGeneralLinks = generalLinks.filter((link) =>
+      isOfferedToSpace(link, eligibility),
+    );
+    const linksById = new Map(
+      [...offeredGeneralLinks, ...spaceLinks].map((link) => [link.id, link]),
+    );
+    return Array.from(linksById.values());
+  }
```
LucieFaire, line 181: nit:
```
const offeredGeneralLinks = generalLinks.filter((link) =>
  isOfferedToSpace(link, eligibility),
);

return [
  ...new Map(
    [...offeredGeneralLinks, ...spaceLinks].map((link) => [link.id, link]),
  ).values(),
];
```

src/modules/entitlements/domain/subscriptions.repository.integration.spec.ts
```diff
@@ -0,0 +1,254 @@
...
+        await subscribe(spaceId, status);
+
+        await expect(
+          subscriptionsRepository.getSubscriptionSummary(spaceId),
+        ).resolves.toStrictEqual({
+          hasEverSubscribed: true,
+          activePlanName: null,
+        });
+      },
+    );
```
LucieFaire: we can remove  `should report a terminal subscription as subscribed but on no plan` test and collapse others into one test which will then also cover the other statuses: `past_due, paused, unpaid`:
```
 // Every status outside ACTIVE_SUBSCRIPTION_STATUSES, listed literally
    // rather than derived from it: the point is to pin which side of the line
    // each one falls on, which a list computed from that constant could not do.
    it.each([
      'canceled',
      'incomplete',
      'incomplete_expired',
      'past_due',
      'paused',
      'unpaid',
    ] as const)(
      'should report a %s subscription as subscribed but on no plan',
      async (status) => {
        const spaceId = await createSpace();
        await subscribe(spaceId, status);

        await expect(
          subscriptionsRepository.getSubscriptionSummary(spaceId),
        ).resolves.toStrictEqual({
          hasEverSubscribed: true,
          activePlanName: null,
        });
      },
    );
```

LucieFaire (dismissed):
minor nits and test improv, otherwise good!

src/modules/billing/domain/payment-link-offer.rules.ts
moisses89: Nested ternary violates our biome lint rule https://github.com/safe-global/safe-client-gateway/blob/main/biome.json#L123

src/modules/billing/routes/__tests__/billing.controller.e2e-spec.ts
moisses89: Done 9a85d312594e622b89b5ce6f250f5699a67e5454

src/modules/billing/routes/billing.service.ts
moisses89: Done 9a85d312594e622b89b5ce6f250f5699a67e5454

src/modules/entitlements/domain/subscriptions.repository.integration.spec.ts
moisses89: Done 9a85d312594e622b89b5ce6f250f5699a67e5454

LucieFaire (approved):

PR #3402, opened 2026-08-30
feat(fees): forward the Safenet check flag and pass its fee line through

Fbartoli:
## Summary

The fee service prices a Safenet check only when the caller states the transaction is
checked. The check is a per-transaction user opt-in: the wallet shows a checkbox
(default off) and the choice arrives on the preview request body as `safenetCheck`.
This gateway forwards `true` to the `gtf/fees` request only when the client sent
`true` AND the viewed chain's `SAFENET_CHECKS` feature says the check is available —
the feature is the availability gate, the user makes the decision. Everything else
(absent, `false`, or a chain without the feature) omits the field: fail closed, and
the wire shape stays byte-identical to today's for unchecked previews, so a fee
service predating the flag keeps working.

Previews that disagree on the flag are safe by construction: the fee service encodes
the Safenet fee into `baseGas`, so the two choices produce different `safeTxHash`es
and distinct quotes upstream. Whichever the user signs is what is billed and checked;
no consistency across previews is required.

For consumers: against the fee-service version this pairs with
(safe-global/safe-fee-service#168), `safenetFeeUsd` and `safenetFeeSponsored` are
always present — unchecked quotes report `0`/`true`. Key off a positive
`safenetFeeUsd` (or the chain feature), never off field presence; the fields are
optional in the schema only so a fee service predating the line still parses.

Deploy order: ship the fee service's `safenetCheck` request DTO before any chain
receives the `SAFENET_CHECKS` feature — reversed, every GTF preview on such a chain
gets a hard 400 from the fee service's own request validation.

Verified with scoped unit + integration suites, type-check, and lint, plus a live
forwarding matrix on a local five-service stack (flag on wire only for
client-true + feature-on; all other cells byte-identical to pre-change requests;
quoted = signed = stored hash at sign time). Wallet consumers:
safe-global/safe-wallet-monorepo#8588 sends the flag; #8590 consumes the read-back
route stacked on this branch.

## Changes

- `GtfFeeBreakdownSchema` and `FeePreviewFeeBreakdown` carry optional `safenetFeeUsd`
  and `safenetFeeSponsored`, passed through unchanged.
- `FeesService` computes the forwarded flag from the request body and
  `chain.features`. The gate is read off the chain object the preview already fetches
  for its relayer type: no new upstream call, no new cache entry; it fails exactly
  when the preview already fails and follows a feature flip through the existing
  chain-update invalidation.
- The `gtf/fees` body is listed field by field instead of spread from the route DTO,
  and `GtfFeesRequestSchema` is strict — a stray or mistyped key is an error rather
  than a silent drop.
- The GTF fee-preview cache key includes the flag.

### DIFF
--- a/src/modules/fees/domain/entities/gtf-fees-request.entity.ts
+++ b/src/modules/fees/domain/entities/gtf-fees-request.entity.ts
@@ -9,13 +9,40 @@ import { NumericStringSchema } from '@/validation/entities/schemas/numeric-strin
 
 export type GtfFeesRequest = z.infer<typeof GtfFeesRequestSchema>;
 
-export const GtfFeesRequestSchema = z.object({
-  to: AddressSchema,
-  value: NumericStringSchema,
-  data: HexSchema,
-  operation: z.enum(Operation),
-  numberSignatures: z.number().int().min(1),
-  nonce: NonNegativeNumericStringSchema,
-  gasToken: AddressSchema,
-  origin: z.enum(Origin).optional(),
-});
+/**
+ * Body of the fee service's `gtf/fees` request.
+ *
+ * Strict: this gateway builds the body itself, so an unknown key is a mistake
+ * here rather than an upstream field to tolerate, and the fee service rejects
+ * one anyway.
+ */
+export const GtfFeesRequestSchema = z
+  .object({
+    to: AddressSchema,
+    value: NumericStringSchema,
+    data: HexSchema,
+    operation: z.enum(Operation),
+    numberSignatures: z.number().int().min(1),
+    nonce: NonNegativeNumericStringSchema,
+    gasToken: AddressSchema,
+    origin: z.enum(Origin).optional(),
+    /**
+     * Whether the fee service must price a Safenet check.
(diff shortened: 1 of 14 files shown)


src/modules/fees/domain/entities/gtf-fees-request.entity.ts
```diff
@@ -9,13 +9,40 @@ import { NumericStringSchema } from '@/validation/entities/schemas/numeric-strin
...
+     * Previews that disagree on the flag are safe by construction: the fee
+     * service encodes the Safenet fee into `baseGas`, so the two choices are
+     * different `safeTxHash`es and distinct quotes; whichever the user signs
+     * is what is billed and checked.
+     *
+     * Sent only when true. Two assumptions about the fee service, neither
+     * verifiable from this repo: absent and `false` are the same value to it,
+     * and a fee service that predates the flag rejects a body carrying a field
+     * its own DTO does not declare.
+     */
```
gjeanmart: nitpick: that's a big comment :)

PR #3419, opened 2026-09-09
feat(tokens): expose token metadata by address

tongyaosafe:
- Closes [WA-3516](https://linear.app/safe-global/issue/WA-3516/implement-get-chainschainidtokens)

## Summary

Adds two public read endpoints on the existing `tokens` module so clients can resolve token metadata (symbol, name, decimals, logo, trusted) from the Transaction Service by address, instead of freezing it into their own source:

- `GET /v1/chains/{chainId}/tokens/{address}` — one token; 404 when the Transaction Service does not know it.
- `GET /v1/chains/{chainId}/tokens?addresses=a,b,c` — up to 20 tokens in request order; addresses the Transaction Service does not know are omitted, any other upstream failure propagates.

First consumer is the Space-level spending-limit token selector ([WA-3149](https://linear.app/safe-global/issue/WA-3149/build-the-chain-aware-token-selector), safe-wallet-monorepo #8662): its popular-token list keeps only per-chain addresses and resolves everything else through these routes, so the metadata can never drift from what the assets page already shows.

Two design points worth a reviewer's judgement. The batch route fans out at most 20 parallel single-token lookups; each is already cached per `(chainId, address)` by the existing datasource and covered by the per-chain circuit breaker, so no new caching was added and the fan-out is not chunked, matching the existing Safe-overview batch. Because that amplification sits behind an unauthenticated route, both routes carry a `TokensRateLimitGuard`: 60 requests per 60 seconds per caller per route, tunable via `TOKENS_RATE_LIMIT_MAX` and `TOKENS_RATE_LIMIT_WINDOW_SECONDS` — the first rate-limited GET in this repo, added after `@claude arch-review` flagged the abuse-controls rule.

The 20-address cap stays a named constant rather than a config value: it is a published API contract bound, interpolated into the OpenAPI descriptions, and two deployments advertising different bounds would be worse than one fixed number. The `Batch and chunk upstream calls` rule asks for config there, but its own canonical example (`CoingeckoApi.MAX_BATCH_SIZE`) is a static class constant, and that rule governs chunk sizes for splitting a large input rather than a validation bound that rejects with 422. Flagged for the reviewer rather than changed unilaterally.

## Changes

- `routes/tokens.controller.ts`: the two routes, every `@Param` and `@Query` through `new ValidationPipe(Schema)`, and a Swagger response for each outcome the routes can produce.
- `routes/tokens.service.ts`: the only new logic. De-duplicates addresses case-insensitively, skips an address the Transaction Service returns 404 for, propagates the first non-404 failure, and preserves request order.
- `routes/entities/token-addresses.dto.entity.ts`: parses the comma-separated `addresses` query into `Address[]`; empty or more than 20 entries fail validation with 422.
- `routes/entities/token.dto.entity.ts`: Swagger classes for the Native / ERC-20 / ERC-721 union. Named `*TokenMetadata` because `components.schemas` is keyed by class name across the whole app and the balances module already owns `NativeToken` / `Erc20Token` / `Erc721Token`, which omit `trusted`.
- `routes/guards/tokens-rate-limit.guard.ts`: `RateLimitGuard` subclass reading `tokens.rateLimit.*`.
- `tokens.module.ts`: registers the controller and the route service.
- `src/config/entities/configuration.ts`, `schemas/configuration.schema.ts`, `entities/__tests__/configuration.ts`, `.env.sample.json`: `TOKENS_RATE_LIMIT_MAX` / `TOKENS_RATE_LIMIT_WINDOW_SECONDS` (default 60 / 60), mapped, validated at boot with `.min(1)`, mirrored in the test config and documented.
- Tests: address-schema spec (7), service spec (8), controller integration spec (14, two of them covering the rate limit).

### DIFF
--- a/src/modules/tokens/routes/tokens.controller.ts
+++ b/src/modules/tokens/routes/tokens.controller.ts
@@ -1,0 +1,98 @@
+// SPDX-License-Identifier: FSL-1.1-MIT
+import { Controller, Get, Param, Query } from '@nestjs/common';
+import {
+  ApiExtraModels,
+  ApiNotFoundResponse,
+  ApiOkResponse,
+  ApiOperation,
+  ApiParam,
+  ApiQuery,
+  ApiTags,
+  ApiUnprocessableEntityResponse,
+  getSchemaPath,
+} from '@nestjs/swagger';
+import type { Address } from 'viem';
+import type { Token } from '@/modules/tokens/domain/entities/token.entity';
+import {
+  Erc20TokenMetadata,
+  Erc721TokenMetadata,
+  NativeTokenMetadata,
+} from '@/modules/tokens/routes/entities/token.dto.entity';
+import {
+  MAX_TOKEN_ADDRESSES,
+  type TokenAddresses,
+  TokenAddressesSchema,
+} from '@/modules/tokens/routes/entities/token-addresses.dto.entity';
+import { TokensService } from '@/modules/tokens/routes/tokens.service';
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { NumericStringSchema } from '@/validation/entities/schemas/numeric-string.schema';
+import { ValidationPipe } from '@/validation/pipes/validation.pipe';
+
+const TOKEN_SCHEMA = {
+  oneOf: [
+    { $ref: getSchemaPath(NativeTokenMetadata) },
+    { $ref: getSchemaPath(Erc20TokenMetadata) },
+    { $ref: getSchemaPath(Erc721TokenMetadata) },
+  ],
+};
+
+@ApiTags('tokens')
+@ApiExtraModels(NativeTokenMetadata, Erc20TokenMetadata, Erc721TokenMetadata)
+@Controller({
+  path: '',
+  version: '1',
+})
+export class TokensController {
+  constructor(private readonly tokensService: TokensService) {}
+
+  @ApiOperation({
+    summary: 'Get token metadata',
+    description:
+      'Symbol, name, decimals, logo and trust flag of a token on the given chain, as known by the Transaction Service.',
+  })
+  @ApiParam({ name: 'chainId', type: 'string', example: '1' })
+  @ApiParam({
+    name: 'address',
+    type: 'string',
+    description: 'Token contract address (0x prefixed hex string)',
+  })
+  @ApiOkResponse({ schema: TOKEN_SCHEMA })
+  @ApiNotFoundResponse({
+    description: 'The Transaction Service does not know this token',
+  })
+  @ApiUnprocessableEntityResponse({
+    description: 'Invalid chain ID or address',
+  })
+  @Get('chains/:chainId/tokens/:address')
+  getToken(
+    @Param('chainId', new ValidationPipe(NumericStringSchema)) chainId: string,
+    @Param('address', new ValidationPipe(AddressSchema)) address: Address,
+  ): Promise<Token> {
+    return this.tokensService.getToken({ chainId, address });
+  }
+
+  @ApiOperation({
+    summary: 'Get metadata for several tokens',
+    description: `Metadata for up to ${MAX_TOKEN_ADDRESSES} tokens of one chain, in request order. Addresses the Transaction Service does not know are omitted.`,
+  })
+  @ApiParam({ name: 'chainId', type: 'string', example: '1' })
+  @ApiQuery({
+    name: 'addresses',
+    type: 'string',
+    description: `Comma-separated token contract addresses, at most ${MAX_TOKEN_ADDRESSES}`,
+    example:
+      '0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48,0xdAC17F958D2ee523a2206206994597C13D831ec7',
+  })
+  @ApiOkResponse({ schema: { type: 'array', items: TOKEN_SCHEMA } })
+  @ApiUnprocessableEntityResponse({
+    description: `Invalid chain ID, malformed address, empty list or more than ${MAX_TOKEN_ADDRESSES} addresses`,
+  })
+  @Get('chains/:chainId/tokens')
+  getTokens(
+    @Param('chainId', new ValidationPipe(NumericStringSchema)) chainId: string,
+    @Query('addresses', new ValidationPipe(TokenAddressesSchema))
+    addresses: TokenAddresses,
+  ): Promise<Array<Token>> {
+    return this.tokensService.getTokens({ chainId, addresses });
+  }
+}
--- a/src/modules/tokens/routes/tokens.controller.ts
+++ b/src/modules/tokens/routes/tokens.controller.ts
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: FSL-1.1-MIT
-import { Controller, Get, Param, Query } from '@nestjs/common';
+import { Controller, Get, Param, Query, UseGuards } from '@nestjs/common';
 import {
   ApiExtraModels,
   ApiNotFoundResponse,
@@ -8,10 +8,12 @@ import {
   ApiParam,
   ApiQuery,
   ApiTags,
+  ApiTooManyRequestsResponse,
   ApiUnprocessableEntityResponse,
   getSchemaPath,
 } from '@nestjs/swagger';
 import type { Address } from 'viem';
+import { ChainIdSchema } from '@/modules/chains/domain/entities/schemas/chain-id.schema';
 import type { Token } from '@/modules/tokens/domain/entities/token.entity';
 import {
   Erc20TokenMetadata,
@@ -23,9 +25,9 @@ import {
   type TokenAddresses,
   TokenAddressesSchema,
 } from '@/modules/tokens/routes/entities/token-addresses.dto.entity';
+import { TokensRateLimitGuard } from '@/modules/tokens/routes/guards/tokens-rate-limit.guard';
 import { TokensService } from '@/modules/tokens/routes/tokens.service';
 import { AddressSchema } from '@/validation/entities/schemas/address.schema';
-import { NumericStringSchema } from '@/validation/entities/schemas/numeric-string.schema';
 import { ValidationPipe } from '@/validation/pipes/validation.pipe';
 
 const TOKEN_SCHEMA = {
@@ -37,6 +39,9 @@ const TOKEN_SCHEMA = {
 };
 
 @ApiTags('tokens')
+// Unauthenticated lookups: the batch route fans out one request into up to
+// MAX_TOKEN_ADDRESSES upstream calls, so it carries its own per-caller bound.
+@UseGuards(TokensRateLimitGuard)
 @ApiExtraModels(NativeTokenMetadata, Erc20TokenMetadata, Erc721TokenMetadata)
 @Controller({
   path: '',
@@ -65,12 +70,13 @@ export class TokensController {
   @ApiNotFoundResponse({
     description: 'The Transaction Service does not know this token',
   })
+  @ApiTooManyRequestsResponse({ description: 'Rate limit exceeded' })
   @ApiUnprocessableEntityResponse({
     description: 'Invalid chain ID or address',
   })
   @Get('chains/:chainId/tokens/:address')
   getToken(
-    @Param('chainId', new ValidationPipe(NumericStringSchema)) chainId: string,
+    @Param('chainId', new ValidationPipe(ChainIdSchema)) chainId: string,
     @Param('address', new ValidationPipe(AddressSchema)) address: Address,
   ): Promise<Token> {
     return this.tokensService.getToken({ chainId, address });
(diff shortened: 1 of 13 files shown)


tongyaosafe:
@claude arch-review

tongyaosafe:
> ## Architecture check
> Guides: [ARCHITECTURE.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/ARCHITECTURE.md), [best-practices.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/best-practices.md), [nestjs-patterns.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/nestjs-patterns.md), [api-dtos-and-validation.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/api-dtos-and-validation.md), [security.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/security.md), [module-structure.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/module-structure.md), [caching-and-performance.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/caching-and-performance.md), [testing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/testing.md), [commits-and-prs.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/commits-and-prs.md), [reviewing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/reviewing.md), [remarks.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/remarks.md)
> 
> 1. **[Abuse controls on unauthenticated endpoints](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/security.md#L134)** — [src/modules/tokens/routes/tokens.controller.ts:103](https://github.com/safe-global/safe-client-gateway/blob/d7f4a964cec1f2f38263744ee5a367413561cb84/src/modules/tokens/routes/tokens.controller.ts#L103) — Both new routes are unauthenticated address lookups and neither attaches a `RateLimitGuard` subclass, so a single call to the batch route fans out up to 20 upstream Transaction Service requests with no per-caller bound. **Fix:** add a `RateLimitGuard` subclass under `src/modules/tokens/routes/guards/` with its own `{ max, windowSeconds }` and declare `@UseGuards(TokensRateLimitGuard)` on the controller.
> 2. **[ValidationPipe everywhere](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/api-dtos-and-validation.md#L11)** — [src/modules/tokens/routes/tokens.controller.ts:73](https://github.com/safe-global/safe-client-gateway/blob/d7f4a964cec1f2f38263744ee5a367413561cb84/src/modules/tokens/routes/tokens.controller.ts#L73) — the two new `chainId` params (line 73 and line 105) pipe bare `NumericStringSchema`, which the rule limits to pre-`ChainIdSchema` routes and states is not precedent to extend. **Fix:** on both handlers use `@Param('chainId', new ValidationPipe(ChainIdSchema)) chainId: string,` with `import { ChainIdSchema } from '@/modules/chains/domain/entities/schemas/chain-id.schema';`.
> 3. **[Batch and chunk upstream calls](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/caching-and-performance.md#L61)** — [src/modules/tokens/routes/entities/token-addresses.dto.entity.ts:6](https://github.com/safe-global/safe-client-gateway/blob/d7f4a964cec1f2f38263744ee5a367413561cb84/src/modules/tokens/routes/entities/token-addresses.dto.entity.ts#L6) — the batch bound applied before the upstream fan-out is a code-level literal, where the rule requires that size to be a named, config-driven value. **Fix:** map it in `src/config/entities/configuration.ts` and declare it in `RootConfigurationSchema` (e.g. `tokens.maxAddressesPerRequest` from `TOKENS_MAX_ADDRESSES_PER_REQUEST`), add the entry to `.env.sample.json`, and read it through `IConfigurationService.getOrThrow`.
> 
> **Verdict:** ❌ 3 finding(s)
> 
> Automated check of this repo's [agent guidelines](https://github.com/safe-global/safe-client-gateway/blob/main/AGENTS.md) — Claude applying the committed guides in `docs/agents/`, not its own opinion. Run by [architecture-check.yml](https://github.com/safe-global/safe-client-gateway/blob/main/.github/workflows/architecture-check.yml) on `@claude arch-review`; background in [the CI gate](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/agent-tooling.md#the-ci-gate). Advisory — reply to a finding by its number to discuss it.

Thanks. 1 and 2 were straight fixes; 3 I'd like to push back on.

**1 — Abuse controls on unauthenticated endpoints.** Fixed in 70e6cb61. Added `TokensRateLimitGuard` under `src/modules/tokens/routes/guards/`, a `RateLimitGuard` subclass reading `tokens.rateLimit.max` / `tokens.rateLimit.windowSeconds`, declared with `@UseGuards` at the controller level so both routes share one budget. Defaults are 60 requests per 60 seconds from `TOKENS_RATE_LIMIT_MAX` / `TOKENS_RATE_LIMIT_WINDOW_SECONDS`, declared in `RootConfigurationSchema` with `.min(1)` so a zero budget fails at boot instead of rejecting every caller, mirrored in the test config and documented in `.env.sample.json`. Two integration tests cover the 429.

One thing that turned up while doing it: these are the first rate-limit variables declared in `RootConfigurationSchema`. The existing `AUTH_RATE_LIMIT_*` and `SPACES_*_RATE_LIMIT_*` are mapped in `configuration.ts` but never declared there, so a malformed value on any of them is a runtime surprise rather than a boot failure. Out of scope here, raising it as a separate gap against non-negotiable 7.

**2 — ValidationPipe everywhere.** Fixed in the same commit. Both `chainId` params now pipe `ChainIdSchema`.

**3 — Batch and chunk upstream calls.** Not changed. I don't think the rule reaches this constant, for three reasons.

The rule governs chunk size: splitting one large input across several upstream calls to respect a provider's documented per-request limit. None of that exists on this path. The Transaction Service has no batch token endpoint, so there is nothing to chunk. Each upstream call carries exactly one address, and `MAX_TOKEN_ADDRESSES` is an input cap at the route boundary that rejects with 422. It does bound the fan-out, which is the concern behind the rule's `why` (the unbounded batch in PR `2532`), and that concern is already satisfied by the cap existing.

Making it config-driven would put a number that is part of the published API contract behind an environment variable. It is interpolated into three Swagger descriptions and reaches the generated wallet-monorepo client verbatim, in `packages/store/src/gateway/AUTO_GENERATED/tokens.ts`, whose CI check compares a sha256 of the entire staging spec. An ops-side tweak to the variable would then turn a check red in a different repo with no code change on either side, and the spec would advertise a bound that another deployment does not honour.

The rule's own canonical example is not config-driven either: `CoingeckoApi.MAX_BATCH_SIZE` is a `private static readonly` literal.

What I did take from the finding: 20 is coupled to the rate-limit budget, the two multiply into the worst-case upstream fan-out per caller per window, and nothing in the code said so. f1679bbd records that coupling on the constant and points at the rate limit as the knob to turn, since that one is config-driven and is not part of the contract.

If you'd rather follow the rule literally I'll add the config value; in that case I'd also drop the number from the Swagger descriptions, so the published spec stops varying with the environment.

tongyaosafe:
@claude arch-review

tongyaosafe:
> ## Architecture check
> Guides applied: [ARCHITECTURE.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/ARCHITECTURE.md), [best-practices.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/best-practices.md), [nestjs-patterns.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/nestjs-patterns.md), [api-dtos-and-validation.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/api-dtos-and-validation.md), [security.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/security.md), [module-structure.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/module-structure.md), [caching-and-performance.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/caching-and-performance.md), [configuration-and-flags.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/configuration-and-flags.md), [testing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/testing.md), [commits-and-prs.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/commits-and-prs.md).
> 
> 1. **[PR body](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/commits-and-prs.md#L82)** — PR description — `## Changes` covers only the controller, service, DTOs, module and tests, so `TokensRateLimitGuard` and the two new env vars land undescribed, and `## Summary` says only that "no new caching and no concurrency limiter were added" without stating that a per-caller rate limit was. **Fix:** add to `## Summary`: "The routes are unauthenticated, so both carry `TokensRateLimitGuard` — 60 requests per 60s per caller per route, tunable via `TOKENS_RATE_LIMIT_MAX` / `TOKENS_RATE_LIMIT_WINDOW_SECONDS`." and add two `## Changes` bullets: "- `routes/guards/tokens-rate-limit.guard.ts`: `RateLimitGuard` subclass reading `tokens.rateLimit.*`." and "- `src/config/` + `.env.sample.json`: `TOKENS_RATE_LIMIT_MAX` / `TOKENS_RATE_LIMIT_WINDOW_SECONDS` (default 60/60) in `configuration.ts`, `RootConfigurationSchema` and the test mirror."
> 2. **[Injected dependencies are `private readonly`](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/best-practices.md#L116)** — [src/modules/tokens/routes/guards/tokens-rate-limit.guard.ts:23](https://github.com/safe-global/safe-client-gateway/blob/f1679bbda1bf0191c75ebf3309097af80f4451c1/src/modules/tokens/routes/guards/tokens-rate-limit.guard.ts#L23) — `configurationService` is a public `readonly` parameter property, exposing an injected dependency the class does not deliberately publish (the sibling rate-limit guards share this drift; it is not precedent). **Fix:** `private readonly configurationService: IConfigurationService,`.
> 
> **Verdict:** ❌ 2 finding(s)
> 
> Automated check of this repo's [agent guidelines](https://github.com/safe-global/safe-client-gateway/blob/main/AGENTS.md) — Claude applying the committed guides in `docs/agents/`, not its own opinion. Run by [architecture-check.yml](https://github.com/safe-global/safe-client-gateway/blob/main/.github/workflows/architecture-check.yml) on `@claude arch-review`; background in [the CI gate](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/agent-tooling.md#the-ci-gate). Advisory — reply to a finding by its number to discuss it.

Both fair. One fixed in code, one was my own bookkeeping.

**1 — PR body.** Correct, and the cause is that the body on this PR was a stale draft: I rewrote the description when the guard went in and hadn't pasted it back. Updated now. `## Summary` states that both routes carry `TokensRateLimitGuard` at 60 requests per 60 seconds per caller per route, tunable via `TOKENS_RATE_LIMIT_MAX` / `TOKENS_RATE_LIMIT_WINDOW_SECONDS`, and `## Changes` has a bullet for the guard file and one for the config surface, including the `.min(1)` boot validation and the test mirror.

**2 — Injected dependencies are `private readonly`.** Fixed in 3eee9a0f, though not literally as suggested. `configurationService` is read only inside the constructor to build `rateLimits` and never touched again, so rather than storing it as a `private readonly` field nothing reads, I dropped the parameter property modifier entirely. That leaves all three injected parameters in the same signature shaped alike — `cacheService` and `loggingService` are already plain parameters — and still satisfies the rule's intent, since no injected dependency is exposed. Say the word if you'd rather have the literal `private readonly` and I'll switch it.

I left the three sibling guards (`oidc-auth-rate-limit`, `spaces-creation-rate-limit`, `spaces-address-book-*-rate-limit`) untouched. Same drift, but changing them here would be scope creep; happy to send a follow-up that cleans all four up together.

tongyaosafe:
@claude arch-review

clovisdasilvaneto (approved):

src/modules/tokens/routes/tokens.controller.ts
```diff
@@ -0,0 +1,118 @@
...
+import { AddressSchema } from '@/validation/entities/schemas/address.schema';
+import { ValidationPipe } from '@/validation/pipes/validation.pipe';
+
+const TOKEN_SCHEMA = {
+  oneOf: [
+    { $ref: getSchemaPath(NativeTokenMetadata) },
+    { $ref: getSchemaPath(Erc20TokenMetadata) },
+    { $ref: getSchemaPath(Erc721TokenMetadata) },
+  ],
+};
```
LucieFaire, line 39: ideally this should not leave here

PR #3420, opened 2026-09-09
feat: add subscription plan change endpoints

falvaradorodriguez:
- Closes [PLA-1950](https://linear.app/safe-global/issue/PLA-1950/cgw-subscription-upgradedowngrade-preview-update-update-endpoints)

## Summary
Add the two endpoints that let a workspace move an existing subscription between the plans it is offered — a preview of the prorated cost, and the change itself. Both are space-scoped: the billing customer is the workspace, not the user.

The target plan is resolved to one of the workspace's offered payment links, reusing the existing offer rules, so a plan change cannot bypass the pre-enforcement filtering from PLA-1870. That link matters beyond authorization: the upstream copies its metadata onto the subscription, and the entitlements are derived from that metadata, so a caller-supplied `paymentLinkId` is validated against the offered set rather than trusted — an unchecked pair would bill one plan and entitle the workspace to another. `prorationBehavior` is a server-side constant and is never read from the request.

Authorization is space membership, matching the pre-existing `createCheckoutUrl`: starting a paid subscription is open to any member, so gating only the change would put the stricter check on the weaker half. The PATCH additionally requires a fresh second factor (`ElevationGuard`) and is recorded in the gated table of `elevation.integration.spec.ts`. The sibling billing routes — `checkout-url`, and `session-url`, whose Stripe portal can cancel a subscription — are knowingly not gated yet; that decision is a separate PR.

Cache invalidation lands here: the PATCH drops the customer's cached subscriptions as soon as the upstream confirms the write. It does not resync the materialized entitlements, which follow the upstream's webhook exactly as they do after a checkout.

Declared follow-ups, deliberately not in this PR:
- The two new datasource methods parse inside the datasource instead of returning `Raw<T>`, consistent with the other seven methods of `billing-api.service.ts`. The module has no repository seam to move the parse to, and converting two methods in isolation would leave a partial boundary; the module-wide conversion (a `BillingRepository` under `src/modules/billing/domain/`) is the follow-up.
- `GET .../preview-update` is not declared in either elevation table, matching 11 other space-scoped GETs across four modules that are equally undeclared. Closing that gap, plus a completeness check, is its own ticket.

## Changes
- `src/modules/billing/routes/` — the two handlers, the request DTO, the response DTOs, and the offer/ownership/conflict checks in `BillingService`
- `src/modules/billing/domain/` — `offersPlan()` beside the existing offer rules
- `src/datasources/billing-api/` — `previewSubscriptionUpdate()`, `updateSubscription()` and `clearSubscriptions()`, plus builders and entities
- `src/domain/interfaces/billing-api.interface.ts` — the three new methods
- `src/datasources/network/` — `patch()` on `INetworkService` and `FetchNetworkService`, generalizing the existing `postWithBody`
- `src/datasources/cache/cache.router.ts` — `getBillingSubscriptionsCacheKey()`
- `src/modules/entitlements/routes/subscription-sync.service.ts` — invalidate via the datasource's `clearSubscriptions()` instead of rebuilding the key locally
- `src/modules/spaces/routes/elevation.integration.spec.ts` — the PATCH in `GATED_ROUTES`
- `docs/agents/security.md` — the `buildUrl` example now lists `patch`


falvaradorodriguez:
@claude arch-review

falvaradorodriguez:
@claude review

falvaradorodriguez:
> ## Architecture check
> Guides applied: [ARCHITECTURE.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/ARCHITECTURE.md), [best-practices.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/best-practices.md), [nestjs-patterns.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/nestjs-patterns.md), [api-dtos-and-validation.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/api-dtos-and-validation.md), [security.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/security.md), [module-structure.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/module-structure.md), [caching-and-performance.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/caching-and-performance.md), [testing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/testing.md), [commits-and-prs.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/commits-and-prs.md), [reviewing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/reviewing.md), [remarks.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/remarks.md).
> 

- 1: IBillingApi has 9 data-returning methods and 12 call sites across two modules, none behind a repository; converting only these two buys a half-seam for most of the price of converting all of them. Leaving it as the module-wide deviation it already is: everything here is validated today, what is missing is the compile-time guarantee, and that is worth doing as one deliberate change rather than piecemeal from a feature PR.

- 2: done: the four pre-existing bindings are back to @Param('spaceId'), uuidPipe only on the two new handlers.

falvaradorodriguez:
> **Claude finished Felipe Alvarado (@falvaradorodriguez)'s task in 3m 46s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/34341409902)
>

- 1:  src/domain/interfaces/ holds datasource contracts, not domain entities, and 18 clear* methods already live there across six of them (11 on ITransactionApi alone). clearSubscriptions is the eighteenth.
It also moved knowledge out of the domain rather than in: SubscriptionSyncService used to build the cache key itself and call deleteByKey; now it knows nothing about Redis. A cache helper it depended on instead would put that back.

- 2: The candidate links share the price, that is what makes the case ambiguous, and the quote depends only on the price, so the preview's answer is the same either way. What differs is the metadata copied on write, which is why only the PATCH has to pick one.

- 3: The upstream has no by-id route — only GET :customerId/subscriptions and the preview-update one — so the find is the only way in. Fetching every status is deliberate too: filtering to active/trialing would make a canceled subscription look absent, and the PATCH would answer 404 instead of 409 "not in an updatable state".

- 4: Done ✅  Documented on the getter itself: "The key, not a dir: the status filter is the hash *field*, so deleting this key drops every filter cached for the customer at once. A caller that needs one filter gone on its own cannot use this." The existing cache-service integration test only exercises a single field, so pinning the multi-field contract belongs with `ICacheService`, not here.

- 5: Done ✅  via Pick. Declaring clearSubscriptions: vi.fn() would have added a mock entry nothing calls, which the arch-review had me delete from spacesRepositoryMock in this same PR. The narrowed type needs as unknown as IBillingApi at the two construction sites, as the sync spec does.

- 6: A TODO here would point at the wrong thing: the preview is one of 12 undeclared space-scoped routes across four modules, so the gap is the table rather than this route, and singling this one out would misdescribe it.

- 7: The nesting is the error funnel, not decoration: `parse` takes the unresolved promise and awaits it inside its own `try`, which is where `HttpErrorFactory` turns a network rejection into a `DataSourceError`. Awaiting the call outside would let that escape as a raw `NetworkResponseError`. Every method in the file passes the promise in the same way.

- 8: `patch` mirrors `post` exactly — same optional `data`, same `JSON.stringify(args.data)` — so a body-less call behaves as it always has on POST. The line I touched in security.md is about `buildUrl` and query strings, not bodies, so it makes no claim either way.

src/datasources/billing-api/entities/subscription-update.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+});
+
+export type SubscriptionUpdatePreview = z.infer<
+  typeof SubscriptionUpdatePreviewSchema
+>;
+
+export const SubscriptionUpdatePreviewSchema = z.object({
+  amountDue: z.number(),
+  currency: z.string(),
+  /** Unix seconds, as everywhere else on a subscription. */
```
yagopv: Having only 1 prop with comments is strange unless super useful. I would remove it

src/datasources/billing-api/entities/subscription-update.entity.ts
```diff
@@ -0,0 +1,39 @@
...
+ * an override, and a caller who could pick `none` would get a plan change
+ * nobody is invoiced for.
+ */
+export const DEFAULT_PRORATION_BEHAVIOR = 'always_invoice';
+
+export type PreviewLineItem = z.infer<typeof PreviewLineItemSchema>;
+
+export const PreviewLineItemSchema = z.object({
+  description: z.string(),
+  /** Signed minor units: negative for credit on unused time. */
```
yagopv: Having only 1 prop with comments is strange unless super useful. I would remove it

yagopv (dismissed):

src/datasources/billing-api/entities/subscription-update.entity.ts
falvaradorodriguez: Done https://github.com/safe-global/safe-client-gateway/pull/3420/commits/ecba65e34f9e15ef6ebe93cfcc33528cd02abe50

src/datasources/billing-api/entities/subscription-update.entity.ts
falvaradorodriguez: Done https://github.com/safe-global/safe-client-gateway/pull/3420/commits/ecba65e34f9e15ef6ebe93cfcc33528cd02abe50

src/modules/billing/routes/billing.service.spec.ts
```diff
@@ -28,25 +37,43 @@ import {
...
+    | 'getPlan'
+    | 'getCustomerSessionUrl'
+    | 'getSubscriptionsByCustomerId'
+    | 'listPaymentLinks'
+    | 'createCheckoutSession'
+    | 'getCheckoutSession'
+    | 'previewSubscriptionUpdate'
+    | 'updateSubscription'
+  >
+>;
```
LucieFaire: im not sure why we need this? Why not have `as MockedObject<IBillingApi>` ?

src/modules/billing/routes/billing.service.ts
```diff
@@ -35,6 +49,16 @@ import { ISpacesRepository } from '@/modules/spaces/domain/spaces.repository.int
...
+/**
+ * Statuses whose plan the upstream will move. Not
+ * `ACTIVE_SUBSCRIPTION_STATUSES`, which is frozen in lockstep with the
+ * `UQ_subscriptions_active_space` index — same two values, different reason.
+ */
+const UPDATABLE_SUBSCRIPTION_STATUSES: ReadonlyArray<SubscriptionStatus> = [
+  'active',
+  'trialing',
+];
+
```
LucieFaire: can we move these to a constant file?

src/datasources/billing-api/billing-api.service.ts
```diff
@@ -32,6 +45,10 @@ import {
   NetworkService,
 } from '@/datasources/network/network.service.interface';
 import type { IBillingApi } from '@/domain/interfaces/billing-api.interface';
+import {
+  type ILoggingService,
+  LoggingService,
+} from '@/logging/logging.interface';
 
 @Injectable()
 export class BillingApi implements IBillingApi {
```
LucieFaire, line 42: Hm, I realized that this class violates our [validation boundary rule](https://github.com/safe-global/safe-client-gateway/blob/felipe/pla-1950-cgw-subscription-upgradedowngrade-preview-update-update/docs/agents/api-dtos-and-validation.md#rawt-at-the-boundary) and [layer placement](https://github.com/safe-global/safe-client-gateway/blob/felipe/pla-1950-cgw-subscription-upgradedowngrade-preview-update-update/docs/agents/module-structure.md#layer-placement).

All billing datasource methods parse and return domain objects directly to BillingService, there is no billing repository that should be in charge or parsing

src/modules/billing/routes/billing.service.spec.ts
LucieFaire: then we dont need to do `as unknown as IBillingApi,`

falvaradorodriguez: Agree, I added the “Pick” after Claude's review comment, and it didn’t add anything. Reverted.

src/modules/billing/routes/billing.service.ts
falvaradorodriguez: Yes, makes sense. Moved.

src/datasources/billing-api/billing-api.service.ts
falvaradorodriguez: You are right, I skipped it from the arch-review because it is the whole class, not just the two methods this PR adds, and that felt like a big change to bring in here. But since the arch-review flagged it too and now you, I have just done it.

src/datasources/billing-api/billing-api.service.ts
```diff
@@ -161,24 +143,21 @@ export class BillingApi implements IBillingApi {
...
-        .object({ subscriptions: z.array(SubscriptionSchema) })
-        .transform((body) => body.subscriptions),
       expireTimeSeconds: this.billingExpireTimeSeconds,
     });
   }
 
   listPaymentLinks(
     args: { upstreamCustomerId?: string } = {},
-  ): Promise<Array<PaymentLink>> {
+  ): Promise<Raw<Array<PaymentLink>>> {
```
LucieFaire: is this type correct? should it not be `Raw<{ paymentLinks: Array<PaymentLink> }>` ?

src/modules/billing/domain/billing-repository.module.ts
```diff
@@ -0,0 +1,17 @@
...
+ * The seam every consumer of the billing service goes through, so `IBillingApi`
+ * stays inside the datasource layer and its `Raw<T>` returns are parsed in
+ * exactly one place.
+ */
+@Module({
+  imports: [BillingApiModule],
+  providers: [{ provide: IBillingRepository, useClass: BillingRepository }],
+  exports: [IBillingRepository],
+})
+export class BillingRepositoryModule {}
```
LucieFaire: do we really need a module for the repo? would it not be part of the billing module? or is it used as standalone somewhere else?

src/datasources/billing-api/billing-api.service.ts
```diff
@@ -49,6 +36,8 @@ import {
   type ILoggingService,
   LoggingService,
 } from '@/logging/logging.interface';
+import type { Raw } from '@/validation/entities/raw.entity';
+import { rawify } from '@/validation/entities/raw.entity';
 
 @Injectable()
 export class BillingApi implements IBillingApi {
```
LucieFaire, line 42: are listPlans and getCustomer dead dead methods or just not wired yet?

I checked the response types, and I think there is mismatch between what is declared and the actual one:

| Method | Declared | Actual body | Evidence |
|---|---|---|---|
| `listPlans` | `Raw<Array<Plan>>` | `{ plans: [...] }` | `src/datasources/billing-api/billing-api.service.spec.ts:138` mocks `{ plans }`, with a comment saying the envelope reaches the repository unopened |
| `getCustomer` | `Raw<Customer>` | `{ customer: {...} }` | `src/datasources/billing-api/billing-api.service.spec.ts:216` mocks `{ customer }` |
| `getSubscriptionsByCustomerId` | `Raw<Array<Subscription>>` | `{ subscriptions: [...] }` | `src/modules/billing/domain/billing.repository.ts:37-39` unwraps it via `SubscriptionsSchema` |
| `listPaymentLinks` | `Raw<Array<PaymentLink>>` | `{ paymentLinks: [...] }` | `src/modules/billing/domain/billing.repository.ts:41-43` unwraps it via `PaymentLinksSchema` |

src/modules/billing/domain/billing.repository.ts
```diff
@@ -0,0 +1,124 @@
...
+ * part of reading the response, so it belongs to the parse rather than to the
+ * transport.
+ */
+const SubscriptionsSchema = z
+  .object({ subscriptions: z.array(SubscriptionSchema) })
+  .transform((body) => body.subscriptions);
+
+const PaymentLinksSchema = z
+  .object({ paymentLinks: z.array(PaymentLinkSchema) })
+  .transform((body) => body.paymentLinks);
```
LucieFaire: i think we should move those to entity declarations

src/datasources/billing-api/billing-api.service.ts
falvaradorodriguez: Good catch!! 

This datasource was created without being 100% clear on its intended use, and it wasn't clear whether those methods were actually being used. I deleted them. If they're needed, we can always add them back. 

As for the types, I think that issue is resolved by fixing the other two comments.

src/datasources/billing-api/billing-api.service.ts
falvaradorodriguez: This was wrong 😓 fixed!

src/modules/billing/domain/billing.repository.ts
falvaradorodriguez: Done.

LucieFaire (approved):

PR #3423, opened 2026-09-09
feat(safe): resolve zerion overview support from zerion's chain list

vseehausen:
## Summary

Fixes WA-2984. The `GET /v2/safes` overview decided per chain whether to use Zerion by checking the config service's `PORTFOLIO_ENDPOINT` feature and reading the Zerion network name from the chain's `balancesProvider`. New Zerion-supported chains therefore needed a manual config change, and a wrong name in config made the overview read another chain's slot of the wallet portfolio (Robinhood showed its Ethereum value). The overview now derives support from Zerion's own chain list, gated only by the global `FF_ZERION_ENABLED` kill switch, the same source the portfolio endpoint already uses.

The chain list is fetched once per environment (mainnet, testnet) present in the request instead of once per chain, and the lookup goes through a new `IZerionRepository` so the route service no longer imports a datasource from another module.

## Changes

- `src/modules/safe/routes/v2/safes.v2.service.ts`: replace the per-chain feature-flag check and the config-based chain name with a chainId to Zerion network map resolved via `IZerionRepository`, once per environment; a failed lookup logs a structured warning and falls back to the balances repository as before.
- `src/modules/safe/safe.module.ts`: import `ZerionModule` instead of `FeatureFlagsModule`.
- `src/modules/zerion/domain/zerion.repository.ts`, `zerion.repository.interface.ts`: new repository exposing `getNetworksByChainId(isTestnet)`; registered and exported in `zerion.module.ts`.
- `src/modules/zerion/datasources/zerion-chain-mapping.service.ts`: `getChainIdToNetworkMapping` becomes public; the unknown-chain log drops from `warn` to `debug` since the overview now asks about every chain.
- Tests: the unit spec mocks the repository per environment and adds a failed-lookup case; the integration spec serves Zerion's chain list in every mock, rewrites the two feature-flag tests as "chain not listed" and "chain list request fails", and drops the config-service v2 chain mock branches this endpoint no longer calls.

Follow-ups outside this PR: the positions endpoint still reads `balancesProvider.chainName` first, and the pre-existing direct `IZerionWalletPortfolioApi` injection in the same service could move behind the same repository.

### DIFF
--- a/src/domain/common/entities/log-type.entity.ts
+++ b/src/domain/common/entities/log-type.entity.ts
@@ -41,4 +41,5 @@ export enum LogType {
   TxnExportFetchRequest = 'TRANSACTION_EXPORT_FETCH_REQUEST',
   TxnExportFetchRequestError = 'TRANSACTION_EXPORT_FETCH_REQUEST_ERROR',
   ZerionCacheInvalidated = 'ZERION_CACHE_INVALIDATED',
+  ZerionChainListError = 'ZERION_CHAIN_LIST_ERROR',
 }
--- a/src/modules/safe/routes/v2/__tests__/safes.v2.controller.overview.integration.spec.ts
+++ b/src/modules/safe/routes/v2/__tests__/safes.v2.controller.overview.integration.spec.ts
@@ -4,7 +4,7 @@ import type { Server } from 'node:net';
 import { faker } from '@faker-js/faker';
 import type { INestApplication } from '@nestjs/common';
 import request from 'supertest';
-import { getAddress } from 'viem';
+import { getAddress, numberToHex } from 'viem';
 import type { MockedObject } from 'vitest';
 import {
   initTestApplication,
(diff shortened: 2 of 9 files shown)


vseehausen:
@claude review

vseehausen:
@claude review

vseehausen:
@claude review again

Addressed in 6142e124: distribution keys lower-cased at the schema (1), single pass without the shared sentinel (2), repository method aligned with the datasource name (4), plus the testnet-only unit test and the balances fallback in the Zerion-failure integration test.

Left as is: the `Chain | null` map (3, pre-existing structure), the thin repository (5, follow-up named in the description), the duplicate overview log message (6, pre-existing), and the per-test chain-list mock (7, this spec's convention).

vseehausen:
@claude review again

Fixed in the latest commit: SPDX header on `zerion-wallet-portfolio.entity.ts` (1) and a tuple type for the two environment maps (2). On (3): we trust the Zerion API to return valid data, so no dedupe handling for mixed-case duplicate keys.

LucieFaire (approved):

PR #3424, opened 2026-09-11
feat(siwe): validate SiWe message domain and URI

TimDaub:
## Summary

Bind a SiWe message to the origins it is expected to come from: the message schema accepts a `domain`, and a `uri` whose authority, that is in `AUTH_ALLOWED_SIWE_DOMAINS`. The check sits on the two fields it applies to, so a message bound elsewhere fails to parse before signature verification and the route answers with its usual generic rejection.

The list is required in production (enforced in `RootConfigurationSchema`'s `superRefine`) and optional everywhere else — staging, development and local setups may leave it unset, which skips the check.

Deployment note: production needs `AUTH_ALLOWED_SIWE_DOMAINS` set before this ships, otherwise configuration validation fails at startup.

## Changes

- `src/modules/siwe/domain/entities/siwe-message.entity.ts` — refine `domain` and `uri` against the allow list; `buildSiweMessageSchema` now takes an options object
- `src/modules/siwe/domain/siwe.repository.ts` — read the allow list at construction, alongside the clock skew
- `src/config/entities/configuration.ts`, `src/config/entities/schemas/configuration.schema.ts` — map and validate `AUTH_ALLOWED_SIWE_DOMAINS`; required in production
- `.env.sample.json`, `src/config/entities/__tests__/configuration.ts` — env docs and the faker mirror
- `src/modules/siwe/domain/entities/__tests__/siwe-message.entity.spec.ts`, `src/config/configuration.validator.spec.ts` — cases for the new validation
- `docs/agents/ARCHITECTURE.md` — note the check in the SIWE login description

https://claude.ai/code/session_01YME1QMaVUAS3V6nFnQDc88

### DIFF
--- a/src/config/entities/schemas/configuration.schema.ts
+++ b/src/config/entities/schemas/configuration.schema.ts
@@ -96,16 +96,7 @@ export const RootConfigurationSchema = z
     AUTH_TOKEN: z.string(),
     AUTH_POST_LOGIN_REDIRECT_URI: z.url(),
     AUTH_ALLOWED_REDIRECT_DOMAIN: DomainSchema.optional(),
-    AUTH_ALLOWED_SIWE_DOMAINS: z
-      .string()
-      .refine(
-        (value) =>
-          value
-            .split(',')
-            .every((domain) => DomainSchema.safeParse(domain.trim()).success),
-        { message: 'Must be a comma-separated list of valid domains' },
-      )
-      .optional(),
+    AUTH_ALLOWED_SIWE_DOMAINS: z.string().optional(),
     AUTH0_API_AUDIENCE: z.string().optional(),
     AUTH0_DOMAIN: DomainSchema.optional(),
     AUTH0_CLIENT_ID: z.string().optional(),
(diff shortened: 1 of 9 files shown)


src/config/entities/schemas/configuration.schema.ts
```diff
@@ -96,6 +96,16 @@ export const RootConfigurationSchema = z
     AUTH_TOKEN: z.string(),
     AUTH_POST_LOGIN_REDIRECT_URI: z.url(),
     AUTH_ALLOWED_REDIRECT_DOMAIN: DomainSchema.optional(),
+    AUTH_ALLOWED_SIWE_DOMAINS: z
```
PooyaRaki: Nit:

```ts
AUTH_ALLOWED_SIWE_DOMAINS: z
  .string()
  .transform((value) => value.split(',').map((domain) => domain.trim()))
  .pipe(z.array(DomainSchema))
  .optional(),
```

PooyaRaki (approved):

PR #3437, opened 2026-09-15
feat: sponsor relays against a workspace's plan allowance

falvaradorodriguez:
- Closes [PLA-1829](https://linear.app/safe-global/issue/PLA-1829/backend-make-gas-sponsoring-plan-aware)

## Summary

Make gas sponsoring plan-aware: a workspace can relay transactions against the sponsored-transaction allowance its plan grants, instead of against the per-Safe daily throttle the chain's own relayer applies. 

The new route is `POST /v1/spaces/:spaceId/chains/:chainId/relay`, authenticated and scoped to a workspace, and the existing `POST /v1/chains/:chainId/relay` is untouched: the two coexist, and a client that relays without a workspace keeps today's behaviour. The workspace is a path parameter verified against the caller's membership — it cannot be a JWT claim, since a user belongs to several workspaces and a claim would freeze an authorization decision for the token's lifetime.

It ships inert. `sponsored_transactions` is seeded with `free_enabled = false` and `free_quota = 0`, so until a plan carries `FEATURE_SPONSORED_TRANSACTIONS` in its upstream metadata every workspace resolves to an allowance of none and the route answers 402. The feature deliberately ignores `ENTITLEMENTS_ENFORCEMENT_STARTS_AT` in both directions: that date exists to grandfather behaviour predating enforcement, and relaying at a workspace's expense has none, so the plan alone decides from day one.

~~Usage is admitted before the relay and recorded after it, which is the ordering `DailyLimitRelayer` already uses in production: a submission that never happened costs the workspace nothing, at the cost of concurrent callers being admitted against the same count and together overshooting the allowance. One relay spends one unit, a batch included — what the allowance buys is the submission.~~ Admitting and spending are one transaction: consumeQuota increments the counter and admits against what the row then holds, so two relays cannot pass the same count and overshoot the allowance together, and a refusal rolls its own reservation back. Because the spend is committed before the call, the allowance is given back when the provider comes home without a taskId, since nothing then says the transaction was submitted. A read-only check runs earlier, so an exhausted workspace is refused before anything expensive happens. One relay spends one unit, a batch included: what the allowance buys is the submission.

`SpaceIdPipe`, `@Auth()` and `assertMember` move out of the `routes/` trees they lived in, so reaching them from another module no longer breaks "Cross-module imports via domain only". The pipe and the decorator go to `src/routes/common/`, the guide's home for a cross-cutting HTTP-layer artifact — not to `domain/`, since a `PipeTransform` is HTTP-layer by construction for the same reason a guard is; the assertion helpers, which are plain functions over a repository, go to `spaces/domain/`. Twenty-six import lines change and no logic does.

Three deliberate deviations, none of which this PR is the right place to resolve:

- `SpaceRelayService` injects `WorkspaceRelayer` rather than a repository. `RelayRepository` exists to choose between relayer strategies; this route has one, so a repository would be a single pass-through method. The relayer sits in the same layer as its three siblings.
- `SpaceRelayModule` lives under `src/modules/relay/` instead of its own top-level module, because the controller reuses `Relay`, `RelayDto` and `RelayDtoSchema` from `relay/routes/entities/` — moving it out would create the cross-module `routes/` imports the first bullet is about.
- `WorkspaceRelayer` does not implement `IRelayer`: that contract is what `RelayManager` dispatches over for a chain-configured relayer, this one is chosen by the route, it needs a workspace, and neither of the contract's other two questions has a caller here.

Two behaviours differ from the public route on purpose. A `createSigner` call is relayed there on any chain, because `RelayManager` short-circuits it past the chain's relayer type; here a chain with no relayer configured is refused before the allowance is touched. And `safeTxHash` is forwarded to the provider but not verified against the calldata — only the relay-fee relayer does that, and it is not on this path.

## Changes

- `modules/entitlements/domain`: `incrementUsage` on `ISpaceFeatureUsageRepository`, an atomic `INSERT … ON CONFLICT DO UPDATE` keyed on `UQ_SFU_space_feature_period`; `counter` on `FeatureGrant`, naming the row a quota is measured against; `isEventMeteredFeature` and `hasClosedWindow` rules; `recordUsage` on `IEntitlementEnforcement`.
- `modules/entitlements/routes`: `countFeatureUsage` reads `space_feature_usage` for event-metered features and refuses what it cannot count; a grant whose window has closed is recomputed rather than served from cache; a feature with no static limit skips the enforcement-date grace.
- `modules/relay/domain`: `WorkspaceRelayer`; `RelaySimulationService` extracted out of `RelayFeeRelayer` unchanged, split into the call and the verdict so the fee relayer keeps running it in parallel with the fee service; `resolveTarget` on `LimitAddressesMapper`, which keeps what each branch already knew about the Safe instead of discarding it — `getLimitAddresses` is now a thin wrapper and the three existing relayers are untouched.
- `modules/relay/routes`: the space-scoped controller and service, and a DTO derived from `RelayDtoSchema` without `gasLimit`, which only the no-fee campaign's declared ceiling ever reads.
- `modules/spaces/domain`: `existsInSpace`, and the blind-index match it shares with `delete` extracted so the two cannot drift.
- `routes/common`: `SpaceIdPipe` and the `@Auth()` decorator moved here from their modules' `routes/` trees, with `space-assert.utils` moved to `spaces/domain/`; every importer updated.
- `migrations`: seed `sponsored_transactions`, inert.


falvaradorodriguez:
@claude arch-review

falvaradorodriguez:
> ## Architecture check
> Guides applied: [ARCHITECTURE.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/ARCHITECTURE.md), [module-structure.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/module-structure.md), [api-dtos-and-validation.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/api-dtos-and-validation.md), [security.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/security.md), [nestjs-patterns.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/nestjs-patterns.md), [best-practices.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/best-practices.md), [caching-and-performance.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/caching-and-performance.md), [database-and-migrations.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/database-and-migrations.md), [configuration-and-flags.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/configuration-and-flags.md), [testing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/testing.md), [commits-and-prs.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/commits-and-prs.md), [reviewing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/reviewing.md), [remarks.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/remarks.md).
> 
- 1: ✅ https://github.com/safe-global/safe-client-gateway/pull/3437/commits/4ff1b96c286f5d96a0a733a25e74ee7bb7b127a7

- 2: 👎 Each extraction exists because this change adds a second consumer to code that had one — landed first they'd be a service extracted for its only caller and a mapper returning a field nobody reads. They're behaviour-preserving: both existing specs keep every assertion and still pass. The fair point is revertability, so if that's the concern I'd rather drop the pre-relay simulation from this PR than split the refactors away from the reason they exist.

- 3: ✅ https://github.com/safe-global/safe-client-gateway/pull/3437/commits/fe652e54281189f2cb0ec5ddf36d29485367634b

- 4: ✅ https://github.com/safe-global/safe-client-gateway/pull/3437/commits/fe652e54281189f2cb0ec5ddf36d29485367634b

falvaradorodriguez:
@claude review

falvaradorodriguez:
> **Claude finished Felipe Alvarado (@falvaradorodriguez)'s task in 3m 29s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/34973722255)

> 
> The arch-review already covered the guide-level items (module boundaries, Swagger completeness, the bare `rejects.toThrow` test). Below are the remaining things I'd flag — focused on correctness, observability, and a small tidy-up. Nothing here is blocking.
> 

- 1, 2, 4, 5: Done ✅ https://github.com/safe-global/safe-client-gateway/pull/3437/commits/abc0f045bfdcc2fc59fe37194dae6343b1486a37

- 3: Skipped. The event needs two relays from the same workspace overlapping the few hundred ms between the check and the record, against a quota of tens per month, and it costs one sponsored transaction when it does — a warn with no alert behind it wouldn't change any decision.

- 6: A typo wouldn't compile: the parameter is typed SimulatedRelayer, so only the two members of the union are accepted, and a rename would break the call sites the same way. Referencing the constant would also mean exporting SIMULATED_RELAYERS again, which the arch-review asked me to stop doing as an unused export.

src/modules/relay/domain/relayers/workspace.relayer.ts
```diff
@@ -0,0 +1,180 @@
...
+      await this.assertHoldsSafe({ ...args, safe });
+    }
+
+    await this.entitlementEnforcement.assertWithinQuota({
+      spaceId: args.spaceId,
+      featureKey: SPONSORED_TRANSACTIONS,
+      delta: RELAYS_PER_CALL,
+    });
+
+    // Only a transaction sent to the Safe itself, as on the public route: a
```
moisses89, line 98: I understand the idea but there is a risk of race condition and give extra sponsored transactions.
Maybe another approach would be increment first and if the relay fails discount.

src/modules/entitlements/domain/entitlements.constants.ts
```diff
@@ -21,11 +24,24 @@ export const STOCK_METERED_FEATURES = [
...
 ): feature is T & { key: StockMeteredFeature } {
-  return (STOCK_METERED_FEATURES as ReadonlyArray<string>).includes(
-    feature.key,
+  return isStockMeteredFeatureKey(feature.key);
+}
+
+/** The other half: metered, and counted in `space_feature_usage`. */
+export function isEventMeteredFeature<T extends { key: string; type: string }>(
+  feature: T,
+): boolean {
```
moisses89: Unlike `isStockMeteredFeature` above, this returns a plain `boolean`, not `feature is T & {...}`. 
No bug today (`eventPeriodStart` takes a plain `Feature`), but worth making it a real predicate for consistency

src/modules/relay/routes/space-relay.controller.ts
```diff
@@ -0,0 +1,122 @@
...
+    status: HttpStatus.NOT_IMPLEMENTED,
+    description: "The chain's relayer type is not supported",
+  })
+  @ApiUnprocessableEntityResponse({
+    description:
+      'Request failed schema validation, or the transaction was simulated and would revert (SIMULATION_FAILED) or could not be simulated (INDETERMINATE_SIMULATION)',
+  })
+  @Post()
+  @UseGuards(AuthGuard)
+  @UseFilters(
```
moisses89: maybe we can put this in a common place. 
```
export const SharedRelayExceptionFilters = () =>
  applyDecorators(
    UseFilters(
      RelayDeniedExceptionFilter,
      RelayerNotAvailableExceptionFilter,
      InvalidMultiSendExceptionFilter,
      InvalidTransferExceptionFilter,
      UnofficialMasterCopyExceptionFilter,
      UnofficialMultiSendExceptionFilter,
      UnofficialProxyFactoryExceptionFilter,
      UnofficialSignerFactoryExceptionFilter,
    ),
  );
```

src/modules/relay/domain/relayers/workspace.relayer.ts
```diff
@@ -0,0 +1,180 @@
...
+    // batch is addressed to MultiSend and a recovery to the DelayModifier,
+    // and neither takes the direct call a simulation makes.
+    if (safe !== null && safe === args.to) {
+      await this.assertSimulates({
```
moisses89, line 102: Could we promise.all the last three awaits? looks like there are not dependencies between them.

falvaradorodriguez: They are independent, but the order matters. Each one can reject the request, and they run from the cheapest to the most expensive, so a Safe that does not belong to the workspace, or an exhausted allowance, is refused before we ever call Tenderly. With Promise.all we would pay for that external call on requests we are going to reject, and it does not cancel the other promises when one of them fails. It would also make the 403 and the 402 depend on which check finishes first, so the same request could return either one. In the happy path the simulation takes most of the time anyway, so running them together would save very little.

src/modules/relay/routes/space-relay.controller.ts
falvaradorodriguez: Makes sense. Applied.

src/modules/relay/domain/relayers/workspace.relayer.ts
falvaradorodriguez: I agree, it's worth the change. Applied.

src/modules/entitlements/domain/entitlements.constants.ts
falvaradorodriguez: Done

moisses89 (dismissed):

src/modules/entitlements/routes/entitlements.service.ts
```diff
@@ -176,6 +187,67 @@ export class EntitlementsService implements IEntitlementEnforcement {
...
+      this.admit({ ...args, grant, used: total - args.delta });
+    });
+  }
+
+  public async refundQuota(args: {
+    spaceId: Space['id'];
+    featureKey: Exclude<FeatureKey, StockMeteredFeature>;
+    delta: number;
+  }): Promise<void> {
+    const { counter } = await this.resolveGrant(args);
```
LucieFaire: Could we preserve the counter selected by `consumeQuota()` and use it for the refund? 
Calling `resolveGrant()` again here can select a different billing period if a subscription webhook arrives while the relay request is in flight. For example, consumption increments period A to 1, the webhook advances to period B, and a failed relay then refunds B to -1 while A remains charged. With a quota of 1, that negative usage permits two successful submissions in B. Returning a reservation containing the original `featureId, periodStart, spaceId and delta` would let the refund reverse the correct increment. Maybe makes sense to add test for such a scenario as well.

src/modules/relay/domain/relayers/workspace.relayer.ts
```diff
@@ -0,0 +1,193 @@
...
+        chainId: args.chainId,
+        to: args.to,
+        data: args.data,
+        safeTxHash: args.safeTxHash,
+      });
+    } catch (error) {
+      // No `taskId` came back, so nothing says it was submitted.
+      if (error instanceof DataSourceError) {
+        await this.refundAllowance(args.spaceId);
+      }
```
LucieFaire, line 126: Maybe we should distinguish confirmed provider rejections from failures where the submission outcome is `unknown`? 
`RhinestoneApi` converts timeouts and response-parsing failures into DataSourceError too, following the repository's `One error funnel per layer`.
The provider may have accepted the transaction before the response is lost or cannot be parsed, so receiving no `taskId` does not establish that submission failed. Refunding in that case removes a potentially successful submission from usage and allows the workspace to exceed its allowance.

src/modules/entitlements/domain/entities/feature-grant.entity.ts
```diff
@@ -6,6 +6,18 @@ const FeatureGrantSchema = z.object({
...
+   * feature, whose usage is a live count in another module's table. Cached
+   * like `resetsAt`, so a Free window rolling over on its own anchor can lag
+   * by the TTL.
+   */
+  counter: z
+    .object({
+      featureId: z.number().int(),
+      periodStart: z.coerce.date(),
+    })
+    .nullable(),
```
LucieFaire, line 20: Should we handle existing cached grants before making counter mandatory? The base branch stores only quota and resetsAt under the same Redis key. After deployment, `CachedGrantsSchema.parse()` rejects those entries because nullable still requires the counter property to exist. This makes quota checks—including existing Safe-seat checks—fail until the entry expires .

src/modules/spaces/routes/elevation.integration.spec.ts
```diff
@@ -155,6 +155,15 @@ const UNGATED_ROUTES: Array<Route> = [
     method: 'get',
     path: (id) => `/v1/spaces/${id}/safes`,
   },
+  // Spends an allowance the Workspace already bought; it changes nothing
+  // anyone else can see, and a second factor per transaction would make
+  // sponsored relaying unusable.
```
LucieFaire: im not sure i understand this comment here

src/modules/relay/domain/exception-filters/relay-calldata.exception-filters.ts
```diff
@@ -0,0 +1,30 @@
...
+import { InvalidTransferExceptionFilter } from '@/modules/relay/domain/exception-filters/invalid-transfer.exception-filter';
+import { RelayDeniedExceptionFilter } from '@/modules/relay/domain/exception-filters/relay-denied.exception-filter';
+import { RelayerNotAvailableExceptionFilter } from '@/modules/relay/domain/exception-filters/relayer-not-available.exception-filter';
+import { UnofficialMasterCopyExceptionFilter } from '@/modules/relay/domain/exception-filters/unofficial-master-copy.exception-filter';
+import { UnofficialMultiSendExceptionFilter } from '@/modules/relay/domain/exception-filters/unofficial-multisend.error';
+import { UnofficialProxyFactoryExceptionFilter } from '@/modules/relay/domain/exception-filters/unofficial-proxy-factory.exception-filter';
+import { UnofficialSignerFactoryExceptionFilter } from '@/modules/relay/domain/exception-filters/unofficial-signer-factory.exception-filter';
+
+/**
+ * What any route that relays rejects: calldata we cannot recognise, and
```
LucieFaire: suggestion:
```
/**
 * Exception filters shared by every relay endpoint (`RelayController` and
 * `SpaceRelayController`). They map the errors thrown while the relay service
 * validates the calldata to HTTP responses:
 *
 * - the relayer refused the request or is unavailable,
 * - the calldata is not a recognised Safe transaction (multiSend, transfer),
 * - the calldata targets a contract that is not an official Safe deployment
 *   (master copy, MultiSend, ProxyFactory, signer factory).
 *
 * Errors specific to one endpoint (e.g. rate limits or space quotas) are not
 * included here; each controller adds those with its own `@UseFilters`.
 */
```

src/modules/relay/domain/relay-simulation.service.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Placeholder EOA used as `from` when simulating a relayed transaction.
+ * On-chain the caller is the relay provider's dispatcher; using a non-Safe
+ * sentinel keeps `msg.sender`/`tx.origin` distinct from the Safe so
+ * refund-receiver-zero flows debit the Safe to a third party (as they would in
+ * production).
+ */
+const SIMULATION_SENDER_SENTINEL: Address =
+  '0x000000000000000000000000000000000000dEaD';
```
LucieFaire: could we please move this out to types/constant files to be consistent :)

LucieFaire (commented):
couple of tiny ones

src/modules/entitlements/routes/entitlements.service.ts
falvaradorodriguez: Good catch! Fixed as you suggested. consumeQuota now returns ConsumedQuota { spaceId, period, delta }, and refundQuota takes it, so the refund credits the row that was charged instead of resolving the grant again. Added the test you mentioned.

src/modules/relay/domain/relayers/workspace.relayer.ts
falvaradorodriguez: You are right about the mechanim, but the existing route (/v1/chains/{chainId}/relay) already behaves this way. DailyLimitRelayer increments the count after relayApi.relay returns, so any throw leaves nothing counted, including a 2xx whose body we cannot read. Keeping both routes consistent seems better than making this one stricter, so if we want to close it I would create a ticket and do the refactor for both.

src/modules/entitlements/domain/entities/feature-grant.entity.ts
falvaradorodriguez: Good point. In practice the key should be empty because this ships together, but staging has a past ENTITLEMENTS_ENFORCEMENT_STARTS_AT, so entries in the old shape can be there, and a rolling deploy would keep writing them while the new instances read. Applied.

src/modules/spaces/routes/elevation.integration.spec.ts
falvaradorodriguez: Fair, it was not saying much. Removed.

src/modules/relay/domain/exception-filters/relay-calldata.exception-filters.ts
falvaradorodriguez: Applied!

src/modules/relay/domain/relay-simulation.service.ts
falvaradorodriguez: Moved!

src/modules/relay/domain/relayers/workspace.relayer.ts
LucieFaire: ko yeah makes sense, lets be consistent then

LucieFaire (approved):

PR #3441, opened 2026-09-16
Revert "feat(safe): resolve zerion overview support from zerion's chain list"

LucieFaire:
Reverts safe-global/safe-client-gateway#3423 cause Sepolia safes return `0` balance


vmosharova (dismissed):
👍🏻

vmosharova (approved):
👍🏻

PR #3442, opened 2026-09-17
feat(circuit-breaker): cap in-flight probes while HALF_OPEN

LucieFaire:
## Summary
While HALF_OPEN the circuit breaker admitted every request until enough of them had failed to re-open the circuit. Against a hanging upstream those failures only land when the request times out, so with the 30 s production HTTP timeout a whole timeout's worth of traffic piled up on the dead service once per half-open window.
This is the pattern behind the data-decoder outage on 2026-09-09, where CGW→decoder traffic amplified 17× and gateway p95 latency reached 30 s.

HALF_OPEN now admits at most as many concurrent probes as the number of failures that would re-open the circuit (`ceil(threshold × halfOpenFailureRateThreshold / 100)`, 3 with the defaults) and rejects the rest with the existing 503. Probe slots are released on both `recordSuccess` and `recordFailure` and reset on every state
transition, so late results from a re-opened circuit cannot leak a slot. 

No new configuration: the cap is derived, so one round of probes is always enough to reach a verdict either way and the two half-open settings cannot drift apart.

## Changes
- `src/datasources/circuit-breaker/circuit-breaker.service.ts` — `halfOpenInFlight` counter; `canProceedInHalfOpenState` admits up to `getEffectiveFailureThreshold` probes; slot release in `recordSuccess`/`recordFailure`; reset on transitions; lazy registration from `recordFailure`; 

- `src/datasources/circuit-breaker/circuit-breaker.service.spec.ts` — probe-cap tests; lazy-registration tests; suite moved to `vi.useFakeTimers()` and a faker-backed circuit name
- `src/datasources/network/network.module.ts` — drop the explicit`getOrRegisterCircuit` call
- `src/config/entities/configuration.ts`, `.env.sample.json` — document that the half-open rate also sets the probe cap
- `docs/agents/ARCHITECTURE.md` — one sentence on the half-open probe bound


PooyaRaki (dismissed):

PooyaRaki (approved):

PooyaRaki:
@claude review

PooyaRaki:
@claude arch-review

PR #3443, opened 2026-09-17
feat: update Nest to 12

vseehausen:
Resolves https://linear.app/safe-global/issue/EPD-509/update-nestjs-version-to-12

Moves the whole Nest line to 12. Supersedes #3427 (Nest core alone) and #3433 (swagger alone) — neither could go green on its own, because swagger 12 needs Nest 12 and Nest 12 breaks swagger 11.

### Why the whole line at once
Nest 12 is ESM with a strict exports map, so directory deep imports no longer resolve — `@nestjs/common/decorators` now points at a non-existent `decorators.js`. `Inject` and `ConfigFactory` move to the package roots. Satellite packages jumped straight to `12.0.0` on the Nest 12 release date; that renumbering aligns their majors with the framework, it is not nine majors of change.

### Behaviour change worth reviewing
`platform-fastify` 12 exact-pins fastify `5.12.1`, which disables the numeric `trustProxy` hop count — [GHSA-3m5p-2c4r-xxw2](https://github.com/fastify/fastify/security/advisories/GHSA-3m5p-2c4r-xxw2) / [CVE-2026-16732](https://www.cve.org/CVERecord?id=CVE-2026-16732). A hop count ignores the connecting address, so the guard added by the parent [CVE-2026-3635](https://www.cve.org/CVERecord?id=CVE-2026-3635) never applied to it.

`parseTrustProxy` now rejects hop counts at startup rather than letting them silently fail closed, which would drop `request.ip` to the socket address and key every client onto one rate-limit bucket. **`EXPRESS_TRUST_PROXY` must be a CIDR or preset** (e.g. `loopback, uniquelocal`) — `0` and unset are unaffected. Needs an infrasec check of the deployed values before merge.

### Verification
tsc, biome and 6467/6467 unit tests pass. Integration is at parity with `main` — the residual failures are timestamp/timing flakes that also fail on Nest 11.

### References
- [Nest 12 release](https://github.com/nestjs/nest/releases/tag/v12.0.0)
- [@nestjs/swagger 12 release notes](https://github.com/nestjs/swagger/releases/tag/12.0.0)
- [fastify `trustProxy` docs](https://github.com/fastify/fastify/blob/main/docs/Reference/Server.md)


gjeanmart (dismissed):
Always been a big fan of NestJS doing clean major version and clean upgrade ;)

.env.sample.json
```diff
@@ -721,7 +721,7 @@
   },
   {
     "name": "EXPRESS_TRUST_PROXY",
```
LucieFaire, line 723: maybe we should rename the var as well

LucieFaire: and we should check what we set it in argo, so it works with a new `trustProxy` function

vseehausen: Added new names in cf71fc42. WDYT?

src/config/entities/configuration.spec.ts
```diff
@@ -16,20 +16,48 @@ describe('configuration - express.trustProxy', () => {
...
 
   it('uses the provided value when set', () => {
     process.env[ENV_KEY] = '10.0.0.0/8';
 
-    expect(configuration().express.trustProxy).toBe('10.0.0.0/8');
+    expect(configuration().httpServer.trustProxy).toBe('10.0.0.0/8');
+  });
+});
+
+describe('configuration - renamed env vars', () => {
```
LucieFaire: do we really need this test suite?

vseehausen: I thought its a nice nudge to update env vars. Should I remove?

src/config/entities/configuration.ts
```diff
@@ -3,7 +3,7 @@ import type { RelayRules } from '@/modules/relay/domain/entities/relay.configura
 
 // Custom configuration for the application
 
-export default () => ({
+const configuration = () => ({
```
LucieFaire: why would you rename this one?

src/config/entities/configuration.ts
```diff
@@ -1113,3 +1111,20 @@ const parseRelayRules = (
   );
   return parsed;
 };
+
+// Renamed in the Express->Fastify migration. A stale name would be ignored and
+// silently fall back to the default, which for `trustProxy` drops `request.ip`
+// to the socket address and keys every client onto one rate-limit bucket.
+const RENAMED_ENV_VARS: Record<string, string> = {
```
LucieFaire: i dont think we need this logic of renaming

src/config/entities/configuration.spec.ts
vseehausen: removed in fcc0974b

src/config/entities/configuration.ts
vseehausen: removed in fcc0974b

src/config/entities/configuration.ts
vseehausen: removed in fcc0974b

LucieFaire (approved):

PooyaRaki (approved):

PR #3444, opened 2026-09-17
fix(circuit-breaker): replace deprecated done() callback with Date.now mocking

PooyaRaki:
## Summary

`circuit-breaker.service.spec.ts` had six tests using the callback-style `(done) => {...}` test signature with a real `setTimeout` wait. Vitest 4 throws `done() callback is deprecated, use promise instead` the moment `done()` is called, which surfaced as unhandled errors in CI on an unrelated PR (https://github.com/safe-global/safe-client-gateway/actions/runs/35244325359/job/105280492571?pr=3223) rather than a clean assertion failure.

## Changes

- `src/datasources/circuit-breaker/circuit-breaker.service.spec.ts` — the six `(done) => {...setTimeout(..., ms)...}` tests (HALF_OPEN transition, timeout gating, and consecutive-success/failure tracking) now mock `Date.now` instead of waiting on the real clock, matching the pattern this file already uses in its Rolling Window and Cleanup tests. `CircuitBreakerService` only ever compares against `Date.now()` internally — it has no timers of its own — so there was never a need to actually wait.

No behavior change to `circuit-breaker.service.ts`. Tests now run in ~0.5s instead of ~2.5s and no longer depend on real elapsed time (which is also what made the failure only intermittently reproducible).

## Test plan

- [x] `yarn vitest run src/datasources/circuit-breaker/circuit-breaker.service.spec.ts` — 29/29 passing
- [x] `yarn format` / `yarn biome check` clean


LucieFaire:
Hehe, I have also addressed it too as part of my [CB improvements PR](https://github.com/safe-global/safe-client-gateway/pull/3442), but I chose Vitest fakeTimers instead of Date.now()

PR #3445, opened 2026-09-18
fix: match the active plan's price, not its display name

moisses89:
## Summary
The payment-link offer filter (PLA-2003) had three gaps, all fixed here:

1. **Sibling plans hidden by name.** The filter compared plans by catalog `planName`, but several price tiers share the same name (e.g. "Business" at 5/10/20 seats), so a workspace on one tier had every sibling price hidden too. It now compares the exact Stripe price id via `offersPlan()`.
2. **No seat-capacity check on downgrade.** A workspace could downgrade to a plan with fewer Safe seats than it currently uses. `createCheckoutUrl` and `updateSubscription` now reject that with 409.
3. **Subscription changes open to any member.** `session-url`, `checkout-url`, `preview-update`, and the `PATCH` subscription-change route now require a workspace admin instead of any member; a non-admin caller that previously got a 200 now gets 403. `checkout-url` also requires a fresh second factor.

## Changes
- `src/modules/billing/domain/payment-link-offer.rules.ts` — `activePlanId` replaces `activePlanName`; matches via `offersPlan()` instead of removed `planNameOf()`
- `src/modules/billing/routes/billing.service.ts` — `getOfferEligibility` passes `activePlanId`; `getSessionUrl`/`createCheckoutUrl`/`previewSubscriptionUpdate`/`updateSubscription` require `assertSpaceAdmin` (not `assertSpaceMember`); `createCheckoutUrl`/`updateSubscription` call `assertSafeSeatCapacity`, which logs a warning on a malformed quota
- `src/modules/billing/routes/billing.controller.ts` — checkout-url gains `ElevationGuard`; Swagger 403/409 descriptions updated across all four routes for the admin-only and seat-capacity outcomes
- `src/modules/entitlements/domain/subscriptions.repository.ts(.interface.ts)` — `getSubscriptionSummary` selects `plan_id` instead of `plan_name`
- `src/modules/entitlements/domain/feature-package.mapper.ts` / `entitlements.constants.ts` / `entities/feature.entity.ts` — `parseSafeSeatQuota` (shared `parseQuotaValue` with `mapFeaturePackage`); `SAFE_SEATS_FEATURE_KEY`/`SAFE_SEATS_METADATA_KEY` named constants, no longer derived by array position
- `src/modules/spaces/routes/elevation.integration.spec.ts` — checkout-url added to the gated table
- Specs updated to price ids; regression case for two links sharing a `planName` but different prices; coverage for the seat-capacity and admin-restriction paths across all four routes


moisses89:
@claude review

moisses89:
@claude review

moisses89:
@claude arch-review

moisses89:
> #### 1. `previewSubscriptionUpdate` skips the seat‑capacity check — inconsistent with its doc
> `billing.service.ts:190-223`. The doc says:
> 
> > Rejects for the same reasons as the PATCH, bar one: it does not require an updatable subscription.
> 
> But `updateSubscription` calls `assertSafeSeatCapacity`, and `previewSubscriptionUpdate` does not. So a downgrade preview succeeds and returns a price, then the follow‑up PATCH 409s. The client sees a quotable number but can never commit it. Either add the same check (probably the right choice — it lets the UI disable/annotate the CTA before the user clicks), or update the doc to state seat capacity is intentionally NOT checked on preview and explain why.
> 
> [Fix this →](https://claude.ai/code?q=In%20src%2Fmodules%2Fbilling%2Froutes%2Fbilling.service.ts%2C%20add%20an%20assertSafeSeatCapacity%20call%20to%20previewSubscriptionUpdate%20so%20a%20downgrade%20whose%20target%20plan%20cannot%20fit%20the%20workspace%27s%20current%20Safes%20is%20rejected%20at%20preview%20time%20too%20-%20matching%20the%20behavior%20its%20docstring%20claims.%20Add%20a%20matching%20spec%20under%20describe%28%27previewSubscriptionUpdate%27%29.%20Branch%3A%20moisesf%2Fpla-2003-cgw-payment-link-offer-filter-excludes-sibling-plans-sharing&repo=safe-global/safe-client-gateway)
> 
> #### 2. `assertSafeSeatCapacity` parses the quota twice
> `billing.service.ts:428-447`. `parseSafeSeatQuota(...)` runs in the guard clause, then `hasSeatCapacity(link, used)` inside its body parses the same metadata again. Not a bug — just wasted work and slightly awkward. Cleaner:
> 
> ```ts
> const quota = parseSafeSeatQuota(
>   args.paymentLink.metadata,
>   (m) => this.loggingService.warn(m),
> );
> if (quota === null) return;
> const used = await this.spaceSafesRepository.countBySpaceId(args.spaceId);
> if (used > quota) {
>   throw new ConflictException(
>     "This plan doesn't offer enough Safe seats for the workspace's current Safes",
>   );
> }
> ```
> 
> `hasSeatCapacity` currently has exactly one caller (this method); if you go this route it can be removed from `payment-link-offer.rules.ts` and its unit tests folded away, keeping the rule module focused on offer selection.
> 
> [Fix this →](https://claude.ai/code?q=In%20src%2Fmodules%2Fbilling%2Froutes%2Fbilling.service.ts%2C%20refactor%20assertSafeSeatCapacity%20to%20call%20parseSafeSeatQuota%20exactly%20once%20-%20store%20the%20result%2C%20guard%20on%20null%2C%20then%20compare%20used%20%3E%20quota%20directly%20-%20and%20remove%20hasSeatCapacity%20from%20src%2Fmodules%2Fbilling%2Fdomain%2Fpayment-link-offer.rules.ts%20and%20its%20tests%20in%20payment-link-offer.rules.spec.ts%20since%20it%20would%20be%20unused.%20Branch%3A%20moisesf%2Fpla-2003-cgw-payment-link-offer-filter-excludes-sibling-plans-sharing&repo=safe-global/safe-client-gateway)
> 
> #### 3. Nit: `ApiConflictResponse` on `PATCH .../subscriptions/:subscriptionId` merges four unrelated 409 causes into one prose sentence
> `billing.controller.ts:274-277`. Fine as documentation, but it forces the client to string‑match the message to distinguish them. If you already return distinct message strings server‑side, consider a short bullet list in the description or dedicated error codes to make the outcomes machine‑distinguishable.
> 
> #### 4. Nit: consider narrowing the OpenAPI 403 wording on `updateSubscription`
> `billing.controller.ts:270-272` says `"Not an admin, or a plan this workspace is not offered"` but the service also throws 403 for a `paymentLinkId` that is not in the offered set (`paymentLinkForPlanOrFail` → `ForbiddenException`). Small clarity gain to mention that third case.

1. No split, all three belong to the same Linear ticket.
2. It is not a breaking changes, the endpoints are disabled in production.  
3. Already fixed, quota parsed once.

yagopv (commented):
Hey @moisses89, I’m missing the same admin protection for the session URL that creates a URL to manage the subscription. A member shouldn’t be able to use it

moisses89:
> Hey Moisés (@moisses89), I’m missing the same admin protection for the session URL that creates a URL to manage the subscription. A member shouldn’t be able to use it

pushed last changes.

yagopv (dismissed):

src/modules/entitlements/domain/feature-package.mapper.ts
```diff
@@ -23,6 +33,33 @@ export function hasFeaturePackageMetadata(
...
 
+/**
+ * The plan's Safe seat quota from its `FEATURE_SAFE_SEATS` metadata, ahead of
+ * a subscription change — before a webhook would otherwise materialize it.
+ * `null` covers unlimited and absent alike: neither bounds the seat count, so
+ * neither blocks a caller comparing against it. A malformed value is also
+ * treated as unbounded, but reported through `onWarning` rather than passed
+ * over in silence — unlike `unlimited`, it is not a value anyone intended.
+ */
+export function parseSafeSeatQuota(
```
LucieFaire, line 64: `parseSafeSeatQuota` only reads uppercase `FEATURE_SAFE_SEATS`, while `mapFeaturePackage` normalizes the suffix to lowercase. For valid metadata `{ FEATURE_safe_seats: '2' }`, this check returns `null` and lets a workspace with three Safes purchase the plan, but the webhook materializes a two-seat quota. 

Use the same key normalization and duplicate-key handling as `mapFeaturePackage`, and add a regression test.

src/modules/billing/routes/billing.service.ts
```diff
@@ -98,13 +105,14 @@ export class BillingService {
     return await this.billingRepository.getPlan({ planId });
   }
 
+  /** Admin-only. */
```
LucieFaire: nit: i think we can remove this comment

src/modules/billing/routes/billing.service.ts
```diff
@@ -124,24 +132,37 @@ export class BillingService {
     return await this.listOfferedPaymentLinks(args);
   }
 
+  /**
+   * Starts a checkout for a workspace payment link.
+   *
+   * Admin-only. A fresh second factor is also required, pinned in the gated
+   * table of `elevation.integration.spec.ts`.
+   */
```
LucieFaire, line 138: I did not understand the part about second factor and an int test

src/modules/entitlements/domain/feature-package.mapper.ts
moisses89: Good catch, fixed. https://github.com/safe-global/safe-client-gateway/pull/3445/changes/a28de0d3cdd3cd9001bfa3a06b01949e7a4a55ce

src/modules/billing/routes/billing.service.ts
moisses89: removed https://github.com/safe-global/safe-client-gateway/pull/3445/changes/a28de0d3cdd3cd9001bfa3a06b01949e7a4a55ce

src/modules/billing/routes/billing.service.ts
moisses89: I updated the comment just mentioning the elevation guard. 
https://github.com/safe-global/safe-client-gateway/pull/3445/changes/919cdb06ec66229a90e298ac212aa129cb3d04ce
The previous comment was added to mention that a second factor is necessary too, because the Auth guard, isAdmin assert and Elevation guard are separated, so I could require that is Admin but not the second factor.

LucieFaire (approved):

PR #3446, opened 2026-09-18
feat: count a Safe on several chains as one seat

falvaradorodriguez:
- Closes [PLA-1975](https://linear.app/safe-global/issue/PLA-1975/count-a-safe-on-multiple-chains-as-one-seat)

## Summary

A Safe address takes one seat in a Workspace however many chains it is on, and releases it only when its last chain is removed. Two rows are the same Safe when the address matches, with no owners or threshold check, which is already how the web app groups them. Safes stay stored per chain and users keep adding and removing per chain; only the counting changes.

The seat key is `COALESCE(address_index, address)`, mirroring the two partial unique indexes the encryption migration already created. `address` is non-deterministic KMS ciphertext, so equality goes through the blind index where a row has one. The blind index is computed whenever `ENCRYPTION_INDEX_KEY` is configured, including with encryption disabled, so a rollback keeps every row on the same key. 

Two things in this diff go beyond the ticket and need saying.

**`SafeSeatsGuard` and its `EntitlementGuard` base are deleted.** The guard asked `assertWithinQuota` with `delta: 0`, and the old `fitsWithinQuota` (`used < quota && …`) refused that at the limit, which is the exact case the ticket requires to admit: a new chain for a Safe the Workspace already holds. It cannot be fixed in place. A guard runs before the validation pipe, so it cannot know what the payload consumes, and giving it the addresses would mean reading seat keys outside the transaction's lock, i.e. duplicating the use case behind an answer that can go stale. The 402 still comes from the check inside the write, which is where it was already decided for a batch. A side effect worth having: the guard ran before any membership assertion, so a non-member could probe a Workspace's remaining seats. At the limit it answered 402 with `quota`, `used` and `resetsAt`, and below it a 403. That path now hits `assertAdmin` first.

**`fitsWithinQuota` drops the `used < quota` conjunct**, and it is shared with `sponsored_transactions`. For `delta >= 1` the two expressions are identical, since `used + delta <= quota` already implies `used < quota`, and `workspace.relayer.ts` passes `RELAYS_PER_CALL`, which is never 0. They differ only at `delta: 0`, which nothing produced before this PR except the deleted guard. A Workspace *over* its quota after a downgrade still gets a 402 for a zero-cost add; that is unchanged from before and outside the ticket, which says "at its seat limit".

No migration. The seat key is a new filter target, so `database-and-migrations.md`'s "Index in the same migration" rule was checked against a measurement rather than assumed: on a 1M-row copy of `space_safes`, adding `("space_id", (COALESCE(address_index, address)))` leaves the plan unchanged. The planner keeps choosing the narrower `IDX_SS_space_id`, and the `NOT EXISTS` becomes a Hash Anti Join (one index scan of the space's rows, hashed, probed by the requested keys). Rows per space are bounded by the seat quota itself, so the index would have cost writes for a structure Postgres ignores.

Two client-visible consequences for whoever ships this: the `used` reported for existing Workspaces drops in one step, and `safeCount` changes meaning. PLA-1976 carries the web app side.

## Changes

- `src/modules/spaces/domain/safes/`: `countBySpaceId` becomes `countSeatsBySpaceId`/`countSeatsBySpaceIds` (`COUNT(DISTINCT …)`), plus `countNewSeats` for what a pending insert would add. Both raw reads parsed with Zod
- `src/modules/spaces/routes/safes/`: the seat check moves inside the transaction, under `lockSeats`, and consumes `{ used, delta }`. `SafeSeatsGuard` off the route
- `src/modules/spaces/routes/spaces.service.ts`: `safeCount` from the repository's seat count. The listing stops loading the `safes` relation, and every Safe's ciphertext with it
- `src/modules/entitlements/domain/`: `fitsWithinQuota` as above. `delta` moves from `prepareQuotaCheck` to the check it returns, because what an action consumes now depends on state read under the caller's lock
- `src/modules/entitlements/routes/guards/`: `entitlement.guard.ts`, `safe-seats.guard.ts`, `space-id-param.schema.ts` and the guard spec deleted. Module providers and exports follow
- `docs/agents/module-structure.md`, `docs/agents/nestjs-patterns.md`: both cited the deleted guards as canonical examples
- Tests: three integration cases for the ticket's requirement (one seat across chains, a held Safe's new chain at the limit, a batch charged only for what it adds), and the `delta: 0` rows restored to the `fitsWithinQuota` table


falvaradorodriguez:
@claude arch-review

falvaradorodriguez:
> ## Architecture check
> Applicable guides: [ARCHITECTURE.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/ARCHITECTURE.md), [best-practices.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/best-practices.md), [nestjs-patterns.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/nestjs-patterns.md), [api-dtos-and-validation.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/api-dtos-and-validation.md), [security.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/security.md), [module-structure.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/module-structure.md), [caching-and-performance.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/caching-and-performance.md), [database-and-migrations.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/database-and-migrations.md), [testing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/testing.md), [commits-and-prs.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/commits-and-prs.md), [reviewing.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/reviewing.md), [remarks.md](https://github.com/safe-global/safe-client-gateway/blob/main/docs/agents/remarks.md).
> 
 1.  ✅ Updated!

falvaradorodriguez:
@claude review

falvaradorodriguez:
@claude review

falvaradorodriguez:
> **Claude finished Felipe Alvarado (@falvaradorodriguez)'s task in 1m 43s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/35338515672)
> 
> ### Code review
> Overall a well-scoped change with careful reasoning in the PR description; the seat-key contract, the lock-before-count sequence, and the guard removal are all defensible and well-tested. A few small things worth considering:
> 
- 1: Not acting on this. A comment would only tell you after you'd broken it, and the real fix (a SAFE constant plus template literals in five more places) costs readability against a rename that can't go unnoticed: both queries have direct integration coverage. 

- 2: The Set is load-bearing. NOT EXISTS dedupes the stored rows, not the requested keys: unnest emits one row per array element and COUNT(*) counts them all, so dropping it charges one seat per chain for a batch adding one address across several, the ticket's central case. Removing it turns space-safes.repository.integration.spec.ts:692 red with Expected 1, Received 2.

- 3 & 5: Both fixed by one change: countNewSeats now takes plaintext addresses and derives the seat key itself through SpaceEncryptionService, so there is no parameter left to key wrong. That also drops the PreparedSpaceSafe dependency, which makes its docstring accurate again as written.

- 4: At the limit it does admit, which is what the ticket asks for.

- 6: Leaving it. The test predates this PR (the diff only renames countBySpaceId in it), and the duplication is deliberate: the simplified used + 1 > quota check isolates what it proves, that lockSeats serializes, from the entitlements machinery. Sharing a helper would pull the whole plan stack in, which space-safes.enforcement.integration.spec.ts already covers.

- 7: resolveInvitedByNames is already batched: two fixed queries over a deduped id set (walletsRepository.find with In(userIds), then findEmailsByIds), so no fan-out there. The per-item work is the decryption instead, decryptAddress in the wallet loop and decryptName per visible member, which scales with members rather than spaces. Predates this PR and untouched by it.

- 8: Good point, and it works here: squash commits on main keep their body, so I'll fold the index measurement into the squash body at merge time. It stays in the PR description too, which is where reviewing.md requires it.

LucieFaire (approved):

PR #3451, opened 2026-09-18
feat(policies): fetch proposers in policies api

akshay-ap:
## Summary

Related to https://linear.app/safe-global/issue/WA-3064/implement-get-policiesactive, https://linear.app/safe-global/issue/WA-3129/extend-the-cgw-policy-payloads-for-spending-limit-and-proposer

- Reports proposers on `GET /v1/spaces/{spaceId}/policies/active`, alongside the
spending limits already served, and adds the `types` query parameter that
which of the caller can use to filter by policy-type. A proposer is a Transaction Service delegate:
it may propose a transaction, never sign or execute one.

- `types` is **required** and rejects an empty value

- Both delegates APIs are reported separately, each as its own policy tagged
`version: v2 | v3`.

- The per-Safe reads are sequential, not `Promise.all`. The Transaction Service
rate-limits on concurrency resulting in `429` errors while testing locally. Reading proposers sequentially leads to slower api response as a trade-off.

Open for discussion:
How to handle v2 and v3 delegates together?

## Changes

**Domain**
- `delegate-api-version.entity.ts` - new `DelegateApiVersion` (`v2`/`v3`).
- `active-policy.entity.ts` - `ProposerPolicyData`, unioned into `ActivePolicyData`.
  `delegatedBy` is a list of `{ delegator, label }`: the label is stored per
  `(delegate, delegator)`, so two owners can label one proposer differently.
- `policy-enforcement.entity.ts` - the missing `offChainEnforcement()` factory.

**Routes**
- `mappers/proposer.mapper.ts` - one policy per `(safe, delegates API)`, grouping
  rows by delegate. `enabled` is always `true`.
- `policies.service.ts` - injects both delegate repositories, reads each Safe at
  `SAFE_TRANSACTION_SERVICE_MAX_LIMIT` (the default page would truncate silently).
- `entities/policy.dto.entity.ts` - proposer DTOs, added to `PolicyDataSchema.oneOf`
  and `@ApiExtraModels`.

**Wiring**
- `policies.module.ts` - both delegates repository modules, plus `ProposerMapper`.

**Tests**
- `proposer.mapper.spec.ts` (new), plus service and e2e cases for a Safe with
  proposers, one without, and an unavailable delegates API.

  Author checklist (docs/agents/reviewing.md — delete this comment before submitting):
  - [x] Read the guides the routing table in AGENTS.md maps this diff to
  - [x] One concern per PR; nothing unrelated refactored or reformatted
  - [x] Env var? Declared in configuration.ts AND RootConfigurationSchema, added to
        .env.sample.json, mirrored in __tests__/configuration.ts, `yarn env:validate` clean
  - [x] Migration? Every FK and WHERE-target column indexed in the same migration; `down` reverts
  - [x] New/changed endpoint? Every input through `new ValidationPipe(Schema)`; DTO implements
        z.infer; guard on state-changing or caller-scoped routes
  - [x] Cached read added or TTL raised? Invalidation lands in this PR, or the trade-off is
        stated in this description
  - [x] SPDX license header on every file this PR touched, not only the new ones
  - [x] No debug scripts, seed helpers, or transient logging left in the branch

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


akshay-ap:
@claude review

akshay-ap:
> **Claude finished Akshay (@akshay-ap)'s task in 2m 37s** —— [View job](https://github.com/safe-global/safe-client-gateway/actions/runs/35616176407)
> 
> ### Code Review
> * [x]  Load review tools
> * [x]  Read PR diff and changed files
> * [x]  Analyze against repo conventions
> * [x]  Post review feedback
> 
> Nice work — DTO/domain split is clean, tests cover the interesting branches (per-version reporting, no-op when nothing asked, atomicity on upstream failure). A few things worth considering:
> 
> ### Correctness / UX
> **1. V2 + V3 both emit the same proposers while Queue Service is off — duplicated to clients.** `src/modules/policies/routes/policies.service.ts:204-210` reads both repos, and while `features.safeQueueService=false` (the current default) `DelegatesV3Repository.getDelegates` (`src/modules/delegate/domain/v3/delegates.v3.repository.ts:60-64`) calls the exact same `getDelegatesV2` tx-service endpoint as `DelegatesV2Repository`. Result: two identical policies per Safe tagged `version: v2` and `version: v3`, and two identical HTTP calls per Safe. The integration test `should report the proposers of a safe, once per delegates api` (`policies.integration.spec.ts:423-465`) actually asserts this duplicated payload as the expected shape, which cements the behavior. Since revocation targets whichever store the grant sits in and the two stores are the same today, consider either (a) reading only V2 while `safeQueueService=false` and only V3 when it flips on, or (b) collapsing to one policy when V3 reports the identical set. The PR body already notes this as open for discussion — flagging it because it will land in every Space response.
> 
> **2. `PolicyDataSchema` union has no discriminator.** `src/modules/policies/routes/entities/policy.dto.entity.ts:170-175` — `EnforcementSchema` above it uses `discriminator: { propertyName: 'via' }`, but `PolicyDataSchema` is a plain `oneOf`. `SpendingLimitPolicyDataDto` and `ProposerPolicyDataDto` share no field with the same name-and-value that a client could switch on inside `data`; the discriminator is `ActivePolicyDto.type`, which lives on the envelope. Generated clients (openapi-typescript-codegen, etc.) will emit an ambiguous union. Options: leave a note that the parent's `type` is the discriminator, or lift `type` into each data DTO.
> 
> ### Performance
> **3. Per-Safe reads are fully sequential.** `policies.service.ts:159-177` iterates safes one at a time, and inside each iteration `delegatesByVersion` (line 191-211) waits for V2 then V3 sequentially — that's `1 indexer + N × (getSafe + v2 delegates + v3 delegates)` serial round-trips for a Space with N Safes. The comment names 429s from the tx service as the driver; that's a per-chain concern, so Safes on **different** chains can safely run in parallel, and within one Safe, `safeRepository.getSafe` + delegates hit different tx-service endpoints and needn't chain. A small `p-limit` per host (or `Promise.all` grouped by chain) would cut wall-clock without reintroducing 429s. The datasource cache also softens repeat calls if TTLs are reasonable.
> 
> ### Small things
> **4. `resolveActivePolicies` type check.** `policies.service.ts:140-141` — `types.includes(...)` twice is fine for two types; if the union grows, `new Set(types)` at the top scales better and reads clearer.
> 
> **5. Missing test for a V2 failure.** `policies.service.spec.ts:430-440` covers `v3` rejecting; since the calls are sequential, a `v2` failure never invokes `v3`. Add a symmetric case so a future refactor that swaps order or parallelizes doesn't silently drop the assertion.
> 
> **6. Dead `Address` import.** `src/modules/policies/routes/mappers/proposer.mapper.ts:3` imports `Address` but only uses it inside a type position via `Map<Address, ...>` — verify `import type` (biome usually catches this, worth a `yarn lint` pass).
> 
> **7. Nit — enum shape.** `DelegateApiVersion` (`src/modules/policies/domain/entities/delegate-api-version.entity.ts:11`) follows the const-object pattern, good. The Swagger enum on `ProposerPolicyDataDto.version` uses `Object.values(DelegateApiVersion)` — same pattern used for `PolicyType`, consistent.
> 
> Nothing here is blocking; (1) and (2) are the two worth resolving before merge. • Branch: `feat/policies-04-proposers`
 
Only `1` needs discussion.

TimDaub:
I wired this against the policies table. Every Safe with a proposer comes back twice, once as `v2` and once as `v3`, with identical `proposers`. [The comment on `map` says](https://github.com/safe-global/safe-client-gateway/blob/f66140e5baa9f20cda6d7357430206370c597829/src/modules/policies/routes/mappers/proposer.mapper.ts#L32-L39) this is expected while the Queue Service is off, because both APIs read the same store.

Why are we doing this?

akshay-ap:
> I wired this against the policies table. Every Safe with a proposer comes back twice, once as `v2` and once as `v3`, with identical `proposers`. [The comment on `map` says](https://github.com/safe-global/safe-client-gateway/blob/f66140e5baa9f20cda6d7357430206370c597829/src/modules/policies/routes/mappers/proposer.mapper.ts#L32-L39) this is expected while the Queue Service is off, because both APIs read the same store.
> 
> Why are we doing this?

I was not sure when developing whether we should use v2 or v3 or both for getting proposers. As checked with team, we will use only v3.

src/modules/policies/routes/policies.service.ts
```diff
@@ -106,38 +116,81 @@ export class PoliciesService {
...
     }
 
-    const [state, enabledModules] = await Promise.all([
-      this.policyIndexerRepository.getState({ safes }),
-      Promise.all(safes.map((safe) => this.enabledModules(safe))),
-    ]);
+    // One upstream read at a time, rather than one `Promise.all` covering every
+    // Safe. This is to avoid 429 errors.
+    //
+    const state = spendingLimitsRequested
```
PooyaRaki, line 148: We should create batches with a fixed, configurable size (e.g. 20) and process each batch with `Promise.all`.

src/modules/policies/routes/__tests__/policies.integration.spec.ts
```diff
@@ -141,6 +154,19 @@ describe('Space Policies Controller', () => {
       ) {
         return Promise.resolve({ data: rawify(polygonSafe), status: 200 });
       }
+      // The proposer half. The delegates API lands here while the Queue
+      // Service is switched off, which is what the test configuration sets.
+      if (url.endsWith('/api/v2/delegates/')) {
```
PooyaRaki, line 159: Why do we even need to check v2?

akshay-ap: This is delegate `v2` from transaction service: e.g. https://api.safe.global/tx-service/sep#/delegates/delegates_list_2 and not CGW's delegate api. If `FF_SAFE_QUEUE_SERVICE` is not set, tests call this api.

src/modules/policies/routes/policies.service.ts
akshay-ap: batch by safe address?

PooyaRaki: It doesn’t matter. The comment points out that we’re currently sending one request at a time, but we should use `Promise.all`. To avoid sending too many requests simultaneously, we can split them into batches of 20 and process each batch with `Promise.all`.

TimDaub:
When we show the [policies proposer detail drawer](https://www.figma.com/design/cOOeHQK12YR2SAKYKiNW5S/Workspaces-%E2%80%93-UX-UI-Review?node-id=16544-16734&m=dev), there is a row called "Last updated" that currently isn't returned by the `/policies/active` endpoint, but we still need it

src/modules/policies/routes/policies.service.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3451/commits/f14324633530dc41083b9b67df54d353d6bb653e

akshay-ap:
> When we show the [policies proposer detail drawer](https://www.figma.com/design/cOOeHQK12YR2SAKYKiNW5S/Workspaces-%E2%80%93-UX-UI-Review?node-id=16544-16734&m=dev), there is a row called "Last updated" that currently isn't returned by the `/policies/active` endpoint, but we still need it

The delegates api does not have `Last updated`. see: https://api.safe.global/tx-service/sep#/delegates/delegates_list_2
Queue service has it though: https://api.safe.global/queue#/Delegates/list_delegates_api_v1_delegates_get

We can me it null when using tx service and return non-null when when using queue service.

src/modules/policies/routes/__tests__/policies.service.spec.ts
```diff
@@ -35,6 +43,14 @@ const mockMembersRepository = {
   findOne: vi.fn(),
 } as unknown as MockedObject<IMembersRepository>;
 
+const mockDelegatesV2Repository = {
+  getDelegates: vi.fn(),
+} as unknown as MockedObject<IDelegatesV2Repository>;
```
PooyaRaki: Do we need `as unknown`?

scripts/others/setup-space.ts
```diff
@@ -0,0 +1,731 @@
...
+import { parseArgs } from 'node:util';
+import { DataSource } from 'typeorm';
+import { type Address, type Hex, isAddress } from 'viem';
+import { privateKeyToAccount } from 'viem/accounts';
+import { createSiweMessage } from 'viem/siwe';
+
+// Read the app's own .env before anything reads process.env, so the gateway
+// port and the Postgres credentials match the running instance. Absent file is
+// fine — every value below has a default.
+try {
```
LucieFaire: are you sure we should add this script to the repo? Is it safe to have defaults in plain text?

src/config/entities/configuration.ts
```diff
@@ -830,6 +830,14 @@ export default () => ({
     indexer: {
       baseUri: process.env.POLICY_INDEXER_BASE_URI || 'POLICY_INDEXER_BASE_URI',
     },
+    delegates: {
+      // How many Safes' delegates are read from the Transaction Service at
+      // once.
+      batchSize: Number.parseInt(
+        process.env.POLICY_DELEGATES_BATCH_SIZE ?? `${1}`,
```
LucieFaire: seems like a very small default :) you can also set a staging default so u dont have to override it in Argo

src/modules/policies/domain/entities/active-policy.entity.ts
```diff
@@ -52,13 +52,31 @@ export type SpendingLimitAllowance = {
...
+ */
+export type ProposerPolicyData = {
+  proposers: Array<{
+    proposer: Address;
+    /**
+     * The owners who granted it, each with the label they gave. The label is
+     * stored per `(delegate, delegator)` row and two owners can label the same
+     * proposer differently, so it cannot be flattened to one.
+     */
+    delegatedBy: Array<{ delegator: Address; label: string }>;
```
LucieFaire: are you matching what proposer is delegated by whom or it doesnt matter?

src/modules/policies/routes/__tests__/policies.service.spec.ts
```diff
@@ -35,6 +42,10 @@ const mockMembersRepository = {
   findOne: vi.fn(),
 } as unknown as MockedObject<IMembersRepository>;
 
+const mockDelegatesV3Repository = {
+  getDelegates: vi.fn(),
+} as unknown as MockedObject<IDelegatesV3Repository>;
```
LucieFaire: idk why unknown is always added, it is not needed

src/modules/policies/routes/mappers/__tests__/proposer.mapper.spec.ts
```diff
@@ -0,0 +1,171 @@
...
+    chainId: SEPOLIA,
+    address: getAddress(faker.finance.ethereumAddress()),
+  };
+
+  beforeEach(() => {
+    target = new ProposerMapper();
+  });
+
+  /** A delegate registration on `safe`. */
+  function delegate(): IBuilder<Delegate> {
```
LucieFaire: not sure what you are saving with those :)

src/modules/policies/routes/policies.service.ts
```diff
@@ -106,38 +126,99 @@ export class PoliciesService {
...
+   * is.
+   *
+   * Read at the Queue Service's max page size i.e 100. That limit is lower
+   * than the Transaction Service i.e. 200.
+   */
+  private async delegates(safe: SafeRef): Promise<Array<Delegate>> {
+    const { results } = await this.delegatesV3Repository.getDelegates({
+      chainId: safe.chainId,
+      safeAddress: safe.address,
+      limit: SAFE_QUEUE_SERVICE_MAX_LIMIT,
```
LucieFaire: why do u limit to 100? `getDelegates()` returns a paginated result, but `delegates()` keeps only the first 100 and discards next. Safes with more delegates receive an incomplete proposer policy, potentially missing proposers or additional delegators. is it expected?

src/modules/policies/routes/policies.service.ts
```diff
@@ -106,38 +126,99 @@ export class PoliciesService {
...
     for (const [index, safe] of safes.entries()) {
-      policies.push(
-        ...this.spendingLimitMapper.map({
-          safe,
-          allowances: this.getAllowancesBySafe(state.allowances, safe),
-          enabledModules: enabledModules[index],
-        }),
-      );
+      if (state) {
+        const enabledModules = await this.enabledModules(safe);
```
LucieFaire: So we await `enabledModules()` inside the per-Safe loop. On cache misses, the endpoint accumulates one upstream request’s latency per Safe, although these reads are independent. Why not fetch module lists in bounded concurrent batches as well?

scripts/others/setup-space.ts
akshay-ap: this was not supposed to be here. i will rewrite history and force push.

src/modules/policies/domain/entities/active-policy.entity.ts
akshay-ap: On front-end we are showing `proposer` address and address of `delegator`

LucieFaire: ah i see now, my bad, disregard :)

src/modules/policies/routes/policies.service.ts
akshay-ap: we are not expecting large list to start with. Limiting to `100` because Queue service limits the page size to `100` currently where as transaction service has limit of `200`. So, choosing the least value that will work for both queue service and transaction service. We not caring about pagination in this iteration.

src/modules/policies/routes/policies.service.ts
LucieFaire: yeah that is what i meant, as you already doing it with others, maybe we can do it here too. Also maybe then u can extract the batch loop into a common func and just pass pther funcs to it? Or its an overkill, just thought as you now do it 3 times :) anyway not a  blocker

scripts/others/setup-space.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3451/commits/ecd7eb6a302d2a670c65384f3a1b2c70ac275755

src/modules/policies/routes/policies.service.ts
akshay-ap: here it is https://github.com/safe-global/safe-client-gateway/pull/3451/commits/42cecd93a6cbcd884a9fee739589c2bb20eaed4f

src/modules/policies/routes/mappers/__tests__/proposer.mapper.spec.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3451/commits/7eb3284f7a490876ae4db1b18f7e4b16dcf183d4

src/config/entities/configuration.ts
akshay-ap: as discussed, we keep it `1` as default. Override in argo for testing and leave that value as it is later unchanged.

LucieFaire (dismissed):

src/modules/policies/routes/__tests__/policies.service.spec.ts
akshay-ap: https://github.com/safe-global/safe-client-gateway/pull/3451/commits/e063802af1af4f683587a2de5e41da0b570bc5eb

PR #3453, opened 2026-09-18
fix(docker-compose): upgrade db to postgres 18 and align db-test SSL key ownership

PooyaRaki:
## Summary
- Upgrade `db` from `postgres:14.8-alpine` to `postgres:18.6-alpine3.24`, switching to the single `/var/lib/postgresql` mount postgres 18+ expects (the cluster lives in a version-specific subdirectory, which also lets `pg_upgrade --link` work across major versions).
- postgres 18 refuses an SSL private key that isn't owned by the `postgres` user or root, and bind-mounted files keep the host's uid. Add an entrypoint to `db` that mounts `./db_config/test` read-only, copies the key/cert into place with the right ownership and mode, then hands off to `docker-entrypoint.sh`.
- Apply the same fix to `db-test`, which had the same bind-mount ownership problem: mount `./db_config/test` read-only and use the same copy-then-exec entrypoint instead of mounting the key/cert directly.
- Simplify both entrypoint scripts by dropping the separate `install -d` step, since busybox's `install -D` already creates leading directories.

## Test plan
- [ ] `docker compose up db` starts cleanly on postgres 18 with SSL enabled
- [ ] `docker compose up db-test` starts cleanly with SSL enabled
- [ ] Confirm the app and test suite can connect to both over SSL


vseehausen (changes requested):
nit: Could we remove the comments?

PooyaRaki:
@vseehausen Done
https://github.com/safe-global/safe-client-gateway/pull/3453/commits/94301fc25cec9f826ca46183723576d340664371

